PR #564, opened 2020-12-11
Allowing users to specify a delete policy based on amount of storage used within the ES cluster

ewolinetz:
Currently there is a gap in that the only way to specify a delete policy is based on the age of the logs which can make it difficult for capacity planning and cost for customers when they need to use more storage than they anticipated. This enhancement seeks to outline a mechanism to let them specify their retention based on a disk usage within the cluster instead.

### DIFF
--- a/enhancements/cluster-logging/elasticsearch-percentage-index-management.md
+++ b/enhancements/cluster-logging/elasticsearch-percentage-index-management.md
@@ -1,0 +1,173 @@
+---
+title: elasticsearch-percentage-index-management
+authors:
+  - "@ewolinetz"
+reviewers:
+  - "@jcantrill"
+  - "@bparees"
+  - "@alanconway"
+approvers:
+  - "@jcantrill"
+  - "@bparees"
+  - "@alanconway"
+creation-date: 2020-12-11
+last-updated: 2021-01-18
+status: implementable
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# elasticsearch-percentage-index-management
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Migration plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Currently our index management within OpenShift Elasticsearch is limited to just being time based for rollover and deletion. The goal of this proposal is to introduce and outline extending this to also include amount of storage used. There are a two different ways this can be measured within an ES cluster such as per index and per node -- this proposal will also discuss the differences.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


enhancements/cluster-logging/elasticsearch-percentage-index-management.md
```diff
@@ -0,0 +1,156 @@
...
+We will add a usage field for our phases to use:
+
+```
+// +k8s:openapi-gen=true
+type IndexManagementDeletePhaseSpec struct {
+	// The minimum age of an index before it should be deleted (e.g. 10d)
+	//
+	MinAge TimeUnit `json:"minAge, optional"`
+
+  // The percent disk usage to wait for before deleting indices for this alias (e.g. 75)
```
alanconway, line 98: Which is more convenient here - percentage or absolute size? I'm not sure of the answer.
Could support both by treating "75" or "75%" as a percent, and "75G" as a size.

ewolinetz: it seems that the request was for a percentage and I think that would actually make it more simple for users. If we wanted to support a static size I would suggest we have a field that is specific for that -- like a k8s resource request

jcantrill: Why not make this more explicit and telling what it represents `MaxUsagePercent` or so that it is in agreement with age should it be `MinUsagePercent`?

enhancements/cluster-logging/elasticsearch-percentage-index-management.md
```diff
@@ -0,0 +1,156 @@
...
+
+| release|Description|
+|---|---|
+|5.1| **GA** - Initial release
+
+## Drawbacks
+
+Currently this is not possible in the OpenDistro Index Management plugin and causes us to deviate from what they offer. We would need to ensure that it is added by us prior to us moving to ES7 (and adopting their plugins).
+
+The naming convention "MinAge" when paired with "Usage" could give a false impression -- we may end up deleting indices before they reach the "MinAge" because they are the oldest indices and with "Usage" being reached. This may just need to be framed differently within the docs and field descriptions.
```
jcantrill, line 139: We can frame this in the same context as the ES docs for the criteria by which they rollover an index: its when any of the 3 possible conditions is met

enhancements/cluster-logging/elasticsearch-percentage-index-management.md
ewolinetz: is "PercentUsage" be explicit and clear enough?

ewolinetz: I don't feel theres a need to say "Min" or "Max"... we have a way for the user to tell us this is how much percent usage they want to allow. So more than that and we curate, less than that and we don't trigger.

enhancements/cluster-logging/elasticsearch-percentage-index-management.md
ewolinetz: that's what I convey in ##Design Details and more calling out here that we should make sure it is clear in the docs.

jcantrill (commented):
/lgtm

PR #566, opened 2020-12-15
Separating provider-specific code in the installer

janoszen:
This change introduces an internal API in the installer code to move provider-specific code into a separate folder and avoid having large switch-case structures in the installer code.

### DIFF
--- a/enhancements/separating-provider-specific-code-in-the-installer.md
+++ b/enhancements/separating-provider-specific-code-in-the-installer.md
@@ -1,0 +1,147 @@
+---
+title: separating-provider-specific-code-in-the-installer
+authors:
+  - "@janoszen"
+reviewers:
+  - "@Gal-Zaidman"
+approvers:
+  - TBD
+creation-date: 2020-11-27
+last-updated: 2020-11-27
+status: provisional
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Separating provider-specific code in the installer
+
+## Release Signoff Checklist
+
+TBD
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+
+## Summary
+
+This change introduces an internal API in the installer code to move provider-specific code into a separate folder and avoid having large switch-case structures in the installer code.
+
+## Motivation
+
+Several parts of the installer code have large, [unwieldy switch-case structures](https://github.com/openshift/installer/blob/254680f6268a7b8e57091d788256c67b81755b5d/pkg/asset/cluster/tfvars.go#L166) that contain provider-specific codes. This makes it hard to keep track of the code that serves a provider, and it makes it impossible to write component-level tests. Creating a separated API reduces friction between teams and makes changes easier to merge because the risk of accidental side effects between providers is reduced. 
+
+### Goals
+
+Create a Go interface and the hooks in the installer code that separatees the installer core code from the provider-specific code.
+
+### Non-Goals
+
+Remove all provider-specific pieces from the code. That would require a complete refactor of the install config data structure and move the Terraform code. 
+
+## Proposal
+
+### Step 1: Create a Go interface
+
+We should create a Go interface that describes the functionality each provider must implement. For example, from the prototype:
+
+```go
+type PlatformV2 interface {
+	// Name returns the name of the platform.
+	Name() string
+
+	// Metadata translates the installConfig to the appropriate entry in clusterMetadata.
+	Metadata(clusterMetadata *types.ClusterMetadata, installConfig *installconfig.InstallConfig) error
+
+    //...
+}
+```
+
+Not all platforms implement all features. Optional features should be hidden behind a feature flag and a separate interface:
+
+```go
+var ErrNotAnIPIPlatform = errors.New("...")
+
+type PlatformV2 interface {
+    //...
+
+    // GetIPI returns the IPI-specific interface, or ErrNotAnIPIPlatform if the platform does not support IPI.
+	GetIPI() (IPIPlatformV2, error)
+}
+
+type IPIPlatformV2 interface {
+    // IPI-specific 
+}
+``` 
+
+### Step 2: Implement registry
+
+In this step we should implement a registry pattern where providers can register with their names.
+
+### Step 3: Implement feature-switch
+
+Once the registry is complete the installer core code must be changed. Provider-specific switch-case structures can be identified by searching for the usage of the `Name` constants in the `types` package. The existing switch-case should be wrapped in an if-else case. This will help with keeping the functionality of providers that have not yet switched.
+
+```go
+if v2, err := registry.Get(providerName); err == nil {
+    v2.CallSpecifiedFunction(...)
+} else if (!errors.Is(err, registry.ProviderNotFound)) {
+    return err
+} else {
+    switch (...) {
+        //...
+    }
+}
+```
+
+### Step 4: Migrate providers
+
+Providers need to be migrated to the new API.
+
+### Step 5: Remove fallback code
+
+Once all providers are migrated the fallback code can be completely removed.
+
+### Risks and Mitigations
+
+As with every refactor, this change contains the risk of breaking existing functionality. This can be mitigated by migrating only small portions of the code to the new API at a time.
+
+## Design Details
+
+### Open Questions
+
+1. How should the API look in detail?
+2. How should the implementation be merged? All-at-once, or bit by bit?
+
+### Test Plan
+
+Same as testing of the installer for regular releases.
+
+### Graduation Criteria
+
+TBD
+
+### Version Skew Strategy
+
+As the API develops it presents the problem of version skew. As new API calls are introduced they need to either be added to all providers. If that is not possible a new interface should be introduced and a compatibility wrapper should be added. Once all providers are migrated the temporary interface can be removed.
+
+## Implementation History
+
+Major milestones in the life cycle of a proposal should be tracked in `Implementation
+History`.
+
+## Drawbacks
+
+### Maintenance of the API
+
+The API needs to be maintained and adding/removing API calls is a multi-step process.
+
+### Slower implementation of a new hook
+
+Until now adding hooks to the core installer code has been easy since the code has been added directly to the installer core code. Now a new hook is time-consuming and API design must be considered. 
+
+## Alternatives
+
+Leave the code as it is.


mdbooth:
I'm in favour of this. The details of each 'API' can be hashed out separately one by one as cloud provider code is refactored into it. I don't think it's worth getting into that detail here, except to agree it in principal.

enhancements/separating-provider-specific-code-in-the-installer.md
```diff
@@ -0,0 +1,147 @@
...
+
+Create a Go interface and the hooks in the installer code that separatees the installer core code from the provider-specific code.
+
+### Non-Goals
+
+Remove all provider-specific pieces from the code. That would require a complete refactor of the install config data structure and move the Terraform code. 
+
+## Proposal
+
+### Step 1: Create a Go interface
```
pierreprinetti, line 46: Instead of defining a single all-encompassing interface, each "installer target" (`install-config`, `manifests`, `machines` etc) could define its own required interface, ideally with just one method each. This way, each logic part can be easily tested and we get cleanly segregated interfaces.

enhancements/separating-provider-specific-code-in-the-installer.md
```diff
@@ -0,0 +1,147 @@
...
+}
+
+type IPIPlatformV2 interface {
+    // IPI-specific 
+}
+``` 
+
+### Step 2: Implement registry
+
+In this step we should implement a registry pattern where providers can register with their names.
```
pierreprinetti, line 81: In the OpenStack provider, we have started the effort of defining a clean provider interface for Installer steps. For example, [for `cloudproviderconfig`](
https://github.com/openshift/installer/blob/ec982ecf5bd847add0f3af82b04f32251972701c/pkg/asset/manifests/cloudproviderconfig.go#L97-L105) we have come up with this signature:

```Go
func GenerateCloudProviderConfig(installConfig types.InstallConfig) (cloudProviderConfigData, cloudProviderConfigCABundleData string, err error)
```

We would be happy to be guinea pigs for this RFE and help defining and testing interface signatures in-vivo for the other steps as well! We also have that [in our backlog](https://issues.redhat.com/browse/OSASINFRA-1901).

enhancements/separating-provider-specific-code-in-the-installer.md
```diff
@@ -0,0 +1,147 @@
...
+
+Same as testing of the installer for regular releases.
+
+### Graduation Criteria
+
+TBD
+
+### Version Skew Strategy
+
+As the API develops it presents the problem of version skew. As new API calls are introduced they need to either be added to all providers. If that is not possible a new interface should be introduced and a compatibility wrapper should be added. Once all providers are migrated the temporary interface can be removed.
```
pierreprinetti, line 128: New function signatures could be implemented as new "optional" interfaces a provider object can implement, with the mechanism of interface assertion.

pierreprinetti (commented):
Thank you for proposing this!

We've been working in this direction in the OpenStack provider already, and I think that it would bring many advantages: clarity, more focus on the orchestration in common parts, a better integration of optional steps (eg pre-flight validation), and moreover, well-defined expectations, written in code.

A big :+1: from me!

enhancements/separating-provider-specific-code-in-the-installer.md
```diff
@@ -0,0 +1,147 @@
...
+	Name() string
+
+	// Metadata translates the installConfig to the appropriate entry in clusterMetadata.
+	Metadata(clusterMetadata *types.ClusterMetadata, installConfig *installconfig.InstallConfig) error
+
+    //...
+}
+```
+
+Not all platforms implement all features. Optional features should be hidden behind a feature flag and a separate interface:
```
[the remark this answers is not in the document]
janoszen: Actually, there is a much better way to do this as I found out. We can create a struct that has receivers for the default functionality and providers can embed this default struct. (Basically, OOP inheritance.)

enhancements/separating-provider-specific-code-in-the-installer.md
janoszen: There ought to be a saying about two guinea pigs being better than one, we (oVirt) are also wanting to have this. Let's do it!

enhancements/separating-provider-specific-code-in-the-installer.md
janoszen: See my comment above, we don't even need to do interface assertion. (Interface assertion is not nice to read and incurs somewhat of a performance penalty.)

enhancements/separating-provider-specific-code-in-the-installer.md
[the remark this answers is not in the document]
pierreprinetti: I am not sure I understand. The way I'm figuring this, Installer-core defines a struct with optional methods defined as no-ops, then the Platform embeds it in a struct where it can override the optional noops with actual logic, then the Installer-core imports that... I see a circular dependency

enhancements/separating-provider-specific-code-in-the-installer.md
pierreprinetti: Why wouldn't it be nice? :grin: Conceptually, it's really what we want here: "dear Platform type, do you implement this method?"

As for the performance penalty: I tend to value readability higher, especially when the total penalty is not expected to pass the 1-second threshold in a 40-minutes process. I would even value aesthetics higher :stuck_out_tongue:

enhancements/separating-provider-specific-code-in-the-installer.md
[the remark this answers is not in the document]
dhellmann: In other places where I've created driver-like APIs of this sort, repeating some code always gave us a result that was easier to maintain over time because a developer working on one driver does not have to untangle an inheritance hierarchy or track down the real implementation that may live somewhere else. In places where we could gain from code reuse, utility functions or types that are instantiated rather than inherited are simpler to test and more flexible.

Along the same lines, no-op implementations, even if they have to be provided by every platform, are simpler to understand than features flags and multiple interfaces.

enhancements/separating-provider-specific-code-in-the-installer.md
```diff
@@ -0,0 +1,147 @@
...
+} else {
+    switch (...) {
+        //...
+    }
+}
+```
+
+### Step 4: Migrate providers
+
+Providers need to be migrated to the new API.
```
dhellmann, line 101: One of the biggest sources of friction we have had with "everyone needs to..." projects is prioritizing the work on the teams' backlogs. How many different teams will be involved in this part of the work? Are they all aware of this plan?

janoszen: @dhellmann I have sent out this plan to all teamleads and managers affected as far as I could tell. The benefit of this plan is that no team *needs to* do anything. In addition, I'm working on a way to make sure that API calls added in the future don't create an immediate requirement to do anything for other teams. PoC will still take a while (probably until new year).

enhancements/separating-provider-specific-code-in-the-installer.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Non-Goals
+
+Remove all provider-specific pieces from the code. That would require a complete refactor of the install config data structure and move the Terraform code. 
+
+## Proposal
+
+### Step 1: Create a Go interface
+
+We should create a Go interface that describes the functionality each provider must implement. For example, from the prototype:
```
staebler, line 48: This interface will have 20-30 functions. Some of these functions are going to be difficult to describe if the interface is far from the asset where it is used. As an example, the [MCS Certificate asset](https://github.com/openshift/installer/blob/6abb3b5a8b687ee38b6c96368c77a305f6f0b563/pkg/asset/tls/mcscertkey.go#L48-L69) has different logic for specifying the IP addresses and DNS names for the cert based on the platform. I personally prefer to have the logic for each asset located close to the asset, particularly where the logic is more closely coupled to the asset than the provider.

The logic for some assets that are very specific to a platform would do better farther away from the asset. One example of this is the [Platform Credentials Check asset](https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/platformcredscheck.go). The credentials check could stand on its own.

enhancements/separating-provider-specific-code-in-the-installer.md
staebler, line 128: This point is a big concern for me. Let's say that a provider needs a new parameter. Either we update every provider at once to add the new parameter. Or we support a new function with the new parameter. In the latter case, I would expect the providers that do not need the new parameter will **never** update to the new version of the function. Then, later, when a new parameter is added again, we get a third function. As time goes on, I could see every provider ending up with their own function. I look at the [machine assets](https://github.com/openshift/installer/blob/master/pkg/asset/machine) as an example. Every provider has a MachineSets and Machines function, but each one has its own unique signature.

staebler (commented):
I still do not see enough value in the monolithic interface to outweigh the effort and churn required to implement it. Perhaps with a PoC where I could see the proposed structure more clearly, I could be convinced. With that said, I am whole-heartedly in favor of the owners of specific providers making an effort to better separate their provider-specific code from the core installer code.

janoszen:
@mdbooth @staebler @dhellmann @pierreprinetti thank you for your input. I have a PoC ready that addresses some of the issues raised here: https://github.com/openshift/installer/pull/4517

enhancements/separating-provider-specific-code-in-the-installer.md
janoszen: This issue has been addressed in the PoC (see below) I will update the document later.

patrickdillon: > As time goes on, I could see every provider ending up with their own function. I look at the [machine assets](https://github.com/openshift/installer/blob/master/pkg/asset/machine) as an example. Every provider has a MachineSets and Machines function, but each one has its own unique signature.

The parameters in an interface's function signature to generate an asset should just match the dependencies for that asset. Each platform/provider should have a signature that takes in all the dependencies and then returns whatever is needed to generate the asset (i.e. whatever is coming out of the switch statement). The PoC could be improved in this regard, I will leave a comment there.

enhancements/separating-provider-specific-code-in-the-installer.md
```diff
@@ -0,0 +1,147 @@
...
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+
+## Summary
+
+This change introduces an internal API in the installer code to move provider-specific code into a separate folder and avoid having large switch-case structures in the installer code.
+
+## Motivation
+
+Several parts of the installer code have large, [unwieldy switch-case structures](https://github.com/openshift/installer/blob/254680f6268a7b8e57091d788256c67b81755b5d/pkg/asset/cluster/tfvars.go#L166) that contain provider-specific codes. This makes it hard to keep track of the code that serves a provider, and it makes it impossible to write component-level tests. Creating a separated API reduces friction between teams and makes changes easier to merge because the risk of accidental side effects between providers is reduced.
```
patrickdillon, line 34: > makes it impossible to write component-level tests

What do you mean by "component-level tests"? Many platforms have platform-specific unit tests. For example, here are [the GCP tests for validating install configs](https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/gcp/validation_test.go).

janoszen: We (oVirt) have several parts of our code where half of the functionality is in a switch-case, the other half is in a switch-case. It would be helpful to have tests that run our code only without needing to test the whole installer. That makes test runs quicker and less flaky.

enhancements/separating-provider-specific-code-in-the-installer.md
patrickdillon: > As an example, the [MCS Certificate asset](https://github.com/openshift/installer/blob/6abb3b5a8b687ee38b6c96368c77a305f6f0b563/pkg/asset/tls/mcscertkey.go#L48-L69) has different logic for specifying the IP addresses and DNS names for the cert based on the platform. I personally prefer to have the logic for each asset located close to the asset, particularly where the logic is more closely coupled to the asset than the provider.

This makes sense for readability. A counter-argument in favor of the interface, which is not included in this enhancement, is that the current implementation creates a dependency issue. We can see in the code that creating an asset depends on all platforms, creating an oVirt MCS Certificate asset depends on the AWS platform. If we inject the dependencies through an interface, we would resolve these dependency issues.

These dependencies are "theoretically wrong," but I'm not sure if they pose a practical problem.

enhancements/separating-provider-specific-code-in-the-installer.md
janoszen: You can't really call them unit tests because they still require external API's, which is a separate issue to be addressed.

patrickdillon: If you look at the GCP example I linked above you can see how we mocked out the GCP APIs.

enhancements/separating-provider-specific-code-in-the-installer.md
janoszen: Actually, the machine config is pretty easy to standardize as far as I can tell. In my previous PoC I did this:

```go
	// GetMasterMachines returns a pool of machine configurations to create for the master nodes.
	GetMasterMachines(
		ic *types.InstallConfig,
		pool *types.MachinePool,
		rhcosImage *rhcos.Image,
		clusterID *installconfig.ClusterID,
	) ([]machineapi.Machine, error)

	// AddWorkerMachines adds a set of machines to the worker pool.
	AddWorkerMachines(
		ic *types.InstallConfig,
		pool *types.MachinePool,
		rhcosImage *rhcos.Image,
		clusterID *installconfig.ClusterID,
		machineSets []runtime.Object,
	) error
```

The API extraction is a rather simple process: there are a number of variables available there the provider-specific code is located and those make it into the API.

PR #567, opened 2020-12-15
Added proposal for remediation history

slintes:
We propose to track a limited history of remediations and showing it in the UI in order to allow admins to detect recurring node problems easier.

/cc @beekhof @JoelSpeed @michaelgugino @enxebre

### DIFF
--- a/enhancements/machine-api/remediation-history.md
+++ b/enhancements/machine-api/remediation-history.md
@@ -90,10 +90,10 @@ type Remediation struct {
 	// the time when the unhealthy condition was detected
 	Detected *metav1.Time `json:"detected"`
 
-	// the time when remediation started
+	// the time when remediation started, will be around "detected" plus the "timeout" configured on the condition
 	Started *metav1.Time `json:"started,omitempty"`
 
-	// the time when the node is remediated
+	// the time when the node is remediated and in a safe state, so that workloads can be rescheduled.
 	Remediated *metav1.Time `json:"remediated,omitempty"`
 
 	// the type of remediation, e.g. "internal" or "machineDeletion", and "external" or the
@@ -140,8 +140,10 @@ status:
     remediated: "2020-11-24T17:34:35Z"
     remediationType: external
     started: "2020-11-24T17:34:04Z"
-    targetKind: Node
-    targetName: worker-1
+    target:
+      kind: node
+      name: worker-1
+      uid: [...]
   remediationsAllowed: 2
 ```
 
--- a/enhancements/machine-api/remediation-history.md
+++ b/enhancements/machine-api/remediation-history.md
@@ -9,7 +9,7 @@ approvers:
   - "@michaelgugino"
   - "@enxebre"
 creation-date: 2020-12-15
-last-updated: 2020-12-15
+last-updated: 2021-02-08
 status: implementable
 ---
 
@@ -26,7 +26,7 @@ status: implementable
 
 ## Summary
 
-Record and show limited remediation history in baremetal environments.
+Record and show limited remediation history.
 
 ## Motivation
 
@@ -38,7 +38,6 @@ reboots at 2am on Wednesdays.
 ### Goals
 
 - Record a limited remediation history.
-
 - Show remediation history in the UI.
 
 ### Non-Goals
@@ -53,8 +52,7 @@ machine health checks to contain the last x remediation events, with information
 condition triggered the remediation, and the timestamps of
 - when the triggering condition was detected
 - when remediation was started
-- when the node was succesfully fenced (deleted / powered off)
-- when the node is healthy again.
+- when the node was successfully fenced (deleted / powered off)
 
 This information can be displayed in the UI in a table on a machine healthcheck details page.
 
@@ -73,13 +71,9 @@ type MachineHealthCheckStatus struct {
 
 // Remediation tracks a remediation triggered by this machine health check
 type Remediation struct {
-	// the kind of the remediation target, usually node or machine
+	// a reference to the target machine or node
 	// +kubebuilder:validation:Type=string
-	TargetKind string `json:"targetKind"`
-
-	// the name of the machine or node which is remediated
-	// +kubebuilder:validation:Type=string
-	TargetName string `json:"targetName"`
+	Target *corev1.ObjectReference `json:"target"`
 
 	// the condition type which triggered this remediation
 	// +kubebuilder:validation:Type=string
@@ -99,13 +93,11 @@ type Remediation struct {
 	// the time when remediation started
 	Started *metav1.Time `json:"started,omitempty"`
 
-	// the time when the node is fenced
-	Fenced *metav1.Time `json:"fenced,omitempty"`
+	// the time when the node is remediated
+	Remediated *metav1.Time `json:"remediated,omitempty"`
 
-	// the time when the machine or node is healthy again
-	Finished *metav1.Time `json:"finished,omitempty"`
-
-	// the type of remediation, e.g. "machineDeletion" or "external"
+	// the type of remediation, e.g. "internal" or "machineDeletion", and "external" or the
+	// name of the external remediation template in future
 	// +kubebuilder:validation:Type=string
 	Type string `json:"remediationType,omitempty"`
 }
@@ -145,8 +137,7 @@ status:
   - conditionStatus: Unknown
     conditionType: Ready
     detected: "2020-11-24T17:33:43Z"
-    fenced: "2020-11-24T17:34:35Z"
-    finished: "2020-11-24T17:35:09Z"
+    remediated: "2020-11-24T17:34:35Z"
     remediationType: external
     started: "2020-11-24T17:34:04Z"
     targetKind: Node
@@ -222,4 +213,4 @@ TBD
 
   - Metrics are a good tool for recording e.g. how often remediations are triggered and how long they take in average, but
   not so much for tracking single remediations in the desired detail.
-  - Parsing getting and parsing metrics, and composing a remediation overview from them, is harder on the UI than to just display the MHC status.
+  - Getting and parsing metrics, and composing a remediation overview from them, is harder on the UI than to just display the MHC status.
--- a/enhancements/machine-api/remediation-history.md
+++ b/enhancements/machine-api/remediation-history.md
@@ -163,8 +163,7 @@ conditions. This will also be the place to track remediations:
 
 - in the `needsRemediation` func:
   - for failed machines and unhealthy node conditions: track detection
-  - for missing nodes after remdiation start: track successful fencing
-  - for healthy nodes: track finished remdiation
+  - for missing nodes after remediation start: track successful fencing
 
 - in the `remediate` and `remediationStrategyExternal` funcs: track remediation start
 
--- a/enhancements/machine-api/remediation-history.md
+++ b/enhancements/machine-api/remediation-history.md
@@ -69,17 +69,19 @@ type MachineHealthCheckStatus struct {
 	RemediationHistory []RemediationEvent `json:"remediationHistory,omitempty"`
 }
 
-// Remediation tracks a remediation triggered by this machine health check
+// RemediationEvent tracks a remediation triggered by this machine health check
 type RemediationEvent struct {
-	// a reference to the target machine or node
+	// a reference to the target machine or node, e.g. a machine in failed phase or without a node,
+	// or a node with a matching condition.
+	 
 	// +kubebuilder:validation:Type=string
 	Target *corev1.ObjectReference `json:"target"`
 
-	// the condition type which triggered this remediation
+	// the condition type which triggered this remediation, in case the target is a node
 	// +kubebuilder:validation:Type=string
 	ConditionType *corev1.NodeConditionType `json:"conditionType,omitempty"`
 
-	// the condition status which triggered this remediation
+	// the condition status which triggered this remediation, in case the target is a node
 	// +kubebuilder:validation:Type=string
 	ConditionStatus *corev1.ConditionStatus `json:"conditionStatus,omitempty"`
 
@@ -98,8 +100,8 @@ type RemediationEvent struct {
 
 	// the type of remediation, can be either "machineDeletion" for the built in remediation strategy, or "external".
 	// In future it could be set to the name of the external remediation template.
-	// In case a node recovers after the remedaiation condition was detected but before the remediation started
-	// (according to the timeout declared on the MHC) the type will be set to `aborted`.
+	// In case a node recovers after the remediation condition was detected but before the remediation started
+	// (according to the timeout declared on the MHC) the type will be set to "aborted".
 	// +kubebuilder:validation:Type=string
 	Type string `json:"remediationType,omitempty"`
 }
--- a/enhancements/machine-api/remediation-history.md
+++ b/enhancements/machine-api/remediation-history.md
@@ -1,0 +1,168 @@
+---
+title: remediation-history
+authors:
+  - "@slintes"
+reviewers:
+  - "@beekhof"
+approvers:
+  - "@JoelSpeed"
+  - "@michaelgugino"
+  - "@enxebre"
+creation-date: 2020-12-15
+last-updated: 2020-12-15
+status: implementable
+---
+
+# Remediation history
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Record and show limited remediation history.
+
+## Motivation
+
+With machine health checks administrators can configure automatic remediation of node problems. There is no good
+overview though for when and why nodes were remediated. Recording and showing a history of remediations would allow
+administrators to review when and why machines were remediated, and to discover patterns like e.g. machineA always
+reboots at 2am on Wednesdays.
+
+### Goals
+
+- Record a limited remediation history.
+
+- Show remediation history in the UI.
+
+### Non-Goals
+
+- Record an unlimited history of remediations.
+
+## Proposal
+
+The machine health check holds the configuration for automatic remediation of nodes. We want to extend the status of
+machine health checks to contain the last x remediation events, with information about which node is affected, which
+condition triggered the remediation, and the timestamps of
+- when the triggering condition was detected
+- when remediation was started
+- when the node was succesfully fenced (deleted / powered off)
+- when the node is healthy again.
+
+This information can be displayed in the UI in a table on a machine healthcheck details page.
+
+Example MHC:
+
+```yaml
+apiVersion: machine.openshift.io/v1beta1
+kind: MachineHealthCheck
+metadata:
+  annotations:
+    machine.openshift.io/remediation-strategy: external-baremetal
+  creationTimestamp: "2020-11-24T16:31:22Z"
+spec:
+  maxUnhealthy: 100%
+  nodeStartupTimeout: 60m
+  selector:
+    matchLabels:
+      machine.openshift.io/cluster-api-machine-role: worker
+  unhealthyConditions:
+  - status: "False"
+    timeout: 20s
+    type: Ready
+  - status: Unknown
+    timeout: 20s
+    type: Ready
+status:
+  conditions:
+  - lastTransitionTime: "2020-11-24T17:17:32Z"
+    status: "True"
+    type: RemediationAllowed
+  currentHealthy: 2
+  expectedMachines: 2
+  remediationHistory:
+  - conditionStatus: Unknown
+    conditionType: Ready
+    detected: "2020-11-24T17:33:43Z"
+    fenced: "2020-11-24T17:34:35Z"
+    finished: "2020-11-24T17:35:09Z"
+    remediationType: external
+    started: "2020-11-24T17:34:04Z"
+    targetKind: Node
+    targetName: worker-1
+  remediationsAllowed: 2
+```
+
+Example UI: TBD (copy from "KNI Node & Host Management Designs" doc?)
+
+### User Stories
+
+#### Story 1
+
+As a cluster administrator I want to have an overview over when and why nodes failed and were remediated, in order
+to detect patterns, and to be able to troubleshoot possible root causes.
+
+### Implementation Details/Notes/Constraints [optional]
+
+The machine healthcheck controller [0] is responsible for triggering remediations based on machine health checks and node
+conditions. This will also be the place to track remediations:
+
+- in the `needsRemediation` func:
+  - for failed machines and unhealthy node conditions: track detection
+  - for missing nodes after remdiation start: track successful fencing
+  - for healthy nodes: track finished remdiation
+    
+- in the `remediate` and `remediationStrategyExternal` funcs: track remediation start
+
+[0] https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+
+### Risks and Mitigations
+
+We are not aware of any risks.
+
+## Design Details
+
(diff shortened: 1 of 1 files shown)


enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,168 @@
...
+  remediationHistory:
+  - conditionStatus: Unknown
+    conditionType: Ready
+    detected: "2020-11-24T17:33:43Z"
+    fenced: "2020-11-24T17:34:35Z"
+    finished: "2020-11-24T17:35:09Z"
+    remediationType: external
+    started: "2020-11-24T17:34:04Z"
+    targetKind: Node
+    targetName: worker-1
```
jewzaam, line 98: Has capturing this in a metric been considered?  With this in prometheus we have options for adding alerts, not risking loss of status if a CR is deleted, and retention already managed by monitoring configuration.  Perhaps for alerting on something like machines replaced often it could be inferred by other means.  Just wondering if it is considered.

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+Example UI: TBD (copy from "KNI Node & Host Management Designs" doc?)
+
+### User Stories
+
+#### Story 1
+
+As a cluster administrator I want to have an overview over when and why nodes failed and were remediated, in order
+to detect patterns, and to be able to troubleshoot possible root causes.
```
jewzaam, line 109: The why often has to be collected elsewhere.  MHC for example simply knows a machine was in status Unknown for many types of scenarios.  Correlating to resource consumption may help but not always.  We have a specific example in the OSD support queue right now that we're hitting a dead end tracking a root cause on because the machine is gone, we can't see what actually was happening.

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Implementation Details/Notes/Constraints [optional]
+
+The machine healthcheck controller [0] is responsible for triggering remediations based on machine health checks and node
+conditions. This will also be the place to track remediations:
+
+- in the `needsRemediation` func:
+  - for failed machines and unhealthy node conditions: track detection
+  - for missing nodes after remdiation start: track successful fencing
```
JoelSpeed, line 118: What happens, if for example, fencing happens correctly, but there is an api conflict and as such, the status update is not persisted, can this be re-calculated on the next reconciliation loop?

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,168 @@
...
+TBD
+
+## Drawbacks
+
+## Alternatives
+
+- Using events
+  
+  The UI already displays events (in its own tab), but they aren't stored for very long, which will make it harder to see which nodes
+  are having recurring issues.
```
JoelSpeed, line 162: What is not very long, can we quantify this at all?

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+- Using metrics
+
+  Metrics are a good tool for recording e.g. how often remediations are triggered and how long they take in average, but
+  not so much for tracking single remediations in the desired detail.
```
JoelSpeed, line 168: How much detail is required? Would it spread across many labels? Perhaps this could be expanded to talk about the different cardinalities that would come from this to help explain why this isn't suitable

enhancements/machine-api/remediation-history.md
slintes: Good question. I think it might be hard to catch all intermediate states in this case (e.g. we record that a condition is unhealthy, but fail to update the remediation start). However, we should always be able to do a "cleanup" when the node is back to healthy,  by triggering reconcile until the status update succeeds. Good enough?

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,168 @@
...
+    type: RemediationAllowed
+  currentHealthy: 2
+  expectedMachines: 2
+  remediationHistory:
+  - conditionStatus: Unknown
+    conditionType: Ready
+    detected: "2020-11-24T17:33:43Z"
+    fenced: "2020-11-24T17:34:35Z"
+    finished: "2020-11-24T17:35:09Z"
+    remediationType: external
```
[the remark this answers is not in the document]
slintes: Not sure what you mean with options here? It is just an indicator that external remediation was used.

enhancements/machine-api/remediation-history.md
slintes: done

enhancements/machine-api/remediation-history.md
slintes: Even when we manage to track the details we want to have as metrics, it still leaves the argument (which I just added ;) ) that parsing metrics and composing a remediation history in the UI will be way harder than by "just" displaying the MHC status.

slintes (commented):
@jewzaam @JoelSpeed Thanks for your review! Please check my comments / improvements.

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,168 @@
...
+  - lastTransitionTime: "2020-11-24T17:17:32Z"
+    status: "True"
+    type: RemediationAllowed
+  currentHealthy: 2
+  expectedMachines: 2
+  remediationHistory:
+  - conditionStatus: Unknown
+    conditionType: Ready
+    detected: "2020-11-24T17:33:43Z"
+    fenced: "2020-11-24T17:34:35Z"
```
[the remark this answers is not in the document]
JoelSpeed: There is a difference if you look at it from a baremetal perspective yes. But on a cloud provider, the only action we take is to delete the Machine. The Machine is then drained and terminated and we get a _new_ Machine. There's no real sense of fencing in MHC for the cloud case (or even OSP/vSphere). I would also define remediated in this case as MHC has completed its task, which for us is just deletion. We can't say the Node goes healthy again because the node dies.

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
JoelSpeed: So what would the value be for this if external wasn't used? This seems like an enum to me, with a set of predefined acceptable values, so I'd like to see a list of the acceptable values defined somewhere (I suspect this list has length 2)

enhancements/machine-api/remediation-history.md
JoelSpeed: I guess the question is, is it bad if the remediation gets restarted here? If there's no prior record and the MHC sees the Node is still unhealthy, will it re-fence/remediate it?

enhancements/machine-api/remediation-history.md
JoelSpeed: > that parsing metrics and composing a remediation history in the UI will be way harder than by "just" displaying the MHC status.

This is a much better argument against using metrics 😉

enhancements/machine-api/remediation-history.md
```diff
@@ -26,7 +26,7 @@ status: implementable
 
 ## Summary
 
-Record and show limited remediation history.
+Record and show limited remediation history in baremetal environments.
```
JoelSpeed, line 29: I see this update 👀  Does this mean it won't be recorded in non baremetal environments?

enhancements/machine-api/remediation-history.md
```diff
@@ -158,11 +213,13 @@ TBD
...
+    defaults to 3 hours in OpenShift), which makes it hard to see which nodes are having recurring issues over a longer duration.
+  - Parsing events and composing a remediation overview is harder on the UI than to just display the MHC status.
 
 
 - Using metrics
 
-  Metrics are a good tool for recording e.g. how often remediations are triggered and how long they take in average, but
+  - Metrics are a good tool for recording e.g. how often remediations are triggered and how long they take in average, but
   not so much for tracking single remediations in the desired detail.
+  - Parsing getting and parsing metrics, and composing a remediation overview from them, is harder on the UI than to just display the MHC status.
```
JoelSpeed, line 225: I think there's a typo at the beginning of this sentence, might wanna double check it

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,225 @@
...
+
+// Remediation tracks a remediation triggered by this machine health check
+type Remediation struct {
+	// the kind of the remediation target, usually node or machine
+	// +kubebuilder:validation:Type=string
+	TargetKind string `json:"targetKind"`
+
+	// the name of the machine or node which is remediated
+	// +kubebuilder:validation:Type=string
+	TargetName string `json:"targetName"`
```
dhellmann, line 82: The `TargetKind` and `TargetName` fields could be combined into one of the object reference types instead of using separate fields (there are a few types, I'm not sure which is most correct). That would make it easier to support remediation outside of the default namespace, which might be useful in a multi-cluster setup.

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
dhellmann: If the point of the `fenced` field is to track when it's in a safe state and that term is only used for metal, maybe `safe` is a better name?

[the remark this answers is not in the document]
n1r1: In my perspective, `fenced` (not sure if it's the best terminology) is when the cluster can safely reschedule the workloads on other node.
From an admin point of view, `fenced` is the time he should expect that the cluster will start recovery (rescheduling) of the workload.

IMO, it also makes sense for cloud env as well. 
Practically, the workload can be scheduled elsewhere when the node got drained or if it's unreachable - when the node is deleted.
`remediatied` (better than `finished` IMO), should be the time where the compute resource is back and ready for running workload. In cloud environment, it would be the time after a new replacement node has joined the cluster.

> There's no real sense of fencing in MHC for the cloud case (or even OSP/vSphere). 

After machine deletion, the machine controller deletes the compute resource from the provider. This is when fencing practically happens in cloud env.

> I would also define remediated in this case as MHC has completed its task, which for us is just deletion. We can't say the Node goes healthy again because the node dies.

True, but from an admin perspective, I guess that `remediated` is when the compute resource is back, i.e. when the node recovered, or when a new replacement node joined the cluster.

> I guess the point here is that for cloud, fenced and remediated are the same, all we do is delete.

As I mentioned above, I'd suggest that `fenced` would be the time the *node* was deleted (or successfully drained) from the cluster, and `remediated` would be the time a new/fixed node rejoins the cluster.

Maybe `fenced` and `remediated` should be renamed to reflect that.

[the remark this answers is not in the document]
JoelSpeed: > In cloud environment, it would be the time after a new replacement node has joined the cluster.

We don't track this in MHC and honestly, I don't think we want to. MHC gives up on a machine (in cloud at least) once it's deleted. We have no replacement concept so there would be no way to say a new Node replaced an old broken one.

> After machine deletion, the machine controller deletes the compute resource from the provider. This is when fencing practically happens in cloud env.

So to be correct, should the MHC track the Machine and look for the Deleting phase? I'm concerned this effort may over complicate the MHC. 

> True, but from an admin perspective, I guess that remediated is when the compute resource is back, i.e. when the node recovered, or when a new replacement node joined the cluster.

This isn't something that we track and I'm not sure we can track in a cloud environment, there is no concept of a 1:1 replacement so the only thing that I could see making sense would be when the Machine no longer exists?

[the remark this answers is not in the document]
n1r1: Yeah, I agree. 
I was mainly taking the customer point of view and less on how things can be implemented.
Not sure if this is worth the effort and complication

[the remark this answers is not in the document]
dhellmann: If the cloud implementation for remediation just deletes a Machine, then how are we going to know when it has been replaced to set the `finished` value? The new Machine will have a different name. The only link to the old Machine will be through the MachineSet, right? But if 2 Nodes based on Machines from the same MachineSet are remediated at the same time how will we know which "replacement" matches which original Machine?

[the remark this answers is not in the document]
JoelSpeed: The only way I can think `remediated/finished` makes sense for the cloud side would be if we set it to the timestamp we first see the Machine has gone away, ie, reconcile MHC, see status references machine not yet finished, does machine exist, no, set finished

> I was mainly taking the customer point of view and less on how things can be implemented.
> Not sure if this is worth the effort and complication

Makes sense. I would prefer we come up with some way to make this work for all providers and to be consistent, though that is proving tricky, perhaps for a cloud it is a case of `fenced` == `deletionTimestamp` and `remediated` matches what I said at the beginning of this comment? Not married to this idea though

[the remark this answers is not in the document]
slintes: Nir explained my initial idea about fenced/remediated/finished better than I did, thanks!
But I understand that it's not as easy than I thought to track everything I wanted to track. I assumed that a new node, which replaces a deleted node, will always have the same name again. Since this is not (always) the case, and following your comments above, I think tracking a `finished/remediated` state is too complicated, and we might just skip it (at least for now, adding fields is always possible). It also wasn't in the requirements, I added it because I thought it will be easy to track. Is this ok for everyone?
About the `fenced/safe` timestamp, quoting Nir: this is "when the cluster can safely reschedule the workloads on other node". In the cloud case this is when the node was deleted.
But I see an issue(?) here as well: while we know that node deletion will happen with the built in remediation strategy, this might not happen with external remediation (e.g. reboot strategy). The fact that external remediation is used is already tracked in the `remediationType` field. Maybe the MHC controller can just leave the fenced/safe field empty in this case, and leave it up to the external remediation controllers to fill it? Thoughts?

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
slintes: Not sure yet. I can think of something like `machineDeletion` for the built in case, and "external" for external remediations at this moment. But a soon as we use the external remediation templates we could use the name of the template. So I thing an enum would not work in the long term. Will add some more info about this.

enhancements/machine-api/remediation-history.md
slintes: The intention of the remediation history is just to track what's happening for documentation purposes. It should not have any effect on deciding what should happen next, that will continue to be only based on the actual cluster state.
Does that answer the question?

enhancements/machine-api/remediation-history.md
slintes: Oh, I think I misunderstood an earlier comment. Will remove it, thanks.

enhancements/machine-api/remediation-history.md
slintes: good idea, thanks!

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
beekhof: Maybe: builtin, external, and (looking forward to the future) the name of the external template being instantiated

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
beekhof: I'll throw `complete` into the mix and claim that it should be set when either:
- the Node is deleted[1], or 
- the Node returns to a healthy state

[1] IIRC this always happens before the MHC deletes the Machine.

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,168 @@
...
+    status: "True"
+    type: RemediationAllowed
+  currentHealthy: 2
+  expectedMachines: 2
+  remediationHistory:
+  - conditionStatus: Unknown
+    conditionType: Ready
+    detected: "2020-11-24T17:33:43Z"
+    fenced: "2020-11-24T17:34:35Z"
+    finished: "2020-11-24T17:35:09Z"
```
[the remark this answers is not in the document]
beekhof: I would claim it means "The node has entered a safe state" and represents when the scheduler was able to consider starting workloads elsewhere.

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
beekhof: Actually (reading further for additional context), what I described applies to "finished".

"detected" should be when the error condition is first observed by MHC
"fenced" (or equivalent) is when the condition persists for longer than the configured timeout and recovery was initiated.  perhaps "triggered" or "recovery initiated" ?
"finished" is when the scheduler can start allocating workloads elsewhere (either because the Node is healthy or has been deleted)

enhancements/machine-api/remediation-history.md
beekhof: Essentially, there is no plan to use these status entries to make decisions in subsequent reconcile loops.
Worst case, it might look like the node came back healthy on its own, or some external entity deleted the Machine for us.

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
slintes: that's almost what I already have at https://github.com/openshift/enhancements/pull/567/files#diff-4e98c03501070758bf9d0a5d747bac5fb64f66ff7d3ae0216546dacaaf066d9fR99
How big is the chance that the interna;l / builtin remediation supports multiple strategies? In case we are sure that's only 1 forever, I'm fine with internal or builtin. Otherwise I'd prefer machineDeletion for something more descriptive

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
slintes: Please note that the snippet above is outdated...
We talk about "finished" here, which I meant as node is back and healthy, but I removed that field already because it's not always possible to detect this.
Your "node has entered safe state" is my "fenced" above, but I renamed it to "remediated" already...

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
slintes: Naming things is hard ;) And looking at outdated code snippets does not make it easier...
your "fenced", which you like to be "triggered" or "initiated" is what I call "started"
your "finished" is "remediated" in the latest version.

slintes:
Before we go on discussing names of timestamps on outdated code snippets... is this fine for everyone?
```
	// the time when the unhealthy condition was detected
	Detected *metav1.Time `json:"detected"`

	// the time when remediation started, will be around "detected" plus the "timeout" configured on the MHC
	Started *metav1.Time `json:"started,omitempty"`

	// the time when the node is remediated and in a safe state, so that workloads can be rescheduled.
	Remediated *metav1.Time `json:"remediated,omitempty"`
```

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
beekhof: Ah, I hadn't seen that you'd already addressed this feedback

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
beekhof: sorry for the confusion... detected, started, and finished looks good to me 👍

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,217 @@
...
+type MachineHealthCheckStatus struct {
+
+	[...]
+
+	// History of remediations triggered by this machine health check
+	RemediationHistory []Remediation `json:"remediationHistory,omitempty"`
+}
+
+// Remediation tracks a remediation triggered by this machine health check
+type Remediation struct {
```
beekhof, line 73: Nit...

```suggestion
type RemediationEvent struct {
```

beekhof:
@JoelSpeed anything holding this back from being merged?

enhancements/machine-api/remediation-history.md
JoelSpeed: +1 I prefer RemediationEvent as well

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,217 @@
...
+	// a reference to the target machine or node
```
JoelSpeed, line 74: Which of these will it be, machine or node? I would have assumed we would only ever point to one of these kinds? (Probably the machine?)

enhancements/machine-api/remediation-history.md
```diff
@@ -0,0 +1,217 @@
...
+	// the condition status which triggered this remediation
+	// +kubebuilder:validation:Type=string
+	ConditionStatus *corev1.ConditionStatus `json:"conditionStatus,omitempty"`
+
+	// the reason for the remediation if not a condition
+	// +kubebuilder:validation:Type=string
+	Reason string `json:"reason,omitempty"`
+
+	// the time when the unhealthy condition was detected
+	Detected *metav1.Time `json:"detected"`
```
JoelSpeed, line 91: What happens if it goes back to healthy before the timeout? Is there some way we signal these near misses?

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
JoelSpeed: I would lean towards calling it `deletion` or `external`, not sure we need the `machine` prefix as we are talking about MachineHealthChecks and in context I think it makes sense?

JoelSpeed (commented):
As I understand it, a machine will get detected, the remediation event will be added, that event will then be updated as time progresses. What happens if that remediation event never actually causes a remediation, does the status then look like we lost track of it? 

Do we need to have a third `Type` that is called `recovered` and then if we detect a machine that we thought would go unhealthy, mark the event as recovered?
(If we go down this avenue, how does this work when nodes are starting, since they appear unhealthy at first?)

Has there been any discussion with the upstream community about this type of enhancement, would be good to have this feature upstreamed as well

enhancements/machine-api/remediation-history.md
beekhof: I'd assume the remediation CR wouldn't be created in this case, that only happens once MHC initiates recovery (at which point it can backfill the Detected value based on the last change time of the condition)

enhancements/machine-api/remediation-history.md
slintes: It depends on what exactly triggers remediation. Machines in failed phase or without a node can do so, then the machine will be referenced. When it's a node condition, it will be the node.

enhancements/machine-api/remediation-history.md
slintes: Good question. I like the idea of a having a `Type` for this. Maybe `aborted` instead of recovered, that would be a better fit for starting nodes than `recovered`?

slintes: > the remediation CR wouldn't be created in this case

yes, but for built in remediation we never have a remediation CR, so I'm not sure how this answers the question?

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
slintes: hm, I see a lot about nodes in the context of MachineHealthChecks as well... so I'd prefer to make it explicit that the machine is deleted by the MHC controller, and not the node (we know this, but maybe not everyone else looking at the MHC status)

slintes:
Renamed `Remediation` to `RemediationEvent` and further clarified values of the `Type` field (`machineDeletion` or `external` or `aborted` now)

enhancements/machine-api/remediation-history.md
slintes: done

enhancements/machine-api/remediation-history.md
JoelSpeed: Right ok, that makes more sense. I wonder if it's worth adding to the comment here why it's machine or node, just that context you added would be good IMO.

enhancements/machine-api/remediation-history.md
JoelSpeed: `aborted` works for me

enhancements/machine-api/remediation-history.md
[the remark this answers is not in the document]
JoelSpeed: Ok, given this and your other comment about why machines or nodes, I'm happy to go with `machineDeletion` if you are?

enhancements/machine-api/remediation-history.md
slintes: done

PR #570, opened 2020-12-17
Enhancement Proposal: API to Forward Logs to CloudWatch

alanconway:
See:
- Epic https://issues.redhat.com/browse/LOG-998
- API issue: https://issues.redhat.com/browse/LOG-1001

/cc @jcantrill
/cc @jeremyeder
/cc @sichvoge


enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+### Non-Goals
+
+Enable Cloudwatch metric collection.
+
+## Proposal
+
+New API fields in the `output.cloudwatch` section:
+
+- `region`: (string) AWS region name.
+- `retentionDays`: (number) Number of days to keep logs. Has a default.
```
portante, line 50: Is the intention that the forwarder owns the CloudWatch instance and sets it up?  If so, then this parameter makes sense.

But if the forwarder is being used to send logs to an instance that is externally managed, what is the log forwarder going to do with this parameter?  Reach out and change the CloudWatch instance to have a retention time that matches this parameter?

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Non-Goals
+
+Enable Cloudwatch metric collection.
+
+## Proposal
+
+New API fields in the `output.cloudwatch` section:
+
+- `region`: (string) AWS region name.
```
portante, line 49: What will be done with the `region` parameter?  Will it be used to construct the the proper API endpoint from [1]?

[1] https://docs.aws.amazon.com/general/latest/gr/cwl_region.html

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+This matches the intended use of logs streams, from the [Cloudwatch documentation][concepts]:
+
+> More specifically, a log stream is generally intended to represent the sequence of events coming from the application instance or resource being monitored.
+> For example, a log stream may be associated with an Apache access log on a specific host.
+
+To group logs from multiple containers (by namespace, label etc.) use *Log Groups* as shown in the User Stories section above.
+
+### Open Questions
+
+Should the cluster logging operator need to delete log streams that are no longer in use? For example when containers are deleted?
```
portante, line 166: Why would we want the logging operator to reach out and control an external service?

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+TODO: is this a requirement? If so define an appropriate set of `secret` keys.
+
+### Implementation Details
+
+Using the [fluentd cloudwatch plugin][plugin].
+Most of the CRD fields have an obvious corresponding fluentd parameter.
+
+Set `auto_create_stream true` to create streams and groups on the fly.
+
+`log-stream-name` is set to the `containerID` of the log record.
```
portante, line 156: Is the `containerID` the UUID of the container or the more human-readable name?

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+This matches the intended use of logs streams, from the [Cloudwatch documentation][concepts]:
+
+> More specifically, a log stream is generally intended to represent the sequence of events coming from the application instance or resource being monitored.
+> For example, a log stream may be associated with an Apache access log on a specific host.
+
+To group logs from multiple containers (by namespace, label etc.) use *Log Groups* as shown in the User Stories section above.
```
portante, line 162: Seems like there are two default modes that would be fairly easy to consider, and then the one advanced mode for groups:

1. Three main groups: "application", "infrastructure", "audit"
2. Group by "infrastructure", "audit", then by "namespace" for application logs
3. Give the user the ability to customize the name of the group from a constant or variable

Seems like we can go a long way with 1 & 2 for now; 3 seems like something to tackle once this is working in production well.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+
+[Cloudwatch][concepts] defines Log Groups and Log Streams.
+The CRD above creates 3 Log Groups named "application", "infrastructure" and "audit".
+A new Log Stream is created for each container producing logs.
+
+**Note**: The installer UI (Addon or OLM) can get AWS credentials from a `cloudcredential.openshift.io/v1`.
+The user only has to provide a `region` to enable cloudwatch forwarding for a cluster.
+Details are out of scope for this proposal.
+
+#### I want to define my own log groups
```
portante, line 92: Seems like a much more complicated feature to approach in the short-term.  Hopefully this won't be offered initially to reduce the complexity.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+     secret:
+        name: mysecret
+  pipelines:
+  - inputRefs: [application, infrastructure, audit]
+  - outputRefs: [CloudwatchOut]
+```
+
+[Cloudwatch][concepts] defines Log Groups and Log Streams.
+The CRD above creates 3 Log Groups named "application", "infrastructure" and "audit".
+A new Log Stream is created for each container producing logs.
```
portante, line 86: ... "within each group".

portante (commented):
This seems a bit complicated for a first pass with the ability to set groups and streams arbitrarily.

Is there a way to just support 3 groups (app, infra, audit) and/or 2 groups (infra & audit) and apps grouped by namespace?  Seems like those would be easier to get right in the very, very short amount of time we have to implement this.

enhancements/cluster-logging/forward_to_cloudwatch.md
jewzaam: Guess this would be useful if the CW endpoint configured (via the credentials I assume) is not in the same region as the cluster.  Is this optional and defaults to cluster's region if not ste?

enhancements/cluster-logging/forward_to_cloudwatch.md
jewzaam: If the forwarder doesn't have permission to manage this and it's externally managed CW would this just not be set and that bit of config skipped?  If not set, do nothing.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+Enable Cloudwatch metric collection.
+
+## Proposal
+
+New API fields in the `output.cloudwatch` section:
+
+- `region`: (string) AWS region name.
+- `retentionDays`: (number) Number of days to keep logs. Has a default.
+- `group`: (string, *nice-to-have*) Cloudwatch log group name.
+   Defaults to the log type (application, infrastructure or audit)
```
jewzaam, line 52: The "group" name in CW defaults to the log type?  Would expect this to default to something unique to the cluster such as infrastructures .status.infrastructureName

enhancements/cluster-logging/forward_to_cloudwatch.md
jewzaam: From a usability point of view, how easy will it be for log consumers to identify which of these streams map to which applications?  Clarity on what `containerID` actually is in terms of mapping to metadata in-cluster will help clarify this.  If it's the UUID of the container it's a non-starter.  The name must map to something that isn't ephemeral.

enhancements/cluster-logging/forward_to_cloudwatch.md
jewzaam: By namespace makes the most sense with the limits imposed by CW on log puts.  Grouping all app logs into a single stream will result in loss of logs as we'll exceed quota on clusters heavily utilizing logs.

enhancements/cluster-logging/forward_to_cloudwatch.md
jewzaam, line 166: No, this is the purpose of `retentionDays` configuration parameter. Let CW manage retention, don't re-implement.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+
+To group logs from multiple containers (by namespace, label etc.) use *Log Groups* as shown in the User Stories section above.
+
+### Open Questions
+
+Should the cluster logging operator need to delete log streams that are no longer in use? For example when containers are deleted?
+
+### Risks and Mitigations
+
+None in particular.
```
jewzaam, line 170: Risks should include that quota in CW can be exceeded if insufficiently granular streams are configured with CW.
See PutLogEvents on https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html:
```
5 requests per second per log stream. Additional requests are throttled. This quota can't be changed.

The maximum batch size of a PutLogEvents request is 1MB.

800 transactions per second per account per Region, except for the following Regions where the quota is 1500 transactions per second per account per Region: US East (N. Virginia), US West (Oregon), and Europe (Ireland). You can request a quota increase.
```

enhancements/cluster-logging/forward_to_cloudwatch.md
jcantrill: I believe @portante originally mentioned to me this would be useful.  I don't pretend to understand what the implications of specifying it.  The default would be to do nothing with region if it is not set

enhancements/cluster-logging/forward_to_cloudwatch.md
jcantrill: This API was worked back from the options available by both the plugin and the Cloudwatch API.  The plugin allows setting this parameter which presumably means it will config auto-created loggroups and logstreams with this value.  The intention is up for debate but the very early prototype assumes the forwarder will create the logroup and logstream under the guidance the only decision ROSA users will make is to opt-into Cloudwatch (e.g. checkbox, true/false)

I don't know the behavior of the CW api so unable to speak to the ramifications of setting the value for externally managed CW.  Seems this would be tangential however as credentials provided, at least in the ROSA case, presumably will be able to access and control whatever loggroup is created by the forwarder

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+
+New API fields in the `output.cloudwatch` section:
+
+- `region`: (string) AWS region name.
+- `retentionDays`: (number) Number of days to keep logs. Has a default.
+- `group`: (string, *nice-to-have*) Cloudwatch log group name.
+   Defaults to the log type (application, infrastructure or audit)
+- `groupLabel`: (string, *nice-to-have*) Kubernetes label to use as group name.
+  If the label is absent, use value of `group`.
+- `fetchIntervalSeconds`: (number, *nice-to-have*) interval to fetch logs in seconds.
```
jcantrill, line 55: Why do we need to be able to fetch logs in order to push them to storage?

enhancements/cluster-logging/forward_to_cloudwatch.md
jcantrill: I would ask the same question why the group would be used used to separate logs in lieu of logStream within a group?

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+
+Existing fields:
+
+- `url`: Not used in production. Can be set for testing purposes.
```
jcantrill, line 59: I saw this in the plugin config and would be curious how it is used for testing.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+- `secret`: AWS credentials, keys `aws_access_key_id` and `aws_secret_access_key`.
+
+*Priority*: The `group`, `groupLabel` and `fetchIntervalSeconds` fields are lower priority.
```
jcantrill, line 62: Priority of what?

enhancements/cluster-logging/forward_to_cloudwatch.md
jcantrill, line 86: We may want to say choose a different word then container as its not applicable to all sources.  infra has both container and journal/node (though SRE has stated journal logs or not used).  I'm not sure how you describe "audit" as those come from specific files

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+  - inputRefs: [application, infrastructure, audit]
+  - outputRefs: [CloudwatchOut]
+```
+
+[Cloudwatch][concepts] defines Log Groups and Log Streams.
+The CRD above creates 3 Log Groups named "application", "infrastructure" and "audit".
+A new Log Stream is created for each container producing logs.
+
+**Note**: The installer UI (Addon or OLM) can get AWS credentials from a `cloudcredential.openshift.io/v1`.
+The user only has to provide a `region` to enable cloudwatch forwarding for a cluster.
```
jcantrill, line 89: Is region really required?  I presume if you define nothing it defaults to whatever region the service is in

enhancements/cluster-logging/forward_to_cloudwatch.md
jcantrill: I agree this may be a stretch goal and maybe not something ever needed, especially since we have other proposals which would allow you to "route" logs to different outputs based on a label

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+```
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogForwarder"
+spec:
+  outputs:
+   - name: CloudwatchOut
+     type: cloudwatch
+     cloudwatch:
+       region: myregion
+	   groupLabel: LogGroup
```
jcantrill, line 109: Not clear to me but does this mean if I label my pod to include the label "LogGroup" its logs route to "CloudwatchOut"

enhancements/cluster-logging/forward_to_cloudwatch.md
jcantrill: logstream per container would be easier to digest if the name was something similar to the fluent tag or the info really we can get from it:
* namespace
* podname
* container ID

Given the containerID is a hash that is more or less meaningless to users I would likely drop this from the stream unless there is a need to uniquely distinguish multi-container pod logs and there is no other mechanism

enhancements/cluster-logging/forward_to_cloudwatch.md
jcantrill: I contend we should give user's fewer choices by introducing strategies that represent the intended data model; more open configuration options leads to more places to make the wrong decisions. I don't have a strong opinion of utilizing a single logroup versus multiple but general idea of strategies:

* Source  - one group (opinionatedly named) per source (e.g. openshift-infrastructure, openshift-application)
* Namespaced - one group per ns (opinionatedy named)? how to name logstreams is open question. 
* Custom - future choice if needed to allow users the config flexability

My original proposal would be a similar concept but utilizing a single logroup and alternate logstreams based on a similar naming strategy. Quota seems to be restricted to logstreams so I'm not certain what advantage is provided by separating by group.  Does it make query and fetch easier?  Does it allow setting permissions differently?

enhancements/cluster-logging/forward_to_cloudwatch.md
jcantrill: LogForwarding does no curation of any other output; I don't see why Cloudwatch would be treated differently

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Risks and Mitigations
+
+None in particular.
+
+## Design Details
+
+### Test Plan
+
+- E2E tests: Need access to AWS logging accounts.
```
jcantrill, line 176: We may be able to do this since CI can utilize the cloud credential operator like what ROSA is using

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: We don't, my mistake.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: The fluentd plugin has this option, and from my reading of the cloudwatch it appears that cloudwatch allows the collector to set retention times, even on individual log entries, which surprised me. Possibly I have misinterpreted? It seems that the collector drives a lot of the cloudwatch configuration, creating groups and so on.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: As I read it, region is required to establish a cloudwatch connection. Its my first reading though, so if that's not correct we'll adjust. Does anyone have a working cloudwatch configuration that we can compare with this proposal?

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: That's how cloudwatch works apparently, the collector is responsible for creating log streams and groups, and can delete them when done - I don't know what happens if the collector doesn't delete them.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: The log entries contain all the metadata that the application might need (container name, uuid, namespace, k8s labels and more) The log stream name only needs to be a unique identifier that lives as long as the source of logs. I'm proposing a log-stream per container, so container UUID fits the bill.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: +1 I'll add namespace and leave the other options as nice-to-have.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: +1, made that more clear.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: There is always a 1-1 log-stream to container relationship regardless of grouping. A log stream can only be part of a single group, which is another reason for the container==log-stream choice since we can't use a single log stream for anything that might span more than one node/collector process.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: Need to check if region is a required parameter.
In the updated proposal the default is infra/app/audit, the user can also enable use of namespace for app logs. We don't have any named sub-categories of infra logs.
When you say "unique to the cluster" do you mean we need to prefix with a cluster-id?

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: I've clarified the subset of stuff that's for first release.
As I read the docs, it is the collector that sets up cloudwatch: creates and deletes groups and streams, sets retentions (even on a per-entry basis??!) etc. Shout if I misread.
Does anyone have a working cloudwatch config we can look at?

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: Retention refers to stored logs. A log stream is a transient artifact created to transmit logs.
I'm not clear on the rules, the documentation says that you "can delete log streams when you are done" (i.e. when a container is destroyed in our case) but doesn't say what happens if you don't.

enhancements/cluster-logging/forward_to_cloudwatch.md
jewzaam: Before locking this down let's be sure from a UX point of view I can get to the logs with just the namespace or container name.  If I first have to know the stream in which to search and the stream is the UUID this is not going to work.

enhancements/cluster-logging/forward_to_cloudwatch.md
jcantrill: I dont see how this is relevent. "the collector is responsible for creating log stream"  How would the collector know about a definition of "done".  It is the user who is responsible for discarding the logs.  The collector is only responsible for pushing them?

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: You can apparently set up a local test cloudwatch server and connect to it for tests. I think we might even be able to use fluetnd's cloudwatch_in plugin to create such a test server, but haven't tried it yet.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: I meant they are "nice to have" - I reworded to make it clearer what's for the first cut and what's for later.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: Good point. I  think it is "container" if there is one and "node" if not. For node-scoped logs we should probably have a log-stream per node per category (node32-infrastructure, node32-application) unless we have any further breakdown of infra andn audit? I'll update.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: If your pod has label "LogGroup:foo" then the cloudwatch log-group for that stream will be "foo"

alanconway:
@jewzaam @portante @jcantrill update taking account of all your comments (thanks for that) but may still be some questions outstanding.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,249 @@
...
+### Cloudwatch streams and groups
+
+[Cloudwatch][concepts] defines *log groups* and *log streams*. To paraphrase the documentation:
+
+>  A log stream is a sequence of log events that share the same source ... For example, an Apache access log on a specific host.
+
+>  Log groups define groups of log streams that share the same retention, monitoring, and access control settings ... For example, if you have a separate log stream for the Apache access logs from each host, you could group those log streams into a single log group called MyWebsite.com/Apache/access_log.
+
+In other words a *log stream* corresponds to the smallest log-producing units.
+*Log groups* are used to group larger sets of related logs together.
```
portante, line 56: Perhaps consider, `*Log groups* are  used to group larger sets of related log streams together.`

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,249 @@
...
+
+We must consider *container logs*, and *node logs*.
+- "application" logs are always container logs
+- "infrastructure" logs are a mix of container and node logs
+- "audit" logs are all node logs.
+
+For *container-scoped* logs we auto-create a log stream for *each container*, using the container UUID as the log stream name.
+- The log stream name is just a unique identifier for the source container.
+- Log *entries* include meta-data (container name, namespace etc.) for the user to search/index.
```
portante, line 65: You might want to mention the minimum set of metadata needed to address the source of the log stream, something like: "cluster", "node", "namespace", "pod", "container".

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,249 @@
...
+  The log stream name is *not* needed for indexing or searching logs.
+- Log streams are for a *single source*, we cannot send logs from multiple nodes on the same log stream.
+
+For *node-scoped* logs we will create "audit"" and "infrastructure" streams for each node: e.g. `node-id/audit`.
+
+By default the forwarder creates 3 *log groups* for each of the log categories:  "application", "infrastructure" and "audit".
+Application logs can also be grouped by namespace, k8s labels, or forwarder configuration, see below.
```
portante, line 72: Perhaps use "stream" in here: `Application log streams can also ...`.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,249 @@
...
+
+### API fields
+
+New API fields in the `output.cloudwatch` section:
+
+- `region`: (string) AWS region name, required to connect.
+- `retentionDays`: (number) Number of days to keep logs. Defaults to cloudwatch default retention.
+- `groupByNamespace`: (bool) If true, set the Cloudwatch group name to the source namespace.
```
portante, line 80: Keep in mind, one of the problems with Kubernetes "namespace" names is that they can be re-used over time.  So if somebody defines a very long retention period in Cloudwatch, but namespaces change, they might have unwanted streams landing in the same group.

Consider using namespace UUID instead.  See also https://github.com/kubernetes/kubernetes/issues/58638.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,249 @@
...
+  If false, group name is set to the log category: "application", "infrastructure" or "audit".
+
+*Nice-to-have*: the following fields may be omitted from the first release and implemented later.
+
+- `group`: (string, *nice-to-have*) Fixed Cloudwatch log group name for all logs leaving this this output.
```
portante, line 85: `s/this this/this/`.

Can users query Cloudwatch logs across groups (looks like it is [available](https://aws.amazon.com/about-aws/whats-new/2019/07/cloudwatch-logs-insights-adds-cross-log-group-querying/))?  If so, then you might need to consider ways for the output stream to add metadata instead of changing the groups so that logs can be combined across groups.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,249 @@
...
+```
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogForwarder"
+spec:
+  outputs:
+   - name: CloudwatchOut
+     type: cloudwatch
+     cloudwatch:
+       region: myregion
+	   groupByLabel: LogGroup
```
portante, line 154: What happens if the user specifies `groupByLabel` and `groupByNamespace`?

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,249 @@
...
+### Open Questions
+
+Should the cluster logging operator need to delete log streams that are no longer in use? For example when containers are deleted?
+
+Is region a required parameter for connecting to cloudwatch? If not make that clear and evaluate if it's necessary.
+
+### Risks and Mitigations
+
+[Cloudwatch quota][quota] can be exceeded if insufficiently granular streams are configured.
+We configure a stream-per-container which is the finest granularity we have for logging.
```
portante, line 215: The throttling can also be "unfair" across nodes.  For large clusters, the risk that certain streams always fail is a real threat.

Having a central forwarder in the cluster would be one way to mitigate this.

alanconway: Do you have any pointers to cloudwatch's fairness strategy? Not clear to me how a central forwarder would mitigate this; a central forwarder would need to replicated, and to split traffic into small log streams given the per-stream quotas; so it would look pretty the same to cloudwatch - a small set of replicas generating a large set of log streams, vs. a small set of nodes generating a large set of log streams.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: Illegal, I'll document it.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: The forwarder does add all the metadata so everything is avaliable for query once cloudwatch has collected & stored the logs. This configuration is only for controlling how logs are *delivered* in terms of log streams and log groups.

alanconway: Metadata is available on all logs regardless of grouping. Whether/how the user choooses to use cloudwatch groups is a separate issue. I'm guessing that namespaces, labels or static groups will cover most use cases but I don't know cloudwatch enough to say what's the best way to group logs.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: That's all part of the general logging model, I do need to get it documented more clearly but it's not specifically tied to this proposal.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: My current assumption is that if you create a forwarder configuration with "namespaces: [foo]",
you want logs from any namespace named "foo", present or future. Presumably the name "foo" means something to you, and if you delete and re-create it, then it is still "the same foo" in some sense. This seems to also be the interpretation of tools like oc, for example: 
    oc get -ns foo cm --watch
will keep printing out configmaps created in namespace 'foo' even if I delete and re-create it.

There may be use cases for supporting namespace UUIDs *as well as* names, but surely not *instead* of? Can you point me to existing k8s APIs/tools that use namespace UUIDs instead of names so I can get an idea of what the use cases are?

enhancements/cluster-logging/forward_to_cloudwatch.md
portante: Under the assumption that no node in the cluster shares the same log stream with another node (stream ID is either container UUID or is the unique host name), then the only 800/1,500 transactions a second quota can be hit (too many "small" PutLogEvent requests combined).

For a 50 node cluster, that would be 16/30 PutLogEvent reqs / sec at 1 MB per req.

So a really chatty node can chew up that up (that is only 16/30 MB/s).

alanconway: Understood, but there's no way we can work around that. If excessively chatty log streams use up a  cluster-wide quota, then others will starve. Doesn't matter if the streams originate from nodes or from replicas of a central forwarder.

alanconway: Rate limiting is still in our plans, when we implement that we'll be able to enforce some policy. Until then we don't control the volume of traffic.

portante: But by having central nodes, you can ensure the throttling is spread out properly, and no one node gets it worse than another, where as you are beholding to how CloudWatch throttles without it.

enhancements/cluster-logging/forward_to_cloudwatch.md
portante: Unfortunately, it is our responsibility to properly communicate that to the users, but this is an "action-at-a-distance" problem.  The retention times can be so long, that one namespace name can go away, and then be recreated by somebody else in charge.  If there is any sensitive information in those logs collected, we are not helping them keep them separate.

Using namespace UUIDs is the only way to be sure.

> My current assumption is that if you create a forwarder configuration with "namespaces: [foo]", you want logs from any namespace named `foo"`, present or future.  Presumably the name `foo` means something to you, and if you delete and re-create it, then it is still "the same `foo`" in some sense.

With short retention times it is very like the "you" is the same person or group.  With long retention times, it is much more likely the "you" changes, and the presumptions we make will have changed.

> This seems to also be the interpretation of tools like `oc`, for example: `oc get -ns foo cm --watch` will keep printing out `configmap`s created in namespace '`foo` even if I delete and re-create it.

True, but there is no history for namespaces in Kubernetes (which is also a problem, the Kube community has not been around long enough to realize the issue yet).  So there is no facility for Kubernetes to do anything different.

But that short retention period does not mean it translates to long term log retention.

It is a design flaw of Kubernetes that operational semantics use name lookups instead of UUIDs.  All of Kubernetes would operate the same using UUIDs, but not the other way around.

> There may be use cases for supporting namespace UUIDs _as well as_ names, but surely not _instead_ of?

You can all solve all use cases with namespace UUIDs, you can't handle all use cases with namespace names because names are reused by Kubernetes.

> Can you point me to existing k8s APIs/tools that use namespace UUIDs instead of names so I can get an idea of what the use cases are?

As stated above, log retention times are much, much longer than operational Kubernetes, so they don't equate or translate.  We have to be thinking about this in our solution to help users make sure they don't expose sensitive information inadvertently.

enhancements/cluster-logging/forward_to_cloudwatch.md
portante: Perhaps consider stating it is the log forwarder's responsibility to ensure the required metadata to address the log entries per-stream uniquely is added.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: I don't understand how you want to use UUIDs. Do you want to 
a) use them in forwarder configuration instead of namespace names? 
b) have the forwarder look up the UUID based on a configured namespace name?

a) makes configuration of the forwarder tricky but I guess if you want to use UUIDs in place of namespace names that's inevitable. 

b) raises the question of *when* the forwarder should look up the UUID. Given the eventually-consistent nature of k8s, we can't assume that the target namespace will exist when the forwarder is started, and even if it does we don't know if it is the _intended incarnation_ of that namespace.

I would note again that the namespace UUID is part of the meta-data we store, so it is possible after the fact to determine which incarnation of a namespace a given log came from.

I'm open to supporting UUIDs but given that using names is such a widespread practice (even if it is a design flaw) I think we would have to support both.

portante: > I don't understand how you want to use UUIDs. Do you want to
> a) use them in forwarder configuration instead of namespace names?

Yes, the target log group would be by namespace UUID instead of namespace name.

> b) have the forwarder look up the UUID based on a configured namespace name?

As you wrote below, this is already happening for the most part with the OpenShift fluentd configuration today.

> a) makes configuration of the forwarder tricky but I guess if you want to use UUIDs in place of namespace names that's inevitable.

You already have that today.

> b) raises the question of _when_ the forwarder should look up the UUID. Given the eventually-consistent nature of k8s, we can't assume that the target namespace will exist when the forwarder is started, and even if it does we don't know if it is the _intended incarnation_ of that namespace.

Can you explain more of what that sentence above means?

The whole purpose of https://github.com/kubernetes/kubernetes/issues/58638 is to make it so that there is no need for a namespace UUID lookup by name when collecting logs: when the container log file is created, it is created with the namespace UUID and the namespace name so that no lookup is required, and the logs can be safely forwarded to another system where the UUID can be used to correctly associate logs from different containers in their namespace.

> I would note again that the namespace UUID is part of the meta-data we store, so it is possible after the fact to determine which incarnation of a namespace a given log entry came from, and to filter on UUID if you need to focus on one incarnation.

The problem is that if you don't use UUID to store the logs remotely in the Log Group, you can't help the user ensure that logs from one namespace are not made visible to the wrong users who have access to a Log Group.

> I'm open to supporting UUIDs but given that using names is such a widespread practice (even if it is a design flaw) I think we would have to support both.

Sure, supporting both is possible, as long as we explain to the users the potential issues with both and how they can avoid them, and how the log forwarder can help or hinder the proper log file association.

alanconway: The current proposal for cloudwatch is to use the container UUID to create log *streams*, which is as unique as using a namespace UUID. 
Log *groups* are created at the user's discretion via configuration of the forwarder. Currently we allow log type (app/infra/auth), namespace *name* and (shortly) POD label selectors as criteria. 
We can easily add namespace UUID as an additional input selection criterion. However if we need to go to that level of detail, perhaps we should add general field selectors rather than treating UUID as a special case, since there might be other fields of interest. 
What do you think? There are 2 separate concerns here: 
1. user requirements/UI design
2. implementation
I realize UUID has special case status in k8s but we should consider 1. as well as 2. to decide between UUID-as-speical-case vs. general-field-selector UIs.

portante: > The current proposal for cloudwatch is to use the container UUID to create log _streams_, which is as unique as using a namespace UUID.

If UUIDs are worth their salt then no container from any node across all possible clusters that could ever exist in our lifetime would end up with the same UUID.  =)

So this proposal does not give the user another way to choose how to create log streams, which is smart.  Too much flexibility in the first implementation leads to complexity, making it much harder to verify.

> Log _groups_ are created at the user's discretion via configuration of the forwarder. Currently we allow log type (app/infra/auth), namespace _name_ and (shortly) POD label selectors as criteria.

Here it might be prudent and avoid providing too many ways to create log groups on the first pass, so adding POD label selectors might be worth waiting on to avoid testing complexity.

There is another case that we have to consider here as well: multiple clusters sending to one CloudWatch instance.  This is likely to be a common use-case given the Kubernetes clusters tend to be small.  If this ends up being true, we'll want to make sure the user is okay with mixing logs from different clusters in the same groups, or adding a way to create groups unique to a cluster.  If shared CloudWatch instances ends up not being true, no real issue.

Is CloudWatch query access is by log group?  If so, for the "log type" and "namespace name" case, clusters sharing the same CloudWatch instance would potentially have a problem limiting access.  Using "namespace UUID" would not be a problem, because the namespaces are unique across the clusters.

> We can easily add namespace UUID as an additional input selection criterion. However if we need to go to that level of detail, perhaps we should add general field selectors rather than treating UUID as a special case, since there might be other fields of interest.
> What do you think? There are 2 separate concerns here:
> 
> 1. user requirements/API design
> 2. implementation details/performance
> 
> I realize UUID has special case status in k8s but we should consider 1. as well as 2. to decide between UUID-as-special-case vs. general-field-selector API.

It seems easier to add support for namespace UUID before the POD label selector or a general field selector.  The problem we have is when a log is collected and ends up not having the particular field to use, what then?  If we ensure a log always has a namespace UUID, namespace name, and log type, then we can have a well behaved system.  If we can't ensure those fields are always present, then we have to define what happens in those cases.

alanconway: Thanks for your patience, that was a worthwhile discussion.
I think the new proposal is exactly what you want, and I have resolved my confusion caused by mixing new cloudwatch features with how existing "generic" forwarder features can be used with cloudwatch. The two are now fully orthogonal which is as it should be.

alanconway:
@jcantrill @portante I think the latest draft may be a winner, please comment if you agree so we can LGTM and unhold.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+
+In other words a *log stream* corresponds to the smallest log-producing units.
+A *log group* is a collection of related *log streams*.
+
+We must consider *container logs*, and *node logs*.
+- "application" logs are always container logs
+- "infrastructure" logs are a mix of container and node logs
+- "audit" logs are all node logs.
+
+For *container-scoped* logs we auto-create a log stream for *each container*, using the container UUID as the log stream name.
```
jcantrill, line 63: The POC is using the fluent tag which is the full file path on the node.  Do we think its necessary to further burden the collector to change it to something else?

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+  This is a common case, many core k8s tools and APIs work this way.
+- **namespace uuid**: group per namespace *object* using the UUID.
+  Destroying then creating a namespace object with the same name results in a *new log group*.  Use when it is important to distinguish logs from successive namespaces instances with the same name. For example, when namespace re-creation is considered a security risk.
+
+### API fields
+
+New API fields in the `output.cloudwatch` section:
+
+- `region`: (string) AWS region name, required to connect.
+- `retentionDays`: (number) Number of days to keep logs. Defaults to cloudwatch default retention.
```
jcantrill, line 85: Per conversation with @sichvoge we really would prefer to not add this to LF as it is a variance from other types.  I think TBD

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+- `groupBy`: (string, default "category") Take group name from logging meta-data. Values:
+   - `category`: category of log entry - one of "application", "infrastructure", or "audit"
+   - `namespace_name`: Container's namespace name
```
jcantrill, line 88: Should be camelcase I think to fit go conventions

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+   - `namespace_uuid`: Container's namespace UUID
+
+Existing fields:
+
+- `url`: Not used in production. Sets the `endpoint` parameter in fluentd for use in testing.
+- `secret`: AWS credentials, keys `aws_access_key_id` and `aws_secret_access_key`.
```
jcantrill, line 94: Maybe more explicit `AWS credentials, the secret must container keys....`

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+```
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogForwarder"
+spec:
+  outputs:
+   - name: CloudwatchOut
+     type: cloudwatch
+     cloudwatch:
+       region: myregion
+	   groupBy: namespace_name
```
jcantrill, line 159: Should be same level as region?

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+To group by namespace UUID, replace `namespace_name` with `namespace_uuid`.
+
+### Implementation Details
+
+Using the [fluentd cloudwatch plugin][plugin].
+Most of the CRD fields have an obvious corresponding fluentd parameter.
+
+Set `auto_create_stream true` to create streams and groups on the fly.
+
+`log-stream-name` is set to the `containerID` of the log record.
```
jcantrill, line 176: underscores

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+
+Log group names are set using `log_group_name` for a static `group` field.
+The `log_group_name_key` can select the log category, namespace or label as group name.
+
+### Open Questions
+
+Which (if any) of the nice-to-have features should be included, and when?
+
+Is the cluster logging operator responsible for deleting log streams that are no longer in use, for example when containers are deleted?
```
jcantrill, line 185: I think the answer to this is "NO". We do not take responsibility for curation any place else in relation to the forwarder

enhancements/cluster-logging/forward_to_cloudwatch.md
portante: Why send the entire fluent tag?  Isn't the UUID much shorter?

enhancements/cluster-logging/forward_to_cloudwatch.md
portante: Why would the log forwarder participate in setting the retention?

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+[Amazon Cloudwatch][aws-cw] is a hosted monitoring and log storage service.
+This proposal extends the `ClusterLogForwarder` API with an output type for Cloudwatch.
```
portante, line 30: s/Cloudwatch/CloudWatch/

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+- [X] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+[Amazon Cloudwatch][aws-cw] is a hosted monitoring and log storage service.
```
portante, line 29: s/Cloudwatch/CloudWatch/

enhancements/cluster-logging/forward_to_cloudwatch.md
portante: And applies to all the rest of them.  It is always capital `W` on the Amazon site.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+[Cloudwatch][concepts] defines *log groups* and *log streams*. To paraphrase the documentation:
+
+>  A log stream is a sequence of log events that share the same source ... For example, an Apache access log on a specific host.
+
+>  Log groups define groups of log streams that share the same retention, monitoring, and access control settings ... For example, if you have a separate log stream for the Apache access logs from each host, you could group those log streams into a single log group called MyWebsite.com/Apache/access_log.
+
+In other words a *log stream* corresponds to the smallest log-producing units.
+A *log group* is a collection of related *log streams*.
+
+We must consider *container logs*, and *node logs*.
```
portante, line 58: Perhaps add audit logs?  `We must consider *container logs*, *node logs*, and *audit logs*.

enhancements/cluster-logging/forward_to_cloudwatch.md
portante: The question is, can the user find out what is the full path of the container log file through some Kube API?  We know they can retrieve the UUID of the container.

portante: Perhaps a rewording to drop "scoped" which hasnt' been introduced before this?  `For *container-logs*, we auto-create a log stream for *each container* using its UUID as the log stream name.`

Also, it is not clear the log forwarder chooses to auto-create, it seems the CloudWatch instance has to be configured to allow such auto-creation.  Might be worth mentioning as a requirement.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+- "application" logs are always container logs
+- "infrastructure" logs are a mix of container and node logs
+- "audit" logs are all node logs.
+
+For *container-scoped* logs we auto-create a log stream for *each container*, using the container UUID as the log stream name.
+- The log stream name is just a unique identifier for the source container.
+- Log *entries* include meta-data (container name, namespace etc.) for the user to search/index.
+  The log stream name is *not* needed for indexing or searching logs.
+- Log streams are for a *single source*, we cannot send logs from multiple nodes on the same log stream.
```
portante, line 67: Not sure this is a "can" or "can not" thing, but more of a "do" or "do not" thing.  We do not *want* to send logs from multiple nodes to the same log stream to avoid hitting hard rate-limits applied to individual log streams.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+
+For *node-scoped* logs we will create "audit"" and "infrastructure" streams for each node: e.g. `node-id/audit`.
```
portante, line 69: This is great for audit, but for node logs which are often a big source of log traffic when Kube is not configured correctly or something goes awry, we might want to take advantage of the Journal tags to subset the node logs by component stream.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+  outputs:
+   - name: CloudwatchOut
+     type: cloudwatch
+     cloudwatch:
+       region: myregion
+     secret:
+        name: mysecret
+  pipelines:
+  - inputRefs: [application, infrastructure, audit]
+  - outputRefs: [CloudwatchOut]
```
portante, line 143: It is worth documenting for this use-case what the default behavior for log group and stream are?

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+   - name: CloudwatchOut
+     type: cloudwatch
+     cloudwatch:
+       region: myregion
+	   groupBy: namespace_name
+     secret:
+        name: mysecret
+  pipelines:
+  - inputRefs: [application, infrastructure, audit]
+  - outputRefs: [CloudwatchOut]
```
portante, line 164: Might be worth explicitly stating what the log stream would be.

enhancements/cluster-logging/forward_to_cloudwatch.md
portante: Is `containerID` the UUID or the fluent log file tag?  And is that existing metadata available today?

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,226 @@
...
+
+Using the [fluentd cloudwatch plugin][plugin].
+Most of the CRD fields have an obvious corresponding fluentd parameter.
+
+Set `auto_create_stream true` to create streams and groups on the fly.
+
+`log-stream-name` is set to the `containerID` of the log record.
+
+Log group names are set using `log_group_name` for a static `group` field.
+The `log_group_name_key` can select the log category, namespace or label as group name.
```
portante, line 179: Lines 86-89 seem to imply we can't use labels, is that true?

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: @jcantrill this is fluentd config not CRD, will clarify.
@portante as discussed earlier, this is an ID that is _as unique as_ the container/node UUID. It could be the actual UUID, a fluent or fluentbit tag, a log file name, or anything with the required uniqueness. Whatever it is, it will be constructed from available metadata. I want to leave this open so we can change it in future to improve performance for static metadata.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: Correct, thanks. Labels are part of "Nice To Have: more options for groupBy", not required for first impl.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: This is not curation of stored logs, it's cleanup of "connection" artifacts used to forward logs to CloudWatch.
We can probably ignore this for now - CloudWatch allows clients to create log streams on the fly so I assume it has a stateless impl and/or garbage collection so it won't leak resources if clients never delete their streams. If that turns out to be wrong, we may need to revisit.

alanconway (commented):
Great feedback, I think we're close. @portante @jcantrill please scan again. I've clarified the role of log stream names, otherwise its small changes.

enhancements/cluster-logging/forward_to_cloudwatch.md
portante: > +1 but this is a generic improvement that applies to all outputs, not just CloudWatch.

Agreed.

> We need to take time to design that with a wider view. If/when we do it will benefit CloudWatch as well.

Sure.

enhancements/cluster-logging/forward_to_cloudwatch.md
sichvoge: I still think we should not expose retention in our API at least. How about we set a default value in our Fluentd for now (e.g. 3d), document that and basically point customers to https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/Working-with-log-groups-and-streams.html#SettingLogRetention so that they can change the retention themselves if they need something different?

That at least doesn't require us to bring something into our API that most people feel not comfortable and if we have to expose it later, fine.

sichvoge (changes requested):
I added an additional comment for `retentionDays` since I am not comfortable to actually expose this in our API. See explanation in my comment.

alanconway:
@jcantrill @portante @sichvoge all comments taken on board, properly updated this time (I managed to check this out twice so brief confusion) Please comment.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,281 @@
...
+  Use when it is important to distinguish logs from successive namespaces instances with the same name.
+  For example, when namespace re-creation is considered a security risk.
+
+### API fields
+
+New API fields in the `output.cloudwatch` section:
+
+- `region`: (string) AWS region name, required to connect.
+- `groupBy`: (string, default "category") Take group name from logging meta-data. Values:
+   - `category`: category of log entry - one of "application", "infrastructure", or "audit"
```
jcantrill, line 101: Should we call this `source` to be consistent with the rest of logforwarding

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,281 @@
...
+   - `namespaceName`: Container's namespace name
+   - `namespaceUUID`: Container's namespace UUID
```
jcantrill, line 103: if there is any error fetching meta info from the API server this will fail? What should the fall back be?  namespace, podname, container hash are available from the container file path and don't rely on an api call

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,281 @@
...
+- *node logs* are logs from a cluster node's journald or syslogd, they do *not* come from any container.
+
+**Note**: Log *sources* are different from log *categories*.
+Category *application* includes only container logs.
+Categories *infrastructure* and *audit* include a mix of selected container and node logs.
+
+#### Generating log streams
+
+This proposal generates a unique *log stream* for each distinct container or node log source.
+The *log stream name* is constructed from meta-data including the container/node UUID.
```
jcantrill, line 70: Naming for audit logs?  im guessing someting with the node id in it and the specific audit log?

jcantrill, line 70: This should also maybe node hostname as thats the info we have

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: No fallback, if the metadata is missing we don't set a group. It will default to whatever AWS does if no group is explicitly set. Note this is the *group* name not the *stream* name.

enhancements/cluster-logging/forward_to_cloudwatch.md
alanconway: Updated to follow your prototype: fluent-key for containers, node-id  + fluent-key for node logs (infra, audit)

alanconway: Followed your prototype - using fluent key.

enhancements/cluster-logging/forward_to_cloudwatch.md
```diff
@@ -0,0 +1,309 @@
...
+Use the [fluentd CloudWatch plugin][plugin].
+Most of the CRD fields have an obvious corresponding fluentd parameter.
+
+- Set log group names using `log_group_name_key` to select the appropriate meta-data value
+- Set `auto_create_stream true` to create streams and groups on the fly.
+- Set `log-stream-name` to a a unique ID for the source of the log (see "CloudWatch streams and groups" above)
+
+For the first implementation, the log stream name will be:
+- *container logs*: use the fluentd tag, via the `use_tag_as_stream true` plugin setting. Unique because it includes container-id.
+- *node logs* (audit, infrastructure): use fluentd tag + node-id. Unique per node and per audit/infrastructure log file.
```
jcantrill, line 178: I don't know what or how to get `node-id` but the node name is same as host name.  Landing on this which would give you journal and all audit logs grouped by node hostname:

```
<label {{.LabelName}}>
  <filter kubernetes.**>
    @type record_transformer
    <record>
      cw_group_name {{.LogGroupName }}
      cw_stream_name ${record["kubernetes"]["container_image_id"]}
    </record>
  </filter>
  <filter journal.** *audit.log>
    @type record_transformer
    <record>
      cw_group_name ${record["hostname"]}
      cw_stream_name ${tag}
    </record>
  </filter>
```

jcantrill (commented):
/hold cancel
/lgtm

PR #575, opened 2020-12-24
Installer Enhacement Proposal: Manifests from STDIN

oglok:
This OpenShift enhancement proposes to allow the installer to read a stream of manifests from STDIN on the main targets that consume manifests.


enhancements/installer/manifests-from-stdin.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Change the existing workflow to deploy an OpenShift cluster.
+- Couple the Openshift Installer with Kustomize or other manifest tooling
+
+
+## Proposal
+
+Modify the openshift-install binary to accept input from STDIN for all targets which consume manifests.  For example:
+
+- Create manifests from customized install-config.yaml: `kustomize build github.com/myorg/myBlueprint/install-config-kustomization | openshift-install create manifests -f -`
```
dhellmann, line 55: Is the intent to make it easier to pass *extra* manifests to the installer, or to make it easier to modify the manifests created by the installer?

copejon: More the later - in the example, the yaml generated by the `kustomize build` would adhere to the current install-config scheme.  A more mundane/contrived example would be `openshift-install create install-config && cat install-config.yaml | openshift-install create manifests -f -`

The idea is that we'd be supporting a simple and powerful pipeline model which could for instance take install-configs created via kustomize.

dhellmann: I guess the source of my confusion is that I didn't really consider the `install-config.yaml` file as "manifest" in the same way that the files created by `openshift-install create manifests`. Would you expect to be able to pass the output of `create manifests` through `kustomize`?

oglok: The idea would be to have a generic `install-config.yaml` describing at a high level view, how the cluster looks like (number of masters and workers, flavor of VMs, etc.) and use kustomize to overwrite specific information from the site (cluster name, domain, ipmi addresses, etc.). Using the base `install-config.yaml` + kustomize patch, the result shall be use to generate the installer manifests.

This approach will give organizations the ability to use a common recipe or `install-config.yaml` and customize only certain parameters. This would be very useful in edge computing usecases at scale.

enhancements/installer/manifests-from-stdin.md
```diff
@@ -0,0 +1,114 @@
+---
+title: manifests-from-sdtin
+authors:
+  - "@oglok"
+  - "@copejon"
+reviewers:
+  - "@fzdarsky"
+approvers:
+  - "@crawford"
```
sdodson, line 9: Please add @staebler

sdodson: @dgoodwin too, I'm not sure if this would have value for Hive.

enhancements/installer/manifests-from-stdin.md
```diff
@@ -0,0 +1,114 @@
...
+## Summary
+
+In many occasions, users need to customize the install-config target and/or the manifests generated by the OpenShift Installer in a reproducible manner. The most popular tool to allow this type of customization is “Kustomize”, and it is already well-known by the Kubernetes ecosystem. The result of any “Kustomization” is sent to standard output as a stream of manifests.
+
+This OpenShift enhancement proposes to allow the installer to read a stream of manifests from STDIN on the main targets that consume manifests.
+
+
+## Motivation
+
+Infrastructure-as-Code (e.g. GitOps) is an operational best practice for provisioning and managing large numbers of clusters and services. A typical IaC workflow involves pulling Kubernetes resource manifest templates of a specific version from a `git` repo, rendering those templates for a given cluster instance, and then running `openshift-install` (resp. `oc apply`) on the rendered manifests to install (resp. update) a cluster.
```
sdodson, line 35: I don't claim to be a GitOps aficionado, so perhaps this is a naive question, but what precludes leaving this entirely outside of the purview of the Installer? Would you not want to source control the actual rendered assets that were used rather than the template and mutations? What if the actual manifests created by `openshift create manifests` change over time given the same install-config.yaml input? Is that desirable or undesirable regardless of whether or not it breaks your pipeline.

enhancements/installer/manifests-from-stdin.md
sdodson: I agree I was thrown off by 'manifests' I assumed it was limited in scope to the output of `openshift-install create manifests` perhaps we should be more generic and speak of "Installer inputs" or "Installer State"?

enhancements/installer/manifests-from-stdin.md
fzdarsky: Both approaches (version controlling the one-off generated manifests vs the install-config.yaml + mutations) are valid and used, each with pros and cons.
In both cases, users would want to pin versions of all installation artefacts (`openshift-install`, `kustomize`, ...) matching the to-be-generated/consumed manifests, such that re-deploying the same git tag (e.g. for disaster recovery) at a later point in time yields the exact same result. In both cases, changes to newer installer versions may require updates to manifests.
One advantage of the "install-config + mutations" approach with `kustomize` is that it minimizes the delta that needs to be (version) controlled. So even when manifests created by the `create manifests` target drift (it's not a stable API after all), the "patches" are more likely to apply cleanly still.

enhancements/installer/manifests-from-stdin.md
dhellmann: Based on the examples of how this would be used, I like "installer inputs" because it differentiates from what the installer reads and what it produces as part of its work.

@oglok describes how install-config.yaml would be read. What about other inputs? Do we want it to be able to read extra content, like machine configs, from stdin as well?

sdodson: I'd assumed that https://github.com/openshift/enhancements/pull/575/files#diff-c776d0a2f45ebd59d0f77c302549387a07069285f130da08a3a66aa24553792cR56-R57 implied we were mutating the output of `openshift-install create manifests` and `openshift-install create ignition-configs` respectively in the full `install-config` `manifests` `ignition-configs` `cluster` create workflow. Maybe I've misunderstood.

oglok: I can certainly replace "manifests" by "inputs". The workflow would require to apply "kustomizations" on the entire workflow. 

- Kustomize overlay (cluster name, number of nodes...) + install-config.yaml | `openshift-install create manifests`
- Kustomize overlay (adding MachineConfigs, patching IngressReplicas...) + openshift installer manifests | `openshift-install create (ignition-configs | cluster )`

If the user wants to deploy following the UPI approach, he/she would create the ignition files to inject in RHCOS. If the user wants to deploy using any IPI available option,  he/she would have to just create the cluster.

enhancements/installer/manifests-from-stdin.md
oglok: Do you want me to add them as reviewers? or approvers?

enhancements/installer/manifests-from-stdin.md
dhellmann: I have limited experience with kustomize, but the only way I've ever seen it used requires the patch files and base files to be in a common directory hierarchy (for example, https://github.com/metal3-io/baremetal-operator/tree/master/config). I've had it report errors when there was not a common parent directory. So, IIUC, modifying the files produced by the installer would involve a separate step to set up the right directory hierarchy (by moving or copying files) before running kustomize.

Since we can already use a multi-step approach with no changes to the installer, I think that means we the proposed approach would only apply to things that are pure inputs to the installer, and we would be assuming that the changes made by kustomize would propagate from those inputs to the manifests the installer generates. A more detailed example of a use case, describing how kustomize would change something more complex than say the cluster name, would help me understand if that assumption is correct.

oglok: From kustomize v3.7.0 it's possible to create reusable "kustomizations" called `Components`.

Let's say that according to this enhacenment proposal, the output of `openshift-install create manifests` generates in the chosen directory the following structure:

- openshift/
- manifests/
- kustomization.yaml (this file will list all manifests as resources)

and we want to apply some remote overlay on top of the manifests generated by the installer. We would need to edit the generated kustomization.yaml with the following `components` section:

```
resources:                                                                                                                                                                                                      
- manifests/openshift-config-secret-pull-secret.yaml                                                                                                                                                            
- manifests/machine-config-server-tls-secret.yaml                                                                                                                                                               
- manifests/kube-system-configmap-root-ca.yaml                                                                                                                                                                  
- manifests/kube-cloud-config.yaml                                                                                                                                                                              
- manifests/etcd-signer-secret.yaml                                                                                                                                                                             
- manifests/etcd-serving-ca-configmap.yaml                                                                                                                                                                      
- manifests/etcd-service.yaml                                                                                                                                                                                   
- manifests/etcd-namespace.yaml                                                                                                                                                                                 
- manifests/etcd-metric-signer-secret.yaml                                                                                                                                                                      
- manifests/etcd-metric-serving-ca-configmap.yaml                                                                                                                                                               
- manifests/etcd-metric-client-secret.yaml                                                                                                                                                                      
- manifests/etcd-host-service.yaml                                                                                                                                                                              
- manifests/etcd-host-service-endpoints.yaml                                                                                                                                                                    
- manifests/etcd-client-secret.yaml                                                                                                                                                                             
- manifests/etcd-ca-bundle-configmap.yaml                                                                                                                                                                       
- manifests/cvo-overrides.yaml                                                                                                                                                                                  
- manifests/cluster-scheduler-02-config.yml                                                                                                                                                                     
- manifests/cluster-proxy-01-config.yaml                                                                                                                                                                        
- manifests/cluster-network-02-config.yml                                                                                                                                                                       
- manifests/cluster-network-01-crd.yml                                                                                                                                                                          
- manifests/cluster-ingress-02-config.yml                                                                                                                                                                       
- manifests/cluster-infrastructure-02-config.yml                                                                                                                                                                
- manifests/cluster-dns-02-config.yml                                                                                                                                                                           
- manifests/cluster-config.yaml                                                                                                                                                                                 
- manifests/cloud-provider-config.yaml                                                                                                                                                                          
- manifests/cloud-controller-uid-config.yml                                                                                                                                                                     
- manifests/04-openshift-machine-config-operator.yaml                                                                                                                                                           
- openshift/99_openshift-machineconfig_99-master-ssh.yaml                                                                                                                                                       
- openshift/99_openshift-cluster-api_master-user-data-secret.yaml                                                                                                                                               
- openshift/99_openshift-cluster-api_master-machines-2.yaml                                                                                                                                                     
- openshift/99_openshift-cluster-api_master-machines-1.yaml                                                                                                                                                     
- openshift/99_openshift-cluster-api_master-machines-0.yaml                                                                                                                                                     
- openshift/99_openshift-machineconfig_99-worker-ssh.yaml                                                                                                                                                       
- openshift/99_openshift-cluster-api_worker-user-data-secret.yaml                                                                                                                                               
- openshift/99_openshift-cluster-api_worker-machineset-2.yaml                                                                                                                                                   
- openshift/99_openshift-cluster-api_worker-machineset-1.yaml                                                                                                                                                   
- openshift/99_openshift-cluster-api_worker-machineset-0.yaml                                                                                                                                                   
- openshift/openshift-install-manifests.yaml                                                                                                                                                                    
- openshift/99_role-cloud-creds-secret-reader.yaml                                                                                                                                                              
- openshift/99_kubeadmin-password-secret.yaml                                                                                                                                                                   
- openshift/99_cloud-creds-secret.yaml                                                                                                                                                                          
                                                                                                                                                                                                                
components:                                                                                                                                                                                                     
- git::https://github.com/oglok/blueprint-test.git//overlay-manifests/ 
```

The manifests located in that remote location will look like:

- etcd-service-port.yaml
- etcd-host-service-port.yaml
- kustomization.yaml

_**etcd-service-port.yaml**_
apiVersion: v1                                                                                                                                                                                                  
kind: Service                                                                                                                                                                                                   
metadata:                                                                                                                                                                                                       
  name: etcd                                                                                                                                                                                                    
  namespace: openshift-etcd                                                                                                                                                                                     
  labels:                                                                                                                                                                                                       
    k8s-app: etcd                                                                                                                                                                                               
spec:                                                                                                                                                                                                           
  ports:                                                                                                                                                                                                        
    - $patch: replace                                                                                                                                                                                           
    - name: etcd                                                                                                                                                                                                
      port: 2378                                                                                                                                                                                                
      protocol: TCP      

_**etcd-host-service-port.yaml**_
apiVersion: v1                                                                                                                                                                                                  
kind: Service                                                                                                                                                                                                   
metadata:                                                                                                                                                                                                       
  name: host-etcd                                                                                                                                                                                               
  namespace: openshift-etcd                                                                                                                                                                                     
spec:                                                                                                                                                                                                           
  ports:                                                                                                                                                                                                        
    - $patch: replace                                                                                                                                                                                           
    - name: etcd                                                                                                                                                                                                
      port: 2378                                                                                                                                                                                                
      protocol: TCP   

_**kustomization.yaml**_
apiVersion: kustomize.config.k8s.io/v1alpha1  # <-- Component notation                                                                                                                                          
kind: Component                                                                                                                                                                                                 
                                                                                                                                                                                                                
patches:                                                                                                                                                                                                        
- etcd-service-port.yaml                                                                                                                                                                                        
- etcd-host-service-port.yaml  

This can be rendered by kustomize. So we propose the installer to accept this kustomized output as input:

```kustomize build generatedManifests/ | openshift-install create cluster```

Comment (oglok) [reply enhancements/installer/manifests-from-stdin.md:55]: @dhellmann do you want me to add this example to the enhancement too?

Comment (dhellmann) [reply enhancements/installer/manifests-from-stdin.md:55]: @oglok I think having a detailed example in the body of the enhancement would be a good addition.

One of the technical challenges I see with implementing this is that, as far as I know, the installer treats manifests as files rather than individual resources. Since a file can contain multiple resources, and the names of the files are both significant and not necessarily derived from the contents in an obvious way, it will be challenging to turn a single input stream from stdin into the right artifacts to replace the cached versions on the filesystem. Which is not to say it can't be done, just that it isn't obvious to me how to do it. Have you started looking at the implementation of this idea already?

Reviewing the way that kustomize is used in the cluster-baremetal-operator repository, I see that it is configure to create output files in a named directory instead of streaming the results to stdout. Using that approach may make it easier to integrate with installer. A user would run `create manifests`, then `kustomize` with the manifests directory as input, then `create cluster`.

Comment (oglok) [reply enhancements/installer/manifests-from-stdin.md:55]: Ok, I will add the example to the enhancement.

We haven't started looking at the implementation, but after one of the meetings we had with @crawford , he said it should be doable.

File: enhancements/installer/manifests-from-stdin.md (hunk 2 above)
Comment (crawford) [reply enhancements/installer/manifests-from-stdin.md:9]: Approvers

File: enhancements/installer/manifests-from-stdin.md (hunk 1 above)
Comment (crawford) [reply enhancements/installer/manifests-from-stdin.md:55]: In order for this to work, the installer would need to start keying assets based on their type and name rather than their on-disk path (since there aren't any paths when reading assets from STDIN). There's probably some annoying gotchas in this approach, but it doesn't fundamentally alter the design of the installer.

File: enhancements/installer/manifests-from-stdin.md [hunk 4]
```diff
@@ -0,0 +1,114 @@
...
+
+As a developer, I want to customize the definition of a production cluster in order to deploy a copy in my test environment. In order to facilitate this task, It would help to have available a `kustomization.yaml` with a list of the manifest files generated by the installer.
+
+### Implementation Details/Notes/Constraints
+
+The installer validates if the input stream contains all the necessary objects required for the installation process. This will ensure that the customization process has not created corrupted objects or non-valid manifests.
+
+#### Resources created by the installer
+
+The `openshift-install create manifests` will create a kustomization.yaml file that will list the resources located in the openshift and the manifest directories. The content of the file should look like this:
```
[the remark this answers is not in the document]
oglok: Sure, I think a global flag in order to trigger the creation of the kustomization.yaml file makes sense. I'll add that.

enhancements/installer/manifests-from-stdin.md
```diff
@@ -0,0 +1,114 @@
...
+
+As an administrator, I want to deploy a set of nearly identical clusters by overlaying minimal changes to a base cluster definition using a simple pipelined CLI command. If I have to apply the same specific configuration on all my clusters, I need an easy way to apply it and deploy. I want to use kustomize or any other manifest manipulation tool and pipe the output to the OpenShift Installer. 
+
+#### Story 2: Developer persona
+
+As a developer, I want to customize the definition of a production cluster in order to deploy a copy in my test environment. In order to facilitate this task, It would help to have available a `kustomization.yaml` with a list of the manifest files generated by the installer.
+
+### Implementation Details/Notes/Constraints
+
+The installer validates if the input stream contains all the necessary objects required for the installation process. This will ensure that the customization process has not created corrupted objects or non-valid manifests.
```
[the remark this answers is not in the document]
oglok: I think your comment is totally understandable based on the proposal. My intention here was to leverage the validation already made by the installer. We don't want that validation to be different from what it exists. I'll rephrase this.

yrobla:
This is an useful feature when some components need to be configured in a different way than the default one, and can cause a failure on the installation or race conditions if we rely on day 2.
Things like modifying number of replicas of the ingress operator, use different storage system for image registry, or installing components that need to be present at day 1 ... could be solved using this approach, and are needed for several use cases. See for example several modifications that we did on Akraino: https://github.com/akraino-edge-stack/kni-blueprint-pae/tree/master/profiles/production.baremetal/01_cluster-mods/manifests , https://github.com/akraino-edge-stack/kni-blueprint-pae/tree/master/base/01_cluster-mods/openshift

copejon:
@dhellmann @sdodson if you're satisfied with the most recent changes, what can we do to move this towards acceptance?

enhancements/installer/inputs-from-stdin.md
```diff
@@ -0,0 +1,182 @@
...
+
+As a developer, I want to customize the definition of a production cluster in order to deploy a copy in my test environment. In order to facilitate this task, It would help to have available a `kustomization.yaml` with a list of the manifest files generated by the installer.
+
+### Implementation Details/Notes/Constraints
+
+The installer would need to keying assets based on their type and name rather than their on-disk path. This approach does not fundamentally alter the design of the installer. It must also apply the existing validations on all of the assets as if they were provided via the filesystem.
+
+#### Resources created by the installer
+
+The `openshift-install create manifests` will create a kustomization.yaml file that will list the resources located in the openshift and the manifest directories if the flag `--kustomize` is passed. The content of the file should look like this:
```
staebler, line 145: I would rather not have the installer own code around supporting a kustomize.yaml file, particularly when the list of resources is simply all of the files.
```
(echo "resources:"; find "${TMPDIR}" -type f -not -name '\.*' -printf '- %P\n') > "${TMPDIR}/kustomization.yaml"
```

staebler (commented):
I am not convinced that there is overall much added value to the installer to support piping from STDIN. I think it introduces a number of grey areas. I have not seen use cases yet where streaming from STDIN makes the installation process significantly easier.

enhancements/installer/inputs-from-stdin.md
```diff
@@ -0,0 +1,182 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+In many occasions, users need to customize the install-config target and/or the manifests generated by the OpenShift Installer in a reproducible manner. The most popular tool to allow this type of customization is “Kustomize”, and it is already well-known by the Kubernetes ecosystem. The result of any “Kustomization” is sent to standard output as a stream of manifests.
```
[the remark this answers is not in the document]
fzdarsky: The problem is `openshift-install`'s assumption on file structure and naming when reading from the manifest directory.

Popular templating tools (kustomize, ksonnet, helm, ...) either output a single stream of manifests to stdout (from where it could be redirected to a single file, of course) or to a manifest directory, but then it's up to the tool how files get named (typically some string concatenation of the resource's GVKN) and organized into subdirs (typically: flat).

Means one cannot currently solve the common use case of templating manifests without additional scripting that reorganizes a tool's output to meet `openshift-install`'s requirements.

enhancements/installer/inputs-from-stdin.md
```diff
@@ -0,0 +1,182 @@
...
+
+- Change the existing workflow to deploy an OpenShift cluster.
+- Couple the Openshift Installer with Kustomize or other manifest tooling
+
+
+## Proposal
+
+Modify the OpenShift Installer to accept inputs from STDIN for all targets. For example: 
+
+- Create manifests from customized install-config.yaml: `kustomize build github.com/myorg/myBlueprint/install-config-kustomization | openshift-install create manifests -f -`
```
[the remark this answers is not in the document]
fzdarsky: The expectation is that the installer can read manifests from stdin like it currrently can read from a `--dir`. The current proposal is to express this to use `-f -` as argument, but if something like `--stdin` makes more sense, that'd be perfectly fine.

Just like the installer doesn't know a priori whether the `--dir` contains an install-config.yaml or additional manifests, it wouldn't know reading from `--stdin`. The assumption we've made is that once the resources read from stdin are mapped back into the same data structure when reading from `--dir`, the following logic would be identical.

Re `openshift-install --kustomization create manifests --dir=$TMPDIR`: This could work and was in fact part of an earlier proposal. However, we've been given the guidance by the architects _not_ to propose an approach that would make the installer opinionated on a given templating tool or even to expect kustomize libs to be added to the installer. This is why we're proposing the stdin approach now.

enhancements/installer/inputs-from-stdin.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Non-Goals
+
+- Change the existing workflow to deploy an OpenShift cluster.
+- Couple the Openshift Installer with Kustomize or other manifest tooling
+
+
+## Proposal
+
+Modify the OpenShift Installer to accept inputs from STDIN for all targets. For example:
```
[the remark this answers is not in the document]
fzdarsky: Just to clarify: It is not only about _adding_ manifests, but also _changing_ installer-generated manifests through config overlays or templating. And I agree the installer doesn't have to / shouldn't be involved _how_ a user achieves this. But the installer can _facilitate_ this by offering some way of reading manifests without mandating a file naming and directory structure convention.

enhancements/installer/inputs-from-stdin.md
```diff
@@ -0,0 +1,182 @@
...
+
+Modify the OpenShift Installer to accept inputs from STDIN for all targets. For example: 
+
+- Create manifests from customized install-config.yaml: `kustomize build github.com/myorg/myBlueprint/install-config-kustomization | openshift-install create manifests -f -`
+- Create ignition files from customized set of manifests: `kustomize build ~/myBlueprint/01_manifests | openshift-install create ignition-configs -f -`
+- Create cluster from customized set of manifests: `kustomize build ~/myBlueprint/01_manifests | openshift-install create cluster -f -`
+
+The syntax and workflow will be similar to `oc apply -f -`.
+
+A straight example of a use case is explained next. From Kustomize v3.7.0 it's possible to create reusable customizations called `Components`. Acording to this enhancement proposal the output of `openshift-install --kustomization create manifests` will generate the following structure in the chosen directory:
```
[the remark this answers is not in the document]
fzdarsky: Thanks for pointing this out. We added this as response to comment https://github.com/openshift/enhancements/pull/575#discussion_r555414105 to add a global flag to enable/disable the behavior.

We need to further clarify this and should probably not call the flag `--kustomization` for the reason I stated in https://github.com/openshift/enhancements/pull/575#discussion_r581813012 but something less opinionated.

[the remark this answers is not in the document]
oglok: The `--kustomization` is not right here. `--stdin` would be a better fit, since we don't want to be opinionated on the tool that the user must leverage to perform customizations on installer inputs.

enhancements/installer/inputs-from-stdin.md
```diff
@@ -0,0 +1,182 @@
...
+- Change the existing workflow to deploy an OpenShift cluster.
+- Couple the Openshift Installer with Kustomize or other manifest tooling
+
+
+## Proposal
+
+Modify the OpenShift Installer to accept inputs from STDIN for all targets. For example: 
+
+- Create manifests from customized install-config.yaml: `kustomize build github.com/myorg/myBlueprint/install-config-kustomization | openshift-install create manifests -f -`
+- Create ignition files from customized set of manifests: `kustomize build ~/myBlueprint/01_manifests | openshift-install create ignition-configs -f -`
```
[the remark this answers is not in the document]
fzdarsky: We were assuming input is either from stdin or from disk. Mixing both is an interesting idea, though. Yes, I think making stdin take precedence would make sense in this case.

Once resources read via stdin are mapped back (via the GVKN) to the installer's internal data structure, the run order would be identical to the installer's default for any resource known to the installer (resp. generated by it). For additional resources, one way could be to simply default to a 99 run order (resp. order them last). Another could be to sort additional manifests in according to the order in the stdin stream.

fzdarsky:
> I am not convinced that there is overall much added value to the installer to support piping from STDIN. I think it introduces a number of grey areas. I have not seen use cases yet where streaming from STDIN makes the installation process significantly easier.

Thank you for your review, @staebler! I hope I could shed some light on the added value in the responses to your detailed comments. Gist is that the installer imposes convention on file naming and directory structure that tooling for config overlays/templating popular with GitOps and Infrastructure as Code isn't aware of. Therefore, users currently need to implement their own logic how to adapt one to the other and our own engineering and field teams have created lots of snowflake solutions for this.

enhancements/installer/inputs-from-stdin.md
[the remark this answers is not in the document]
staebler: Then let's consider as an alternative whether openshift-install can work without some of those assumptions on file structure and naming. If openshift-install is going to support streaming from STDIN, then it would need to rid itself of those assumptions for that input path anyway. So this is not a compelling argument to me to allow for reading from STDIN.

enhancements/installer/inputs-from-stdin.md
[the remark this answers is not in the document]
staebler: There is more in the assets directory than just manifests. Is your expectation that **everything** in the assets directory is going to be sent in via STDIN? Or is, for example, the asset store going to remain in the assets directory?

You are correct that the installer does not know a priori whether the assets director contains and install-config.yaml or additional manifests. The installer builds the asset graph up to the targets desired in the command and tries to load from the asset directory whatever assets are needed. The installer relies on the names of files to ascertain whether the files for a particular asset are on disk. It is not clear to me how the "mapp[ing] back into the same data structure" is supposed to work without the file names. Let's take the ignition config assets as an example. How is the installer expected to differentiate between the master ignition config and the worker ignition config when read from STDIN?

Re `openshift-install --kustomization create manifests --dir=$TMPDIR`: You can ignore the use of the `--kustomization` flag here. That was lifted from on of the use cases later in the document. It has no bearing on the behavior of the installer as it relates to input.

enhancements/installer/inputs-from-stdin.md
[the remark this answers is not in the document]
staebler: I don't follow how your argument relates to whether we accept STDIN for all commands or only for the `create manifests` command.

staebler:
> > I am not convinced that there is overall much added value to the installer to support piping from STDIN. I think it introduces a number of grey areas. I have not seen use cases yet where streaming from STDIN makes the installation process significantly easier.
> 
> Thank you for your review, @staebler! I hope I could shed some light on the added value in the responses to your detailed comments. Gist is that the installer imposes convention on file naming and directory structure that tooling for config overlays/templating popular with GitOps and Infrastructure as Code isn't aware of. Therefore, users currently need to implement their own logic how to adapt one to the other and our own engineering and field teams have created lots of snowflake solutions for this.

Can you give me some concrete examples of installer-imposed conventions that are onerous for GitOps and Infrastructure as Code?

Here are the only conventions that I am aware of that the installer imposes for manifests.
1. The manifests must be in either the manifests folder or the openshift folder in the assets directory. There is no practical difference between the two folders, so manifests can be placed arbitrarily in either.
2. The manifests must have one of the following extensions: yaml, yml, or json.

Would it be helpful if the installer read recursively through the manifests and openshift folders?

PR #576, opened 2021-01-06
Move ES cert management into Elasticsearch Operator

ewolinetz:
This proposal seeks to outline moving cert management from the Cluster Logging and Jaeger operators into the Elasticsearch Operator to make that the single source of truth for certificates when communicating with elasticsearch.

It also highlights the mechanism by which components that use the EO can have mTLS certificates injected into a secret for them to continue using mTLS for communicating with an elasticsearch cluster.

/cc @jcantrill @alanconway

### DIFF
--- a/enhancements/cluster-logging/elasticsearch-cert-management.md
+++ b/enhancements/cluster-logging/elasticsearch-cert-management.md
@@ -1,0 +1,230 @@
+---
+title: elasticsearch-cert-management
+authors:
+  - "@ewolinetz"
+reviewers:
+  - "@jcantrill"
+  - "@bparees"
+  - "@alanconway"
+  - "@jpkrohling"
+approvers:
+  - "@jcantrill"
+  - "@bparees"
+  - "@alanconway"
+  - "@jpkrohling"
+creation-date: 2021-01-05
+last-updated: 2021-01-20
+status: implementable
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# elasticsearch-cert-management
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Migration plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Currently to use the Elasticsearch Operator, one would need to generate secrets in a specific manner which creates a high amount of complexity for use.
+This proposal seeks to outline a mechanism where the Elasticsearch Operator creates and maintains these certificates and the elasticsearch/kibana secret instead of other operators (e.g. Cluster Logging and Jaeger), and allows an annotation mechanism on secrets for injecting required keys and certificates for mTLS with Elasticsearch.
+
+## Motivation
+
+Currently our managed Elasticsearch cluster requires anything using our operator to generate, manage, and provide certs and secrets for the operator and its cluster to use.
+This leads to many potential sources to create certificates and can cause a gap between what the managed Elasticsearch requires and what something may provide.
+This would instead make the Elasticsearch Operator a single source of truth for how certificates would be generated.
+
+### Goals
+The specific goals of this proposal are:
+
+* Outline the responsibility for the Elasticsearch Operator regarding certificates
+
+* Determine an annotation that the Elasticsearch Operator will watch for on secrets in the same namespace as an existing elasticsearch CR
+
+* Discuss upgrade strategies from the current LTS mechanism.
+
+We will be successful when:
+
+* Users of Elasticsearch Operator no longer need to maintain certificates for Elasticsearch and can have certificates injected into their secrets
+
+* There is a clear path forward for automatic upgrades
+
+### Non-Goals
+
+* This does not seek to implement work to adhere to the cluster level TLS definitions, however this does make the implementation less complex for the Elasticsearch use-case.
+
+### Current workflow vs Proposed
+
+Currently, a consumer like Jaeger or Cluster Logging is responsible for the following for their ES cluster:
+
+1. Generating and managing their own keys/certificates including those that are used in Elasticsearch as server serving and inter-cluster so that Elasticsearch can join and form clusters.
+1. Generating and managing any client keys/certs to communicate with Elasticsearch. This would include their forwarders, and if applicable, their UIs.
+1. Generating and managing secrets to be used by: EO, Elasticsearch, IndexManagement cronjobs, and any clients as denoted in point 2.
+1. Creating and updating their elasticsearch CR(s) so that EO can manage their cluster.
+
+With this proposal the workflow would be:
+
+1. Create their elasticsearch CR with an annotation to denote EO will be managing certs for the cluster.
+1. Create an empty secret in the same namespace as the elasticsearch CR with a name that matches the annotation they will use.
+1. Add an additional annotation on the elasticsearch CR for each client they need to have certs injected into a specific secret (EO will be responsible for rotating the cert).
+
+For example, CLO would use the following annotations to tell EO to do cert management and also inject client keys/certs into the `fluentd` secret with the CN `system.logging.fluentd`:
+
+```yaml
+apiVersion: logging.openshift.io/v1
+kind: Elasticsearch
+metadata:
+  name: elasticsearch
+  annotations:
+    logging.openshift.io/elasticsearch-cert-management: true
+    logging.openshift.io/elasticsearch-cert/fluentd: system.logging.fluentd
+```
+
+Similarly, Jaeger would do the following to have a client cert with the CN `user.jaeger` injected into the secret `jaeger`:
+
+```yaml
+apiVersion: logging.openshift.io/v1
+kind: Elasticsearch
+metadata:
+  name: jaeger-cluster
+  annotations:
+    logging.openshift.io/elasticsearch-cert-management: true
+    logging.openshift.io/elasticsearch-cert/jaeger: user.jaeger
+```
+
+## Proposal
+
+Instead of the current implementation where any consumer of the Elasticsearch Operator is responsible for providing certificates and secrets for Elasticsearch and the operator to use, we want to shift this responsibility to the Elasticsearch Operator instead.
+
+This will greatly improve the use case of the operator and make the operator the single source of truth for its certificates.
+
+To allow other components to still be able to communicate with the managed cluster via mTLS the operator will watch for a specific annotation on the elasticsearch CR and will inject into a specified secret that is in the same namespace as the CR.
+
+As part of the responsibility for generating, managing, and injecting these certificates, the EO would also be responsible for rotating the certificates and updating the content of the secrets when necessary.
+
+### User Stories
+
+#### As an user of EO, I do not want to have to generate and provide secrets for the Elasticsearch cluster
+
+We want to further simplify using the Elasticsearch Operator and create a single source of truth for the certs the cluster uses.
+
+#### As a consumer of EO, I want a mechanism to have certificates injected into a secret for me to continue mTLS communcation with my Elasticsearch cluster
+
+We want to continue allowing mTLS as this supports the Cluster Logging and Jaeger use cases, however they shouldn't need to do more than notify they need certs injected by EO.
+
+### Implementation Details
+
+#### Assumptions
+
+* The Elasticsearch Operator will implement this in golang (rather than the current script used in CLO)
+
+* The operator will need to be able to store generated certificates for each namespace/unique Elasticsearch cluster so that there is no security concerns of clusters cross communicating.
+
+#### Security
+
+Ensuring that signing certificates and configurations are isolated for different clusters is a must for this implementation. It will also need to be able to do this thread safe manner.
+
+### Risks and Mitigations
+
+## Not Breaking Other Operators who Depend on EO
+
+Currently there are two operators that depend on the Elasticsearch Operator and would need to ensure that their legacy functionality [read: their continuing to do cert management] is not broken.
+For instances where one operator is upgraded along with EO, but the other is not, we would need to provide an opt-in model for Elasticsearch to perform cert management (if we make it opt-out, we would break the legacy functionality).
+
+The most simple mechanism for this would be to gate Elasticsearch Operator performing cert management based on an annotation on the elasticsearch CR.
+
+## Design Details
+
+From a user perspective, there should be no change in any of the CRDs. The only major changes would be no longer generating secrets for EO to use, annotating their elasticsearch CR, and providing to an empty secret in the case they want to communicate with ES in a mTLS way.
+
+### elasticsearch cr annotation
+
+We will require an annotation value for opting in to the Elasticsearch Operator generating and managing certificates for the particular cluster:
+
+```yaml
+apiVersion: logging.openshift.io/v1
+kind: Elasticsearch
+metadata:
+  name: elasticsearch
+  annotations:
+    logging.openshift.io/elasticsearch-cert-management: true
+```
+
+### secret injection annotation
+
+Note: to utilize this will require the above annotation is also on the elasticsearch CR.
+For the sake of simplifying the example of this annotation, it has been left off below.
+
+We will add an annotation value to check for on the elasticsearch CR to denote that we would need to inject keys into a named secret:
+
+```yaml
+apiVersion: logging.openshift.io/v1
+kind: Elasticsearch
+metadata:
+  name: elasticsearch
+  annotations:
+    logging.openshift.io/elasticsearch-cert/fluentd: system.logging.fluentd
+```
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: fluentd
+type: Opaque
+data: {}
+```
+
+Where `fluentd` in the key name is the name of the secret and the value will be optional (defaults to the name of the secret) and corresponds to the CN used for the generated certificate.
+
+The keys used for injecting into the secret will be `tls.crt`, `tls.key`, and `ca-bundle.crt` to corespond with the [log forwarding proposal](./cluster-logging-log-forwarding.md).
+
+
+### Test Plan
+
+#### Unit Testing
+
+* We will need to ensure we have unit tests that verify certificates are generated by the EO
+
+* We will need to ensure we have unit tests that verify secrets are generated by EO with the correct certificates
+
+* We will need to ensure we have unit tests that verify EO generates and injects into secrets specified in an annotation
+
+#### Integration and E2E tests
+
+* Given we are moving logic from one operator to another, we need to ensure that there are no regressions with communication between Fluentd, Kibana, IM cronjobs, and Elasticsearch (this should be handled by our smoke test).
+
+### Graduation Criteria
+
+#### GA
+
+* End user documentation
+* Sufficient test coverage
+* Sufficient LTS upgrade strategy
+
+### Version Skew Strategy
+
+We need to have a proper upgrade strategy from our LTS version to the version this is released on (currently 4.6).
+
+## Implementation History
+
+| release|Description|
+|---|---|
+|5.2| **GA** - Initial release
+
+## Drawbacks
+
+Given the OCP cert signing service does not allow for mTLS configuration with custom fields that the ES cluster requires, we are unable to use it so we have to implement something similar to a pre-existing service.https://bugzilla.redhat.com/show_bug.cgi?id=1925627
+None
+
+## References
+
+* [Preliminary CLO script to golang reimplementation](https://github.com/openshift/cluster-logging-operator/pull/831)


enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,158 @@
...
+## Motivation
+
+Currently our managed Elasticsearch cluster requires anything using our operator to generate, manage, and provide certs and secrets for the operator and its cluster to use. This leads to many potential sources to create certificates and can cause a gap between what the managed Elasticsearch requires and what something may provide. This would instead make the Elasticsearch Operator a single source of truth for how certificates would be generated.
+
+### Goals
+The specific goals of this proposal are:
+
+* Outline the responsibility for the Elasticsearch Operator regarding certificates
+
+* Expose the desired annotation that the operator would watch for on secrets where an elasticsearch CR exists
```
syedriko, line 47: on secrets in the namespace where an elasticsearch CR exists?

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,158 @@
...
+
+We want to continue allowing mTLS as this supports the Cluster Logging and Jaeger use cases, however they shouldn't need to do more than notify they need certs injected by EO.
+
+### Implementation Details
+
+#### Assumptions
+
+* The Elasticsearch Operator will implement this in golang (rather than the current script used in CLO)
+
+* The operator will need to be able to store generated certificates for each namespace/unique cluster so that there is no security concerns of clusters cross communicating.
```
syedriko, line 83: Elasticsearch cluster?

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,158 @@
...
+
+#### Security
+
+Ensuring that signing certificates and configurations are isolated for different clusters is a must for this implementation. It will also need to be able to do this thread safe manner.
```
syedriko, line 87: "in a thread safe manner". Thread safety is not really a security concern though.

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,158 @@
...
+kind: Secret
+metadata:
+  name: fluentd
+  annotations:
+    secret.openshift.io/elasticsearch-cert-component-name: system.logging.fluentd
+type: Opaque
+data: {}
+```
+
+Where the value will be optional (defaults to the name of the secret) and corresponds to the CN used for the generated certificate.
```
syedriko, line 110: We need to specify the contents of these secrets as it's part of the contract. The first decision to make is whether to keep compatibility with the current secrets. If we're keeping compatibility, the annotations need to be more expressive. Specifying just the CN is not enough, we need the names of the CA[s], the certs and the keys. Some secrets currently hold multiple copies of CA and several cert/key pairs. Please take a look at https://github.com/openshift/cluster-logging-operator/blob/c36a5589e7bc272f44153886ab1e4aba645d0638/pkg/k8shandler/certs.go#L73, this is a definition of what's currenly found in the secrets with certs as generated by the CLO.
If we're breaking compatibility with the existing format of the secrets, we can say that there's going to be a separate secret for every cert we need to generate, the name of the secret is going to be the CN unless overridden by the annotation (would need to chase down the specs to make sure any CN is representable as at least one of these), and the secret is going to contain three entries: ca, cert and key. Then there's an outlier that the kibana-proxy secret that also has an entry for its session secret, but we can find a place for it somewhere.

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,158 @@
...
+|---|---|
+|5.2| **GA** - Initial release
+
+## Drawbacks
+
+None
+
+## Alternatives
+
+None
```
syedriko, line 150: 1. Leave the contract between the CLO and EO and Jaeger and EO the way it is and document it more clearly.
2. Take a step further and move certificate generation into an OCP-cluster-wide CA, as there's nothing specific to EO, CLO or Jaeger in this code.

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Implementation History
+
+| release|Description|
+|---|---|
+|5.2| **GA** - Initial release
+
+## Drawbacks
+
+None
```
syedriko, line 146: Arguably this EP doesn't go far enough, as the proposed functionality is not specific to EO, cluster logging or Jaeger, but belongs rather in in a OCP-cluster-wide CA.

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: this is a little clunky, i'll rephrase. essentially the operator will watch secrets in namespaces where someone has created an elasticsearch CR. If it has the annotation then we'll perform the cert generation and then inject it back into that secret.

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: yep, i'll clarify

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: can't it be if we're reading in certificates that we've generated for another cluster? we want to ensure things are completely isolated

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: i was thinking we would just have a `cert`, `key`, `ca` entry which should cover most cases. Kibana can be excluded as we wouldn't use an opt in model for it given its managed by the Elasticsearch Operator (like-wise, the ES secret wouldn't need to be opt-in)

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: can you expand more on this? we've tried to get mTLS functionality into the OCP cert signing service and it seems that isn't going to be implemented so we would need to still sign our own certs

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: true, I should be clear that the alternative is to leave it as it currently stands... "None" is misleading

enhancements/cluster-logging/elasticsearch-cert-management.md
syedriko: I thought under thread safety you meant absence of race conditions or deadlocks. Can you help me understand how k8s objects from a namespace and a thread in the operator are related? On the surface it seems it should be enough during reconcile to look at the namespace the objects being reconciled are from and keep things isolated based on that. On the other hand, would running an EO instance per Elastic cluster alleviate such concerns? I'm sure I'm missing important details.

enhancements/cluster-logging/elasticsearch-cert-management.md
syedriko: I don't really know anything about the OCP cert signing service, it's just an observation that what we're talking about in this EP is a piece of generic functionality and there may be other components that sign their own certs out there that could use this. But if you've been down that road and its a dead end (for now?) so be it. It might be worth mentioning this in the EP.

enhancements/cluster-logging/elasticsearch-cert-management.md
jpkrohling: That's similar to how the service-ca works, right?

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,158 @@
...
+
+* There is a clear path forward for automatic upgrades
+
+### Non-Goals
+
+* This does not seek to implement work to adhere to the cluster level TLS definitions, however this does make the implementation less complex for the Elasticsearch use-case.
+
+## Proposal
+
+Instead of the current implementation where any consumer of the Elasticsearch Operator is responsible for providing certificates and secrets for Elasticsearch and the operator to use, we want to shift this responsibility to the Elasticsearch Operator instead. This will greatly improve the use case of the operator and make the operator the single source of truth for its certificates.
```
jpkrohling, line 63: I don't think this is going to be a breaking change, but I don't remember how it's all tied together from the top of my head. Could you state explicitly in the doc that we will still be able to use the current mechanism, as we have two release streams that we'd not change to use this new feature?

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: @jpkrohling correct

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: my thought was given the Elasticsearch operator can process es events from multiple namespaces it may be possible that it is trying to resolve two different namespaces at the same time. so in that case it would be processing certificates "at the same time". this may not end up being a valid concern, but it was my original thought behind the original statement.

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: currently Jaeger Operator runs a script that will create the certs and secrets that are provided for the elasticsearch cluster that is to be used. Off the top of my head I'm not sure how we could support the currently used model and this proposed one at the same time.

I would suggest that the Jaeger Operator not use an updated EO version, but given it is a dependency of (at least) two operators that isn't a likely solution.

jpkrohling: We'll need a good coordination then before this feature gets released, and include the Maistra team (cc @knrc and @rcernich) in the loop.

rcernich: Thanks for the head's up @jpkrohling.  Maistra has no direct integration with elasticsearch.  Anything we might configure is done through the jaeger cr.

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: @syedriko does what i added address your comments?

syedriko: @ewolinetz In the Drawbacks section, yes ;) I think we really need to nail down the contents of the secrets ^^^ EO is going to populate as it's part of the contract with the its consumers.

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -143,11 +143,11 @@ We need to have a proper upgrade strategy from our LTS version to the version th
 
 ## Drawbacks
 
-None
+Given the OCP cert signing service does not allow for mTLS configuration with custom fields that the ES cluster requires, we are unable to use it so we have to implement something similar to a pre-existing service.
 
 ## Alternatives
 
-None
+Stay with the crrent implementation where any consumer of the Elasticsearch Operator is responsible for cert and secret management.
```
jpkrohling, line 150: I see now. If we don't use the annotation at all, we can just use our current strategy, right? We'll eventually move to this new approach, but there will be scenarios where the ES Operator will be at a newer version than the Jaeger, so, wanted to make sure they'll still work.

ewolinetz: not quite, "Alternatives" would be alternatives to the enhancement... not alternatives to opting in. This would still be a coordinated effort between Elasticsearch Operator, Cluster Logging Operator, and Jaeger Operator.

I can try to think of a way to somehow opt out for the Jaeger use case while they try to catch up... that may be what we need to do

jpkrohling: My concern is that we have currently 2 streams out there, 1.17 and 1.20, both of which currently provision ES clusters generating certs on their own. Just releasing an ES Operator that breaks those installations doesn't sound like a good plan: even if we decide to backport the required changes to those branches, we still risk having folks on an older Z version that is going to break.

To be clear: the main goal for this proposal is sound and I look forward to offloading the cert management from the Jaeger Operator to the ES Operator, but we do have a good installation base out there to take care of.

ewolinetz: Ack, 100% agree and understand. Let me figure out and add something about opting in (perhaps it can be an annotation on the elasticsearch CR) so we can preserve legacy systems and not introduce a huge break

ewolinetz: @jpkrohling added some new language to hopefully capture and address this concern. Please let me know if this makes sense.
@jcantrill please read though this too as it would change steps CLO needs to take (likely makes this easier)

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: @syedriko I'm not sure I follow... it would be the same as they are currently populated today in CLO because the purpose is that of communicating with ES with mTLS (and ES will expect certain domain name patterns so that they are correctly matched to backing roles for authn/authz)

syedriko: @ewolinetz The contents of the secrets CLO populates is irregular - the names and number of CAs, the names of the cert and key are different in different secrets. I captured these at https://github.com/openshift/cluster-logging-operator/blob/c36a5589e7bc272f44153886ab1e4aba645d0638/pkg/k8shandler/certs.go#L73. I'm figuring EO will populate secrets in a uniform way, in which case we need to document how.

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -88,9 +88,15 @@ Ensuring that signing certificates and configurations are isolated for different
 
 ### Risks and Mitigations
 
+## Not Breaking Other Operators who Depend on EO
+
+Currently there are two operators that depend on the Elasticsearch Operator and would need to ensure that their legacy functionality [read: their continuing to do cert management] is not broken. For instances where one operator is upgraded along with EO, but the other is not, we would need to provide an opt-in model for Elasticsearch to perform cert management (if we make it opt-out, we would break the legacy functionality).
+
+The most simple mechanism for this would be to gate Elasticsearch Operator performing cert management based on an annotation on the elasticsearch CR.
```
jpkrohling, line 95: Sounds good!

jpkrohling (approved):

enhancements/cluster-logging/elasticsearch-cert-management.md
jcantrill: Is there not a path where by which we can support both where one takes precedence over the other.  For example, if you tell me the secret, then I will use it unless it has say an annotation on it, then the EO injects the files into the secret?  I believe the convention now is we just expect the secret to exist and be named the same as the `elasticsearch` resource

enhancements/cluster-logging/elasticsearch-cert-management.md
jcantrill: In lieu of these values, why not use those defined for CLF secrets.  We had a lengthy discussion in CLF before we settled on: tls.key, tls.crt, ca-bundle.crt  or something like that

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,177 @@
...
+
+We will require an annotation value for opting in to the Elasticsearch Operator generating and managing certificates for the particular cluster:
+
+```
+apiVersion: logging.openshift.io/v1
+kind: Elasticsearch
+metadata:
+  name: elasticsearch
+  annotations:
+    logging.openshift.io/elasticsearch-cert-management: true
```
jcantrill, line 128: The pattern elsewhere (e.g. cert gen service we use for metrics) to tell the service which secret will be injected in lieu of watching all secrets

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: @jcantrill that's file, the document doesn't call out any specific values yet, but I can add them based on what we had in the other enhancement proposal

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: how would we know who created the secret though? we would see that there is a secret out there which is the expected outcome no matter who is managing certs...

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: it does that because it creates certs specific for that service. this is the annotation that would be on the elasticsearch CR, which follows a similar pattern to what you're describing... we don't need to know the name of the secret to inject into because we will be responsible for creating the secret here (and given it needs to match the name of the elasticsearch CR name, we can infer it)

enhancements/cluster-logging/elasticsearch-cert-management.md
jcantrill: After considering my ask here https://github.com/openshift/enhancements/pull/576/files#r566939322 , why does the caller who needs the secret to contact the service need to have any notion of what the CN will be?  It just needs certs that allow it to talk to the service.  As an aside, this task will be tricky because it presumably requires a cluster restart whenever a new client is added

enhancements/cluster-logging/elasticsearch-cert-management.md
jcantrill: > it does that because it creates certs specific for that service. this is the annotation that would be on the elasticsearch CR, which follows a similar pattern to what you're describing...

If we make providers define the name thats the trigger that indicates and sets expectations the operator is managing the content of the secret, or at least the keys it provides

> we don't need to know the name of the secret to inject into because we will be responsible for creating the secret here (and given it needs to match the name of the elasticsearch CR name, we can infer it)

This statement sounds contrary to at least what is described above which is there is some secret, presumably created by the CLO (e.g. fluentd) and not EO.  The CLO (or any client) needs to know the name of this secret because its used to mount into some other service (e.g. fluentd).  It makes more sense for the requestor to tell the provider where to inject certs, for how else would we know the name of the created secret?   This is why the cert service pattern works well for consumers.

jcantrill (commented):
/hold

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: > This statement sounds contrary to at least what is described above which is there is some secret, presumably created by the CLO (e.g. fluentd) and not EO. The CLO (or any client) needs to know the name of this secret because its used to mount into some other service (e.g. fluentd). It makes more sense for the requestor to tell the provider where to inject certs, for how else would we know the name of the created secret? This is why the cert service pattern works well for consumers.

You had responded on the annotation for the es CR so I thought you were referring to that...

```
apiVersion: v1
kind: Secret
metadata:
  name: fluentd
  annotations:
    secret.openshift.io/elasticsearch-cert-component-name: system.logging.fluentd
type: Opaque
data: {}
```

so given the idea that we annotate the elasticsearch CR to follow the pattern of the cert signing service, how would we get the cn and the name of the secret to inject into - something like the following?
```
apiVersion: logging.openshift.io/v1
kind: Elasticsearch
metadata:
  name: elasticsearch
  annotations:
    logging.openshift.io/elasticsearch-cert-management: true
    logging.openshift.io/client.secret/system.logging.fluentd: fluentd
```

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Implementation Details
+
+#### Assumptions
+
+* The Elasticsearch Operator will implement this in golang (rather than the current script used in CLO)
+
+* The operator will need to be able to store generated certificates for each namespace/unique Elasticsearch cluster so that there is no security concerns of clusters cross communicating.
+
+#### Security
```
jcantrill, line 85: Following up here as I don't know if we addressed it.  We need to either express a way for users to request an ES permission set/role or explicitly list what permission/set or role you get when get these certs.  Do we need to be concerned with restricting how can ask for certs or is it sufficient to say if you have the permissions to request then that is restrictive enough

enhancements/cluster-logging/elasticsearch-cert-management.md
jcantrill: The existing pattern is the requester creates the secret and the service injects and knows about it because of the annotation on the service like https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/fluentd.go#L79-L96
> 
> so given the idea that we annotate the elasticsearch CR to follow the pattern of the cert signing service, how would we get the cn and the name of the secret to inject into - something like the following?
> 

Why does the requester care about the CN name?  Why cant the provider generate a CN name using a known combination of NS, ES name, secret name (e.g. openshift-logging.elasticsearch.fluentd) or similiar?  

This might be something like:
```
 apiVersion: logging.openshift.io/v1
 kind: Elasticsearch
 metadata:
   name: elasticsearch
   namespace: openshift-logging
   annotations:
      logging.openshift.io/elasticsearch-secret-name: fluentd
 ```

Maybe the requester also needs to request the roles he requires so we can properly seed the permissions: 

```
  logging.openshift.io/elasticsearch-roles: read,write
```
or as I mentioned in a previous comment we need to be explicit about the roles we give.  I presume we should not have to do this for kibana deployments because user's express they need kibana and  the certs are only needed between kibana and elasticsearch

File: enhancements/cluster-logging/elasticsearch-cert-management.md (hunk 11 above)
Comment (ewolinetz) [reply enhancements/cluster-logging/elasticsearch-cert-management.md:85]: > We need to either express a way for users to request an ES permission set/role or explicitly list what permission/set or role you get when get these certs

do we have a requirement for this anywhere? We don't currently offer this functionality

File: enhancements/cluster-logging/elasticsearch-cert-management.md (hunk 10 above)
Comment (ewolinetz) [reply enhancements/cluster-logging/elasticsearch-cert-management.md:130]: > Why does the requester care about the CN name? Why cant the provider generate a CN name using a known combination of NS, ES name, secret name (e.g. openshift-logging.elasticsearch.fluentd) or similiar?

this would make our implementation that much more difficult then, we would be updating and seeding the elasticsearch security config on the fly for each cluster because we've decided to allow a new configuration of DN on the client certs. For a new cluster it is already complicated but if we have to upgrade a cluster that is moving from CL/Jaeger managing certs to EO doing so and we also need to reseed their cluster... i think this is well outside the scope of what we are looking to achieve here

File: enhancements/cluster-logging/elasticsearch-cert-management.md (hunk 4 above)
Comment (ewolinetz) [reply enhancements/cluster-logging/elasticsearch-cert-management.md:116]: > As an aside, this task will be tricky because it presumably requires a cluster restart whenever a new client is added

how so? we have a signing cert that is used as the CA for the cluster, we don't need to restart at all because we'd trust that.

File: enhancements/cluster-logging/elasticsearch-cert-management.md (hunk 11 above)
Comment (jcantrill) [reply enhancements/cluster-logging/elasticsearch-cert-management.md:85]: I don't know that we do need someone to request a certain permission;  I only put it out there as a possibility.  I think if, with this proposal, we explicitly state the role you will be given when you request certs, I believe that is sufficient and we are covered.

File: enhancements/cluster-logging/elasticsearch-cert-management.md (hunk 4 above)
Comment (jcantrill) [reply enhancements/cluster-logging/elasticsearch-cert-management.md:116]: > > As an aside, this task will be tricky because it presumably requires a cluster restart whenever a new client is added
> 
> how so? we have a signing cert that is used as the CA for the cluster, we don't need to restart at all because we'd trust that.

Stretching my cert limits here... is mutual auth tangential to this question since that is what we have enabled?  Wouldn't the server need to present its public key to the caller meaning it has to be in the key store?  If so could we just disable mutual auth to avoid the issue? Is it a non issue?

File: enhancements/cluster-logging/elasticsearch-cert-management.md (hunk 10 above)
Comment (jcantrill) [reply enhancements/cluster-logging/elasticsearch-cert-management.md:130]: > > Why does the requester care about the CN name? Why cant the provider generate a CN name using a known combination of NS, ES name, secret name (e.g. openshift-logging.elasticsearch.fluentd) or similiar?
> 
> this would make our implementation that much more difficult then, we would be updating and seeding the elasticsearch security config on the fly for each cluster because we've decided to allow a new configuration of DN on the client certs. For a new cluster it is already complicated but if we have to upgrade a cluster that is moving from CL/Jaeger managing certs to EO doing so and we also need to reseed their cluster... i think this is outside the scope of what we are looking to achieve here

I guess we need to take this in context of how far work will go with the operator.  The realistic expectation is that only jaeger and CLO will make use of cert gen meaning we can continue to rely on the static security config that exist today.  The implications are there is the possibility that either of those consumers don't specify what matches the static config and thus are broken.  That risk is likely small.  In this context the following is probably sufficient:

```apiVersion: logging.openshift.io/v1
kind: Elasticsearch
metadata:
  name: elasticsearch
  annotations:
    logging.openshift.io/elasticsearch-client-secret/system.logging.fluentd: fluentd
```
I believe the existence of the one annotation is sufficient to indicate a consumer needs certs injected

enhancements/cluster-logging/elasticsearch-cert-management.md
jcantrill: Couldnt  we assume for a smoother transition:
* follow existing pattern of a named secret that matches named elasticsearch resource 
or
* elasticsearch resource with annotation takes precedence over previous pattern?

I think we want the requestor (CL/Jaeger) to own the secret regardless as it has to exist to be mounted into something like the collector pod, and assume the secret will be injected the explicit keys.  I defer to @alanconway here

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: this is a non-issue... we will trust the signing cert (which will be used in signing all of the client/server certs). Meaning, as long as someone trusts this original signing cert they would trust anyone that is using a cert we've signed.

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: @syedriko ack, i updated the document to describe the field names we'll use in the secrets for anyone that is opting in to have client certs generated for them. I believe it matches what is expected for the log forward use case.

Also, given I've updated the intention to make using this an opt-in configuration (by annotation on the elasticsearch CR) we don't have to worry about breaking anyone prior to them being ready to make all necessary changes that would go along with using certs managed by EO.

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: this enhancement outlines that to have EO create and manage certs for your cluster, you'll need to annotate your elasticsearch CR. If you are doing that, then you are opting in to this process and it would replace the current pattern (where CL and Jaeger manage certs/secrets).

> I think we want the requestor (CL/Jaeger) to own the secret regardless as it has to exist to be mounted into something like the collector pod, and assume the secret will be injected the explicit keys. I defer to @alanconway here

I think we can do this either way... I would rather lean towards doing what the cert signing service already does though to be consistent.

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: > I believe the existence of the one annotation is sufficient to indicate a consumer needs certs injected

Except for the case of Kibana and ES are stood up and managed by EO... we would need to know if EO is going to be responsible for creating the certs so that they can communicate with one another or if someone else is going to provide that for us. Hence the other annotation (I'm thinking about use cases where someone wants to just stand up ES and Kibana together... its not a typical use case, however it is specific for testing in the EO repo)

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: I think this would be considered a stretch goal/completely separate proposal... we are currently only looking to move cert management into the operator. This starts to get into the realm of EO dynamically seeding ES.

enhancements/cluster-logging/elasticsearch-cert-management.md
jcantrill: >  logging.openshift.io/elasticsearch-cert-management: true

I'm missing the value here?  If I define a singular annotation is that not sufficient to indicate I want the EO to provide and manage certs? If i define nothing isn't the expectation that I am managing the certs?

enhancements/cluster-logging/elasticsearch-cert-management.md
jcantrill: Some way we still need to be explicit about what role will be granted even if it is to state you will get the roles that existed prior to implementation.  We know what they are so we should be able to state them.  My main concern is we are bound to have someone ask how to get some alternate ES permission.  Defining what you get with this proposal provides us a pointer for when someone asks

ewolinetz: > My main concern is we are bound to have someone ask how to get some alternate ES permission. Defining what you get with this proposal provides us a pointer for when someone asks

We can adjust the proposal if/when that happens

enhancements/cluster-logging/elasticsearch-cert-management.md
ewolinetz: > If i define nothing isn't the expectation that I am managing the certs?

the intention of this is that eventually we always move it to where EO is creating/maintaining certs for its clusters. however, it will need to be an opt in, hence having the second annotation which will eventually be ignored/deprecated. We would still need the other annotation to denote that you need a client cert generated so you can communicate with ES.

jcantrill: Tying off this discussion from our verbal chat:
* one annotation says:  I want EO to manage certs for EO and the cluster itself to communicate
* one annotation says: I want EO to give me certs so a client can talk to the cluster

bparees:
@ewolinetz can you outline the workflow for a consumer like Jaeger today (what resources they create/consume/mount) and the workflow for Jaeger after this is implemented(assuming they switch to using it)?

there seemed to be a lot of back and forth between you and @jcantrill about the right annotation experience and i think walking through an example of how a consumer uses this would help shed light on whether it "looks right" or not.

ewolinetz:
/hold cancel
@bparees added a section to specifically call out the differences between current workflow and proposed

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,225 @@
...
+1. Generating and managing any client keys/certs to communicate with Elasticsearch. This would include their forwarders, and if applicable, their UIs.
+1. Generating and managing secrets to be used by: EO, Elasticsearch, IndexManagement cronjobs, and any clients as denoted in point 2.
+1. Creating and updating their elasticsearch CR(s) so that EO can manage their cluster.
+
+With this proposal the workflow would be:
+
+1. Create their elasticsearch CR with an annotation to denote EO will be managing certs for the cluster
+1. Add an additional annotation on the elasticsearch CR for each client they need to have certs injected into a specific secret.
+
+For example, CLO would use the following annotations to tell EO to do cert management and also inject client keys/certs into the `fluentd` secret with the CN `system.logging.fluentd`:
```
bparees, line 78: the secret is assumed to live in the same namespace as the ES CR?

ewolinetz: correct, I'll add that clarification

enhancements/cluster-logging/elasticsearch-cert-management.md
```diff
@@ -0,0 +1,225 @@
...
+
+1. Generating and managing their own keys/certificates including those that are used in Elasticsearch as server serving and inter-cluster so that Elasticsearch can join and form clusters.
+1. Generating and managing any client keys/certs to communicate with Elasticsearch. This would include their forwarders, and if applicable, their UIs.
+1. Generating and managing secrets to be used by: EO, Elasticsearch, IndexManagement cronjobs, and any clients as denoted in point 2.
+1. Creating and updating their elasticsearch CR(s) so that EO can manage their cluster.
+
+With this proposal the workflow would be:
+
+1. Create their elasticsearch CR with an annotation to denote EO will be managing certs for the cluster
+1. Add an additional annotation on the elasticsearch CR for each client they need to have certs injected into a specific secret.
```
bparees, line 76: do these injected secrets get rotated or expire?

ewolinetz: since EO would be managing them, they should be rotated. I'll clarify

jpkrohling (approved):

ewolinetz:
@bparees I hope I addressed your comments... please let me know if there's more I need to clarify on

bparees:
> @bparees I hope I addressed your comments... please let me know if there's more I need to clarify on

yup, thanks

I do wonder how you plan to handle expiration to ensure there's not outage during the rotation.  I assume you're going to handle it similar to how other components do, by having a window during which both certs are valid.

ewolinetz:
> I do wonder how you plan to handle expiration to ensure there's not outage during the rotation. I assume you're going to handle it similar to how other components do, by having a window during which both certs are valid.

Currently CLO and JO (they use the same base underlying script) wait until the cert is actually expired to regenerate it, I think it would be a better experience to regenerate it if it is going to expire within 24 hours. This way we get that overlap period for their component to respond to their secret updating.

PR #577, opened 2021-01-08
describe one-stop-shopping for exposing customized routes

deads2k:
## Summary

Add `CustomizeableRoute{Spec,Status}` to `IngressSpec` in a way that allows multiple operators to provide information about Routes (or Ingresses, or maybe even something else), that a cluster-admin wants to provide custom names and/or custom serving certificates for.

## Motivation

Some customers do not allow wildcard serving certificates for the openshift ingress to use, so they have to provide individual
serving certificates to each component. Those customers need a way to see all the routes they need to customize and a way to configure them. A stock installation has the following routes which may need customization of both names and serving cert/keys
1. openshift-authentication   oauth-openshift
2. openshift-console          console
3. openshift-console          downloads
4. openshift-monitoring       alertmanager-main
5. openshift-monitoring       grafana
6. openshift-monitoring       prometheus-k8s
7. openshift-monitoring       thanos-querier

It is just as easy to create a generic solution that provides one stop shopping for cluster-admins as it is to produce a series of one-off solutions. This proposes a single point of configuration because we like cluster-admins and we dislike writing multiple pages of slightly different documentation.

@sttts @Miciah @ironcladlou @stlaz @mfojtik

### DIFF
--- a/enhancements/ingress/custom-route-configuration.md
+++ b/enhancements/ingress/custom-route-configuration.md
@@ -7,6 +7,8 @@ reviewers:
   - "@miciah"
   - "@standa"
   - "@spadgett"
+  - "@sur"
+  - "@lilic"
 approvers:
   - "@sttts"
   - "@miciah"
@@ -48,6 +50,8 @@ A stock installation has the following routes which may need customization of bo
 5. openshift-monitoring       grafana
 6. openshift-monitoring       prometheus-k8s
 7. openshift-monitoring       thanos-querier
+8. image-registry (this isn't appearing in my installation, but @Miciah says it's there)
+9. openshfit-logging (this would be natural, though I think it comes via OLM, so we may have to definitively answer that)
 
 It is just as easy to create a generic solution that provides one stop shopping for cluster-admins as it is to produce
 a series of one-off solutions.
@@ -56,7 +60,7 @@ slightly different documentation.
 
 ### Goals
 
-1. provide a way for cluster-admins to see all the customizeable routes in the cluster.
+1. provide a way for cluster-admins to see all the routes that have flexible names and serving certificates in the cluster.
 2. provide a single API that is consistent for every route they need to configure
 3. provides a way to have an operator with scope limited permissions read the serving cert/key pairs
 4. allow a cluster-admin to specify a different DNS name.
@@ -72,73 +76,88 @@ slightly different documentation.
 type IngressSpec struct {
     // other fields
 
-    // CustomizeableRoutes is a list of routes that a cluster-admin wants to customize.  It is logically keyed by
-    // .spec.customizeableRoutes[index].{namespace,name}.
-    // To determine the set of possible keys, look at .status.customizeableRoutes where participating operators place
+    // ComponentRoutes is a list of routes that a cluster-admin wants to customize.  It is logically keyed by
+    // .spec.componentRoutes[index].{namespace,name}.
+    // To determine the set of possible keys, look at .status.componentRoutes where participating operators place
     // current route status keyed the same way.
-    CustomizeableRoutes []CustomizeableRouteSpec
+    // If a ComponentRoute is created with a namespace,name tuple that does not match status, that piece of config will
+    // not have an effect.  If an operator later reads the field, it will eventually (but not necessarily immediately)
+    // honor the pre-existing spec values.
+    ComponentRoutes []ComponentRouteSpec
 }
 
 type IngressStatus struct {
     // other fields
 
-    // CustomizeableRoutes is where participating operators place the current route status for routes which the cluster-admin
+    // ComponentRoutes is where participating operators place the current route status for routes which the cluster-admin
     // can customize hostnames and serving certificates.
     // How the operator uses that serving certificate is up to the individual operator.
-    CustomizeableRoutes []CustomizeableRouteStatus
+    // An operator that creates entries in this slice should clean them up during removal (if it can be removed).
+    // An operator must also handle the case of deleted status without churn.
+    ComponentRoutes []ComponentRouteStatus
 }
 
-type CustomizeableRouteSpec struct{
-    // namespace is the namespace of the route to customize.  It must be a real namespace.
+type ComponentRouteSpec struct{
+    // namespace is the namespace of the route to customize.  It must be a real namespace.  Using an actual namespace
+    // ensures that no two components will conflict and the same component can be installed multiple times.
     Namespace string
     // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
     // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
     // this name.
     Name string
-    // desiredHost is the name that a cluster-admin wants to specify
-    DesiredHost string
+    // Hostname is the host name that a cluster-admin wants to specify
+    Hostname string
     // ServingCertKeyPairSecret is a reference to a secret in namespace/openshift-config that is a kubernetes tls secret.
     // The serving cert/key pair must match and will be used by the operator to fulfill the intent of serving with this name.
     // That means it could be embedded into the route or used by the operand directly.
     // Operands should take care to ensure that if they use passthrough termination, they properly use SNI to allow service
     // DNS access to continue to function correctly.
+    // SANs in the certificate are ignored, but SNI can be used to make operator managed certificates (like internal load balancers
+    // and service serving certificates) serve correctly.
     ServingCertKeyPairSecret SecretNameReference
 
     // possible future, we could add a set of SNI mappings.  I suspect most operators would not properly handle it today.
 }
 
-type CustomizeableRouteStatus struct{
-    // namespace is the namespace of the route to customize.  It must be a real namespace.
+type ComponentRouteStatus struct{
+    // namespace is the namespace of the route to customize.  It must be a real namespace.  Using an actual namespace
+    // ensures that no two components will conflict and the same component can be installed multiple times.
     Namespace string
     // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
     // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
     // this name.
     Name string
-    // defaultHost is the normal name of this route.
-    DefaultHost string
+    // defaultHostname is the normal host name of this route.  It is provided in case cluster-admins find it more recognizeable
+    // and having it here makes it possible to answer, "if I remove my configuration, what will the name be".
+    DefaultHostname string
     // ConsumingUsers is a slice of users that need to have read permission on the secrets in order to use them.
     // This will usually be an operator service account.
     ConsumingUsers []string
-    // currentHost is the current name used to by the route.  Routes can have more than one exposed name, even though we
+    // currentHostnames is the current name used to by the route.  Routes can have more than one exposed name, even though we
     // only allow one route.spec.host.
-    CurrentHosts []string
+    CurrentHostnames []string
 
-    // conditions are available, degraded, progressing.  This allows consistent reporting back and feedback that is well
+    // conditions are degraded and progressing.  This allows consistent reporting back and feedback that is well
     // structured.  These particular conditions have worked very well in ClusterOperators.
+    // Degraded == true means that something has gone wrong trying to handle the ComponentRoute.  The CurrentHostnames
+    // may or may not be operating successfully.
+    // Progressing == true means that the component is taking some action related to the ComponentRoute
     Conditions []ConfigCondition
 
     // relatedObjects allows listing resources which are useful when debugging or inspecting how this is applied.
-    // They may be aggregated into an overal status RelatedObjects to be automatically by oc adm inspect
+    // They may be aggregated into an overall status RelatedObjects to be automatically shown by oc adm inspect
     RelatedObjects []ObjectReference
     
     
 
     // This API does not include a mechanism to distribute trust, since the ability to write this resource would then 
     // allow interception.  Instead, if we need such a mechanism, we can talk about creating a way to allow narrowly scoped
-    // updates to a configmap containing ca-bundle.crt for each CustomizeableRoute.
+    // updates to a configmap containing ca-bundle.crt for each ComponentRoute.
     // CurrentCABundle []byte
 }
 ```
+
+Validation rules to be specified in the openshift/api PR, but basic restrictions on strings as you'd expect.
 
 
 ### User Stories
@@ -147,7 +166,7 @@ type CustomizeableRouteStatus struct{
 To use this, a cluster-admin would
 1.  Either read docs to find the namespace,name tuples or look at an existing cluster and read ingresses.config.openshift.io.
 2.  Create the serving cert/key pair secret in ns/openshift-config
-3.  Create an entry in `ingresses.config.openshift.io.spec.customizeableRouteSpec` for their customization
+3.  Create an entry in `ingresses.config.openshift.io.spec.componentRouteSpec` for their customization
 4.  Wait to see the corresponding change in status.
 
 
@@ -157,8 +176,8 @@ To use this, a cluster-admin would
 
 #### Control loop to manage secret read permissions in openshift-config
 List/watch for individual names has been possible for several releases.
-A control loop in either the ingress operator or the cluster config operator will watch for ingress.spec.customizeableRouteSpec.servingCertKeyPairSecret
-and will create a role/rolebinding for the corresponding ingress.status.customizeableRouteStatus.consumingUsers.
+A control loop in either the ingress operator or the cluster config operator will watch for ingress.spec.componentRouteSpec.servingCertKeyPairSecret
+and will create a role/rolebinding for the corresponding ingress.status.componentRouteStatus.consumingUsers.
 This will allow an operator (or other binary with that user) to get/list/watch on the particular secret.
 
 This means that the power to mutate ingresses.config.openshift.io and ingresses.config.openshfit.io/status
@@ -173,7 +192,8 @@ copying the secret so it can be mounted.
 It would also be possible (though I'm less sure someone would use it), to provide a route injection option for serving cert/key
 pairs.
 
-Library-go is also an ideal spot to provide functionality to provide request route hosts.
+Library-go is also an ideal spot to provide functionality to inject route hosts.  Something like our StaticResource
+controller which automatically handles the API status and the required route changes.
 
 ### Risks and Mitigations
 
--- a/enhancements/ingress/custom-route-configuration.md
+++ b/enhancements/ingress/custom-route-configuration.md
@@ -24,8 +24,8 @@ superseded-by:
 
 ## Release Signoff Checklist
 
-- [ ] Enhancement is `implementable`
-- [ ] Design details are appropriately documented from clear requirements
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
 - [ ] Test plan is defined
 - [ ] Operational readiness criteria is defined
 - [ ] Graduation criteria for dev preview, tech preview, GA
@@ -51,7 +51,6 @@ A stock installation has the following routes which may need customization of bo
 6. openshift-monitoring       prometheus-k8s
 7. openshift-monitoring       thanos-querier
 8. image-registry (this isn't appearing in my installation, but @Miciah says it's there)
-9. openshfit-logging (this would be natural, though I think it comes via OLM, so we may have to definitively answer that)
 
 It is just as easy to create a generic solution that provides one stop shopping for cluster-admins as it is to produce
 a series of one-off solutions.
@@ -72,7 +71,7 @@ slightly different documentation.
 
 ## Proposal
 
-```
+```go
 type IngressSpec struct {
     // other fields
 
@@ -112,8 +111,7 @@ type ComponentRouteSpec struct{
     // That means it could be embedded into the route or used by the operand directly.
     // Operands should take care to ensure that if they use passthrough termination, they properly use SNI to allow service
     // DNS access to continue to function correctly.
-    // SANs in the certificate are ignored, but SNI can be used to make operator managed certificates (like internal load balancers
-    // and service serving certificates) serve correctly.
+    // Operators are not required to inspect SANs in the certificate to set up SNI.
     ServingCertKeyPairSecret SecretNameReference
 
     // possible future, we could add a set of SNI mappings.  I suspect most operators would not properly handle it today.
@@ -164,10 +162,10 @@ Validation rules to be specified in the openshift/api PR, but basic restrictions
 
 #### Cluster-admin wants to customize a route
 To use this, a cluster-admin would
-1.  Either read docs to find the namespace,name tuples or look at an existing cluster and read ingresses.config.openshift.io.
-2.  Create the serving cert/key pair secret in ns/openshift-config
-3.  Create an entry in `ingresses.config.openshift.io.spec.componentRouteSpec` for their customization
-4.  Wait to see the corresponding change in status.
+1. Either read docs to find the namespace, name tuples or look at an existing cluster and read ingresses.config.openshift.io.
+2. Create the serving cert/key pair secret in ns/openshift-config
+3. Create an entry in `ingresses.config.openshift.io.spec.componentRouteSpec` for their customization
+4. Wait to see the corresponding change in status.
 
 
 #### Story 2
@@ -216,7 +214,7 @@ The helpers suggested above can ease that transition, but it still exists.
 1. an operator test should be written for this
 2. QE testing on configuration changes.
 
-##### Removing a deprecated feature
+#### Removing a deprecated feature
 
 - Announce deprecation and support policy of the existing feature
 - Deprecate the feature
@@ -226,7 +224,6 @@ The helpers suggested above can ease that transition, but it still exists.
 For components that did not try to solve this in a one-off way, the upgrade is easy, the new feature becomes available.
 
 For components that created a one-off solution, the upgrade will vary depending on the steps they used.
-I seem to recall that monitoring allowed direct route manipulation and the console created its own fields.
 
 On downgrade, the customizations will be removed by the old versions of the operators.
 While this is annoying, it is consistent with our general downgrade story of new features require the new product.
@@ -247,7 +244,7 @@ The idea is to find the best form of an argument why this enhancement should _no
 ## Alternatives
 
 ### cluster-admin created roles and rolebindings
-This avoids the transitive privilege escalation to read secrets based on update rights to ingresses.config.openshift.io 
+This avoids the transitive privilege escalation to read secrets based on update rights to ingresses.config.openshift.io
 and ingresses.config.openshift.io/status.
 However, those permissions are closely held.
 
--- a/enhancements/ingress/custom-route-configuration.md
+++ b/enhancements/ingress/custom-route-configuration.md
@@ -1,0 +1,242 @@
+---
+title: custom-route-configuration
+authors:
+  - "@deads2k"
+reviewers:
+  - "@danmace"
+  - "@miciah"
+  - "@standa"
+  - "@spadgett"
+approvers:
+  - "@sttts"
+  - "@miciah"
+creation-date: 2021-01-08
+last-updated: 2021-01-08
+status: implementable
+see-also:
+replaces:
+superseded-by:
+---
+
+# Custom Route Configuration
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Add `CustomizeableRoute{Spec,Status}` to `IngressSpec` in a way that allows multiple operators to provide information about
+Routes (or Ingresses, or maybe even something else), that a cluster-admin wants to provide custom names and/or custom
+serving certificates for.
+
+## Motivation
+
+Some customers do not allow wildcard serving certificates for the openshift ingress to use, so they have to provide individual
+serving certificates to each component.
+Those customers need a way to see all the routes they need to customize and a way to configure them.
+A stock installation has the following routes which may need customization of both names and serving cert/keys
+1. openshift-authentication   oauth-openshift
+2. openshift-console          console
+3. openshift-console          downloads
+4. openshift-monitoring       alertmanager-main
+5. openshift-monitoring       grafana
+6. openshift-monitoring       prometheus-k8s
+7. openshift-monitoring       thanos-querier
+
+It is just as easy to create a generic solution that provides one stop shopping for cluster-admins as it is to produce
+a series of one-off solutions.
+This proposes a single point of configuration because we like cluster-admins and we dislike writing multiple pages of
+slightly different documentation.
+
+### Goals
+
+1. provide a way for cluster-admins to see all the customizeable routes in the cluster.
+2. provide a single API that is consistent for every route they need to configure
+3. provides a way to have an operator with scope limited permissions read the serving cert/key pairs
+4. allow a cluster-admin to specify a different DNS name.
+
+### Non-Goals
+
+1. provide a way to specify SNI serving certificates for operands.
+2. dictate how an operator must expose itself, the termination policy on a route, or how the secret is used.
+
+## Proposal
+
+```
+type IngressSpec struct {
+    // other fields
+
+    // CustomizeableRoutes is a list of routes that a cluster-admin wants to customize.  It is logically keyed by
+    // .spec.customizeableRoutes[index].{namespace,name}.
+    // To determine the set of possible keys, look at .status.customizeableRoutes where participating operators place
+    // current route status keyed the same way.
+    CustomizeableRoutes []CustomizeableRouteSpec
+}
+
+type IngressStatus struct {
+    // other fields
+
+    // CustomizeableRoutes is where participating operators place the current route status for routes which the cluster-admin
+    // can customize hostnames and serving certificates.
+    // How the operator uses that serving certificate is up to the individual operator.
+    CustomizeableRoutes []CustomizeableRouteStatus
+}
+
+type CustomizeableRouteSpec struct{
+    // namespace is the namespace of the route to customize.  It must be a real namespace.
+    Namespace string
+    // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
+    // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
+    // this name.
+    Name string
+    // desiredHost is the name that a cluster-admin wants to specify
+    DesiredHost string
+    // ServingCertKeyPairSecret is a reference to a secret in namespace/openshift-config that is a kubernetes tls secret.
+    // The serving cert/key pair must match and will be used by the operator to fulfill the intent of serving with this name.
+    // That means it could be embedded into the route or used by the operand directly.
+    // Operands should take care to ensure that if they use passthrough termination, they properly use SNI to allow service
+    // DNS access to continue to function correctly.
+    ServingCertKeyPairSecret SecretNameReference
+
+    // possible future, we could add a set of SNI mappings.  I suspect most operators would not properly handle it today.
+}
+
+type CustomizeableRouteStatus struct{
+    // namespace is the namespace of the route to customize.  It must be a real namespace.
+    Namespace string
+    // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
+    // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
+    // this name.
+    Name string
+    // defaultHost is the normal name of this route.
+    DefaultHost string
+    // ConsumingUsers is a slice of users that need to have read permission on the secrets in order to use them.
+    // This will usually be an operator service account.
+    ConsumingUsers []string
+    // currentHost is the current name used to by the route.  Routes can have more than one exposed name, even though we
+    // only allow one route.spec.host.
+    CurrentHosts []string
+
+    // conditions are available, degraded, progressing.  This allows consistent reporting back and feedback that is well
+    // structured.  These particular conditions have worked very well in ClusterOperators.
+    Conditions []ConfigCondition
+
+    // relatedObjects allows listing resources which are useful when debugging or inspecting how this is applied.
+    // They may be aggregated into an overal status RelatedObjects to be automatically by oc adm inspect
+    RelatedObjects []ObjectReference
+    
+    
+
+    // This API does not include a mechanism to distribute trust, since the ability to write this resource would then 
+    // allow interception.  Instead, if we need such a mechanism, we can talk about creating a way to allow narrowly scoped
+    // updates to a configmap containing ca-bundle.crt for each CustomizeableRoute.
+    // CurrentCABundle []byte
+}
+```
+
+
+### User Stories
+
+#### Cluster-admin wants to customize a route
+To use this, a cluster-admin would
+1.  Either read docs to find the namespace,name tuples or look at an existing cluster and read ingresses.config.openshift.io.
+2.  Create the serving cert/key pair secret in ns/openshift-config
+3.  Create an entry in `ingresses.config.openshift.io.spec.customizeableRouteSpec` for their customization
+4.  Wait to see the corresponding change in status.
+
+
+#### Story 2
+
+### Implementation Details/Notes/Constraints [optional]
+
+#### Control loop to manage secret read permissions in openshift-config
+List/watch for individual names has been possible for several releases.
+A control loop in either the ingress operator or the cluster config operator will watch for ingress.spec.customizeableRouteSpec.servingCertKeyPairSecret
+and will create a role/rolebinding for the corresponding ingress.status.customizeableRouteStatus.consumingUsers.
+This will allow an operator (or other binary with that user) to get/list/watch on the particular secret.
+
+This means that the power to mutate ingresses.config.openshift.io and ingresses.config.openshfit.io/status
+will imply the power to read secrets in openshift-config.
+If we decide we do not wish to do this, then it will incumbent upon the cluster-admin to create the role and rolebinding.
+
+#### library-go usage
+Because this will be leveraged by the authentication operator (and possibly the console operator), we can provide a config
+observer that reads the ingresses.spec, handles the specified names, sets up the specific list/watch, and takes care of
+copying the secret so it can be mounted.
+
+It would also be possible (though I'm less sure someone would use it), to provide a route injection option for serving cert/key
+pairs.
+
+Library-go is also an ideal spot to provide functionality to provide request route hosts.
+
+### Risks and Mitigations
+
+Some components have attempted to resolve similar issues in one off ways.
+This has produced a disjointed experience across the stack and challenging documentation to follow.
+Since these other ways exist, a migration to a unified mechanism may be challenging for those components.
+The helpers suggested above can ease that transition, but it still exists.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. Which components should migrate?
+2. What timeline should that migration happen?
+3. Which components can start from this unified API?
+4. Do we allow OLM operators to participate?
+
+### Test Plan
+
+1. an operator test should be written for this
+2. QE testing on configuration changes.
+
+##### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
+
+### Upgrade / Downgrade Strategy
+
+For components that did not try to solve this in a one-off way, the upgrade is easy, the new feature becomes available.
+
+For components that created a one-off solution, the upgrade will vary depending on the steps they used.
+I seem to recall that monitoring allowed direct route manipulation and the console created its own fields.
+
+On downgrade, the customizations will be removed by the old versions of the operators.
+While this is annoying, it is consistent with our general downgrade story of new features require the new product.
+
+### Version Skew Strategy
+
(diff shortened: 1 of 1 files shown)


enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+
+type CustomizeableRouteSpec struct{
+    // namespace is the namespace of the route to customize.  It must be a real namespace.
+    Namespace string
+    // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
+    // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
+    // this name.
+    Name string
+    // desiredHost is the name that a cluster-admin wants to specify
+    DesiredHost string
```
Miciah, line 99: This is in `spec`, so isn't "desired" redundant?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+    // Operands should take care to ensure that if they use passthrough termination, they properly use SNI to allow service
+    // DNS access to continue to function correctly.
+    ServingCertKeyPairSecret SecretNameReference
+
+    // possible future, we could add a set of SNI mappings.  I suspect most operators would not properly handle it today.
+}
+
+type CustomizeableRouteStatus struct{
+    // namespace is the namespace of the route to customize.  It must be a real namespace.
+    Namespace string
```
Miciah, line 112: `Namespace` in status is slightly less redundant than `Namespace` in spec, but it has similar risks of causing confusion.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+    // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
+    // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
+    // this name.
+    Name string
+    // defaultHost is the normal name of this route.
+    DefaultHost string
```
Miciah, line 118: Is this needed?  We don't usually have status fields to say what value we _would_ use if spec didn't specify a value.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+    // ConsumingUsers is a slice of users that need to have read permission on the secrets in order to use them.
+    // This will usually be an operator service account.
+    ConsumingUsers []string
+    // currentHost is the current name used to by the route.  Routes can have more than one exposed name, even though we
+    // only allow one route.spec.host.
+    CurrentHosts []string
+
+    // conditions are available, degraded, progressing.  This allows consistent reporting back and feedback that is well
+    // structured.  These particular conditions have worked very well in ClusterOperators.
+    Conditions []ConfigCondition
```
Miciah, line 128: Makes sense to re-use status condition types, but we need to define their semantics in this context.  Does the owning operator (and only that operator) manage status conditions (e.g., console-operator exclusively manages status of the "console" and "console-downloads" entries)?  Is the same operator expected to perform health probes and set `Degraded=True` if the route is unreachable or has misconfigured TLS?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+
+#### library-go usage
+Because this will be leveraged by the authentication operator (and possibly the console operator), we can provide a config
+observer that reads the ingresses.spec, handles the specified names, sets up the specific list/watch, and takes care of
+copying the secret so it can be mounted.
+
+It would also be possible (though I'm less sure someone would use it), to provide a route injection option for serving cert/key
+pairs.
+
+Library-go is also an ideal spot to provide functionality to provide request route hosts.
```
Miciah, line 176: What does it mean "to provide request route hosts"?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+## Proposal
+
+```
+type IngressSpec struct {
+    // other fields
+
+    // CustomizeableRoutes is a list of routes that a cluster-admin wants to customize.  It is logically keyed by
+    // .spec.customizeableRoutes[index].{namespace,name}.
+    // To determine the set of possible keys, look at .status.customizeableRoutes where participating operators place
+    // current route status keyed the same way.
```
Miciah, line 78: Is an entry in spec that does not have a corresponding entry in status dead config?  Will an admission controller inhibit such entries in spec?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+    CustomizeableRoutes []CustomizeableRouteSpec
+}
+
+type IngressStatus struct {
+    // other fields
+
+    // CustomizeableRoutes is where participating operators place the current route status for routes which the cluster-admin
+    // can customize hostnames and serving certificates.
+    // How the operator uses that serving certificate is up to the individual operator.
```
Miciah, line 87: Is an operator responsible for deleting any entries it has added here when it is removed?

Miciah (commented):
Looks reasonable.  Is allowing configuration using installer manifests a goal or non-goal?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+This means that the power to mutate ingresses.config.openshift.io and ingresses.config.openshfit.io/status
+will imply the power to read secrets in openshift-config.
+If we decide we do not wish to do this, then it will incumbent upon the cluster-admin to create the role and rolebinding.
+
+#### library-go usage
+Because this will be leveraged by the authentication operator (and possibly the console operator), we can provide a config
+observer that reads the ingresses.spec, handles the specified names, sets up the specific list/watch, and takes care of
+copying the secret so it can be mounted.
+
+It would also be possible (though I'm less sure someone would use it), to provide a route injection option for serving cert/key
```
stlaz, line 173: Isn't that the only option if your routes use reencrypt/edge termination?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+1. provide a way to specify SNI serving certificates for operands.
+2. dictate how an operator must expose itself, the termination policy on a route, or how the secret is used.
+
+## Proposal
+
+```
+type IngressSpec struct {
+    // other fields
+
+    // CustomizeableRoutes is a list of routes that a cluster-admin wants to customize.  It is logically keyed by
```
stlaz, line 75: language corner: doesn't English usually omit the "e" when the "-able" suffix gets used?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+
+    // CustomizeableRoutes is where participating operators place the current route status for routes which the cluster-admin
+    // can customize hostnames and serving certificates.
+    // How the operator uses that serving certificate is up to the individual operator.
+    CustomizeableRoutes []CustomizeableRouteStatus
+}
+
+type CustomizeableRouteSpec struct{
+    // namespace is the namespace of the route to customize.  It must be a real namespace.
+    Namespace string
```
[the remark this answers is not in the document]
stlaz: You want to modify a route in a given namespace, no? Having namespace therefore seems natural to me. My 2c

enhancements/ingress/custom-route-configuration.md
stlaz: +1, it would be nice to have a section describing what available/degraded/progressing mean in this context

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+    // currentHost is the current name used to by the route.  Routes can have more than one exposed name, even though we
+    // only allow one route.spec.host.
+    CurrentHosts []string
+
+    // conditions are available, degraded, progressing.  This allows consistent reporting back and feedback that is well
+    // structured.  These particular conditions have worked very well in ClusterOperators.
+    Conditions []ConfigCondition
+
+    // relatedObjects allows listing resources which are useful when debugging or inspecting how this is applied.
+    // They may be aggregated into an overal status RelatedObjects to be automatically by oc adm inspect
```
stlaz, line 131: nit: to be automatically shown (?) by oc adm inspect

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+and will create a role/rolebinding for the corresponding ingress.status.customizeableRouteStatus.consumingUsers.
+This will allow an operator (or other binary with that user) to get/list/watch on the particular secret.
+
+This means that the power to mutate ingresses.config.openshift.io and ingresses.config.openshfit.io/status
+will imply the power to read secrets in openshift-config.
+If we decide we do not wish to do this, then it will incumbent upon the cluster-admin to create the role and rolebinding.
+
+#### library-go usage
+Because this will be leveraged by the authentication operator (and possibly the console operator), we can provide a config
+observer that reads the ingresses.spec, handles the specified names, sets up the specific list/watch, and takes care of
```
stlaz, line 170: What about validation of the specific spec fields, would library-go serve as the model implementation even for operators that don't use it? Consider adding a section on validation of the specific fields.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+The helpers suggested above can ease that transition, but it still exists.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. Which components should migrate?
+2. What timeline should that migration happen?
+3. Which components can start from this unified API?
+4. Do we allow OLM operators to participate?
```
stlaz, line 192: I have doubts about having both auto-RBAC assign and allowing OLM in.

enhancements/ingress/custom-route-configuration.md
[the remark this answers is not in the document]
Miciah: Is it possible to have two routes with the same logical CustomizeableRoute name but in different namespaces?

enhancements/ingress/custom-route-configuration.md
Miciah: I assume that for passthrough routes, the operator (which has a serviceaccount in `ConsumingUsers`) reads the certificate and configures the operand to use it as the serving certificate.

enhancements/ingress/custom-route-configuration.md
Miciah: +1, if you want consumers to do any validation, make sure it is documented, not just codified in library-go (which many core operators do not use).

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+### Goals
+
+1. provide a way for cluster-admins to see all the customizeable routes in the cluster.
+2. provide a single API that is consistent for every route they need to configure
+3. provides a way to have an operator with scope limited permissions read the serving cert/key pairs
+4. allow a cluster-admin to specify a different DNS name.
+
+### Non-Goals
+
+1. provide a way to specify SNI serving certificates for operands.
```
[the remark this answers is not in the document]
deads2k: > What is the distinction between specifying SNI serving certificates for operands and specifying `ServingCertKeyPairSecret`?

Yes.  The cert/key pair is a single serving certificate, but an SNI configuration requires mapping ServerName to cert/key pairs.

enhancements/ingress/custom-route-configuration.md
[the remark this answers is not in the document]
deads2k: > Is it possible to have two routes with the same logical CustomizeableRoute name but in different namespaces?

`api` or `apiserver` seemed like natural ones that would conflict if you don't use namespaces, but when namespace qualified do not conflict.

Namespaces are a way to subdivide our operators and operands today to ensure no conflicts.

enhancements/ingress/custom-route-configuration.md
deads2k: > Is this needed? We don't usually have status fields to say what value we _would_ use if spec didn't specify a value

Recognizeability.  One mapping is namespace/name.  Another likely mapping is "this is the name of the thing I want to customize".  Displaying both in status allows a user to find names based on either.

enhancements/ingress/custom-route-configuration.md
deads2k: > Is an entry in spec that does not have a corresponding entry in status dead config? Will an admission controller inhibit such entries in spec?

I was thinking dead, not prohibited.  If you prohibit, then you have an ordering problem. I will update the enhancement.

enhancements/ingress/custom-route-configuration.md
deads2k: > Is an operator responsible for deleting any entries it has added here when it is removed?

Yes, will clarify.

enhancements/ingress/custom-route-configuration.md
deads2k: > language corner: doesn't English usually omit the "e" when the "-able" suffix gets used?

heh, now I'm slightly embarassed. ;)

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
+---
+title: custom-route-configuration
+authors:
+  - "@deads2k"
+reviewers:
+  - "@danmace"
+  - "@miciah"
+  - "@standa"
+  - "@spadgett"
```
sttts, line 9: @sur @lilic as well

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+Those customers need a way to see all the routes they need to customize and a way to configure them.
+A stock installation has the following routes which may need customization of both names and serving cert/keys
+1. openshift-authentication   oauth-openshift
+2. openshift-console          console
+3. openshift-console          downloads
+4. openshift-monitoring       alertmanager-main
+5. openshift-monitoring       grafana
+6. openshift-monitoring       prometheus-k8s
+7. openshift-monitoring       thanos-querier
+
```
ewolinetz, line 51: do we need to also include the openshift-logging kibana route? we currently don't allow any configuration (both with route name and cert)

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+    CustomizeableRoutes []CustomizeableRouteSpec
+}
+
+type IngressStatus struct {
+    // other fields
+
+    // CustomizeableRoutes is where participating operators place the current route status for routes which the cluster-admin
+    // can customize hostnames and serving certificates.
+    // How the operator uses that serving certificate is up to the individual operator.
+    CustomizeableRoutes []CustomizeableRouteStatus
```
simonpasquier, line 88: I'm missing details on how an operator would update the route's status. Would it be via a strategic merge patch?

enhancements/ingress/custom-route-configuration.md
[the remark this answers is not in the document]
sttts: Do you mean that the ServerName is taken from the cert, but there is no custom configuration beyond that?

enhancements/ingress/custom-route-configuration.md
sttts: Or just via normal update. Read, change, Update.

enhancements/ingress/custom-route-configuration.md
sttts: what is a name? Be more precise. A name is two lines above already.

enhancements/ingress/custom-route-configuration.md
sttts: > Is this needed?

When console is accessed via `console.foo.com`, how does it know which auth and which api name it should use? I understand it would pick `DefaultHost` here and forward the user to the right auth URL or to use the right api URL in the kubeconfig the user can download.

@deads2k is this your understanding?

sttts: This also means that if you have multiple domains, e.g. `console.public-internet.com` and `console.intranet`, the console would have no way to know the matching auth and api URL. @deads2k is that correct?

@Anandnatraj do we have the requirement for such a "polymorphic" mapping of URLs, depending on which console URL the user accesses?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+    // this name.
+    Name string
+    // desiredHost is the name that a cluster-admin wants to specify
+    DesiredHost string
+    // ServingCertKeyPairSecret is a reference to a secret in namespace/openshift-config that is a kubernetes tls secret.
+    // The serving cert/key pair must match and will be used by the operator to fulfill the intent of serving with this name.
+    // That means it could be embedded into the route or used by the operand directly.
+    // Operands should take care to ensure that if they use passthrough termination, they properly use SNI to allow service
+    // DNS access to continue to function correctly.
+    ServingCertKeyPairSecret SecretNameReference
```
sttts, line 105: @deads2k will this replace the existing cert? I.e. if the installer creates `console.asdfhzwdsf4322.openshift.com` as external cluster URL, and on the second day the admin sets `console.my-pretty-cluster.com` by providing a matching cert, the old name is gone?

And if this cert is broken, he would have lost access to the cluster?

And if the admin wants both the pretty and the installer created URL to access the console, he would have to provide a multi-host certificate (via SANs)?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+a series of one-off solutions.
+This proposes a single point of configuration because we like cluster-admins and we dislike writing multiple pages of
+slightly different documentation.
+
+### Goals
+
+1. provide a way for cluster-admins to see all the customizeable routes in the cluster.
+2. provide a single API that is consistent for every route they need to configure
+3. provides a way to have an operator with scope limited permissions read the serving cert/key pairs
+4. allow a cluster-admin to specify a different DNS name.
```
sttts, line 62: 5. allow a cluster-admin to specify a different DNS name, but keep the installer provided one
6. provide a fall-back in some way for the admin to access the cluster even if the cert is broken

@Anandnatraj what about 5 and 6?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+7. openshift-monitoring       thanos-querier
+
+It is just as easy to create a generic solution that provides one stop shopping for cluster-admins as it is to produce
+a series of one-off solutions.
+This proposes a single point of configuration because we like cluster-admins and we dislike writing multiple pages of
+slightly different documentation.
+
+### Goals
+
+1. provide a way for cluster-admins to see all the customizeable routes in the cluster.
```
sttts, line 59: what does customize mean? Be precise. Is it only about giving a custom tls bundle? Or is it more? You wording sounds like the operator stops touching its routes objects and the user can do whatever he likes.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+serving certificates to each component.
+Those customers need a way to see all the routes they need to customize and a way to configure them.
+A stock installation has the following routes which may need customization of both names and serving cert/keys
+1. openshift-authentication   oauth-openshift
+2. openshift-console          console
+3. openshift-console          downloads
+4. openshift-monitoring       alertmanager-main
+5. openshift-monitoring       grafana
+6. openshift-monitoring       prometheus-k8s
+7. openshift-monitoring       thanos-querier
```
sttts, line 50: where is the api in this list?

enhancements/ingress/custom-route-configuration.md
simonpasquier: Ack. I wasn't sure initially about conflicting updates but IIUC it's implied that each operator deals with potential conflict resolution. Correct?

sttts: yes, that's the expectation for kube clients.

pweil-:
@jhadvig FYI for console/downloads.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+    Namespace string
+    // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
+    // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
+    // this name.
+    Name string
+    // defaultHost is the normal name of this route.
+    DefaultHost string
+    // ConsumingUsers is a slice of users that need to have read permission on the secrets in order to use them.
+    // This will usually be an operator service account.
+    ConsumingUsers []string
```
mfojtik, line 121: Why is this in status? Can this be configured? How?

enhancements/ingress/custom-route-configuration.md
jhadvig: We already have a similar logic in the `console-operator` where, in case of a custom route a hostname and a secret reference has to be specified.
Check: https://github.com/openshift/api/blob/master/operator/v1/types_console.go#L51-L63

Would drop the `desired` and just use hostname as we use in https://github.com/openshift/api/blob/master/operator/v1/types_console.go#L53

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+
+It is just as easy to create a generic solution that provides one stop shopping for cluster-admins as it is to produce
+a series of one-off solutions.
+This proposes a single point of configuration because we like cluster-admins and we dislike writing multiple pages of
+slightly different documentation.
+
+### Goals
+
+1. provide a way for cluster-admins to see all the customizeable routes in the cluster.
+2. provide a single API that is consistent for every route they need to configure
```
jhadvig, line 60: Does this mean that after this is implemented, we can get rid of [ConsoleConfigRoute](https://github.com/openshift/api/blob/master/operator/v1/types_console.go#L51) field that is configuring a custom route for the console ?

enhancements/ingress/custom-route-configuration.md
jhadvig: I suppose the ingress-operator will be setting those statuses ?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+    CustomizeableRoutes []CustomizeableRouteStatus
+}
+
+type CustomizeableRouteSpec struct{
+    // namespace is the namespace of the route to customize.  It must be a real namespace.
+    Namespace string
+    // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
+    // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
+    // this name.
+    Name string
```
jhadvig, line 97: I guess these names should be unique. Currently we have two routes in the console:
- console (default route name)
- console-custom (custom route name)

Should there be a list of these well-known routes for each concerned operator, so the given operator can match a behaviour based on the set `CustomizeableRouteSpec`?

enhancements/ingress/custom-route-configuration.md
Miciah: The section "Control loop to manage secret read permissions in openshift-config" references this field.  As I understand it, the operator that manages the route and its associated secrets sets this field to specify that its or its operand's service account needs access to the secret so that the operator can read the certificate and inject it into the route or into the operand.  For example, the console operator needs to read the secret in order to build the console route, and if the console route were passthrough, the console operator might also need inject the certificate into the console pod.  So the console operator needs to specify the console operator's service account in `ConsumingUsers` for the console CustomizeableRoute.

enhancements/ingress/custom-route-configuration.md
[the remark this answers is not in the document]
deads2k: > Do you mean that the ServerName is taken from the cert, but there is no custom configuration beyond that?

I wouldn't even take that.  I would use this as the default certificate for serving and configure SNI for the certificates that an operator controls, like service serving cert.

enhancements/ingress/custom-route-configuration.md
deads2k: > When console is accessed via `console.foo.com`, how does it know which auth and which api name it should use? I understand it would pick `DefaultHost` here and forward the user to the right auth URL or to use the right api URL in the kubeconfig the user can download.
> 
> @deads2k is this your understanding?

I would probably choose the first `.status.currentHosts` value.



> This also means that if you have multiple domains, e.g. `console.public-internet.com` and `console.intranet`, the console would have no way to know the matching auth and api URL. @deads2k is that correct?

I don't think this makes that worse one way or the other.

enhancements/ingress/custom-route-configuration.md
deads2k: I think #6 is up to the admin to decide whether to manually accept a certificate

enhancements/ingress/custom-route-configuration.md
deads2k: > where is the api in this list?

Currently uses a different API.  I'm open to collapsing, but we would have to solve how to specify SNI certificates to do so.

enhancements/ingress/custom-route-configuration.md
deads2k: > Does this mean that after this is implemented, we can get rid of [ConsoleConfigRoute](https://github.com/openshift/api/blob/master/operator/v1/types_console.go#L51) field that is configuring a custom route for the console ?

Yes, but at your leisure.  I don't have a plan that requires immediate refactoring.

deads2k:
Pushed updates to comments. Clarifications where requested and some renames.

@sttts appears concerned about SNI overall, anyone else share those concerns?

enhancements/ingress/custom-route-configuration.md
deads2k: > Should there be a list of these well-known routes for each concerned operator, so the given operator can match a behaviour based on the set `CustomizeableRouteSpec`?

We could write an e2e test that effectively has the list based on reading the state of the cluster at the end.  I'm ok documenting a list as best we can, but I think the most effective solution is just to look at the .status.componentRoutes

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,262 @@
+---
+title: custom-route-configuration
+authors:
+  - "@deads2k"
+reviewers:
+  - "@danmace"
+  - "@miciah"
+  - "@standa"
```
stlaz, line 8: nit: even all the way back when I started with GitHub, this handle was already taken, so I had to go with @stlaz

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,262 @@
...
+Those customers need a way to see all the routes they need to customize and a way to configure them.
+A stock installation has the following routes which may need customization of both names and serving cert/keys
+1. openshift-authentication   oauth-openshift
+2. openshift-console          console
+3. openshift-console          downloads
+4. openshift-monitoring       alertmanager-main
+5. openshift-monitoring       grafana
+6. openshift-monitoring       prometheus-k8s
+7. openshift-monitoring       thanos-querier
+8. image-registry (this isn't appearing in my installation, but @Miciah says it's there)
```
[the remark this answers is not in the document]
pweil-: @dmage @ricardomaraschini can comment on registry related items :+1:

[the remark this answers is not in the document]
dmage: the registry doesn't have routes by default, but you can enable it though config.imageregistry

enhancements/ingress/custom-route-configuration.md
```diff
@@ -72,82 +76,97 @@ slightly different documentation.
...
 
     // This API does not include a mechanism to distribute trust, since the ability to write this resource would then 
     // allow interception.  Instead, if we need such a mechanism, we can talk about creating a way to allow narrowly scoped
-    // updates to a configmap containing ca-bundle.crt for each CustomizeableRoute.
+    // updates to a configmap containing ca-bundle.crt for each ComponentRoute.
     // CurrentCABundle []byte
 }
 ```
 
+Validation rules to be specified in the openshift/api PR, but basic restrictions on strings as you'd expect.
```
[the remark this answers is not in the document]
deads2k: > These all are not necessarily API validation of the object, but something that should still exist in the operator before it pushes whatever suspicious configuration it is supplied. Some of these checks might make more sense even after the configuration is pushed on.

You cannot do these checks at API validation time, since references can change and the system is supposed to be eventually consistent with creation in any order.  I'm willing to recommend, "don't apply stuff that's broken", but it must not be done at API validation time because other changes (like secret updates) can take the content in an unworkable direction.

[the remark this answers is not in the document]
stlaz: Sure, my point is that we should add these recommendations as to how an operator should treat this new configuration before and after applying it, and how to report problems it finds.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,262 @@
...
+
+type IngressStatus struct {
+    // other fields
+
+    // ComponentRoutes is where participating operators place the current route status for routes which the cluster-admin
+    // can customize hostnames and serving certificates.
+    // How the operator uses that serving certificate is up to the individual operator.
+    // An operator that creates entries in this slice should clean them up during removal (if it can be removed).
+    // An operator must also handle the case of deleted status without churn.
+    ComponentRoutes []ComponentRouteStatus
```
dmage, line 97: So all operator that want to report about their routes will update status of the same resource? It can cause conflicts as the number of operators grow.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+    // desiredHost is the name that a cluster-admin wants to specify
+    DesiredHost string
+    // ServingCertKeyPairSecret is a reference to a secret in namespace/openshift-config that is a kubernetes tls secret.
+    // The serving cert/key pair must match and will be used by the operator to fulfill the intent of serving with this name.
+    // That means it could be embedded into the route or used by the operand directly.
+    // Operands should take care to ensure that if they use passthrough termination, they properly use SNI to allow service
+    // DNS access to continue to function correctly.
+    ServingCertKeyPairSecret SecretNameReference
+
+    // possible future, we could add a set of SNI mappings.  I suspect most operators would not properly handle it today.
```
s-urbaniak, line 119: correct me if I'm wrong: TLS termination is not done by the operators but still done by the router. The only capability operators have to offer is a) reconcile `CustomizeableRoute` objects and b) generate or modify existing `Route` resources they usually deploy as part of their manifests.

enhancements/ingress/custom-route-configuration.md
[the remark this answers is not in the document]
s-urbaniak: afaik today you would have to create a separate Route to host a different SNI certificate pointing pointing to a different hostname.

The reality though is that (in case of monitoring) we are deploying just **one** `Route` object and thus allow only tweaking only one hostname.

So keeping a 1:1 relation between a `Route` and a `CustomizeableRoutes` makes sort of sense. If you want to add additional SNI hosts, you'd have to support multiple routes in the original operator.

enhancements/ingress/custom-route-configuration.md
sttts: > I don't think this makes that worse one way or the other.

the choice is important if not all URLs can be accessed from the internet.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,242 @@
...
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
+
+### Upgrade / Downgrade Strategy
+
+For components that did not try to solve this in a one-off way, the upgrade is easy, the new feature becomes available.
+
+For components that created a one-off solution, the upgrade will vary depending on the steps they used.
+I seem to recall that monitoring allowed direct route manipulation and the console created its own fields.
```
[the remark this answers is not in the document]
s-urbaniak: Correct, CMO does not support route customization as of today.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -72,82 +76,97 @@ slightly different documentation.
...
+    // ensures that no two components will conflict and the same component can be installed multiple times.
     Namespace string
     // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
     // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
     // this name.
     Name string
-    // desiredHost is the name that a cluster-admin wants to specify
-    DesiredHost string
+    // Hostname is the host name that a cluster-admin wants to specify
+    Hostname string
```
sttts, line 109: still open question: what if there are multiple?

enhancements/ingress/custom-route-configuration.md
sttts: @deads2k this is still open

enhancements/ingress/custom-route-configuration.md
sttts: Depending on the component. If you use terminating routes now, your operator has to wire the cert into the route.

s-urbaniak:
Sorry for the late reply from my side, but could we add some more guidelines for operators how they are supposed to consume this API? From what I understand, based on cluster-monitoring-operator as an example:

a) Document all well-known logical `CustomizeableRoute` names in the Monitoring documentation, i.e. `prometheus`, `thanos-querier`, `grafana`, etc. This is the "API" part from the perspective of CMO.
a) Watch for those well-known `CustomizeableRoute` resources in the `openshift-monitoring` namespace.
b) Modify existing already deployed `Route` objects by leveraging the metadata in the logical `CustomizeableRoute` and more or less "copying" the relevant fields. Especially for this point it would be great to have a bit more guidance in this OEP.

Generally: the operator watching this API (in our case cluster-monitoring-operator) does not handle any low level things like TLS termination, or even reading out secrets, but just copying secret _references_ only.

Please correct me if I'm saying anything silly here :-)

enhancements/ingress/custom-route-configuration.md
deads2k: > @deads2k will this replace the existing cert? I.e. if the installer creates `console.asdfhzwdsf4322.openshift.com` as external cluster URL, and on the second day the admin sets `console.my-pretty-cluster.com` by providing a matching cert, the old name is gone?

I don't think we expect operators to handle the creation of multiple routes.  This cert would be used for serving content via the route and if the cluster-admin wants to adjust DNS to use a CNAME, they still can, right?

enhancements/ingress/custom-route-configuration.md
deads2k: > So all operator that want to report about their routes will update status of the same resource? It can cause conflicts as the number of operators grow.

well, operators are controllers with conflict resolution, so I don't expect any issues.

enhancements/ingress/custom-route-configuration.md
deads2k: > still open question: what if there are multiple?

Routes can only specify a single name (see https://github.com/openshift/api/blob/master/route/v1/types.go#L80).  I don't understand the question.  Are you proposing an API that coerces operators into the creation of multiple routes?

enhancements/ingress/custom-route-configuration.md
deads2k: > > I don't think this makes that worse one way or the other.
> 
> the choice is important if not all URLs can be accessed from the internet.

The route API doesn't know which it is either.

deads2k:
> Sorry for the late reply from my side, but could we add some more guidelines for operators how they are supposed to consume this API? From what I understand, based on cluster-monitoring-operator as an example:
> 
> a) Document all well-known logical `CustomizeableRoute` names in the Monitoring documentation, i.e. `prometheus`, `thanos-querier`, `grafana`, etc. This is the "API" part from the perspective of CMO.
> a) Watch for those well-known `CustomizeableRoute` resources in the `openshift-monitoring` namespace.
> b) Modify existing already deployed `Route` objects by leveraging the metadata in the logical `CustomizeableRoute` and more or less "copying" the relevant fields. Especially for this point it would be great to have a bit more guidance in this OEP.

That is the flow proposed here.  I didn't/don't have plans to make operators expose multiple routes to access their components. Though such a thing could be built, the requirement as I understand it is to create a way to avoid wildcard certs, allow vanity DNS names, and allow serving certificate customization.  Supporting multiple routes appears to be outside the scope of the  need.



> Generally: the operator watching this API (in our case cluster-monitoring-operator) does not handle any low level things like TLS termination, or even reading out secrets, but just copying secret _references_ only.

The route directly embeds the secret content, so the operator is responsible for doing reasonable validation on input, reflecting status back into this API, and injecting secrets where they are required.

enhancements/ingress/custom-route-configuration.md
s-urbaniak: i thought about the same and (sorry for xrefing) as commented here https://github.com/openshift/enhancements/pull/577#discussion_r565146419 a 1:1 relationship between a `Route` and `CustomizeableRoute` seems reasonable.

If the user wants multiple hostnames for a given use case (i.e. multiple external hostnames for an externally visible Prometheus), the underlying operator a) would need to support creating multiple routes (one for each hostname) and b) support for allowing customizing each of them. This is equivalent to what the user (to my knowledge) needs to do, namely create seperate one route per hostname.

Does that sound reasonable?

enhancements/ingress/custom-route-configuration.md
s-urbaniak: Ah, that makes sense, thanks! We use reencrypt routes pointing to services so I guess the service CA is being used implicitly then as we don't set it explicitly in cluster-monitoring-operator.

enhancements/ingress/custom-route-configuration.md
sttts: Not talking about the route, but about console and oauth-server choosing the right URL.

deads2k:
/hold

waiting for information about whether we need to support multiple routes per logical component. Such an API would likely look different and significantly raise the bar for implementation by operators.  At the very least, we're likely to key it differently and have a different multiplicity.

Anandnatraj:
> /hold
> 
> waiting for information about whether we need to support multiple routes per logical component. Such an API would likely look different and significantly raise the bar for implementation by operators. At the very least, we're likely to key it differently and have a different multiplicity.

@deads2k @sttts based on customer requirements, we only need to only replace default route with a custom route, there is no need to support multiple routes.

sttts:
@danmace @miciah @standa @stlaz @sur @lilic can everybody please do a final pass so that we can finalize the enhancement?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,262 @@
...
+    Namespace string
+    // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
+    // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
+    // this name.
+    Name string
+    // defaultHostname is the normal host name of this route.  It is provided in case cluster-admins find it more recognizeable
+    // and having it here makes it possible to answer, "if I remove my configuration, what will the name be".
+    DefaultHostname string
+    // ConsumingUsers is a slice of users that need to have read permission on the secrets in order to use them.
+    // This will usually be an operator service account.
```
awgreene, line 134: Is the plan for the users/SA to always exist in the same namespace as the route?

awgreene: It seems like an operators could create a route outside of their namespace.

awgreene, line 134: From my understanding the secret will be created in the openshift/config namespace. The consumer of this secret needs to be granted RBAC to read this secret. The only namespace information we get from the `ComponentRouteStatus` is the namespace of the `Route`. Is there a requirement for the ServiceAccount to exist in the same namespace as the route?

deads2k: a serviceaccount username is `system:serviceaccount:<namespace>:<sa-name>`



> Is there a requirement for the ServiceAccount to exist in the same namespace as the route?

no.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,262 @@
...
+# Custom Route Configuration
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
```
sttts, line 32: it needs some checks here (follow-up is fine)

enhancements/ingress/custom-route-configuration.md
sttts: @deads2k ^

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,262 @@
...
+    // this name.
+    Name string
+    // Hostname is the host name that a cluster-admin wants to specify
+    Hostname string
+    // ServingCertKeyPairSecret is a reference to a secret in namespace/openshift-config that is a kubernetes tls secret.
+    // The serving cert/key pair must match and will be used by the operator to fulfill the intent of serving with this name.
+    // That means it could be embedded into the route or used by the operand directly.
+    // Operands should take care to ensure that if they use passthrough termination, they properly use SNI to allow service
+    // DNS access to continue to function correctly.
+    // SANs in the certificate are ignored, but SNI can be used to make operator managed certificates (like internal load balancers
```
sttts, line 115: what does "SANs in the cert are ignore" mean? Please clarify.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,262 @@
...
+
+    // possible future, we could add a set of SNI mappings.  I suspect most operators would not properly handle it today.
+}
+
+type ComponentRouteStatus struct{
+    // namespace is the namespace of the route to customize.  It must be a real namespace.  Using an actual namespace
+    // ensures that no two components will conflict and the same component can be installed multiple times.
+    Namespace string
+    // name is the *logical* name of the route to customize.  It does not have to be the actual name of a route resource.
+    // Keep in mind that this is your API for users to customize.  You could later rename the route, but you cannot rename
```
sttts, line 127: misplaced in API docs. You don't talk to the dev here, but the admin.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,262 @@
...
+    // relatedObjects allows listing resources which are useful when debugging or inspecting how this is applied.
+    // They may be aggregated into an overall status RelatedObjects to be automatically shown by oc adm inspect
+    RelatedObjects []ObjectReference
+    
+    
+
+    // This API does not include a mechanism to distribute trust, since the ability to write this resource would then 
+    // allow interception.  Instead, if we need such a mechanism, we can talk about creating a way to allow narrowly scoped
+    // updates to a configmap containing ca-bundle.crt for each ComponentRoute.
+    // CurrentCABundle []byte
```
sttts, line 156: what is this? Scratch pad? Future enhancement?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+Validation rules to be specified in the openshift/api PR, but basic restrictions on strings as you'd expect.
+
+
+### User Stories
+
+#### Cluster-admin wants to customize a route
+To use this, a cluster-admin would
+1.  Either read docs to find the namespace,name tuples or look at an existing cluster and read ingresses.config.openshift.io.
```
sttts, line 167: nit: `,<space>`

enhancements/ingress/custom-route-configuration.md
sttts: but technically nothing stops them, just permissions.

sttts:
Some nits, and style issues left (first person voice is irritating in this document). Other than that SGTM.

@Miciah ready to merge?

sttts:
@deads2k can you fix the first person voice? This is strange in a technical document. This document is not the voice of a person, but a technical document describing what teams are building and why.

enhancements/ingress/custom-route-configuration.md
awgreene: At this time, OLM [does not support shipping routes](https://github.com/operator-framework/operator-registry/blob/master/pkg/lib/bundle/supported_resources.go#L27-L44) in Operator Bundles, so the operator would need to implement support for this feature.

OLM could certainly add support mind you.

sttts: What does this mean for security? If I can write Ingress status, I can let me give permissions to any openshift-config secret? @deads2k this needs a discussion why this is ok or how to mitigate the problem

awgreene: That point is called out in the enhancement [here](https://github.com/openshift/enhancements/pull/577/files#diff-5a2afb792e11918fa5b74f9d217fc641a3af897066968cd2449e373f25f964abR183-R185) but doesn't seem to have garnered much conversation.

>This means that the power to mutate ingresses.config.openshift.io and ingresses.config.openshfit.io/status
will imply the power to read secrets in openshift-config. If we decide we do not wish to do this, then it will incumbent upon the cluster-admin to create the role and rolebinding.

awgreene: Note that you need to write both ingresses.config.openshift.io and ingresses.config.openshift.io/status.

enhancements/ingress/custom-route-configuration.md
deads2k: > misplaced in API docs. You don't talk to the dev here, but the admin.

I'm sure it will work itself out in the API PR.  Tihs expresses the intent pretty clearly.

enhancements/ingress/custom-route-configuration.md
deads2k: > what is this? Scratch pad? Future enhancement?

Something to consider.  Again, this will work itself out in the api PR.

deads2k:
> @deads2k can you fix the first person voice? This is strange in a technical document. This document is not the voice of a person, but a technical document describing what teams are building and why.

updated for you.

sttts:
/lgtm
/approve

Further changes as follow-up.

PR #581, opened 2021-01-11
Add network flows export support proposal

rcarrillocruz:


enhancements/network/netflow.md
```diff
@@ -0,0 +1,110 @@
...
+### Non-Goals
+
+- This document only applies to OVN-Kubernetes, not OpenShift SDN.
+- This document will not the flow logs store solutions. Typically, customers store the NetFlow data in a data warehouse system (e.g. ClickHouse) or search and analytics system (e.g. ElasticSearch). It is out of the scope of this document to discuss which option should be used and if we should manage that solution. 
+
+## Proposal
+
+The Cluster Network Operator (CNO) should expose options to enable the NetFlow data from OVS bridges.
+OVS supports NetFlow v5, IPFIX and sflow. NetFlow v5 does not support IPv6, therefore we should only consider as protocol options
+both IPFIX/sflow or pick one of them. Sflow is a more performant as it is a sampling protocol and does not rely on templates.
```
russellb, line 54: Up to this point, the enhancement implies NetFlow is the protocol.  Would you like to include a section in this doc about netflow vs. ipfix vs. sflow?  I have some notes on this I can share with you.  I see the goflow README has a high level overview, as well.  I think explaining it in this enhancement would be helpful though.

It seems we may need to support all of them anyway if we want to support customers' own collectors.  Then we can just pick a default to use for our own collector that dumps the info to logs?

... and if we're supporting all the protocols, then maybe the language should be changed throughout the enhancement to say something generic like "network flow collection" instead of "NetFlow" specifically?

enhancements/network/netflow.md
```diff
@@ -0,0 +1,110 @@
...
+
+* OpenShiftSrcPod
+* OpenShiftDstPod
+* OpenShiftSrcPodNamespace
+* OpenShiftDstPodNamespace
+* OpenShiftSrcNode (when the source address is a node IP)
+* OpenShiftDstNode (when the destination address is a node IP)
+
+This enrichment could be done by either modifying the collector source code or by using an adapter container pattern, where the collector dumps the stdout of the flows
+collected to a volume and that is consumed and enriched by a Go Kubernetes client that dumps the enriched flow to stdout/file.
```
russellb, line 83: What are some other examples of this "adapter container pattern" ?

enhancements/network/netflow.md
rcarrillocruz: yeah,makes sense.
sflow has a lesser performance impact at the expense of being sampled (i.e. it won't take all packets).
But I can see customers taking both, so maybe we should change the title to network flow collection as you suggest.

enhancements/network/netflow.md
rcarrillocruz: yeah, I can probably reference, the k8s official docs even have a page explaining the different container patterns and adapter is also explained into detail.

enhancements/network/netflow.md
```diff
@@ -0,0 +1,107 @@
...
+
+OVS (Open vSwitch) supports exporting network flow protocols like NetFlow, IPFIX and sflow.
+OpenShift uses OVN/OVS for OVN-Kubernetes network plugin. 
+OpenShift should have the ability to export cluster traffic using any of this protocols.
+
+## Motivation
+
+* Customers need to be able to get network flows out of OpenShift to be consumed by their collectors.
+* Customers need to be able to see a log of the network traffic that flows thru OpenShift cluster.
+* Customers need to get metrics and monitoring data of the OpenShift cluster network traffic to make decisions on capacity and/or security issues.
```
russellb, line 34: Instead of Customers, how about Cluster Administrators?  I think that more directly identifies the target persona

enhancements/network/netflow.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Goals
+
+- Add support for exporting all the network flows protocols supported by OVS from OpenShift.
+- Add an optional collector that consumes the network flows data and enriches it with Kubernetes metadata.
```
russellb, line 39: We should probably set a goal for what we want our optional collector to be able to do with this data.  

Maybe ... The first target use case for our optional collector is to dump flow data to logs.

and then later you clarify that log handling is out of scope for this enhancement.  I'm tempted to at least give cluster logging as an obvious optional solution here.

enhancements/network/netflow.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Goals
+
+- Add support for exporting all the network flows protocols supported by OVS from OpenShift.
+- Add an optional collector that consumes the network flows data and enriches it with Kubernetes metadata. 
+- Add Prometheus metrics from the data collected by the optional collector for visualization and alerting purposes.
+
+### Non-Goals
+
+- This document only applies to OVN-Kubernetes, not OpenShift SDN.
```
russellb, line 44: or any other network type

enhancements/network/netflow.md
```diff
@@ -0,0 +1,107 @@
...
+
+OVS supports NetFlow v5, IPFIX and sflow.
+NetFlow v5 does not support IPv6, thus it should not be considered as a default protocol
+sFlow is the most performant protocol as it is a sampling protocol, which polls periodically interfaces to get samples of packets. The drawback is that it does not capture all packets.
+IPFIX is capable of capturing all traffic at line rate and supports IPv6, therefore it should be the default protocol if this option is not specified.
+
+The Cluster Network Operator (CNO) should expose:
+* A *exportNetworkFlows* option to enable the export of network flows data from OVS bridges
+* A *exportNetworkFlowsProtocol* to specify which protocol should be used. Valid values should be *netflowV5*, *ipfix* and *sflow*, being the default ipfix.
+* A *exportNetworkFlowsCollectorIP* to specify the IP of the collector that will consume the flow data
```
russellb, line 57: Would it be expected to only have a single collector IP?  How does this scale out?  Do we need to plan ahead for when we need to shard the cluster across multiple collectors?

enhancements/network/netflow.md
```diff
@@ -0,0 +1,107 @@
...
+* Customers need to be able to see a log of the network traffic that flows thru OpenShift cluster.
+* Customers need to get metrics and monitoring data of the OpenShift cluster network traffic to make decisions on capacity and/or security issues.
+
+### Goals
+
+- Add support for exporting all the network flows protocols supported by OVS from OpenShift.
+- Add an optional collector that consumes the network flows data and enriches it with Kubernetes metadata. 
+- Add Prometheus metrics from the data collected by the optional collector for visualization and alerting purposes.
+
+### Non-Goals
```
russellb, line 42: Another non-goal to call out is that we're not doing any visualization in this first step

enhancements/network/netflow.md
rcarrillocruz: done

enhancements/network/netflow.md
rcarrillocruz: done

enhancements/network/netflow.md
```diff
@@ -0,0 +1,107 @@
...
+- This document will not discuss the flow logs store solutions. Typically, customers store the NetFlow data in a data warehouse system (e.g. ClickHouse) or search and analytics system (e.g. ElasticSearch). It is out of the scope of this document to discuss which option should be used and if we should manage that solution. 
+
+## Proposal
+
+OVS supports NetFlow v5, IPFIX and sflow.
+NetFlow v5 does not support IPv6, thus it should not be considered as a default protocol
+sFlow is the most performant protocol as it is a sampling protocol, which polls periodically interfaces to get samples of packets. The drawback is that it does not capture all packets.
+IPFIX is capable of capturing all traffic at line rate and supports IPv6, therefore it should be the default protocol if this option is not specified.
+
+The Cluster Network Operator (CNO) should expose:
```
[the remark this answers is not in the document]
rcarrillocruz: Exporting netflow/sflow off OVS is an OVS thing, not SDN or OVN-Kubernetes. OTOH, my understanding is that SDN is no longer getting new features.
It's not hard to add to SDN, just a matter of templatize in the SDN manifests.

enhancements/network/netflow.md
rcarrillocruz: Ack, will add.

enhancements/network/netflow.md
rcarrillocruz: The way I wrote the POC for this is to have a Kubernetes Service of type ClusterIP in front of the collectors.
OVS does not allow multiple target collector IPs as a parameter to load balance, thus we need to do that ourselves.

enhancements/network/netflow.md
rcarrillocruz: done

enhancements/network/netflow.md
rcarrillocruz: done

enhancements/network/netflow.md
```diff
@@ -0,0 +1,107 @@
...
+* A *exportNetworkFlowsCollectorIP* to specify the IP of the collector that will consume the flow data
+* A *exportNetworkFlowsCollectorPort* to specify the port of the collector that will consume the flow data
+
+Under the covers, these options will make the CNO to perform an ovs-vsctl command on the bridge br-int, which is the bridge the containers are connected to on OVN-Kubernetes, and its management port ovn-k8s-mp0.
+As an example, if sFlow was enabled this is the command that would be executed.
+
+`ovs-vsctl -- --id=@sflow create sflow agent=ovn-k8s-mp0 target="\"${exportNetworkFlowsCollectorIP}:{exportNetworkFlowsCollectorPort}\"" header=128 sampling=64 polling=10 -- set bridge br-int sflow=@sflow`
+
+If *exportNetworkFlowsCollectorIP* is not provided, we deploy a collector based in the open-source GoFlow project, expose it with a service and we set the service ClusterIP as its value.
+There are many open-source NetFlow collectors. Just to name a few:
```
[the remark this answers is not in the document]
rcarrillocruz: done

enhancements/network/netflow.md
rcarrillocruz: Not quite follow you on this, what you mean by optional solution here.
I see cluster logging as an add-on, i.e. the optional collector dumps the flow data into stdout which then will be handled by cluster logging for aggregation, analytics, searching etc.
You mean suggesting that as a future use case?

enhancements/network/netflow.md
rcarrillocruz: As discussed, removing the optional collector and enrichment to scope just the export of flows.

enhancements/network/netflow.md
rcarrillocruz: Resolving as we are re-scoping the proposal for just the flows export.

enhancements/network/netflow.md
```diff
@@ -0,0 +1,107 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+OVS (Open vSwitch) supports exporting network flow protocols like NetFlow, IPFIX and sflow.
+OpenShift uses OVN/OVS for OVN-Kubernetes network plugin. 
+OpenShift should have the ability to export cluster traffic using any of this protocols.
```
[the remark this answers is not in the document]
rcarrillocruz: done

enhancements/network/netflow.md
```diff
@@ -0,0 +1,92 @@
...
+There are existing protocols for how this metadata is transmitted between systems, including NetFlow, IPFIX, and sFlow.
+Open vSwitch (OVS) supports these protocols and OVS is a fundamental building block of the OVN-Kubernetes network type.
+
+This proposal discusses how we start providing this functionality by allowing cluster administrators to enable the export of network flow metadata using one of these protocols.
+
+## Motivation
+
+* Cluster administrators need to be able to get network flows out of OpenShift to be consumed by their collectors.
+* Cluster administrators need to be able to see a log of the network traffic that flows thru OpenShift cluster.
+* Cluster administrators need to get metrics and monitoring data of the OpenShift cluster network traffic to make decisions on capacity and/or security issues.
```
russellb, line 40: I think these last 2 motivations are probably for the follow-up enhancement where we discuss our built-in collector

enhancements/network/netflow.md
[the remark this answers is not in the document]
russellb: Right - I figured we'd only support for OVN-Kubernetes.  i was wondering if it would be exposed through the configuration in a way that's OVN-Kube specific to make that clear.

enhancements/network/netflow.md
russellb: want to capture that detail in the proposal here?

russellb (commented):
this is looking good.  I just had a few minor comments

enhancements/network/netflow.md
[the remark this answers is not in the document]
rcarrillocruz: Ah right.  Yeah, it will be under OVNKubernetesConfig, to point that out.

https://github.com/openshift/api/pull/846/files#diff-2e98c189fdf696589c6d28124557b9cf96a329e82f790b4e61209351cc68648dR330

enhancements/network/netflow.md
rcarrillocruz: done

enhancements/network/netflow.md
rcarrillocruz: I take that back, I randomly found a blogpost where apparently it is possible, yet it's not mentioned on the docs:

https://blog.shin.do/2014/12/netflow-on-open-vswitch-2/

I'm torn about having a list of collectors. It feels to me an anti-pattern to make OVS send flow data to an arbitrary list of IPs as that puts the churn on us.
Maybe mention here 'even though OVS is capable to send to multiple collectors, we just support one IP to avoid CPU consumption on OpenShift side. If multiple collectors are used a K8S Service or LB can be used as the collector IP to distribute the load' or something along those lines.
Thoughts?

russellb: I checked the OVS code.  Specifying multiple collectors will have OVS send the data to ALL of them.  It seems "free" to support this.  If you want to send to more than one, you list multiple IPs and OVS will send to all of them.  Otherwise, people can use a LB IP (whether that's a kube Service or external LB, whichever).  We can document the options for people.

enhancements/network/netflow.md
```diff
@@ -0,0 +1,89 @@
...
+
+This proposal discusses how we start providing this functionality by allowing cluster administrators to enable the export of network flow metadata using one of these protocols.
+
+## Motivation
+
+* Cluster administrators need to be able to get network flows out of OpenShift to be consumed by their collectors.
+
+### Goals
+
+- Add support for exporting the network flows traffic in an OpenShift cluster by leveraging the NetFlow/sflow/IPFIX protocols supported by OVS.
```
jackevans43, line 42: IMO IP address will need to be resolved to namespace/podname, otherwise I can't see how anybody could make any use of the flows given the transient nature of IPs in Kubernetes? Maybe new IPFIX octetArray entity would need to be added for source and destination pod?

rcarrillocruz: The enrichment with K8S data is out of the scope on this proposal, this is just for exporting the data.
There are a number of ways to enrich depending on tools used, I wrote a POC by injecting the data directly within the collector by using Cloudflare GoFlow:

https://github.com/rcarrillocruz/goflow/tree/enrich_k8s

If skydive is used, we can write a flow exporter that enriches it with the K8S discovered data.

TL;DR it will come, in other proposal.

enhancements/network/netflow.md
rcarrillocruz: Alright then, will change the options to accept a list.

enhancements/network/netflow.md
russellb: Yeah, agree that we need to do it, and also agree with covering that in a follow-up next-step enhancement

enhancements/network/netflow.md
```diff
@@ -0,0 +1,121 @@
...
+- This document will not discuss the flow logs store solutions. Typically, customers store the NetFlow data in a data warehouse system (e.g. ClickHouse) or search and analytics system (e.g. ElasticSearch). It is out of the scope of this document to discuss which option should be used and if we should manage that solution.
+- This document will not discuss the aggregation of flow data with corresponding K8S pods.
+- This document will not discuss flow visualization or flow dashboards.
+
+## Proposal
+
+OVS supports NetFlow v5, IPFIX and sflow.
+NetFlow v5 does not support IPv6, thus it should not be considered as a default protocol
+sFlow is the most performant protocol as it is a sampling protocol, which polls periodically interfaces to get samples of packets. The drawback is that it does not capture all packets.
+IPFIX is capable of capturing all traffic at line rate and supports IPv6, therefore it should be the default protocol if this option is not specified.
```
russellb, line 56: This seems reasonable.  Measurement of the performance impact could change our choice on this, I suppose.  I also wonder if having a default value is that valuable in this case since whoever is turning this on is going to want to make the choice, I would think?

enhancements/network/netflow.md
```diff
@@ -0,0 +1,121 @@
...
+
+The Cluster Network Operator (CNO) would expose:
+* *exportNetworkFlowsConfig* dict to contain the options to export the network flows
+  * *exportNetworkFlowsProtocol* to specify which protocol should be used. Valid values should be *netflow*, *ipfix* and *sflow*, being the default ipfix.
+  * *exportNetworkFlowsCollectors* list to contain the collectors IPs and ports
```
russellb, line 61: maybe add a note clarifying that specifying more than one IP here means the flows are sent to ALL of them

russellb (commented):
/approve

I'm happy with the direction here.  My last comments are trivial.  I'll leave the lgtm to one more person to help ACK your proposed config layout, but the details you've included seem good to me.

enhancements/network/netflow.md
rcarrillocruz: done

enhancements/network/netflow.md
rcarrillocruz: Alright, then let's make it required and not default.
I agree, the choice has greater implications and also it seems the most popular is netflow (at least is what I get most hits when looking for blogposts, articles, configuration doc pages, etc).

TL;DR there doesn't seem to be an easy choice, so let's make the admin responsible for that.

enhancements/network/netflow.md
rcarrillocruz: k, will do

rcarrillocruz: done

rcarrillocruz:
@russellb when you get a chance, can you please do a final pass?

enhancements/network/netflow.md
```diff
@@ -0,0 +1,121 @@
...
+- This document will not discuss the aggregation of flow data with corresponding K8S pods.
+- This document will not discuss flow visualization or flow dashboards.
+
+## Proposal
+
+OVS supports NetFlow v5, IPFIX and sflow as export flows protocol. It also supports a list of collectors by specifying IP and port to send the flow data.
+As such, we would support the three protocols and an arbitrary list of collectors.
+
+The Cluster Network Operator (CNO) would expose:
+* *exportNetworkFlowsConfig* dict to contain the options to export the network flows
```
danwinship, line 57: (style nitpick; use backticks, not asterisks, around API fields)

enhancements/network/netflow.md
```diff
@@ -0,0 +1,121 @@
...
+    hostPrefix: 23
+  defaultNetwork:
+    ovnKubernetesConfig:
+      exportNetworkFlowsConfig:
+        exportNetworkFlowsCollectors:
+        - exportNetworkFlowsCollectorIP: 172.30.158.150
+          exportNetworkFlowsCollectorPort: 2056
+        - exportNetworkFlowsCollectorIP: 172.30.54.103
+          exportNetworkFlowsCollectorPort: 2056
+        exportNetworkFlowsProtocol: netflow
```
danwinship, line 78: Not sure how much I should be arguing about the API details here rather than in the eventual API PR, but...

1. Not sure this belongs under `ovnKubernetesConfig`? Even though we only support it in ovn-kubernetes currently, these are generic protocols which *could* be supported by any network plugin. eg, like how we have `kubeProxyConfig` as a top-level rather than inside `openshiftSDNConfig`. Not totally sure about this one...
2. You probably don't need `Config` in the name? Just `exportNetworkFlows` explains what that section is doing
3. Get rid of the naming redundancy in the subfields. eg, consider what accessing this would look like from golang code:

        ip := defaultNetwork.exportNetworkFlowsConfig.exportNetworkFlowsCollectors[0].exportNetworkFlowsCollectorIP

    You can just have `collectors` and `ip`/`port`

4. Kubernetes API enums always have initial caps and then proper case. So, `NetFlow`, `IPFIX`, `SFlow`
5. Is there any reason we allow multiple collectors of a single type but not multiple collectors of different types? If you put the protocol, ip, and port together, then it would be more flexible, and also slightly simpler to write out in the case where you only have a single collector (which I think is the common case?):

        exportNetworkFlows:
          - protocol: NetFlow
            ip: 172.30.158.150
            port: 2056

6. The big feature of sFlow over NetFlow is that it lets you export only a sample of the traffic rather than all the traffic. Any plans to allow configuring that, now or later?

enhancements/network/netflow.md
```diff
@@ -0,0 +1,121 @@
...
+  disableNetworkDiagnostics: false
+  logLevel: Normal
+  managementState: Managed
+  observedConfig: null
+  operatorLogLevel: Normal
+  serviceNetwork:
+  - 172.30.0.0/16
+```
+
+Under the covers, these options will make the CNO to perform an ovs-vsctl command on the bridge br-int, which is the bridge the containers are connected to on OVN-Kubernetes, and its management port ovn-k8s-mp0.
```
danwinship, line 93:
```suggestion
Under the covers, these options will make the CNO perform an ovs-vsctl command on the bridge br-int, which is the bridge the containers are connected to on OVN-Kubernetes, and its management port ovn-k8s-mp0.
```

danwinship: Maybe not worth arguing about here but would this actually be done by CNO? I'd expect CNO to write out something to the ovn-kube config, and ovn-kube to modify `br-int`

enhancements/network/netflow.md
```diff
@@ -0,0 +1,121 @@
...
+As an example, if netflow was enabled per the above example this is the command that would be executed.
+
+`ovs-vsctl -- --id=@netflow create netflow targets=\[\"172.30.158.150:2056\",\"172.30.54.103:2056\"\] -- set bridge br-int netflow=@netflow`
+
+This would make OVS to send the flow data to both collectors 172.30.158.150 and 172.30.54.103 on port 2056 by using netflow protocol.
+
+### Test Plan
+
+- Unit tests for the feature
+- e2e tests covering the feature
```
danwinship, line 103: Test plan is pretty vague; do we even know what software we could use to act as the collector?

enhancements/network/netflow.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Goals
+
+- Add support for exporting the network flows traffic in an OpenShift cluster by leveraging the NetFlow/sflow/IPFIX protocols supported by OVS.
+
+### Non-Goals
+
+- This document only applies to OVN-Kubernetes, not OpenShift SDN or any other other network plugin.
+- This document will not discuss the flow logs store solutions. Typically, customers store the NetFlow data in a data warehouse system (e.g. ClickHouse) or search and analytics system (e.g. ElasticSearch). It is out of the scope of this document to discuss which option should be used and if we should manage that solution.
+- This document will not discuss the aggregation of flow data with corresponding K8S pods.
```
danwinship, line 48: Do we have any plan for that? Might be good to at least hint about it if so

danwinship (commented):
This is basically good and maybe all of my comments are better addressed in the actual implementation PRs rather than in the enhancement PR?
/lgtm
/hold
if you want to merge it as-is just cancel the hold

enhancements/network/netflow.md
rcarrillocruz: 1. Right, I thought about that. Thing is I saw IPSEC within ovnKubernetesConfig and conceptually it's not really an OVN-K related feature, it could also be leveraged by other plugin. Given that it's the latest thing added, I just cargo culted. I don't have a strong opinion.
2. I cargo culted HybridOverlayConfig and IPSecConfig
3. Yeah :D , I was actually looking for this name bikeshedding feedback. I see hybrid overlay options prepended with HybridOverlay, thus I just cargo culted. But I agree, probably sounds better not to prepend, since it's a dict it's clear those options apply to 'export network flows' .
4. I'd rather accept any case and just canonicalize as we do with other options. Note that sflow is usually written as "sFlow" per https://sflow.org/ .
5. Here you get me off-guard. The docs don't clarify if you can have multiple collector protocols defined, I'd need to test. I take you suggest having something like "endpoint: 172.30.158.140:2056"
5. Yes it does, however I was thinking of going light with exposing per-protocol options. In others proposals you were against having toggles on stuff to avoid end-user overconfigurability, e.g. the network diagnostics interval, so I thought of just hardcoding them and if there's end-user demand/RFE then expose . But I'm cool with exposing sflow sampling and polling params now.

enhancements/network/netflow.md
rcarrillocruz: @russellb I think you also raised about hinting at something here? Thoughts?

enhancements/network/netflow.md
rcarrillocruz: Yeah, I guess we can just leverage goflow, since it has a docker image in the hub and is easy to use. Will add.

enhancements/network/netflow.md
rcarrillocruz: Why? To enable any of the export protocols you just run an ovs-vsctl command.
In my POC, I put it on the 006 -ovs-node.yaml template, seems simple enough and it's also the same pattern that ipsec feature followed.

enhancements/network/netflow.md
danwinship: > Not sure this belongs under `ovnKubernetesConfig`?

I guess another counter-argument is that even if we supported exactly the same flow collecting options in openshift-sdn, we could do that by just adding the same `ExternalNetworkFlowsConfig` struct to `openshiftSDNConfig` too. OK.

> I cargo culted HybridOverlayConfig and IPSecConfig

which maybe we should have done differently...

`exportNetworkFlowsConfig` struck me as particularly awkward though. If you had said `networkFlowExportConfig` I probably wouldn't have said anything... I guess "verb phrase" = OK, "noun phrase + Config" = OK, "verb phrase + Config" = :slightly_frowning_face: 

> I see hybrid overlay options prepended with HybridOverlay, thus I just cargo culted.

`HybridOverlayVXLANPort` is definitely wrong and I think I even complained about that

> I'd rather accept any case and just canonicalize as we do with other options.

We mostly don't do that any more; we only originally did that for the network plugin type because the original capitalization was wrong (`OpenshiftSDN` rather than `OpenShiftSDN`) and we fixed it right before 4.1 code freeze and didn't want to break anything. But then this led to other bugs because while *CNO* handles either variant, stuff in other components didn't (eg, MCO deciding whether to enable system OVS) and so random things would break if you had used the wrong case. In the end, it's easier to just require everyone to get the string correct (and use validation to only accept the correct spelling/casing).

> The docs don't clarify if you can have multiple collector protocols defined

oh, hm, I just assumed you could since they're totally separate tables...

> In others proposals you were against having toggles on stuff to avoid end-user overconfigurability

It's true. I guess we can add more options later if we need them. My vague understanding though was that pretty much the main reason anyone used sFlow or IPFIX rather than NetFlow was to get sampling.

enhancements/network/netflow.md
danwinship: 1. `br-int` is an internal detail of ovn-kubernetes and CNO should not be making assumptions about it. That's the sort of thing that could easily change in the future. (Or maybe local gateway and shared gateway do it differently. Or maybe some update to the external gateway handling will result in some traffic not crossing `br-int`. Etc.)
2. Non-OCP users of ovn-kubernetes may want to enable flow export too

rcarrillocruz: Oh, so you mean making this an ovn-k change:

@trozet @russellb can you please chime in to get consensus?
I always thought on doing this from CNO, per https://github.com/openshift/cluster-network-operator/pull/982 .

But if you all agree this should be upstreamed to OVN-K, then that's fine. I assume the change would be on https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/node/node.go ?

enhancements/network/netflow.md
rcarrillocruz: It is, it's just I'm hardcoding the sampling/polling values, from my WIP branch:

https://github.com/openshift/cluster-network-operator/pull/982/files#diff-d1edeaf791eef7719efb5bb733f48f3a201cc1000ce6f8d09a917be2accc9cb4R86

```
 ovs-vsctl -- --id=@sflow create sflow agent=ovn-k8s-mp0 targets={{ .ExportNetworkFlowsCollectors }} header=128 sampling=64 polling=10 -- set bridge br-int sflow=@sflow
```

enhancements/network/netflow.md
rcarrillocruz: done

enhancements/network/netflow.md
rcarrillocruz: Anyway, it's not worth discussing the implementation detail here.
Let's have that discussion over slack or PRs.

enhancements/network/netflow.md
rcarrillocruz: @danwinship please take a look, I put up an alternative based slightly on your suggestions.
Since there can be either netflow, ipfix or sflow exports, have those as dicts and as values for those dicts a list of strings encompassing the IP and the port , i.e. no need to add yet another key.

rcarrillocruz: In the end, I'm making the exportNetworkFlows option top-level, I'll check on runtime OVN-K is used.
Cos otherwise, we will just keep piling on ovnKubernetesConfig and it won't be sustainable.

enhancements/network/netflow.md
rcarrillocruz: done

enhancements/network/netflow.md
rcarrillocruz: done

enhancements/network/netflow.md
```diff
@@ -0,0 +1,132 @@
...
+  disableMultiNetwork: false
+  disableNetworkDiagnostics: false
+  logLevel: Normal
+  managementState: Managed
+  observedConfig: null
+  operatorLogLevel: Normal
+  serviceNetwork:
+  - 172.30.0.0/16
+  exportNetworkFlows:
+    netflow:
```
danwinship, line 90: `netFlow`, `sFlow`, `ipfix`. Sticking to the standard rules makes it easier for people to remember the right capitalization to use when creating/editing the struct

enhancements/network/netflow.md
```diff
@@ -0,0 +1,132 @@
...
+    - 172.30.158.150:2056
+    - 172.30.54.103:2056
+    sflow:
+    - 172.30.158.150:6343
+    - 172.30.54.103:6343
```
danwinship, line 95: Above you describe `sflow` as an "optional dict", but here it's an array, not a dict, and in particular, there's nowhere under `sflow` where you could add sampling/polling config. You'd need something like

    sFlow:
      endpoints:
        - 172.30.158.150:2056
        - 172.30.54.103:2056

so you can add other fields later too. (I don't know if "endpoint" is the right terminology here.)

But then, would you necessarily want the same sampling config for every endpoint?

danwinship (approved):
I was going to just merge this and say we can address the other issues in a later follow-up, but then I remembered that I'm not an approver in enhancements anyway

danwinship:
oh, Russell approved it.
/lgtm

enhancements/network/netflow.md
rcarrillocruz: yeah, the description needs fixing, it should be an array just like in the example.

enhancements/network/netflow.md
rcarrillocruz: sgtm

PR #585, opened 2021-01-14
EUS Upgrades MVP

sdodson:
An enhancement that focuses on general upgrade safety, speed, and
a minimal amount of EUS to EUS specific work.

TODO: 
- [ ] Find a board / team to take on the disconnected upgrade planning tools

### DIFF
--- a/enhancements/update/eus-upgrades-mvp.md
+++ b/enhancements/update/eus-upgrades-mvp.md
@@ -1,0 +1,400 @@
+---
+title: eus-upgrades-mvp
+authors:
+  - "@sdodson"
+reviewers:
+  - @darkmuggle
+  - @soltysh
+  - @sttts
+  - @deads2k
+  - @ecordell
+  - @dgoodwin
+approvers:
+  - "@eparis"
+  - "@derekwaynecarr"
+  - "@crawford"
+  - "@pweil-"
+creation-date: 2020-01-12
+last-updated: 2020-01-12
+status: provisional
+see-also:
+replaces:
+superseded-by:
+---
+
+# EUS Upgrades MVP
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement outlines a set of cross platform improvements meant to ensure
+the safety of multiple back-to-back minor version upgrades associated with EUS
+to EUS upgrades. Additionally, we outline improvements applicable to all
+upgrades meant to reduce the duration of upgrades and workload disruption.
+
+All of the work detailed herein is prerequisite to enabling upgrades which skip
+reboots. However where non-overlapping resources exist we can pursue reboot
+removal in parallel such as the Node team validating upstream's stated Kubelet
+version skew policies.
+
+## Motivation
+
+The introduction of EUS creates a subset of clusters which we expect will run
+4.6 for a year or more then upgrade rapidly, though serially, from 4.6 to 4.10.
+This rapid upgrade introduces the risk that those clusters may upgrade faster
+than is safe due to constraints imposed by OpenShift, the upstream components of
+OpenShift, or deployed workloads.
+
+This also creates a scenario where admins wish to reduce both the duration and
+the disruption to workload associated with the upgrade.
+
+### Goals
+
+- We will inform admins of incompatibilities between their current usage and the
+final upgrade target
+- We will provide disconnected admins tools that help them plan their entire
+upgrade path from 4.6 to 4.10 so that they may mirror all content necessary
+- We will provide guidance on expected upgrade duration scaled by node count exclusive
+of the variability associated with workload rescheduling (ie: everything except
+Worker MachineConfigPool rollout)
+- We will inhibit upgrades to the next minor version whenever necessary to preserve
+cluster stability and supportability, ie:
+  - Respecting version skew policies between OS managed components (kubelet, crio, RHCOS)
+  and operator managed components (Kubernetes API, OpenShift API, etc)
+  - Respecting version skew policies of OLM managed Operators
+  - Ensuring APIs removed by components in the release payload are no longer in
+  use
+- We will ensure that CVO managed operators upgrade as quickly as is safe to do so
+- We will reduce workload disruption associated with pod rescheduling during rolling
+reboots
+
+
+### Non-Goals
+
+- This enhancement does not attempt to remove *any* steps along the serial
+  upgrade path from 4.6 to 4.7 to 4.8 to 4.9 to 4.10 *including* reboots. A
+  follow-up enhancement will address those concerns.
+
+## Proposal
+
+Before we dive into user stories it's useful that you're familiar with CVO and
+Console features which are relevant to how upgrades and channel options are
+presented to the admin. The CVO and Console as of 4.6 are now aware of other
+channels containing the cluster's current version. This means that if 4.6.50 is
+a non EUS version and 4.6.51 is an EUS version both the CVO and Console can
+display a different set of channel options. This means that 4.6.50 would see
+candidate-4.6, fast-4.6, stable-4.6, eus-4.6, candidate-4.7, fast-4.7, and
+stable-4.7 channels.
+
+Where as 4.6.51 would see only eus-4.6 and eus-4.10 assuming we have enabled EUS
+4.6 to 4.10 upgrades. After switching to eus-4.10 the cluster would offer
+upgrades to 4.7 versions present in the eus-4.10 channel, when that completes
+the cluster would offer 4.8 versions present in the eus-4.10 channel, repeat
+until the cluster has completed its upgrade to 4.10.
+
+### User Stories
+
+#### APIServer - Notify cluster admins of API use which will be removed between 4.7 and 4.10
+
+We have pre-existing mechanisms to alert admins to their use of APIs which are
+slated for removal in a future release. Currently those mechanisms do not look
+forward far enough to span 4.6 to 4.10, we will need to add the ability to
+backport that API metadata to 4.6 once we have a firm understanding of all
+removals between 4.6 and 4.10.
+
+In addition to Info level alerts core API providers will also need to set
+Upgradeable=False whenever the next minor version removes an API which is in
+use.
+
+This allows us to *inform* the admin for removals that are more than one minor
+version away and *block* upgrades for removals which are imminent.
+
+### MCO - Enforce OpenShift's defined host component version skew policies
+
+The MCO, will set Upgradeable=False whenever any MachineConfigPool has one more
+more nodes present which fall outside of a defined list of constraints. For
+instance, if OpenShift has a defined Kubelet Version Skew of N-1, the node
+constraints enforced by the MCO defined in OCP 4.7 could be as follows:
+
+```yaml
+node.status.nodeInfo.kubeletVersion:
+- v1.20
+- v1.19
+```
+
+If the policy were to change to allow for a version skew of N-2, v1.18 would be
+added to the list of acceptable matches. The MCO is not responsible for defining
+these constrains and constraints are only widened whenever we have CI testing
+proves them to be safe.
+
+These changes will need to be backported to 4.7 prior to 4.7 EOL.
+
+#### OLM - Allow Operators to define inclusive compatible version ranges
+
+We should allow maintainers to define an inclusive version range for their Operator.
+
+#### OLM - Enforce defined OLM managed Operator compatibility
+
+If a managed Operator defines a maxKubeVersion or maxOCPVersion which would be
+violated by upgrading to the next minor OLM must set Upgradeable=False and
+enumerate the constraint in the condition message.
+
+Note, this is up for debate as to what the behavior should be in absence of a
+defined maximum version.
+
+#### OTA - Inhibit minor version upgrades when an upgrade is in progress
+
+We should inhibit minor version upgrades via Upgradeable=False whenever an existing
+upgrade is in progress. This prevents retargetting of upgrades before we've reached
+a safe point.
+
+Imagine:
+
+1. Be running 4.6.z.
+1. Request an update to 4.7.z'.
+1. CVO begins updating to 4.7.z'.
+1. CVO requests recommended updates from 4.7.z', and hears about 4.8.z".
+1. User accepts recommended update to 4.8.z" before the 4.7.z' OLM operator had come out to check its children's max versions against 4.8 and set Upgradeable=False.
+1. Cluster core hits 4.8.z" and some OLM operators fail on compat violations.
+
+This should not inhibit further z-stream upgrades, but we should be sure that
+we catch the case of 4.6.z to 4.7.z to 4.7.z+n to 4.8.z whenever 4.7.z was not
+marked as Complete.
+
+#### OTA - Optimize & Model Minor Version Upgrade Duration
+
+With support of the PerfScale team the Over The Air team will measure and
+analyze upgrade timings using simulated workloads at a defined capacity and
+load. This analysis will be used to identify optimizations and set expectations
+on a per minor version basis as to the duration of the upgrade.
+
+Due to the variability in workload deployment these measurements will be
+*exclusive of Worker MCP rollout*. The simulated workload is intended only to
+ensure that our measurements more closely mimic real world usage, not upgrades
+of clusters with zero workload.
+
+##### SDN - Canary and Optimize SDN DaemonSet rollout
+
+In 4.6 and 4.7 we identified a number of DaemonSets including dns, mcd, node-tuned,
+node-exporter which were not critical to workload availability which could rollout
+in a more parallel manner. Those operators were changed to utilize maxUnavailable
+of 10% rather than absolute value of 1, on a cluster of 250 nodes that generally
+reduced rollout time from 80 minutes to 10 minutes. The multus, sdn, ovs DaemonSets
+have not been updated but likely could be now that workload critical processes
+have been moved to the host. Before doing this given the risk we should devise
+a method to canary updates to these DaemonSets. Any other options to speed up
+updates to these DaemonSets which are deployed to all Linux nodes should also be
+considered.
+
+Once these new deployment patterns have been vetted in 4.8 or 4.9 these changes
+should be backported through to 4.6.
+
+#### Workloads - Optimize scheduling for rolling upgrades
+
+Currently the MCO cycles nodes in a deterministic but unintentional manner. The
+workload on that host which can be relocated scatters to other hosts according
+to scheduling rules. This has been measured and shown that approximately 70% of
+all reschedulable pods are rescheduled more than necessary if they were
+rescheduled in an optimal manner. Optimal rescheduling should be closer to (1 /
+node_count_per_failure_domain) percentage of pods being rescheduled more than
+once during a rolling restart of a given failure domain.
+
+We should make the MCO topology aware so that it upgrades all of the host in one
+failure domain because anti-affinity scheduling is used in most scenarios. We
+should provide some mechanism to rank upgraded nodes as slightly higher priority
+in scheduler configuration. Perhaps a timestamp of configuration generation that
+the scheduler then biases toward nodes with the greatest timestamps as a last in
+chain priority.
+
+We should consider enabling the descheduler as well to ensure that when the
+upgrade completes we rebalance so that the last node ends up with proportionate
+workload.
+
+#### ??? - Provide a hosted graph viewer which will assist disconnected admins in mirroring all content between 4.6 and 4.10
+
+Assume that 4.6.eus-z -> 4.7.z -> 4.8.z -> 4.9.z -> 4.10.z are all in the eus-4.10
+channel providing a chain of upgrades that eventually lead to 4.10. We should build
+a graph viewer / solver somewhere which allows an admin of a 4.6 cluster hoping
+to upgrade to 4.10 to discover all content which must be mirrored between their
+current version and their eventual 4.10 target.
+
+In general the graph viewer should provide a complete path from the cluster's
+current version to the latest .z in the selected channel assuming such a path exists.
+If someone were to provide cluster version 4.4.3 but select the stable-4.6 channel
+no such path exists and it should indicate that, if they were to select stable-4.5
+it would provide the upgrade path of 4.4.3 to 4.4.29 to 4.5.24 which is the latest 4.5.z.
+
+### Implementation Details/Notes/Constraints [optional]
+
+The main caveat of what's outlined here is that it does not reduce reboots
+which, for many reasons, is highly desired by some admins. The rationale for not
+addressing that in this enhancement is that this enhancement targets
+improvements which benefit all upgrades and are requisite in both scenarios.
+
+The epic / user story to backport API removal notifications to 4.6 assumes that
+we will require anyone upgrading along an EUS specific upgrade path to update to
+a 4.6 version which is current as of the time that 4.9 goes GA. This date is chosen
+because it's the point in time where we must know with certainty every API which
+is removed in 4.10. EUS customers may assume that EUS means "I only need to upgrade
+every 18 months" which will be at odds with our expectation that they upgrade
+into a late 4.6.z before starting their upgrade to 4.10.
+
+### Risks and Mitigations
+
+The improvements here do not seem to introduce risk unto themselves. However
+failure to implement some of these such as the version skew constraints would
+impose significant risk to supportability in clusters which may upgrade more
+rapidly than is safe to do so.
+
+Additionally, the timeframe for final delivery extends three releases into the
+future. This allows for design iteration during the 4.8 release cycle with
+implementation focused on the 4.9 and 4.10 release cycles.
+
+Upstream generally pays little attention to concerns which are present in EUS
+clusters where they stall for a year or more and then expect to rapidly upgrade
+across 3 or more minor versions to the next EUS version. Therefore decisions are
+made upstream which may hinder our ability to deliver this pattern to EUS
+customers.
+
+## Design Details
+
+### Open Questions [optional]
+
+- As this is a bit of a meta-enhancement calling out epic level work it should be
+no surprise that there will be implementation details unaccounted for today. We
+will need to decide on an epic by epic level which require enhancements of their
+own.
+
+- While I believe all of this work to be prerequisite to taking on work to skip
+upgrade steps that may not be widely agreed upon.
+
+- Do we need to consider CCO for similar treatment that we're providing for API
+compatibility? This seems like a nice to have that we could consider for followup.
+If we remova an API workloads break, if credential requests are not capable of
+being met upgrades to a subsequent release will not be possible but the workload
+should remain unaffected.
+
+- Should the graph solver attempt to assist with mirroring of catalog content or
+only core platform content?
+
+- Should we externalize the API removal bits? Perhaps via a marketplace operator
+that's broadly scoped as "EUS 4.6 to EUS 4.10 Validator"?
+
+### Test Plan
+
+- CI tests are necessary which attempt to upgrade while violating kubelet to API
+compatibility, ie: 4.6 to 4.7 upgrade with MachineConfigPools paused, then check
+for Upgradeable=False condition to be set by MCO assuming that our rules only allow
+for N-1 skew.
+- CI tests are necessary which install an OLM Operator which expresses a maxKubeVersion
+or maxOCPVersion equal to the current cluster version and checks for Upgradeable=False
+on OLM
+- CI tests are necessary which check for Info level alerts when to be removed
+APIs are in use.
+- Once we've determined the target upgrade duration we should tighten the existing
+`Upgrades should be fast` test to those values so that we catch regressions on
+our upgrade duration targets. We should also add infrequent test jobs which
+perfom the same upgrade duration tests at cluster sizes and workloads that more
+closely represent our deployed fleet.
+
+### Graduation Criteria
+
+**Note:** *Section not required until targeted at a release.*
+
+Define graduation milestones.
+
+These may be defined in terms of API maturity, or as something else. Initial proposal
+should keep this high-level with a focus on what signals will be looked at to
+determine graduation.
+
+Consider the following in developing the graduation criteria for this
+enhancement:
+
+- Maturity levels
+  - [`alpha`, `beta`, `stable` in upstream Kubernetes][maturity-levels]
+  - `Dev Preview`, `Tech Preview`, `GA` in OpenShift
+- [Deprecation policy][deprecation-policy]
+
+Clearly define what graduation means by either linking to the [API doc definition](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning),
+or by redefining what graduation means.
+
+In general, we try to use the same stages (alpha, beta, GA), regardless how the functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+#### Examples
+
+These are generalized examples to consider, in addition to the aforementioned [maturity levels][maturity-levels].
+
+##### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+- Enumerate service level indicators (SLIs), expose SLIs as metrics
+- Write symptoms-based alerts for the component(s)
+
+##### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+
+**For non-optional features moving to GA, the graduation criteria must include
+end to end tests.**
+
+##### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
+
+### Upgrade / Downgrade Strategy
+
+Despite the topic area, this work does not actually change Upgrade or Downgrade
+strategy.
+
+### Version Skew Strategy
+
+N/A
+
+## Implementation History
+
+See the noted CVO and Console features related to channel presentation outlined
+above. Otherwise nothing specifically tied to this effort.
+
+## Drawbacks
+
+The idea is to find the best form of an argument why this enhancement should _not_ be implemented.
+
+## Alternatives
+
+Rather than having MCO enforce version skew policies between OS managed
+components and operator managed components it could simply set Upgradeable=False
+whenever a rollout is in progress. This would preclude minor version upgrades in
+situations where a z-stream rollout stalls for some reason or another. We also
+know of some customers who have maintenance windows such that they may be in a
+perpetual state of MCP rollout.
+
+We could also consider simply waiting for all MachineConfigPool rollouts to have
+completed before the MCO considers itself upgraded. This has the downside of
+shifting highly variable workload dependent operation into the scope of the
+upgrade which is something we track very closely both in terms of success and
+duration.
+
+
+## Infrastructure Needed [optional]
+
+N/A
--- a/enhancements/update/eus-upgrades-mvp.md
+++ b/enhancements/update/eus-upgrades-mvp.md
@@ -207,12 +207,13 @@ rescheduled in an optimal manner. Optimal rescheduling should be closer to (1 /
 node_count_per_failure_domain) percentage of pods being rescheduled more than
 once during a rolling restart of a given failure domain.
 
-We should make the MCO topology aware so that it upgrades all of the host in one
-failure domain because anti-affinity scheduling is used in most scenarios. We
-should provide some mechanism to rank upgraded nodes as slightly higher priority
-in scheduler configuration. Perhaps a timestamp of configuration generation that
-the scheduler then biases toward nodes with the greatest timestamps as a last in
-chain priority.
+We should study this more deeply to understand what improvements could be made
+to reduce pod rescheduling during the rolling reboot of a cluster associated with
+upgrades. This may require coordination between MCO and the scheduler such that
+the scheduler can bias slightly towards hosts which were more recently updated.
+
+We should consider whether or not we're only concerned with rolling reboots due
+to machineConfig updates or if we're attempting to optimize all roling reboots.
 
 We should consider enabling the descheduler as well to ensure that when the
 upgrade completes we rebalance so that the last node ends up with proportionate


enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,296 @@
...
+on a per minor version basis as to the duration of the upgrade.
+
+Due to the variability in workload deployment these measurements will be
+*exclusive of Worker MCP rollout*. The simulated workload is intended only to
+ensure that our measurements more closely mimic real world usage, not upgrades
+of clusters with zero workload.
+
+#### As an admin, upgrades of CVO managed components are as fast as safe
+
+In 4.6 and 4.7 we identified a number of DaemonSets which were not critical to
```
pweil-, line 136: List these out?

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,296 @@
...
+The main caveat of what's outlined here is that it does not reduce reboots
+which, for many reasons, is highly desired by some admins. The rationale for not
+addressing that in this enhancement is that this enhancement targets
+improvements which benefit all upgrades.
+
+The epic / user story to backport API removal notifications to 4.6 assumes that
+we will require anyone upgrading along an EUS specific upgrade path to update to
+a 4.6 version which is specific to EUS. This is necessary because only those who
+follow the EUS upgrade path have 4.10 as their next logical target version. We
+should not alert non EUS clusters of removals in 4.10 until they've upgraded to
```
pweil-, line 182: Maybe this is the intention driven experience I was asking about above.  It isn't that they've picked 4.10 (they still have to pick 4.7 as the next target) it's that they've upgraded to this 4.6 EUS version that has all the notification items backported to it?

In that case would there be an EUS 4.7, 4.8, and 4.9 or would the backported notifications show for people not on the EUS path too? (or the alternative above that only shows {next releases notifications} with expansion to {next EUS notifications}

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,296 @@
...
+  upgrade path from 4.6 to 4.7 to 4.8 to 4.9 to 4.10 *including* reboots. A
+  follow-up enhancement will address those concerns.
+
+## Proposal
+
+### User Stories
+
+#### As an admin preparing for 4.6 to 4.10 EUS upgrade I'm alerted to API removals
+
+We have pre-existing mechanisms to alert admins to their use of APIs which are
```
pweil-, line 87: I'm assuming this is the notification and alert section.

For UX related teams: I think we need to clarify the experience on if this would be incorporated to part of the upgrade visualization (in addition to alerting) and if so how would that API work?

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,296 @@
...
+no surprise that there will be implementation details unaccounted for today. We
+will need to decide on an epic by epic level which require enhancements of their
+own.
+
+While I believe all of this work to be prerequisite to taking on work to skip
+upgrade steps that may not be widely agreed upon.
+
+### Test Plan
+
+Each epic should prescribe their own test plans.
```
pweil-, line 210: Should we list out the organizational finish line for CI on this effort?

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,296 @@
...
+
+## Proposal
+
+### User Stories
+
+#### As an admin preparing for 4.6 to 4.10 EUS upgrade I'm alerted to API removals
+
+We have pre-existing mechanisms to alert admins to their use of APIs which are
+slated for removal in a future release. Currently those mechanisms do not look
+forward far enough to span 4.6 to 4.10, we will need to add the ability to
```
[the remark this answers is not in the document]
sdodson: Yes, if we want alerting to be different based on EUS vs not we need to have some mechanism to know that they're EUS. Right now I'd planned on that being a requirement that they upgrade into a 4.6 version that falls after the EUS transition and we'd only backport these alerting rules to those versions.

Requiring them to upgrade deep into 4.6.z may be a tough sell but I think if we move on to eventually removing reboots it's a decent exchange because it allows us to assume a higher baseline of maturity because they've got at least all of the quality improvements that happened during the first year of 4.6. This requirement also allows us to limit the upgrade graph and thus the versions we must test forcing them deep into 4.7, 4.8, and 4.9. Only 4.10 will be relatively new at that point and they'd also have the option of waiting another 3-6 months before completing the 4.9 and 4.10 upgrades if they wish.

enhancements/update/eus-upgrades-mvp.md
sdodson: The backporting to 4.7, 4.8, and 4.9 is a good point to raise. I hadn't thought about that, I would assume we leave alerting rules in those versions whatever is appropriate for non-EUS upgrades. We know that 4.6 EUS upgrades will be upgrading through those versions but if we raise the red flags in 4.6 EUS versions and they start the upgrade anyway I feel like they were given enough warning.

I'm also envisioning an escalation of alerts based on when the API is removed. APIs removed in N+1 are warning, N+2 or greater are info level only. When running an EUS specific version N+1 should be equivalent to the alert set of your eventual target minus one, ie: 4.9 in the case that 4.6 and 4.10 are the two EUS versions.

enhancements/update/eus-upgrades-mvp.md
sdodson: It may be useful if the CVO scraped up alerts of this nature and raised them and then the console could only display alerts firing from CVO on the upgrades panel.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,333 @@
...
+
+- As this is a bit of a meta-enhancement calling out epic level work it should be
+no surprise that there will be implementation details unaccounted for today. We
+will need to decide on an epic by epic level which require enhancements of their
+own.
+
+- While I believe all of this work to be prerequisite to taking on work to skip
+upgrade steps that may not be widely agreed upon.
+
+- Do we need to consider CCO for similar treatment that we're providing for API
```
sdodson, line 231: @dgoodwin @joelddiaz Your thoughts here please

joelddiaz: As long as we visit each minor release between 4.6->4.10, I think we're okay. We just need to make sure the minor release visited is a release where CCO knows about the next minor release requirements. So for example: you can't go from 4.6.10 (which can alert about upgradeability to 4.7) to 4.7.2 (which doesn't know about upgradeability to 4.8 b/c 4.7.2 was too early to know about what was coming in 4.8).

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,346 @@
...
+
+Where as 4.6.51 would see only eus-4.6 and eus-4.10 assuming we have enabled EUS
+4.6 to 4.10 upgrades. After switching to eus-4.10 the cluster would offer
+upgrades to 4.7 versions present in the eus-4.10 channel, when that completes
+the cluster would offer 4.8 versions present in the eus-4.10 channel, repeat
+until the cluster has completed its upgrade to 4.10.
+
+### User Stories
+
+#### As an admin preparing for 4.6 to 4.10 EUS upgrade I'm alerted to API removals
```
sdodson, line 103: @deads2k @sttts PTAL at this section as well as the enhancement in general.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,346 @@
...
+
+We have pre-existing mechanisms to alert admins to their use of APIs which are
+slated for removal in a future release. Currently those mechanisms do not look
+forward far enough to span 4.6 to 4.10, we will need to add the ability to
+backport that API metadata to 4.6 EUS versions once we have a firm understanding
+of all removals between 4.6 and 4.10.
+
+#### As an admin I cannot violate OpenShift's defined Kubelet to API version skew
```
sdodson, line 111: @darkmuggle PTAL at this proposed epic as well as the enhancement in full

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,346 @@
...
+nodes had upgraded to 4.7.
+
+Note, OpenShift may choose to define different version skew constraints than
+upstream Kubernetes. We may also change these in the future when our internal
+testing indicates it is safe to do so, therefore this should be implemented in a
+manner which allows us to change the constraints in a future update to the MCO.
+
+These changes will need to be backported to 4.7 prior to 4.10 GA.
+
+#### As the maintainer of a Market Place Operator I may define either maxKubeVersion or maxOCPVersion
```
sdodson, line 127: @ecordell PTAL at this proposed epic as well as the enhancement in full

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,346 @@
...
+
+We should allow maintainers to define an inclusive version range for their Operator.
+
+#### As an admin I cannot violate defined OLM managed Operator compatibility
```
sdodson, line 131: @ecordell PTAL at this proposed epic as well as the enhancement in full

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,346 @@
...
+
+If a managed Operator defines a maxKubeVersion or maxOCPVersion which would be
+violated by upgrading to the next minor OLM must set Upgradeable=False and
+enumerate the constraint in the condition message.
+
+Note, this is up for debate as to what the behavior should be in absence of a
+defined maximum version.
+
+#### As an admin I have estimated upgrade duration data scaled by node count
```
sdodson, line 140: @LalatenduMohanty PTAL at this proposed epic as well as the enhancement in full

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,352 @@
...
+have not been updated but likely could be now that workload critical processes
+have been moved to the host. Before doing this given the risk we should devise
+a method to canary updates to these DaemonSets. Any other options to speed up
+updates to these DaemonSets which are deployed to all Linux nodes should also be
+considered.
+
+Once these new deployment patterns have been vetted in 4.8 or 4.9 these changes
+should be backported through to 4.6.
+
+#### As an admin, my workloads within a failure domain are rescheduled a minimum number of times
```
sdodson, line 168: @soltysh @darkmuggle PTAL

sdodson:
@darkmuggle @soltysh @sttts  @deads2k @ecordell @dgoodwin @crawford @eparis @derekwaynecarr PTAL

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,366 @@
...
+of all removals between 4.6 and 4.10.
+
+#### As an admin I cannot violate OpenShift's defined Kubelet to API version skew
+
+The MCO, or another suitable component, must set Upgradeable=False in order to
+prevent upgrades whenever a managed node would violate Kubelet to API version
+skew defined by OpenShift. For instance, assume OCP 4.6, 4.7, and 4.8 map to
+Kubernetes versions 1.19, 1.20, 1.21. If OpenShift chooses to support only
+version skew of N-1 the MCO would inhibit upgrades to 4.8 until all managed
+nodes had upgraded to 4.7.
```
darkmuggle, line 118: I *think* what is being requested here is that we finish the "Wait for all pools" work where we block upgrades until everything has been upgraded; we're planning on that work finishing in 4.8.

I am leary about the MCO gaining any awareness of versions or brokering if a specific version is compatible with the running Kubelet versions. A better path would for all components to report their API versions and the CVO blocks upgrades until the components show the compatible state.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,366 @@
...
+
+Note, OpenShift may choose to define different version skew constraints than
+upstream Kubernetes. We may also change these in the future when our internal
+testing indicates it is safe to do so, therefore this should be implemented in a
+manner which allows us to change the constraints in a future update to the MCO.
+
+These changes will need to be backported to 4.7 prior to 4.10 GA.
```
darkmuggle, line 125: This is a tall-ask that makes me _very_ nervous; I will not commit to backporting unknowns to the MCO. We need to get very specific about the technical details and architecture. 

Why would this knowledge be encoded in the MCO instead of the CVO?

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,366 @@
...
+
+Currently the MCO cycles nodes in a deterministic but unintentional manner. The
+workload on that host which can be relocated scatters to other hosts according
+to scheduling rules. This has been measured and shown that approximately 70% of
+all reschedulable pods are rescheduled more than necessary if they were
+rescheduled in an optimal manner. Optimal rescheduling should be closer to (1 /
+node_count_per_failure_domain) percentage of pods being rescheduled more than
+once during a rolling restart of a given failure domain.
+
+We should make the MCO topology aware so that it upgrades all of the host in one
```
darkmuggle, line 178: The MCO should not be topology-aware, IMHO. Topology is very much platform-specific. Really, what is being asked is that the MCO would have to infer sub-pools based on cloud regions. The understanding of cloud-fault domains is well above the level of the MCO. The MCO would need a platform-agnostic means of enumerating the fault-domain and upgrade priorities.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,366 @@
...
+
+In general the graph viewer should provide a complete path from the cluster's
+current version to the latest .z in the selected channel assuming such a path exists.
+If someone were to provide cluster version 4.4.3 but select the stable-4.6 channel
+no such path exists and it should indicate that, if they were to select stable-4.5
+it would provide the upgrade path of 4.4.3 to 4.4.29 to 4.5.24 which is the latest 4.5.z.
+
+### Implementation Details/Notes/Constraints [optional]
+
+The main caveat of what's outlined here is that it does not reduce reboots
```
darkmuggle, line 205: I think we need to more precise about the desires to eliminate reboots. In various conversations I have had, we seem to be conflating workload interruption with reboots. If we are talking about workload interrupts then the solutions that we come up might look _very_ different. For example, live-migration for workload interruption over reboot which might be a _capacity_ problem.

enhancements/update/eus-upgrades-mvp.md
sdodson: I think the MCO is the right Operator but we can be flexible on implementation for how MCO knows when to inhibit minor version upgrades. No other operator manages the OS and thus manages the Kubelet.

enhancements/update/eus-upgrades-mvp.md
sdodson: We'd always be respecting minimum version requirements, we'll be upgrading along very late 4.6.z, 4.7.z, and 4.8.z. While 4.9.z will be relatively young it must meet the minimum version requirements to upgrade to 4.10.

enhancements/update/eus-upgrades-mvp.md
sdodson: I've updated this proposed epic to reflect what we discussed regarding constraints being defined by policy and enforced by MCO.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,368 @@
...
+compatibility? This seems like a nice to have that we could consider for followup.
+If we remova an API workloads break, if credential requests are not capable of
+being met upgrades to a subsequent release will not be possible but the workload
+should remain unaffected.
+
+- Should the graph solver attempt to assist with mirroring of catalog content or
+only core platform content?
+
+- Should we externalize the API removal bits? Perhaps via a marketplace operator
+that's broadly scoped as "EUS 4.6 to EUS 4.10 Validator"?
```
crawford, line 257: If we've written our operators correctly, an EUS Validator would be entirely redundant. All operators must be able to reason about (and take action against!) the validity of their operands in the context of the larger system. This is what makes them an operator as opposed to a controller.

enhancements/update/eus-upgrades-mvp.md
mfojtik: Do we need to consider using upgradeable=false if we detect usage of an API that is removed in next version?

iow.:

4.6
4.7
4.8: API server go "upgradeable=false" if there are cronjobs in the cluster
4.9: CronJobs are gone, but cluster has still 1000 cron jobs created
4.10:

sttts: Info alerts are planned for 4.7, showing up in the console notification side-bar. We will have to backport to 4.6.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,368 @@
...
+we will require anyone upgrading along an EUS specific upgrade path to update to
+a 4.6 version which is specific to EUS. This is necessary because only those who
+follow the EUS upgrade path have 4.10 as their next logical target version. We
+should not alert non EUS clusters of removals in 4.10 until they've upgraded to
+a later minor version.
+
+As with all upgrades, new features enabled by interim versions are not expected
+enabled during the upgrade path. This means that for instance new MachineConfig
+features enabling complex root devices while introduced in 4.7 may not be utilized
+until the cluster has completed its upgrade.
```
sttts, line 222: This won't be possible, at least on the upstream side. If the admin is quick enough to use new APIs, he will succeed. Clarify what you mean with "features" here.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,368 @@
...
+
+The improvements here do not seem to introduce risk unto themselves. However
+failure to implement some of these such as the version skew constraints would
+impose significant risk to supportability in clusters which may upgrade more
+rapidly than is safe to do so.
+
+Additionally, the timeframe for final delivery extends three releases into the
+future. This allows for design iteration during the 4.8 release cycle with
+implementation focused on the 4.9 and 4.10 release cycles.
+
```
sttts, line 234: I miss the risk of upstream making our efforts hard or infeasible.

enhancements/update/eus-upgrades-mvp.md
pweil-: @sdodson interesting idea from @sferich888 today: if the APIs are deprecated in what we generate our API docs from can we use that info to also generate big, scary release notes?

Another suggestion for collaboration: @jwmatthews and the container migration teams may have thoughts on how we can help admins work through deprecations and other ways to make sure they're front and center during our upgrade process.

jwmatthews: @pweil- FYI, we have some beginning validation work in context of Migrations from MTC to warn a user when they will face a problem that a GVK they relied on in the source cluster isn't present in the target cluster (https://github.com/konveyor/mig-controller/blob/master/pkg/gvk/gvk.go#L66)  This is a problem space we are interested in, we expect to see it with future migrations, open to collaborating.    (cc @djwhatle @eriknelson @shawn-hurley @jortel )

soltysh: k8s also has https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1693-warnings which kubectl & oc consume informing users that they are using deprecating APIs. The question is if/how we could also consume that elsewhere.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,368 @@
...
+once during a rolling restart of a given failure domain.
+
+We should make the MCO topology aware so that it upgrades all of the host in one
+failure domain because anti-affinity scheduling is used in most scenarios. We
+should provide some mechanism to rank upgraded nodes as slightly higher priority
+in scheduler configuration. Perhaps a timestamp of configuration generation that
+the scheduler then biases toward nodes with the greatest timestamps as a last in
+chain priority.
+
+We should consider enabling the rescheduler as well to ensure that when the
```
soltysh, line 187: s/rescheduler/descheduler

enhancements/update/eus-upgrades-mvp.md
soltysh: I think the idea here is to inform about user workloads wrt deprecated/removed APIs between 4.6 and 4.10, not with our own.

enhancements/update/eus-upgrades-mvp.md
sdodson: @pweil- @sferich888  Yeah I definitely agree that we'll want to have a release notes if not an upgrade section of docs that's unique to 4.6 to 4.10 upgrades. I'll add an epic place holder to ensure that as part of 4.10 dev cycle we pull this together.

I'll find a slot next week to sync up with @sanchezl @deads2k and you, @jwmatthews.

enhancements/update/eus-upgrades-mvp.md
sdodson: Correct, OLM is just the delivery mechanism to install an optional component which could decouple the EUS specific API removal notification from the release payload. I don't think we want to attempt to remediate any issues identified but just inform the admin.

enhancements/update/eus-upgrades-mvp.md
sdodson: My intent was to ensure that if we introduce a new capability in 4.9 which depends on updated host bound components (kernel, crio, kubelet, etc) and we forego updating RHCOS to 4.9 then that new capability is unavailable until host components are updated to 4.10.

enhancements/update/eus-upgrades-mvp.md
soltysh: Cards are already on our board.

enhancements/update/eus-upgrades-mvp.md
sdodson: Updated text mentions simply that MCO will enforce defined policy. Going to mark this thread as resolved, it would need to be backported to 4.8 if we determine kubelet n-2 version skew is acceptable, if N-1 is all we can tolerate we'd need to get it back to 4.7 but it will be quite some time before we determine that.

enhancements/update/eus-upgrades-mvp.md
sdodson: @sttts specifically around items outlined here, mainly api removals, or something else?

I do intend to call out upstream's apparent lack of kubelet N-2 version skew testing despite the stated policy that it's supported.

soltysh: > @sttts specifically around items outlined here, mainly api removals, or something else?

afaict, all of them. We only have this amount of time we can juggle between upstream and downstream. It happened in the past that we missed changes upstream that caused us pain. They are rare, but they still can happen without our sufficient upstream presence to mitigate them.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,368 @@
...
+
+### Goals
+
+- We will inform admins of incompatibilities between their current usage and the
+final upgrade target
+- We will provide disconnected admins tools that help them plan their entire
+upgrade path from 4.6 to 4.10 so that they may mirror all content necessary
+- We will provide guidance on expected upgrade duration scaled by node count exclusive
+of the variability associated with workload rescheduling (ie: everything except
+Worker MCP rollout)
```
wking, line 67: nit: expand MCP to MachineConfigPool?  I was scratching my head for a minute until that clicked.

enhancements/update/eus-upgrades-mvp.md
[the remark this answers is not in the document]
wking: > Requiring them to upgrade deep into 4.6.z may be a tough sell...

For folks running from 4.6.{shortly-after-EUS-but-before-some-forward-looking-warning}, there's really not all that much difference between the first hop taking them to 4.6.{huge} or 4.7.{huge} and picking up the warning then, if we decide that 4.6.* -> 4.7.{huge} seems safe enough to add an update edge.  So this just comes down to our usual "how many update edges are we willing to stretch tests over?", which is a new-release-build-time decision, and doesn't need any enhancement-time decisions.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,368 @@
...
+We have pre-existing mechanisms to alert admins to their use of APIs which are
+slated for removal in a future release. Currently those mechanisms do not look
+forward far enough to span 4.6 to 4.10, we will need to add the ability to
+backport that API metadata to 4.6 EUS versions once we have a firm understanding
+of all removals between 4.6 and 4.10.
+
+### MCO - Enforce OpenShift's defined host component version skew policies
+
+The MCO, will set Upgradeable=False whenever any MachineConfigPool has one more
+more nodes present which falloutside of a define list of constraints. For
```
wking, line 114: nit: "falloutside of a define list" -> "fall outside of a defined list"

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,368 @@
...
+
+#### OLM - Allow Operators to define inclusive compatible version ranges
+
+We should allow maintainers to define an inclusive version range for their Operator.
+
+#### OLM - Enforce defined OLM managed Operator compatibility
+
+If a managed Operator defines a maxKubeVersion or maxOCPVersion which would be
+violated by upgrading to the next minor OLM must set Upgradeable=False and
+enumerate the constraint in the condition message.
```
wking, line 137: this, and a few other guards in the enhancement, are vulnerable to update-retarget vulnerabilities that openshift/cluster-version-operator#460 has been in flight for a while trying to address.  For example, today you could:

1. Be running 4.6.z.
2. Request an update to 4.7.z'.
3. CVO begins updating to 4.7.z'.
4. CVO requests recommended updates from 4.7.z', and hears about 4.8.z".
5. User accepts requested update to 4.8.z" before the 4.7.z' OLM operator had come out to check it's children's max versions against 4.8 and set `Upgradeable=False`.
6. Cluster core hits 4.8.z" and some OLM operators fail on compat violations.

This isn't a blocker for this effort; we've wanted to address the CVO-retarget issue for a while.  But it's been slow, low-priority work, and we probably want to get it going again and drive it to a conclusion as a blocker for updating out of eus-4.6.  I dunno if we have a structured way to represent blockers like that; maybe a whiteboard keyword?

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,368 @@
...
+all reschedulable pods are rescheduled more than necessary if they were
+rescheduled in an optimal manner. Optimal rescheduling should be closer to (1 /
+node_count_per_failure_domain) percentage of pods being rescheduled more than
+once during a rolling restart of a given failure domain.
+
+We should make the MCO topology aware so that it upgrades all of the host in one
+failure domain because anti-affinity scheduling is used in most scenarios. We
+should provide some mechanism to rank upgraded nodes as slightly higher priority
+in scheduler configuration. Perhaps a timestamp of configuration generation that
+the scheduler then biases toward nodes with the greatest timestamps as a last in
```
wking, line 184: This seems achievable and super-useful.  Currently the MCO annotates nodes with current and target MachineConfig references.  It could also annotate them with a rank or timestamp or some such of the current MachineConfig, and the scheduler could use that rank when breaking ties to increase the chances that drained workloads land on the already-updated nodes.  With anti-affinity rules and all that usual stuff protecting workloads from getting overly wedged into a particular failure domain, even if that means temporarily scheduling some workload on still-to-be-updated nodes.

enhancements/update/eus-upgrades-mvp.md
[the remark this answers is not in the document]
sdodson: While we know the risk is low it's still a cluster roll and a lot of customers think "EUS" means "I only have to reboot or upgrade once every 18 months."

Based on the call with API Server team it sounds like we're going to pursue only Info level alerts whenever we know of an API removal happening in the next four minor releases. We'll also pursue Upgradeable=False whenever the removal is in the next minor, so an API removed in 4.10 would get Info alerts in 4.6, 4.7, 4.8, 4.9 and Upgradeable=False in 4.9, that way the admin *knows* while running 4.6 but isn't *blocked* from forward progress until something would actually break when they take that step.

That means it becomes safe to backport all of the logic related to these alerts to 4.6 regardless of whether or not it's an EUS cluster, therefore we can set a deadline of having this backported to 4.6 when 4.9 goes GA which will be at least 12 weeks before potential 4.10 upgrades. I believe we can reasonably assert that they must be no more than 12 weeks old at the time we start providing EUS upgrades.

enhancements/update/eus-upgrades-mvp.md
sdodson: done

enhancements/update/eus-upgrades-mvp.md
sdodson: Lets call that out as a story here and make sure that we're giving that the priority and attention necessary.

enhancements/update/eus-upgrades-mvp.md
sdodson: I'll write up a separate enhancement that focuses on removal of reboots and we can propose alternatives there knowing that it's a mix of both concern over actual reboot and workload disruption due to rescheduling.

enhancements/update/eus-upgrades-mvp.md
sdodson: I think now that we know we'll only use Info level alerts for all removals between 4.6 and 4.10 and then Upgradeable=False for removals between 4.N and 4.N+1 we no longer require this intent signal. It's safe to tell all 4.6 customers about removals between 4.6 and 4.10 and it's safe/required to inhibit upgrades from 4.6 to 4.7 when APIs removed in 4.7 are in use.

enhancements/update/eus-upgrades-mvp.md
sdodson: I'm going to remove this, it's specifically tied to any efforts to remove host level updates between releases and this enhancement makes no attempt to deliver on that. I'll add this to an enhancement that's focused on that.

enhancements/update/eus-upgrades-mvp.md
sdodson: I'm no longer certain this is necessary but I'm going to leave this in until I hear back from the workloads team that this isn't necessary.

enhancements/update/eus-upgrades-mvp.md
sdodson: I've added a paragraph about EUS to EUS not being something that upstream considers. PTAL.

enhancements/update/eus-upgrades-mvp.md
sdodson: new story here https://github.com/openshift/enhancements/pull/585/files#diff-91c1b36fc770bc12ef9147ff77360aa00fb1f79d3df986a471d75d1c4a33891cR153-R170

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,400 @@
...
+chain priority.
+
+We should consider enabling the descheduler as well to ensure that when the
+upgrade completes we rebalance so that the last node ends up with proportionate
+workload.
+
+#### ??? - Provide a hosted graph viewer which will assist disconnected admins in mirroring all content between 4.6 and 4.10
```
sdodson, line 221: I'm still unsure which board this needs to land on, trying to track that down.

enhancements/update/eus-upgrades-mvp.md
darkmuggle: I think using a timestamp, while easy enough, is likely going to result in a problem as its does not indicate whether the nodes have the preferred state. I *think* having the MCO introduce the idea of an "update-epoch" would be clear and allow the scheduler to send workloads based on nodes with the highest epoch. Something like `machineconfiguration.openshift.io/update-epoch: 20` (this would be interesting for telemetry, too).

However, if someone has paused their machine config pools, it would have the effect of preferring to run workloads on unpaused pools (since they would have a higher epoch).

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,400 @@
...
+constraints enforced by the MCO defined in OCP 4.7 could be as follows:
+
+```yaml
+node.status.nodeInfo.kubeletVersion:
+- v1.20
+- v1.19
+```
+
+If the policy were to change to allow for a version skew of N-2, v1.18 would be
+added to the list of acceptable matches. The MCO is not responsible for defining
```
yuqi-zhang, line 134: How will these constrains be defined, and how does the MCO retroactively apply them to prevent the upgrade in the first place?

It sounds like the 4.6 MCO would first have to understand what the constraints are to prevent the upgrade to 4.7, despite the restrictions defined in the 4.7 MCO?

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,400 @@
...
+In addition to Info level alerts core API providers will also need to set
+Upgradeable=False whenever the next minor version removes an API which is in
+use.
+
+This allows us to *inform* the admin for removals that are more than one minor
+version away and *block* upgrades for removals which are imminent.
+
+### MCO - Enforce OpenShift's defined host component version skew policies
+
+The MCO, will set Upgradeable=False whenever any MachineConfigPool has one more
```
kikisdeliveryservice, line 122: This feels like a bit of an overreach onto the MCO. We should be setting upgradeable=false bc an upgrade isn't finished/a pool is degraded and reporting our general status but we shouldn't have to read into kubelet versions etc... to set anything. This feels like overcomplicating the matter? Why can the CVO not make the decision about the next acceptable upgrade based on operator statuses and known upgrade graphs?

enhancements/update/eus-upgrades-mvp.md
kikisdeliveryservice: If this is not necessary would prefer to skip as this feels like a separate enhancement altogether.

enhancements/update/eus-upgrades-mvp.md
yuqi-zhang: Actually on second thought, there is no real need for the MCO to expand on its current block on pools, since a previous upgrade not finishing would block further upgrades anyways. This means we would never have a kubelet version skew > 1, making this check pointless (unless you wanted to initiate a direct update from 4.6->4.8, which as I understand, shouldn't happen regardless).

enhancements/update/eus-upgrades-mvp.md
kikisdeliveryservice: Also for this we should have the narrowest path possible for these upgrades which would make it safer and easier to support.

enhancements/update/eus-upgrades-mvp.md
sdodson: The specific implementation needs to be hashed out, but it would be expected to be defined in the release payload of the version of the MCO which is running, a configmap most likely.

enhancements/update/eus-upgrades-mvp.md
sdodson: The MCO manages the host os, the host os "manages" the kubelet, therefore MCO seems like the right component to enforce constraints of the host os.

If an MCP were paused before an upgrade is initiated what happens in your scenario where we set Upgradeable=False when a pool isn't finished or degraded?

The CVO already defines cluster version upgrade paths, but the long term goal is to introduce a decoupling of cluster version from host components while still enforcing version skew policies. This enhancement doesn't propose that, just systems to enable us to change that policy in the future. The future would look like this

1. Start with 4.6 cluster, Pause Worker MCP
2. Upgrade 4.6 to 4.7
3. Upgrade 4.7 to 4.8
4. MCO is now Upgradeable=False because constraints are violated, kubelet is 1.19 but we're only upgradeable in 1.20 and 1.21, can't upgrade to 4.9.
5. Worker MCP is unpaused
6. When all hosts in MCPs now meet the constraints Upgradeable=True
7. Pause Worker MCP again
8. Upgrade to 4.9
9. Upgrade to 4.10
10. MCO is now Upgradeable=False because constraints are violated, kubelet is 1.21 but we're looking for 1.22 or 1.23.
11. Worker MCP is unpaused

enhancements/update/eus-upgrades-mvp.md
sdodson: This is really a rollup of many different enhancements across many teams, each likely to lead to an enhancement of their own if sufficiently complex to require an ehancement. I believe we should try to minimize pod rescheduling and I'm currently asking that the workloads team drive that investigation, we'll see what they come up with. I'll change this just to mention that this *may* require coordination between MCO and Scheduler in some form.

enhancements/update/eus-upgrades-mvp.md
yuqi-zhang: Is the plan then to skip versions for the worker nodes (e.g. 4.6->4.8->4.10 directly), and that is what will be in the separate proposal you mentioned? And this is setting up for that eventuality?

Put another way, this implementation is for safety only and has nothing to do with EUS. Today you could screw up your cluster by pausing worker pool, upgrading your master from 4.4->4.7, and unpause the worker pool only to have it blow up since it can't make that jump directly. In which case, perhaps we should consider adding some sort of additional check for the wait for all pools work, where paused worker pools block upgrades too?

sdodson: @yuqi-zhang Yes, exactly, this closes that gap by defining constraints that all nodes in an MCP must meet before MCO agrees that the cluster can be upgraded to the next minor. We want to decouple that decision from whether or not the pool is up to date in order to provide more flexibility. If someone changed /etc/motd we wouldn't want that to block them from upgrading to the next minor while that rollout is in progress.

kikisdeliveryservice: I think my question is why does the MCO have to even look at kubelet versions to determine this?

sdodson: I'm open to other suggestions. We had thought of maintaining a list of machine-os-content pullspecs or however that's represented, but that leaves RHEL workers out of the picture and they are managed via MCPs even if that comes via assistance of openshift-ansible. I'm also not sure how we'd maintain that list, where as simple string matching or semantic version comparison would more accurately express the constraints.

Another item I hadn't mentioned is there may become a time when we need to express these constraints on kernel versions, specifically for RHEL Workers.

Is the concern the cluster load required to examine node objects, growth in RBAC necessary to do so, or being able to implement these rules effectively? Anything else?

crawford (approved):

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -0,0 +1,401 @@
...
+until the cluster has completed its upgrade to 4.10.
+
+### User Stories
+
+#### APIServer - Notify cluster admins of API use which will be removed between 4.7 and 4.10
+
+We have pre-existing mechanisms to alert admins to their use of APIs which are
+slated for removal in a future release. Currently those mechanisms do not look
+forward far enough to span 4.6 to 4.10, we will need to add the ability to
+backport that API metadata to 4.6 once we have a firm understanding of all
```
shawn-hurley, line 110: How does an OLM operator backport the same knowledge so that we can also alert customers to breaking API's?

sdodson: @shawn-hurley Do you mean CRDs installed and managed via OLM or OLM managed operators which use deprecated/removed core APIs? I would expect the latter to generate alerts already once this work lands, but the former would not be covered.

pweil-:
Based on current unresolved comments I believe that this PR is in a good state to merge as the umbrella proposal with subsequent iterations on any guidance this proposal needs to and the following sub-topic implementations being continued discussions for owning teams:

MCO safety checks: 
	https://github.com/openshift/enhancements/pull/585/files#r569816300
	https://github.com/openshift/enhancements/pull/585/files#r569802804
	

cluster update-retarget: https://github.com/openshift/enhancements/pull/585/files#r569160529

Hosted graph viewer: https://github.com/openshift/enhancements/pull/585/files#r569688098

Eliminate reboots (separate enhancement): https://github.com/openshift/enhancements/pull/585/files#r558461059

User workload alerting: https://github.com/openshift/enhancements/pull/585/files#r558612148

OLM Backports: https://github.com/openshift/enhancements/pull/585#discussion_r572069342

/lgtm

enhancements/update/eus-upgrades-mvp.md
yuqi-zhang: > MCO is now Upgradeable=False because constraints are violated, kubelet is 1.19 but we're only upgradeable in 1.20 and 1.21, can't upgrade to 4.9.

@sdodson Just to add a question, if we add something like this for the general case, what are we doing to determine the next upgrade will be 4.9 and not another 4.8 version (which should be allowed?). i.e. when your control plane are on 4.8, your workers are on 4.6, you should block upgrades to 4.9 but not another 4.8 version correct? Without pre-flight checks (maybe in the CVO itself) for the overall upgrade, we won't know in the MCO what the next version would be (until an upgrade gets to the MCO, and its already too late to block by then)

sdodson: @yuqi-zhang CVO only inhibits minor version upgrades when Upgradeable=False is set, it does not inhibit z-stream/patch upgrades. So an upgrade into a later 4.8.z would be allowed.

We do this so that if we screw up setting Upgadeable=False you're still allowed to upgrade into a later z-stream where we may fix that.

PR #560, opened 2020-12-10
single-node production deployment approach

dhellmann:
This enhancement describes the approach to deploying single-node
production OpenShift instances without using a cluster profile.

PR #587, opened 2021-01-14
add risk about disruptive changes to single-node-production-deployment-approach

dhellmann:
As discussed in
https://github.com/openshift/enhancements/pull/560/files#r554177184,
this change adds more detail to the risk section to highlight the fact
that configuration changes are likely to be quite disruptive to
single-node deployments.


markmc:
/approved
/lgtm

Thanks Doug

PR #590, opened 2021-01-19
authentication: add 'Allowing URI Scheme in OIDC sub claims'

stlaz:
/cc @deads2k 
/assign @sttts

### DIFF
--- a/enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
+++ b/enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
@@ -39,7 +39,8 @@ providers that use the URI scheme in the `sub` claim of their ID tokens.
 Per the [OIDC specification](https://openid.net/specs/openid-connect-core-1_0-final.html#IDToken)
 of the ID token and its more thorough details described in
 [RFC7519 - JSON Web Token (JWT)](https://tools.ietf.org/html/rfc7519#section-4.1.2),
-the `sub` claim may be represented as both a string or a URI. However, URI by
+the `sub` claim may be represented as both a string or a URI (as defined by the above RFC,
+URI is any string containing a `:`). However, URI by
 definition may contain characters such as `/` and `:` that are hard to digest
 for the `identity.oauth.openshift.io` API, which prevents integration with OIDCs
 that make use of this URI scheme. However, an identity provider that follows all
@@ -57,12 +58,24 @@ None.
 
 ## Problem Summary
 
-With OpenShift 4.x, the oauth-server no longer allows [configuring the claim
+With OpenShift 4.x, the OAuth configuration no longer allows [specifying the claim
 of ID tokens](https://github.com/openshift/api/blob/670ac3fc997c2f1d19b8c29ef04f70d6e3d4a59e/osin/v1/types.go#L341)
 retrieved from the OIDC provider to use as the unique identifier
-for the given user. This identifier is later used to construct the name of an
-`Identity` object that describes the mapping between the identity provider the
-user is a member of, and the user.
+for the given user. Allowing this configuration might end up in
+users configuring this to fields that either:
+
+- could be manually changed by the user on the OIDC provider side
+- was not unique per the configured client
+
+The OIDC specification directly specifies the `sub` claim of ID tokens
+to be the perfect candidate for use as an identifier, which is therefore
+enforced in 4.x in order to prevent the above issues.
+
+This identifier from an ID token's `sub` claim retrieved by OpenShift is used to
+construct the name of an `Identity` object that describes the mapping between
+the identity provider the user is a member of, and the user. It can also
+be used as the name of the respective `User` object if the claim for preferred
+username is not configured.
 
 The name of the `Identity` object is constructed by using the following pattern:
 
@@ -70,7 +83,7 @@ The name of the `Identity` object is constructed by using the following pattern:
 <identity provider name from config>:<unique user identifier from the identity provider>
 ```
 
-This naming is further validated so that neither the username, nor the identity provider
+This naming is validated so that neither the username, nor the identity provider
 name may contain the `:` character, and, according to Kubernetes naming schemes,
 neither can also contain the `/` sign, yet both of these commonly appear as a
 part of URI/URL schemes.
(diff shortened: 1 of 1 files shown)


matthyx:
LGTM, if you need testers, I can volunteer
/cc @baldator

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
```diff
@@ -0,0 +1,166 @@
+---
+title: neat-enhancement-idea
```
deads2k, line 2: really?

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
```diff
@@ -0,0 +1,167 @@
...
+
+Validation of the names of the `Identity` objects need to be loosened in order
+to allow three sections separated by the `:` character, where the only allowed
+value for the second section is `url`.
+
+### Risks and Mitigations
+
+When a cluster with existing `Identity` objects of the new naming schemes is downgraded,
+logging in with OIDC providers using URI scheme is their ID tokens' `sub` claims
+will be broken again.
```
s-urbaniak, line 136: tldr: this seems to be acceptable.

generally, afaik the idiom is that downgrades should generally not draw the cluster dysfunctional as downgrade scenarios are mostly used as temporary timeframes to upgrade again so this seems acceptable.

but to be sure: is it possible that those oidc id tokens appear in some form of service-to-service communication within our control plane or user workloads? I don't think so but just asking to make sure.

matthyx:
Glad to see this moving :-)

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
stlaz: There is no way to keep the authentication working for the afflicted clusters on downgrade as they would not work prior to the upgrade either. An upgrade means a switch from "this has always been broken" to fixed.

The ID tokens should only be visible to the oauth-server and none other.

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
```diff
@@ -0,0 +1,167 @@
...
+
+It is also possible that some OpenShift users already started using their OIDC
+providers that were broken due to this, with the use of `unsupportedConfigOverrides`.
+While we generally do not care about these cases, it is fair to say that the
+proposed schema should allow for a scripted conversion of `Identity` and `User`
+objects for them.
```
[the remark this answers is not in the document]
stlaz: I am not aware of a specific location where we'd publish scripts we think people can use. I guess we might be able to provide such a script in a form of a KB article on request.

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
```diff
@@ -0,0 +1,167 @@
...
+
+Make the oauth-server capable of setting up identities for OIDC identity
+providers that use the URI scheme in the `sub` claim of their ID tokens.
+
+## Motivation
+
+Per the [OIDC specification](https://openid.net/specs/openid-connect-core-1_0-final.html#IDToken)
+of the ID token and its more thorough details described in
+[RFC7519 - JSON Web Token (JWT)](https://tools.ietf.org/html/rfc7519#section-4.1.2),
+the `sub` claim may be represented as both a string or a URI. However, URI by
```
sttts, line 42: nit: an URI is a string.

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
```diff
@@ -0,0 +1,167 @@
...
+   to OpenShift successfully.
+
+### Non-Goals
+
+None.
+
+## Problem Summary
+
+With OpenShift 4.x, the oauth-server no longer allows [configuring the claim
+of ID tokens](https://github.com/openshift/api/blob/670ac3fc997c2f1d19b8c29ef04f70d6e3d4a59e/osin/v1/types.go#L341)
```
sttts, line 61: to understand: in the 3.x past those customers could configure another claim than `sub` and work around the problem?

matthyx: yes, this is what we are doing currently in our v3 clusters

stlaz: yes, that's how it used to work

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
```diff
@@ -0,0 +1,167 @@
...
+that had its `Identity` created prior to the version containing this fix.
+
+Proposal: if a `sub` claim of an ID token contains either `:` or `/`, url-encode
+the value of the claim and create the new identity object like so:
+
+```yaml
+apiVersion: user.openshift.io/v1
+kind: Identity
+metadata:
+  name: <identity provider name>:url:<url-encoded sub claim value>
```
sttts, line 93: What is the resource path of such an object with encoding?

sttts: `/apis/user.openshift.io/v1/foo:url:https%3A%2F%2Fheise.de%0A` is not valid I assume, is it?

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
```diff
@@ -0,0 +1,167 @@
...
+objects for them.
+
+## Design Details
+
+### Test Plan
+
+Configure an OIDC provider that uses URL scheme for the `sub` claim in its
+ID tokens (like Keycloak in some configurations), set it up as an OIDC identity
+provider for an OpenShift cluster, and attempt to log in as a user of this
+OIDC provider.
```
sttts, line 151: in automated test or just once manually?

stlaz: I think this could be automated

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
stlaz: that'd be something like: thisbeuri%3A%3A32423423%2F23432
I should probably change that to base64: dGhpc2JldXJpOjozMjQyMzQyMy8yMzQzMg==

WDYT?

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
stlaz: the linked RFC separates them in this sense:
```
   StringOrURI
      A JSON string value, with the additional requirement that while
      arbitrary string values MAY be used, any value containing a ":"
      character MUST be a URI [RFC3986].  StringOrURI values are
      compared as case-sensitive strings with no transformations or
      canonicalizations applied.

```

sttts: please clarify in the text here

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
sttts: please clarify in the text

enhancements/authentication/allowing-uri-scheme-in-oidc-sub-claims.md
sttts: b64 certainly makes it easier to reason about. Have you tried to understand exactly and write down here how url encoding will end up in the URL path? Am not saying that it does not work. Am just curious to see an example here.

stlaz: Hm, you're right, %s really won't make for a good, valid name. The more I think about it the more I want to use url-safe base64 instead. I'll change that.

sttts:
/lgtm
/approve

/hold

for PM approval.

PR #591, opened 2021-01-19
authentication: useroauthaccesstokens - don't remove system:oauth-token-deleters

stlaz:
This clusterrolebinding is a remnant of the Olds and changing it
might cause side-effects.

/assign @deads2k


deads2k:
This change looks good to me.  I don't see a reason to change permission at this point.  The existing CR and CRB don't cause any issue and keeping them ensures API behavior remains consistent.

/approve
/assign @sttts

PR #363, opened 2020-06-05
Enhancement for adding upgrade preflight checks for operators

LalatenduMohanty:
Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

PR #357, opened 2020-06-02
Supporting out-of-tree drivers on OpenShift

zvonkok:
Support out-of-tree drivers on OpenShift for enabling custom hardware and third-party kernel modules on container and non-container based operating systems. 

In OpenShift v3.x out-of-tree drivers could be easily installed on the nodes, since the node was a full RHEL node with a subscription and needed tools were installed easily with yum. 
In OpenShift v4.x this changed with the introduction of RHCOS. There are currently two different documented ways to enable out-of-tree drivers. One way is using an SRO based operator such as the nvidia GPU operator and the other is using kmods-via-containers. 
We want to come up with a unified solution that works for our customers across RHCOS and RHEL. The solution should also help customers that are currently on RHEL7 to consider moving to RHCOS which is fully managed and easier to support in OpenShift.

(As a side note we had DriverContainers running with OCP-3.11 on AtomicHost)

PR #592, opened 2021-01-19
Safer Cluster Upgrades for OLM Managed Operators

awgreene:
Introduces proposal to update OLM to consider if the collection of
Operators on cluster will continue to work on the next Minor OpenShift
version before reporting whether or not it is upgradeable to CVO.

### DIFF
--- a/enhancements/olm/max-openshift-versions-for-operators.md
+++ b/enhancements/olm/max-openshift-versions-for-operators.md
@@ -1,0 +1,284 @@
+---
+title: Allow OLM Managed Operators to Specify a Max OpenShift Version
+authors:
+  - "@awgreene"
+reviewers:
+  - TBD
+  - "@kevinrizza"
+approvers:
+  - "@ecordell"
+  - "@spadgett"
+creation-date: 2021-01-11
+last-updated: 2021-01-19
+status: provisional
+see-also:
+  -  N/A
+replaces:
+  - N/A
+superseded-by:
+  - N/A
+---
+
+# Allow OLM Managed Operators to Specify a Max OpenShift Version
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Definitions
+
+**OLM Managed Operators**: Operators that are installed, managed, and upgraded by the [Operator Lifecycle Manager (OLM)](https://github.com/operator-framework/operator-lifecycle-manager) project.
+
+**Operator Author**: The team responsible for the development of an operator.
+
+**Index Maintainers**: The tean that maintains a collection of [Operator Bundles](https://olm.operatorframework.io/docs/glossary/#bundle) in the form of an [Index](https://olm.operatorframework.io/docs/glossary/#index).
+
+**Minor OpenShift Upgrade**: When the OpenShift Cluster is upgraded to the next minor version, for example the upgrade from version 4.6 to 4.7. See [Semantic Versioning](https://semver.org/) for more details.
+
+**Patch OpenShift Upgrades**: When the OpenShift Cluster is upgraded to the next patch version, for example the upgrade from version 4.6.1 to 4.6.2. See [Semantic Versioning](https://semver.org/) for more details.
+
+**maxOpenShiftVersion**: The maximum `{major}.{minor}` OpenShift version that an operator supports, example: `4.5`.
+
+## Summary
+
+OpenShift Cluster Admins introduce new services to their clusters by way of OLM Managed Operators. When performing a Minor OpenShift Upgrade, there is no guarantee that these operators will continue to run on the upgraded cluster version.
+
+Many operators shipped by Red Hat are rigorously tested on a specific set of OpenShift versions. The teams responsible for this testing know exactly which `{major}.{minor}` versions of OpenShift that their operators can run on. OLM should allow Operator Authors to specify the maximum `{major}.{minor}` version of OpenShift that their operator may run on.
+
+With this information OLM should:
+
+- Prevent Minor OpenShift Upgrades when it is possible to determine that one or more installed operators will not run on the next minor OpenShift version.
+- Prevent the operator from being installed on OpenShift clusters whose version is greater than the `maxOpenShiftVersion` specified by the operator.
+
+## Motivation
+
+The primary purpose of this enhancement is to define how OpenShift can protect existing services introduced by OLM Managed Operators when performing Minor OpenShift Upgrades.
+
+### Goals
+
+- Prevent Minor OpenShift Upgrades when an installed operator declares that it does not support the next minor OpenShift version.
+- Warn Cluster Admins when an installed operator does not explicitly declare support for the next minor OpenShift version.
+- Prevent OLM from installing operators whose `maxOpenShiftVersion` is less than the current [ClusterVersion](https://github.com/openshift/api/blob/a9e731090f5ed361e5ab887d0ccd55c1db7fc633/config/v1/types_cluster_version.go#L11-L13).
+
+### Non-Goals
+
+- Ensure that services will not be disrupted from a cluster upgrade.
+- Guarantee that all installed OLM Managed Operators will continue to run after a Minor OpenShift Upgrade.
+- Prevent Patch OpenShift Upgrades.
+
+## Proposal
+
+The [ClusterVersionOperator](https://github.com/openshift/cluster-version-operator) provides OLM with the means to prevent Minor OpenShift Upgrades by setting the `operator-lifecycle-manager's` [ClusterOperator's upgradeable condition](https://github.com/openshift/api/blob/5935a5beec4bb8e1e81dd0fe9ebc2af36b9a09ae/config/v1/types_cluster_operator.go#L171-L175) to `False`.
+
+OLM currently reports that it is upgradeable as soon its deployments are successfully rolled out. OLM should instead set its `upgradeable condition` to reflect if any OLM Managed Operators indicate that they will not run on the next minor OpenShift version.
+
+The bulk of this enhancement focuses on defining:
+
+- How operators will define their `maxOpenShiftVersion`.
+- The steps that OLM will take to determine upgrade safety based on the collection of installed operators.
+- The steps that OLM will take to prevent the operator from being installed on OpenShift clusters whose version is greater than the `maxOpenShiftVersion` specified by the operator.
+
+### Allowing Operator Authors to Define a Maximum OpenShift Version
+
+Given that this is an OpenShift specific feature, the OLM team would rather not expand the `Spec` of the [ClusterServiceVersion (CSV)](https://olm.operatorframework.io/docs/concepts/crds/clusterserviceversion/), used on vanilla Kubernetes clusters, to include a `maxOpenShiftVersion` field.
+
+Instead, OLM will rely on the presence of an annotation on a CSV to specify the operator's `maxOpenShiftVersion`. This annotation can be added in one of two ways:
+
+1. By defining the `maxOpenShiftVersion` as a property of the bundle using the [Declarative Index Config](https://github.com/operator-framework/enhancements/blob/1cf4a0363d918e810f638036539388622265d466/enhancements/declarative-index-config.md) (recommended approach).
+2. By adding the annotation to the CSV directly.
+
+Let's review the benefits associated with defining an operator's `maxOpenShiftVersion` using the Declarative Index Config.
+
+#### Defining the Maximum OpenShift Version using the Declarative Index Config
+
+The Declarative Index Config provides the means to define [indexes](https://olm.operatorframework.io/docs/glossary/#index) in a declarative way. When this feature is available, OLM will allow Index Maintainers to specify each bundles `maxOpenShiftVersion`, an example of which can be seen below:
+
+```json=
+"etcd": {
+  "name": "etcd",
+  "bundles": [
+    {
+      "path": "quay.io/foo/etcdv0.0.1",
+      "version": "v0.0.1",
+      "channels": ["alpha"],
+      "operators.coreos.com/maxOpenShiftVersion": "4.6", # Prevent upgrades to 4.7
+    },
+    {
+        "path": "quay.io/foo/etcdv0.0.2",
+        "version": "v0.0.2",
+        "channels": ["alpha"],
+        "operators.coreos.com/maxOpenShiftVersion": "4.7", # Prevent upgrades to 4.8
+    },
+  ],
+  ...
+  ...
+  ...
+}
+```
+
+Like all bundle properties, these properties will be propagated to the CSV as an annotation. The value of the bundle's `maxOpenShiftVersion` can be easily updated without releasing a new bundle as soon as the existing bundle has been tested against a new version of OpenShift.
+
+#### How OLM Determines Upgradeability Status
+
+When OLM reconciles a CSV it will check the set of annotations for a key that matches `operators.coreos.com/maxOpenShiftVersion`. The value associated with this key specifies the maximum `{major}.{minor}` OpenShift version that the operators supports. An example of this annotation can be seen below:
+
+```yaml=
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    # Prevent upgrades to OpenShift Version 4.9
+    operators.coreos.com/maxOpenShiftVersion: "4.8" 
+```
+
+Assume that the CSV shown above is the only CSV on the Cluster. In this case, OLM would compare the value of the `operators.coreos.com/maxOpenShiftVersion` annotation against the version of the cluster to decide its upgradeable status.
+
+As this is an opt-in feature, there are instances where OLM will be unable to determine if the upgrade is safe:
+
+- The CSV is missing the `operators.coreos.com/maxOpenShiftVersion` annotation
+- The value associated with the `operators.coreos.com/maxOpenShiftVersion` is not a valid `{major}.{minor}` Semantic Version (Valid Example: x.y).
+
+OLM will report the following upgradeable conditions based on the collection of CSVs present on cluster:
+
+| Key                  | Meaning                                                                                            |
+| ---------------------| -------------------------------------------------------------------------------------------------- |
+| Upgradeable    CSV   | A CSV with a valid upgradeable annotation that will run on the next Minor version of OpenShift     |
+| Non-Upgradeable  CSV | A CSV with a valid upgradeable annotation that will not run on the next Minor version of OpenShift |
+| Undeterminable CSV   | A CSV that does not include a valid upgradeable annotation                                         |
+
+| Upgradeable CSV Present | Indeterminate CSV Present | Non-Upgradeable CSV Present | Status          | Message                                                                                            |
+| ----------------------- | ------------------------- | --------------------------- |---------------- | -------------------------------------------------------------------------------------------------- |
+| Yes                     | No                        | No                          | Upgradeable     | Ready for upgrade                                                                                  |
+| Don't Care              | Yes                       | No                          | Upgradeable     | The following operators may not run on the next OpenShift Version: namespace/foo, namespace/bar"   |
+| Don't Care              | Don't Care                | Yes                         | Not Upgradeable | The following operators will not run on the next OpenShift Version: namespace/foo, namespace/bar"  |
+
+### Preventing OLM Managed Operators Installation on Unsupported OpenShift Cluster Versions
+
+OLM Managed Operators can already specify a `minKubeVersion` on their Operator Bundles which prevents OLM from installing the operator on a Kubernetes Clusters whose version is less than the specified value.
+Similarly, the resolver will be updated to use the `operators.coreos.com/maxOpenShiftVersion` annotation to prevent operators from being installed on an OpenShift Version that is not supported. The resolver will be updated to:
+
+1. Check for the presence of the ClusterVersion API.
+2. If present, the resolver will check if the operator being installed includes a `maxOpenShiftVersion`. If a `maxOpenShiftVersion` is specified and the cluster version is less than the specified version, OLM will prevent the operator from being installed.
+
+### User Stories
+
+#### Story 1
+
+As an OpenShift cluster admin, I want OLM to block cluster upgrades if one or more of the operators that it manages will will not run on the next minor OpenShift version.
+
+#### Story 2
+
+As an operator author using OLM to manage the lifecycle of my operator, I want OLM to prevent upgrades to specific `{major}.{minor}` versions of OpenShift that I know are not supported by my operator.
+
+#### Story 3
+
+As an operator author using OLM to manage the lifecycle of my operator, I want OLM to prevent installations of my operators on OpenShift Clusters whose versions are greater the `maxOpenShiftVersion` defined by my operator.
+
+#### Story 4
+
+As the index maintainer, I want to be able to dynamically change which `{major}.{minor}` OpenShift versions that are supported by my operator without releasing a new version of the CSV.
+
+### Risks and Mitigations
+
+- The most immediate risk to this solution is the fact that OLM will prevent Minor OpenShift Upgrades when an OLM Managed Operator does not support the next minor OpenShift version.
+This risk is mitigated given that admins will still be able to apply security fixes with patch updates whether or not OLM is blocking Minor OpenShift Upgrades. This gives them a safe window to either remove the minor-blocking operator or update it to a version that is compatible with the next minor OpenShift version.
+In extreme cases, cluster admins can override the CVO's upgrade checks via means documented elsewhere.
+
+## Design Details
+
+### Open Questions
+
+- When a CSV is present that does not include a `maxOpenShiftVersion` annotation, how will this information be surfaced by the UI and CLI?
+
+### Test Plan
+
+#### Proposed Unit Tests
+
+- Logic used to determine upgrade safety should be thoroughly tested
+
+#### Proposed E2E Tests
+
+- The cluster upgrade is not blocked by CSVs that do not report a `maxOpenShiftVersion`.
+- The cluster upgrade is not blocked by CSVs that report an invalid `maxOpenShiftVersion`.
+- The cluster upgrade is not blocked by CSVs that report a `maxOpenShiftVersion` greater than the current OpenShift version.
+- The cluster upgrade is blocked by CSVs that report a `maxOpenShiftVersion` less than or equal to the current OpenShift version..
+
+For operators that specify a `maxOpenShiftVersion`:
+
+- The operator can be installed on a Kubernetes cluster.
+- The operator can be installed on OpenShift clusters whose version is less than or equal to the `maxOpenShiftVersion` specified by the operator.
+- The operator can be installed on OpenShift clusters whose version is greater than the `maxOpenShiftVersion` specified by the operator.
+
+For operators that do not specify a `maxOpenShiftVersion`:
+
+- The operator continues to be installed on both OpenShift and vanilla Kubernetes clusters.
+
+### Graduation Criteria
+
+The goal of this enhancement is to provide Cluster Admins some level of assurance that service made available via OLM Managed Operators continue to remain available on a targeted OpenShift Cluster Version.
+
+This feature will initially be introduced as a Generally Available Feature, but upgrade compatibility checks may be added overtime.
+
+Proposed GA Features:
+
+- OLM determines upgrade safety based on CSV annotations.
+- OLM prevents operators from being installed on OpenShift clusters whose versions are greater than than the specified `maxOpenShiftVersion`.
+- Console highlights upgrade safety in the UI.
+- User facing documentation available at olm.operatorframework.io.
+- Comprehensive unit tests exist.
+- Comprehensive e2e tests exist.
+
+#### Removing a deprecated feature
+
+Not applicable.
+
+#### Upgrade / Downgrade Strategy
+
+Not applicable.
+
+#### Version Skew Strategy
+
+The proposed feature should only require integration with the CVO component and OLM Managed Operators.
+
+## Implementation History
+
+- Initial enhancement proposal created.
+
+## Drawbacks
+
+- OpenShift users might be upset when a Minor OpenShift Upgrade is block by OLM if they are unfamiliar with this feature.
+- There are instances where an operator might not be supported on the targeted version that the operator is being upgraded to. If the customer is unable to remove the existing operator, this may place them in a position where they must choose between upgrading the cluster or removing the operator.
+This is arguably better than upgrading the cluster only to find out that a core service is no longer available on cluster, but I suspect that multiple tickets will be opened against OLM and the operators it installs when the customer finds themselves unable to upgrade.
+
+## Possible Future Work
+
+### Determine upgrade safety using OLM Managed Operator RBAC
+
+In addition to giving Operator Authors the ability to declare a maximum OpenShift Version, OLM could determine upgrade readiness based on default OpenShift GVKs and those required by installed operators. At a high level, the following workflow could be implemented:
+
+1. CVO provides OLM with a list of default GVKs at OpenShift version `x.y.z`. This information should be provided by the CVO team, they could possibly retrieve it from the same source that stores OpenShift upgrade channels.
+2. OLM retrieves the list of default GVKs available on OpenShift version `x.y.z`.
+3. OLM generates a list of required APIs by aggregating the GVKs defined as permissions and ClusterPermissions in existing ClusterServiceVersions. Any operators that uses a wildcard (*) for the Group, Version, or Kind in its permissions/clusterPermissions would automatically get flagged as OLM would be unable to guarantee that all GVK required by the operator would be available on cluster version `x.y.z`.
+4. OLM would then compare the list of apis on the cluster at version `x.y.z` against those required by installed OLM Managed Operators. If a required GVK is unavailable, OLM would report that it is no longer upgradeable.
+If the required GVKs are available, OLM would mark itself as upgradeable, possibly using a special Message/Reason to signified that it has compared required GVKs against available GVKs at target version.
+If OLM is unable to guarantee that the required GVKs are available on OpenShift version `x.y.z` that could be highlighted as well, the cluster admin could then force the upgrade if they feel comfortable doing so.
+
+#### Pros
+
+- This feature is available by default and does not require any additional effort from operator authors, giving some level of assurance that the upgrade is safe.
+
+#### Cons
+
+- This solution forces the CVO (or API team?) to host and provide available GVK information for each `{major}.{minor}` OpenShift version
+- This solution could falsely identify an upgrade as safe. Rather than tell customers that their operators will keep working, we should specify that all GVKs required by the operator are available on the `x.y.z` version of OpenShift.
+- There are many instances where an operator may not be supported on a specific version of OpenShift outside of required GVKs versus available GVKs.
+- Operator authors must specify complete GVKs in their CSVs if OLM is to report true/false for upgrade safety.
+
+## Infrastructure Needed
+
+No additional infrastructure is needed.


enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,267 @@
...
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Definitions
+
+**OLM Managed Operators**: Operators that are installed, managed, and upgraded by the [Operator Lifecycle Manager (OLM)](https://github.com/operator-framework/operator-lifecycle-manager) project.
+
+**Minor OpenShift Upgrade**: When the OpenShift Cluster is upgraded to the next minor version, for example the upgrade from version 4.6 to 4.7. See [Semantic Versioning](https://semver.org/) for more details.
```
gallettilance, line 37: is it always sequential or is it whatever upgrades OCP supports?

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Proposal
+
+The [ClusterVersionOperator](https://github.com/openshift/cluster-version-operator) provides OLM with the means to prevent Minor Semantic Version cluster upgrades by setting the  `operator-lifecycle-manager` [ClusterOperator's](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md) [upgradeable condition](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#what-should-an-operator-report-with-clusteroperator-custom-resource) to `False`.
+
+OLM currently reports that it is upgradeable as soon as the OLM Deployment is successfully rolled out. This existing logic should be updated to monitor the collection of OLM Managed Operators on cluster and update the `upgradeable condition` to reflect if any of these operators will not run on the next Minor OpenShift Version.
+
+The bulk of this enhancement focuses on defining the steps that OLM will take to determine upgrade safety based on the collection of installed operators.
+
+### Allowing Operator Authors to define supported OpenShift Versions
```
gallettilance, line 67: operators that publish to the redhat operators, isv and community operator catalogs that are on cluster use a label on their bundle image to indicate which OCP versions they support (see: https://docs.engineering.redhat.com/display/CFC/Delivery). It seems redundant to have them also add a field in the CSV.

gallettilance (commented):
I would prefer if OLM could inspect the bundle labels to see what it supports. All bundles that don't support a specific OCP version will have a max version set moving forward. The limitation is OLM won't be able to tell when a bundle has been deprecated or if it will be able to continue to receive updates after cluster upgrade. But it sounds like that's not the goal of the proposal.

enhancements/olm/safe-cluster-upgrades.md
awgreene: At this time OpenShift ALWAYS moves through each Minor OpenShift version when upgrading.

enhancements/olm/safe-cluster-upgrades.md
awgreene: @gallettilance thanks for sharing that delivery document. We can absolutely pursue that approach since this is an OpenShift only feature. However, there are two concerns that I would like to address before moving forward:

1. The "Max OpenShift Version" could only be inferred when the bundle image is shipped to a specific version of OpenShift (`V4.6`) or a range of OpenShift Version (`v4.5-v4.7`, where we would assume the end of the range as the Max OpenShift Version, 4.7 in this case) as all other forms are shipped to future OpenShift versions. This means that the "Max OpenShift Version" could not be inferred for a set of labels that are currently in use so OLM would still need to provide the ability for this subset of operators to specify a "Max OpenShift Version".
2. Currently, these labels simply identify which OpenShift versions they should be available on. These labels will instead prevent the Cluster from upgrading to the next version. I suspect there might be a number of operators that will unexpectedly prevent cluster upgrades if we make this change. Consider a 4.,6 cluster where a bundle is shipped labeled with the `=4.6` label and the next version of the bundle is shipped with `=4.7`. The 4.6 cluster would not be able to upgrade to 4.7 as the existing bundle would set the max OpenShift Version to 4.6. We could instead assume that the Max OpenShift Version is equal to 1 more than the provided version (Example: a bundle labeled with `=4.6` equals a "Max OpenShift Version" of `4.7`) but that would not handle the usecase where a bundle made available on 4.6 is known not to run on 4.7.

Overall, I am concerned that if we commandeer these labels make unfair assumptions about intended operator behavior.

gallettilance: 1. Operator authors can't modify their CSV after they are released. In both cases (label and CSV annotation) they have to use their best guess (which is why there is a deprecation mechanism in the pipeline) and that guess will be wrong exactly in the same way for both right? let me know if I misunderstood what you were trying to say
2. I think we need to distinguish "installable" and "upgradeable". In the case where I set a max version in the labels, the intention is for the version to be no longer installable on OCP versions after that max version. However, it may still be able to run and upgrade on OCP versions after that max version - if it cannot maybe this is where deprecation comes into play (given that the resolver has access to that information via properties on the bundle, it could surface that somehow). To be honest we haven't explicitely made that distinction in the pipeline but it definitely feels like a reasonable ask given this enhancement. What do you think?

awgreene: Thanks for the detailed response @gallettilance 

>1. Operator authors can't modify their CSV after they are released. In both cases (label and CSV annotation) they have to use their best guess (which is why there is a deprecation mechanism in the pipeline) and that guess will be wrong exactly in the same way for both right? let me know if I misunderstood what you were trying to say

There is no need to modify the CSV after it is released, consider the following example:
- Operator Foo has been developed for and tested against a 4.7 cluster.
- The Author sets the maxOpenShiftVersion annotation on Operator Foo's CSV to 4.7
- OpenShift 4.8 becomes available.
- Clusters with Operator Foo are unable to upgrade to 4.8
- A new version of Operator Foo is released in the existing 4.7 index / channel. This version of Operator Foo has been tested on OpenShift Version 4.8 and has had the maxOpenShiftVersion annotation set to 4.8.
- (Optional Step) All earlier versions of Operator Foo are marked as deprecated. (unsure if this is needed).
- Cluster Admins can now upgrade Operator Foo to the latest version, unblocking Cluster Upgrades.

>  I think we need to distinguish "installable" and "upgradeable"

Agreed!

>In the case where I set a max version in the labels, the intention is for the version to be no longer installable on OCP versions after that max version. However, it may still be able to run and upgrade on OCP versions after that max version - if it cannot maybe this is where deprecation comes into play (given that the resolver has access to that information via properties on the bundle, it could surface that somehow). To be honest we haven't explicitly made that distinction in the pipeline but it definitely feels like a reasonable ask given this enhancement. What do you think?

I believe that you are suggesting that we could prevent cluster upgrades based on whether or not the CSV has been deprecated. This is certainly possible, but I am concerned that:
- Operator Authors would be forced to mark a number of CSVs as deprecated before each OpenShift release to prevent cluster upgrades. Missing this step could allow Cluster Upgrades that disrupt operator services. Note that the workflow above does not include this downside.
- Upgrades would only be disabled if the index on the cluster is at the earliest version that includes all the deprecation changes.. 

If you want to pull in eyes from the pipeline team to discuss please do, I welcome any and all reviews :)

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,267 @@
...
+
+OLM must make an effort to prevent cluster upgrades when it is possible to determine that one or more installed operators will not run on the next version of OpenShift.
+
+## Motivation
+
+The primary purpose of this enhancement is to define how OpenShift can protect existing services introduced by OLM Managed Operators when performing a Minor OpenShift upgrade.
+
+### Goals
+
+- Prevent Minor OpenShift Upgrades when OLM detects that an installed operator will not run on the next Minor OpenShift version.
```
bparees, line 51:
```suggestion
- Prevent Minor OpenShift Upgrades when OLM detects that an installed operator may not run on the next Minor OpenShift version.
```

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,267 @@
...
+
+- Ensure that services will not be disrupted during a cluster upgrade.
+- Guarantee that all installed operator will continue to run after a cluster upgrade.
+- Prevent Patch OpenShift Upgrades.
+
+## Proposal
+
+The [ClusterVersionOperator](https://github.com/openshift/cluster-version-operator) provides OLM with the means to prevent Minor Semantic Version cluster upgrades by setting the  `operator-lifecycle-manager` [ClusterOperator's](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md) [upgradeable condition](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#what-should-an-operator-report-with-clusteroperator-custom-resource) to `False`.
+
+OLM currently reports that it is upgradeable as soon as the OLM Deployment is successfully rolled out. This existing logic should be updated to monitor the collection of OLM Managed Operators on cluster and update the `upgradeable condition` to reflect if any of these operators will not run on the next Minor OpenShift Version.
```
bparees, line 63:
```suggestion
OLM currently reports that it is upgradeable as soon as the OLM Deployment is successfully rolled out. This existing logic should be updated to monitor the collection of OLM Managed Operators on cluster and update the `upgradeable condition` to reflect if any of these operators indicate that they will not run on the next Minor OpenShift Version.
```

(TBD: what we want to do if the operator gives no indication either way)

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,267 @@
...
+| Upgradeable    CSV | A CSV with a valid upgradeable annotation that will run on the next Minor version of OpenShift     |
+| Unupgradeable  CSV | A CSV with a valid upgradeable annotation that will not run on the next Minor version of OpenShift |
+| Undeterminable CSV | A CSV that does not include a valid upgradeable annotation                                         |
+
+| Upgradeable CSV | Indeterminate CSV | Unupgradeable CSV | Is Upgradeable | Upgradeable Message |
+| --------------- | ----------------- | ----------------- | -------------- | ------------------------------------------------------------------------- |
+| X               |                   |                   | True           | All operators will run on the next Minor OpenShift Version                |
+| X               | X                 |                   | True           | One or more operator(s) might not run on the next Minor OpenShift Version |
+| X               | X                 | X                 | False          | Not all operators will run on the next Minor OpenShift Version            |
```
bparees, line 118: not sure why multiple columns have Xs here.  If there's a single "indeterminate" CSV, it doesn't matter if there are any "upgradeable" CSVs.

similarly if there's a signle unupgradeable CSV, it doesn't matter if there are any indeterminate or upgradeable CSVs.

Instead of Xs, should those boxes have a "don't care" value?

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,267 @@
...
+- The CSV is missing the `operators.coreos.com/maxOpenShiftVersion` annotation
+- The value associated with the `operators.coreos.com/maxOpenShiftVersion` is not a valid Semantic Version (Valid Example: x.y.z).
+
+OLM will report the following upgradeable conditions based on the collection of CSV present on cluster:
+
+| Key                | Meaning                                                                                            |
+| -------------------| -------------------------------------------------------------------------------------------------- |
+| Upgradeable    CSV | A CSV with a valid upgradeable annotation that will run on the next Minor version of OpenShift     |
+| Unupgradeable  CSV | A CSV with a valid upgradeable annotation that will not run on the next Minor version of OpenShift |
+| Undeterminable CSV | A CSV that does not include a valid upgradeable annotation                                         |
```
bparees, line 112: how will these conditions be presented to the admin?  In particular the "indeterminate" state needs to be clear to the admin since we are not(?) going to block upgrades when we are in that state.

enhancements/olm/safe-cluster-upgrades.md
awgreene: @bparees I believe that the wording is correct in this case, OLM will prevents upgrades when it can guarantee that one or more operator will not run on the next version of operator.because of the mentioned annotation.

enhancements/olm/safe-cluster-upgrades.md
awgreene: > not sure why multiple columns have Xs here. 

X was meant to indicate the presence of a CSV of that type.

>  If there's a single "indeterminate" CSV, it doesn't matter if there are any "upgradeable" CSVs.

OLM would not prevent an upgrade if one or more "indeterminate" CSVs are present provided no "unupgradeable" csv exists.

As long as you agree that that's the logic that we want I can work on improving the table to convey that point.

enhancements/olm/safe-cluster-upgrades.md
bparees: perhaps i'm being too cute here, but just because an operator says "I don't want to be run on 4.7" doesn't mean it can't actually run on 4.7.  That was the intent of my suggested change.

I was also trying to cover the case where the operator doesn't tell us anything.  Perhaps that needs to be added as a second goal ("Inform the admin when one or more operators do not explicitly declare support for the next OCP version so that the admin can decide whether or not to upgrade")

bparees: how about these two goals:

1) prevent minor version ocp upgrades when an installed operator declares that it does not support the next ocp minor version
2) warn the admin when an installed operator does not explicitly declare support for the next ocp minor version

?

enhancements/olm/safe-cluster-upgrades.md
bparees: > X was meant to indicate the presence of a CSV of that type.

yeah i don't think that's the best way to represent the logic, because the presence of an upgradeable CSV literally does not matter (and the presence of an indeterminate CSV does not matter if you have a non-upgradeable CSV present).  Hence my suggestion of "don't care" as the value to use.  As written the table doesn't explicitly state what happens if "there are no upgradeable CSVs and some indeterminate CSVs" because it only states what happens if "there are upgradeable CSVs and indeterminate CSVs"  (second row).


> OLM would not prevent an upgrade if one or more "indeterminate" CSVs are present provided no "unupgradeable" CSVs exists.

right.  but it doesn't matter if there are (or are not) "upgradeable" CSVs for that case.

enhancements/olm/safe-cluster-upgrades.md
awgreene: There are two ways to initiate a Cluster Upgrade, via the CLI and OpenShift Console UI.

# CLI
I am unsure how we could notify users that there is an operator that is "indeterminate". I believe the Cluster Admin would need to look at  the OLM ClusterOperator.

# OpenShift Console UI
The messages in ClusterOperators t is already available in the Cluster Admin panel:
ClusterOperators

 The UI Team could work to make this more visible.

enhancements/olm/safe-cluster-upgrades.md
awgreene: > yeah i don't think that's the best way to represent the logic, because the presence of an upgradeable CSV literally does not matter (and the presence of an indeterminate CSV does not matter if you have a non-upgradeable CSV present).

I disagree, the presence of an indeterminate CSV impacts the ClusterOperator message.

>  Hence my suggestion of "don't care" as the value to use. As written the table doesn't explicitly state what happens if "there are no upgradeable CSVs and some indeterminate CSVs" because it only states what happens if "there are upgradeable CSVs and indeterminate CSVs" (second row).


Perhaps instead of a table we should make it clear that:

- "Unupgradeable" operators set Degraded to true
- "Indeterminable" CSVs does not change Degraded to true but will update the message.
- -"Upgradeable" CSVs that reports a maxOpenShiftVersion

enhancements/olm/safe-cluster-upgrades.md
bparees: yeah i don't think there's much we can do about the cli.  Which is all the more reason we need to ensure that if we are in that state, the warning is very clearly presented to the admin.  (We can also document as part of the upgrade process that we recommend you check the OLM status before starting an upgrade).

For the UI, I was thinking we might want an explicit alert when we are in this state, rather than relying on the user to go look at the operator status.  But perhaps that is too aggressive/noisy.

bparees: Another option for the UI is to present/warn the user when they are on the upgrade panel, that OLM is not happy.

awgreene: > For the UI, I was thinking we might want an explicit alert when we are in this state, rather than relying on the user to go look at the operator status. But perhaps that is too aggressive/noisy.

By "alert" do you mean a warning box or a prometheus alert? I don't know that we need this information picked up by telemeter.

> Another option for the UI is to present/warn the user when they are on the upgrade panel, that OLM is not happy.

I think this is the best option. We could define a specific `reason` that alerts the UI to show a warning box, perhaps @spadgett or someone from his team can chime in.

bparees: > By "alert" do you mean a warning box or a prometheus alert? I don't know that we need this information picked up by telemeter.

i meant a prometheus alert since those get presented to the admin on the console.  i don't know that we care that it will be picked up by telemeter, but i do agree  that it may be too big of a hammer in general.  We can definitely start with a lighter touch (something on the upgrade UI page, and docs that tell the admin to check the status before starting an upgrade.  I think we already have docs that tell admins to upgrade all their OLM operators before starting an upgrade)

enhancements/olm/safe-cluster-upgrades.md
awgreene: I accepted the suggested changes, please mark the conversation as resolved if applicable.

enhancements/olm/safe-cluster-upgrades.md
awgreene: I accepted the suggested changes, please mark the conversation as resolved if applicable.

enhancements/olm/safe-cluster-upgrades.md
awgreene: Updated the table to reflect changes that @bparees and I discussed offline:

```

| Upgradeable CSV Present | Indeterminate CSV Present | Non-Upgradeable CSV Present | Status          | Message            |
| ----------------------- | ------------------------- | --------------------------- |---------------- | ------------------ |
| Yes                     | No                        | No                          | Upgradeable     | Ready for upgrade  |
| Don't Care              | Yes                       | No                          | Upgradeable     | Warning            |
| Don't Care              | Don't Care                | Yes                         | Not Upgradeable | Upgrade is blocked |
```

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,265 @@
...
+| Non-Upgradeable  CSV | A CSV with a valid upgradeable annotation that will not run on the next Minor version of OpenShift |
+| Undeterminable CSV   | A CSV that does not include a valid upgradeable annotation                                         |
+
+| Upgradeable CSV Present | Indeterminate CSV Present | Non-Upgradeable CSV Present | Status          | Message            |
+| ----------------------- | ------------------------- | --------------------------- |---------------- | ------------------ |
+| Yes                     | No                        | No                          | Upgradeable     | Ready for upgrade  |
+| Don't Care              | Yes                       | No                          | Upgradeable     | Warning            |
+| Don't Care              | Don't Care                | Yes                         | Not Upgradeable | Upgrade is blocked |
+
+Ben Parees  11:49 AM|
```
bparees, line 116: cut+paste issue?  :)
also i like the sample messages you provided in slack, so i'd go ahead and put them in.

bparees: (I was just being lazy when i wrote up the sample table, so i didn't write out the more detailed sample message)

awgreene: Ha yes a copy paste issue! Updated in the latest version of the PR! Please mark as resolved if we're all set!

enhancements/olm/safe-cluster-upgrades.md
ecordell: I think we should plan to make use of declarative index for this.

- A team publishes an operator v1 during openshift 4.6. It is tested on 4.6 and stamped with maxOCPVersion: 4.6
- 4.7 is released. Some combination of QE / automated tests / static analysis determines to some degree of confidence that the operator should or does run on 4.7. 
- The index is updated to overwrite `maxOCPVersion` to be 4.7 for v1.

The goal of the declarative config is to allow us to overwrite metadata (like maxOCPVersion) for operators after release.

Certainly what @awgreene said will allow us to make use of maxOCPVersion without declarative config - but we know that we want to deliver declarative config for 4.8 already.

bparees: @ecordell are you proposing that adopting declarative config is the only way operator authors will be able to specify maxVersion?

kevinrizza: Can we just include a bit about dropping properties into the bundle and making sure that this workflow works that way as well?

awgreene: I will bump the `Declaring Max OpenShift Version via the Declarative Index Config` section into the main proposal and outline the advantages that @ecordell discussed.

enhancements/olm/safe-cluster-upgrades.md
wking: there's been some musing about possibly allowing direct 4.6 -> 4.8 hops for folks leaving eus-4.6, but nobody has really dug into that yet.  So while it's generally going to be 4.{y} -> 4.{y+1}, it's good to keep in mind that that may get softened in the future.

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,263 @@
...
+
+OLM must make an effort to prevent cluster upgrades when it is possible to determine that one or more installed operators will not run on the next version of OpenShift.
+
+## Motivation
+
+The primary purpose of this enhancement is to define how OpenShift can protect existing services introduced by OLM Managed Operators when performing a Minor OpenShift upgrade.
+
+### Goals
+
+- Prevent Minor OpenShift Upgrades  when an installed operator declares that it does not support the next ocp minor version
```
wking, line 51: nit: "Upgrades  when" -> "Upgrades when" (doubled space).  And while we're touching this line, probably "ocp" -> "OCP" or "OpenShift" throughout the doc?

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Non-Goals
+
+- Ensure that services will not be disrupted during a cluster upgrade.
+- Guarantee that all installed operator will continue to run after a cluster upgrade.
+- Prevent Patch OpenShift Upgrades.
+
+## Proposal
+
+The [ClusterVersionOperator](https://github.com/openshift/cluster-version-operator) provides OLM with the means to prevent Minor Semantic Version cluster upgrades by setting the  `operator-lifecycle-manager` [ClusterOperator's](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md) [upgradeable condition](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#what-should-an-operator-report-with-clusteroperator-custom-resource) to `False`.
```
wking, line 62: nit: `master` links are unreliable, because things might be changed or moved.  Probably best to pin to a specific commit, and also probably better to link the official API instead of CVO's dev docs.  Like [this][1].

[1]: https://github.com/openshift/api/blob/5935a5beec4bb8e1e81dd0fe9ebc2af36b9a09ae/config/v1/types_cluster_operator.go#L171-L175

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Allowing Operator Authors to define supported OpenShift Versions
+
+Many Operators shipped by Red Hat are rigorously tested on a specific set of OpenShift Versions. The teams responsible for this testing know exactly which versions of OpenShift that their Operators can run on. OLM should allow Operator Authors to specify the maximum Major and Minor version of OpenShift that their operator may run on.
+
+Given that this is an OpenShift specific feature, OLM would rather not expand the Spec of the [ClusterServiceVersion (CSV)](https://olm.operatorframework.io/docs/concepts/crds/clusterserviceversion/), used on vanilla Kubernetes clusters, to include a `maxOpenShiftVersion` field. Instead, OLM will rely on the presence of an annotation on the CSVs to specify the max OpenShift Version that the operator may run on. This annotation can be added in one of two ways:
+
+1. By defining the Max OpenShift Version as a property of the bundle using the [Declarative Index Config](https://github.com/operator-framework/enhancements/blob/1cf4a0363d918e810f638036539388622265d466/enhancements/declarative-index-config.md) (recommended approach)
+2. By adding the annotation to the CSV Directly
```
wking, line 75: Is this the only option for #357, where the compatibility with the next OpenShift version is less about the OLM-managed operator and more about that OLM-child's operands (in the case of #357, the kernel modules being installed)?  I guess the special-resource operator would check version compat of the kernel modules it managed vs. some compiled-into-the-SRO knowledge of the kernel version that was likely coming with the next OCP minor, and then manage the annotation in its own CSV to pass the word on to OLM?

enhancements/olm/safe-cluster-upgrades.md
ecordell: > are you proposing that adopting declarative config is the only way operator authors will be able to specify maxVersion?

Nope - just that with the declarative config, there is no need to publish a _new_ bundle to support a new ocp versions. Existing bundles can be annotated with new support statements as we discover them.

enhancements/olm/safe-cluster-upgrades.md
ecordell: I don't think OLM can know which next y-stream version the user will pick with the current APIs provided by CVO.

If skipping releases is possible, we likely need to provide some configuration mechanism for OLM (a `targetOCPRelease` squirreled away somewhere). And it will be a docs effort to ensure anyone attempting to skip a release sets that field before update, to give OLM a chance to review the operators it manages.

Has there been any discussion of a two-stage ack for CVO, where the user will set a desired target version but the update won't actually start? That would allow OLM to read the desired target version and adjust accordingly.

enhancements/olm/safe-cluster-upgrades.md
bparees: > Nope - just that with the declarative config, there is no need to publish a new bundle to support a new ocp versions. Existing bundles can be annotated with new support statements as we discover them.

ack, ok.

enhancements/olm/safe-cluster-upgrades.md
bparees: > Has there been any discussion of a two-stage ack for CVO, where the user will set a desired target version but the update won't actually start? That would allow OLM to read the desired target version and adjust accordingly.

sounds like the preflight checks discussion, but i'm not sure where that stands at this point.

enhancements/olm/safe-cluster-upgrades.md
bparees: > I guess the special-resource operator would check version compat of the kernel modules it managed vs. some compiled-into-the-SRO knowledge of the kernel version that was likely coming with the next OCP minor, and then manage the annotation in its own CSV to pass the word on to OLM?

Not sure if you're suggesting the operator updates its own CSV at runtime to change its maxVersion.  If so, i don't think we had that in mind (not even sure if you can do that, or if OLM will stomp your attempted changes).

if you're just saying the operator team would set the annotation on their CSV prior to shipping the operator bundle, then yes, that's the expectation.

enhancements/olm/safe-cluster-upgrades.md
awgreene: Updated this portion of the document to outline the benefits of declaring the operators Max OpenShift Version using the Declarative Index Config.

enhancements/olm/safe-cluster-upgrades.md
awgreene: > if you're just saying the operator team would set the annotation on their CSV prior to shipping the operator bundle, then yes, that's the expectation.

Or the annotation could be set using the declarative index config.

enhancements/olm/safe-cluster-upgrades.md
sdodson: @wking the only thing on the table right now is an upgrade that doesn't apply updates to MachineConfigPools and that's dependent on us getting comfortable with kubelet version skew of N-2.

The preflights [enhancement](https://github.com/openshift/enhancements/pull/363) didn't get a warm reception though we do plan to try again in the future. They'll take the form of executing jobs defined in the target release image, so if at some point we were to skip a release it would be a 4.N cluster that fetches the target release image and runs jobs from that image.

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,270 @@
...
+> NOTE: OLM should allow users to specify Major.Minor.Patch versions in this field but OLM can only prevent Major.Minor OpenShift upgrades. By allowing Major.Minor.Patch versions to be specified, OLM will not need to modify the existing annotations if OpenShift ever allows ClusterOperators to prevent patch upgrades.
+
+Assume that the CSV shown above is the only CSV on the Cluster. In this case, OLM would compare the value of the `operators.coreos.com/maxOpenShiftVersion` annotation against the version of the Cluster to decide it's upgradeable status.
+
+Examples can be found below:
+
+| OpenShift Cluster Version   | CSV's Max OpenShift Version | Is Upgradeable?|
+| --------------------------- | --------------------------- | -------------- |
+| 4.6.z                       | 4.8.0                       | True           |
+| 4.7.z                       | 4.8.0                       | True           |
```
wking, line 151: This enhancement proposal claims to be asking for patch-level `operators.coreos.com/maxOpenShiftVersion` values above ("By allowing Major.Minor.Patch versions to be specified...").  But if you're running 4.7.30 (or whatever), your only 4.8 update targets might be 4.8.4 (or whatever).  That would be over the claimed 4.8.0 max.  If you want to keep the door open for patch-level specificity (because you don't trust the OpenShift core to preserve [the required backwards compatibility][1]?), you should probably give folks a way to say "I'm ok floating without claiming a particular patch will break me" like `4.8.*`.  If you're ok having everyone float, you can restrict the `maxOpenShiftVersion` to just `{major}.{minor}`.

[1]: https://semver.org/spec/v2.0.0.html#spec-item-6

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,270 @@
...
+
+| Upgradeable CSV Present | Indeterminate CSV Present | Non-Upgradeable CSV Present | Status          | Message                                                                                            |
+| ----------------------- | ------------------------- | --------------------------- |---------------- | -------------------------------------------------------------------------------------------------- |
+| Yes                     | No                        | No                          | Upgradeable     | Ready for upgrade                                                                                  |
+| Don't Care              | Yes                       | No                          | Upgradeable     | The following operators may not run on the next OpenShift Version: namespace/foo, namespace/bar"   |
+| Don't Care              | Don't Care                | Yes                         | Not Upgradeable | The following operators will not run on the next OpenShift Version: namespace/foo, namespace/bar"  |
+
+#### Pros
+
+- This approach provides allows Operator Authors to specify the exact versions of OpenShift that their operator supports.
```
wking, line 175: nit: "provides allows" -> "allows"

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### Cons
+
+- This makes upgrade safety opt-in for each Operator, meaning that OLM cannot guarantee upgrade safety unless each installed operator specifies a maxOpenShiftVersion.
```
wking, line 179: And even then, these are declarations about compatibility with a version you are not yet running.  Folks can get declarations wrong.  Even if we test operator A on OpenShift B, and it works under testing, that does not guarantee it will work on B for every customer.  Bugs happen.  I'd soften "guarantee" -> "calculate" or something.

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,270 @@
...
+
+### User Stories
+
+#### Story 1
+
+As an OpenShift Cluster Admin, I want OLM to block cluster upgrades if it can determine that at least one operator will not run on the next Minor OpenShift version.
+
+### Risks and Mitigations
+
+- The most immediate risk with this solution is the fact that OLM will prevent the Cluster from upgrading if an OLM Managed Operator will not run on the next Minor OpenShift Version. This is easily circumvented if the Cluster Admin initiates the upgrade with the `force` command or by removing the offending Operator.
```
wking, line 190: `force` is a blunt instrument and also waives signature checks on the incoming release and other important things.  Can we replace the second sentence with something like:

> Admins will still be able to apply up security fixes with patch updates, regardless of OLM operators.  This gives them a safe window to either remove the minor-blocking operator or update it to a version that is compatible with the next minor OpenShift version.

And not mention forcing at all?  Or say something like "In extreme cases, admins can `force` to waive the `Upgradeable=False` guard (and all other guards) and update minor OpenShift versions anyway." if you feel like you really want to call out the safety-valve.

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Design Details
+
+### Open Questions
+
+- How will OLM rejecting an upgrade be surfaced in both the UI and the CLI?
```
wking, line 196: Should be no need for anything special.  The message OLM sets in its ClusterOperator will be picked up by the CVO and stored in ClusterVersion.  And both `oc adm upgrade ...` and the web console expose ClusterVersion's `Upgradeable` condition today, and there's an alert for `Upgradeable=False`.

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,270 @@
...
+4. OLM would then compare the list of apis on the cluster at version `t` against those required by installed OLM Managed Operators. If a required GVK is unavailable, OLM would report that it is no longer upgradeable. If the required GVKs are available, OLM would mark itself as upgradeable, possibly using a special Message/Reason to signified that it has compared required GVKs against available GVKs at target version. If OLM is unable to guarantee that the required GVKs are available on OpenShift version `t` that could be highlighted as well, the cluster admin could then force the upgrade if they feel comfortable doing so.
+
+#### Pros
+
+- This feature is available by default and does not require any additional effort from Operator Authors, giving some level of gurantee that the upgrade is safe.
```
wking, line 259: nit: "gurantee" -> "assurance"

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,270 @@
...
+- There are instances where an Operator might not be supported on the targeted version that the operator is being upgraded to. If the customer is unable to remove the existing operator, this may place them in a position where they must choose between upgrading the cluster or removing the operator. This is arguably better than upgrading the cluster only to find out that a core service is no longer available on cluster, but I suspect that multiple tickets will be opened against OLM and the operators it installs when the customer finds themselves unable to upgrade.
+
+## Possible Future Work
+
+### Determine upgrade safety using OLM Managed Operator RBAC
+
+In addition to giving Operator Authors the ability to provide the Max OpenShift Version, OLM could determine upgrade readiness based on default OpenShift GVKs and those required by installed Operators. At a high level, the following workflow could be implemented:
```
wking, line 250: Something like this exists today, where operators which will remove a particular API in the next OpenShift minor will set `Upgradeable=False` if they see any current consumers of that API.  If the only compat issue you're worried about is Kube-objects, maybe we don't need this enhancement?  You would need this enhancement for non-Kube things like RHCOS-kernel compat (#357), so if you grow OLM-side Kube API checks in the future, there should be ways for OLM-child operators to opt-into declaring additional restrictions beyond the ones the Kube API checks turn up.

enhancements/olm/safe-cluster-upgrades.md
bparees: This is a good point, @wking.  Given we can't (won't) actually enforce the z-stream level, we should probably not allow it to be specified in the field.

if that changes someday, we can loosen the validation on the field.

enhancements/olm/safe-cluster-upgrades.md
bparees: We definitely have to mention it.  But your caveat about the additional implications of using `force` makes sense (though i'd say that's more relevant to product documentation than the EP).

enhancements/olm/safe-cluster-upgrades.md
bparees: Not sure which scenario this bullet was aimed at, but there's not just "rejecting" updates, there's also the case where we are warning the admin that we can't be sure (because one or more operators doesn't make any declaration at all).

enhancements/olm/safe-cluster-upgrades.md
kevinrizza: Isn't this a bit ugly for the EUS->EUS case? So every intermediate upgrade will need to stop, check the UI to see if there is anything that's showing the warning, then resume assuming nothing is wrong? Or could we potentially have something more explicit in the case of EUS->EUS?

bparees: There is no "EUS->EUS" upgrade, it's just a series of upgrades from 4.6->4.70->4.8->4.9->4.10. 

So the EUS->EUS experience is already "make sure you're ready to upgrade.  run an upgrade to the next minor.  make sure it succeeded. make sure you're ready for the next tone.  run the next minor.  rinse/repeat until you get to 4.10".

So while this adds an extra step in what it means to make sure you're ready to upgrade, it's not fundamentally changing the fact that EUS->EUS is still a series of independent upgrades that have to be managed properly by an admin.  

What we (as an org) have been asked to do in support of it is:
1) do everything we can to ensure those upgrades will be successful/inform the admins when the upgrade may not be successful
2) do everything we can to ensure those upgrades are smooth/fast were possible

enhancements/olm/safe-cluster-upgrades.md
sdodson: Yeah, above it's stated that this is meant to allow a future state where we'd allow inhibition of patch upgrades without changing existing annotations but that's only true if they chose an inflated value from the start, ie: 4.8.999999

While there have been past proposals for `UpgradeableMicro` or `UpgradeablePatch` to inhibit patch version upgrades those have been rejected because we always want to allow ourselves the ability to alter the logic for setting `Upgradeable` in a later patch version if necessary. If we ever inhibit patch versions we then pushing admins toward forcing an upgrade and we never want to encourage that.

enhancements/olm/safe-cluster-upgrades.md
sdodson: > This is easily circumvented by removing the offending Operator or by overriding the CVO's checks via means documented elsewhere.

It seems necessary to make it clear it's possible but not provide documentation on how to do it here.

enhancements/olm/safe-cluster-upgrades.md
```diff
@@ -0,0 +1,270 @@
...
+OLM must make an effort to prevent cluster upgrades when it is possible to determine that one or more installed operators will not run on the next version of OpenShift.
+
+## Motivation
+
+The primary purpose of this enhancement is to define how OpenShift can protect existing services introduced by OLM Managed Operators when performing a Minor OpenShift upgrade.
+
+### Goals
+
+- Prevent Minor OpenShift Upgrades  when an installed operator declares that it does not support the next ocp minor version
+- Warn the Cluster Admin when an installed operator does not explicitly declare support for the next ocp minor version
```
bparees, line 52: it seems like we should also prevent installation of operators that specify a maxKube/OCP version that is lower than the current cluster version.  Can we add that as a goal here? (similar to what minKubeVersion does)

kevinrizza: Do we want to do that though? I understand the reasoning of minkubeversion because it is easy to be explicit about, but I could imagine a world where an operator is just not maintained anymore and the maxkubeversion isn't accurate but a user really wants to install it, and we do not have a way today of overriding a constraint like that.

bparees: > we do not have a way today of overriding a constraint like that.

i think that's the actual problem, right?  We need to start introducing ways for admins to override these blocking requirements.

enhancements/olm/safe-cluster-upgrades.md
awgreene: Understood and thank you for the information. I will update the PR to only allow `{major}.{minor}` declarations.

enhancements/olm/safe-cluster-upgrades.md
awgreene: Rewording for clarity:

> - When a CSV is present that does not include a maxOpenShiftVersion annotation, how will this information be portrayed via the Console UI and CLI?

enhancements/olm/safe-cluster-upgrades.md
awgreene: This enhancement is focused on preventing unsafe cluster upgrades, not preventing installation of operators onto a OpenShift version that is high than their defined max OpenShift version. I would rather not dilute this enhancement by expanding its scope. If this feature is needed, could we create an enhancement that focuses on expanding the resolver to include it?

enhancements/olm/safe-cluster-upgrades.md
awgreene: Thanks for the review @wking!

> Something like this exists today, where operators which will remove a particular API in the next OpenShift minor will set Upgradeable=False if they see any current consumers of that API. If the only compat issue you're worried about is Kube-objects, maybe we don't need this enhancement? You would need this enhancement for non-Kube things like RHCOS-kernel compat (#357), so if you grow OLM-side Kube API checks in the future, there should be ways for OLM-child operators to opt-into declaring additional restrictions beyond the ones the Kube API checks turn up.

The existing enhancement is focused on providing operator authors or index owners a way to specify a maximum OpenShift version per bundle and is required to prevent an cluster upgrade form disabling a service provided by an operator. Based on the frequency of conversations on this topic with OLM customers this feature is highly requested.

The possible expansion of this feature to have OLM to inspect RBAC required by installed operators is simply a nice to have that allows OLM to derive when a cluster upgrade may disable services offered by operators due to the deprecation of GVKs required by the operators.

> And there's also a chance that an OLM-child operator isn't using a particular, vulnerable Kube API right now, but will try to start using it once a user touches some knob. So there is a benefit to declaring a complete set of possible consumed APIs, instead of relying entirely on current-use detection.

In general, the RBAC required by an operator is declared in the `permission` and `clusterPermission` fields of its CSV. Its possible that a knob exists which causes the operator to perform privilege escalation if it has role/rolebinding rbac, but that seems like an abusive usecase.

enhancements/olm/safe-cluster-upgrades.md
bparees: > This enhancement is focused on preventing unsafe cluster upgrades, not preventing installation of operators onto a OpenShift version that is high than their defined max OpenShift version. 

I have mixed feelings.  I agree that's what this EP is focused on, but i wonder if the EP *should* be focused on "adding a maxocpversion metadata" and all that implies (blocking upgrades, blocking operator installs).

I'm generally happy with iterative development and i don't want to make it harder to get this landed than it needs to be, but it feels weird/confusing to introduce a "maxocp/kubeversion" field for operator authors to specify that doesn't actually enforce that restriction on install, especially when that is exactly what the existing minKubeVersion api field does.  It's also slightly problematic in my mind to add future restrictions based on that field after operator teams start adopting it (we generally try to avoid changing the behavior of apis).

Net being:
1) it can certainly be a separate EP
2) but we need to think long and hard about what it means if we only land the implementation of one of those EPs and not the other, when we introduce the api.

From an implementation perspective, am I naive to think it's pretty trivial to add this constraint to the resolver?  Adding the ability for admins to override the constraint i'm sure is harder, but i don't think we need that on day 1.

kevinrizza: You are right to insinuate that implementing the constraint is trivial, but the override is a much more hotly debated and nebulous thing that will require significantly more work. I think you do have a good point that introducing this without actually blocking is going to be an unexpected experience for operator devs though.

IMO I think it's reasonable to include that constraint in this enhancement proposal and deal with how to override it in a future enhancement. We aren't actually going to see too many folks with an explicit need to override this until at least one or two releases after we even ship a feature like this.

bparees: >  We aren't actually going to see too many folks with an explicit need to override this until at least one or two releases after we even ship a feature like this.

yeah, that was my thinking in saying it'd be ok to have the constraint w/o a way to override, initially.

enhancements/olm/safe-cluster-upgrades.md
awgreene: Addressed in the latest version of the enhancement.

enhancements/olm/safe-cluster-upgrades.md
awgreene: Addressed in the latest version of the enhancement.

awgreene: Please mark as resolved if applicable.

enhancements/olm/safe-cluster-upgrades.md
awgreene: Addressed in the latest version of the enhancement.

enhancements/olm/safe-cluster-upgrades.md
awgreene: Addressed in the latest version of the enhancement.

enhancements/olm/safe-cluster-upgrades.md
awgreene: Addressed in the latest version of the enhancement.

enhancements/olm/max-openshift-versions-for-operators.md
```diff
@@ -0,0 +1,276 @@
...
+## Implementation History
+
+- Initial enhancement proposal created.
+
+## Drawbacks
+
+- OpenShift users might be upset when a Minor OpenShift Upgrade is block by OLM if they are unfamiliar with this feature.
+- There are instances where an operator might not be supported on the targeted version that the operator is being upgraded to. If the customer is unable to remove the existing operator, this may place them in a position where they must choose between upgrading the cluster or removing the operator. This is arguably better than upgrading the cluster only to find out that a core service is no longer available on cluster, but I suspect that multiple tickets will be opened against OLM and the operators it installs when the customer finds themselves unable to upgrade.
+
+## Possible Future Work
```
bparees, line 252: another bit of future work that's at least related to this is providing the ability to override the install requirement.

enhancements/olm/max-openshift-versions-for-operators.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Risks and Mitigations
+
+- The most immediate risk to this solution is the fact that OLM will prevent Minor OpenShift Upgrades when an OLM Managed Operator does not support the next minor OpenShift version. This risk is mitigated given that admins will still be able to apply security fixes with patch updates whether or not OLM is blocking Minor OpenShift Upgrades. This gives them a safe window to either remove the minor-blocking operator or update it to a version that is compatible with the next minor OpenShift version. In extreme cases, cluster admins can override the CVO's upgrade checks via means documented elsewhere.
+
+## Design Details
+
+### Open Questions
+
+- When a CSV is present that does not include a `maxOpenShiftVersion` annotation, how will this information be surfaced by the UI and CLI?
```
bparees, line 191: has the UI team been looped into this so they know work is coming for them?
cc @spadgett

bparees (commented):
looks good to me.

sdodson:
/approve
This looks good and meets the needs I have for EUS, thank you all.

enhancements/olm/max-openshift-versions-for-operators.md
```diff
@@ -0,0 +1,284 @@
...
+
+OLM currently reports that it is upgradeable as soon its deployments are successfully rolled out. OLM should instead set its `upgradeable condition` to reflect if any OLM Managed Operators indicate that they will not run on the next minor OpenShift version.
+
+The bulk of this enhancement focuses on defining:
+
+- How operators will define their `maxOpenShiftVersion`.
+- The steps that OLM will take to determine upgrade safety based on the collection of installed operators.
+- The steps that OLM will take to prevent the operator from being installed on OpenShift clusters whose version is greater than the `maxOpenShiftVersion` specified by the operator.
+
+### Allowing Operator Authors to Define a Maximum OpenShift Version
```
aravindhp, line 86: Can I specify both `minKubeversion`and `maxOpenShiftVersion`? If yes, what happens if `minKubeversion` is `v0.19` and `maxOpenShiftVersion` is `v4.6`? I suggest you call out the various permutations / combinations and the expected result.

bparees: they're independent constraints, I would expect both of them must be met to install, along with any other constraints.

enhancements/olm/max-openshift-versions-for-operators.md
```diff
@@ -0,0 +1,284 @@
+---
+title: Allow OLM Managed Operators to Specify a Max OpenShift Version
```
camilamacedo86, line 2: Hi @awgreene, 

I know that this EP is merged already. However, it only came to my attention now.
I raised some concerns/questions over it in the channel which might be valid we address here. 

1. We have the OCP label `LABEL com.redhat.openshift.versions: v4.5-v4.7` that can be used in the image index already. What happens if a user adds the `label com.redhat.openshift.versions` in the image and the `maxOCPversion` annotation in the CSV? Also, as described in the https://github.com/openshift/enhancements/pull/592/files#r570641713 how will we ensure that the value in the annotation is compatible with minKubeversion?

- Should not CVP verify if it?
- Should not this annotation and its compatibility with `minKubeversion` be validated via the operator-framework/api and then, in this way, it is checked via `operator-sdk bundle validate` and etc? 

2. Could also we add the annotation for maxKubeVersion in order to make it compatible? Also, would not make sense deprecated the attr minKubeVersion and add it via annotation as well? And then, we have a minOCPVersion too?  

3. Would be possible to make this annotation be generic and valid for any vendor instead of being specific for openshift? 

c/c @kevinrizza @dmesser

awgreene: Hello @camilamacedo86 - thanks for the review!

> We have the OCP label LABEL com.redhat.openshift.versions: v4.5-v4.7 that can be used in the image index already. What happens if a user adds the label com.redhat.openshift.versions in the image and the maxOCPversion annotation in the CSV? 

Can you elaborate on this label and its impact? I believe that @gallettilance and I [discussed](https://github.com/openshift/enhancements/pull/592#discussion_r560518168) this earlier and I had some concerns with overloading the existing field.

> Also, as described in the https://github.com/openshift/enhancements/pull/592/files#r570641713 how will we ensure that the value in the annotation is compatible with minKubeversion? Should not CVP verify if it?

There is no guarantee that these annotation will not conflict. If either of these constraints are not met the operator will not be installed.

> Should not this annotation and its compatibility with minKubeversion be validated via the operator-framework/api and then, in this way, it is checked via operator-sdk bundle validate and etc?

Again, these are independent constraints. Also, if this approach was pursued, are you proposing that the validation library hardcodes the relationship between OCP and K8s version?

> Could also we add the annotation for maxKubeVersion in order to make it compatible?

This could certainly be added.

> Also, would not make sense deprecated the attr minKubeVersion and add it via annotation as well? And then, we have a minOCPVersion too?

We could remove this field from existing CRDs but I do not believe that it is worth doing so until we release a new version of the CRD as existing CRs would need to be migrated or OLM would need to ship a conversion webhook.

> Would be possible to make this annotation be generic and valid for any vendor instead of being specific for openshift?

While I agree with this in principle, I am not sure how users could provide specific versions for different Kubernetes variations without unique annotations. For example, how could an Operator Author generically specify that their operator's `maxVersion` is `4.7` for openshift but `1.21` for vanilla k8s?

camilamacedo86: Hi @awgreene, 

Sorry, if I could not clarifies. My mainly questions/concerns are about:

**Is it aligned with what exists already or are we planning to change what exists to be aligned with?** 

Otherwise, how it works and how it should be used can end up bring confusion and not to be intuitive to the end-users and us indeed. (e.g skipRange which does not work exactly as skips )  

**How users will know that they did not make something wrong? How we will know what should be the expected behaviour? How can we prevent an operator be published with an OCP label and max version or minKubeversioon that is not compatible and will never be installed?**

Example A:
- LABEL com.redhat.openshift.versions: v4.5-v4.7
- maxOCPversion=4.6 

Will the operator be installed or not in 4.7? 

Example B:
- LABEL com.redhat.openshift.versions: v4.5-v4.7
- maxOCPversion=4.8

Will the operator be installed in 4.8 or not? 

- And then, will the `maxOCPversion` prevent the operator be in the index catalogue as it is done for skips or will it works as skipRange? 
- How the users will know that thy added an invalid value for this field before publish and check that the operator is not installed? How can we prevent this scenario? Would be possible to add a check to the bundle validate for that? 

**Could we provide a solution that is valid for upstream and downstream?**

Do we really need to create a specific annotation for OCP or could we have for example maxVendorVersion and compare that version with the vendor version one? I am not sure if it would be possible, however, did we check if we could use k8s api to get the vendor version and do the check no matter the vendor kind or this information cannot be obtained by k8s api?

bparees: @camilamacedo86 the label decides which catalogs the bundle will be in.  the min/maxVersion field decides which clusters it can be installed on.

They are related, but independent.  There is nothing invalid about saying "put this bundle in the v4.8 catalog" and then saying "this bundle can't be installed on v4.8".  It's a poor choice by the operator team, but there's no need to explicitly forbid it. (Among other things, a cluster admin could still reference the v4.8 catalog from another cluster version which might be able to install the operator.  Also someday we're going to have the ability for admins to force the installation despite requirements not being met).


> Do we really need to create a specific annotation for OCP or could we have for example maxVendorVersion and compare that version with the vendor version one? I am not sure if it would be possible, however, did we check if we could use k8s api to get the vendor version and do the check no matter the vendor kind or this information cannot be obtained by k8s api?

what would operator teams set "maxVendorVersion" to?  Openshift has both an OCP version and a kube version.  Both versions can be retrieved and compared.

camilamacedo86: Thank you for the clarifications.

PR #593, opened 2021-01-19
Add proposal for hiding container mountpoints from systemd

lack:


enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,139 @@
+---
+title: hide-container-mountpoints
+authors:
+  - "@lack"
+reviewers:
+  - TBD
+approvers:
+  - TBD
```
haircommander, line 8: you can add myself, @mrunalp  and @umohnani8 here (to cover reviews for both cri-o and MCO bits)

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,139 @@
...
+overhead with no difference in functionality.
+
+## Motivation
+
+Systemd scans and re-scans mountpoints many times, adding a lot to the CPU
+utilization of systemd and overall overhead of the host OS running OpenShift.
+Changing systemd to reduce its scanning overhead is tracked in [BZ
+1819868](https://bugzilla.redhat.com/show_bug.cgi?id=1819868), but we can work
+around this exclusively within openshift. Using a separate mountpoint to for
+both crio-o and kubelet mounts can completely segregate all container-specific
```
haircommander, line 39: nit: s/crio-o/CRI-O

in general, for docs like this, we should use CRI-O (and I believe Kubelet should be capitalized as well)

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,139 @@
...
+
+The end-user experience should not be affected in any way by this proposal, as
+there is no outward API changes. There is some supportability change though,
+since anyone attempting to inspect the cri-o or kubelet mountpoints externally
+would need to be aware that these are now available in a different namespace
+than the default top-level systemd mount namespace.
+
+### Implementation Details/Notes/Constraints
+
+Here is an example of the container-mount-namespace.service:
```
haircommander, line 82: should have a note that these will most likely be configured by the MCO

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,139 @@
...
+    PrivateMounts=on
+    ExecStart=bash -c 'while :; do sleep infinity; done'
+
+This needs to be managed separately from either crio or kubelet to avoid the
+namespaces getting out of sync if either of those services restart. The systemd
+'PrivateMounts=on' facility does a nice job of creating a separate mountpoint,
+and both crio and kubelet can find the associated namespace via 'nsenter' as
+follows:
+
+    nsenter -m -t $(systemctl show --property MainPID --value container-mount-namespace.service) $ORIGINAL_EXECSTART
```
haircommander, line 97: both of the service files will have to also have
`After=container-mount-namespace.service`

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Risks and Mitigations
+
+Any other external or 3rd-party tools would need to change to match.
```
haircommander, line 101: a field should be added in open questions to investigate who may be affected by this

enhancements/hide_container_mountpoints.md
haircommander: also all mentions of kubernetes and openshift should be Kubernetes and Openshift, I think

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,139 @@
...
+
+The fact that there is a kubernetes test that explicitly tests that the
+container mountpoints are available to the parent operating system implies that
+this may have been desirable to someone at some level at one time in the past.
+ > 1. What is the reason for the kubernetes test? Is it okay to just skip or
+ >    disable the test in OpenShift?
+
+### Test Plan
+
+- Disable the existing Kubernetes e2e test that checks that all mountpoints are
```
haircommander, line 115: or, should the tests be run in this namespace as well?

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,139 @@
...
+- Initial proof-of-concept example (here)[https://github.com/lack/redhat-notes/tree/main/crio_unshare_mounts]
+
+## Drawbacks
+
+- Requires re-wrapping cri-o and kubelet services in bash and nsenter.
+- Requires get-and-set of the ExecStart stanza from both of those services.
+- If the namespace service restarts and then either cri-o or kubelet restarts,
+  there will be a mismatch between the mount namespaces and containers will
+  start to fail. Could be mitigated by enhancing cri-o pinns utility to pin and
+  reuse mount namespaces.
```
haircommander, line 131: this option should be in the alternatives, and I think some version of it should be considered. having the external service could be prone to timing issues, whereas having no process to keep the namespace open will reduce the risk, IMO.

romfreiman:
+1 we so in single node openshift flow that systemd consumes significant cpu resources

enhancements/hide_container_mountpoints.md
lack: Probably `Requires=` too?

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Motivation
+
+Systemd scans and re-scans mountpoints many times, adding a lot to the CPU
+utilization of systemd and overall overhead of the host OS running OpenShift.
+Changing systemd to reduce its scanning overhead is tracked in [BZ
+1819868](https://bugzilla.redhat.com/show_bug.cgi?id=1819868), but we can work
+around this exclusively within openshift. Using a separate mountpoint to for
+both crio-o and kubelet mounts can completely segregate all container-specific
+mounts away from any systemd interaction whatsoever.
```
mrunalp, line 40: kubelet uses systemd to setup mounts. Any impact to that?

enhancements/hide_container_mountpoints.md
lack: That's a good point.

I don't think timing would be an issue, provided both crio.service and kubelet.service have `After=` this service, but it would be more robust to have a systemd service that uses pinns to create the namespace one time idempotently, and then nothing need rely on that.

enhancements/hide_container_mountpoints.md
mrunalp: Yeah big worry is about anything external that cares about the mounts.

enhancements/hide_container_mountpoints.md
lack: Can you expand a bit more on what you mean by this?  What is the current kubelet/systemd interaction like?

enhancements/hide_container_mountpoints.md
lack: My current proof-of-concept does that, but I was thinking for the real implementation we could create them at the same time we currently create the `kubelet.service`, maybe?  Open to suggestions about where this would be best done.

enhancements/hide_container_mountpoints.md
lack: That's a good idea.

enhancements/hide_container_mountpoints.md
lack: Added a paragraph about this as well.

lack:
@romfreiman : Some ad-hoc testing we did with my workaround shows a noticable reduction in systemd CPU usage. No hard numbers yet.

enhancements/hide_container_mountpoints.md
lack: Added a discussion about approach into the "open questions" section.

enhancements/hide_container_mountpoints.md
derekwaynecarr: i thought we hit perf issue with these mounts and being inspected in systemd cadvisor handler.

https://github.com/google/cadvisor/pull/1573

i am not aware who actually cares about the mounts in question at this point...

rphillips:
cc @jsafrane

rphillips:
cc @kolyshkin @giuseppe

enhancements/hide_container_mountpoints.md
jsafrane: In OCP 3.x we allowed kubelet to run in a docker container. Therefore it called `nsenter` to escape the container's mount namespace (mount propagation was not used at that time) and `systemd-run --scope -- mount -t <type> <what> <where>` to run any potential fuse mount daemons outside of kubelet's pid namespace / kubelet service cgroup.

`nsenter` was removed, but we still drag `systemd-run ... mount  <what> <where>`, to run fuse daemons outside of kubelet's cgroup. `systemctl restart kubelet.service` would kill them.

jsafrane: Btw, this means that there is a short-lived systemd service that does the mount and dies. We do not support any fuse-based storage in 4.x. Upstream kubernetes does though. I don't think it affects how system handles mounts.

enhancements/hide_container_mountpoints.md
giuseppe: Podman/Skopeo/Buildah will also need to run in the same mount namespace, otherwise they will end-up setting up the storage in the host mount namespace, possibly causing corruption of the storage

jsafrane: CSI driver pods / containers typically mount `/var/lib/kubelet` into their pods with `rshared` mount propagation. They must mount volumes there and kubelet + the container runtime must see them there. Any change in this behavior will likely break storage in OCP clusters.

What will a pod with `/var/lib/kubelet` HostPath volume actually see? The systemd or kubelet mount namespace?

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,198 @@
...
+ >    wrap the command invocation for CRI-O and Kubelet, or should both CRI-O
+ >    and Kubelet be altered to accept a mount namespace PID or file as a
+ >    commandline option?
+
+For testing, the fact that there is a Kubernetes test that explicitly tests
+that the container mountpoints are available to the parent operating system
+implies that this may have been desirable to someone at some level at one time
+in the past.
+ > 1. What is the reason for the Kubernetes test? Is it okay to just skip or
+ >    disable the test in OpenShift?
```
jsafrane, line 132: Is it [Mount propagation](https://github.com/kubernetes/kubernetes/blob/82ebcd171909c4cd6007007fc07b025d0950153d/test/e2e/node/mount_propagation.go#L79-L79) test? I wrote the test, together with mount propagation support in Kubernetes.

Documentation says:

```
    // MountPropagationBidirectional means that the volume in a container will
    // receive new mounts from the host or other containers, and its own mounts
    // will be propagated from the container to the host or other containers.
    // Note that this mode is recursively applied to all mounts in the volume
    // ("rshared" in Linux terminology).
```

It's not specified there what the "host" is. The test defines it as something that's available under `nsenter  --mount=/rootfs/proc/1/ns/mnt ... ` and that's what Kubernetes has guaranteed so far. Changing this behavior may have consequences that we do not know about. I mentioned CSI drivers above, but we never know what people use mount propagation for.

lack: Thank you for that information!

Yes, that is the test I was concerned about because it implied some kind of relationship between that host mount namespace and the container's.  Basically instead of being under /rootfs/proc/1/ns/mnt, kubelet and the container runtime would be under a slightly different namespace, on level below that one.

So yes, this proposal would intentionally change that guarantee, but not all of it.

What would stay the same:
- The volume in a container will receive new mounts from other containers
- Its own mounts will be propagated from the container to other containers

What I'm pretty sure we can keep though I have to double-check if my current proof-of-concept does this today:
- The volume in a container will receive new mounts from the host

What would necessarily change:
- Its own mounts would no longer be propagated up to the host

I'll have to do some further experiments with my proof-of-concept to be sure, but I suspect the CSI case you mentioned should be okay, since it's in a container and therefore already under the lower-level umbrella namespace, thus will continue to interoperate fine with any other containers also rooted in that namespace.

enhancements/hide_container_mountpoints.md
lack: Good question. I'll double-check this!

enhancements/hide_container_mountpoints.md
jsafrane: > What would necessarily change:
> * Its own mounts would no longer be propagated up to the host

As you wrote, CSI drivers might be fine with that - we can test the ones that we ship and we could give 3rd party vendors some time to test theirs.

Still, it's a significant a change from upstream. Everyone who uses mount propagation for anything else than `/var/lib/kubelet`  will be broken. cc @openshift/openshift-architects

smarterclayton: > Its own mounts would no longer be propagated up to the host

Give me some hypothetical examples of things that go into mounts that admins on the host wouldn't be able to see?

* Mounted cloud disks?
* Empty dirs?
* Host level containers that create a directory in a host dir? such as an agent that creates logs on the host?

The first two if true would be frustrating but not end of world for admins.  The last one would be a blocker if true.  If the last one is true, is there a way to a) identify workloads where sharing with the host is important and b) separate those pods from regular workload pods when mounting?

enhancements/hide_container_mountpoints.md
lack: I hope I've done this right, but when I run a pod with a `/var/lib/kubelet` HostPath, the container is in its own mount namespace, but since this is inherited from the container runtime namespace (which is identical to the kubelet mount namespace), the pod does indeed see all the mounts under `/var/lib/kubelet`.

Do you have any pointers to a simple CSI pod I can play with to do some deeper testing of this?  Or is there a specific e2e suite I can run that specifically exercises the CSI architecture?

enhancements/hide_container_mountpoints.md
yuvalk: I dont think that a big problem
1. we can document for admins how to `nsenter` to the right namespace to see the mounts
2. and pod created by crio would be in the same mount namespace, so will see the mounts (incl. oc debug, etc)
3. If in the future we'll ever want to start each pod in its own mount namespace, we can add a privileged flag to inherit crio namespace, so privileged pods would still be able to see that

lack: I have expanded the `Risks and Mitigations` section in the proposal to be clearer about what's changing and what's not.

@smarterclayton : The answer to your specific questions depend a lot on exactly where these directories you're mentioning are created, and where the underlying filesystems are mounted.
- If the cloud disks are mounted by the Host OS, administrators and containers will see them.  If they're mounted by containers, administrators will not see them unless they enter the right mount namespace.
- The visibility of empty dirs or files to administrators will depend entirely upon the namespace of the underlying mountpoint. For example, if /var/lib is mounted in a Host OS, and a CRI-O creates /var/lib/foo and then mounts a filesystem at /var/lib/foo and creates /var/lib/foo/bar inside it, an administrator would see /var/lib/foo as an empty directory until they enter the right namespace at which point they would see that /var/lib/foo has a filesystem mounted and have access to the contents.
- If I'm understanding this correctly, the 'oc debug' container counts as a host-level container, I think.  As in the previous example, if /var/lib is a filesystem mounted by systemd, if a container creates a file or directory beneath /var/lib it will be visible to administrators and the Host OS, unless that container (or another container) has mounted an additional filesystem beneath that mountpoint.  If that's the case, anything below that mountpoint will not be visible to the administrator unless they enter the right mount namespace. Alternatively, the host level container could run `nsenter -m -t 1` to join the top-level systemd mount namespace, at which point any mounts it create would be visible to the host (and CRI-O and Kubelet and other containers), but it would lose visibility of anything mounted exclusively within that lower mount namespace until it exits the top-level namespace.

Does that help explain the changes and limitations a bit more clearly?

enhancements/hide_container_mountpoints.md
markmc: There's still only reviewers listed here, and no approvers

@haircommander @mrunalp @umohnani8 maybe one of you could agree to take point as final approver, and give your sense of where things stand so far? Thanks

lack: I have now added @haircommander @mrunalp and @umohnani8 as approvers, happy to add/remove any of them as required :)

enhancements/hide_container_mountpoints.md
lack: @jsafrane : I have a potential update to the "Mount propagation" test [here](https://github.com/openshift/kubernetes/compare/master...lack:hide_container_mountpoints) that may also add some clarity about exactly what the changes here are.

enhancements/hide_container_mountpoints.md
lack: For a proof-of-concept, I have changed the e2e test so that when it notices a segregated container mount namespace it keeps most tests the same, but then diverges when checking about the host's top-level namespace.

enhancements/hide_container_mountpoints.md
jsafrane: IMO you should write an enhancement upstream about this change and get approval there. While it may seem to be a simple and innocent change to you, it may break some crazy users.

Quick search in github.com/openshift revealed these users of mount propagation outside of /var/lib/kubelet:
* https://github.com/openshift/cucushift/blob/5925f0c1fcb021d98b78a24dcc2598072f0f7404/testdata/storage/hostpath/propashare.yaml
* https://github.com/openshift/special-resource-operator/blob/60770c14fea0a37a150c5aacf1ddb77d40869ebe/config/recipes/nvidia-gpu/manifests/1000-state-driver.yaml
* https://github.com/openshift/special-resource-https://github.com/embercsi/ember-csi/blob/6a0f5d41564ab6cb5a699cdf914c6c374d9b69cc/examples/kubernetes/kubeyml/node.yml#L105

Search over whole github, I got 3,185 hits in yaml files. Lot of them are CSI drivers using /var/lib/kubelet. But some of them mount `/dev`, `/run`, `/sys` and other directories.

enhancements/hide_container_mountpoints.md
zvonkok: @giuseppe Could this be related to the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1903553 how systemd handles mount points?

enhancements/hide_container_mountpoints.md
lack: I think it would depend a lot on the purpose of each `mountPropagation: Bidirectional`, even if it's dealing with areas outside of `/var/lib/kubelet`.  If the purpose of the `rshared` propagation is to mount a filesystem that's then visible to other pods, this will continue to work as before (even if it's acting in `/run` or `/sys`).  If the purpose is that systemd or something outside of CRI-O/Kubelet needs access to those pod-created mountpoints, that would indeed be broken by this proposal.  

Hard to know which is which without evaluating all of the ~3000 usages. So I take your point about needing more input on this to flush out any places where it may indeed be the case of a host process needing access to a container-mounted filesystem.

enhancements/hide_container_mountpoints.md
lack: @zvonkok : Can you talk to me a bit more about the need for the bidirectional mount in `/run/nvidia/driver`?  I get the sense that this is mounted by a container and intended to be seen outside that container.  Is the contents needed by other containers?  Some piece of software running on the host outside of the container runtime?

giuseppe: @zvonkok hiding such mounts from systemd could probably help in the issue you are seeing

zvonkok: @lack The situation I have is a Pod running privileged + rbind + systemd in the container. Now if I delete the container systemd is stopping all services in the container but also on the node and the node is rendered NotReady, only a reboot can "save" it. If I am using the Pod without the rbind meaning privileged + systemd it works as expected. Looks like the rbind is introducing something that throws off systemd. I will test your POC on my cluster and see if it helps or not.

lack: Ah, that's interesting... if the cause is an instance of systemd beneath the container runtime namespace that gets over-eager because it sees state from the top-level systemd, my proposal may not be the help you need; I think the rbind will still see the state that's properly owned by the parent systemd, so the container systemd would still do too much.

If the cause is the top-level systemd seeing state it shouldn't from mountpoints owned by the container runtime or the container itself, then my fix might be helpful, since sytsemd won't see anything under mountpoints created by the container runtime or containers themselves.

enhancements/hide_container_mountpoints.md
lack: After some conversation with @zvonkok, in the case of the special-resource-operator, the rshared mount of /run is specifically so that OCI hooks (like prestart) can mount things under /run that are then consumed by other containers.  And because OCI hooks run inside the container runtime namespace (aka, a process somewhere within the process tree of CRI-O), within my proposal the mount will end up in the lower-level namespace and still be visible to all containers. It seems based on my understanding so far that the host doesn't actually need access to the mountpoints, as long as the OCI prestart hook and containers have common access to them.

lack: I've updated the proposal to explicitly mention that OCI hooks will run inside this container-mount-namespace.

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,273 @@
...
+container-specific mountpoints into a private namespace reduced systemd
+overhead with no difference in functionality.
+
+## Motivation
+
+Systemd scans and re-scans mountpoints many times, adding a lot to the CPU
+utilization of systemd and overall overhead of the host OS running OpenShift.
+Changing systemd to reduce its scanning overhead is tracked in [BZ
+1819868](https://bugzilla.redhat.com/show_bug.cgi?id=1819868), but we can work
+around this exclusively within OpenShift. Using a separate mountpoint to for
```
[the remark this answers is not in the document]
lack: Nice catch, thank you!

enhancements/hide_container_mountpoints.md
lack: I have also just posted a summary of this proposal to a few mailing lists to see if there are any specific use-cases or solutions people know about that this proposal would interfere with.

enhancements/hide_container_mountpoints.md
lack: @jsafrane : Coming back to your original comment regarding `systemd-run`, if kubelet uses `systemd-run` to mount a filesystem, that mount will end up happening in the top-level systemd mount namespace, which will then be propagated down and seen by CRI-O/Kubelet/etc because their namespace is a 'slave' to that top-level mount namespace.  So this proposal doesn't interfere with that working, and in fact that method of mounting a filesystem circumvents this proposal's mechanism and makes such mounts visibile to systemd and the host OS again.

enhancements/hide_container_mountpoints.md
lack: @giuseppe : Can you give me a little more context or an example of how Podman/Skopeo/Buildah normally run in conjunction with OpenShift?  If they set up storage in the host mount namespace this will be propagated down to the container mount namespace (because it's a slave of the host mount namespace); so I'm unsure how much of a concern this is.

kolyshkin:
I very much like the overall idea, and the implementation seems to be decent, too.

It might (and will) break some setups in which people expect the container mounts to be visible on the host (which seems trivial to fix -- use the same nsenter wrapper as for k8s and cri-o). For this reason (of breakage) perhaps it is better to make this change in the upstream k8s and cri-o, rather than by using an operator.

enhancements/hide_container_mountpoints.md
giuseppe: I think a potential issue is with `/var/lib/containers/storage/overlay`, as it is always mounted `MS_PRIVATE` and it is not propagated to different namespaces.

giuseppe: this idea is very similar to rootless `podman unshare` that executes the specified command in a different mount (and user in this case) namespace where all the containers are running.  For rootless it is a necessity because it needs a user namespace to gain capabilities.

I think in future we could teach other container tools to check if there are images/containers mounted in a different mount namespace and try to join it, or at least raise a warning.

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,276 @@
...
+mountpoints" guarantee a deal-breaker?
+
+The implementation could take a few different forms. The proof-of-concept work
+that preceded this work was a set of
+[MCO](https://github.com/openshift/enhancements/pull/github.com/openshift/machine-config-operator)
+changes. This could be carried forward into the new implementation, or we could
+approach it another way:
+ > 1. Should these services and overrides be installed by MCO?
+ > 2. If so, should they be part of the main `00-worker` profile, or in a
+ >    separate `02-worker-container-mount-namespace` object?
```
umohnani8, line 187: For the MCO part, this would depend on whether this is going to be the default behavior for OpenShift. If it is, then we should use templates for the various service files and it will be a part of the `00-worker` MC. If we are giving the user a knob to be able to enable/disable this then it should be a separate MC like `02-worker-container-mount-namespace`.

lack: I would suggest that it should be enabled by default but can be disabled (in case there are use-cases where this mount hiding is undesirable).

In the MCO world, what's the best way to have a feature toggle for something like this?  Is it enough for MCO to create this `02-worker-container-mount-namespace` object by default that an admin could manually delete?

lack:
> I very much like the overall idea, and the implementation seems to be decent, too.
> 
> It might (and will) break some setups in which people expect the container mounts to be visible on the host (which seems trivial to fix -- use the same nsenter wrapper as for k8s and cri-o). For this reason (of breakage) perhaps it is better to make this change in the upstream k8s and cri-o, rather than by using an operator.

@kolyshkin: Thanks for the comments; and I have two questions for you here:
- What setups do you know about where people expect the container mounts to be visible on the host?  I would expect that this would be extremely uncommon and an anti-pattern in OpenShift (though perhaps less so in upstream k8s)
- Without a centralized top-level installer or configuration management system like MCO, how do you think we would package something like this, which requires coordination between the init system, kubelet and whatever container runtime a user chooses for k8s?

enhancements/hide_container_mountpoints.md
lack: @giuseppe : When you mention `/var/lib/containers/storage/overlay`, I'd love to learn more about that filesystem and how it's used and shared between various tools, especially since it's mounted `MS_PRIVATE`.

If you don't mind taking the time to explain, I'd like to understand more about which process mounts the filesystem there, and which other processes expect to be able to act on it.

giuseppe: any containers tool based on containers/storage (so all of CRI-O, Podman, Skopeo, Buildah).

The first access to the storage makes sure that `/var/lib/containers/storage/$DRIVER` is a mount point and it is made `MS_PRIVATE`.  All containers and images are mounted under this directory.  There is also some cleanup performed when there are no more containers and images mounted.

The reason for `MS_PRIVATE` is to prevent that each mountpoint will be propagated to the mount namespaces for each newly created container and avoid the cost of unmounting them.

enhancements/hide_container_mountpoints.md
mrunalp: @lack I agree with @jsafrane re: opening an enhancement upstream. We can then discuss it in sigs and mailing list to make sure we aren't breaking any use cases unintentionally.

enhancements/hide_container_mountpoints.md
```diff
@@ -0,0 +1,276 @@
...
+  cleanup its pinned namespace but instead idempotently re-use a
+  previously-created namespace.  However, given that the namespace service as
+  implemented today is a systemd `oneshot` with no actual process that needs
+  keeping alive, the risk of this terminating unexpectedly is very low.
+
+## Alternatives
+
+- Enhance systemd to support unsharing namespaces at the slice level, then put
+  `crio.service` and `kubelet.service` in the same slice
+- Alter both CRI-O and Kubelet executables to take a mount namespace via
```
mrunalp, line 273: I prefer that for CRI-O we support joining the mount namespace as an option. We should be able to make it work with an init in C similar to runc.

mrunalp (changes requested):
I think that if we do proceed with this once we know that we don't have edge cases we worry about it, it should still be treated as something that we back out once kernel / systemd have the fixes. Could that be captured in the enhancement?

enhancements/hide_container_mountpoints.md
umohnani8: Yes, the MCO can create the `02-worker-container-mount-namespace` MC and an admin could manually delete it. But once deleted, this MC will not be created unless the admin decides to manually recreate it again. If you want the MCO to automatically create this and delete it based on a toggle, then we will have to add logic to one of the controllers in the MCO to watch for the toggle and create/delete accordingly.

enhancements/hide_container_mountpoints.md
lack: Understood!  I am now taking this discussion to sig-node, sig-storage, and the upstream community for comments.

lack:
@mrunalp : I am not sure if making this temporary is necessarily the right option.  Once systemd/kernel are fixed and this isn't for CPU utilization purposes, I think there's a good case to be made for both tidiness/encapsulation reasons (ie, the mounts owned by OpenShift shouldn't be needed by the host) and security (if unprivileged users aren't allowed to see the container-specific mountpoint locations, this removes one source of potentially damaging knowledge about container contents).

That said, I do think it would make sense for this to be optional within OpenShift, perhaps via some trigger in MCO which would be on by default, but could be turned off to restore the older system behavior at the behest of an administrator.  Would this be a reasonable compromise to your request?

enhancements/hide_container_mountpoints.md
lack: The benefit to sticking with the 'nsenter' mechanism is that it would be the same for Kubelet, CRI-O, and easily adaptable to any other container runtime, but I am happy to consider natively supporting mount-namespace-entering for CRI-O as well.  It's just more work.

enhancements/hide_container_mountpoints.md
lack: That's great!  Do you have any suggestion of which controller in MCO would be best suited to have this toggle?  I'd be happy to try my hand at coding it up.

mrunalp:
> That said, I do think it would make sense for this to be optional within OpenShift, perhaps via some trigger in MCO which would be on by default, but could be turned off to restore the older system behavior at the behest of an administrator. Would this be a reasonable compromise to your request?

That sounds good to have an escape hatch if needed :+1:

lack:
I will be reviving this shortly.

I've been working with the upstream sig-node team to get make sure this idea doesn't break any present or future assumptions in kubernetes upstream.

lack:
So it turns out this proposed mechanism likely prevents containers (even privileged containers) from exploiting CVE-2021-33910!

lack:
/reopen

I'm reviving this given the request from customers for this feature, as well as some time spend on revamping the proposal to be a bit clearer.

PR #597, opened 2021-01-21
Add groupPrefix field to cloudwatch output type

alanconway:
User-defined prefix for all log group names.
Use to avoid name clashes when multiple clusters forward to a single CloudWatch instance.


alanconway:
/cc @jcantrill 
/cc @igor-karpukhin 
/assign @jcantrill

Added a simple groupPrefix field - default "" (no prefix), can be set by user or by the add-on to any value (including cluster name). No built-in support for picking cluster name, cluster UUID or any other string - but user is free to chose any of those or something else.

jcantrill (commented):
/lgtm

PR #601, opened 2021-01-22
continue working on markdownlint

dhellmann:
This PR includes a few fixes for files that were merged before the CI
job was set up and a change to base image in the Dockerfile.

### DIFF
--- a/hack/Dockerfile.markdownlint
+++ b/hack/Dockerfile.markdownlint
@@ -1,4 +1,4 @@
-FROM centos
+FROM fedora:29
 WORKDIR /workdir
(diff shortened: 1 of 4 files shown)


hack/Dockerfile.markdownlint
```diff
@@ -1,4 +1,4 @@
-FROM centos
+FROM fedora:29
```
russellb, line 1: not ideal since fedora 29 isn't maintained anymore, but I know you said you saw this used elsewhere.  We should come back around and change it again at some point though ...

dhellmann: Yeah, I wanted 33, but that wasn't available in CI and I thought it would be better to be consistent.

PR #603, opened 2021-01-25
Initial proposal of allow mtu and overlay port changes

juanluisvaladas:
For [SDN-1365](https://issues.redhat.com/browse/SDN-1365). Proposes changes to CNO to make it possible for customers to change the MTU, geneve port, and vxlan port in both OVN Kubernetes, and openshift SDN.

/cc @mccv1r0 
For MTU changes
/cc @trozet 
For OVN kubernetes
/cc @danwinship @knobunc @abhat 
For general thoughts

### DIFF
--- a/enhancements/network/allow-mtu-and-overlay-port-changes.md
+++ b/enhancements/network/allow-mtu-and-overlay-port-changes.md
@@ -3,18 +3,20 @@ title: allow-mtu-and-overlay-port-changes
 authors:
   - "@juanluisvaladas"
 reviewers:
-  - TBD
+  - "@danwinship"
+  - "@mccv1r0"
+  - "@russellb"
 approvers:
   - TBD
 creation-date: 2021-01-25
-last-updated: 2021-01-25
+last-updated: 2022-02-012
 status: provisional
 ---
 
 # Allow MTU and overlay port changes
 
 This covers adding the capability to the cluster network operator of changing
-the MTU, and the vxlan and geneve ports post installation.
+the MTU, and the VXLAN and Geneve ports post installation.
 
 ## Release Signoff Checklist
 
@@ -27,32 +29,35 @@ the MTU, and the vxlan and geneve ports post installation.
 
 ## Summary
 
-Customers may need to change the MTU, or the ports used for vxlan or geneve
+Customers may need to change the MTU, or the ports used for VXLAN or Geneve
 tunnels post-installation. However these changes aren't trivial and may cause
 downtime, hence the CNO forbids currently them.
 
-We propose a brand new daemonset that will run on every node of the cluster and
-will make the necessary changes in an ordered, and coordinated manner with a
-service disruption within the TCP timeout.
+We propose a brand new daemonset that will be launched on demand and will
+run on every node of the cluster. This daemonset will make the necessary
+changes in an ordered, and coordinated manner with a service disruption
+within the TCP timeout (best effort).
 
 ## Motivation
 
-While customers usually set the MTU, and the overlay ports some times they
-need to change them afterwards for unpredictable reasons such as performance,
-or changes in the underlay.
+While cluster administrators usually set the MTU, and the overlay ports
+correctly during the installation sometimes they need to change them
+afterwards for reasons such as changes in the underlay, or because they were
+set incorrectly at install time.
 
 ## Goals
 
-* Allow to change MTU post install on both OpenShift SDN and OVN Kubernetes.
+* Allow to change MTU of the tunnels post install on both OpenShift SDN and
+  OVN Kubernetes.
 
 * Allow to change the overlay network ports on the underlay in both OpenShift
   SDN and OVN Kubernetes.
 
-* Allow to change both vxlan and geneve ports in a single operation.
+* Allow to change both VXLAN and Geneve ports porst install.
 
 ## Non goals
 
-* Allow to change MTU configuration in te underlay interfaces.
+* Allow to change MTU configuration in the underlay interfaces.
 
 * Allow to change MTU, and either port in a single operation.
 
@@ -60,68 +65,67 @@ or changes in the underlay.
 
 ## Proposal
 
-When the cluster network operator detects the change it will deploy a new
-deaemonset with restartPolicy Mever which is responsible for doing the changes.
-
 Even though both changing the MTU, and changing the port have a lot in common,
 there are some key differences and therefore the daemonset will be slightly
 different.
 
-When decreasing the MTU each of the deamonset will:
+When the cluster network operator detects the change it will:
+1. Set the `clusteroperator/network` conditions:
+   - Progressing: true
+   - Upgradeable: false
+2. Deploy a deaemonset with `restartPolicy: Mever` which is responsible for
+   validating the preconditions. If the preconditions are met the pod will
+   exit with code 0. Some of the preconditions that we will check are:
+   - Chrony is synchronized
+   - For MTU changes, the new MTU is valid to apply on that node.
+   - For Port changes, the new port is not being used by other process
+3. Once all the pods in the daemonset are finished, if any of the pods has
+   exited with a code different than 0, the CNO will set the
+   `clusteroperator/network` conditions to:
+   - Progressing: false
+   - Degraded: true
+   At this point we require manual intervention.
 
-1. Verify that the new MTU is valid for that node. If it is valid add to
-   itself the annotation `openshift.io/cno-able-to-change-mtu`.
-2. Wait until the timestamp defined in `CHANGE_MTU_INIT_TS`
-3. Verify that every pod of that daemonset is running and has the annotation
-   `openshift.io/cno-able-to-change-mtu` if any of the pods is not running, or
-   is missing that annotation the pod will abort the procedure.
-4. Cordon the node where its running.
-5. Wait a grace period. This grace period is based on
+Once the preconditions are met the steps to change the MTU and the ports are
+different, for MTU changes the CNO will:
+1. Cordon every node, we don't want pods created during the process.
+2. Deploy a new daemonset that will run on every node which will apply the
+   changes that must be done manually in that node in particular.
+3. So that we don't have nodes doing things a t different times and we have
+   everything synchronized, the pods will wait until a timestamp defined in
+   the environment variable `CHANGE_MTU_INIT_TS`, which is defined by the CNO.
+4. The pod deployed by the daemonset, will enter the network namespace of
+   every pod in the pod network and change the MTU of `eth0`.
+5. The pod will wait a grace period. This grace period is based on
    `CHANGE_MTU_INIT_TS` rather than the time of finalization of the
    previous step.
-6. Proceed to change the MTU of `eth0` on each one of the pods netnamespace.
-7. Wait a grace period. This grace period is based on
+6. The pod will change the MTU of the end of the veth pair in the virtual
+   switch.
+7. The pod will wait a grace period. This grace period is based on
    `CHANGE_MTU_INIT_TS` rather than the time of finalization of the
    previous step.
-8. Proceed to change the MTU of each pod in the virtual switch.
-9. Wait a grace period. This grace period is based on
-   `CHANGE_MTU_INIT_TS` rather than the time of finalization of the
-   previous step.
-10. In the case of OVN Kubernetes proceed to change the MTU of the interfaces
+8. For OVN Kubernetes the pod will change the MTU of the interfaces
    ovn-k8s-mp0, br-local, and ovn-k8s-gw0 sequentially without any delay in
    between.
+   For OpenShift SDN the pod will change the MTU of the interface tun0.
+9. If all these steps failed, the pod will exit with code 1, if all were
+   successful it will exit with code 0.
 
-   In the case of OpenShiftSDN proceed to change the MTU of the interfaces
-   vxlansys.
-
-13. Uncordon the node
-
-14. Annotate itself with `openshift.io/cno-succesuflly-changed-mtu` and exit
-    with code 0.
-
-When increasing the MTU the same process applies, except steps 10, 8, and 6 are done in reverse order.
-
-Changing the overlay ports is a similar procedure, we will depoy a daemonset
-where each pod will:
-
-1. Verify that the port is not being used at the moment by anything else and
-   add to itself the annotation `openshift.io/cno-able-to-change-port`
-2. Wait until the timestamp defined in `CHANGE_PORT_INIT_TS`
-3. Verify that every pod of that daemonset is running and has the annotation
-   `openshift.io/cno-able-to-change-port` if any of the pods is not running, or
-   is missing that annotation the pod will abort the procedure.
-4. Cordon the node where its running.
-5. Wait a grace period. This grace period is based on
-   `CHANGE_PORT_INIT_TS` rather than the time of finalization of the
-   previous step.
-6. Actually change the port:
+For port changes the CNO will:
+1. Cordon every node, we don't want pods created during the process.
+2. Deploy a new daemonset that will run on every node which will apply the
+   changes that must be done manually in that node in particular.
+3. So that we don't have nodes doing things a t different times and we have
+   everything synchronized, the pods will wait until a timestamp defined in
+   the environment variable `CHANGE_PORT_INIT_TS`, which is defined by the CNO.
+4. Actually change the port:
    For OpenShift SDN we'll delete the port:
 
    `ovs-vsctl --if-exists del-port br0 vxlan0`
 
    And create it again with no delay at all:
 
-   ```
+   ```sh
    ovs-vsctl add-port br0 vxlan0 \
      -- set Interface vxlan0 ofport_request=1 \
      type=vxlan options:remote_ip="flow" \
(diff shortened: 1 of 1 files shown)


enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+9. Wait a grace period. This grace period is based on
+   `CHANGE_MTU_INIT_TS` rather than the time of finalization of the
+   previous step.
+10. In the case of OVN Kubernetes proceed to change the MTU of the interfaces
+   ovn-k8s-mp0, br-local, and ovn-k8s-gw0 sequentially without any delay in
+   between.
+
+   In the case of OpenShiftSDN proceed to change the MTU of the interfaces
+   vxlansys.
+
```
mccv1r0, line 96: Do you mean e.g. `vxlan_sys_4789` ?  If so, we probably want to leave the vxlan and geneve interface MTU at 65000

```
4: genev_sys_6081: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 65000 qdisc noqueue master ovs-system state UNKNOWN group default qlen 1000
    link/ether 7e:a4:52:23:c4:e8 brd ff:ff:ff:ff:ff:ff
```
```
5: vxlan_sys_4789: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 65000 qdisc noqueue master ovs-system state UNKNOWN group default qlen 1000
    link/ether 2a:6f:6d:e9:0d:20 brd ff:ff:ff:ff:ff:ff
    inet6 fe80::286f:6dff:fee9:d20/64 scope link
       valid_lft forever preferred_lft forever
```

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+13. Uncordon the node
+
+14. Annotate itself with `openshift.io/cno-succesuflly-changed-mtu` and exit
+    with code 0.
+
+When increasing the MTU the same process applies, except steps 10, 8, and 6 are done in reverse order.
+
```
mccv1r0, line 103: When Increasing we should be able to just set the MTU values larger.  Any existing TCP (or SCTP) connection will not be able to make use of the larger MSS at all.  The active MSS used  has long ago been negotiated when the 3WHS established the TCP connection, or been since set via an ICMP triggered along the path.   Likewise for UDP.  Either the new larger MTU is used for the next transmit, or the old one (e.g. if cached) is used, and will still work.

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Customers may need to change the MTU, or the ports used for vxlan or geneve
+tunnels post-installation. However these changes aren't trivial and may cause
+downtime, hence the CNO forbids currently them.
+
+We propose a brand new daemonset that will run on every node of the cluster and
```
danwinship, line 34: maybe make it clearer that you're talking about a daemonset that would be launched on demand, not a daemonset that would run all the time

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+will make the necessary changes in an ordered, and coordinated manner with a
+service disruption within the TCP timeout.
+
+## Motivation
+
+While customers usually set the MTU, and the overlay ports some times they
```
danwinship, line 40: missing the word "correctly" before the comma?

(except really, customers usually _don't_ set them; they leave them to be defaulted/autodetected)

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+need to change them afterwards for unpredictable reasons such as performance,
+or changes in the underlay.
```
danwinship, line 42: I don't think anyone is like "oh, I have a 9000 MTU network but I'll just set the MTU to 1450 anyway... oops, now I have bad performance!". It's more like, they screwed up. Or they changed the underlay.

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Goals
+
+* Allow to change MTU post install on both OpenShift SDN and OVN Kubernetes.
```
danwinship, line 46: maybe clarify that you're only talking about tunnel MTU not underlay MTU here

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+
+* Allow to change the overlay network ports on the underlay in both OpenShift
+  SDN and OVN Kubernetes.
+
+* Allow to change both vxlan and geneve ports in a single operation.
```
danwinship, line 51: "On OVN Kubernetes, allow changing both the hybrid overlay network VXLAN port and the pod network Geneve port in a single operation"

(ok so I didn't complain about this anywhere else, but if you fixed the capitalization of "VXLAN" and "Geneve" everywhere, I wouldn't hate it.)

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Non goals
+
+* Allow to change MTU configuration in te underlay interfaces.
```
danwinship, line 55:
```suggestion
* Allow to change MTU configuration in the underlay interfaces.
```

also, this means we need to clearly document what the administrator needs to do here. (ie, fix the underlay MTU _first_ when increasing the MTU, fix it _after_ when decreasing the MTU)

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+
+* Allow to change MTU, and either port in a single operation.
+
+* Other safe or unsafe configuration changes.
+
+## Proposal
+
+When the cluster network operator detects the change it will deploy a new
+deaemonset with restartPolicy Mever which is responsible for doing the changes.
```
danwinship, line 64:
```suggestion
daemonset with `restartPolicy: Never` which is responsible for doing the changes.
```

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+
+Even though both changing the MTU, and changing the port have a lot in common,
+there are some key differences and therefore the daemonset will be slightly
+different.
+
+When decreasing the MTU each of the deamonset will:
+
+1. Verify that the new MTU is valid for that node. If it is valid add to
+   itself the annotation `openshift.io/cno-able-to-change-mtu`.
```
danwinship, line 73: This would require the DS to have a ServiceAccount... it would be easier to do this if the DS didn't need to interact with the apiserver at all.

One way to do that would be to have a `readinessProbe`, such that the pod doesn't become ready unless it validates the MTU (but also see below)

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+2. Wait until the timestamp defined in `CHANGE_MTU_INIT_TS`
+3. Verify that every pod of that daemonset is running and has the annotation
+   `openshift.io/cno-able-to-change-mtu` if any of the pods is not running, or
+   is missing that annotation the pod will abort the procedure.
```
danwinship, line 77: again, it would be better to have the CNO validate this, and kill the DS if any of the pods aren't ready...

although then you have race conditions. Another way to do this would be to launch the DS first to do the pre-checks, and have them all immediately exit with status 0 or 1, and then if they all exit successfully, launch the DS a second time with second args to do the change?

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+4. Cordon the node where its running.
```
danwinship, line 78: We _definitely_ do not want the DS to have cluster-admin power. At any rate, you're talking about cordoning the entire cluster anyway, which could be done more efficiently in a single step by the CNO.

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+5. Wait a grace period. This grace period is based on
+   `CHANGE_MTU_INIT_TS` rather than the time of finalization of the
+   previous step.
```
danwinship, line 81: explain the reasoning on the waits... I'm not sure they're all needed

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+6. Proceed to change the MTU of `eth0` on each one of the pods netnamespace.
```
danwinship, line 82:
```suggestion
6. Proceed to change the MTU of `eth0` on each one of the pod's netnamespaces.
```

enhancements/network/allow-mtu-and-overlay-port-changes.md
danwinship: `vxlan_sys_4789` is a virtual interface that we shouldn't ever touch directly. You can look at the existing code to see what interfaces we set the MTU on at startup. (I think it's `br0` and `tun0`)

enhancements/network/allow-mtu-and-overlay-port-changes.md
danwinship: existing connections won't notice the change, but new connections could; if we raised the pod MTU first, then a pod might try to open a new connection using the new MTU and then smash into the lower MTU on the bridge

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+   vxlansys.
+
+13. Uncordon the node
+
+14. Annotate itself with `openshift.io/cno-succesuflly-changed-mtu` and exit
+    with code 0.
+
+When increasing the MTU the same process applies, except steps 10, 8, and 6 are done in reverse order.
+
+Changing the overlay ports is a similar procedure, we will depoy a daemonset
```
danwinship, line 104:
```suggestion
Changing the overlay ports is a similar procedure, we will deploy a daemonset
```

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+where each pod will:
+
+1. Verify that the port is not being used at the moment by anything else and
+   add to itself the annotation `openshift.io/cno-able-to-change-port`
```
danwinship, line 108: (similar comments as above)

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+4. Cordon the node where its running.
+5. Wait a grace period. This grace period is based on
+   `CHANGE_PORT_INIT_TS` rather than the time of finalization of the
+   previous step.
+6. Actually change the port:
+   For OpenShift SDN we'll delete the port:
+
+   `ovs-vsctl --if-exists del-port br0 vxlan0`
+
+   And create it again with no delay at all:
```
danwinship, line 122: Is this procedure actually already tested or is this just a guess/placeholder? If the latter, you don't need to specify it in that much detail here.

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+   And if we need to change the geneve port we'll do the same thing with it.
+
+7. In the case of OpenShift SDN each node will uncordon iself.
+8. EAch pod will annotate the node where it's running with
+   `openshift.io/cno-succesuflly-changed-mtu` and exit
+
+Once the daemonsets are finished, for OVN Kubernetes the CNO will verify that
+the configmap ovnkube-config is synchronized, and will force a rollout of the
+ovnkube-master daemonset. After that it will also force a rollout of the
+ovnkube-node daemonset.
```
danwinship, line 145: and for openshift-sdn you need to update the `ClusterNetwork` object

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+
+If any of the nodes fails during the procedure we expect the administrator to
+manually reboot the nodes. We could reboot automatically, but given the risks
+involved, and that this change is a very uncommon intervention, we consider
+safer making the human responsible for it.
+
+### Test Plan
+For MTU changes start a TCP server and client in two different nodes
+communicating continuosuly. Decrease the MTU and increase it once its finished.
```
danwinship, line 154:
```suggestion
communicating continuously. Decrease the MTU and increase it once its finished.
```

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+The TCP connection must be alive after both changes.
```
danwinship, line 155: would also be good to have a process repeatedly opening new connections, transferring a bit, closing, and looping.

Also, maybe explicitly do something involving TLS and large certificates, since that's a common MTU failure mode...

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,165 @@
...
+
+For vxlan or geneve start a TCP server and client in two different nodes
+communicating continuously. Change both ports in OVN kuberentes and vxlan
```
danwinship, line 158:
```suggestion
communicating continuously. Change both ports in OVN Kubernetes and vxlan
```

enhancements/network/allow-mtu-and-overlay-port-changes.md
juanluisvaladas: I meant `vxlan_sys_4789`, anyway Dan's right and I removed this step as it's not necessary, I just realized it has a MTU of 65000 by default.

enhancements/network/allow-mtu-and-overlay-port-changes.md
juanluisvaladas: I tested this manually. Should I make it less detailed anyway?

enhancements/network/allow-mtu-and-overlay-port-changes.md
juanluisvaladas: I've decided to rewrite the whole test plan with your feedback, I think it now addresses all your thoughts.

juanluisvaladas:
Just fixed a problem with markdownlint, the actual content doesn't change at all.

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,185 @@
...
+run on every node of the cluster. This daemonset will make the necessary
+changes in an ordered, and coordinated manner with a service disruption
+within the TCP timeout (best effort).
+
+## Motivation
+
+While customers usually set the MTU correctly, and the overlay ports some
+times they need to change them afterwards for unpredictable reasons such as
+changes in the underlay or because they set it to a wrong value when they
+installed the cluster.
```
russellb, line 45: Cluster administrators sometimes need to change the MTU or VXLAN/Geneve ports post-installation
for reasons such as changes in the underlay network or because they were set incorrectly at install time.
```

File: enhancements/network/allow-mtu-and-overlay-port-changes.md [hunk 23]
```diff
@@ -0,0 +1,185 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Customers may need to change the MTU, or the ports used for VXLAN or Geneve
```
russellb, line 31: I'd suggest "Cluster administrators" instead of "Customers" to be a bit more direct about the target persona

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Goals
+
+* Allow to change MTU of the tunnels post install on both OpenShift SDN and
+  OVN Kubernetes.
+
+* Allow to change the overlay network ports on the underlay in both OpenShift
+  SDN and OVN Kubernetes.
+
+* Allow to change both VXLAN and Geneve ports in a single operation.
```
russellb, line 55: The wording of this goal is a bit confusing.  It sounds like both Geneve and VXLAN are used at the same time?

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,185 @@
...
+Once the daemonsets are finished, for OVN Kubernetes the CNO will verify that
+the configmap ovnkube-config is synchronized, and will force a rollout of the
+ovnkube-master daemonset. After that it will also force a rollout of the
+ovnkube-node daemonset. For OpenShift SDN we'll make sure the ClusterNetwork
+object is synchronized and we'll force a rollout of the sdn daemonset.
+
+If any of the nodes fails during the procedure we expect the administrator to
+manually reboot the nodes. We could reboot automatically, but given the risks
+involved, and that this change is a very uncommon intervention, we consider
+safer making the human responsible for it.
```
russellb, line 152: How will a cluster administrator know the status of their requested change?  How will they know that it has failed vs. still in progress?  How will they know a reboot is required, and of which nodes?

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Test Plan
```
russellb, line 154: Will this be in CI?  Can you clarify when and where this test plan will be executed?

enhancements/network/allow-mtu-and-overlay-port-changes.md
juanluisvaladas: My understanding is that they can, for hybrid overlay

russellb:
I'm good with the direction of this.  I'll leave it to someone from the SDN team to lgtm the details of the plan.

/approve

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,196 @@
...
+   between.
+   For OpenShift SDN the pod will change the MTU of the interface tun0.
+9. If all these steps failed, the pod will exit with code 1, if all were
+   successful it will exit with code 0.
+
+For port changes the CNO will:
+1. Cordon every node, we don't want pods created during the process.
+2. Deploy a new daemonset that will run on every node which will apply the
+   changes that must be done manually in that node in particular.
+3. So that we don't have nodes doing things a t different times and we have
```
kedark3, line 118: `s/things a t different/things at different/`

enhancements/network/allow-mtu-and-overlay-port-changes.md
kedark3: @juanluisvaladas I'd like to leave detailed steps where possible. Could be helpful for QE to understand the flow of the things. Right now it seems useful to understand what's being planned.

enhancements/network/allow-mtu-and-overlay-port-changes.md
```diff
@@ -0,0 +1,196 @@
...
+   established.
+3. An HTTPS server with a very large certificate, and multiple clients
+   doing a single HTTPS request. The acceptance criteria is TLS negotiation
+   suceeds and HTTPS request returns 200.
+
+The acceptance criteria for this test is that the long lived TCP connection is
+kept alive, and the short lived connections get established.
+
+Packet loss, TCP retransmissions, increased latency, and reduced bandwidth are
+considered acceptable while the chane is happening.
```
kedark3, line 179: `s/the chane/the change/`

kedark3 (commented):
Hi @juanluisvaladas thanks for providing detailed info on enhancements. Some minor spelling nitpicks. Hope you don't mind. Thanks.

kedark3:
Maybe a good idea to close this as a new one #730 is now opened.

russellb:
> Maybe a good idea to close this as a new one #730 is now opened.

can you clarify why a new enhancement was opened instead of updating this one?

russellb:
original author of this enhancement is no longer working on this

/close

PR #605, opened 2021-01-25
cluster-high-availability-mode-api: clarify use of "infrastructure"

dhellmann:
Make it clear that the `infrastructureTopology` applies to
infrastructure *services* rather than infrastructure *nodes*.


derekwaynecarr:
this improves understanding that its not tied to literal node role label, but supporting service for workload on the cluster.

/lgtm
/approve

PR #205, opened 2020-02-07
Add enhancement for collecting installer telemetry metrics.

patrickdillon:
Authored together with @rna-afk 

This is a first draft of a design for collecting telemeter metrics in the installer.

PR #609, opened 2021-01-27
tools: include standard comments in PR activity count

markmc:
Since every PR is also an issue, the standard comments in PR are available via the issue comments API. We were only counting "review comments", which are comments associated with a particular diff.

(And "review comments" are different from the comments that are associate with PR reviews. Not confusing.)

openshift/enhancements#205 is an example of PR with lots of standard comments

### DIFF
--- a/tools/stats/stats.go
+++ b/tools/stats/stats.go
@@ -17,16 +17,26 @@ type GithubClientSource func() *github.Client
 // PullRequestDetails includes the PullRequest and some supplementary
 // data
 type PullRequestDetails struct {
-	Pull                *github.PullRequest
-	Reviews             []*github.PullRequestReview
-	RecentReviewCount   int
-	Comments            []*github.PullRequestComment
-	RecentCommentCount  int
+	Pull *github.PullRequest
+
+	// These are groups of comments, submited with a review action
+	Reviews           []*github.PullRequestReview
+	RecentReviewCount int
+
+	// These are "review comments", associated with a diff
+	PullRequestComments  []*github.PullRequestComment
+	RecentPRCommentCount int
+
+	// PRs are also issues, so these are the standard comments
+	IssueComments           []*github.IssueComment
+	RecentIssueCommentCount int
+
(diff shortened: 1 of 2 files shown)


tools/stats/stats.go
```diff
@@ -120,7 +130,34 @@ func (s *Stats) IteratePullRequests() error {
 	return nil
 }
 
-func (s *Stats) getComments(pr *github.PullRequest) ([]*github.PullRequestComment, error) {
+func (s *Stats) getIssueComments(pr *github.PullRequest) ([]*github.IssueComment, error) {
+	c := s.clientSource()
+	ctx := context.Background()
+	opts := &github.IssueListCommentsOptions{
+		Since: &s.earliestDate,
```
markmc, line 137: This is what we do for `PullRequestComment` too, but it seems wrong - why query comments since this date if we want to have both recent/all counts?

markmc:
```
tools/README.md:17 MD040/fenced-code-language Fenced code blocks should have a language specified
```

Objection!

(fixed)

tools/stats/stats.go
dhellmann: Yeah, this does seem like a bug. Do you want to fix it in this PR or should we use another one?

dhellmann (approved):
I think the original version of the tool looked at both, and we decided to drop the regular comments as a way to cut down on the number of API calls. Now that we're making fewer calls in parallel, we're less likely to run into problems with throttling, so it does make sense to collect everything.

What do you think, @russellb ?

russellb:
> I think the original version of the tool looked at both, and we decided to drop the regular comments as a way to cut down on the number of API calls. Now that we're making fewer calls in parallel, we're less likely to run into problems with throttling, so it does make sense to collect everything.
> 
> What do you think, @russellb ?

Counting both sounds like an improvement if it doesn't blow up!

dhellmann:
Let's take this as-is and fix the query argument in a follow-up.

/lgtm

PR #610, opened 2021-01-27
tools: refactor to add subcommands

markmc:
This is a bunch of refactoring, to prepare for adding a tool to look at who has been actively merging enhancements.


markmc:
/hold

The mergers tool isn't very useful yet, but perhaps it's worth merging the refactoring separately.

tools/mergers/mergers.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	for _, c := range comments {
+		user := *c.GetUser().Login
+		scanner := bufio.NewScanner(strings.NewReader(c.GetBody()))
+		for scanner.Scan() {
+			line := scanner.Text()
+			if strings.HasPrefix(line, "/lgtm") {
+				s.lgtmByUser[user]++
+			} else if strings.HasPrefix(line, "/approve") {
```
dhellmann, line 86: This will miss approvals added via the GitHub UI instead of a comment, like I did on https://github.com/openshift/enhancements/pull/609

To catch those, you also need to look at the "state" field in the review comment. https://docs.github.com/en/rest/reference/pulls#list-reviews-for-a-pull-request

tools/util/query.go
```diff
@@ -0,0 +1,178 @@
...
+			break
+		}
+		opts.Page = response.NextPage
+	}
+
+	fmt.Fprintf(os.Stderr, "\n")
+
+	return nil
+}
+
```
dhellmann, line 94: The rest of the methods below this point are tied more to the client than the query being run, so they probably don't belong on this type long term. Maybe they can be standalone functions that take the client source and context as args? That can happen in another PR, since this one is already fairly large.

tools/cmd/report.go
```diff
@@ -19,13 +19,12 @@ func newReportCommand() *cobra.Command {
...
+			query := util.NewPullRequestQuery(
+				daysBack, staleMonths, orgName, repoName, devMode,
 				util.NewGithubClientSource(configSettings.Github.Token))
-			if err != nil {
-				return errors.Wrap(err, "Could not create stats")
-			}
 
-			err = theStats.IteratePullRequests()
+			theStats := stats.New(query)
+			err := query.IteratePullRequests(theStats.Process)
```
dhellmann, line 27: It feels wrong to have to pass the query to `stats.New()` and then pass the method to `IteratePullRequests`. Can't the stats type pull the data it needs from the query itself? Or can we avoid storing the query in the stats type?

tools/util/github.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/google/go-github/v32/github"
+	"golang.org/x/oauth2"
+)
+
+// GithubClientSource describes a function that can create a
+// github.Client without any other inputs.
+type GithubClientSource func() *github.Client
+
+func NewGithubClientSource(token string) GithubClientSource {
```
dhellmann, line 14: I honestly no longer remember why I didn't just instantiate the client and pass it to the thing that needs it, instead of using a factory like this. Do you think we can eliminate the extra layer?

tools/cmd/report.go
markmc: Good call. I've made the interface simply `theStats := stats.Generate(query)` now

tools/mergers/mergers.go
markmc: Yeah, thanks - I actually noted that when I saw what you did :)

tools/util/github.go
markmc: It seems to work fine without it, I'll remove

tools/util/query.go
markmc: Yeah, it's not brilliant, but I think of them kinda like accessors on the query result

markmc:
/unhold

Thanks for the review. I've dropped the new tool from this PR for now, so it can just be the refactoring

PR #613, opened 2021-01-28
NetworkPolicies for System Namespaces

danwinship:
For https://issues.redhat.com/browse/RFE-701; add a mode in which system services are protected by NetworkPolicies (in addition to the existing TLS certificate authentication), for "defense in depth".

### DIFF
--- a/enhancements/network/system-networkpolicies.md
+++ b/enhancements/network/system-networkpolicies.md
@@ -1,0 +1,398 @@
+---
+title: Network Policies for System Namespaces
+authors:
+  - "@danwinship"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2021-01-18
+last-updated: 2021-01-18
+status: provisional
+---
+
+# Network Policies for System Namespaces
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+An OCP cluster contains various services that should not be accessible
+to unprivileged users.
+
+Currently, authorization for these services is handled via TLS
+certificates; unauthorized users can _connect_ to the service, but
+they cannot (usefully) _authenticate_ to it, and so the service is
+protected.
+
+Some customers, especially in the financial sector, have objected to
+this on the grounds that attackers may be able to exploit bugs in the
+TLS implementation to gain access to these services (or the services'
+underlying credentials). They would prefer that untrusted users were
+prevented from reaching the services entirely. (eg, see [RFE-701])
+
+This enhancement proposes adding NetworkPolicies to an OCP cluster to
+(optionally) ensure that restricted services are not reachable except
+by the system services that are expected to access them. Each team
+would responsible for defining the policies for their own operators.
+
+[RFE-701]: https://issues.redhat.com/browse/RFE-701
+
+## Motivation
+
+### Goals
+
+- Create a global configuration option indicating that OCP operators
+  should restrict access to themselves.
+
+- When the option is enabled, have various OCP operators create
+  NetworkPolicies limiting access to their services to the appropriate
+  other components. (Alternatively: have operators always create the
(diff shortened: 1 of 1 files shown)


danwinship:
> For https://issues.redhat.com/browse/RFE-701

(There's no actual epic filed for this yet, but since it's going to take several releases to move all the pieces into place I wanted to start early...)

enhancements/network/system-networkpolicies.md
```diff
@@ -0,0 +1,386 @@
...
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+An OCP cluster contains various services that should not be accessible
+to unprivileged users.
+
+Currently, authorization for these services is handled via TLS
+certificates; unauthorized users can _connect_ to the service, but
+they cannot _authenticate_ to it, and so the service is protected.
```
squeed, line 32: if you want to be pedantic, they can _authenticate_, they're just not _authorized_ to access the resource.

squeed:
Hmm - one thought about how we might be able to incrementally enhance this: all new system namespaces need to have a network policy in them (which we can enforce with an e2e test). This would be regardless of "restricted mode".

enhancements/network/system-networkpolicies.md
```diff
@@ -0,0 +1,386 @@
...
+
+(The presence of `kube-system` there is probably a mistake, since
+there are no pods in `kube-system`...)
+
+These lists occasionally need to be adjusted. It would be good if we
+could find a way to figure these out automatically, either based on
+analyzing the NetworkPolicies that the operators create, or by having
+the operators annotate their namespaces with hints to openshift-sdn.
+
+### Risks and Mitigations
```
squeed, line 257: The biggest risk, though, is that we block something unexpected not covered by e2e tests. Something, for example, shipped via OLM.

enhancements/network/system-networkpolicies.md
```diff
@@ -0,0 +1,386 @@
...
+
+#### Dev Preview -> Tech Preview
+
+- The periodic e2e test that runs in fully-restrictive mode can now
+  pass. We add some presubmit jobs.
+
+- We update the documentation on the "restrictive mode" config option
+  to note that the feature is in Tech Preview and is expected to work,
+  but is not yet fully supported.
+
```
squeed, line 307: Perhaps: namespaces that are "obviously" not user-facing, but existed prior to the feature, _could_ have policies created.

enhancements/network/system-networkpolicies.md
```diff
@@ -0,0 +1,386 @@
...
+#### Tech Preview -> GA
+
+- More testing
+- Sufficient time for feedback
+
```
squeed, line 312: Does it make sense for all new clusters to be restricted when the feature reaches GA? I think so. Or, perhaps, there is a GA + Default step here.

enhancements/network/system-networkpolicies.md
```diff
@@ -0,0 +1,386 @@
...
+deny" or "allow to all" baseline policy. (See Appendix for gory
+details.)
+
+Of the two sub-options, 2b ("CNO creates permissive policies") seems
+nice because it means that individual operators wouldn't need to track
+the restricted-vs-open option; they'd just always configure themselves
+as restricted, and CNO would fix things up if they were supposed to be
+open.
+
+However, 2b would make upgrades more complicated, since we have to
```
squeed, line 205: Do we _really_ need to worry about this? The only connections that would be disrupted would be 

1. Connections we somehow missed when writing network polices, and 
2. Only between operator S upgrade and CNO upgrade.

enhancements/network/system-networkpolicies.md
danwinship: We don't do "old clusters have default X but new clusters have default Y". It's too confusing.

enhancements/network/system-networkpolicies.md
danwinship: I didn't mean "upgrades in general", I meant specifically the case of upgrading from a version of OCP that doesn't implement this feature, to the version of OCP that does, in a cluster where the administrator does not want to be using this feature, and has components of their own that we don't know about accessing random OpenShift components. In that case, during the upgrade, user workloads would be blocked from accessing OpenShift components, possibly creating outages.

If you're going to argue that that's not a real problem, then you're essentially arguing that we don't actually need to preserve the permissive option at all; we can just start switching all components to be restrictive with no way to override it.

PR #615, opened 2021-01-30
refactor report tool

dhellmann:
Make the stats package more general by defining a Bucket type with a
Rule to indicate which pull requests belong in the bucket. This opens
us up to moving the logic to organize the data and then print the report
to the same place, which makes it easier to change and extend.

/cc @russellb @markmc

### DIFF
--- a/tools/cmd/report.go
+++ b/tools/cmd/report.go
@@ -1,9 +1,13 @@
 package cmd
 
 import (
+	"fmt"
+	"os"
+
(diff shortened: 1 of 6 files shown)


markmc:
Looks like a nice abstraction, definitely an improvement. I'd have been inclined to move the bucket definitions into `tools/report`, and keep the reporting printing code there, keeping `tools/cmd` minimal ... but not a big deal

/approve
/lgtm

PR #616, opened 2021-01-30
add a debug command to the report tool

dhellmann:
For now the command only prints the group name based on the files in
the PR, but we can extend it to print more detailed information later.

/cc @russellb @markmc


markmc:
Cool. `show` or `dump` would have made a bit more sense to me. A `--debug` flag for debug logging is what I guessed this was about at first

/approve
/lgtm

PR #620, opened 2021-02-01
tools: show old and stale enhancements by default

dhellmann:
We discussed making the default report include more categories. This
moves "old" and "stale" into the default. That leaves only "idle" and
"revived", so I wonder if maybe we want to include those, too.

/cc @russellb

### DIFF
--- a/tools/cmd/report.go
+++ b/tools/cmd/report.go
@@ -17,7 +17,7 @@ import (
 func newReportCommand() *cobra.Command {
 	var (
 		daysBack, staleMonths int
-		devMode, full         bool
+		devMode               bool
 	)
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


dhellmann:
When I run the tool with `--full` it reports 54 idle PRs, which are less than 3 months old and have had no activity in the last 7 days. Only 3 are stale (older than 3 months with no activity in the last 7 months).

russellb:
> When I run the tool with `--full` it reports 54 idle PRs, which are less than 3 months old and have had no activity in the last 7 days. Only 3 are stale (older than 3 months with no activity in the last 7 months).

maybe for stale we should switch to watching for the `lifecycle/stale` label?  It might be confusing if we report on it but with a different definition for "stale".

Otherwise, I think it makes sense to turn it all on and let's see how it looks in the coming weeks.  It's easy to just cut stuff when we talk about it together on Fridays.

dhellmann:
@russellb I don't see any lifecycle labels available for PRs. Are those only allowed on issues in this repo?

dhellmann:
Oh, fun, it just looks like I don't have permission to apply lifecycle labels, but we do have some PRs with `lifecycle/stale`.

dhellmann:
I've updated this PR to remove the `--full` flag completely. #621 redefines the stale section of the report.

PR #621, opened 2021-02-01
tools: redefine stale in the report as having the lifecycle/stale label

dhellmann:
Instead of doing the calculation ourselves and potentially having
different results, use the existing label applied by the bot.

Related to #620

/cc @russellb


russellb:
@dhellmann has a merge conflict now

dhellmann:
@russellb rebased

PR #622, opened 2021-02-03
WIP: sts-by-default proposal

joelddiaz:
initial commit


joelddiaz:
will revisit this when we have more data on STS usage

PR #443, opened 2020-08-19
Support a provisioning token for the Machine Config Server

cgwalters:
See https://github.com/openshift/machine-config-operator/pull/784

The Ignition configuration can contain secrets, and we want to avoid it being accessible both inside and outside the cluster.

PR #626, opened 2021-02-04
WIP: enhancements/machine-config: securing MCS

crawford:
We're getting heat for blocking arbitrary ports on the pod network (TCP 22623/22624). This is one potential path forward and is an alternative to https://github.com/openshift/enhancements/pull/443.

### DIFF
--- a/enhancements/machine-config/securing-mcs.md
+++ b/enhancements/machine-config/securing-mcs.md
@@ -1,0 +1,112 @@
+---
+title: Securing the Machine Config Server
+authors:
+  - @crawford
+reviewers:
+  - @darkmuggle
+  - @kikisdeliveryservice
+  - @michaelgugino
+  - @danwinship
+  - @cgwalters
+  - @russellb
+  - @derekwaynecarr
+approvers:
+  - @darkmuggle
+  - @russellb
+creation-date: 2021-02-01
+last-updated: 2021-02-01
+status: provisional
+---
+
+# Securing the Machine Config Server
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The bootstrapping process for new cluster nodes is a tricky problem, primarily due to the fact that a new nodes start from a position of almost zero knowledge. These nodes need to fetch their machine configuration and authenticate themselves with the cluster before they can access most resources and have workloads scheduled. At the same time, a malicious actor pretending to be a new node cannot be allowed to join the cluster. Otherwise, they would be able to get access to sensitive resources and could have workloads and their secrets scheduled to them. Today, we protect against this by preventing cluster workloads from accessing the Machine Config Server, the component of the Machine Config Operator which is responsible for serving Ignition configs, derived from a set of Machine Configs, to new machines. This approach presents a problem though: workloads on OpenShift are not allowed to make use of TCP ports 22623 and 22624, the ports used by the Machine Config Server.
+
+## Motivation
+
+This blanket network policy prevents traffic destined for ports 22623 and 22624, regardless of the source and destination IP addresses. This has the potential to impact customer workloads, but also internal initiatives (e.g. OpenShift on OpenShift is complicated by this restriction).
+
+### Goals
+
+- Remove the port restriction on 22623 and 22624
+- Ensure that the node bootstrap flow remains secure
+
+### Non-Goals
+
+- Allow customers to use Machine Configs to provide pre-bootstrap configuration
+
+## Proposal
+
+If it can be ensured that no one is able to access sensitive data through the Machine Config Server, the port restrictions can be safely removed. In order to achieve this, Machine Configs can be split into two groups: pre-node-approved configs and post-node-approved configs. The pre-node-approved configs are ones which do not contain sensitive information and contain just enough to configure new nodes to submit a certificate signing request (CSR) to the API server. The post-node-approved configs contain everything else, including the overwhelming majority of customer-provided Machine Configs. When a new machine is created, it fetches its Ignition config from the Machine Config Server, which only serves the pre-node-approved configs. At this point, the machine has enough information to submit a CSR, but nothing sensitive (note that for the purposes of this proposal, the CSR bootstrap token and pull secret are not considered sensitive). Once the CSR has been approved and a kubeconfig returned, the Machine Config Daemon is able to directly enumerate Machine Configs, including the post-node-approved ones, so it then applies any remaining configuration to the machine and reboots if necessary. At this point, the machine is fully configured.
+
+Certain, exotic use cases may require control over which Machine Configs are considered pre-node-approved and post-node-approved. To facilitate those use cases, the Machine Config CRD will need to be updated to include a new boolean field: `neededBeforeNodeApproved`. Since this defaults to `false`, all customer-provided Machine Configs will be placed into the post-node-approved set after the cluster has been updated.
+
+### User Stories
+
+#### Story 1
+
+#### Story 2
+
+### Implementation Details/Notes/Constraints [optional]
+
+The Machine Config Operator will begin rendering two sets of Machine Configs: pre- and post-node-approved; the former being a strict subset of the latter. This means that the post-node-approved rendered Machine Configs will be identical to the rendered Machine Configs today. This is the set of rendered Machine Configs that the Machine Config Daemon will use to reconcile. The pre-node-approved Machine Configs, on the other hand, will only be used by Ignition and they will be to configure the machine to the point that the kubelet can successfully submit a CSR. Because these configs are served without any authentication, care must be taken to ensure that no sensitive information is contained within.
+
+### Risks and Mitigations
+
+- If we make this change to existing clusters, new nodes that are provisioned will no longer include configuration specified in customer-provided Machine Configs until after the CSR has been approved. There is a chance that clusters would fail to auto-scale after applying this update, but that should be very rare. In order for a machine to fetch its Ignition config, it needs to be able to communicate with the API server, specifically, the Machine Config Server. If that's possible, it should also be possible for the kubelet to communicate with that same host.
+
+- This effectively doubles the number of rendered Machine Configs that need to be managed. This might be a concern for clusters with a huge number of Machine Config Pools, but this also seems rare.
+
+- This will introduce a second reboot into the provisioning process in cases where there are post-node-approved Machine Configs. Unfortunately, this is likely to be the case in the environments which are most impacted by reboot times (i.e. bare metal). The second reboot comes from the fact that `machine-config-daemon-firstboot.service` needs to run before the kubelet can start, and this service reboots the machine because it updates the RHCOS installation. After that reboot, the kubelet starts, submits a CSR, and then after approval, the Machine Config Daemon reconciles the post-node-approved Machine Configs and then drains and reboots. We could potentially avoid the reboot by teaching `machine-config-daemon-firstboot.service` to live pivot the machine, allowing the kubelet get its kubeconfig, and then letting the Machine Config Daemon do its reconciliation pass before rebooting. It's not clear to me how feasible this would be.
+
+## Design Details
+
+### Open Questions
+
+1. Can we avoid the extra reboot when post-node-approved Machine Configs are used?
+
+### Test Plan
+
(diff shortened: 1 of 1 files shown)


enhancements/machine-config/securing-mcs.md
```diff
@@ -0,0 +1,112 @@
...
...sitive data through the Machine Config Server, the port restrictions can be safely removed. In order to achieve this, Machine Configs can be split into two groups: pre-node-approved configs and post-node-approved configs. The pre-node-approved configs are ones which do not contain sensitive information and contain just enough to configure new nodes to submit a certificate signing request (CSR) to the API server. The post-node-approved configs contain everything else, including the overwhelming majority of customer-provided Machine Configs. When a new machine is created, it fetches its Ignition config from the Machine Config Server, which only serves the pre-node-approved configs. At this point, the machine has enough information to submit a CSR, but nothing sensitive (note that for the purposes of this proposal, the CSR bootstrap token and pull secret are not considered sensitive). Once the CSR has been approved and a kubeconfig returned, the Machine Config Daemon is able to directly enumerate Machine Configs, including the post-node-approved ones, so it then applies any remaining configuration to the machine and reboots if necessary. At this point, the machine is fully configured.
```
russellb, line 51: I think it would be helpful to pull out some of the assumptions and state them very clearly up front.  It's included in here, but perhaps before this paragraph, a new paragraph could say:

Security assumptions:
* The CSR bootstrap token is not considered sensitive and being accessible to all applications in the cluster is accessible.
* The pull secret is not considered sensitive and being accessible to all applications in the cluster is accessible
* Contents of MachineConfig resources are considered (potentially) sensitive, so this proposal introduces a change to protect them from unauthorized access through the MCS

Do I have that right?

This proposal would have to depend on some other enhancement that would allow that first assumption to hold.  I believe right now you could DoS a cluster with the ability to issue CSRs.

danwinship, line 51: > At this point, the machine has enough information to submit a CSR, but nothing sensitive (note that for the purposes of this proposal, the CSR bootstrap token and pull secret are not considered sensitive). Once the CSR has been approved

Wait... we're saying that

  - the MCS cannot distinguish a real proto-node from an attacker, _but_
  - the certificate approver _can_ distinguish a CSR sent by a real proto-node that has received non-secret information from the MCS, from a CSR sent by an attacker that has received non-secret information from the MCS

That doesn't make sense; if the certificate approver can distinguish them, then the MCS ought to be able to distinguish them in the same way. Specifically, the MCS should only respond to connections from IPs that correspond to not-yet-approved `Machine` resources (or something like that). Right?

russellb: The logic there makes sense.  If we're willing to automatically approve a CSR, that same logic should be sufficient for accepting a connection to MCS.

However, the MCS doesn't have all the same information that the CSR approver has.  It uses the contents of the CSR itself as part of the validation, so we can't copy it easily, unless the MCS request includes more info than it does today.

For reference, see the README in https://github.com/openshift/cluster-machine-approver/ and most of the important logic is in `csr_check.go`

danwinship: > However, the MCS doesn't have all the same information that the CSR approver has. It uses the contents of the CSR itself as part of the validation, so we can't copy it easily, unless the MCS request includes more info than it does today.

But that other information doesn't matter. If the MCS can see that a request is coming from an IP that isn't the IP of a legitimate "proto-node", then it knows that the requester cannot possibly construct an approvable CSR (that points to itself), and so there is no reason for the MCS to talk to it.

russellb: Is the source IP good enough?  Let's say untrusted-pod-A on node1 is trying to reach the MCS on node2.  Won't that request be SNAT'd on node1, so node2 is going to see node1's IP address and consider it legitimate?

enhancements/machine-config/securing-mcs.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Risks and Mitigations
+
+- If we make this change to existing clusters, new nodes that are provisioned will no longer include configuration specified in customer-provided Machine Configs until after the CSR has been approved. There is a chance that clusters would fail to auto-scale after applying this update, but that should be very rare. In order for a machine to fetch its Ignition config, it needs to be able to communicate with the API server, specifically, the Machine Config Server. If that's possible, it should also be possible for the kubelet to communicate with that same host.
```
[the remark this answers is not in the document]
crawford: It's an open question in my mind, though I'd definitely prefer a solution that allows us to keep new clusters and existing clusters in sync. I'm not too worried about leaving the network filter in place on existing clusters. If it was going to be a problem, the customer would have noticed by now.

enhancements/machine-config/securing-mcs.md
crawford: > That doesn't make sense; if the certificate approver can distinguish them, then the MCS ought to be able to distinguish them in the same way.

The client for the MCS is Ignition, which has significantly less functionality than the client for the CSR endpoint, the kubelet. Eventually, the approver could be enhanced (by us or the customer) to make use of [AWS signed requests](https://docs.aws.amazon.com/general/latest/gr/signing_aws_api_requests.html) or [GCP instance identity verification](https://cloud.google.com/compute/docs/instances/verifying-instance-identity). Keeping the MCS out of the authentication path allows the customer to decide how much security they need for their deployment. If we did everything in the MCS, they wouldn't easily be able to swap out the implementation.

enhancements/machine-config/securing-mcs.md
```diff
@@ -0,0 +1,112 @@
...
+
+These have been enumerated in other parts of the proposal, but to get them all in one place:
+
+  - This will negatively impact the scale-up time of clusters which make use of post-node-approved Machine Configs.
+  - This has the potential to break a cluster's ability to scale up.
+
+## Alternatives
+
+  - Do nothing and just remove the network policy. This would expose the contents of (what would be) post-node-approved Machine Configs to everything that has network access to the control plane.
+  - Add support for an authentication token to the Machine Config Server (https://github.com/openshift/enhancements/pull/443). The real downside to this approach is that it makes troubleshooting more difficult, as it would likely need to happen in the initramfs of RHCOS. This is an extremely limited environment and requires console or serial access, which isn't available in a number of environments.
```
derekwaynecarr, line 111: is the experience different here versus when a user provides an invalid pull-secret at install?

smarterclayton: What is the difference between screwing up 

`https://some-ip/some/value/ignition/master

and 

`https://some-ip/some/value/ignition/master?token=xxxxxxxxxxxxxx`

You either cut and paste the secret correctly or not.  At some point we also have to verify the remote ip (if you ignore validation errors on some-ip you are going to get MITM'd anyway, so security is not your problem).

The person provisioning a box is ultimately responsible for establishing a verifiable, secure, and verified channel to the thing that will establish the basic trust foundation for the rest of that box.  

Why is troubleshooting the problem?  Is that the only argument against requiring a shared secret between provisioner and provisionee to establish mutual verification?

crawford: @derekwaynecarr if the user provides an invalid pull-secret at install, the bootstrap machine will spin, trying to pull the release image. Eventually, the install will time out and the installer will collect logs. From the perspective of a single compute node, the two scenarios are indistinguishable, but I would expect an admin to look at more than just that. For example, if they SSH into the bootstrap node and run the recommended `journalctl` invocation, it will be immediately obvious that the pull secret is incorrect.

crawford: @smarterclayton `https://some-ip/some/value/ignition/master` will never change. `https://some-ip/some/value/ignition/master?token=xxxxxxxxxxxxxx` will change regularly (specifically, the token). The concern around troubleshooting is the fact that it will invariably be required more often when the Ignition URL is being changed periodically. There is a deeper question in there as well though. How will this token be rotated? This is easy in deployments which make use of the Machine API Operator, but everywhere else, this will have to be a manual process. And as we all know, if there is a periodic, manual task required for security, we can just assume it's not going to happen. I truly believe that the auth token will hurt the security of our clusters, all things considered.

enhancements/machine-config/securing-mcs.md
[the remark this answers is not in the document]
crawford: I added this to the "Open Questions" section.

enhancements/machine-config/securing-mcs.md
danwinship: > Is the source IP good enough? Let's say untrusted-pod-A on node1 is trying to reach the MCS on node2. Won't that request be SNAT'd on node1, so node2 is going to see node1's IP address and consider it legitimate?

node1's address is not legitimate, because node1 is an _actual_ node, not a "proto-node". An actual functioning node has no reason to talk to the MCS (as I understand it); an IP should only be contacting the MCS if there is a `Machine` that claims that IP but no `Node` that claims that IP; ie, if the cluster is expecting a node to come up with that IP, but it knows that the node hasn't actually come up yet.

(Even if node1 gets accidentally nuked and you want to recreate it, then (AIUI), you would have to delete the old node1 Node/CSR first, since otherwise the new node1's new CSR would fail to match and the approver would consider it to be an imposter. So even for reinstalls, the rule that "only IPs that correspond to a Machine but not a Node should talk to the MCS" should work.)

> Eventually, the approver could be enhanced (by us or the customer) to make use of [AWS signed requests](https://docs.aws.amazon.com/general/latest/gr/signing_aws_api_requests.html) or [GCP instance identity verification](https://cloud.google.com/compute/docs/instances/verifying-instance-identity).

Do you mean adding that to the existing approval mechanism, or removing the existing approval mechanism entirely and switching it to be based entirely on cloud APIs? If you just mean adding, then yes, this would let the CSR approver distinguish "this is actually a real node VM" from "this is someone pretending to be a new node VM". But as long as we still have `Machine` objects, and the CSR approver still requires that `Node`s come up with IPs that match their corresponding `Machine`s, then the MCS can check those IPs itself too, and refuse to talk to any IP that isn't a valid proto-node IP.

squeed: >  An actual functioning node has no reason to talk to the MCS (as I understand it).

Correct, unless an existing node is re-provisioning for some reason (e.g. disk failure), and the Node object still exists. (Is this a case we need to handle? Will the new CSR even be signed?)

wking: > ...an IP should only be contacting the MCS if there is a Machine that claims that IP but no Node that claims that IP...

Folks can also provision their own instances, without going through the machine API.  In that case, they're the hook to handle CSR signing for those machines.  With the "MCS includes CSR-approver-style-verification" approach, those folks would need a hook so they could tell the MCS, yes, really serve requests to my new proto-node I manually provisioned".

enhancements/machine-config/securing-mcs.md
```diff
@@ -0,0 +1,121 @@
...
...sitive data through the Machine Config Server, the port restrictions can be safely removed. In order to achieve this, Machine Configs can be split into two groups: pre-node-approved configs and post-node-approved configs. The pre-node-approved configs are ones which do not contain sensitive information and contain just enough to configure new nodes to submit a certificate signing request (CSR) to the API server. The post-node-approved configs contain everything else, including the overwhelming majority of customer-provided Machine Configs. When a new machine is created, it fetches its Ignition config from the Machine Config Server, which only serves the pre-node-approved configs. At this point, the machine has enough information to submit a CSR, but nothing sensitive (note that for the purposes of this proposal, the CSR bootstrap token and pull secret are not considered sensitive). Once the CSR has been approved and a kubeconfig returned, the Machine Config Daemon is able to directly enumerate Machine Configs, including the post-node-approved ones, so it then applies any remaining configuration to the machine and reboots if necessary. At this point, the machine is fully configured.
```
cgwalters, line 59: Today we carry effectively 1.63 implementations of Ignition - one in the upstream project that runs in the initramfs, and  63% of Ignition in the MCD that runs in the real root.  If do this, then we'd hit all limitations of the 63% MCD-Ignition implementation upfront.  For example, while this isn't important to OpenShift, today the MCO doesn't support adding/changing users.

We may hit edge cases around e.g. parent directory creation, file write ordering etc.  Those are all bugs that *should* be fixed of course, but we also tend to not hit them today because use of "day 2" MCs is so limited.

If we go this route it may be worth a spike around how we can have the MCO actually reuse Ignition upstream code directly for this.

enhancements/machine-config/securing-mcs.md
cgwalters: I think it's a no-brainer to include node IP validation in the MCS - that's what my original PR has in fact here: https://github.com/openshift/machine-config-operator/pull/784

But that said, we also need to care about UPI on-premise installations where the network may be exposed to machines outside the cluster.  To me, this is where even a non-rotating auth token is extremely helpful - it provides a really strong first layer of defense.

PR #504, opened 2020-10-19
single-node production deployments

dhellmann:
This enhancement describes a new single-node cluster profile for
production use in "edge" deployments that are not considered to be
resource-constrained, such as telecommunications bare metal
environments.

PR #628, opened 2021-02-04
automated resource request scaling

dhellmann:
This enhancement describes an approach to allow us to scale the
resource requests for the control plane services to reduce consumption
for constrained environments. This will be especially useful for
single-node production deployments, where the user wants to reserve
most of the CPU resources for their own workloads and needs to
configure OpenShift to run on a fixed number of CPUs within the host.

/cc @smarterclayton @derekwaynecarr @markmc @browsell

### DIFF
--- a/enhancements/automated-resource-request-scaling.md
+++ b/enhancements/automated-resource-request-scaling.md
@@ -1,0 +1,886 @@
+---
+title: automated-cpu-resource-request-scaling
+authors:
+  - "@dhellmann"
+  - "@csrwng"
+  - "@wking"
+  - "@browsell"
+reviewers:
+  - TBD
+approvers:
+  - "@smarterclayton"
+  - "@derekwaynecarr"
+  - "@markmc"
+creation-date: 2021-02-04
+last-updated: 2021-02-04
+status: provisional
+see-also:
+  - "/enhancements/single-node-production-deployment-approach.md"
+---
+
+# Automated CPU Resource Request Scaling for Control Plane
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement describes an approach to allow us to scale the
+resource requests for the control plane services to reduce consumption
+for constrained environments. This will be especially useful for
+single-node production deployments, where the user wants to reserve
+most of the CPU resources for their own workloads and needs to
+configure OpenShift to run on a fixed number of CPUs within the host.
+
+One example of this use case is seen in telecommunication service
+providers implementation of a Radio Access Network (RAN). This use case
+is discussed in more detail below.
+
+## Motivation
+
+The resource requests for cluster operators and their operands are
+based on performance analysis on multiple cloud VMs using the
+end-to-end test suite to gather data. While the resulting numbers work
+well for similar cloud environments and even multi-node bare metal
+deployments, they result in over-provisioning resources for
+single-node deployments.
+
+### Goals
+
+* This enhancement describes an approach for configuring OpenShift
+  clusters to run with lower CPU resource requests than its default
+  configuration. The current resource requirements for an OpenShift
+  control plane exceed the size desired by some customers running on
+  bare metal.
+* Clusters built in this way should pass the same Kubernetes and
+  OpenShift conformance and functional end-to-end tests as single-node
+  deployments that have not been scaled down.
+* We have a goal of 4 hyperthreads today, but we do not know what
+  constraints we might be given in the future so we need a solution
+  that is not tied to the current known limit.
+* We want a general approach, that can be applied to all OpenShift
+  control plane components.
+
+### Non-Goals
+
+* This enhancement is focused on CPU resources. Other compressible
+  resource types may need to be managed in the future, and those are
+  likely to need different approaches.
+* This enhancement does not address non-compressible resource
+  requests, such as for memory.
+* This enhancement does not attempt to describe application-specific
+  APIs for tuning the configuration of individual operators or their
+  workloads.
+* This enhancement does not address ways to disable operators or
+  operands entirely.
+* Although the results of this enhancement may be useful for
+  single-node developer deployments, multi-node production
+  deployments, and Internet-of-things devices, those use cases are not
+  addressed directly in this document.
+* This enhancement describes a way to limit the CPU resource
+  requests. It does not address reducing actual utilization. There is
+  no expectation that a cluster configured to use fewer than the
+  default resources would offer exactly the same performance as the
+  default. It must be stable and continue to operate reliably, but may
+  respond more slowly.
+* This enhancement does not address scaling the control plane **up**.
+
+## Proposal
+
+There are two aspects to manage scaling the control plane resource
+requests. Most of the cluster operators are deployed by the
+cluster-version-operator using static manifests that include the
+resource requests. Those operators then deploy other controllers or
+workloads dynamically. We need to account for both sets of components.
+
+### Deployments managed by the cluster-version-operator or operator-lifecycle-manager
+
+To start, we will [reduce the default
+requests](https://bugzilla.redhat.com/show_bug.cgi?id=1920159) for
+cluster operators deployed by the cluster-version-operator with static
+manifests. By re-tuning those settings by hand, we may be able to
+avoid implementing a more complex system to change them dynamically
+within the cluster.
+
+A similar approach can be taken for operators installed via the
+operator-lifecycle-manager.
+
+### Deployments managed by cluster operators
+
+We are considering 2 approaches for operands of cluster operators.
+
+#### Kubelet over-subscription configuration option
+
+Today, kubelet subtracts reserved CPUs from the available resources
+that it reports, under the assumption that they will be entirely
+consumed by workloads. Sometimes workloads do not entirely consume
+those CPUs, however, and non-guaranteed pods can also use the reserved
+CPU pool. A change to kubelet to provide a static over-subscription
+configuration option would provide a global approach to allowing the
+CPU resources reported by kubelet to be adjusted by a factor defined
+by the user to account for this difference, based on the workloads
+planned for the cluster.
+
+The new configuration option needs to be added to kubelet upstream in
+order to be supported long-term in OpenShift. A KEP will be written
+and submitted independently of this enhancement document.
+
+In OpenShift, the option would be set during installation and would be
+added to the control plane `MachineConfig`.
+
+We would need to carefully consider how to document the use of this
+option. Users will need to understand the minimum CPU resource
+requests of a base OpenShift deployment and any additional add-on
+operators before they will be able to tune the over-subscription
+setting so that all the components will fit onto the number of CPUs
+they require. We will need to keep the documentation of minimum CPU
+resource requests for OpenShift up to date for each release, or
+provide a tool to calculate the value from a default cluster.
+
+Refer to the implementation history section below for details of a
+proof-of-concept implementation of this approach.
+
+#### Cluster-wide resource request limit API
+
+The kubelet change in option 1 requires work to ensure the option is
+accepted upstream. That requirement introduces risk, if the upstream
+community does not like the option or if it takes a long time to reach
+agreement. As an alternative approach, we are considering a new
+cluster-wide API.
+
+Tuning for operator-managed workloads should be controlled via an API,
+so that operators can adjust their settings correctly. We do not want
+users to change the setting after building the cluster (at least for
+now), so the API can be a status field. As with other
+infrastructure-related settings, the field will be added to the
+`infrastructure.config.openshift.io` [informational
+resource](https://docs.openshift.com/container-platform/4.6/installing/install_config/customizations.html#informational-resources_customizations).
+
+The new `controlPlaneCPULimit` field will be a positive integer
+representing the whole number of [Kubernetes
+CPUs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu)
+into which the control plane must fit. Expressing the API this way
+makes it easy for the user to declare the requirements up front when
+installing the cluster, without having to understand the basis for the
+default settings and without having to calculate a scaling factor
+themselves.
+
+In order for operators to know how to use `controlPlaneCPULimit` to
+compute a scaling factor, they need to know the default number of CPUs
+assumed for a control plane that produced the default resource
+limit. Rather than encoding that information separately in every
+operator implementation, we should put it in one place so it can be
+managed consistently. Using a `const` defined in the library with the
+Infrastructure type makes it easy to upgrade operators if the value
+changes across releases. So, a new `DefaultControlPlaneCPUCount`
+constant will be added to the module with the Infrastructure type.
+
+If `controlPlaneCPULimit == 0`, operators should use their
+default configuration for their operands.
+
+If `0 < controlPlaneCPULimit < DefaultControlPlaneCPUCount`, operators
+should scale their operands down by around `(controlPlaneCPULimit /
+DefaultControlPlaneCPUCount)` or more, if appropriate. When scaling,
+the operators should continue to follow the other guidelines for
+resources and limits in
+[CONVENTIONS.md](CONVENTIONS.md#resources-and-limits).
+
+If `controlPlaneCPULimit >= DefaultControlPlaneCPUCount`, operators
+should use their default configuration for their operands. We do not
+currently have a need to allow the control plane to scale up, although
+this may change later. If we do have a future requirement to support
+that, we could remove this restriction.
+
+The logic described in this section could also be captured in a
+library function. Something like
+
+```go
+cpuRequestValue, err := GetScaledCPURequest(kubeClientConfig, myDefaultValue)
+```
+
+where the function fetches the Infrastructure resource and applies the
+algorithm to a default value provided by the caller. That would make
+the code change to implement the approach in each operator smaller,
+since we could conceivably wrap the existing static values with the
+function call.
+
+Users will specify the `controlPlaneCPULimit` when building a cluster
+by passing the value to the installer in `install-config.yaml`.
+
+### User Stories
+
+#### Radio Access Network (RAN) Use Case
+
+In the context of telcocommunications service providers' 5G Radio Access
+Networks, it is increasingly common to see "cloud native" implementations
+of the 5G Distributed Unit (DU) component. Due to latency constraints,
+this DU component needs to be deployed very close to the radio antenna for
+which it is responsible. In practice, this can mean running this
+component on anything from a single server at the base of a remote cell
+tower or in a datacenter-like environment serving several base stations.
+
+A hypothetical DU example is an unusually resource-intensive workload,
+requiring 20 dedicated cores, 24 GiB of RAM consumed as huge pages,
+multiple SR-IOV NICs carrying several Gbps of traffic each, and
+specialized accelerator devices. The node hosting this workload must
+run a realtime kernel, be carefully tuned to ensure low-latency
+requirements can be met, and be configured to support features like
+Precision Timing Protocol (PTP).
+
+The most constrained resource for RAN deployments is CPU, so for now
+the focus of this enhancement is scaling CPU requests.  Kubernetes
+resource requests for CPU time are measured in fractional "[Kubernetes
+CPUs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu)". For
+bare metal, 1 Kubernetes CPU is equivalent to 1 hyperthread.
+
+Due to the resource-intensive workload, the overhead for platform
+components such as the OpenShift platform and add-on operators is
+severely restricted when compared to other uses. For example, one
+customer has allocated 4 hyperthreads for all components other than
+their own workloads. OpenShift currently requires 7.
+
+### Implementation Details/Notes/Constraints
+
+1. Add the new field to the `Infrastructure` API.
+2. Expose the `controlPlaneCPULimit` setting through the installer via
+   `install-config.yaml`.
+3. Update operators to respond to the API, as described above.
+
+### Risks and Mitigations
+
+If we choose to add a cluster-wide API and update operators to use it
+to configure their operands, then we may have to update a lot of
+different operators. We could mitigate that by starting with some of
+the most expensive operators and either having them adjust their
+workloads based on the new API or use the existing infrastructure
+topology API. Using the existing API would mean we could take more
+time to think about what API we do want, but we would be giving up the
+ability to tune the configuration of operators for more scenarios.
+
+The operators with the highest excess requests compared to idle use
+include:
+
+| Namespace | Pod | OCP 4.6 Average Use at rest | Request Size | Difference | % Difference |
+| --------- | --- | ----------------------------| -------------| ---------- | ------------ |
+| openshift-etcd | etcd | 200 | 350 | 150 | 43% |
+| openshift-oauth-apiserver | apiserver | 6 | 150 | 144 | 96% |
+| openshift-apiserver | apiserver | 16 | 125 | 109 | 87% |
+| openshift-controller-manager | controller-manager | 3 | 100 | 97 | 97% |
+| openshift-ingress | router-default | 4 | 100 | 96 | 96% |
+| openshift-dns | dns-default | 4 | 65 | 61 | 94% |
+| openshift-kube-apiserver | kube-apiserver | 230 | 290 | 60 | 21% |
+| openshift-kube-controller-manager | kube-controller-manager | 23 | 80 | 57 | 71% |
+
+## Design Details
+
+### Open Questions
+
+1. If manually re-tuning the settings for the static manifests managed
+   by the cluster-version-operator is not sufficient, what should we
+   do?
+2. How do we express to the cluster operators that they should scale
+   their operands?
+3. How dynamic do we actually want the setting to be? Should the user
+   be able to change it at will, or is it a deployment-time setting?
+4. How do we handle changes in the base request values for a cluster
+   operator over an upgrade?
+5. Are operators allowed to place a lower bound on the scaling? Is it
+   possible to request too few resources, over subscribe a CPU, and
+   cause performance problems for the cluster?
+
+### Test Plan
+
+**Note:** *Section not required until targeted at a release.*
+
+Consider the following in developing a test plan for this enhancement:
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+- What additional testing is necessary to support managed OpenShift service-based offerings?
+
+No need to outline all of the test cases, just the general strategy. Anything
+that would count as tricky in the implementation and anything particularly
+challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations).
+
+### Graduation Criteria
+
+**Note:** *Section not required until targeted at a release.*
+
+Define graduation milestones.
+
+These may be defined in terms of API maturity, or as something else. Initial proposal
+should keep this high-level with a focus on what signals will be looked at to
+determine graduation.
+
+Consider the following in developing the graduation criteria for this
+enhancement:
+
+- Maturity levels
+  - [`alpha`, `beta`, `stable` in upstream Kubernetes][maturity-levels]
+  - `Dev Preview`, `Tech Preview`, `GA` in OpenShift
+- [Deprecation policy][deprecation-policy]
+
+Clearly define what graduation means by either linking to the [API doc definition](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning),
+or by redefining what graduation means.
+
+In general, we try to use the same stages (alpha, beta, GA), regardless how the functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+**Examples**: These are generalized examples to consider, in addition
+to the aforementioned [maturity levels][maturity-levels].
+
+#### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+- Enumerate service level indicators (SLIs), expose SLIs as metrics
+- Write symptoms-based alerts for the component(s)
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+
+**For non-optional features moving to GA, the graduation criteria must include
+end to end tests.**
+
+#### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
+
+### Upgrade / Downgrade Strategy
+
+If applicable, how will the component be upgraded and downgraded? Make sure this
+is in the test plan.
+
+Consider the following in developing an upgrade/downgrade strategy for this
+enhancement:
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade in order to keep previous behavior?
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade in order to make use of the enhancement?
+
+Upgrade expectations:
+- Each component should remain available for user requests and
+  workloads during upgrades. Ensure the components leverage best practices in handling [voluntary disruption](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/). Any exception to this should be
+  identified and discussed here.
+- Micro version upgrades - users should be able to skip forward versions within a
+  minor release stream without being required to pass through intermediate
+  versions - i.e. `x.y.N->x.y.N+2` should work without requiring `x.y.N->x.y.N+1`
+  as an intermediate step.
+- Minor version upgrades - you only need to support `x.N->x.N+1` upgrade
+  steps. So, for example, it is acceptable to require a user running 4.3 to
+  upgrade to 4.5 with a `4.3->4.4` step followed by a `4.4->4.5` step.
+- While an upgrade is in progress, new component versions should
+  continue to operate correctly in concert with older component
+  versions (aka "version skew"). For example, if a node is down, and
+  an operator is rolling out a daemonset, the old and new daemonset
+  pods must continue to work correctly even while the cluster remains
+  in this partially upgraded state for some time.
+
+Downgrade expectations:
+- If an `N->N+1` upgrade fails mid-way through, or if the `N+1` cluster is
+  misbehaving, it should be possible for the user to rollback to `N`. It is
+  acceptable to require some documented manual steps in order to fully restore
+  the downgraded cluster to its previous state. Examples of acceptable steps
+  include:
+  - Deleting any CVO-managed resources added by the new version. The
+    CVO does not currently delete resources that no longer exist in
+    the target version.
+
+### Version Skew Strategy
+
+Clusters deployed from versions that do not support scaling will show
+a value of `0`, causing operators that support scaling to use their
+default settings.
+
+Operators from versions that do not support scaling will ignore the
+new field until they are upgraded.
+
+The value of `DefaultControlPlaneCPUCount` constant may change between
+releases. The value from the library will be vendored into each
+operator, so the scaling formula for a given version of an operator
+will be correct at all times during an upgrade, even if different
+operators are using different base values.
+
+## Implementation History
+
+Some [work to reduce the static default requests for cluster
+operators](https://bugzilla.redhat.com/show_bug.cgi?id=1920159) has
+been done in the 4.8 release.
+
+### Proof-of-concept for kubelet over-subscription option
+
+A proof-of-concept implementation of a kubelet configuration option to
+declare a single static over-subscription value has been
+built.
+
+#### Code changes
+
+We want to increase (compensate) allocatable by the
+`overSubscriptionCapacity` amount.  A change was made to the
+`node_container_manager_linux.go` by introducing a static
+oversubscription capacity of 1 CPU.
+
+```diff
+[root@cnfdd3-installer kubernetes]# git diff
+diff --git a/pkg/kubelet/cm/cpumanager/cpu_manager.go b/pkg/kubelet/cm/cpumanager/cpu_manager.go
+index 6acd8ebc660..6a8d2e94043 100644
+--- a/pkg/kubelet/cm/cpumanager/cpu_manager.go
++++ b/pkg/kubelet/cm/cpumanager/cpu_manager.go
+@@ -160,6 +160,9 @@ func NewManager(cpuPolicyName string, reconcilePeriod time.Duration, machineInfo
+                // exclusively allocated.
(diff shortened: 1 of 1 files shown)


enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,517 @@
...
+
+Consider including folks that also work outside your immediate sub-project.
+
+## Design Details
+
+### Open Questions
+
+1. The cluster-version-operator manages cluster operators using static
+   manifests. How would we make changes to the settings of Deployments
+   defined in those manifests?
```
wking, line 153: I'm still not convinced that we actually need to tune requests for CVO-managed operators, because operator consumption seems like it would be fairly independent of cluster workload (although I expect operands like etcd member and API-server consumption to scale with cluster workload).  Do we have an example CVO-managed deployment where we have one number we like for the general case and a different number we like for the single-node case?

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,517 @@
...
+resource requests. Those operators then deploy other controllers or
+workloads dynamically. We need to account for both sets of components.
+
+
+
+This is where we get down to the nitty gritty of what the proposal actually is.
+
+### User Stories
+
+Detail the things that people will be able to do if this is implemented.
```
wking, line 120: nit: `Radio Access Network (RAN) Use Case` section can probably be copy/pasted down into a subsection here under `User Stories`

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,517 @@
...
+
+1. The cluster-version-operator manages cluster operators using static
+   manifests. How would we make changes to the settings of Deployments
+   defined in those manifests?
+2. How do we express to the cluster operators that they should scale
+   their operands?
+3. How dynamic do we actually want the setting to be? Should the user
+   be able to change it at will, or is it a deployment-time setting?
+4. What happens during an upgrade? The new manifests from the new
+   release payload would have unscaled limits.
```
wking, line 159: Updates aren't all that special from the perspective of individual resources.  It's just a CVO reconciling manifests like the CVO always does, just from a different bag of manifests.  I expect 2nd-level operators are similar in how they roll out updates to their own operands.  However we have these limits applied in the steady state should apply on updates, although it will be easier to be sure once we have a steady-state plan.  And a plan for single-node updates ;).

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,517 @@
...
+Updating existing operators to read the new field and scale their
+operands and documenting an entirely new API might be more work than
+using the Infrastructure API that would only be useful if the operator
+responding to the API was an add-on.
+
+### Dynamic vs. Static Setting
+
+It is not likely that users would need to change the setting after
+building a cluster, so we could have the installer set a status field
+in whatever CRD we choose for the API. If we do need to support
```
wking, line 373: Would the chain be user -> installer -> static setting?  Because teaching the installer to automatically know what the setting should be strikes me as harder than teaching some in-cluster component to automatically know what the setting should be (because an in-cluster component could actually look, while the installer will have to guess ahead of time).  If the installer is just forwarding a user choice, that seems achievable, but we'd need to train users in selecting an appropriate value.

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,517 @@
...
+operators to be starved for resources, triggering degraded cluster
+performance.
+
+The static settings only apply to the cluster operators Deployments,
+and changing them does not signal to the operators that they need to
+change the settings for their operands.
+
+#### Have a new operator in the release payload perform the scaling for control plane operators
+
+It would use the new API field to determine how to scale (see above).
```
wking, line 417: sort of ambiguous in the current enhancement is whether the requested scaling is:

> Single-node workloads are smaller, so control-plane consumption is smaller, and we'd like the requests reduced to reflect that.

or if it is:

> Single-node workloads are smaller, so control-plane consumption is smaller.  But it's not smaller enough.  So we need an extra knob to say "hey, control-plane components, we need to to be extra lean on CPU, so feel free to swap in some long poll durations, add heavy internal rate-limiting, and so on to get your consumption down.  We'll accept the decreased performance".

Do we know which of those it is yet?  If we do, can we make it explicit in the `Proposal` and (non-)goal sections?

enhancements/automated-resource-request-scaling.md
dhellmann: Yeah, I was thinking that if some other actor changes the requests, then the CVO gets a new manifest it would overwrite the value when applying the new thing, then the other actor would come a long and change it again. Maybe that's not what would happen, though, if the request fields are ignored when the manifest and resource are compared?

enhancements/automated-resource-request-scaling.md
dhellmann: Yes, I was thinking a new field in install-config.

enhancements/automated-resource-request-scaling.md
dhellmann: @browsell does have some examples of operators that request more resources than they need in the *RAN* use case. I don't know if changing those requests for all other use cases makes sense.

The console operator requests 50m and uses 5m, when the console itself is disabled.

The insights operator requests 10m and uses 1m, in a disconnected environment.

The full list is a bit long to paste here, but there are several others and all together they add up to requests of 210m where the average use is only a total of 22m.

enhancements/automated-resource-request-scaling.md
dhellmann: It is definitely the latter. Actual use is less, but the requests have to fit into less space than they do right now (4 CPU instead of 7).

enhancements/automated-resource-request-scaling.md
dhellmann: Done

enhancements/automated-resource-request-scaling.md
dhellmann: I added some text to clarify that the setting would come from the installer's input and not need to be derived.

enhancements/automated-resource-request-scaling.md
dhellmann: I updated the text of the first goal and added a new non-goal to explain this. I'm holding off on writing the proposal section until there's more feedback about the various options listed in the alternatives section.

wking: New (non-)goal text looks good to me.  Feel free to mark this thread resolved.

enhancements/automated-resource-request-scaling.md
wking: yeah, if we teach the CVO to not care about these requests (which it would have to do in order to allow the non-CVO controller to manage them), the incoming CVO would continue to ignore the requests, so I don't expect any CVO stomping to take place, even on updates.

enhancements/automated-resource-request-scaling.md
wking: I'm curious about the difference.  I guess there are some API requests they're saving when they only have to monitor their config to notice they're off, and don't have to manage any operands.  Things like Insights I expect to be extremely bursty when they're doing real work, and then sleeping in between, and that's hard to represent with `requests` in a sane way, because you can fit a lot of bursty things in the same space if the bursts don't overlap, but it gets tight if the bursts do overlap.  Maybe it would help with scheduling if the Insights operator launched higher-request jobs when it was configured to run, and then the operator itself could lower its request settings?

enhancements/automated-resource-request-scaling.md
wking: Mark this thread resolved so it doesn't distract future reviewers, now that it's been addressed in the text?

dhellmann: Yep, I wanted to get an ack from you before resolving it.

enhancements/automated-resource-request-scaling.md
dhellmann: OK, excellent, I'll update the text to take that into account.

enhancements/automated-resource-request-scaling.md
dhellmann: The _requests_ really do matter, too. If we have to fit into N CPUs and we're requesting N + 0.5 CPUs, trimming the requests down can help squeeze into that space. Then, as you point out, the bursty nature of actual use may cause problems, too, and so we may end up having to solve those on a case-by-case basis. Hence the large matrix of options to consider at the end of the doc.

wking: I am not saying requests don't matter.  This enhancement is about finding ways to request OCP components to trim down even more beyond what they'd otherwise want to consume for performance reasons.  But the console/insights examples you gave are just plain over-requesting.  You're not asking the console operator to do less than it wants to do; you're just pushing back on it requesting 50m when in this case it only needs 5m to do everything it wants to do.  So the more situations like that we can find, and make adjustments to get requests in line with actual consumption, the better job the scheduler can do finding folks space, and the fewer places we'll need to tell operators to tone down what they're actually doing.  A console operator managing an active console operand may actually need 50m in most cases; I dunno.  But I'm pretty sure that the bulk of the time, the Insights operator is just sleeping between gathers.  If we can restructure the Insights operator into something with a constant, low CPU consumption, it can lower its requests on all deployments, not just single-node deployments.  And it can use jobs for the periodic, bursty gathers that set higher requests.  And those jobs will never fire on single-node deployments, so those folks don't have to budget CPU for them.  And the jobs will fire for other folks, but maybe we set them low priority, so they only get scheduled into gaps created by other bursty OCP-core workload.  And once we've tuned down that sort of over-request issue, we can come back and get a better feel for how and where we want to ask operators to reduce beyond their current single-node usage.

imiller0: One thought is that the appropriate cpu request value for an operator on an edge deployed single-node cluster may not be the same as for a larger multi-node cluster. The advantage of the global options approach below is that (when limited to infrastructure/system-critical pods) it can be used to help bring the infrastructure requests in line with actual usage without having to update each operator. This would limit the need for modifications to operator behavior to those cases where actual CPU load exceeds the available capacity budgeted for overhead.

dhellmann: @wking I think your suggestion is at least partly covered by "Lower the resource requests for cluster operators statically in the default configuration" on line 397. That section does not cover having those operators change the way they manage tasks that require CPU bursts, and what you describe here makes me think a one-size-fits-all approach isn't likely to be as effective, long-term, as more thoughtful design changes. Do you think it's sufficient to extend that section to say that we may need to rewrite some of the bursty operators, or should that be a separate alternative entirely?

@imiller0 Would your idea be covered by either "Use a cluster profile to change the resource requests for control plane operators" or one of the alternatives for scaling dynamically?

imiller0: I would echo the same concern with cluster profiles that you called out below (potential for a lot of customer specific profiles). Dynamic scaling is more along the lines of what I was thinking, with the question being what mechanism/where to do the scaling. Using kubelet to do the scaling or increasing allocatable capacity (as described under the global options) when the pod is scheduled seems to avoid some of the issues with modifying the pod's request and then dealing with reconciliation.

dhellmann: @wking & @imiller0, I think I have captured the details from this thread in the latest draft. Please let me know if I've missed anything.

dhellmann:
I have updated the enhancement with more details in the proposal section. Please take another look.

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,635 @@
...
+
+### Deployments managed by the cluster-version-operator
+
+To start, we will [reduce the default
+requests](https://bugzilla.redhat.com/show_bug.cgi?id=1920159) for
+cluster operators deployed with static manifests. By re-tuning those
+settings by hand, we may be able to avoid implementing a more complex
+system to change them dynamically within the cluster.
+
+### Deployments managed by cluster operators
```
browsell, line 100: The same strategy would apply to OLM operators as well ?

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,635 @@
...
+Infrastructure type makes it easy to upgrade operators if the value
+changes across releases. So, a new `DefaultControlPlaneCPUCount`
+constant will be added to the module with the Infrastructure type.
+
+If `controlPlaneCPULimit == 0`, operators should use their
+default configuration for their operands.
+
+If `0 < controlPlaneCPULimit < DefaultControlPlaneCPUCount`, operators
+should scale their operands down by around `(controlPlaneCPULimit /
+DefaultControlPlaneCPUCount)` or more, if appropriate. When scaling,
```
browsell, line 134: One consideration is that there may not be a one size fits all scaling factor based on the use case. For Telco use cases, some operators are not required whereas they may be for other use cases. The console would be a good example. In the former case, we would want to be more aggressive with the scale down.

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,635 @@
...
+can be described in another enhancement.
+
+### Managing settings for operators installed by cluster-version-operator
+
+The alternatives in this subsection cover the need to change the
+settings for the static manifests currently managed by the
+cluster-version-operator. These approaches would only apply to the
+operators themselves, and not their operands.
+
+1. Use a cluster profile to change the resource requests for control
```
browsell, line 464: Recognize the support costs but this option would also allow the customization of the installed operators to exclude any not required.

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,635 @@
...
+   the changes we may lower values too much and cause important
+   cluster operators to be starved for resources, triggering degraded
+   cluster performance.
+
+   The static settings only apply to the cluster operators
+   Deployments, and changing them does not signal to the operators
+   that they need to change the settings for their operands.
+
+   With lowered CPU requests, the consumption of some operators may
+   appear more sporadic ("bursty") than it does currently. One way to
```
browsell, line 500: Not sure I understand this statement. The absolute cpu usage of any spikes should not change.

enhancements/automated-resource-request-scaling.md
dhellmann: Yes, good point.

enhancements/automated-resource-request-scaling.md
dhellmann: Operators that can be disabled will have their own API to do that, and those changes would be covered by other enhancements. So for the purposes of this doc, assume we're talking only about how to scale the things we haven't disabled or re-configured in some application-specific way. I can update the introduction to clarify that.

The goal here was to describe a general API that would apply to everything running in the cluster. If we need finer control over scaling for some specific operators to handle special use cases, those would be application-specific APIs, like the ones to disable operators. I can update the non-goals section to talk about creating those in separate enhancements.

enhancements/automated-resource-request-scaling.md
dhellmann: See #504 for some history of that discussion. I'll update this section with a link to that older enhancement.

enhancements/automated-resource-request-scaling.md
dhellmann: I phrased this poorly. What I was trying to capture was @wking's comments about restructuring some operators so that they live within their requested resources (so those requests continue to look accurate), and the periodic "bursts" of work are run in a way that means they can be scheduled separately.

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,635 @@
...
+based on performance analysis on multiple cloud VMs using the
+end-to-end test suite to gather data. While the resulting numbers work
+well for similar cloud environments and even multi-node bare metal
+deployments, they result in over-provisioning resources for
+single-node deployments.
+
+### Goals
+
+* This enhancement describes an approach for configuring OpenShift
+  clusters to run with lower CPU resource requests than its default
```
smarterclayton, line 57: We should probably broaden this or qualify this as "compressible resources" - by definition this wouldn't help with incompressible resources.

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,635 @@
...
+Due to the resource-intensive workload, the overhead for platform
+components such as the OpenShift platform and add-on operators is
+severely restricted when compared to other uses. For example, one
+customer has allocated 4 hyperthreads for all components other than
+their own workloads. OpenShift currently requires 7.
+
+### Implementation Details/Notes/Constraints
+
+1. Add the new field to the `Infrastructure` API.
+2. Expose the `controlPlaneCPULimit` setting through the installer via
```
smarterclayton, line 183: And 3: all components respect this.  I think this is the problem with this design - it requires EVERY component perform the same calculation (it denormalizes something inefficiently).  If the calculation is standard, I would expect components to have to do nothing.

I would prefer an approach that centralizes the responsibility for scaling these components, not divides them out inefficiently.  This isn't about operators, this is about a set of target resource constraints.  That could be to the node, an admission plugin, or to a more general purpose solution.  I don't think this scales correctly.

Also, because this is about compressible resources, eventually we will have other constraint types besides CPU, so this already hits problems with being too CPU specific.

dhellmann: Centralizing it makes a lot of sense. There are a few options below for doing that, starting on line 457. 

I like the in-cluster operator approach, because it doesn't turn resource requests into something we would need to support users modifying directly. If that isn't a big concern, then I like the "external actor" approach, since we can probably do it with an ACM policy and not have to write any scaling code at all. What do you think?

As far as other compressible resources, I think we're likely to need to be able to express each of those in their own units somehow. That implies separate API fields, and I would like to keep this enhancement focused on scaling the thing we know we need to scale today (I should update the title and summary accordingly). Are you thinking of a more general API that does not use separate fields for things like CPU and network bandwidth? Or do you foresee issues waiting to consider how we might scale things other than CPU?

dhellmann: I guess another benefit of the external-actor approach is that it would let us tune things differently for different use cases, which is something @browsell brought up in a comment above.

browsell: Also one less component consuming resources on the node as you point out below.

dhellmann: Thinking about the centralization more, I'm not sure an independent actor, either inside the cluster or outside, would work.

It could change the settings of Pods from Deployments managed by the CVO without a problem. But as soon as it tried to change a workload being managed by another operator, the two things would be in contention. So if each operator doesn't support scaling on its own, then we have to do the scaling in a way that's completely transparent to the Pod definitions.

dhellmann: Another argument in favor of having each operator deal with scaling is that it may need to change more than the resource requests. For example, if CPU requests are reduced and we expect that might lead to slower operation (if there is no overhead for bursting), we may need to change timeouts dynamically as well.

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,635 @@
...
+DefaultControlPlaneCPUCount)` or more, if appropriate. When scaling,
+the operators should continue to follow the other guidelines for
+resources and limits in
+[CONVENTIONS.md](CONVENTIONS.md#resources-and-limits).
+
+If `controlPlaneCPULimit >= DefaultControlPlaneCPUCount`, operators
+should use their default configuration for their operands. We do not
+currently have a need to allow the control plane to scale up, although
+this may change later. If we do have a future requirement to support
+that, we could remove this restriction.
```
dhellmann, line 143: The logic described in this section could also be captured in a library function. Something like

```
cpuRequestValue, err := GetScaledCPURequest(kubeClientConfig, myDefaultValue)
```

where the function fetches the Infrastructure resource and applies the algorithm to `myDefaultValue`.

That would make the update in each operator smaller, since we could conceivably wrap the existing static values with the function call.

enhancements/automated-resource-request-scaling.md
dhellmann: An outside actor could be made to work if the operators ignored the resource requests when checking for equality -- basically every operator would have to do what we said up on pin 92 that the CVO has to do.

That complicates upgrades, though, because how does an operator ever change the resource requests for the things it is managing? Maybe we use an annotation on the Deployment to set the baseline value(s)?

imiller0: Could we turn the up-front scaling (options on line 457) around by decoupling the requests from the “applied” impact on the node? If we scale the request when updating the remaining allocatable space during pod admission to a node we preserve scheduling space for the user’s pods. In practical terms this is somewhere between the two global options (lines 593 and 602), and may meet your criteria of being transparent to the Pod definition.

As noted below this does complicate debug/analysis of the system but would decoupling requests from the applied state be beneficial for other scenarios? For example scheduling a pod on a hyperthreaded node where, for security or performance reasons, it may be necessary to exclude all code execution on the sibling CPU. If such a pod requests 1 CPU and we offline the sibling the “applied” impact is 2 CPU, a similar scaling approach (although scaled up) as could be used here.

dhellmann: That sounds a lot like the option described on line 610, where kubelet scales the request before reporting the available capacity on a node. Unless there's a subtle difference that I'm missing?

imiller0: Yes, apologies, I misread that option as explicitly adding in extra capacity. Re-framing my comment then would be that having kubelet scale the requests at the node (as in the option on line 610) might meet the goal of being transparent to the Pod definition avoiding some of the thrashing scenarios and may have applications in other scenarios as well.

dhellmann: No worries, I'll try to clarify the language for that one. There are a lot of options to keep straight. :-)

I think the main challenge with doing it inside kubelet is then we're carrying a fork of kubelet, at least until/unless we can convince upstream to take the feature. But, I'm not a kubelet expert so perhaps there's a way to make a plugin work?

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,635 @@
...
+   Presenting false data to influence the schedule may be very tricky to
+   get right and would complicate debugging and support. We would need
+   some way to expose the difference between the actual capacity and the
+   capacity value being used for scheduling.
+
+   It's not clear if we could implement these behaviors in a plugin, or
+   if we could push the changes upstream. Carrying a fork of kubelet to
+   implement the feature would result in additional release management
+   work indefinitely.
+
```
vitus133, line 632: We did a POC with adding a static oversubscription capacity.

## Code changes
We want to increase (compensate) allocatable by the overSubscriptionCapacity amount.
A change was made to the node_container_manager_linux.go by introducing a static oversubscription capacity of 2.5 cpu.

```
[root@cnfdd3-installer kubernetes]# git diff
diff --git a/pkg/kubelet/cm/cpumanager/cpu_manager.go b/pkg/kubelet/cm/cpumanager/cpu_manager.go
index 6acd8ebc660..6a8d2e94043 100644
--- a/pkg/kubelet/cm/cpumanager/cpu_manager.go
+++ b/pkg/kubelet/cm/cpumanager/cpu_manager.go
@@ -160,6 +160,9 @@ func NewManager(cpuPolicyName string, reconcilePeriod time.Duration, machineInfo
                // exclusively allocated.
                reservedCPUsFloat := float64(reservedCPUs.MilliValue()) / 1000
                numReservedCPUs := int(math.Ceil(reservedCPUsFloat))
+               if specificCPUs.Size() > 0 {
+                       numReservedCPUs = specificCPUs.Size()
+               }
                policy, err = NewStaticPolicy(topo, numReservedCPUs, specificCPUs, affinity)
                if err != nil {
                        return nil, fmt.Errorf("new static policy error: %v", err)
diff --git a/pkg/kubelet/cm/node_container_manager_linux.go b/pkg/kubelet/cm/node_container_manager_linux.go
index fc7864dacc2..78d8c7bbde6 100644
--- a/pkg/kubelet/cm/node_container_manager_linux.go
+++ b/pkg/kubelet/cm/node_container_manager_linux.go
@@ -211,6 +211,7 @@ func (cm *containerManagerImpl) getNodeAllocatableInternalAbsolute() v1.Resource
 func (cm *containerManagerImpl) GetNodeAllocatableReservation() v1.ResourceList {
        evictionReservation := hardEvictionReservation(cm.HardEvictionThresholds, cm.capacity)
        result := make(v1.ResourceList)
+       overSubscriptionCapacity := resource.NewMilliQuantity(1000, resource.DecimalSI)
        for k := range cm.capacity {
                value := resource.NewQuantity(0, resource.DecimalSI)
                if cm.NodeConfig.SystemReserved != nil {
@@ -222,6 +223,9 @@ func (cm *containerManagerImpl) GetNodeAllocatableReservation() v1.ResourceList
                if evictionReservation != nil {
                        value.Add(evictionReservation[k])
                }
+               if k == "cpu" {
+                      value.Sub(*overSubscriptionCapacity)
+                }
                if !value.IsZero() {
                        result[k] = *value
                }

```

### Results
It can be seen that oversubscription capacity (1 cpu) has been added to node allocatable ( Capacity = 104, reserved = 4, oversubscription = 1, allocatable = capacity - reserved + oversubscription = 104 - 4 + 1 = 101)
Node report:
```
Capacity:
  cpu:                104
...
Allocatable:
  cpu:                101
```
## Productization aspects
### Provisioning of the oversubscription parameter
The approach of providing the node a fixed amount of oversubscription capacity requires a new parameter provision to the kubelet. This new parameter must be added to both kubelet command line arguments and a corresponding parser should be added to the KubeletConfig resource handler. The parameter should be optional and can be set to zero if undefined.
The deployment will look as follows:
```
apiVersion: machineconfiguration.openshift.io/v1
kind: KubeletConfig
spec:
  kubeletConfig:
    cpuOverSubscriptionCapacity: 1.5

```
### Visibility
#### Node report
The oversubscription capacity will alter the node reports and can potentially mislead cluster administrators and developers. Therefore it is essential that oversubscription capacity will be reflected in the node report, for example:
Capacity:
  cpu:                     104
  cpu oversubscription:    1.000                    
#### Logging
Kubelet logs reporting Allocatable, Capacity and Reservations should include the new parameter where applicable.
### Testing
Additional testing will be required to verify that oversubscription parameter addition does not cause regression in different configuration cases (at least all permutations of kube / system / explicitly reserved)
### Documentation
The parameter must be documented in the official documentation
### Upgrades and downgrades
The oversubscription capacity parameter should be seamless for upgrades and downgrades. Need to verify that kubelet doesn’t crash in case of an unknown command line argument.

enhancements/automated-resource-request-scaling.md
dhellmann: I clarified that the goal of this enhancement is to only address CPU, and that we would look at approaches for other compressible resources in other enhancements. I've given the doc a new title, but was reluctant to rename the file and lose all of the review threads. I will do that before we merge it, or in another PR.

enhancements/automated-resource-request-scaling.md
dhellmann: I added some text to cover OLM operators in this section.

enhancements/automated-resource-request-scaling.md
dhellmann: I updated the goals and non-goals lists to clarify the scope based on this thread.

enhancements/automated-resource-request-scaling.md
dhellmann: Thanks for the excellent level of detail here. I added this to the implementation history section, since I thought it fit better there.

dhellmann:
I have updated the enhancement based on the outcome of the meeting from earlier today, to include the fact that we are currently considering 2 implementation options.

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,855 @@
...
+default settings.
+
+Operators from versions that do not support scaling will ignore the
+new field until they are upgraded.
+
+The value of `DefaultControlPlaneCPUCount` constant may change between
+releases. The value from the library will be vendored into each
+operator, so the scaling formula for a given version of an operator
+will be correct at all times during an upgrade, even if different
+operators are using different base values.
```
wking, line 404: I don't think this is quite right.  Based on your goal:

> The current resource requirements for an OpenShift control plane exceed the size desired by some customers running on bare metal.

where you have *requirements*, not *requests*, and [this comment][1], the current proposal would result in:

1. Update initiated.
2. New CVO comes in, with new `DefaultControlPlaneCPUCount`.  Say it's lower.  CVO starts working through its [manifest graph][2].
3. CVO gets to operator A's Deployment, and it updates to match the new manifest, with the scaled request value.
4. New operator A pod comes up with its new `DefaultControlPlaneCPUCount`.  It fetches the Infrastructure object, sees a low `controlPlaneCPULimit`, and says "ooh, I'm really going to have to reign it in", and triples some poll periods and adds some sleeps or whatever.

Having the `request` reduction in (3) match the actual-consumption reduction from (4) means that we really want `DefaultControlPlaneCPUCount` to be uniform throughout the cluster.  That's a lot of components that will need to bump vendored libraries to pick this up; I'm not convinced we won't screw it up ;).

And more broadly, there is going to be some resource consumption drop in most processes just because their single-node workload is lighter than their Standard Cluster workload.  But this proposal is not asking the process to figure out single-node-ness.  If the process throttles back polling periods and whatnot by the amount needed to take its Standard Cluster CPU consumption down by the requested amount, and then we run it with single-node workloads, it may end up starving itself more than it needs to, resulting in even higher latencies, etc. than necessary.  Without something actually looking at measured consumption in order to tune whatever the in-process knobs end up being ("ah, I'm over my `requests` more often than under them in this past second, better reign it in"), it seems like it is going to be hard to reliably approach the `controlPlaneCPULimit` target.

[1]: https://github.com/openshift/enhancements/pull/628#discussion_r570578301
[2]: https://github.com/openshift/cluster-version-operator/blob/dfc8494a4ed9b4e68d7511ae989cbf3225d65e65/docs/user/reconciliation.md#manifest-graph

enhancements/automated-resource-request-scaling.md
wking: Bursty comment you have in the current draft works for me.  I'd still like to see a breakdown of:

1. Current requests.
2. Current consumption under Standard Cluster workloads (whatever we feel like a standard cluster is 🤷).
3. Current consumption under single-node workloads.
4. Ideal consumption under single-node workloads (e.g. with slower polling, or whatever [you expect containers to do][1] to further reduce their consumption below 3).

Before and after this sort of consumption stabilization effort.  Because my naive gut still feels like 2 and 3 are going to end up pretty close together for most containers, in which case once we align 1, we don't need a mechanism for tuning it per-cluster.  But I don't think we need to hold this thread open for that; we can pick it up in [this new thread][2] ;)

[1]: https://github.com/openshift/enhancements/pull/628#discussion_r570578301
[2]: https://github.com/openshift/enhancements/pull/628#discussion_r579631867

enhancements/automated-resource-request-scaling.md
dhellmann: I thought a `const` in a library would expose fewer implementation details, but your counterargument is good. If we exposed `DefaultControlPlaneCPUCount` via the Infrastructure API, what would be responsible for updating it during an upgrade? Do we have anything that writes to that API? Should we use a different resource, so we can ship a static manifest with the value and have the CVO apply it as part of the upgrade?

@browsell correct me if I'm wrong, but I thought this was mostly about being able to schedule the workloads, and that they would be pinned to CPUs to throttle them without individual components having to constantly adjust their resource consumption.

As far as scaling the operators requests, I had been expecting the work done to reduce the resource constraints on operators deployed by the CVO to be enough that the CVO would not need to take further action, so the API would only be used by those operators but not "on" them. Maybe that's not a safe assumption, though.

enhancements/automated-resource-request-scaling.md
dhellmann: @browsell does your team have the data @wking is asking for? If you can share it with me, I can update the enhancement.

enhancements/automated-resource-request-scaling.md
browsell: Target application for this deployment model would mostly use guaranteed/exclusive cpus. The expectation for the remaining applications pods (i.e. burstable) is that they would be tightly engineered to optimize the cpu footprint.

dhellmann: > Target application for this deployment model would mostly use guaranteed/exclusive cpus. The expectation for the remaining applications pods (i.e. burstable) is that they would be tightly engineered to optimize the cpu footprint.

When you say "application pods" do you mean only user-provided workloads that are not part of OpenShift or other add-ons that make up "the platform"?

browsell: User provided workloads that are not part of openshift

dhellmann: OK, good. So, I think @wking's question applies to things that are part of OpenShift, and how we ensure that those actually fit within the number of CPUs allowed. We've been talking about scaling the requests, but the concern is about components that use more CPU than they request, at least some of the time. Do we anticipate the scaling behavior, in whatever implementation, to provide enough overhead to support those spikes in use without excessive throttling leading to latency issues?

browsell: On average the actual usage of openshift pods would be less than the targeted number of allowed cpu's. There are a small number of spikes that would exceed it. We should consider providing some overhead for spikes but at the same time I don't think we should engineer for the worst case spike, rather we should look at the cause and address it.

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,871 @@
...
+their own workloads. OpenShift currently requires 7.
+
+### Implementation Details/Notes/Constraints
+
+1. Add the new field to the `Infrastructure` API.
+2. Expose the `controlPlaneCPULimit` setting through the installer via
+   `install-config.yaml`.
+3. Update operators to respond to the API, as described above.
+
+### Risks and Mitigations
```
dhellmann, line 254: @browsell as we discussed earlier today, I've added info about starting by not updating all operators at once. Rather than listing it as a complete separate 3rd option, I have it here as mitigation of the risk of the amount of effort of implementing the same approach across all operators. Please let me know what you think.

browsell: Thanks. lgtm

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,871 @@
...
+   changes.
+
+   Some [work to reduce default
+   requests](https://bugzilla.redhat.com/show_bug.cgi?id=1920159) has
+   been done in the 4.8 release.
+
+3. Have a new operator in the release payload perform the scaling for
+   control plane operators
+
+   It would use the new API field to determine how to scale (see above).
```
cynepco3hahue, line 735: We already have some operators for it - https://github.com/openshift/cluster-resource-override-admission-operator, we can improve it and add new API fields.

dhellmann: OK, so we can modify this section to say it doesn't need to be in the release payload but that it would need to be installed separately and link to the existing operator.

We'll also need to say that each openshift operator will need to make the appropriate adjustments to ignore changes to the resource requests to avoid contention with the cluster-resource-override-admission web hook. It looks like the way to enable the admission web hook is to add a label to the namespace and then update the code to ignore resource requests on operands when comparing for equality. Is that right?

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,886 @@
...
+   1. Kubelet could accept an option to report a static amount of extra
+      capacity, for example always reporting 2 more CPUs than are present.
+   2. Kubelet could not subtract the requests from some workloads, or
+      only subtract part of the requests, leaving the capacity higher
+      than it would if the request was included fully in the
+      calculation. *See implementation history above for details about
+      a proof-of-concept implementation of this approach.*
+   3. Kubelet could add capacity to match the requests of some workloads,
+      so that the apparent capacity of a node grows dynamically when
+      system workloads are added.
```
dhellmann, line 855: Today we discussed this approach, using a regex or list of regexes to identify namespaces to which the behavior would apply. Namespaces give us RBAC, which prevents normal users from abusing the behavior to circumvent quota or DOS the system.

enhancements/automated-resource-request-scaling.md
```diff
@@ -0,0 +1,886 @@
...
+operators before they will be able to tune the over-subscription
+setting so that all the components will fit onto the number of CPUs
+they require. We will need to keep the documentation of minimum CPU
+resource requests for OpenShift up to date for each release, or
+provide a tool to calculate the value from a default cluster.
+
+Refer to the implementation history section below for details of a
+proof-of-concept implementation of this approach.
+
+#### Cluster-wide resource request limit API
```
dhellmann, line 149: Today we discussed a variation of this approach in which we do not add a new API but use the topology APIs to have operators reconfigure their operands to alternative static values. We would likely start with the top consumers and not try to update all operators in one release. This approach gives us more time to consider what a better scaling API should look like.

dhellmann:
The latest draft of this enhancement reflects the changes to the approach that we came up with today, using https://github.com/openshift/cluster-resource-override-admission. There are still some details to work out, but I think the basics are all here now. All of the old approaches have been folded into the alternatives section at the end of the document.

@derekwaynecarr @browsell @markmc please take a look and let me know if I've missed anything.

enhancements/automated-resource-request-scaling.md
dhellmann: @cynepco3hahue  The latest draft elevates the cluster-resource-override-admission-operator to the primary approach being proposed. Let me know what you think.

dhellmann:
I forgot to mention that I left the changes in a separate patch in case that made reviewing easier. I will squash the commits before we merge this enhancement.

enhancements/automated-resource-request-scaling.md
```diff
@@ -247,10 +151,55 @@ their own workloads. OpenShift currently requires 7.
...
+so they are recreated and the admission webhook can modify their
+settings.
+
+An additional benefit of using the separate component is that it is
+not part of the OpenShift release payload, and therefore could be
+released independently of OpenShift releases. Since we are looking at
+a tight timeline for delivering the work described in this
+enhancement, that may help with scheduling.
+
+1. Add a new field to `ClusterResourceOverride` API with a name like
```
browsell, line 187: For a general case where a pod has a limit set, would need to decide the precendence if both cpuRequestPercent and cpuRequestToLimitPercent are set.

dhellmann: Good point. It seems like the limit-based scaling (the existing behavior) should take precedence, and the new field should be ignored if the old field is present. What do you think?

browsell: Agreed.

deads2k:
This enhancement doesn't appear to mention the runlevel restrictions created in the clusterresourceoverride's MutatingWebhookConfiguration which allows us to ensure re-bootstrapping a cluster that has been shut down.  We label namespaces in openshift with `openshift.io/run-level=0` and `openshift.io/run-level=1` to indicate namespaces that are required to have pods before admission webhooks can be called.

This includes 36 out of 61 openshift-* namespaces on our 4.7 CI cluster.  Removing this restriction to shrink the pod resource requests/limits would be fraught.

One example flow that will work in a demo and be at risk in a production cluster is 
1. install cluster
2. stop skipping runlevel=0 namespaces.  this will include the SDN and kube-apiserver namespaces, among others
3. delete pods in these namespaces and they will be restarted successfully
4. shutdown the cluster
5. start the cluster
6. there is no SDN running
7. etcd starts from a static pod
8. the kube-apiserver starts from a static pod
9. the kube-controller-manager starts from a static pod
10. the kube-controller-manager tries to create a daemonset pod for the SDN
11. the kube-apiserver enforces admission webhook configuration, which tries to call the cluster-resource-overrides admission plugin
12. there is no pod running for the cluster-resource-overrides admission plugin and even if there was, it would unreachable because the SDN pod doesn't exist
13. the cluster is live-locked without the ability to run pods

We created the runlevels and skips to ensure that we did not wedge in such a situation.  While it may be possible to tune the labeling slightly tighter for a single-node use case (I don't know one way or the other), having an accident at this level can result in situations where clusters cannot be safely restarted.

Knowing that more than half of the namespaces we have today will not be restricted and knowing that we need a way to scale resource requests and limits on pods in these 36 namespaces and on static pods, having a single solution for the additional  25 namespaces doesn't seem so bad.

dhellmann:
Closing this in favor of #703

PR #635, opened 2021-02-08
IR-57: API changes for manifest list support

ricardomaraschini:
This commits adds a proposal for Openshift's manifest list support.

### DIFF
--- a/enhancements/manifestlist/manifestlist-support.md
+++ b/enhancements/manifestlist/manifestlist-support.md
@@ -360,6 +360,62 @@ $ oc get isimage <imagestream>@<sha of any of the child images>
 The output of the `oc describe isimage` command line will need to be changed to display
 information present in the DockerImageManifests property of the parent Image.
 
+### Proposed pruner changes
+
+Pruner gathers information about images in use within a cluster by inspecting the following
+objects:
+
+- Pods
+- Replication controllers
+- Deployment configs
+- Replica sets
+- Deployments
+- Daemon sets
+- Builds
+- Build configs
+- Stateful sets
+- Jobs
+- Cronjobs
+
+If a given image is in use by any of these entities it is not pruned. This operation will need
+to be refactored to take into account the extra level of indirection that the "child images"
+layer brings. For example, in a scenario where we have the following:
+
+```text
+ImageStream X -> Image A -> Image B
+                            Image C
+                            Image D
+
+```
+
+If a reference for, let's say, `Image D` exists in the cluster pruner will maintain `Image A`,
+`Image B`, `Image C`, and `Image D` as well. The same logic also applies when deleting tags from
+an `ImageStream` during prune: all child images are also deleted whenever a tag ceases to exist.
+
+The extra cost for this indirection analysis will be paid during the initial evaluation of images
+in use. During this operation, we already have all Images and ImageStreams available in memory and
+during evaluation, we fill in two maps indicating the images in use and who is using them using
+a struct like the following:
+
+```text
+p.usedImages = map[imageStreamImageReference][]resourceReference{}
+```
+
+By considering the above example scenario again (Pod called `pod0` using `Image D`), after
+the initial evaluation we would have the following resulting struct (edited for easier reading):
+
+```text
+p.usedImages = {
+	"Image A": [ "pod0" ],
+	"Image B": [ "pod0" ],
+	"Image C": [ "pod0" ],
+	"Image D": [ "pod0" ]  <- the actual image in use
+}
+```
+
+This would ensure that we treat the whole manifest list as a single entity at a cost of
+maintaining in the cluster images that may not be in use but are part of a manifest list.
+
 ### Test Plan
 
 #### Validate we can deploy manifest lists
--- a/enhancements/manifestlist/manifestlist-support.md
+++ b/enhancements/manifestlist/manifestlist-support.md
@@ -379,7 +379,8 @@ objects:
 
 If a given image is in use by any of these entities it is not pruned. This operation will need
 to be refactored to take into account the extra level of indirection that the "child images"
-layer brings. For example, in a scenario where we have the following:
+layer brings. For example, in a scenario where we have a tag in ImageStream X pointing to a
+manifest list A containing images B, C, and D:
 
 ```text
 ImageStream X -> Image A -> Image B
--- a/enhancements/manifestlist/manifestlist-support.md
+++ b/enhancements/manifestlist/manifestlist-support.md
@@ -1,0 +1,410 @@
+---
+title: manifest-list-support
+authors:
+  - "@ricardomaraschini"
+reviewers:
+  - "@dmage"
+approvers:
+  - "@dmage"
+creation-date: 2021-02-03
+last-updated: 2021-02-03
+status:
+see-also:
+replaces:
+superseded-by:
+---
+
+# Add support to manifest list image imports
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+A manifest list is a group of image manifests, it is used to keep track of references to images
+targeting different platforms/architectures in a single entity. OpenShift does not support this
+type of manifest, it currently supports only a 1 to 1 mapping (one manifest pointing to one
+image). Manifest lists are identified by their media types:
+
+```text
+application/vnd.docker.distribution.manifest.list.v2+json
+application/vnd.oci.image.index.v1+json
+```
+
+Their representation is very simple, it contains (aside from the media type and schema version)
+an array of manifests with their respective platform information:
+
+```json
+{
+  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
+  "schemaVersion": 2,
+  "manifests": [
+    {
+      "digest": "sha256:0b159cd1ee1203dad901967ac55eee18c24da84ba3be384690304be93538bea8",
+      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+      "platform": {
+        "architecture": "amd64",
+        "os": "linux"
+      },
+      "size": 1362
+    },
+    {
+      "digest": "sha256:0492a0bfe6585aee9f80b476b8d432a03968bb0a2b67d00a340cac69956e6c50",
+      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+      "platform": {
+        "architecture": "arm",
+        "os": "linux",
+        "variant": "v5"
+      },
+      "size": 1362
+    }
+  ],
+}
+```
+
+This enhancement proposal designs a plan to add support for this new type of manifest in
+OpenShift.
+
+## Motivation
+
+When running on multi-architecture clusters it makes sense to be able to allow users to work with
+images that support different architectures as well. By adding support for manifest lists we are
+also adding support for hosting these multi-architecture images in OpenShift's internal registry.
+
+### Goals
+
+- Allow users to import manifest lists using the `oc` client.
+- Allow users to execute pull-through of manifest lists.
+- Make image pruner aware of manifest lists.
+- Allow users to push manifest lists into the integrated registry.
+- Improve the `oc describe` command to present manifest list info.
+
+### Non-Goals
+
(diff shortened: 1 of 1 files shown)


enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,277 @@
...
+
+Images are indexed (named) in OpenShift by the hash of their content. Currently
+an Image is represented by the following structure:
+
+```go
+type Image struct {
+	metav1.TypeMeta
+	metav1.ObjectMeta
+	DockerImageReference string
+	DockerImageMetadata DockerImage
```
dmage, line 187: `DockerImageMetadata` is `runtime.RawExtension`: https://github.com/openshift/api/blob/a9e731090f5ed361e5ab887d0ccd55c1db7fc633/image/v1/types.go#L41

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,277 @@
...
+	OS           string
+	Variant      string
+}
+```
+
+The Image pointed by the `Digest` property contain all regular information
+(i.e. Config) we can see nowadays in a regular Image.
+
+Not all fields in an Image are relevant for a manifest list, therefore other
+proposed change is to add an "omitempty" flag on `DockerImageLayers` property
```
dmage, line 222: I don't think this is a backward compatible change. Old clients (oc v4.(y-1).z) may expect them to be always present.

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,277 @@
...
+to end test. This test would execute the following operations:
+
+1. Deploy a temporary image registry.
+2. Publish a manifest list containing images for different platforms.
+3. Create an ImageStream object to import the published manifest list.
+4. Create a Deployment using the ImageStream.
+
+A second test would be very similar to this one but at this time leveraging
+pull through. Podman currently supports managing manifest lists and we can
+use its internals for step 2.
```
dmage, line 257: We also need to test that a manifest list can be mirrored into the integrated registry using `oc image mirror`.

dmage (changes requested):
Will we have changes in ImageStreamImport? ImageStreamLayers? Will it be possible to get ImageStreamImages only for top-level SHAs (i.e. for image A) or will it work for children manifests too (i.e. for images B, C, D)?

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Goals
+
+- Allow users to import manifest lists using `oc image-import`.
+- Allow users to execute pull through using manifest lists.
+- Make image pruner aware of manifest lists.
+
+### Non-Goals
+
+- Modify OpenShift's web console to support manifest lists.
```
adambkaplan, line 89: Another non-goal - Multiarch OpenShift builds which push a manifest list as output.

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Test Plan
+
+Other to regular unit tests we gonna need to back this feature with an end
+to end test. This test would execute the following operations:
+
+1. Deploy a temporary image registry.
+2. Publish a manifest list containing images for different platforms.
+3. Create an ImageStream object to import the published manifest list.
+4. Create a Deployment using the ImageStream.
```
adambkaplan, line 253: I think builds need to be included here as well. I'm pretty sure buildah can handle building from an image that references a manifest list, but we should verify that.

enhancements/manifestlist/manifestlist-support.md
ricardomaraschini: Added, thanks.

enhancements/manifestlist/manifestlist-support.md
ricardomaraschini: "Make sure a build that leverages an image using ManifestList works"
Something like this?

enhancements/manifestlist/manifestlist-support.md
ricardomaraschini: Added.

enhancements/manifestlist/manifestlist-support.md
ricardomaraschini: What is the main source of concern here?

1. `DockerImageLayers` being an empty slice or
2. `DockerImageLayers` not even being presented

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,277 @@
...
+When running on multi architecture clusters it makes sense to be able to allow users
+to import images that support different architectures as well. By adding support for
+manifest lists we are also adding support for hosting these multi architecture images
+in the internal registry.
+
+### Goals
+
+- Allow users to import manifest lists using `oc image-import`.
+- Allow users to execute pull through using manifest lists.
+- Make image pruner aware of manifest lists.
```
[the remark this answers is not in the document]
ricardomaraschini: Added.

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Risks and Mitigations
+
+#### Console usage of ImageStreamImport
+
+OpenShift console uses ImageStreamImport to obtain information about images
+in order to present them to users. When dealing with manifest lists not all
+information will be presented back to the console (the Config portion of an
+image is absent in a manifest list). Depending on how manifest lists support
+is going to be implemented in the console a second request to the API may be
```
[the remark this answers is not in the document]
ricardomaraschini: Could we tag someone from Console team's here? I need to get someone from that side involved.

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,404 @@
...
+application/vnd.oci.image.index.v1+json
+```
+
+Their representation is very simple, it contains (aside from the media type and schema version)
+an array of manifests with their respective platform information:
+
+```json
+{
+  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
+  "schemaVersion": 2
```
dmage, line 45:
```suggestion
  "schemaVersion": 2,
```
nit

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,404 @@
...
+enhancement proposes the addition of another field where child images (manifests within a
+manifes list) would be listed, leaving `ImageImportStatus` struct looking like:
+
+
+```go
+type ImageImportStatus struct {
+	Status metav1.Status
+	Image *Image
+	Tag string
+	DockerImageManifests []Image
```
dmage, line 312:
```suggestion
	Manifests []Image
```
or something like that, we don't want to have brand names in our API.

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,404 @@
...
+So that I can keep my storage resource usage under control
+
+### Implementation Details/Notes/Constraints
+
+#### On pushing manifests into integrated registry
+
+Openshifft's internal registry will need to accept image pushes by SHA, not only by tag. The
+difference here is that the registry is not supposed to create an Image object every time it
+sees a push by SHA but only when it sees a push by tag. In the latter scenario the registry may
+need to create more than one image (if the pushed manifest is of type manifest list).
```
dmage, line 149: Risk: the pruner won't see these images, the hard pruner doesn't know about its age and whether this manifest has been just uploaded for a manifest list.

dmage:
I think this proposal is ready for a review by a broader audience.

/assign @bparees @smarterclayton

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,403 @@
...
+
+As an OpenShift user
+I want to be able to automatically clean unused manifest lists
+So that I can keep my storage resource usage under control
+
+### Implementation Details/Notes/Constraints
+
+#### On pushing manifests into integrated registry
+
+Openshifft's internal registry will need to accept image pushes by SHA, not only by tag. The
```
bparees, line 146:
```suggestion
Openshift's internal registry will need to accept image pushes by SHA, not only by tag. The
```

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,403 @@
...
+
+As an OpenShift user
+I want to be able to push manifest lists to Openshift's internal registry
+So that I can publish my multi-architecture images
+
+#### Support for manifest lists during prune
+
+As an OpenShift user
+I want to be able to automatically clean unused manifest lists
+So that I can keep my storage resource usage under control
```
bparees, line 140: need more detail on how prune will work.  If any image within the manifestlist is in use, will the full manifestlist be kept around?  Or only if the manifestlist itself is referenced the manifestlist will be kept around?

or something in between?  (if 2 of 3 images in the manifestlist are in use, the manifestlist is kept around but the unused image is removed from the manifestlist)

enhancements/manifestlist/manifestlist-support.md
bparees: yeah.  more importantly, "using a manifestlist hosted in the internal registry"  (it better already work for manifestlists in general)

ricardomaraschini: I have added a new section with a test to validate this. Thanks.

enhancements/manifestlist/manifestlist-support.md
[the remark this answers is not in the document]
ricardomaraschini: Added.

enhancements/manifestlist/manifestlist-support.md
ricardomaraschini: I think that the best way of dealing with this would be to maintain the whole struct consistent by not removing anything that is part of the manifest list.  I have added a [section](https://github.com/openshift/enhancements/pull/635/files#diff-b4a80ec15b1d903d24929518bc457be2aea5a1ba81d43eccd586314da952a168R363) to the proposal specifying this, PTAL.

/cc @dmage you may have some input on this regard.

bparees: keeping the entire manifestlist intact if any image it contains is ref'd makes sense to me.

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,466 @@
...
+
+If a given image is in use by any of these entities it is not pruned. This operation will need
+to be refactored to take into account the extra level of indirection that the "child images"
+layer brings. For example, in a scenario where we have the following:
+
+```text
+ImageStream X -> Image A -> Image B
+                            Image C
+                            Image D
+
```
bparees, line 388: it's not clear to me what the relationship is between A,B,C,and D here.

are they all part of the same manifestlist?  is B the base image for A?  Is A a manifestlist that contains B,C, and D?

I think it's that last option, but it'd help to clarify the scenario.

ricardomaraschini: Yes, you are right. I will clarify it. The last option is the correct one.

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,467 @@
...
+Both of the commands below would be valid, the first command would not return that much
+information but it works as a means of obtaining a list of all child images:
+
+```console
+$ oc get isimage <imagestream>@<sha of the parent image>
+$ oc get isimage <imagestream>@<sha of any of the child images>
+```
+
+The output of the `oc describe isimage` command line will need to be changed to display
+information present in the DockerImageManifests property of the parent Image.
```
dmage, line 361: /cc @soltysh 

fyi, we want to update `oc describe` for some image related types

enhancements/manifestlist/manifestlist-support.md
```diff
@@ -0,0 +1,467 @@
...
+	Status metav1.Status
+	Image *Image
+	Tag string
+	Manifests []Image
+}
+```
+
+Here the image that refers to the manifest list would be presented on `Image` property (as it is
+today for regular manifests) while all inner manifests within the manifest list would be listed
+in the `Manifests` slice.
```
dmage, line 318: /cc @spadgett 

fyi, ImageStreamImports are going to show more information about manifest lists, afaik Console uses them

enhancements/manifestlist/manifestlist-support.md
soltysh: No objections from my end.

dmage:
No objections, so let's merge it.
/lgtm

PR #636, opened 2021-02-08
API Removal Notifications

sanchezl:


enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,159 @@
...
+
+Kubernetes is being more aggressive with regards to the removal of APIs that have been deprecated or pre-release APIs that have reached a dead-end and will never reach GA. Additionally, the ability to re-enable a previously removed API has been removed.
+
+Customers should be notified that an API that will be removed in the next release is in use so that they can take action to avoid failure when upgrading to the next release. 
+
+### Goals
+
+1. Notify customers an API that will be removed in the next release is in use.
+2. Document how to identify the workload that is using the API to be removed.
+
```
sttts, line 41: I would expect a goal for EUS here.

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,159 @@
...
+```go
+// +k8s:prerelease-lifecycle-gen:introduced=1.1
+// +k8s:prerelease-lifecycle-gen:deprecated=1.8
+// +k8s:prerelease-lifecycle-gen:removed=1.18
+// +k8s:prerelease-lifecycle-gen:replacement=apps,v1,Deployment
+```
+
+The `apiserver_requested_deprecated_apis` metrics provides the metadata needed by this enhancement to "look into the future" and identify APIs, currently in use, that will be removed in the next release.
+
+In order to provide the metadata needed to look forward to the next EUS release, `prerelease-lifecycle-gen` tags updated between EUS releases will need to be back-ported to the latest EUS release.
```
sttts, line 83: same metric will do?

enhancements/kube-apiserver/stability-api-removal-notifications.md
sttts: What about filtering workloads by Redhat-provided and user workloads. Can we use APIs in a EUS release in our component that will be gone in EUS+1 ? I would expect this either in goals or in non-goals.

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,159 @@
...
+
+1. Automatically identify the workload using the API to be removed.
+2. Prevent the customer from attempting to upgrade when an API that will be removed in the next release is in use.
+3. Notify customers that an upstream API is not supported on OpenShift.
+
+## Proposal
+
+Fire an `APIRemovedInNextReleaseInUse` alert when an API that will be removed in the the next release in in use.
+
+On EUS releases, also fire an `APIRemovedInNextEUSReleaseInUse` alert when an API that will be removed in the next extended support release is in use.
```
sttts, line 52: what should be the behaviour of these. Will they be conditional in some sense? I.e. non-EUS customers won't see the second alert?

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,159 @@
...
+#### Determinine Workload
+
+Documentation will be provided on how to use the audit logs to identify the workload using the API that will be removed.
+
+
+### Risks and Mitigations
+
+#### Storage Version Migration
+
+During an EUS upgrade, where multiple "normal" releases are applied back-to-back, there is a risk that there are persisted instances of resources that can longer be decoded from storage due to the removal of the storage encoder for a removed version. We should either provide customers with documentation on upgrading storage versions or provide a storage version migration tool to perform the migrations automatically.
```
sttts, line 115: IMO kube-apiserver operator should make sure that this migration is done, behind the scenes.

enhancements/kube-apiserver/stability-api-removal-notifications.md
sanchezl: Updated Goal/Non-Goals.
> What about filtering workloads by Redhat-provided and user workloads.
This cannot currently be done if relying on alerts for notifications due to the lack of client information in the metrics. Maybe something for a followup enhancement.

enhancements/kube-apiserver/stability-api-removal-notifications.md
sanchezl: Yes. The `removed_release` label will be enough to distinguish between next release vs next EUS release.

enhancements/kube-apiserver/stability-api-removal-notifications.md
sanchezl: > non-EUS customers won't see the second alert?
@sttts AFAIK there is no "EUS flag" available. So non-EUS customers on an EUS compatible release (4.6 or 4.10), would also see this alert.

Without an "EUS flag", the alert definition would be added/removed in the specific release branches.

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,162 @@
...
+Documentation will be provided on how to use the audit logs to identify the workload using the API that will be removed.
+
+### Risks and Mitigations
+
+#### Storage Version Migration
+
+During an EUS upgrade, where multiple "normal" releases are applied back-to-back, there is a risk that there are
+persisted instances of resources that can longer be decoded from storage due to the removal of the storage encoder for a
+removed version. We should either provide customers with documentation on upgrading storage versions or provide a
+storage version migration tool to perform the migrations automatically.
```
deads2k, line 126: We won't ask a customer to do this for openshift/api or k8s.io/api.  We will handle it automatically if it comes up for those two repos.

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,162 @@
...
+removed.
+
+Customers should be notified that an API that will be removed in the next release is in use so that they can decide if
+there is a need to take action in order to ensure success when upgrading to the next release.
+
+### Goals
+
+1. Notify customers that an API that will be removed in the next release is in use.
+2. Notify EUS customers that an API that will be removed in the next EUS release is in use.
+3. Document how to identify the workload that is using the API to be removed.
```
deads2k, line 42: I want a section exploring at least two possibilities of how this could be done at an API level so we don't have to have a manual process.  I think I'll lose my mind trying to fix these if I have to manually dig into audit logs.

enhancements/kube-apiserver/stability-api-removal-notifications.md
deads2k: > What about filtering workloads by Redhat-provided and user workloads.

how do you know the difference?

sttts: > how do you know the difference?

That's the $100,000 question to make these alert usable.

enhancements/kube-apiserver/stability-api-removal-notifications.md
sttts: Has @sdodson 's EUS enhancement specified this "flag"?

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,162 @@
...
+Kubernetes is being more aggressive with the removal of APIs that have been deprecated or pre-release APIs that have
+reached a dead-end and will never reach GA. Additionally, the ability to re-enable a previously removed API has been
+removed.
+
+Customers should be notified that an API that will be removed in the next release is in use so that they can decide if
+there is a need to take action in order to ensure success when upgrading to the next release.
+
+### Goals
+
+1. Notify customers that an API that will be removed in the next release is in use.
```
sdodson, line 40: Is there a reason this cannot be expanded to the full three minor versions required of the deprecation policy? Are we concerned that there's too much capitulation here, as in we'd say something is removed in 1.24 today but there's risk that it is not? To me that's tolerable, we'd just update the ruleset when we become aware of that change and it's only an Info alert.

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,162 @@
...
+2. Notify EUS customers that an API that will be removed in the next EUS release is in use.
```
sdodson, line 41: If we get the three minor versions notification in point number one, then this just becomes backporting from data from 4.7 to 4.6 in the case of 4.6 and 4.10 being EUS versions.

enhancements/kube-apiserver/stability-api-removal-notifications.md
sdodson: I don't think it matters. We're responsible for cleaning up our mess in the platform, ideally with CI jobs failing on alerts firing. They're responsible for cleaning up their workloads mess, as long as we make it clear which namespace the offending api use is present in is that good enough? I say this assuming that this is knowable with very little knowledge of how this is to be implemented.

Perhaps we also make this a criteria for being in the Red Hat Marketplace, no OLM operators get approved if they're using deprecated APIs.

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,162 @@
...
+3. Document how to identify the workload that is using the API to be removed.
+
+### Non-Goals
+
+1. Automatically identify the workload using the API to be removed.
+2. Prevent the customer from attempting to upgrade when an API that will be removed in the next release is in use.
```
sdodson, line 47: I think this should become a goal, if an object just vanishes during upgrade then we should inhibit that upgrade. So Info level alerts on anything removed in N through N+3, Upgradeable=False on removal in N+1.

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,162 @@
...
+3. Notify customers that an upstream API is not supported on OpenShift.
+
+## Proposal
+
+Fire an `APIRemovedInNextReleaseInUse` alert when an API that will be removed in the the next release in in use.
+
+On EUS releases, also fire an `APIRemovedInNextEUSReleaseInUse` alert when an API that will be removed in the next
+extended support release is in use.
```
sdodson, line 55: If we were to switch to Alerts over N+3 and Upgradeable=False on N+1 then I think we'd change this to simply be `APIRemovedInFutureReleaseInUse` with a message that rolls up all of the APIs in use and their planned removal release.

sdodson:
My view is that I'd prefer we provide notifications as far out as we're sure of. That seems to be N+3 in general and in the case of EUS that would mean that we'd need to backport updated rules to 4.6 at the time that 4.9 goes GA, that specific date is chosen because that's the point at which 4.6 would EOL if not for EUS. That means that anyone upgrading into a 4.6.z after that point is an EUS cluster. In conjunction with that we will require that in order to upgrade along the EUS-to-EUS path they upgrade to a 4.6.z that's at least as new as 4.9 GA date.

While I think notifying on N+3 is the right thing to do, I'm fine with it being N+1 for now as long as it's easy to change in the future and we agree to the EUS specific backporting I mention above.

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Drawbacks
+
+## Alternatives
+
+### Console Recommendations vs. Alerts
+
+Alerts are not ideal for these notifications as they are not necessarily immediately actionable. There is the
+possibility in the future of moving from using alerts to "console recommendations", but "console recommendations" are
+still in the request for enhancement stage.
```
dhellmann, line 152: Are console recommendations a GUI-only thing?

sanchezl: Reach out to @cshinn, he might have more information.

enhancements/kube-apiserver/stability-api-removal-notifications.md
deads2k: While building a mechanism to see those is reasonable, we think that building a mechanism to alert on them, even at an info level, would teach customers that we alert on things that won't be a problem for a year+ and teach them to ignore these warnings.

Given the API resources being added in this design, there is a mechanism for an interested party to look at usage of APIs slated for future removal.

enhancements/kube-apiserver/stability-api-removal-notifications.md
deads2k: > If we get the three minor versions notification in point number one, then this just becomes backporting from data from 4.7 to 4.6 in the case of 4.6 and 4.10 being EUS versions.

I don't anticipate making that change, since it will produce noise that is unnecessary.  If you have a signal for determining a customer planning on using 4.6 as an EUS, then we could reconsider, but without such a signal looking three releases out will simply produce noise.

enhancements/kube-apiserver/stability-api-removal-notifications.md
deads2k: > I think this should become a goal, if an object just vanishes during upgrade then we should inhibit that upgrade. So Info level alerts on anything removed in N through N+3, Upgradeable=False on removal in N+1.

I'm willing to debate that on a larger arch call, but without the ability to selectively override conditions for an upgrade, such a condition is likely to become a regularly performed and indiscriminately used --force.

For the purposes of this enhancement, an non-goal is a very reasonable starting point to get the majority of development done.

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,293 @@
...
+
+Example:
+
+```yaml
+version: api.openshift.io/v1
+kind: DeprecatedAPIRequest
+meta-data:
+  name: 'flowschemas.v1alpha1.flowcontrol.apiserver.k8s.io'
+status:
+  removedRelease: '1.21'
```
deads2k, line 66: spec

enhancements/kube-apiserver/stability-api-removal-notifications.md
```diff
@@ -0,0 +1,293 @@
...
+| ----------------- | --------------- | --------------------- |
+| 4.6 / 4.6 EUS     | 1.20            | 1.23                  |
+| 4.7               | 1.21            |                       |
+| 4.8               | 1.22            |                       |
+| 4.9               | 1.23            |                       |
+| 4.10 / 4.10 EUS   | 1.24            | TBD                   |
+
+#### DeprecatedAPIRequest Resource
+
+Patch the apiserver such that requests to a deprecated API are logged to a `DeprecatedAPIRequest` resource corresponding
```
soltysh, line 155: I know we're talking about kube-apiserver, but I'm not everyone :wink:

enhancements/kube-apiserver/stability-api-removal-notifications.md
sttts: What about resources that don't exist in GA version in EUS1, but do in EUS2, with the beta being removed somewhere in the middle? This is a real danger and I would not call this "mess", but rather something we are in full control to fix during the EUS upgrade path.

I would expect this either be a goal or non-goal to hide these. Making it a non-goal means that we have lots of ugly and the alert/info is always firing.

enhancements/kube-apiserver/stability-api-removal-notifications.md
wking: > If you have a signal for determining a customer planning on using 4.6 as an EUS, then we could reconsider...

We have that signal today by looking for folks in the `eus-4.6` channel.  Although folks can always change channels. 
 Once `stable-4.6` goes EOL, anyone on later 4.6.z will have to be `eus-4.6` folks who will eventually either tear down their clusters or try to update out to at least 4.8, and giving those folks a bit of lead time on things which might block their transition into 4.8 seems a lot nicer than surprising them once they land in 4.7 (which they probably planned to pass through extremely quickly on their way from 4.6 to 4.8).

enhancements/kube-apiserver/stability-api-removal-notifications.md
wking: +1 to hiding alerts about OCP-core consumers, however we need to do that.  We'd only want to do that in production though.  Not bothering an admin about something they can't fix outside of updating is great, but forgetting to update our own components and having an API removed that we were still consuming would be really bad.  But before we get too deep into this... how are we planning on distinguishing clients without exploding Prom cardinality?  Would we just filter out particular requests in the metrics-generator and just never submit deprecated-consumption-by-OCP-core-clients to Prom at all?

sttts: This is still open

deads2k:
replaced by https://github.com/openshift/enhancements/pull/764

PR #640, opened 2021-02-09
Ingress: Add GCP Global Access LB Proposal

sgreene570:
This commit adds the enhancement proposal for the GCP
Global Access option for Ingress Controller's exposed on GCP via
an internal Load Balancer.

Supports https://issues.redhat.com/browse/NE-447

### DIFF
--- a/enhancements/ingress/gcp-lb-global-access.md
+++ b/enhancements/ingress/gcp-lb-global-access.md
@@ -1,0 +1,226 @@
+---
+title: gcp-internal-load-balancer-global-access
+authors:
+  - "@sgreene570"
+reviewers:
+  - "@danehans"
+  - "@frobware"
+  - "@knobunc"
+  - "@miciah"
+  - "@candita"
+  - "@rfredette"
+  - "@miheer"
+approvers:
+  - "@danehans"
+  - "@frobware"
+  - "@knobunc"
+  - "@miciah"
+creation-date: 2021-02-09
+last-updated: 2021-02-12
+status: implementable
+---
+
+# GCP Internal Ingress Load Balancer Global Access Option
+
+## Release Signoff Checklist
+
+- [X] Enhancement is `implementable`
+- [X] Design details are appropriately documented from clear requirements
+- [X] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement extends the IngressController API to allow the user to enable
+the "Global Access" option for Ingress Controllers exposed via an Internal Load Balancer
+on GCP.
+
+## Motivation
+
+On GCP, Ingress Controllers that specify a `LoadBalancerScope` of `InternalLoadBalancer` are only externally accessible
+when a client is in the same VPC network and Google Cloud Region as the Load Balancer. GCP exposes the
+[Global Access](https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balancing#global_access) option specifically for
+Internal Load Balancers via an annotation on the given LoadBalancer service. When enabled, the Global Access option allows clients from any
+region within the cluster's VPC to connect to the Load Balancer.
+
+
+### Goals
+
+1. Expose the GCP Internal Load Balancer Global Access Option via a new field in the existing IngressController API `ProviderLoadBalancerParameters` type.
+
+
+### Non-Goals
+
+1. Expose a similar API field for other cloud providers (AWS, Azure, etc.) as GCP is the only provider that provides this configuration option.
+2. Change the current default behavior (no Global Access) of Ingress Controllers exposed via Internal Load Balancers on GCP.
+
+## Proposal
+
+The IngressController API is extended by adding a `GCP` field of type `GCPLoadBalancerParameters` to the `ProviderLoadBalancerParameters` type.
+
+```go
+// ProviderLoadBalancerParameters holds desired load balancer information
+// specific to the underlying infrastructure provider.
+// +union
+type ProviderLoadBalancerParameters struct {
+        // <snip>
+
+        // aws provides configuration settings that are specific to AWS
+        // load balancers.
+        //
+        // If empty, defaults will be applied. See specific aws fields for
+        // details about their defaults.
+        //
+        // +optional
+        AWS *AWSLoadBalancerParameters `json:"aws,omitempty"`
+
+        // gcp provides configuration settings that are specific to GCP
+        // load balancers.
+        //
+        // If empty, defaults will be applied. See specific gcp fields for
+        // details about their defaults.
+        //
+        // +optional
+        GCP *GCPLoadBalancerParameters `json:"gcp,omitempty"`
+}
+```
+
+Then, the new `GCPLoadBalancerParameters` type is created with a single field, `ClientAccess`.
+
+
+```go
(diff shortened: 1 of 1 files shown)


enhancements/ingress/gcp-lb-global-access.md
```diff
@@ -0,0 +1,223 @@
...
+## Summary
+
+This enhancement extends the IngressController API to allow the user to enable
+the "Global Access" option for Ingress Controllers exposed via an Internal Load Balancer
+on GCP.
+
+## Motivation
+
+On GCP, Ingress Controllers that specify a `LoadBalancerScope` of `InternalLoadBalancer` are only externally accessible
+when a client is in the same VPC network and Google Cloud Region as the provider's Load Balancer. GCP exposes the
```
Miciah, line 43: What does "the provider" refer to in this context?  Does it matter, or could we remove "provider's"?  
```suggestion
when a client is in the same VPC network and Google Cloud Region as the Load Balancer. GCP exposes the
```

enhancements/ingress/gcp-lb-global-access.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Goals
+
+1. Expose the GCP Internal Load Balancer Global Access Option via a new field in the existing IngressController API `ProviderLoadBalancerParameters` type.
+
+
+### Non-Goals
+
+1. Expose a similar API field for other cloud providers (AWS, Azure, etc.) as GCP is only provider that provides this configuration option.
+2. Change the current default behavior (no Global Access) of Ingress Controller's expose via an Internal Load Balancers on GCP.
```
Miciah, line 57:
```suggestion
2. Change the current default behavior (no Global Access) of Ingress Controller's expose via an Internal Load Balancers on GCP.
```
```suggestion
2. Change the current default behavior (no Global Access) of Ingress Controllers exposed via Internal Load Balancers on GCP.
```

enhancements/ingress/gcp-lb-global-access.md
```diff
@@ -0,0 +1,223 @@
...
+
+
+### Goals
+
+1. Expose the GCP Internal Load Balancer Global Access Option via a new field in the existing IngressController API `ProviderLoadBalancerParameters` type.
+
+
+### Non-Goals
+
+1. Expose a similar API field for other cloud providers (AWS, Azure, etc.) as GCP is only provider that provides this configuration option.
```
Miciah, line 56:
```suggestion
1. Expose a similar API field for other cloud providers (AWS, Azure, etc.) as GCP is the only provider that provides this configuration option.
```

enhancements/ingress/gcp-lb-global-access.md
```diff
@@ -0,0 +1,223 @@
...
+        ClientAccess GCPClientAccess `json:"clientAccess"`
+}
+```
+
+The following `GCPClientAccess` string constants are defined to support the new `ClientAccess` field.
+
+
+```go
+
+// GCPClientAccess describes how client acecss is restricted for internal
```
Miciah, line 123:
```suggestion
// GCPClientAccess describes how client access is restricted for internal
```

enhancements/ingress/gcp-lb-global-access.md
```diff
@@ -0,0 +1,223 @@
...
+
+How often do cluster administrators want an internal load balancer on GCP that can communicate with clients in a different region?
+
+### Test Plan
+
+The Ingress Operator has existing unit tests to verify that the operator renders the correct service definition given an Ingress Controller's `Spec.LoadBalancer` field.
+These unit test will be extended to verify that Ingress Operator applies the Global Access LoadBalancer service annotations for Ingress Controllers with
+`Spec.LoadBalancer.ProviderParameters.GCP.ClientAccess` and `Spec.LoadBalancer.Scope` properly set.
+
+The Ingress Operator also has existing e2e tests. These e2e tests will be extended to verify that Ingress Controller's created with the Global Access option enabled
```
Miciah, line 179:
```suggestion
The Ingress Operator also has existing e2e tests. These e2e tests will be extended to verify that Ingress Controllers created with the Global Access option enabled
```

enhancements/ingress/gcp-lb-global-access.md
```diff
@@ -0,0 +1,223 @@
...
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+On upgrade, internal LoadBalancer services on GCP will not be updated with the Global Access annotation unless
+an Ingress Controller's spec is updated to require so.
+
+On downgrade, the Ingress Operator will ignore the new `GCPLoadBalancerParameters` field for an Ingress Controller and not set any
+Global Access LoadBalancer service annotations.
```
Miciah, line 206: We don't have logic in 4.7 that would revert the annotation on the service, so the annotation would not be reverted on downgrade, and therefore any setting made on a 4.8 cluster would remain intact after downgrade to 4.7, right?

enhancements/ingress/gcp-lb-global-access.md
```diff
@@ -0,0 +1,223 @@
...
+N/A
+
+## Implementation History
+
+None.
+
+## Alternatives
+
+A cluster administrator is already free to use a load balancer not managed by the Cluster Ingress Operator to configure these kinds of settings on an Ingress Controller's
+LoadBalancer service.
```
Miciah, line 219: This could be clearer:
```suggestion
A cluster administrator can already create a custom service that is not managed by the Cluster Ingress Operator.  The administrator can then configure these kinds of settings on the
custom, administrator-managed LoadBalancer service.
```

enhancements/ingress/gcp-lb-global-access.md
sgreene570: We can remove "provider's". Thanks!

enhancements/ingress/gcp-lb-global-access.md
sgreene570: fixed, thanks!

enhancements/ingress/gcp-lb-global-access.md
sgreene570: fixed.

enhancements/ingress/gcp-lb-global-access.md
sgreene570: whoops! fixed.

enhancements/ingress/gcp-lb-global-access.md
sgreene570: fixed

enhancements/ingress/gcp-lb-global-access.md
sgreene570: You are correct. I included this information in the enhancement (and also mentioned that this behavior is not detrimental and easily corrected need be).

enhancements/ingress/gcp-lb-global-access.md
sgreene570: Thanks for the suggestion! fixed.

enhancements/ingress/gcp-lb-global-access.md
```diff
@@ -0,0 +1,226 @@
...
+N/A
+
+## Implementation History
+
+None.
+
+## Alternatives
+
+A cluster administrator can already create a custom service that is not managed by the Cluster Ingress Operator. The administrator can then configure
+these kinds of settings on the custom, administration-managed LoadBalancer service.
```
Miciah, line 222:
```suggestion
these kinds of settings on the custom, administrator-managed LoadBalancer service.
```

Miciah:
/approve
/lgtm
/hold
in case @frobware, @danehans, @knobunc, or others want to review.

PR #641, opened 2021-02-10
EP: Support for Cluster High-availability mode API

varshaprasad96:
Enhancement on providing helper functions that
expose the high availability expectations of
OpenShift cluster.

### DIFF
--- a/enhancements/support-cluster-HA-API.md
+++ b/enhancements/support-cluster-HA-API.md
@@ -1,0 +1,134 @@
+---
+title: support-cluster-HA-API
+authors:
+  - "@varshaprasad96"
+reviewers:
+  - "@jmrodri"
+  - "@ecordell"
+  - "@estroz"
+approvers:
+  - "@jmrodri"
+  - "@ecordell"
+  - "@estroz"
+creation-date: 2021-02-01
+status: implementable
+---
+
+# Support for Cluster High-availability Mode API
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+
+## Summary
+
+The goal of this enhancement is to describe the support for [cluster high-availability mode API][enhancement_cluster_operators] introduced by OpenShift. The enhancement proposes to introduce necessary helper functions which will enable users to identify if the OpenShift cluster can support high availability deployment mode of their operator or not.
+
+## Motivation
+
+OpenShift intends to provide an API to expose the high-availability expectations to operators in a cluster. The API will accomplish this by embedding additional fields in the [informational resource][informational_resource]. Providing the necessary helper functions which utilize the OpenShift API to expose the intended cluster topology, will enable users to identify the mode of operator deployment which the cluster can support. Operator authors can leverage this functionality and configure their operators to react accordingly.
+
+### Goals:
+
+- List the helper functions that read the OpenShift informational resource and return appropriate values to expose the high availability mode that the cluster can support.
+- Discuss on the appropriate location of the helper functions.
+
+### Non-Goals:
+
+- Discuss how OpenShift determines if the cluster can support HA/non-HA mode.
+- Discuss the behavior of operators in HA/non-HA mode.
+
+#### Story 1
+
+As an Operator author, I would like to know if the cluster can support HA or non-HA modes, so that I can configure my operator's behavior accordingly.
+
+## Design Details
+
+As an operator author it is helpful to know the cluster topology so that one can modify the behavior of operands. The library where this functionality resides, will provide users with a `GetClusterConfig` function that returns a `ClusterConfig` struct.
+
+The `ClusterConfig` struct can be defined to be:
+
+```go
+import (
+    ...
+    apiv1 "github.com/openshift/api/config/v1"
+    ...
+)
+
+// ClusterConfig contains the configuration information of the
+// cluster.
+type ClusterConfig struct {
+    // ControlPlaneTopology contains the expectations for operands that 
+    // run on control plane nodes.
+    ControlPlaneTopology apiv1.TopologyMode
+
+    // InfrastructureTopology contains the expectations for infrastructre
+    // services.
+    InfrastructureTopology apiv1.TopologyMode
+}
+```
+
+The `ClusterConfig` contains the high availability expectations for control plane and infrastructure nodes. The `apiv1.TopologyMode` is defined by OpenShift API and refers to the [topology mode][api_topology_mode_pr] in which the cluster is running. It can have the following two values:
+
+1. `ControlPlaneTopology` - refers to the expectations for operands that normally run on [control plane][control_plane] (or master) nodes.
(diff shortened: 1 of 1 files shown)


enhancements/support-cluster-HA-API.md
```diff
@@ -0,0 +1,134 @@
...
+
+#### Story 1
+
+As an Operator author, I would like to know if the cluster can support HA or non-HA modes, so that I can configure my operator's behavior accordingly.
+
+## Design Details
+
+As an operator author it is helpful to know the cluster topology so that one can modify the behavior of operands. The library where this functionality resides, will provide users with a `GetClusterConfig` function that returns a `ClusterConfig` struct.
+
+The `ClusterConfig` struct can be defined to be:
```
dhellmann, line 52: What benefit do we get from defining a new type instead of using the existing Infrastructure type?

romfreiman:
@dhellmann  @ravidbro

enhancements/support-cluster-HA-API.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+**NOTE:**
+Operators cannot modify the informational resources present in OpenShift clusters. The function can only be used to retrieve information about the high availability expectations of the cluster and not to modify it.
+
+### Scenarios:
+
+1. Operator author tries to fetch the `ClusterConfig` for a non-openshift cluster:
+
+In this case, an `ErrAPINotAvailable` error is returned. The error can be wrapped, such that users can utilize the `errors.Is` method to identify this specific case.
```
dhellmann, line 107: When the Infrastructure resource is not present, why return an error instead of default or nil values?

dhellmann:
@russellb maybe you can apply the priority label, since the bot doesn't want me to?

enhancements/support-cluster-HA-API.md
jmrodri: @dhellmann how will the API defined in this EP https://github.com/openshift/enhancements/blob/master/enhancements/single-node/cluster-high-availability-mode-api.md be accessible? I'm wondering if there's a need for a "wrapper" in operator-lib or library-go at all.

ravidbro: This is already implemented here:
https://github.com/openshift/api/blob/master/config/v1/0000_10_config-operator_01_infrastructure.crd.yaml#L182

and example of usage:
https://github.com/openshift/cluster-authentication-operator/pull/391/files#diff-b6f4bf160fd7e801eadfbfac60b84bd00fcad21a0979c2d730549f10db015645R108

dhellmann: As Ravid says, the API is implemented as 2 new fields on the Infrastructure configuration resource. It sounds like the proposal is to provide a convenience function to read that resource and extract the 2 fields to return in a new struct. Some level of help for operators that run on non-OpenShift clusters might be useful, since it means we could provide some defaulting behavior for those platforms. That's not what this proposal says, though (see my other suggestions elsewhere on the review).

enhancements/support-cluster-HA-API.md
varshaprasad96: Isn't it better to explicitly state out that this particular feature is for openshift clusters only? Users can chose to handle the error gracefully and proceed if they use this in a non-openshift cluster, since this entire feature is dependent on a specific CRD which is present only in openshift. Would it be fine to say that we have a HA/non-HA detector for the cluster which does not expose such a detail itself.

dhellmann: > Isn't it better to explicitly state out that this particular feature is for openshift clusters only? 
> Users can chose to handle the error gracefully and proceed if they use this in a non-openshift cluster, since this entire feature is dependent on a specific CRD which is present only in openshift. 

I would expect an helper function for an OpenShift cluster to return the Infrastructure resource directly. But that helper wouldn't be all that useful, since it's just the Get() method of the client.

> Would it be fine to say that we have a HA/non-HA detector for the cluster which does not expose such a detail itself.

I'm not sure what you mean here. Can you give an example of what that might look like?

varshaprasad96: > I would expect an helper function for an OpenShift cluster to return the Infrastructure resource directly. But that helper wouldn't be all that useful, since it's just the Get() method of the client.

+1! I agree

> I'm not sure what you mean here. Can you give an example of what that might look like?

What I meant was - if we return nil or any other default value for non-openshift clusters, it would mean that we are trying to tell the user if the cluster supports HA or non HA mode when the cluster doesn't even expose such a metric. Would giving a specific error for this case be better?

dhellmann: > > I'm not sure what you mean here. Can you give an example of what that might look like?
> 
> What I meant was - if we return nil or any other default value for non-openshift clusters, it would mean that we are trying to tell the user if the cluster supports HA or non HA mode when the cluster doesn't even expose such a metric. Wouldn't giving a specific error for this case be better?

I think I would need to see how we would expect a caller of this function to use it to answer. It sounds like the caller is going to have to do a bunch of work to check the return value and then decide how to handle error cases by default. Could we make it more helpful than that?

For example, could we trap all errors and return a default value provided by the caller if we can't retrieve a value from the Infrastructure CR, for any reason? That would at least let the caller avoid any extra logic when calling it. They could do something like

```go
topology := GetClusterTopology(HighlyAvailable)
```

without having to check for errors because we would promise to always return something.

enhancements/support-cluster-HA-API.md
jmrodri: @dhellmann after investigating more I've come to the consensus that we probably don't need a convenience library at the moment. The existing API as pointed out by @ravidbro above is quite easy to access via normal k8s constructs and controller-runtime.

Simple controller-runtime example (just the act of getting it).
```golang
	// Simple query
	nn := types.NamespacedName{
		Name: "cluster",
	}
	infraConfig := &configv1.Infrastructure{}
	err = crClient.Get(context.Background(), nn, infraConfig)
	if err != nil {
		return err
	}
	fmt.Printf("using crclient: %v\n", infraConfig.Status.ControlPlaneTopology)
	fmt.Printf("using crclient: %v\n", infraConfig.Status.InfrastructureTopology)
```

And similarly with normal k8s constructs
```golang
	operatorConfigInformer := configinformer.NewSharedInformerFactoryWithOptions(configClient, 2*time.Second)
	infrastructureLister = operatorConfigInformer.Config().V1().Infrastructures().Lister()
	infraConfig, err := configClient.ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
	if err != nil {
		return err
	}

	// fmt.Printf("%v\n", infraConfig)
	fmt.Printf("%v\n", infraConfig.Status.ControlPlaneTopology)
	fmt.Printf("%v\n", infraConfig.Status.InfrastructureTopology)
```

Any helper library would simply be a layer above what is already a seemingly simple API call. So I'm advocating we simply close this EP.

dhellmann: @jmrodri I think your conclusion makes sense.

@varshaprasad96, what do you think?

varshaprasad96: Yes, I agree. Closing this EP

PR #642, opened 2021-02-11
Auto Node Sizing

harche:
This enhancement attempts to add a capability to dynamically select node sizing values for the kubelet. 

### DIFF
--- a/enhancements/kubelet/kubelet-node-sizing.md
+++ b/enhancements/kubelet/kubelet-node-sizing.md
@@ -1,0 +1,268 @@
+---
+title: auto-node-sizing
+authors:
+  - "@harche"
+reviewers:
+  - "@rphillips"
+approvers:
+  - "@rphillips"
+creation-date: 2021-02-11
+last-updated: 2021-02-11
+status: implementable
+see-also:
+  - https://bugzilla.redhat.com/show_bug.cgi?id=1857446
+replaces:
+superseded-by:
+---
+
+# Kubelet Auto Node Sizing
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+
+## Summary
+
+Nodes should have an automatic sizing calculation mechanism, which could give kubelet an ability to scale values for memory and cpu system reserved based on machine size.
+
+Today the sizing values are passed manually to kubelet using `--kube-reserved` and `--system-reserved` flags. Many cloud providers provide reference values for their customers to help them select optimal values based on the node sizes. e.g. [GKE](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#memory_cpu), [AKS](https://docs.microsoft.com/en-us/azure/aks/concepts-clusters-workloads#resource-reservations)
+
+This enhancement proposes a mechanism to automatically determine the optimal sizing values for any node size irrespective of the cloud provider.
+
+## Motivation
+
+Kubelet’s `system reserved` and `kube reserved` play a crucial role in the OOMKilling the resource intensive pods. Without an adequate enough `system reserved` and `kube reserved` we risk freezing the node making it completely unavailable for other pods.
+
+We have observed that scaling the value of `system reserved` and `kube reserved` with respect to the installed capacity of the node helps to deduce optimal values. Larger nodes have capacity for more pods and will require larger system reserved values.
+
+Currently, the only way to customize the `system reserved` and `kube reserved` limits is to pre-calculate the values manually prior to Kubelet start.
+
+### Goals
+
+* Enable Kubelet systemd service to determine the value of the `system reserved` automatically during start up.
+
+### Non-Goals
+
+* For now the systemd service will only be used for calculating the values of `system reserved`. Similar approach can be taken to dynamically fetch the values of other parameters of the kubelet (e.g. `evictionHard`) but they are out of scope of this enhancement.
+* Strictly from the OpenShift's point of view, we only need to take care of `system reserved`, and not `kube reserve`. Hence this proposal will not deal with generating optimal values for `kube reserve`
+
+### User Stories
+
+* User wants to enable auto node sizing on nodes of the cluster to start the kubelet with optimal system reserved values.
+
+## Proposal
+
+* New script that will be placed on the node that can calculate the system reserved values based on the node capacity
+* New auto node sizing service to execute that script which will result in storing the system reserved values in a file.
+* Modify kubelet service to read that file and use the generated values to start kubelet daemon.
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+* Successfully calculate and set the optimal system reserved values.
+* End user documentation
+
+#### Tech Preview -> GA
+* More testing (upgrade, downgrade, scale)
+* Optinally make it available during installation
+
+#### Removing a deprecated feature
+
+N/A
+
+## Design Details
+
+### Auto Node Sizing Enabler
+
+During the cluster installation a file will be placed at the location `/etc/node-sizing-enabled.env` with following content,
+
+```bash
+NODE_SIZING_ENABLED=false
+SYSTEM_RESERVED_MEMORY=1Gi
+SYSTEM_RESERVED_CPU=500m
+```
+Initially we would like the `Auto Node Sizing` to be an optional feature, so the value of the variable `NODE_SIZING_ENABLED` will be set to `false` during the installation along with the existing default values for system reserved memory and cpu. To enable this feature, the value of the variable `NODE_SIZING_ENABLED` can be set to `true` by using following `KubeletConfig`.
+
+```yaml
+kind: KubeletConfig
+metadata:
+  name: dynamic-node
+spec:
+  autoSizingReserved: true
+  machineConfigPoolSelector:
+    matchLabels:
+      pools.operator.machineconfiguration.openshift.io/worker: ""
+```
+This will enable `Auto Node Sizing` on all the worker nodes. A similar approach can be taken to enable it on the `master` nodes or on a custom machine config pool.
+
+### Auto Node Sizing Script
+
+This script can be found on the node at the location, `/usr/local/sbin/dynamic-system-reserved-calc.sh`
+
+When the `Auto Node Sizing` is enabled, script will probe the host to get the installed resource capacity (such as, installed amount of RAM) and use well tested guidance on the optimal values for the corresponding system reserved.
+
+Some of the examples of the guidance values for system reserved provided by [GKE](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#memory_cpu) and [AKS](https://docs.microsoft.com/en-us/azure/aks/concepts-clusters-workloads#resource-reservations)
+
+And when the `Auto Node Sizing` is disabled, the script will output the existing default values for the system reserved.
+
+The script will output the values in the following format at the location `/etc/node-sizing.env`,
+
+```bash
+$ cat /etc/node-sizing.env
+SYSTEM_RESERVED_MEMORY=3.5Gi
+SYSTEM_RESERVED_CPU=0.09
+```
+### Kubelet Auto Node Sizing Service
+
+A new service `kubelet-auto-node-size.service` that will run `before` the existing kubelet service to calculate the optimal values of system reserved.
+
+```toml
+[Unit]
+Description=Dynamically sets the system reserved for the kubelet
+Wants=network-online.target
+After=network-online.target ignition-firstboot-complete.service
+Before=kubelet.service crio.service
+[Service]
+# Need oneshot to delay kubelet
+Type=oneshot
+RemainAfterExit=yes
+EnvironmentFile=/etc/node-sizing-enabled.env
+ExecStart=/bin/bash /usr/local/sbin/dynamic-system-reserved-calc.sh ${NODE_SIZING_ENABLED}
+[Install]
+RequiredBy=kubelet.service
+```
+This service will write recommended values of system reserved to the location `/etc/node-sizing.env`. It depends on another systemd environment file `/etc/node-sizing-enabled.env` mentioned above to determine if the user has enabled the `Auto Node Sizing` feature. In case user has not opted to enable it, this service will output the default values of the system reserved used today in `/etc/node-sizing.env`.
+
+### Changes to Existing Kubelet Service
+
+```toml
+[Unit]
+Description=Kubernetes Kubelet
+Wants=rpc-statd.service network-online.target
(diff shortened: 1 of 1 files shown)


enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,211 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Open Questions [optional]
+
+## Summary
+
+Kubelet should have an automatic sizing calculation mechanism, which could give kubelet an ability to dynamically set sizing values for memory and cpu reservations.
```
rphillips, line 32: memory and cpu system reserved*

enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,211 @@
...
+
+This enhancement proposes a mechanism to automatically determine the optimal sizing values for any node size irrespective of the cloud provider.
+
+## Motivation
+
+Kubelet’s `system reserved` and `kube reserved` play a crucial role in the OOMKilling the resource intensive pods. Without an adequate enough `system reserved` and `kube reserved` we risk freezing the node making it completely unavailable for other pods. 
+
+We have observed that varying the value of `system reserved` and `kube reserved` with respect to the installed capacity of the node helps to deduce optimal values.
+
+Currently, the only way to customize the `system reserved` and `kube reserved` limits is to pre-calculate the values manullay prior to Kubelet start.
```
rphillips, line 44: nit: manually

enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,211 @@
...
+controlPlane:
+  architecture: amd64
+  hyperthreading: Enabled
+  name: master
+  platform: {}
+  replicas: 3
+metadata:
+  creationTimestamp: null
+  name: testcluster1
+  autoNodeSizing: true
```
rphillips, line 79: +1 nice

enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,211 @@
...
+Restart=always
+RestartSec=10
+
+[Install]
+WantedBy=multi-user.target
+```
+
+Node sizing values will be calculated by the script `/etc/kubernetes/node-sizing`. This script will set the environment variables `CPU_RESERVED`, `MEMORY_RESERVED` and `EPHEMERAL_RESERVED` which will be used for setting the kubelet parameter `system-reserved`
+
+The script `/etc/kubernetes/node-sizing` will be executed using `ExecStartPre`
```
rphillips, line 141: I did a quick audit of MCO... Let write the script to /usr/local/bin. `_base/files/configure-ovs-network.yaml` is one example that writes to /usr/local/bin that already exists.

rphillips (changes requested):

enhancements/kubelet/kubelet-node-sizing.md
harche: Thanks, updated.

enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,212 @@
...
+### Non-Goals
+
+* For now the systemd service will only be used for calculating the values of `system reserved`. Similar approach can be taken to dynamically fetch the values of other parameters of the kubelet (e.g. `evictionHard`) but they are out of scope of this enhancement. 
+* Strictly from the OpenShift's point of view, we only need to take care of `system reserved`, and not `kube reserve`. Hence this proposal will not deal with generating optimal values for `kube reserve` 
+
+## Proposal
+
+### Installer Changes
+
+If the user wishes to enable `auto node sizing` during the installation of the cluster itself, they will have to indicate that using a field `metadata.autoNodeSizing` in installation configuration. e.g.
```
sdodson, line 59: What percentage of clusters would you expect to enable this versus disable it? Does it require in depth analysis to decide? If we have 80% one way or the other I don't think we want to promote this into install-config.
CC @staebler

rphillips: We will eventually want this in the default install, because cluster memory reservations can change the workload capacity within the system. It'll be defaulted to off (for now), but some clouds - like Azure - may be enabled by default.

enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,212 @@
...
+  No
+
+## Drawbacks
+
+This solution utilizes kubelet command line flags. Kubelet command line flags have been deprecated in favour of config file, so there is risk for this solution if those flags are actually purged. Having said that, those flags are quite widely used today. So there has not been much traction on actually removing those flags even though they have been marked deprecated. 
+
+## Alternatives
+
+1. Enhance kubelet itself to be more smart about calculating node sizing values. We have an actively debated [KEP](https://github.com/kubernetes/enhancements/pull/2370) in sig-node around this idea. 
+2. Modify MCO the way it handles kubeletconfig. Instead of passing `--system-reserved` argument to the kubelet, maybe there is a possibility to make sure MCO is more tolerant of changes to the kubelet config file. This way we will modify the config file to add system reserve values instead of passing them as `--system-reserved`.
```
staebler, line 212: This is how I would have expected node sizing to be handled. I would have expected a KubeletConfig resource add by the user to either set explicit values or set it to use auto-sizing.

enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,212 @@
...
+      --v=${KUBELET_LOG_LEVEL}
+
+Restart=always
+RestartSec=10
+
+[Install]
+WantedBy=multi-user.target
+```
+
+Node sizing values will be calculated by the script `/usr/local/bin/node-sizing.sh`. This script will set the environment variables `CPU_RESERVED`, `MEMORY_RESERVED` and `EPHEMERAL_RESERVED` in `/etc/kubernetes/node-sizing-env` which will be used for setting the kubelet parameter `system-reserved`
```
staebler, line 140: Can you provide a bit more detail about how the actual values used will be decided in the script?

enhancements/kubelet/kubelet-node-sizing.md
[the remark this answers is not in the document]
staebler: My first thought is that this should be in the machine pool.

enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,212 @@
...
+
+The script `/usr/local/bin/node-sizing.sh` will be executed using `ExecStartPre`
+
+#### Node sizing script
+
+Irrespective of `auto node sizing` is enabled or not, MCO will always place the script file `/usr/local/bin/node-sizing.sh` on the node.
+
+However the content of the script file will vary depending upon whether the `auto node sizing` is enabled or disabled. When the `auto node sizing` is disabled, `/usr/local/bin/node-sizing.sh` will set variables `CPU_RESERVED`, `MEMORY_RESERVED` and `EPHEMERAL_RESERVED` to their existing default values, but when the `auto node sizing` is enabled the script will dynamically select the optimal value based on the resources available on the node.
```
staebler, line 148: What is the plan for how the MCO will know whether auto node sizing is enabled? Is this something that we will allow the user to change as a day-2 operation?

enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,212 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Open Questions [optional]
+
+## Summary
+
+Kubelet should have an automatic sizing calculation mechanism, which could give kubelet an ability to dynamically set sizing values for memory and cpu system reserved.
```
ehashman, line 32:
```suggestion
Nodes should have an automatic sizing calculation mechanism, which could give kubelet an ability to scale values for memory and cpu system reserved based on machine size.
```

I don't think the mechanism necessarily needs to live inside the kubelet; some other component of a cluster could be responsible for doing the calculation.

enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,212 @@
...
+
+Today the sizing values are passed manually to kubelet using `--kube-reserved` and `--system-reserved` flags. Many cloud providers provide reference values for their customers to help them select optimal values based on the node sizes. e.g. [GKE](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#memory_cpu), [AKS](https://docs.microsoft.com/en-us/azure/aks/concepts-clusters-workloads#resource-reservations)
+
+This enhancement proposes a mechanism to automatically determine the optimal sizing values for any node size irrespective of the cloud provider.
+
+## Motivation
+
+Kubelet’s `system reserved` and `kube reserved` play a crucial role in the OOMKilling the resource intensive pods. Without an adequate enough `system reserved` and `kube reserved` we risk freezing the node making it completely unavailable for other pods. 
+
+We have observed that varying the value of `system reserved` and `kube reserved` with respect to the installed capacity of the node helps to deduce optimal values.
```
ehashman, line 42:
```suggestion
We have observed that scaling the value of `system reserved` and `kube reserved` with respect to the installed capacity of the node helps to deduce optimal values. Larger nodes have capacity for more pods and will require larger system reserved values.
```

enhancements/kubelet/kubelet-node-sizing.md
ehashman, line 59: Because this name is end user facing, it's going to be extremely confusing I think; I would expect this to mean "the node automatically will scale in size with respect to my workloads", and not "the system reserved will scale to size of the node". I would suggest another name, like `autoSystemReserved`.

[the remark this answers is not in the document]
harche: We want to have a cluster wide option that's applicable to all nodes. Either all nodes (workers and masters) have this enabled or not. We didn't want user to specify this option for `compute` or `controlPlane` individually.

enhancements/kubelet/kubelet-node-sizing.md
harche: For the day-1 scenario, until the kubelet is deployed and node has joined the cluster the MCO controller does not know the node size.

enhancements/kubelet/kubelet-node-sizing.md
harche: Thanks, I will update the name to `autoSystemReserved`

enhancements/kubelet/kubelet-node-sizing.md
harche: Thanks, I have added a subsection `Node sizing script` that talks more about how the script will use the guidance values to determine optimal system reserved.

enhancements/kubelet/kubelet-node-sizing.md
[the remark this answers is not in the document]
staebler: What is the reasoning for that? Is it required that it be applied cluster-wide for some technical reason? Or is it just for user convenience?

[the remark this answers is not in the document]
harche: This enhancement will set the optimal values of the `system-reserved` for the kubelet on the node. Having the optimal value of `system-reserved` makes a huge difference when node is under severe resource crunch, such as pod running with extremely high memory requirements. 

Without the right value of `system-reserved` the risk of node getting into lock up with `NotReady` status increases. You can set the `system-reserved` value only on specific nodes, since it's just a parameter to a kubelet. But this would not be very ideal from the customer support point of view. Customers will see a non-uniform behaviour in performance within their cluster nodes. Nodes with optimal `system-reserve` will perform substantially better than those without optimal values. 

We want to avoid any ambiguity that may arise if customers see their workload run fine on some nodes but not others. This is the reason, we want to make this a cluster-wide option. That way any in future should we receive any customer support ticket related to node lock ups, it's easy for us to understand if the `system-reserve` was set across the cluster properly or not.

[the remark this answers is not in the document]
harche: /cc @rphillips

enhancements/kubelet/kubelet-node-sizing.md
```diff
@@ -0,0 +1,215 @@
...
+- Will any other components on the node change? For example, changes to CSI, CRI
+  or CNI may require updating that component before the kubelet.
+
+  No
+
+## Drawbacks
+
+This solution utilizes kubelet command line flags. Kubelet command line flags have been deprecated in favour of config file, so there is risk for this solution if those flags are actually purged. Having said that, those flags are quite widely used today. So there has not been much traction on actually removing those flags even though they have been marked deprecated.
+
+## Alternatives
```
romfreiman, line 212: Isnt IPI has the knowledge in advance? Installer it the one who spawns the VMs in the cloud, so it knows the size. Same  for BM with the assisted installer where it knows the values in advance. Not sure about Metal3/ironic though.

rphillips: We need a solution that works across all the clouds and install flavors. Metal would be problematic here to figure out up front, and sometimes machines have different hardware layouts. We plan on having this script run on each node which should alleviate cloud to bare metal discrepancies.

harche:
Not sure why the job `ci/prow/markdownlint` is failing, when I run it locally it doesn't show any error. 

```bash
$ hack/markdownlint.sh 
+ '[' -z '' ']'
+ case "$-" in
+ __lmod_vx=x
+ '[' -n x ']'
+ set +x
Shell debugging temporarily silenced: export LMOD_SH_DBG_ON=1 for this output (/usr/share/lmod/lmod/init/bash)
Shell debugging restarted
+ unset __lmod_vx
+ markdownlint-cli2 '**/*.md'
markdownlint-cli2 v0.0.15 (markdownlint v0.23.1)
Finding: **/*.md
Linting: 218 file(s)
Summary: 0 error(s)
```

harche:
Oh got it,

```
enhancements/kubelet/kubelet-node-sizing.md missing "### User Stories"
enhancements/kubelet/kubelet-node-sizing.md missing "### Risks and Mitigations"
enhancements/kubelet/kubelet-node-sizing.md missing "## Design Details"
enhancements/kubelet/kubelet-node-sizing.md missing "### Graduation Criteria"
enhancements/kubelet/kubelet-node-sizing.md missing "#### Dev Preview -> Tech Preview"
enhancements/kubelet/kubelet-node-sizing.md missing "#### Tech Preview -> GA"
enhancements/kubelet/kubelet-node-sizing.md missing "#### Removing a deprecated feature"
enhancements/kubelet/kubelet-node-sizing.md missing "### Upgrade / Downgrade Strategy"
enhancements/kubelet/kubelet-node-sizing.md missing "## Implementation History"
```

mrunalp (approved):

PR #643, opened 2021-02-11
Add Reduced Reboots enhancement

sdodson:
This enhancement builds upon the MCO node constraints(https://github.com/openshift/enhancements/blob/master/enhancements/update/eus-upgrades-mvp.md#mco---enforce-openshifts-defined-host-component-version-skew-policies) to enable host component version skew of N-2 which allows reboots across multiple minor version upgrades to be skipped.

### DIFF
--- a/enhancements/update/reduced_reboots.md
+++ b/enhancements/update/reduced_reboots.md
@@ -78,11 +78,22 @@ define and implement upstream testing changes which give us an appropriate level
 of confidence that N-2 version skew issues would be identified in the community
 whenever possible.
 
-#### Node - Implement Downstream Kubelet Version Skew Testing
+#### OTA - Implement Downstream Paused Worker Pool Upgrade Tests
 
-In addition to upstream version skew testing is in place we must also implement
-downstream version skew testing which includes any additional tests required for
-OpenShift specific implementation details.
+In parallel with efforts to revamp upstream version skew testing we must also
+implement downstream version skew testing which includes any additional tests
+required for OpenShift specific implementation details.
+
+We will achieve this by delivering upgrade jobs which pause the Worker MachineConfigPool
+then upgrade from 4.x to 4.x+1 to 4.x+2. We will run conformance tests from 4.x
+after upgrading to 4.x+1 in order to ensure that we continue to provide a baseline
+feature set, then again after upgrading to 4.x+2, finally after unpausing the
+Worker MCP we will run 4.x+2 tests.
+
+Given the complexity of these test jobs we should expect that they may take
+longer than the current four hour limit for test jobs. Rather than compromising
+on test completeness we will seek to extend test duration limits or find other
+ways to meet these testing demands.
 
 #### Teams with Host Components - Allow N-2 Host Component Version Skew
 
@@ -139,12 +150,18 @@ be a good place to talk about core concepts and how they relate.
 
 This imposes significant risk due to a number of factors:
 - We're currently not confident in upstream's testing matrix and our specific
-feature sets
-- We're never before expected teams to offer broader than N-1 compatibility.
-Teams have always assumed at most N-1 and even then, especially early after a
-GA release, it's not uncommon to find problems in N-1 compatibility.
+  feature sets
+- We've never before expected teams to offer broader than N-1 compatibility.
+  Teams have always assumed at most N-1 and even then, especially early after a
+  GA release, it's not uncommon to find problems in N-1 compatibility.
 - While N-1 is tested in the context of upgrades it's not been tested in long
-term use.
+  term use.
+- If external integrations depend on host components of all nodes having been
+  updated then we'll run into problems. For instance, if there's an upgrade
+  scenario where RHV cloud provider integration needs to be upgraded between 4.6
+  and 4.10 in order to ensure compatibility and the components which interface
+  with RHV are components of RHCOS then we may not upgrade those components at
+  the same minor version expected previously.
 
 We may mitigate some of this by further delaying EUS-to-EUS upgades after normal
 minor version upgrades have been promoted to stable and allocating significantly
--- a/enhancements/update/reduced_reboots.md
+++ b/enhancements/update/reduced_reboots.md
@@ -12,6 +12,8 @@ reviewers:
   - @mrunalp
   - @zvonkok
   - @pweil-
+  - @wking
+  - @vrutkovs
 approvers:
   - @derekwaynecarr
   - @crawford
--- a/enhancements/update/reduced_reboots.md
+++ b/enhancements/update/reduced_reboots.md
@@ -1,0 +1,309 @@
+---
+title: reduced-reboot-upgrades
+authors:
+  - "@sdodson"
+reviewers:
+  - @darkmuggle
+  - @rphillips
+  - @derekwaynecarr
+  - @crawford
+  - @dcbw
+  - @miabbott
+  - @mrunalp
+  - @zvonkok
+  - @pweil-
+approvers:
+  - @derekwaynecarr
+  - @crawford
+creation-date: 2020-01-21
+last-updated: 2020-01-21
+status: provisional
+see-also:
+  - https://github.com/openshift/enhancements/pull/585
+  - "/enhancements/eus-mvp.md"
+
+---
+
+# Reduced Reboot Upgrades
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement is intended to reduce host reboots when upgrading across two or
+more OpenShift minor versions by enabling an N-2 version skew policy between all
+host components and cluster scoped resources.
+
+## Motivation
+
+While OpenShift is designed to minimize workload disruption and risk associated
+with rolling reboots there exist a class of customers and workloads where
+reboots remain a disruptive and time consuming activity. Additionally, with the
+introduction of Extended Update Support (EUS) a new upgrade pattern will emerge
+where clusters run 4.6 for a year or more then rapidly upgrade across multiple
+minor versions in a short period of time. Those customers wish to complete their
+upgrades in a condensed time frame and with as few reboots as possible, they do
+not intend to run each minor version for an extended period of time.
+
+### Goals
+
+- Define testing requirements for N-2 host to cluster resource version skew
+- Define version skew policies for host and cluster scoped resources
+- Reduce reboots in accordance with our new tested policies
+
+### Non-Goals
+
+- Exceeding upstream's documented Kubelet version skew policies
+
+## Proposal
+
+### User Stories
+
+#### Node - Improve Upstream Kubelet Version Skew Testing
+
+Kubernetes defines a version skew policy(https://kubernetes.io/docs/setup/release/version-skew-policy/#kubelet)
+which allows for kubelet N-2 to be compatible with kube-apiserver version N. At
+this point in time OpenShift is not comfortable with the level of testing upstream
+and the intersection of the specific features of OpenShift. We should work to
+define and implement upstream testing changes which give us an appropriate level
+of confidence that N-2 version skew issues would be identified in the community
+whenever possible.
+
+#### Node - Implement Downstream Kubelet Version Skew Testing
+
+In addition to upstream version skew testing is in place we must also implement
+downstream version skew testing which includes any additional tests required for
+OpenShift specific implementation details.
+
+#### Teams with Host Components - Allow N-2 Host Component Version Skew
+
+All teams which own components that directly interface with or ship host based
+components will need to ensure that they're broadening their compatibility to
+allow for N-2 version skew between host and cluster scoped resources.
+
+This would include for example the SDN DaemonSets in 4.10 remaining compatible
+with OVS and any other host components in 4.10, 4.9, and 4.8. On a case by case
+basis teams should decide whether it makes more sense to maintain a broader
+compatibility matrix or that N-1 bits and MachineConfig are backported to N-2 and
+upgrade graph is amended with these new minimum version requirements.
+
+For instance, if 4.9.12 is the minimum version for 4.9 to 4.10 upgrades we'd
+ensure that the next 4.8.z shipping after 4.9.12 has RHCOS bits and MachineConfig
+which offer parity with 4.9.12 so that it's not required that we reboot into
+4.9.12. If teams choose to pursue this option they will need to continue to ensure
+that 4.7 to 4.8.z and 4.8.z-n to 4.8.z upgrades continue to work as well.
+
+Teams which believe this is not achievable or the level of effort is extremely
+high should document those findings.
+
+Thus far RHCOS, Node, MCO, SDN, Containers, and PSAP teams are known to fall into
+this group of teams which have components coupled to the host.
+
+#### MCO - Widen Node Constraints to allow for N-2
+
+Building upon the EUS-to-EUS upgrade MVP work(https://github.com/openshift/enhancements/blob/master/enhancements/update/eus-upgrades-mvp.md#mco---enforce-openshifts-defined-host-component-version-skew-policies)
+to allow MCO to enforce host constraints we will broaden those constraints to
+enable host component version skew.
+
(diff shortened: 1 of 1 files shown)


sdodson:
CC @darkmuggle @rphillips @crawford @dcbw @miabbott @mrunalp @zvonkok @wking @vrutkovs 
PTAL, I know we still need to reach consensus on the MCO node constraints stuff but hopefully we can find a path forward on that while we discuss how we'd move on to the next objective of reducing reboots

enhancements/update/reduced_reboots.md
```diff
@@ -0,0 +1,328 @@
...
+#### OTA - Implement Downstream Paused Worker Pool Upgrade Tests
+
+In parallel with efforts to revamp upstream version skew testing we must also
+implement downstream version skew testing which includes any additional tests
+required for OpenShift specific implementation details.
+
+We will achieve this by delivering upgrade jobs which pause the Worker MachineConfigPool
+then upgrade from 4.x to 4.x+1 to 4.x+2. We will run conformance tests from 4.x
+after upgrading to 4.x+1 in order to ensure that we continue to provide a baseline
+feature set, then again after upgrading to 4.x+2, finally after unpausing the
```
wking, line 90: I'm not clear on why we couldn't run the 4.x+1 suite on the cluster once the control-plane reached 4.x+1.  It seems unlikely to me that the test suite is tightly tied to kubelet-side features, and that when it is, it requires 4.x+1 _compute_ nodes.  The distinction isn't critical, because leaving a pool paused long enough to port workloads to new APIs is unwise, but running the tests that match the current control plane seems convenient if it works.

sdodson: Yeah, it's not a straight forward question to answer. We know that under our current upgrade process most operators have to tolerate some incomplete upgrade state without exploding because MCO updates nodes (kubelets) toward the very end of the upgrade, but dodging problems there isn't quite the same as running a full test suite. At the same time it's not truly a 4.x+1 cluster so if we do choose to run the tests we should not be surprised if there's some portion which leverage new features that fail.

I have a feeling that where we'll run into problems is when we upgrade the control plane to 4.x+2 but we still have kubelets at 4.x. This is where we'll likely run into operators which have always assumed a baseline feature set of 4.x+1.

Perhaps node team have an opinion here? @rphillips @ehashman @harche

enhancements/update/reduced_reboots.md
```diff
@@ -0,0 +1,328 @@
...
+
+Thus far RHCOS, Node, MCO, SDN, Containers, and PSAP teams are known to fall into
+this group of teams which have components coupled to the host.
+
+#### MCO - Widen Node Constraints to allow for N-2
+
+Building upon the EUS-to-EUS upgrade MVP work(https://github.com/openshift/enhancements/blob/master/enhancements/update/eus-upgrades-mvp.md#mco---enforce-openshifts-defined-host-component-version-skew-policies)
+to allow MCO to enforce host constraints we will broaden those constraints to
+enable host component version skew.
+
```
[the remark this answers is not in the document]
zvonkok: Another constraint to consider would be `maxUnavailable: 0` in a MCP?

enhancements/update/reduced_reboots.md
```diff
@@ -0,0 +1,328 @@
...
+Admins who choose to would then be able to skip a host reboot by following this
+pattern:
+
+1. Starting with a 4.8 cluster, pause Worker MachineConfigPool
+1. Upgrade to 4.9
+1. Upgrade to 4.10
+1. Unpause Worker MachineConfigPool
```
zvonkok, line 134: What about custom MachineConfigPools?

zvonkok: One example would be e.g. real-time kernel workers that are in a separate MCP with additional MachineConfigs

enhancements/update/reduced_reboots.md
```diff
@@ -0,0 +1,328 @@
...
+more time and effort to testing efforts. Ultimately this introduces another
+dimension to an already complex testing matrix.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. Should we make these between specific named versions. ie: 4.6-4.8, and 4.8-4.10
+or should this be a standard N-2 rule, ie: 4.6-4.8, 4.7-4.9, 4.8-4.10?
+
```
zvonkok, line 177: What happens with an MCP which is "never" unpaused? Will MCO force an upgrade if the MCP violates the constraint N-1 or N-2

enhancements/update/reduced_reboots.md
vrutkovs: All pools should inherit from `worker`, so if users have custom pools these would be updated when MCO generates a new rendered config for worker on every step. Custom pools needs to be paused as well

enhancements/update/reduced_reboots.md
```diff
@@ -0,0 +1,328 @@
...
+  the same minor version expected previously.
+
+We may mitigate some of this by further delaying EUS-to-EUS upgades after normal
+minor version upgrades have been promoted to stable and allocating significantly
+more time and effort to testing efforts. Ultimately this introduces another
+dimension to an already complex testing matrix.
+
+## Design Details
+
+### Open Questions [optional]
```
zvonkok, line 173: To prevent an upgrade of specific nodes, the high-value assets of a cluster with special resources and constraints (kernel, os, ... ), one could create a custom MCP and set either `pause: true` or `maxUnavailable: 0` this would prevent MCO from updating this very MCP for the next 1 or 2 release depending on the constraint. 

An operator that does preflight checks on newer kernel, os could use the new information that is rolled out to the other nodes to check if the special resource would work on the update.

enhancements/update/reduced_reboots.md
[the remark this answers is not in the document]
sdodson: We're iterating on design with MCO, the original proposal to examine node specific details like kubelet version or kernel version may not be accepted in favor of something more abstract like the version of MachineConfig templates currently applied to nodes being greater than or equal to some version. At very high level we're just seeking to ensure that we enforce whatever host component version skew policy we come up with by setting Upgradeable=False to prevent upgrades that would violate those policies.

In 1) above MCO would only set Upgradeable=False when those constraints would be violated through a minor version upgrade, within this context we don't actually care if the pools are paused just that they're of some minimum version.

All operators are enabled to set Upgradeable=False, it's the primary mechanism used by operators to assert that the cluster is not in a state which would allow for minor version upgrades.

enhancements/update/reduced_reboots.md
sdodson: Do those inherit from Worker pool? This does make me wonder if we need to define our policies on a per pool basis, especially if we're considering that control plane pool must be updated.

@darkmuggle another point to consider here.

enhancements/update/reduced_reboots.md
sdodson: No, we'd just inhibit upgrades that fall outside of our defined policy. That can still be forced around but anytime an admin chooses to force an upgrade they become responsible for the disaster they create.

enhancements/update/reduced_reboots.md
sdodson: I think if an operator has tigheter constraints than can be expressed by either minimum version of MachineConfig, kubelet version, or kernel then that logic should live outside of the MCO. The MCO is caring about this only because it can affect the host component versions.

enhancements/update/reduced_reboots.md
zvonkok: Ok just for clarification, if I pause a custom MCP this property will not be backpropagated to the parent worker MCP. MCO will rollout the upgrade on the worker MCP and since my custom MCP is paused it will wait until availability.

enhancements/update/reduced_reboots.md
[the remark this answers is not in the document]
zvonkok: @sdodson That is fine, for such information we have an RFE for the machine-os-content to provide such information via annotations as part of the extension system. Just trying to understand the high-level constraints and the best way to "guard" special resources in a cluster from unintended upgrades.

enhancements/update/reduced_reboots.md
sdodson: @zvonkok reference for that? I'd be interested in the set of annotations that would provide.

zvonkok: @sdodson This is needed to support https://github.com/openshift/enhancements/pull/357, https://issues.redhat.com/browse/GRPA-2713

zvonkok: @sdodson This was a question. :) Forgot the question mark.

enhancements/update/reduced_reboots.md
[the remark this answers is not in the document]
zvonkok: The annotations are needed to support https://github.com/openshift/enhancements/pull/357 https://issues.redhat.com/browse/GRPA-2713

enhancements/update/reduced_reboots.md
zvonkok: @sdodson Understood, this is related to this: https://github.com/openshift/enhancements/pull/357 my questions around this is, how SRO can prevent upgrades and guard the special resources from unwanted upgrades. Some kmods are sensitive to any kernel version change and some kmods (kABI whitelisted symbols) only care of about OS major changes (8.x -> 9.x) .

enhancements/update/reduced_reboots.md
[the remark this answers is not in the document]
zvonkok: `Upgradeable=False` How and where is this exactly set?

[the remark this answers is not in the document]
zvonkok: @sdodson By all operators you mean operators that are managed via CVO and OLM? I had a question on my enhancement if OLM managed operators are allowed to create a ClusterOperator object to "control" upgradability.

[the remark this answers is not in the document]
wking: >  I had a question on my enhancement if OLM managed operators are allowed to create a ClusterOperator object to "control" upgradability.

I didn't think this was possible before, but turns out I was wrong, and the CVO [uses an unfiltered list of all ClusterOperator objects][1], regardless of whether they are from CVO manifests or added by third parties, when it calculates whether minor bumps are blocked.

[1]: https://github.com/openshift/cluster-version-operator/blob/bac69e03599f8c57f46585246618e462c31c2e93/pkg/cvo/upgradeable.go#L145

enhancements/update/reduced_reboots.md
wking: > Custom pools needs to be paused as well

I don't think any pools _need_ to be paused.  Folks interested in reducing reboots can just pause the pools in which they want to reduce reboots, for the short time it takes to make a few consecutive update hops.  I don't understand how pool-pausing interacts between the `worker` pool and its custom decedents; maybe the MCO folks can speak to that.

enhancements/update/reduced_reboots.md
wking: Never unpausing a pool will sooner or later fail the nodes out as the kubelets stop getting Kube-API CA rotations, even if the admins don't force an update.  We should be very clear that pool pauses are acceptable practice for a few hours, maybe into days, but that the whole time pools are paused, you are not getting CA rotations, bumped pull secrets, RHCOS bugfixes, and all the other good stuff that comes with having an unpaused pool.  Folks who want to leave a pool paused for longer should be considering all of those consequences, and deciding for themselves if the reduced disruption is worth the risk.

enhancements/update/reduced_reboots.md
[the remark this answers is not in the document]
zvonkok: @wking Thanks for pointing this out, started to read the code as well and doing some tests with custom MCPs and custom ClusterOperator objects.

enhancements/update/reduced_reboots.md
zvonkok: Users could have self-defined constraints that are not yet applied or implemented in a piece of software. Either an operator is preventing the upgrade or an admin wishes to do some "day 2" customizations before upgrading, knowing that an upgrade could break something.

zvonkok: What about workloads that are running _longer_ (days, weeks) that cannot be checkpointed, wouldn't this also be a reason to pause an MCP? PodDisruptionBudgets with minAvailable=100% would prevent draining to finish but the Node would already be cordoned. The workload may need additional Pods to be scheduled on the node to finish the task.

wking: > PodDisruptionBudgets with minAvailable=100% would prevent draining to finish but the Node would already be cordoned.

There is a balance between wasting resources by not scheduling work on the node while you wait for the slow pod to wrap up, and getting stuck in an endless drain because the new work you let onto the node ends up also being guarded by a PDB or other mechanism and taking longer than the initial slow pod.  Currently we weight in favor of quickest-possible-drain at the expense of underutilizing the available capacity.  Folks who want to minimize wasted capacity can set `maxUpgradeable: 1` (the default) on their MachineConfigPools, although the net effect is just spreading the waste out over a longer wall-clock duration as the nodes roll one after the other.  That spreading may still be useful if the overall pool doesn't have much extra capacity to spare.  And if folks feel really strongly, they can arrange to fill each node with workloads that will all wrap up around the same point in time, or twiddle tolerations to allow faster work onto the cordoned node until the slow work gets closer to wrapping up.

> The workload may need additional Pods to be scheduled on the node to finish the task.

It's up to the actor managing that workload to either say "you know what, we should clear this slow work off so that poor node can update" or "I am unwilling to abandon this slow work, so I'm going to set sufficient tolerations and PDB guards on these new Pods so they can go join the slow worker on the cordoned node and help it push through to completion".  I don't think that's something generic OpenShift tooling that doesn't understand the details of your workload can help you with out of the box.  In both of these cases, pausing the pool is one way to reduce the likelihood of cordons and node reboots.  But MachineConfigs are not the only reason your nodes might get cordoned or rolled.  I think it's better for folks who have really expensive, slow work to write their own really smart controllers who can shepherd the work through a generic Kube-node environment instead of using the coarse pool-pausing knob to try to orchestrate the dance they want to see.

zvonkok: @wking Right, we do not have to lengthen this. Either you want to update or you do not want to. If you want to update then make a plan on how to fit the upgrade procedure into your daily/weekly business.  We can offer the tools and mechanics but "you" should do "your" homework.

zvonkok:
@wking @sdodson I have the following use-case for upgrading. In my case, I can safely upgrade the operator but not the node. If I set `Upgradeable=False` the operator and the node will not be updated. I would like to have an updated operator but not the node. Should we keep here simple and just say the operator just sets Upgradeable=False and does not any special cases with pausing an MCP but letting the cluster upgrade all operators. 

The benefit of pausing a special MCP that all operators and all other MCPs would be up to date. If the admin is "brave" he can still unpause if he is willing to take the risk of having an incompatible "driver".

enhancements/update/reduced_reboots.md
```diff
@@ -0,0 +1,328 @@
+---
+title: reduced-reboot-upgrades
+authors:
+  - "@sdodson"
+reviewers:
+  - @darkmuggle
```
darkmuggle, line 6:
```suggestion
  + @crawford 
```

darkmuggle: Alex will be driving the MCO side of this discussion.

enhancements/update/reduced_reboots.md
```diff
@@ -0,0 +1,328 @@
...
+
+#### Node - Improve Upstream Kubelet Version Skew Testing
+
+Kubernetes defines a version skew policy(https://kubernetes.io/docs/setup/release/version-skew-policy/#kubelet)
+which allows for kubelet N-2 to be compatible with kube-apiserver version N. At
+this point in time OpenShift is not comfortable with the level of testing upstream
+and the intersection of the specific features of OpenShift. We should work to
+define and implement upstream testing changes which give us an appropriate level
+of confidence that N-2 version skew issues would be identified in the community
+whenever possible.
```
ehashman, line 79: Note that this is not currently verified in upstream CI -- I raised this for the [SIG Architecture agenda tomorrow](https://groups.google.com/g/kubernetes-sig-architecture/c/QX-4qq2krm0/m/998T3cJUBQAJ) to ensure that this has test coverage or to clarify the support policy.

brenton:
@crawford, @derekwaynecarr, Anything blocking this from merge from your end?  On the OTA side we're ready for this to merge.

sdodson:
This was meant to be follow-on work from the MVP work that was outlined in #762 and #585 however, actual [implementation](https://github.com/openshift/machine-config-operator/pull/2658) has moved forward so i'm going ahead and closing this. I will follow up with changes to existing content for anything that needs to be pulled over.

/close

PR #645, opened 2021-02-11
eus-upgrades-mvp: Fix MCO constraints examples

sdodson:


enhancements/update/eus-upgrades-mvp.md
```diff
@@ -122,19 +122,24 @@ version away and *block* upgrades for removals which are imminent.
...
 ```yaml
 node.status.nodeInfo.kubeletVersion:
 - v1.20
-- v1.19
 ```
 
-If the policy were to change to allow for a version skew of N-2, v1.18 would be
+If the policy were to change to allow for a version skew of N-2, v1.19 would be
 added to the list of acceptable matches. The MCO is not responsible for defining
 these constrains and constraints are only widened whenever we have CI testing
```
dhellmann, line 134: While we're in here...

```suggestion
these constraints and constraints are only widened whenever we have CI testing
```

dhellmann (approved):
This update looks right. I noticed one typo near a change you were already making so proposed an update.

enhancements/update/eus-upgrades-mvp.md
sdodson: Thanks! I gave this section an edit after reading over it again and I've addressed both of those.

sdodson:
@darkmuggle PTAL, I just noticed this error which may have led to some confusion for you and MCO team.

kikisdeliveryservice:
would like this discuss this in our meeting today :smile:

wking (approved):
/lgtm

This makes sense to me and adds clarity to Scott's plan.  Any follow-up pivots can come in via follow-up PRs.

PR #649, opened 2021-02-12
conventions: update the resources and limits section

dhellmann:
Expand the "Resources and Limits" section of the conventions document
with more detail about why we have the conventions we do, where the
formula for CPU resource requests came from, and links to the history
of work on resource requests.

/cc @smarterclayton @markmc

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -141,20 +141,154 @@ it is used to manage all elements of the cluster.
 
 #### Resources and Limits
 
-All cluster components must declare resource requests for CPU and memory, and should not describe limits.
+Kubernetes allows Pods to declare their CPU and memory resource
+requirements in advance using [requests and
+limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/). Requests
+are used to ensure minimum resources are provided and to influence
+scheduling, while limits prevent Pods from consuming resources
+excessively.
 
-The CPU request of components scheduled onto the control plane should
-be proportional to existing etcd limits for a 6 node cluster running
-the standard e2e suite (if etcd on a control plane component uses 600m
-during an e2e run, and requests 100m, and the component uses 350m,
-then the component should request `100m/600m * 350m`).  Components
-scheduled to all nodes should be proportional to the SDN CPU request
-during the standard e2e suite.
+Unlike with user workloads, setting limits for cluster components is
+problematic for several reasons:
 
-The memory request of cluster components should be set at 10% higher than their p90 memory usage over a standard e2e suite execution.
+* Components cannot anticipate how they scale in usage in all customer
+  environments, so setting one-size-fits-all limits is not possible.
+* Setting static limits prevents administrators from responding to
+  changes in their clusters’ needs, such as by resizing control plane
+  nodes to provide more resources.
+* We do not want cluster components to be restarted based on their
+  resource consumption (for example, being killed due to an
+  out-of-memory condition). We need to detect and handle those cases
+  more gracefully, without degrading cluster performance.
 
-Limits should be not set because components cannot anticipate how they scale in usage in all customer environments and a crashlooping OOM component is something we must detect and handle gracefully regardless. Setting memory limits leaves administrators with no option to react to valid changes usage dynamically.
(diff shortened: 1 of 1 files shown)


CONVENTIONS.md
```diff
@@ -141,20 +141,129 @@ it is used to manage all elements of the cluster.
...
+the requirement to be efficient and support small-footprint
+deployments. If the request settings for a component are too low, it
+could be starved for resources in a very busy cluster, resulting in
+degraded performance and lower quality of service. If the request
+settings for a component are too high, it could mean that the
+scheduler cannot find a place to run it, leading to crash looping or
+preventing a cluster from deploying. If the combined resource requests
+for the cluster components are too high, users may not have enough
+resources available to use OpenShift at all, resulting in lost
+business (see especially the [single-node](enhancements/single-node)
```
smarterclayton, line 184: This one feels on the nose, you might say "inability to run in reasonable footprints for end users"

CONVENTIONS.md
```diff
@@ -141,20 +141,129 @@ it is used to manage all elements of the cluster.
...
+use cases).
+
+Kubernetes resource requests for CPU time are measured in fractional
```
smarterclayton, line 187: I think having a paragraph before this to distinguish and set tone for future use cases is important - distinguishing between best practice for setting requests on compressible resources (CPU) and best practices on incompressible resources (memory).  For incompressible we wish to set a request that is roughly aligned to what we will use.  For compressible resources we are more concerned about balancing between proportional users - ensuring system or user workloads are not unfairly preferred (if system compressible requests are too low, user workloads will take priority and lead to impact to the user workload, vice versa some user workloads will not schedule) as well as ensuring that in resource constrained situations there is enough proportional CPU for the system to make progress.

CONVENTIONS.md
```diff
@@ -141,20 +141,129 @@ it is used to manage all elements of the cluster.
...
+"[Kubernetes
+CPUs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu)"
+expressed using the unit “millicpu”, or 1/1000th of a CPU. For
```
smarterclayton, line 190: Hrm, we almost always call these `millicores`.  I was surprised you used `millicpu`.

CONVENTIONS.md
```diff
@@ -141,20 +141,129 @@ it is used to manage all elements of the cluster.
...
+to other components based on their actual consumption. There are two
+classes of cluster components to consider, those deployed only to the
+control plane nodes and those deployed to all nodes.
+
+The etcd database is the backbone of the entire cluster. If etcd
+becomes starved for CPU resources, other components like the API
+server and controllers that use the API server will suffer cascading
+issues that will eventually cause the cluster to fail. Therefore, we
+use etcd’s requirements as the baseline for a formula for other
+components running on the control plane.
```
smarterclayton, line 208: ... for other compressible resources

smarterclayton:
Few minor things, otherwise looks good.  Thank you for diving into this.

CONVENTIONS.md
dhellmann: millicpu came from https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu

It does say that millicore is often used to mean the same thing, but a core and a CPU are only the same when hyperthreading is not enabled. I've updated this section to mention both names.

CONVENTIONS.md
dhellmann: Done

CONVENTIONS.md
dhellmann: I don't fully understand what you mean by "compressible" and "incompressible". Can you explain those terms?

sdodson: CPU is compressible because if there's less of it you generally just go slower, storage resources (RAM, disk) are incomprehensible because when you run out of those you halt.

https://en.wikipedia.org/wiki/System_resource#Categories

dhellmann: TIL, thanks @sdodson 

I've added a paragraph combining what you both said at line 187 in the latest draft.

PR #650, opened 2021-02-12
Add ClusterOperator Scheduling

michaelgugino:


enhancements/scheduling/clusteroperator-scheduling.md
```diff
@@ -0,0 +1,108 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+To ensure system components, especially ClusterOperators, have the correct
+set of tolerations.
```
wking, line 29: Possibly PR with adjustments to [this existing enhancements content][1], instead of creating a new, parallel enhancement?

[1]: https://github.com/openshift/enhancements/blob/feed05c5c9838fa9476efd650cc968a15ac6b42b/CONVENTIONS.md#taints-and-tolerations

enhancements/scheduling/clusteroperator-scheduling.md
```diff
@@ -0,0 +1,108 @@
...
+This is a special key that is applied when a node is cordoned, such as during
+drain.  Since this is a NoSchedule taint, anything without this taint will
+not automatically be evicted, it will only prevent future scheduling.
+
+During upgrades, we cordon and drain all nodes.  Sometimes the scheduler will
+place pods tolerating NoSchedule onto a node quicker than the loop for
+draining a node can complete.  This results in hot-looping between draining
+and the scheduler.
+
+Any other NoSchedule toleration is fine.
```
smarterclayton, line 88: In the spirit of Chesterton fence, I’d like to see a review of the components that thought they *should* set this and why (representative sample).

That ensures we correct misconceptions about when the toleration is appropriate.

If it was accidental, what would have prevented the accident?

enhancements/scheduling/clusteroperator-scheduling.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Open Questions [optional]
+
+
+
+### Test Plan
+
```
smarterclayton, line 95: I assumed we would enforce this via a test?  Or did I miss that above?

enhancements/scheduling/clusteroperator-scheduling.md
michaelgugino: I should have used grep!  I looked in the scheduling folder but didn't find this obviously.

enhancements/scheduling/clusteroperator-scheduling.md
michaelgugino: The one I know of for sure is the CVO.  I don't have data if anything else is doing it or not.  I will take a look.

michaelgugino: Looks like the CVO is the only-non daemonset pod that has this toleration, at least from a typical CI run.  The daemonsets get that toleration dynamically, so everything else looks correct.

ravisantoshgudimetla: There was a discussion [sometime ago](https://github.com/openshift/cluster-version-operator/pull/182#discussion_r280956544) on whether CVO should tolerate `NoSchedule` taint and the answer was `yes`. Has something changed now? I was always in favour of specific tolerations instead of blanket tolerations. How about adding a taint called `drain-in-progress` to node with `NoSchedule` effect. That won't be tolerated by CVO which should help in proceeding with draining smoothly.

enhancements/scheduling/clusteroperator-scheduling.md
ravisantoshgudimetla: I think we [do](https://github.com/openshift/origin/blob/53784e6548d24422f9c227dbd466a85082db2f59/test/extended/operators/tolerations.go)

ravisantoshgudimetla (commented):
Thank you for initiating this discussion @michaelgugino. This looks like a step in right direction.

enhancements/scheduling/clusteroperator-scheduling.md
wking: Looks like that `ensure control plane operators do not make themselves unevictable` does not cover `node.kubernetes.io/unschedulable` today?  Just unreachable and not-ready.

enhancements/scheduling/clusteroperator-scheduling.md
ravisantoshgudimetla: I am not entire sure of that, IIUC I got a firm `yes` on `unschedulable` however if you think this was not needed, I can change the tolerations of CVO.

CONVENTIONS.md
```diff
@@ -222,6 +225,10 @@ spec:
       - operator: Exists
 ```
 
+Tolerating all taints should be reserved for DaemonSets and static
```
ravisantoshgudimetla, line 228: Which static pods are you referring to?

michaelgugino: Is there more than one type?

enhancements/scheduling/clusteroperator-scheduling.md
michaelgugino: There are many taints that have the effect 'NoSchedule' but there is one 'node.kubernetes.io/unschedulable' taint with the effect 'NoSchedule'.  It appears in the previous discussion only NoSchedule was considered rather than the specific taints (at the least, node.kubernetes.io/unschedulable was not itself considered directly).

CONVENTIONS.md
ravisantoshgudimetla: I meant which specific static pods in OpenShift. IIUC, static pods should be tolerating all `NoExecute` taints not `NoSchedule` as their manifests are directly present on the node, there is no need for them to go through scheduling cycle.

enhancements/scheduling/clusteroperator-scheduling.md
ravisantoshgudimetla: any `NoSchedule` taint != `unschedulable` condition. `unschedulable` condition == `node.kubernetes.io/unschedulable` taint. unschedulable is a node condition which is being represented via a special well known taint called `node.kubernetes.io/unschedulable`.

michaelgugino: I'm trying to use the k8s precise language.  There are taints, and those taints have effects.  The effect we're discussing is NoSchedule.  That effect is caused by one or more taints.

I did not see the taint node.kubernetes.io/unschedulable discussed in the link you provided.  There were only discussions around the NoSchedule effect.

CONVENTIONS.md
michaelgugino: There isn't an option to tolerate all taints of a certain effect, just "All" generally.

CONVENTIONS.md
```diff
@@ -189,6 +189,9 @@ following taints if doing so is necessary to form a functional Kubernetes node:
 Operators should not specify tolerations in their manifests for any of the taints in
 the above list without an explicit and credible justification.
 
+Operators should never specify the following toleration:
```
dhellmann, line 192: The other guidelines in this section include justification, so that someone coming to the doc without all of the background can learn why we have each restriction. This addition would benefit from a sentence or two explaining why this value is so special that it should never be used.

dhellmann: Thanks, that looks good.

smarterclayton:
/lgtm
/approve

In the future we may come up with exceptions, which are reasonable to propose as new enhancements.  We may also wish to take this feedback back to sig-scheduling and gather other perspectives on whether tolerations are missing something in between "list" and "all".

PR #651, opened 2021-02-12
Implement forwarder-label-selector in 2 phases.

alanconway:
Recommend that this be implemented in 2 phases to get a useful feature as quickly as possible.
Updated to reflect the implementation approch of the PR [https://github.com/openshift/cluster-logging-operator/pull/865

/cc @jcantril
/cc @sichvoge

### DIFF
--- a/enhancements/cluster-logging/forwarder-label-selector.md
+++ b/enhancements/cluster-logging/forwarder-label-selector.md
@@ -22,7 +22,7 @@ status: provisional
 
 ## Summary
 
-Add an input selector to the ClusterLogForwarder (CLF) to forward application
+Add an input selector to the ClusterLogForwarder (CLF) to collect application
 logs only from pods identified by labels.
 
 Kubernetes has two ways to identify pods: namespaces and labels.  The CLF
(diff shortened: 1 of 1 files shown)


sichvoge:
Ah this PR just moves the original proposal to two separate phases. Right?

alanconway:
> Ah this PR just moves the original proposal to two separate phases. Right?

Right, along with some minor clarifications - no substantial changes. Care to /lgtm ?

jcantrill (commented):
/lgtm

enhancements/cluster-logging/forwarder-label-selector.md
```diff
@@ -22,7 +22,7 @@ status: provisional
 
 ## Summary
 
-Add an input selector to the ClusterLogForwarder (CLF) to forward application
+Add an input selector to the ClusterLogForwarder (CLF) to collect application
```
sichvoge, line 25: Shouldn't that be still "forward" instead of "collect"? We will not "exclude" anything at the collector level right?

alanconway: From the users point of view, we _are_ excluding logs from being collected by that input - no pipeline connected to that input will see those logs. 
What we are doing under the covers is not the users business and could change in future - you're correct that presently we collect everything and filter inside fluetnd, but it would be a great optimization to avoid reading some log files altogether if we know their contents will be completely filtered out before we get to forwarding.

PR #652, opened 2021-02-15
WIP: Enable cgroup v2 support

harche:
Add support for cgroup v2 in OpenShift


m-yosefpor:
Any updates on this? I think k8s v1.22 is the version which officially supports cgroupv2 https://github.com/kubernetes/enhancements/issues/2254 (although it was possible to make k8s work with cgroupv2 from prior versions). Also FCOS 34 and later have cgroupv2 enabled by default (however it is disabled in Openshift images)

Is there a plan for Openshift/OKD 4.9 to have an option for enabling that?

P.S. If I understand correctly, it is even possible just now to switch to cgroup v2 by providing kernel args as MC. @giuseppe  Is it the only required change to make openshift work with cgroup v2?

giuseppe:
> P.S. If I understand correctly, it is even possible just now to switch to cgroup v2 by providing kernel args as MC. @giuseppe Is it the only required change to make openshift work with cgroup v2?

yes, that is the only change required for Kubernetes to switch to cgroup v2

enhancements/node/node-cgroupv2.md
```diff
@@ -0,0 +1,174 @@
...
+
+Cgroup v2 brings in bunch of improvements over it's predecessor. Please refer to the official kernel [documentation](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html) for more information.
+
+### Goals
+
+* Enable cgroup v2 across the cluster as an optional feature
+
+### Non-Goals
+
+* Until we have more in-depth understanding of the impact of enabling cgroup v2 on the workloads, we won't make it default and replace it with cgroup v1.
```
m-yosefpor, line 46: Shouldn't we start considering it as default now?

enhancements/node/node-cgroupv2.md
```diff
@@ -0,0 +1,174 @@
...
+    - systemd.unified_cgroup_hierarchy=1
+    - cgroup_no_v1="all"
+    - psi=1
+
+```
+
+#### Enable cgroup v2 on master nodes
+```yaml
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
```
m-yosefpor, line 129:
```suggestion
kind: MachineConfig
```

enhancements/node/node-cgroupv2.md
```diff
@@ -0,0 +1,174 @@
...
+`launch https://github.com/openshift/installer/pull/4648`
+
+### Day-2 operation scenario
+
+Users can enable cgroup v2 in OpenShift cluster on Day-2 by using the [existing support](https://github.com/openshift/machine-config-operator/blob/906673c4cab9dc5ad3f50b7481c8f02d26240a5d/docs/MachineConfiguration.md#kernelarguments) for passing the kernel arguments in MCO.
+
+#### Enable cgroup v2 on worker nodes
+```yaml
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
```
m-yosefpor, line 112:
```suggestion
kind: MachineConfig
```

m-yosefpor:
@harche  I think we need to add a `### Risks and Mitigations` sections.

All apps which are deployed on OKD should be revisited to ensure they are not depending on cgroupv1 hierarchy. E.g. right now cluster logging operator has an issue with cgroup v2: see https://github.com/openshift/origin-aggregated-logging/pull/2182.

Right now we are running our OKD clusters with cgroup v2 enabled, and everything seems to work fine (except for CLO).

m-yosefpor:
In a recent OKD working groups, there is this item:

> 4.9 - enable cgroupsv2 on new installs? Autoenable on 4.10?

https://hackmd.io/YJBn04R5TDi5Sm9XbOGwZA

vrutkovs:
@m-yosefpor this enhancement is for both OCP and OKD. In OKD we have a liberty of enabling some features before they have landed in OCP

PR #653, opened 2021-02-15
Adding owners file for cluster-logging subdirectory

ewolinetz:
As the cluster-logging group moves towards further independence from core OCP and becomes more of a second day operator group, it would be useful for our leads/architect to be able to approve our own enhancement proposals.


russellb:
Makes sense to me.  In general, reviewers of a given enhancement can change based on the scope.  For things internal to cluster logging, leads of that team make sense.

/approve

PR #654, opened 2021-02-15
ARO private DNS zone resource removal

jim-minter:
ARO private DNS zone resource removal

### DIFF
--- a/enhancements/dns/aro-private-dns-zone-resource-removal.md
+++ b/enhancements/dns/aro-private-dns-zone-resource-removal.md
@@ -1,0 +1,180 @@
+---
+title: aro-private-dns-zone-resource-removal
+authors:
+  - "@jim-minter"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2021-02-15
+last-updated: 2021-02-15
+status: provisional
+see-also:
+  - "/enhancements/dns/plugins.md"
+replaces:
+superseded-by:
+---
+
+# ARO private DNS zone resource removal
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This proposal covers removing the Azure private DNS zone resource from the
+architecture of Azure Red Hat OpenShift clusters.
+
+## Motivation
+
+Today on Azure, OCP (and hence ARO) clusters use an Azure private DNS zone to
+resolve {api,api-int,\*.apps}.*cluster domain*.  For this to work, it is a
+requirement that the Azure vnet in which the VMs are provisioned be configured
+to use the default Azure DNS resolver.  This is problematic in ARO for multiple
+reasons:
+
+1. Customer DNS configuration errors are a cause of toil and unactionable
+   incidents for ARO SREs.  DNS configuration errors also block cluster
+   upgrades.
+
+   ARO customers expect to be able to reconfigure vnet DNS and regularly do.
+   The ARO service can neither prevent this nor roll back a customer
+   configuration change because it may adversely affect non-ARO workloads
+   running in the same vnet.
+
+   Any process which involves restarting an ARO VM is dependent on DNS
+   configuration to be correct.  If DNS is misconfigured, VMs which restart will
+   fail to rejoin the cluster because they cannot resolve api-int.*cluster
+   domain*.  Impacted processes include VM upgrade, VM configuration change,
+   cluster scale up, VM crash recovery, cluster cold start, manual VM restart,
+   etc.
+
+   ARO monitors cluster node readiness and automatically creates an incident
+   when a cluster node is not ready.  If the root cause is DNS misconfiguration,
+   the service is dependent on communicating with the customer and the customer
+   resolving the problem.  This is an expensive process.  During this time the
+   incident is unactionable and key service processes are blocked.
+
+1. ARO customers regularly want to configure their clusters to use custom DNS
+   servers.  They can use [OCP DNS
+   forwarding](https://docs.openshift.com/container-platform/4.6/networking/dns-operator.html#nw-dns-forward_dns-operator)
+   to achieve this from pod context, but this does not work from the node
+   context and hence cannot be used to resolve on-premise container registries.
+   It also represents additional configuration load that the customer must
+   carry.
+
+   Some customers need to resolve on-premise registries with wildcard DNS names,
+   preventing the use of MachineConfig/DaemonSet workarounds which would modify
+   /etc/hosts.
+
+1. ARO intends to enable zero Internet egress by default for ARO clusters.  It
+   is likely that in order to implement this, ARO will need to be able to
+   selectively override DNS resolution of certain domains in both node and pod
+   contexts.
+
+### Goals
+
+* ARO cluster functionality should not be dependent on Azure vnet DNS
+  configuration.  Azure private DNS zone resources should not be part of the ARO
+  architecture.  The changes required are implemented in ARO, not OCP.
+
+### Non-Goals
+
+## Proposal
+
+ARO would like OCP's approval to remove the private DNS zone resource from its
+cluster architecture.  Instead, we propose to resolve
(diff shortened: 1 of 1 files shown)


jim-minter:
@derekwaynecarr @knobunc @frobware @ironcladlou @nstielau
(cc @mjudeikis @bennerv @jewzaam)

enhancements/dns/aro-private-dns-zone-resource-removal.md
```diff
@@ -0,0 +1,180 @@
...
+cluster architecture.  Instead, we propose to resolve
+{api,api-int,\*.apps}.*cluster domain* DNS addresses via an inline dnsmasq proxy
+resolver running as a systemd service on each cluster VM.  The dnsmasq proxy
+resolver would be backed by whichever DNS server(s) the customer specifies in
+their vnet configuration.
+
+A working PoC-quality implementation of this approach can be found at
+https://github.com/Azure/ARO-RP/pull/1296.  This implementation is notably
+similar to the following [Machine Config Operator
+code](https://github.com/openshift/machine-config-operator/blob/master/templates/common/on-prem/files/coredns-corefile.yaml),
```
cybertron, line 101: We've actually proposed a similar thing for on-prem platforms: https://github.com/openshift/machine-config-operator/pull/2374

Reading through this I'm wondering if that could replace the existing coredns implementation entirely. The only thing we need coredns for at this point is resolution of the other nodes, which it seems is not actually required anymore either. My current PR still pulls a container to template a config value, but if we removed coredns then it wouldn't need that. This may be useful cross-platform behavior.

I'll try updating my change to move everything into dnsmasq to verify that will work for us too. Maybe there won't be much overlap since we get our IPs from a different place, but it's worth looking into.

romfreiman: cc: @tsorya

enhancements/dns/aro-private-dns-zone-resource-removal.md
```diff
@@ -0,0 +1,180 @@
...
+IP address changes.  This is not expected to be a likely occurrence.  ARO
+already protects against this case by preventing customers from manually
+modifying the configuration of the Azure cluster load balancer resources.  ARO
+would be responsible for automatically handling *.apps IP address changes (also
+unlikely) as well as defining a process to recover from api-int IP address
+changes if ever needed.
+
+Removing the ability to resolve VM hostnames via DNS could conceivably impact
+end user workloads, but the likelihood of this is considered low.  We don't
+consider the ability to resolve VM hostnames via DNS part of the ARO platform
```
romfreiman, line 152: what about oc logs / oc debug commands - will they work without hosts being resolvable?

cybertron: I checked oc logs, debug, and rsh in my local env without node names being resolvable and they all worked fine.

mjudeikis:
ARO been running this configuration for last 2 months. We can confirm it works very well. Please considering merging it as we can own aligning ARO with wider azure community.

enhancements/dns/aro-private-dns-zone-resource-removal.md
```diff
@@ -0,0 +1,180 @@
...
+  the Kubelet config, we don't believe this is currently a hard platform
+  requirement.
+
+* ARO machine config server URL and serving certificate would use api-int IP
+  address instead of DNS.  This is because DNS resolution of api-int would not
+  be available at ignition time.
+
+* The ARO cluster installation process would prepopulate the
+  openshift-ingress/router-default LoadBalancer service early with the
+  appropriate IP address.
```
Miciah, line 134: What purpose does this change serve?

enhancements/dns/aro-private-dns-zone-resource-removal.md
```diff
@@ -0,0 +1,180 @@
...
+
+* Ingress operator dynamic DNS would be disabled on ARO via overridden
+  configuration of the DNS custom resource.
```
Miciah, line 137: Do you mean the operator's publishing to the private zone would be disabled by overriding the private zone in the cluster DNS config, or are you trying to disable all DNS publishing by the operator?

enhancements/dns/aro-private-dns-zone-resource-removal.md
mjudeikis: Routes are resolvable eary in the process of cluster creation AND we avoid double node rotation by removing need to inject this later in the process and causing nodes to rotate twice.

enhancements/dns/aro-private-dns-zone-resource-removal.md
mjudeikis: Yes. I managed services customer does not own or manage DNS and cluster does not have any access to do so. So it is managed from outside. This should not be for scope for everybody but we need this for ARO specifically.

sdodson:
Where we can control parameters of the cluster which are critical to cluster health we should. I believe we should not leave internal, critical dns, up to externally managed dns and would prefer that if we pursue using dnsmasq we continue to use private dns where available and forward requests for critical resources to that resolver using dnsmasq's server directive. This would allow for flexible future dns management without requiring all nodes be reconfigured.

While there are other separate requests to remove cloud private dns zones they're addressing different concerns and the two solutions could target different cluster configuration requirements. Unless we believe there's near total overlap we can address the two needs separately.

mjudeikis:
So what is the next actions, considering we are not the team doing the work?

bennerv:
This has been closed for awhile, but this ask is still valid from the ARO team.  Wondering if we can get a re-review and status update.

bennerv:
@patrickdillon #1276  is a bit different than what we were looking for in this enhancement.  

We're looking to allow customers to set node-level DNS resolution while maintaining resolvability api, api-int, and *.apps within the cluster.  We already do allow this and have been running this infra for almost 2 years: https://learn.microsoft.com/en-us/azure/openshift/howto-custom-dns

>   ARO customers expect to be able to reconfigure vnet DNS and regularly do.
   The ARO service can neither prevent this nor roll back a customer
   configuration change because it may adversely affect non-ARO workloads
   running in the same vnet.
   

#1276 would be better named as "Add option to not provision DNS records in Route53 for AWS IPI Install".  That's not changing node-level DNS resolution, but rather: "hey we don't want to use route53, Azure DNS zones, ... please allow us to create the api and *.apps records on our own DNS server.  

They do overlap a bit, but we already have this option on ARO with customers bringing their own domain: https://learn.microsoft.com/en-us/azure/openshift/tutorial-create-cluster#prepare-a-custom-domain-for-your-cluster-optional

patrickdillon:
> @patrickdillon #1276 is a bit different than what we were looking for in this enhancement.
> 
> We're looking to allow customers to set node-level DNS resolution while maintaining resolvability api, api-int, and *.apps within the cluster. We already do allow this and have been running this infra for almost 2 years: https://learn.microsoft.com/en-us/azure/openshift/howto-custom-dns
> 

Allowing customers to set the node-level DNS resolution is the ultimate goal, yes, but the blocker that is preventing customers from doing that is the DNS private zones.  #1276 proposes removing private dns zones and instead resolving the records with coredns, which is a different implementation than #1276/ARO, but should be sufficient...

It would be helpful to understand what else is missing or if we are overlooking anything.

bennerv:
@patrickdillon will that be a successor enhancement?  There's more to consider when swapping out node DNS servers in `/etc/resolv.conf` over just allowing customers to create the DNS A records within their own DNS service.  
- Resolution of quay, ARM, cloud.redhat.com, api.redhat.com, etc. (if a customer sets the node DNS resolver and it doesn't have resolvability of these domains)
- Ensuring resolution of api-int from a node standpoint when a new worker joins the cluster
- probably more

patrickdillon:
> will that be a successor enhancement? There's more to consider when swapping out node DNS servers in /etc/resolv.conf over just allowing customers to create the DNS A records within their own DNS service.
> 
>     Resolution of quay, ARM, cloud.redhat.com, api.redhat.com, etc. (if a customer sets the node DNS resolver and it doesn't have resolvability of these domains)
>     Ensuring resolution of api-int from a node standpoint when a new worker joins the cluster
>     probably more I can't think of right now

These questions should be answered in #1276 in the context of AWS, but the pattern should be almost identical to Azure and the main discussion will take place in this enhancement. It would be great to have your feedback there. 

API-int resolution will be handled by coredns, and that solution should be described in #1276.

I would expect Quay, cloud.redhat.com etc to all be publicly resolvable and should be handled by upstream DNS servers. I'm not sure I see a problem here... is the concern that users would switch to custom DNS that can't resolve these?

cybertron:
Is that valid though? I don't think it's unreasonable to say that the upstream servers need to be able to resolve the necessary records if you're doing something custom.

PR #656, opened 2021-02-15
CONSOLE-2355: Update Quick Start proposal with i18n

rebeccaalpert:
I added an approach for handling internationalized Quick Start content.

Sam outlined [several possible approaches](https://issues.redhat.com/browse/CONSOLE-2355?focusedCommentId=15184163&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15184163). I went with this approach for the proposal update because it's dead simple; let me know if another approach is preferred. We should be able to use this approach for other extension resources like ConsoleLink and ConsoleCLIDownload as well.

We can always store the value of the i18next.languages vs. supported languages check in localStorage (either in bridge/language or a different attribute) so we don't need to run it every time we decide whether to render an extension resource.

Fixes https://issues.redhat.com/browse/CONSOLE-2355.

### DIFF
--- a/enhancements/console/quick-starts.md
+++ b/enhancements/console/quick-starts.md
@@ -2,6 +2,7 @@
 title: quick-starts
 authors:
   - "@jhadvig"
+  - "@rebeccaalpert"
 reviewers:
   - "@spadgett"
   - "@alimobrem"
@@ -9,12 +10,13 @@ reviewers:
 approvers:
   - "@spadgett"
 creation-date: 2020-06-02
-last-updated: 2020-09-14
+last-updated: 2021-03-10
 status: implementable
 see-also:
   - "https://issues.redhat.com/browse/CONSOLE-2255"
   - "https://issues.redhat.com/browse/CONSOLE-2232"
   - "https://issues.redhat.com/browse/SRVLS-262"
+  - "https://issues.redhat.com/browse/CONSOLE-2355"
 ---
 
 # Quick Starts
@@ -91,12 +93,15 @@ As a operator creator I want to provide operator consumers with a guide on how t
    ![help-menu](https://raw.githubusercontent.com/jhadvig/images/master/help-menu.png)
 
 
-QuickStarts CR for [Explore Serverless](https://marvelapp.com/236ge4ig/screen/69908905):
+Quick Starts CR for [Explore Serverless](https://marvelapp.com/236ge4ig/screen/69908905):
 ```yaml
 apiVersion: console.openshift.io/v1
 kind: QuickStart
 metadata:
   name: explore-serverless
+  labels:
+    console.openshift.io/lang: en
+    console.openshift.io/name: explore-serverless
 spec:
   displayName: Explore Serverless
   tags:
@@ -104,7 +109,7 @@ spec:
   duration: 10
   description: Install the Serverless Operator to enable containers, microservices and functions to run "serverless"
   prerequisites: Release requirements if any Install X number of resources.
-  introduction: Redhat OpenShift Serverless is a service based on the open source Knative project. It provides ...
+  introduction: Red Hat OpenShift® Serverless is a service based on the open source Knative project. It provides ...
   tasks:
     - title: Install Serverless Operator
       description: The OperatorHub is where you can find a catalog of available Operators to install on your cluster ...
@@ -150,6 +155,62 @@ tour. User's access to the Quick Start is reviewed based on array of
 resource actions. In order for the user to see the Quick Start and
 take its tour, his access review needs to pass all the listed
 resources actions.
+
+#### Quick Start Internationalization
+
+OpenShift can be toggled between multiple languages. The Quick Start
+CRD contains a series of labels to specify the Quick Start name and language:
+```yaml
+console.openshift.io/lang: en
+console.openshift.io/country: gb
+console.openshift.io/name: explore-serverless
+```
+
+The optional `console.openshift.io/lang` label specifies the two-letter
+[ISO 639-2 Language Code](https://www.loc.gov/standards/iso639-2/php/code_list.php))
+(i.e. "en" for English) for the Quick Start. Quick Starts without this label will be treated
+as English Quick Starts.
+
+The optional `console.openshift.io/country` label specifies the alpha-2
+[ISO-3166 Country Code](https://www.iso.org/iso-3166-country-codes.html)
+(i.e. "gb" for the United Kingdom of Great Britain and Northern Ireland).
+Quick Starts without this label will only rely on the `console.openshift.io/lang`
+behavior.
+
+The `console.openshift.io/name: explore-serverless` label specifies the name of the
+Quick Start. This should be consistent across all translations of a given Quick Start,
+as it allows us to avoid showing the same Quick Start content in multiple languages.
+If a Quick Start is not translated into a user's language, the English version will be shown.
+
+OpenShift can display the correct
+Quick Starts for each language by comparing the value of the
+`console.openshift.io/lang` and `console.openshift.io/country` labels to the currently displayed language in OpenShift.
+See the [internationalization enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/console/internationalization.md) for more information.
+This approach will not handle languages we don't offer support for. If Quick Starts
+in non-supported languages become a need, we will need to modify this approach.
+
+Example Explore Serverless Quick Starts CR snippet for Japanese:
+```yaml
+apiVersion: console.openshift.io/v1
+kind: QuickStart
+metadata:
+  name: explore-serverless
+  labels:
+    console.openshift.io/lang: ja
+    console.openshift.io/name: explore-serverless
+```
+
+Example Explore Serverless Quick Starts CR snippet for English (the United Kingdom of Great Britain and Northern Ireland):
+```yaml
+apiVersion: console.openshift.io/v1
+kind: QuickStart
+metadata:
+  name: explore-serverless
+  labels:
+    console.openshift.io/lang: en
+    console.openshift.io/country: gb
+    console.openshift.io/name: explore-serverless
+```
 
 #### Air Gapped Environments
 


enhancements/console/quick-starts.md
```diff
@@ -151,6 +154,23 @@ resource actions. In order for the user to see the Quick Start and
 take its tour, his access review needs to pass all the listed
 resources actions.
 
+#### Quick Start Internationalization
+
+OpenShift can be toggled between multiple languages. The
+Quick Start CRD `spec` contains a `language` field to specify what
```
jhadvig, line 160: So this goes with `Have separate ConsoleQuickStart resources for each language and use a label to indicate the language.` approach. Right ?

jhadvig (commented):
Looks good but but thinking that we should also introduce an i18n label that our CRs will carry. Would make fetching of of resources in a specific language more easy, both in CLI  and UI.
Thoughts ?

enhancements/console/quick-starts.md
spadgett, line 160: Would you hide any QuickStarts that are in English when another language is selected in the UI? It would mean any QuickStarts that haven't been translated yet would be hidden entirely. I think it's probably better to show the English version if it's all we have. But if you always show English QuickStarts, then you will have duplicates potentially when those are also translated. The UI has no way to know that they're the same QuickStarts in different languages.

We might either need all of the translations in one QuickStart resource or a property that says this QuickStart translates another English QuickStart so we can hide the English one.

spadgett, line 160: As @jhadvig suggests, we might make this a label so we can exclude other languages from the response using a label query. That's the advantage to having each translation being in a separate resource: you don't request QuickStarts for languages that you don't need.

enhancements/console/quick-starts.md
```diff
@@ -151,6 +154,23 @@ resource actions. In order for the user to see the Quick Start and
...
+language the Quick Start is in. OpenShift can display the correct
+Quick Starts for each language by comparing the value of the
+`language` field to the currently displayed language in OpenShift.
```
spadgett, line 163: What happens if a QuickStart specifies a language that the rest of the UI doesn't support, but is set as the user's language in the system or browser settings? Should we try to handle that case?

enhancements/console/quick-starts.md
spadgett: Another option is to have a console backend endpoint that filters by `spec.language` if we don't need to watch QuickStarts.

spadgett:
cc @andrewballantyne

enhancements/console/quick-starts.md
jhadvig: Good point, should console maybe have a list of supported languages?

enhancements/console/quick-starts.md
jhadvig: Thinking if the label would not be sufficient, since the label would be basically the same as the proposed field.
The only thing is that we will need to have it documented, thats the label is required, otherwise the QS won't be fetched.

rebeccaalpert: Yeah, I was thinking we'd hide them, so the duplicate issue wouldn't come up. But if you want to show English ones by default, we need an approach similar to what you mentioned.

spadgett: A missing label could be treated as the default language (English).

I don't think we can require that the label is set since it's effectively API breaking. Existing QuickStarts would disappear on upgrade.

enhancements/console/quick-starts.md
rebeccaalpert: With the approach outlined here, they would only see English Quick Starts. i18next.languages vs. supported languages will output 'en' if your browser is set to German, for example.

Is it something that third-party Quick Start contributors may want? How often do you think it would come up? Making the Quick Start language match OpenShift creates consistency, but you lose the ability to add pockets of different-language content. I can see the value of both approaches (consistent user experience vs. user experience augmented for your language).

christianvogt:
We reference quick starts by resource name as part of the "next quick start" and possibly other doc links. Now that the _same_ quick start can be represented by different languages, how does one reference the quick start now? Perhaps we need a naming convention where the language is appended to the resource name so that lookups split the name on a delimiter.

enhancements/console/quick-starts.md
christianvogt: I like the idea of using a label so that we can limit the results when fetching resources.

rebeccaalpert:
Updated the proposal doc to use the label approach and a language in the name of the Quick Start. I went with a tilda ("~") for the delimiter because it's what i18next uses as a delimiter in other contexts and won't cause problems with dash-separated names or language codes.

enhancements/console/quick-starts.md
```diff
@@ -96,15 +98,17 @@ QuickStarts CR for [Explore Serverless](https://marvelapp.com/236ge4ig/screen/69
 apiVersion: console.openshift.io/v1
 kind: QuickStart
 metadata:
-  name: explore-serverless
+  name: explore-serverless~en
+  labels:
+    language: en
```
spadgett, line 103: This needs to have a namespace prefix. I'd suggest

```suggestion
    console.openshift.io/lang: en
```

enhancements/console/quick-starts.md
```diff
@@ -151,6 +155,30 @@ resource actions. In order for the user to see the Quick Start and
 take its tour, his access review needs to pass all the listed
 resources actions.
 
+#### Quick Start Internationalization
+
+OpenShift can be toggled between multiple languages. The
+Quick Start CRD `label` contains a `language` field to specify what
```
spadgett, line 161:
```suggestion
Quick Start CRD contains an optional `console.openshift.io/lang` label to specify what
```

enhancements/console/quick-starts.md
```diff
@@ -151,6 +155,30 @@ resource actions. In order for the user to see the Quick Start and
...
+language the Quick Start is in. CRDs without this label will be treated
+as English CRDs. OpenShift can display the correct
```
spadgett, line 163: I'd say specifically "Quick Starts" here so it's clear this is specific to Quick Starts and not other CRDs.

```suggestion
language the Quick Start is in. Quick Starts without this label will be treated
as English Quick Starts. OpenShift can display the correct
```

enhancements/console/quick-starts.md
```diff
@@ -151,6 +155,30 @@ resource actions. In order for the user to see the Quick Start and
...
+Quick Starts for each language by comparing the value of the
+`language` label to the currently displayed language in OpenShift.
+(The display language is determined by `localStorage.getItem('bridge/language')`
```
spadgett, line 166: We might move this to the user settings config map. I would leave this detail out and instead link to the i18n enhancement for more information.

enhancements/console/quick-starts.md
```diff
@@ -151,6 +155,30 @@ resource actions. In order for the user to see the Quick Start and
...
+
+The language label must be a two-letter
+[ISO 639-2 Language Code](https://www.loc.gov/standards/iso639-2/php/code_list.php))
+(i.e. "en" for English) or an ISO language code hyphenated with an alpha-2
+[ISO-3166 Country Code](https://www.iso.org/iso-3166-country-codes.html)
+(i.e. "en-us").
+
+The language is also appended to the name of the CRD so we can avoid showing
+duplicates of the same CRD content in different languages. The name and language are
+separated by a tilda ("~") so we can split them on a consistent delimiter.
```
spadgett, line 180: `~` is not typically allowed in k8s names. It's also used as our internal delimiter for group/version/kind in console, so I'd suggest not using it here. I'd say just `my-quick-start-{lang}`. It would be good to show an example YAML metadata with the label and name.

enhancements/console/quick-starts.md
```diff
@@ -96,15 +98,17 @@ QuickStarts CR for [Explore Serverless](https://marvelapp.com/236ge4ig/screen/69
 apiVersion: console.openshift.io/v1
 kind: QuickStart
 metadata:
-  name: explore-serverless
+  name: explore-serverless~en
```
spadgett, line 101: I'd suggest just leaving the postfix off for English Quick Starts. Then we don't have to worry about renaming existing Quick Starts. (We should mention that explicitly below.)

rebeccaalpert:
Thanks; addressed feedback!

enhancements/console/quick-starts.md
```diff
@@ -151,6 +155,28 @@ resource actions. In order for the user to see the Quick Start and
...
+
+The language label must be a two-letter
+[ISO 639-2 Language Code](https://www.loc.gov/standards/iso639-2/php/code_list.php))
+(i.e. "en" for English) or an ISO language code hyphenated with an alpha-2
+[ISO-3166 Country Code](https://www.iso.org/iso-3166-country-codes.html)
+(i.e. "en-us").
+
+The language is also appended to the name of non-English Quick Starts so we can avoid showing
+duplicates of the same Quick Start content in different languages. The format is
+my-quick-start for English and my-quick-start-{language code} for other languages.
```
spadgett, line 178: Can you show an example YAML snippet with the metadata changes for another language here?

enhancements/console/quick-starts.md
```diff
@@ -151,6 +155,28 @@ resource actions. In order for the user to see the Quick Start and
...
+in non-supported languages become a need, we will need to modify this approach.
+
+The language label must be a two-letter
+[ISO 639-2 Language Code](https://www.loc.gov/standards/iso639-2/php/code_list.php))
+(i.e. "en" for English) or an ISO language code hyphenated with an alpha-2
+[ISO-3166 Country Code](https://www.iso.org/iso-3166-country-codes.html)
+(i.e. "en-us").
+
+The language is also appended to the name of non-English Quick Starts so we can avoid showing
+duplicates of the same Quick Start content in different languages. The format is
```
spadgett, line 177: I'd mention that if a Quick Start is not translated into the users' language, the English version will be shown.

rebeccaalpert:
All set!

enhancements/console/quick-starts.md
```diff
@@ -176,6 +176,17 @@ The language label must be a two-letter
...
 duplicates of the same Quick Start content in different languages. The format is
 my-quick-start for English and my-quick-start-{language code} for other languages.
+If a Quick Start is not translated into a user's language, the English version will be shown.
+
+Example Explore Serverless Quick Starts CR snippet for Japanese:
+```yaml
+apiVersion: console.openshift.io/v1
+kind: QuickStart
+metadata:
+  name: explore-serverless-{ja}
```
spadgett, line 186: Sorry, I meant to use `{lang}` as the replacement variable for `ja` (without the braces). Braces are not allowed in most resource names. This is why the YAML snippet really helps :)

```suggestion
  name: explore-serverless-ja
```

enhancements/console/quick-starts.md
christianvogt: I agree. But should we also be accepting of the `en` postfix?

enhancements/console/quick-starts.md
christianvogt: Note that if we grow our localization, the language code can be 2 part separated by `-`. Perhaps our name to language code separator should be a `.` instead?
`<name>.fr-FR`

enhancements/console/quick-starts.md
christianvogt: i guess at the end of the day the name only matters when we reference quick starts in other fields and try to resolve to the the correct quick start. We should look for both with and without the `en` postfix.

spadgett:
Putting this out as alternative. We leave the label as a recommendation to help us query, but we have a `translates` stanza in the quick start content:

```yaml
apiVersion: console.openshift.io/v1
kind: QuickStart
metadata:
  name: explore-serverless-ja
  labels:
    console.openshift.io/lang: ja
spec:
  ...
  translates:
    quickStart: explore-serverless
    lang: ja
```

The value from `translates` could be the authoritative value. Then we don't have problems trying to split on the name

If the label is there, great. We don't query it unnecessarily. If not, it's still simple to filter client side. We'll always query for quick starts with no label or lang=en anyway. This allows us to provide API doc and validation for the new properties as well.

@christianvogt ?

spadgett:
The `translates` stanza could be used for other types like `ConsoleLink` as well.

christianvogt:
While I'm not against such a stanza, I prefer a convention that can be followed for CRs without the need to alter the schema. By forcing a label on authors we can rely on querying by label and reduce over fetching.

If using labels I'd suggestion having 3 for the different labels. Having a `name` label could allow us to more easily query for all the same localized quick starts.
```
metadata:
  name: explorer-serverless.en-gb
  labels:
    console.openshift.io/lang: en
    console.openshift.io/country: gb
    console.openshift.io/name: explore-serverless
```

rebeccaalpert:
@spadgett - Are you ok with the approach above? Thanks!

rebeccaalpert:
@jhadvig and @spadgett - I updated this to the labels approach mentioned in scrum; let me know if you have feedback. Thanks!

christianvogt:
lgtm

We'll need to ensure that the documentation for the `nextQuickStart` is updated such that it is a reference to the quick start `name` label and no longer the resource name.

enhancements/console/quick-starts.md
```diff
@@ -151,6 +156,62 @@ resource actions. In order for the user to see the Quick Start and
...
+Quick Start. This should be consistent across all translations of a given Quick Start,
+as it allows us to avoid showing the same Quick Start content in multiple languages.
+If a Quick Start is not translated into a user's language, the English version will be shown.
+
+OpenShift can display the correct
+Quick Starts for each language by comparing the value of the
+`console.openshift.io/lang` and `console.openshift.io/country` labels to the currently displayed language in OpenShift.
+See the [internationalization enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/console/internationalization.md) for more information.
+This approach will not handle languages we don't offer support for; if Quick Starts
+in non-supported languages become a need, we will need to modify this approach.
```
jhadvig, line 190: What happens when a QS contains a language in the `console.openshift.io/lang` label that we don't support, eg. `cz` for czech? Is there a fallback? Won't those be used at all?
In this case I would suggest no to use, list, ... those Quick Start CRs. Just use those that are are supported, or don't have the label set, which means that they are english.

jhadvig (commented):
Just couple of nits.
REally like that we could bypass the need of an additional spec field or even CRD and make this happen with resources we have 👍  really nice work

enhancements/console/quick-starts.md
rebeccaalpert: If we use the existing console approach for i18n, Czech will fall back to English. (We'd be using the same language logic, so the behavior would match.)

We'd get the same ultimate outcome as just using supported languages/Quick Starts that don't have the label set. Happy with either approach.

rebeccaalpert:
Addressed PR feedback!

enhancements/console/quick-starts.md
jhadvig: Yeah, I would go with fallback to english in both cases 👍

jhadvig (approved):
LGTM. Please squash so we are ready for final tag 👍

spadgett (approved):
lgtm, please squash

rebeccaalpert:
Squashed!

spadgett (approved):
/lgtm

PR #657, opened 2021-02-16
Add managementState field to the DNS operator

rfredette:
Enhancement proposal to address https://issues.redhat.com/browse/NE-369

### DIFF
--- a/enhancements/dns/managementstate.md
+++ b/enhancements/dns/managementstate.md
@@ -43,27 +43,34 @@ doesn't re-enable the DNS operator.
 
 The DNS operator API should provide a `managementState` field, which will
 prevent the DNS operator from overwriting fields in managed deployments and
-daemonsets when the `managementState` field set to `Unmanaged` 
+daemonsets when the `managementState` field set to `Unmanaged`
 
 ## Motivation
 
 ### Goals
 
-Provide a method to disable the DNS operator without disabling the CVO
+Provide a method to disable the DNS operator without disabling the CVO.
 
 ### Non-Goals
 
-TODO
+- Do not support the `Force` option for `managementState` that is provided for
+  other operators. The `Force` state overrides other upgrade-blocking checks,
+  and can lead to unsuccessful upgrades.
+- Do not support the `Removed` option for `managementState` that is provided
+  for other operators. The `Removed` state means that the operator is actively
+  removing itself, which should not be done for DNS.
 
 ## Proposal
+
+Several other operators, including the apiserver, scheduler, storage, network,
+and console operators already implement the `managementState` field in order to
+disable the operator without needing to disabling the CVO. I propose that the
+DNS operator utilize the same field for consistency.
 
 The DNS operator spec shall include the `managementState` field. The field will
 accept two values:
 - `Managed`: The DNS operator will manage CoreDNS configuration.
 - `Unmanaged`: The DNS operator will not manage CoreDNS configuration.
-
-(TODO: is there anything the DNS operator will do regardless of the state of
-`managementState`?)
 
 ### User Stories
 
@@ -85,7 +92,10 @@ TODO
 
 ### Risks and Mitigations
 
-TODO
+By adding the `managementState` field, it becomes easier for cluster admins to
+put cluster DNS into an unsupported, potentially broken state. In order to
+mitigate this, it is important to prominently state that setting
+`managementState` to `Unmanaged` is inherently unsupported.
 
 ## Design Details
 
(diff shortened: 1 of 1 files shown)


enhancements/dns/managementstate.md
```diff
@@ -0,0 +1,126 @@
...
+
+Provide a method to disable the DNS operator without disabling the CVO
+
+### Non-Goals
+
+TODO
+
+## Proposal
+
+The DNS operator spec shall include the `managementState` field. The field will
```
sgreene570, line 60: By "DNS operator spec", do you mean the cluster DNS config API?

enhancements/dns/managementstate.md
```diff
@@ -0,0 +1,126 @@
...
+
+As a cluster admin, I have reported an issue with CoreDNS, but until a fix is
+released, I need a workaround. I set the DNS operator's `managementState` field
+to `Unmanaged`, then apply the workaround.
+
+### Implementation Details/Notes/Constraints [optional]
+
+TODO
+
+### Risks and Mitigations
```
sgreene570, line 86: The customer could potentially break cluster DNS services (and therefore workloads) when they set DNS to an unmanaged state, right?

enhancements/dns/managementstate.md
```diff
@@ -0,0 +1,126 @@
...
+  `Unmanaged`. Verify that the change is rejected.
+
+### Graduation Criteria
+
+TODO
+
+### Upgrade / Downgrade Strategy
+
+Upgrades will leave `managementState` unset, which should be treated as being
+in a `Managed` configuration.
```
sgreene570, line 111: Should we be blocking upgrades when a customer has the DNS config in an unmanaged state? Or do we want to force the config back to a managed state? How do other components handle this?

enhancements/dns/managementstate.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Motivation
+
+### Goals
+
+Provide a method to disable the DNS operator without disabling the CVO
+
+### Non-Goals
+
+TODO
```
candita, line 56: As non-goals, maybe explain why there are two excluded management states: `Force` and `Removed`

enhancements/dns/managementstate.md
```diff
@@ -0,0 +1,126 @@
...
+When diagnosing a DNS issue, sometimes it is helpful or even necessary to
+disable the DNS operator and patch the CoreDNS daemonset. Currently, this
+requires disabling the cluster version operator (CVO) as well so that the CVO
+doesn't re-enable the DNS operator.
+
+The DNS operator API should provide a `managementState` field, which will
+prevent the DNS operator from overwriting fields in managed deployments and
+daemonsets when the `managementState` field set to `Unmanaged` 
+
+## Motivation
```
candita, line 48: Motivation can include the other operators that employ the `managementState`

enhancements/dns/managementstate.md
```diff
@@ -0,0 +1,126 @@
...
+
+TODO
+
+## Proposal
+
+The DNS operator spec shall include the `managementState` field. The field will
+accept two values:
+- `Managed`: The DNS operator will manage CoreDNS configuration.
+- `Unmanaged`: The DNS operator will not manage CoreDNS configuration.
+
```
candita, line 64: It might be useful to have a `TimedUnmanaged`, which would keep the operator unmanaged for a specific window of time.  Then it would automatically revert at the end of the maintenance window.  The benefit would be the admin wouldn't overlook reverting the `managementState` and ending with a stale config.  The drawback is that it may not be of much value to users.

enhancements/dns/managementstate.md
rfredette: I'm not sure about the terminology, but the common method of including management state is in the operator config, as `spec.managementState`, like it is in this example for the API server operator: https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/README.md#debugging

enhancements/dns/managementstate.md
rfredette: I'm not sure how useful it would be to automatically revert to a `Managed` state, since most likely the reason that a user would set it to `Unmanaged` is that the `Managed` state is not working for whatever reason. There's probably an argument to be made that it would be useful to have as a backstop when debugging an issue, but if the operator is set to `Unmanaged` because there's a temporary patch to coredns, then it may need to stay `Unmanaged` until the fix is released, and it would be hard to pin a particular amount of time for that.

Miciah: The enhancement says that `managementState: Unmanaged` will cause the clusteroperator to report `Upgradeable=False`, and CVO will raise a warning alert if the clusteroperator continues to report `Upgradeable=False` for an hour or longer:  https://github.com/openshift/cluster-version-operator/blob/07598f405066871e3869df6a36a1eafba2598ee5/install/0000_90_cluster-version-operator_02_servicemonitor.yaml#L65-L72
That should suffice to keep the administrator from overlooking the situation.

PR #658, opened 2021-02-18
ovn-kubernetes-ipsec: Update EP with small feature enhancements for OCP 4.8

markdgray:
* Enhance OpenShift IPsec by adding support for enabling and disabling on a running cluster without having to enable it at install time.
* Enable dual stack IPv6 support for IPsec.

### DIFF
--- a/enhancements/network/ovn-kubernetes-ipsec.md
+++ b/enhancements/network/ovn-kubernetes-ipsec.md
@@ -68,12 +68,12 @@ an entire Kubernetes cluster that has been configured with OVN Kubernetes.
   - Assume that all nodes will be provisioned with certificates that have been
   signed by a OpenShift Container Platform internal certificate authority
   (CA).
-- This option will only be available at cluster installation time.
 - All components used to implement this must be FIPS-compliant cryptographic
 components.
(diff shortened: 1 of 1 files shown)


russellb (commented):
Is the additional 4.8 work already implemented?  If not, I guess it should be "implementable" instead of "implemented" ?

markdgray:
> Is the additional 4.8 work already implemented? If not, I guess it should be "implementable" instead of "implemented" ?

No, it isn't but I wasn't sure what to put as the original feature is implemented. I will update and fix the linting issue.

Is it correct to just update this enhancement rather than adding 2 new enhancements. These really aren't large changes.

russellb:
> > Is the additional 4.8 work already implemented? If not, I guess it should be "implementable" instead of "implemented" ?
> 
> No, it isn't but I wasn't sure what to put as the original feature is implemented. I will update and fix the linting issue.
> 
> Is it correct to just update this enhancement rather than adding 2 new enhancements. These really aren't large changes.

Yeah, updating this one is fine.  And maybe leaving it as "implemented" is fine.  4.8 isn't out anyway, and even if plans change, it would be far from the only doc slightly out of date!

so just fix the linting thing and we'll merge

PR #660, opened 2021-02-19
Flow control API enhancements, first draft.

alanconway:
Expecting discussion and changes.

https://issues.redhat.com/browse/LOG-1043

/cc @pportante
/cc @jcantrill
/cc @sichvoge
/assign @jcantrill

### DIFF
--- a/enhancements/cluster-logging/flow-control-api.md
+++ b/enhancements/cluster-logging/flow-control-api.md
@@ -1,0 +1,164 @@
+---
+title: flow-control-api
+authors:
+  - "@alanconway"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2021-02-05
+status: provisional
+see-also:
+replaces:
+superseded-by:
+---
+
+# Flow control API for Logging
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+If the log collector cannot keep up with the rate that logs are written, some logging data will be lost.
+Presently the logging admin has little control over when or how logs get lost.
+
+This proposal defines an API to let cluster administrators to limit logging rates, or ignore some logs entirely.
+Logs will still be lost if the collector cannot keep up, but administrators have more control over *what* is lost.
+
+Rate limits mean that:
+
+* The cost and volume of logging can be predicted more accurately in advance.
+* Noisy containers cannot produce unbounded log traffic and unfairly drown out other containers.
+* High-value logs can be preferred over low-value logs by assigning higher rate limits.
+* Log streams that are not needed can be ignored, which reduces the load on the logging infrastructure.
+
+The API allows rates and policies to be applied _per-container_ by namespaces or pod labels.
+New types of selector that are added in future should also be applicable.
+
+## Motivation
+
+### Goals
+
+### Non-Goals
+
+The following are *not* goals for this proposal but *may* be addressed in future proposals:
+
+- `block` policy: containers that exceed their rate limit are back-pressured by leaving their stdout/stderr unread until they are within the target rate.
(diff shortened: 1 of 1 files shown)


sichvoge:
@alanconway I don't think that we should put this into the ClusterLogForwarder API. My vote would be to do that either via a dedicated CRD or through the ClusterLogging CRD. My reason would be that the ClusterLogForwarder only focuses on defining "what logs should go where" not "governing" the ingestion. WDYT?

alanconway:
> @alanconway I don't think that we should put this into the ClusterLogForwarder API. My vote would be to do that either via a dedicated CRD or through the ClusterLogging CRD. My reason would be that the ClusterLogForwarder should only focus on one thing and that's "what logs should go where" not "governing" the ingestion, IMHO. WDYT?

+1 I debated it, you've convinced me. I originally thought of associating flow control with forwarder inputs, but on reflection I don't think that's correct. E.g. I want all the application logs to go to the same Loki instance, but I want the flow control rates to be set based on  per-namespace SLAs. Will update.

enhancements/cluster-logging/flow-control-api.md
```diff
@@ -0,0 +1,180 @@
...
+List the specific goals of the proposal. How will we know that this has succeeded?
+
+### Non-Goals - FIXME
+
+The following are not goals for this proposal but may to addressed by future proposals building on this one:
+
+- `block` policy: containers that exceed their rate limit are back-pressured by leaving their stdout/stderr unread until they are within the target rate.
+- combined rate limits: specify a total rate limit for a pipeline, enforced by dynamically adjusting rate limits on all containers contributing to the pipeline.
+
+## Proposal - FIXME
```
portante, line 61: Are these applied at the fluentd level?  Or at the conmon level?

portante (commented):
Certainly more details need to be added to the document as indicated by the various `FIXME` comments, but can we add what sub-systems will implement the policies and rates?

enhancements/cluster-logging/flow-control-api.md
alanconway: Probably a bit of both. The model I have in  my mind is that the collector is reading pipes or sockets from conmon rather than files from conmon and can push back by not reading. That means comon only needs to poll for write, and provide an option to use blocking pipes or sockets, it doesn't need complex config for controlling per-container rates & blocking - they can be imposed from outside.  I'm not sure that will be the way it ends up - there's thought needed.

alanconway: The doc is mainly intended to describe the API,  not the impl, but I agree at least some "possible impl" notes might help clarify

portante: I have been thinking that the only way to really do this well is in the kernel, which hopefully my presentation for container plumbers will be accepted where I can present that idea.

enhancements/cluster-logging/flow-control-api.md
```diff
@@ -0,0 +1,164 @@
...
+[LabelSelector]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#labelselector-v1-meta
+[quantity]: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/
+[memory]: (https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory)
+
+### User Stories
+
+#### Cluster admin wants to set global per-container default of 1024 bytes/second
+
+``` yaml
+- maxByteRate: 1024
```
sichvoge, line 86: When you do not configure a `policy`, is the default "drop"?

enhancements/cluster-logging/flow-control-api.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+#### Cluster admin wants to ignore logs from containers with certain labels
```
sichvoge, line 89: Is that the scenario where you could basically "exclude" collecting logs from certain systems or are you still collecting but you don't do anything?

enhancements/cluster-logging/flow-control-api.md
alanconway: Yes, same as existing behaviour, least distrubance to apps.

enhancements/cluster-logging/flow-control-api.md
alanconway: Exclude collecting, unless there's some compelling reason why we can't (don't know of any right now)

alanconway:
Moved to #1013

PR #661, opened 2021-02-20
New OpenshiftCatalogueValidator

camilamacedo86:
**Description**

See that our goal here is;

- We have OLM, and it has rules/criteria to work with operators in upstream(k8s) and (ocp) - We offer these standard options in ( operatorhub.io )
- Then, we are proposing here have a validator for OCP such as we have for operatorhub.io

### DIFF
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -16,7 +16,7 @@ config:
 
   # We like to use really long lines
   line-length:
-    line_length: 400
+    line_length: 800
 
   # Sometimes we repeat headings, and it's easier to just turn those
   # into emphasis
--- a/enhancements/operator-framework-api/catalogue-check.md
+++ b/enhancements/operator-framework-api/catalogue-check.md
@@ -1,0 +1,405 @@
+---
+title: openshift-validator
+authors:
+  - "@camilamacedo86"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2021-02-19
+last-updated: 2021-03-10
+status: implementable
+see-also:
+  - "/enhancements/olm/max-openshift-versions-for-operators.md"
+  - new checks for OperatorHubValidator:https://github.com/operator-framework/enhancements/pull/65
+---
+
+# OpenShiftValidator
+
+This proposal describes a new Validator to perform specific *linters* checks which are specifically to publish the operator bundle on the OpenShift Catalogue. The requirements are similar to the [OperatorHubValidator][operatorhubvalidator] implemented in [operator-framework/api][oper-api].
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [x] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Currently, [operator-framework/api][oper-api] centralizes the implementations of validator(s) & suite(s) of checks used by tools. For further information check [here](https://github.com/operator-framework/api/blob/master/pkg/validation/validation.go#L38-L46) the existing validators. Now see:
+
+- `BundleValidator` checks if the operator bundle spec is according to spec. For further information see [Operator Bundle](https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md) document.
+- `OperatorHubValidator` checks if the operator is respecting the required criteria to be published in the [OperatorHub.io](https://operatorhub.io/).
+
+The new `OpenShiftValidator` proposed here has the goal of perform checks which ensure the requirements to publish the operator bundle on the OpenShift Catalogue. The validator will raise issues or warnings accordingly. These checks will also be available to be used by the users via the tools such as SDK and consequently also called in the pipeline. Note, SDK provides the ability to optionally run validates using the `--select-optional` parameter to the bundle validate command.
+
+We can utilize this feature allowing the `OpenShiftValidator` to be called:
+
+```sh
+$ operator-sdk bundle validate ./bundle --select-optional name=openshift`
+```
+
+The new validator will allow verifying the operator bundle configuration works for a specific OCP version where it is intend to be published:
+
+```sh
+$ operator-sdk bundle validate ./bundle \
+    --select-optional name=openshift \
+    --optional-values="ocp=4.8"
+```
+
+## Motivation
+
+- Allow an operator developer or scheduled jobs a means to check required criteria for publishing an operator on OpenShift.
+- Validate the operator bundle manifests and notify users and maintainers with warning and errors about what is not compliant to the OpenShift criteria
+- Allow operator developers, CI jobs such as the pipeline/CVP, and admins of OLM catalogues audit and validate the bundles regarding the specific criteria for OpenShift versions, which are supported by the bundle or which the user is intend to distribute the operator.
+
+### Goals
+
+Perform static checks on the operator bundle manifests to avoid the scenarios where the operator bundle:
+
+- is published but will not be installable
+- is not prepared to work well on all OpenShift cluster versions and lacks any configuration to address its limitation
+- is using deprecated or no longer supported K8S APIs
+- is not following good practices and conventions
+
+**NOTE** These motivations are similar to [Add new checks to the operatorhub validator][ep-operatorhub-checks], however, for OpenShift Catalog.
+
+Also:
+- [operator-framework/api][oper-api] with new checks in `OperatorHubValidator`
+- SDK tool using [operator-framework/api][oper-api] and return errors or warnings regarding the new checks detailed in this proposal via the command `operator-sdk bundle validate ./bundle --select-optional name=operatorhub`
+- SDK tool providing new argument `--optional-values` that allows pass a string map.
+- Pipeline performing the new checks since it uses `operator-sdk bundle validate ./bundle --select-optional name=openshift`
+
+### Non-Goals
+
+- Add checks which are not specific to `OpenShift/OKD` and that could be part of `OperatorHubValidator`
+- Provide a way for the users check specific criteria or perform custom checks
+
+## Proposal
+
+### User Stories
+
+- As an operator developer, I'd like to use the SDK tool to validate my operator bundle before publishing on OpenShift Catalogue. Validating will allow me to quickly get its errors and warnings locally so that, I can provide the required fixes and improvements before trying to publish it.
+- As an operator framework maintainer, I'd like to see CI checking the operator bundle criteria when a user pushes a Pull Request to add the operator bundle on the OpenShift so that, we can easily ensure a better quality of the projects before they are published.  
+- As an operator developer or operator framework maintainer, I'd like to use the validator to check if my operator bundle will work on versions of OpenShift that I am intend to publish it to.
+- As an operator framework maintainer, I'd like to see the `OpenshiftValidator` in [operator-framework/api][oper-api] so that, I can easily keep it maintained since the rules and criteria are centralized and then, I can ensure that all projects and tools are using the same. In this way, if something change I only need to update one project instead of many of them.
+
+### Implementation Details/Notes/Constraints
+
+#### OpenShift version supportability features
+
+Before we describe the criteria for the validator we need to check some options available to specify the OpenShift and Kubernetes version(s) which the operator bundle will be supported by its configuration. Following a summary.
+
+**Annotation maxOpenShiftVersion**
+
+See [here](/enhancements/olm/max-openshift-versions-for-operators.md) a proposal for the operator bundles can use the following annotation:
+
+```yaml
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    # Prevent upgrades to OpenShift Version 4.9
+    operators.coreos.com/maxOpenShiftVersion: "4.8"
+```
+
+**OCP Labels in the image index**
+
+By using the label `LABEL com.redhat.openshift.versions`, it is possible to define range criteria of OpenShift catalogue(s) where the operator bundle should be published, see:
+
+```yaml
+FROM scratch
+
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=your-operator
+LABEL operators.operatorframework.io.bundle.channels.v1=your-channel
+LABEL operators.operatorframework.io.bundle.channel.default.v1=your-channel
+LABEL com.redhat.openshift.versions: v4.5-v4.7
+
+COPY manifests /manifests/
+COPY metadata /metadata/
+```
+
+The versioning scheme available in the above OpenShift versions label accepts:
+
+- Min version (e.g `v4.5`)
+- Range (e.g `v4.5-v4.7`)
+- A specific version (e.g `=v4.6`)
+
+**Examples**
+
+| Label | Description |
+| ------ | ----- |
+| `LABEL com.redhat.openshift.versions:v4.5` | It means valid for OCP `4.5` and later versions |
+| `LABEL com.redhat.openshift.versions:=v4.5` | Valid **ONLY** for `4.5`
+| `LABEL com.redhat.openshift.versions:v4.5-v4.7` | Valid **ONLY** for `4.5`, `4.6` and `4.7` |
+
+However, note that nothing prevents a customer from taking a `4.6` catalog and installing it on a `4.8` ocp cluster (they may or may not actually be able to install the operators that exist in that catalog, of course).
+
+**CSV spec for minimal k8s version**
+
+The spec `minKubeVersion` describe what is the minimal version of Kubernetes that is supported by the operator bundle, see:
+
+```yaml
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+...
+spec:
+  ... 
+  minKubeVersion: 1.11.0
+```
+
+The above features can be used alone and/or combined.
+
+#### Criteria valid which will not checked against to a specific OCP version
+
+The motivation for the following check is to ensure that operators authors understand that when the `maxOpenShiftVersion` and `minKubeVersion` is not defined it means allow to distribute the operator bundle for any OpenShift Catalog version available and then, tha the operator projects has no limitations regarding the supported OCP/Kubernetes versions which probably is not the common scenario.
+
+| Type | Description |
+| ----- | ----- |
+|warning|Check if we can found the `maxOpenShiftVersion` annotation or `minKubeVersion` spec in CSV. Then, if none of this options be filled raise a warning such as; `Your operator bundle is not specifying the OpenShift or Kubernetes version(s) which are supported by it. Ensure that it is installable in all possible provided options or configure it accordingly.`.
+
+The motivation for the following checks is only to ensure that the syntax informed is valid and avoid issues usually caused by a typo mistake.
+
+| Type | Description |
+|  ----- | ----- |
+| error| Check if `maxOpenShiftVersion` in CSV was informed then, if yes ensure that is valid. Return the message with the error(s) found.
+| error| Check if the label `com.redhat.openshift.versions` is used in the index image, if yes then ensure that is a valid range/syntax. Return the message with the error(s) found.
+
+The motivation for the following check is to ensure that versions informed are actually compatible which will not let the operator face problems to be installed or to be available for the desired criteria.
+
+| Type | Description |
+|  ----- | ----- |
+| error | Check if `maxOpenShiftVersion` and `minKubeVersion` in CSV as the `com.redhat.openshift.versions` in the index image are informed then, check if them are a valid combination. Return the error accordingly inform what option combination is not valid as its values. (E.g If the `maxOpenShiftVersion` is = `4.5` and the label range is `LABEL com.redhat.openshift.versions:v4.5-v4.7` it is invalid or if `minKubeVersion` == `1.22` and the `maxOpenShiftVersion` is == `4.5` then it is invalid.)
+
+#### Criteria valid which will be checked against to a specific OCP version
+
+For the checks which should be done accordingly for some specific version(s) it will use the OCP version which has been provided to the validator via flag or via the bundle operator configuration. For example, if we run:
+
+```sh
+ operator-sdk bundle validate ./bundle \
+     --select-optional name=openshift \
+     --optional-values="ocp=4.8"
+```
+
+Then, it means that the check will validate the operator bundle configuration accordingly to the OCP version informed where it is intend to be published to. In this case, the validator will consider the OpenShift version `=4.8`.
+
+If the `--optional-values="ocp=4.8"` not be used then, the validator will check the operator bundle configuration and do the checks considering the OCP version(s) which are supported by it. So, the validator will look for the values in the `maxOpenShiftVersion` and the `minKubeVersion` which are in the CSV and in the `com.redhat.openshift.versions` which is found in the index image.
+
+- To check if the operator bundle is using a CRD API version which is deprecated or not supported we will need to see if the CRD manifests in the bundle are using `apiVersion: apiextensions.k8s.io/v1beta1`. (See [here](https://github.com/operator-framework/operator-sdk/blob/v1.4.2/testdata/go/v2/memcached-operator/bundle/manifests/cache.example.com_memcacheds.yaml#L1)). Note that an operator bundle can have Many CRD(s).
+
+- To check if the operator bundle is using the Webhook API version which is deprecated we will need to see if the CSV has the spec `webhookdefinitions.<*>.v1beta1`. (e.g See [here](https://github.com/operator-framework/operator-sdk/blob/v1.4.2/testdata/go/v2/memcached-operator/bundle/manifests/memcached-operator.clusterserviceversion.yaml#L203-L205). Note that an operator bundle can have Many Webhook(s) and with more than one type and this field is a list of.
+
+Note that the output message should clarify and informed when the API was deprecate and when was or will be removed and what version should be used instead. Also, the respective OCP version should also be provided in the message.
+
+The error raised will be accordingly with the OCP version that we have, see:
+
+| Conditional |Type |
+| ------ | ----- |
+| operator bundle supports OpenShift version which is using k8s => `1.22` and do not uses `apiextensions/v1` ÒR has [webhookdefinitions](https://github.com/operator-framework/operator-sdk/blob/v1.4.2/testdata/go/v2/memcached-operator/bundle/manifests/memcached-operator.clusterserviceversion.yaml#L203-L205) and do not have the API `v1` listed | error |
+| operator bundle supports OpenShift version which is using k8s `=> 1.16 <= 1.22` and uses `apiextensions/v1beta1` ÒR `admissionregistration.k8s.io/v1beta1` | warning |
+
+**Why we need to check if the api v1 is not used instead of only check if it has any CRD or Webhook with the deprecated `v1beta1`?**
+		
+The operator author might provide a solution that works for both scenarios which means that will work when installed in the previous version of `Kubernetes < 1.16` and also in the upper versions `=> 1.22`. However, for the operator works well in any Kubernetes version `>=1.22`, we know that it is mandatory it has the latest APIs versions `v1`. See [here](https://github.com/kubernetes-sigs/kubebuilder/blob/v3.0.0-beta.0/test/e2e/v3/plugin_cluster_test.go#L99-L101), for example, that in the e2e test for Kubebuilder, we check what is the cluster version to apply and test the API versions accordingly. And then, note that we can find both APIs in the webhooks fields of the CSV as well such as:
+
+```yaml
+  webhookdefinitions:
+  - admissionReviewVersions:
+    - v1beta1
+    - v1   
+```
+
+**IMPORTANT**
+
+> Kubernetes version `1.22` was not released until now which means that the deprecated APIs still supported for all clusters. Also, at the same way none OpenShift version which is using Kubernetes API `1.22` was released. Then, we cannot raise errors until it happens for the scenarios where we the `maxOpenShiftVersion` and the `minKubeVersion` which are in the CSV and in the `com.redhat.openshift.versions` which is found in the index image are not provided because we are assuming all possible versions.
+> However, for the checks where the `--optional-values="ocp=4.X"` which will be using Kubernetes API `1.22` was used we know that the goal is to verify if the operator bundle will work òn `1.22` then, we can already raise the `error` result.
+
+Then, it means that:
+- `--optional-values="ocp=4.X"` flag with a value of an OpenShift version using K8S `=> 1.16 <= 1.22` the validator will return result as warning.
+- `--optional-values="ocp=4.X"` flag with a value OpenShift version using K8S `=> 1.22` the validator will return result as error.
+- `minKubeVersion >= 1.22` return the error result.
+- `maxOpenShiftVersion >= OCP version with 1.22` return the error result.
+- `com.redhat.openshift.versions` with a range which has an OCP version with 1.22` return the error result.
+
+**Currently (OCP versions released with the Kubernetes `<1.22`)**
+- `minKubeVersion` and `maxOpenShiftVersion` and `com.redhat.openshift.versions` empty/nil return the warning result.
+
+**After OCP version using the `1.22` Kubernetes release**
+- `minKubeVersion` and `maxOpenShiftVersion` and `com.redhat.openshift.versions` empty/nil return the error result.
+
+In a scenario where none version has been provided and the operator bundle has not been using the provided features to define the OpenShift/Kubernetes versions which are supported by it then, it means that the bundle operator should work well in all available possible versions, indeed `>= 1.22` after its release roll out of course. The only check above which will return an error is that one which ensure that the APIs used are supported, however, for this scenario the validator should return **warning** instead.
+
+Note that the OpenshiftValidator needs to be able to receive a string map and check if it contains any valid key for the specific check to address this requirement.
+
+In the code implementation, it should be only one check which will raise error(s) and warning(s) as a respective message according to the OCP/K8S version used. See that the same checks with its specific outputs has been proposed to the `OperatorHubValidator` in the EP [Add new checks to the operatorhub validator][ep-operatorhub-checks]. Therefore, the code could be done without duplicate the rules and instead of that, used for both validators where only type and message changes if possible.
+
+### Risks and Mitigations
+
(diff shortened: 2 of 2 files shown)


enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,180 @@
...
+```yaml
+FROM scratch
+
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=your-operator
+LABEL operators.operatorframework.io.bundle.channels.v1=your-channel
+LABEL operators.operatorframework.io.bundle.channel.default.v1=your-channel
+LABEL com.redhat.openshift.versions: v4.5-v4.7
```
bparees, line 118: this could be used to decide what validation checks to perform on the bundle(which ocp versions to validate against), but this is *not* a mechanism for an operator team to constrain what versions of OCP their operator can run on.  Nothing prevents a customer from taking a 4.6 catalog and installing it on a 4.8 ocp cluster (they may or may not actually be able to install the operators that exist in that catalog, of course)

So while it's true that we should not put an operator into the 4.6 catalog if it fails validation on 4.6, and it's true that the pipeline won't put an operator into the 4.6 catalog if this version label doesn't include 4.6, those things are not the same as making an explicit statement about what versions of OCP the operator runs on.  At best it's a hint.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,180 @@
...
+  minKubeVersion: 1.11.0
+```
+
+The above features can be used alone and/or combined.
+
+#### Criteria
+
+| Check | Type | Description |
+| ------ | ----- | ----- |
+|To ensure that the operator bundle specified the OCP versions which the project should be supported.|warning|Check if `maxOpenShiftVersion` in CSV **OR** the `com.redhat.openshift.versions` in the Dockerfile is used.
```
bparees, line 157: see above, i don't consider these two things equivalent.

checking for both is reasonable, but the error/warning/implications of not having each are different.

and i suspect lots of operators are not going to specify maximum ocp versions, nor an upper bound on their ocp catalog version.  So we'll need to think about exactly what we want the error/warning/guidance to be here.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,180 @@
...
+|To check if the operator bundle is using the Webhook API version which is no longer supported|error|Check if `maxOpenShiftVersion` in CSV **OR** the `com.redhat.openshift.versions` in the Dockerfile is used and has the OCP version which is using K8S API >= `1.22` then, check if the CSV has the spec `webhookdefinitions.admissionReviewVersions.v1beta1` . (See [here](https://github.com/operator-framework/operator-sdk/blob/v1.4.2/testdata/go/v2/memcached-operator/bundle/manifests/memcached-operator.clusterserviceversion.yaml#L203-L205). Note that in this cases users should be informed that the `admissionregistration.k8s.io/v1beta1` was deprecated in Kubernetes `1.16` and is removed in `1.22` and the respective OCP versions as well.
```
bparees, line 167: I had a slightly different model in mind.  Rather than internalizing the logic that says "which versions does this bundle claim to support, now validate the bundle against rules for that version" I was anticipating something more decoupled in which the user specifies which versions(rules) they want to validate against.  

There could reasonably be an auto/auto-detect mode that chooses which version/rules based on the bundle metadata, but i think it's useful to allow any bundle to be validated against any ocp version's rules and see what errors/warnings come out.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,180 @@
...
+
+- `BundleValidator` check if the operator bundle spec is accordingly to spec definition. For further information see [here](https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md).
+- `OperatorHubValidator` checks of the operator is respecting the required criteria to be published in the [OperatorHub.io](https://operatorhub.io/).
+
+The new `OpenShiftCatalogueValidator` proposed here has the purpose of perform checks which are specifically to publish the operator bundle on the OpenShift Catalogue. The validator will raise issues or warnings accordingly.
+
+This checks will also be available to be used by the users via the tools such as SDK and consequently also called in the pipeline. Note that SDK provides currently:
+
+```sh
+To validate a bundle with a superset of requirements for operatorhub.io specifically:
```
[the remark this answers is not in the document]
camilamacedo86: The superset are the kind of validators. See that if we run only `operator-sdk bundle validate` it will use only the BundleValidator implemented in the operator-framework/api. However, if we inform in the select-option other "superset" such as OperatorHubValidator then it will run its checks. I will try to clarify that before in the EP.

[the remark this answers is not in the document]
camilamacedo86: It is the SDK output the validators are called/describe as "superset". (it is the current tool output from the help) and then, see: 

- operator-sdk bundle validate -> calls the BundleValidator and ensure that the handle spec def are OK
- operator-sdk bundle validate ./bundle --select-optional name=operatorhub -> calls the OperatorHubValidator to check if the bundle is passing in the criteria to publish it in OperatorHub

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,180 @@
...
+   
+  $ operator-sdk bundle validate ./bundle --select-optional name=operatorhub
+``` 
+
+Then, the idea proposed here would allow the tool also provide:
+
+```sh
+To validate a bundle with a superset of requirements for OpenShift Catalogue specifically:
+   
+$ operator-sdk bundle validate ./bundle --select-optional name=openshiftcatalogue
```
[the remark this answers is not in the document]
camilamacedo86: > i think we need an openshift version, because the validation rules will vary based on the openshift version (e.g. whether or not you are allowed to use v1beta1 CRDs)

If the bundle has not minKubeVersion, or maxOCPVersion or a set up in the index `com.redhat.openshift.versions` then it means that the operator bundle should work in all versions. In this way, look that we are doing these checks accordingly with the configuration of the bundle. 

- and arguably we can leave "catalog" off the name. It's validating whether the bundle is acceptable for openshift clusters.

Agreed 👍 I think it is a good suggestion.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,180 @@
...
+``` 
+
+**NOTE** This validator and its usage are demanding to address a further requirement proposal to audit all bundles of a provided catalogue.
+
+## Motivation
+
+- Provide an operator bundle validator to raise issues and warnings regarding specific configurations and specifications which are only valid for `OpenShift/OKD` cluster types
```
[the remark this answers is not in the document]
camilamacedo86: My idea here is that the OpenshiftValidator will do any check to ensure any criteria which are specific for OCP vendor. Then, it can be related to the OCP versions supportability as any other scenario that we might have in the future. E.g a specific spec/annotation that is only required when the operator will be published in OCP.

However, if the bundle is configured to work only in OCP 4.5 we should consider that to make the checks as well. I will add your suggestion and try to clarify it better.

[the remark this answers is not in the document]
camilamacedo86: I changed for `Provide an operator bundle validator to raise issues and warnings regarding specific configurations and specifications which are only valid for `OpenShift/OKD` cluster types accordingly with the OCP versions supported by the bundle.`

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: I understand your motivation and I agree that it is a very reasonable requirement too. However, I think we need to address 2 scenarios:
- I want to check my bundle according to its configuration regards the versions supported
- I want to check my bundle according to its configuration regards against a specific OCP version that I intend to publish. 

I will try to address that here. Really thank you for your support and suggestion.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,180 @@
...
+```sh
+To validate a bundle with a superset of requirements for operatorhub.io specifically:
+   
+  $ operator-sdk bundle validate ./bundle --select-optional name=operatorhub
+``` 
+
+Then, the idea proposed here would allow the tool also provide:
+
+```sh
+To validate a bundle with a superset of requirements for OpenShift Catalogue specifically:
```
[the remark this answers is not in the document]
camilamacedo86: It is only the SDK output/description ttps://github.com/openshift/enhancements/pull/661#discussion_r579792135. Let me know if it answers your question.

enhancements/operator-framework-api/catalogue-check.md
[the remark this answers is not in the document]
camilamacedo86: addressed.

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: I improved that and tried to clarify its better. Let me know if it makes more sense now.

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: Note that the checks which use this option are: 
- To verify if the operator bundle version has not any restriction and is actually configured to be installed in ANY OCP/K8S version (return a warning with the goal to only notify the users and let them beware of)
- To ensure that the range values informed for this option are valid
- To check if the operator bundle is using an API version that is deprecated or not supported in the range/versions informed when it is used and the --version suggested by you to check the config for the intend catalogue version was not provided. 

Also, I supplement the summary with the info added here by you in this comment. So, I think it is fine now.

camilamacedo86:
HI @bparees, 

Really thank you for your time and inputs. I understand that all comments are addressed because of this I close them. However, feel free to re-open and/or raise new suggestions and questions.

enhancements/operator-framework-api/catalogue-check.md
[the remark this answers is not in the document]
bparees: ok i think i understand.  it might be clearer if this was described as `to perform validation against operatorhub.io requirements, in addition to the standard bundle validation:`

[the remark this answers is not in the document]
camilamacedo86: It has a full description that might clarify better. See: 

```
 $ operator-sdk bundle validate -h
The 'operator-sdk bundle validate' command can validate both content and format of an operator bundle
image or an operator bundle directory on-disk containing operator metadata and manifests. This command will exit
with an exit code of 1 if any validation errors arise, and 0 if only warnings arise or all validators pass.

A valid bundle is defined by the bundle spec (linked below), therefore the default validator ensures a bundle conforms to	
that spec. If you want to ensure that your bundle is valid for an optional superset of requirements such as to those
required to publish your operator on operatorhub.io, then you will need to run one or more supported optional validators.
Set '--list-optional' to list which optional validators are supported, and how they are grouped by label.
	
More information about operator bundles and metadata:
https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md

NOTE: if validating an image, the image must exist in a remote registry, not just locally.

Usage:
  operator-sdk bundle validate [flags]

Examples:
This example assumes you either have a *pullable* bundle image,
or something similar to the following operator bundle layout present locally:

  $ tree ./bundle
  ./bundle
  ├── manifests
  │   ├── cache.my.domain_memcacheds.yaml
  │   └── memcached-operator.clusterserviceversion.yaml
  └── metadata
      └── annotations.yaml

To validate a local bundle:

  $ operator-sdk bundle validate ./bundle

To build and validate a *pullable* bundle image:

  $ operator-sdk bundle validate <some-registry>/<operator-bundle-name>:<tag>

To list and run optional validators, which are specified by a label selector:

  $ operator-sdk bundle validate --list-optional
  NAME           LABELS                     DESCRIPTION
  operatorhub    name=operatorhub           OperatorHub.io metadata validation
                 suite=operatorframework

To validate a bundle with a superset of requirements for Operator Framework:
	
  $ operator-sdk bundle validate ./bundle --select-optional suite=operatorframework

To validate a bundle with a superset of requirements for operatorhub.io specifically:
	
  $ operator-sdk bundle validate ./bundle --select-optional name=operatorhub


Flags:
  -h, --help                     help for validate
  -b, --image-builder string     Tool to pull and unpack bundle images. Only used when validating a bundle image. One of: [docker, podman, none] (default "docker")
      --list-optional            List all optional validators available. When set, no validators will be run
      --select-optional string   Label selector to select optional validators to run. Run this command with '--list-optional' to list available optional validators

Global Flags:
      --plugins strings          plugin keys of the plugin to initialize the project with
      --project-version string   project version
      --verbose                  Enable verbose logging
```

However, I think that the best approach would be to raise an issue against the SDK repo with we decide to improve it. I will try to add a better context to the EP as well. Please, feel free to ping me for we see how we can make it better in SDK. I will happy in try to work on that.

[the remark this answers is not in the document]
bparees: yes, thanks for showing where the term was coming from.  I agree that the SDK help text could be improved, along the lines i suggested.  I don't think `superset` is being used correctly here.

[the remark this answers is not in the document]
camilamacedo86: Done: https://github.com/operator-framework/operator-sdk/issues/4577

kevinrizza:
High level concern: I'm concerned about whether this stuff should really land in the upstream operator-framework project. Are we attempting to explicitly tie the operator-sdk with downstream OCP? There is a downstream version of the sdk, should these bits just land there?

njhale (commented):
In general, I think having a set of validations that OpenShift cares about is a good idea, but I do have some reservations about the specifics of the design in this proposal. Mainly, as a maintainer of operator-framework, I **would not** want to maintain OpenShift-specific validators in an operator-framework repository. Our intention from the beginning was for validations to be pluggable so that folks could define and layer new validations that they care about. Currently, this amounts to defining an implementation of the [validator interface](https://github.com/operator-framework/api/blob/master/pkg/validation/interfaces/validator.go#L8) and [adding it to the set of validators to be run](https://github.com/operator-framework/api/blob/master/cmd/operator-verify/manifests/cmd.go#L51). In the future, I was hoping we could offload validation entirely to something like [Cue](https://cuelang.org/docs/about/), so that we don't incur the overhead of maintaining custom tooling (IMO, this also keeps us from being brittle).

camilamacedo86:
Hi @kevinrizza and @njhale, 

I try to clarify the points discussed. And then, I will work on the follow-up proposals to describe that. Let me know if you think that has anything else that we would need to cover here.

camilamacedo86:
/hold 

I have been thinking about a better approach for we address this need.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,344 @@
...
+
+In the code implementation, it should be only one check which will raise error(s) and warning(s) as a respective message according to the OCP/K8S version used. See that the same checks with its specific outputs has been proposed to the `OperatorHubValidator` in the EP [Add new checks to the operatorhub validator][ep-operatorhub-checks]. Therefore, the code could be done without duplicate the rules and instead of that, used for both validators where only type and message changes if possible.
+
+### Risks and Mitigations
+
+- Same risks and/or mitigations described in [Add new checks to the operatorhub validator][ep-operatorhub-checks]
+
+## Design Details
+
+### Open Question
```
camilamacedo86, line 254: Hi @kevinrizza @njhale I tried to updated here based on the discussions and I added all these open questions to let us know clarifies our thought an take the decision. I hope that helps. If you have time would be great get your inputs as well.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,344 @@
...
+
+d) Have we exactly the same risks/concerns regards address specific features for OpenShift in [operator-framework/api][oper-api] that we have in OLM? Since it is an API would not actually for this case easier the process to decouple what is specific for OpenShift if/when that starts to be required in the future?
+
+3. How much value allow users develop their own validators which only can be used by SDK feature bundle validation can bring to project and OF? Have we some example scenario(s) or requirements that would justify this effort? 
+
+4. Have SDK any concern about show the specific validator for OpenShift as one option of the bundle validation feature? If yes, what are the concerns and reasons? Is SDK also planning have an [operator-lib](https://github.com/operator-framework/operator-lib) API specific for downstream to address the OpenShift specific helpers and needs that we know about (e.g see [here](https://github.com/operator-framework/operator-sdk/issues/2770)?
```
camilamacedo86, line 274: @jmrodri this open question you might able to help us. WDYT?

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Design Details
+
+### Open Question
+
+1. Will SDK be affected by this solution?
+
+The proposal describes the new Validator ideally implemented in [operator-framework/api][oper-api] which means that SDK would only need to provide its usage.
+
+However, see the [Alternative][#alternative] section. Some options could result in a high effort for SDK.
```
camilamacedo86, line 260: @jmrodri this one I think needs came to your attention.

camilamacedo86:
Hi @njhale, 

Really thank you for your input. I have some questions I hope that you can help me with. 

> Our intention from the beginning was for validations to be pluggable so that folks could define and layer new validations that they care about. Currently, this amounts to defining an implementation of the validator interface and adding it to the set of validators to be run. In the future, I was hoping we could offload validation entirely to something like Cue, so that we don't incur the overhead of maintaining custom tooling (IMO, this also keeps us from being brittle).

I understand that besides increasing the maintainability by centralizing the code in operator-framework/api which is common  for all projects another motivation to do that is to avoid the scenario where for example: 
- User uses the bundle validation and all worked 
- Then, the user faces issues when pushing the operator to b published in operatorrHub io because the rule/code implementation is not centralized and not the same. 

I mean, by following your suggestion when something change we will need to remember to change all possible "clu" implementations which can be in all parts of the org instead of just one place.

- So, what is the advantage of we have the API but do not have the rules and criteria of the validators that should be used for any solution on it?
- How can we ensure that all projects are following the same criteria when we have a "clu" config for each one instead of centralizing it in the api? 
- How much is the effort to keep the validators maintained without centralizing the code?

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,344 @@
...
+|To ensure that the index image label `com.redhat.openshift.versions` value is valid|error|Check if the label `com.redhat.openshift.versions` is used in the index image, if yes then ensure that is a valid range/syntax. Return the message with the error(s) found.
+
+The motivation for the following check is to ensure that versions informed are actually compatible and the operator will not face problems to be installed or to be available for the desired criteria.
+
+| Check | Type | Description |
+| ------ | ----- | ----- |
+|To ensure the OpenShift and Kubernetes options to define the versions which are supported by the operator bundle in use are compatible|error|Check if `maxOpenShiftVersion` and `minKubeVersion` in CSV as the `com.redhat.openshift.versions` in the index image are informed then, check if them are a valid combination. Return the error accordingly inform what option combination is not valid as its values. (E.g If the `maxOpenShiftVersion` is = `4.5` and the label range is `LABEL com.redhat.openshift.versions:v4.5-v4.7` it is invalid or if `minKubeVersion` == `1.22` and the `maxOpenShiftVersion` is == `4.5` then it is invalid.)
```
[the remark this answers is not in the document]
jmrodri: These checks seemed wordy especially since you are describing them in the Description field.

jmrodri (changes requested):
A few changes. I will continue to review the second half soon.

.markdownlint-cli2.yaml
```diff
@@ -16,7 +16,7 @@ config:
 
   # We like to use really long lines
   line-length:
-    line_length: 400
+    line_length: 800
```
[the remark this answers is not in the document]
camilamacedo86: to allow us to do a table with bigger content. See its comment ` # We like to use really long lines`. however, it was not long enough :-P

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,344 @@
...
+
+- Add checks which are not specific to `OpenShift/OKD` and that could be part of `OperatorHubValidator` to this check.
+
+## Proposal
+
+### User Stories
+
+- As an operator developer, I'd like to use the SDK tool to validate my operator bundle, which will be published on OCP Catalogue so that, I could quickly get its errors and warnings locally to be able to provide the required fixes and improvements before try to publish it.
+- As a member of the Pipeline team, I'd like to use the SDK tool to validate the bundles so that, I could easily avoid issues in the projects published on the catalogue.
+- As an operator framework maintainer, I'd like to see the `OpenShiftValidator` in [operator-framework/api][oper-api] so that, I can easily keep it maintained and use it in any project and tool which please me.
```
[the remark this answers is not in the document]
camilamacedo86: changed. Tks for the feedback.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,362 @@
...
+
+The above features can be used alone and/or combined.
+
+#### Criteria valid which will not checked against to a specific OCP version
+
+The motivation for the following check is to ensure that operators authors understand that when the `maxOpenShiftVersion` and `minKubeVersion` is not defined it means allow to distribute the operator bundle for any OpenShift Catalog version available and then, tha the operator projects has no limitations regarding the supported OCP/Kubernetes versions which probably is not the common scenario.
+
+| Type | Description |
+| ----- | ----- |
+|warning|Check if we can found the `maxOpenShiftVersion` annotation or `minKubeVersion` spec in CSV. Then, if none of this options be filled raise a warning such as; `Your operator bundle is not specifying the OpenShift or Kubernetes version(s) which are supported by it. Ensure that it is installable in all possible provided options or configure it accordingly.`.
```
camilamacedo86, line 156: @bparees, your concern raised in https://github.com/openshift/enhancements/pull/661#discussion_r579720717 is addressed here.

.markdownlint-cli2.yaml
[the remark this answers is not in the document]
dhellmann: Please do not change this value. 400 is already too long. I set it that way so that when I added the linter job I did not have to wrap every single line of every file in the repository, but we don't actually want it that long. Please wrap paragraphs to a reasonable width to make reviewing easier.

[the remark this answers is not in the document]
camilamacedo86: @dhellmann, @jmrodri, 

A markdown table breaks the lines automatically and makes it easier for us to view the data.
However, the lint here is static and does not ignore tables. 
Also, I do not sure what is the motivation of this check. line_length: 400 requires to scroll such ass 500 such as 800. so, why it needs to be 400 and can't be 800?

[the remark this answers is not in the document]
dhellmann: I chose 400 as the smallest value that allowed me to turn on the linter job as required without editing every file in the repository first. 80 or 100 would be my preference, but I haven't wanted to come back and edit all of those files to make that limit work. Raising the value to 800 makes it completely useless as a check and encourages authors to write documentation in a format that is hard to review, hard to track edits, etc.

It looks like the tables in this document could easily be turned into bullet lists and retain the semantic value of special formatting beyond regular paragraphs.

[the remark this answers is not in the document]
bparees: linter rules that prevent people from using reasonable markdown structures seems like the tail wagging the dog.

Tables have headers and columns.  Bullet lists are not the same thing.

jmrodri:
> In general, I think having a set of validations that OpenShift cares about is a good idea, but I do have some reservations about the specifics of the design in this proposal. Mainly, as a maintainer of operator-framework, I **would not** want to maintain OpenShift-specific validators in an operator-framework repository. Our intention from the beginning was for validations to be pluggable so that folks could define and layer new validations that they care about. Currently, this amounts to defining an implementation of the [validator interface](https://github.com/operator-framework/api/blob/master/pkg/validation/interfaces/validator.go#L8) and [adding it to the set of validators to be run](https://github.com/operator-framework/api/blob/master/cmd/operator-verify/manifests/cmd.go#L51). In the future, I was hoping we could offload validation entirely to something like [Cue](https://cuelang.org/docs/about/), so that we don't incur the overhead of maintaining custom tooling (IMO, this also keeps us from being brittle).

From @kevinrizza and @njhale concerns, I think we might need to do what Nick suggested is take this opportunity to figure out how to create pluggable "rules" that can be pulled in dynamically and be used by operator-sdk. That way if someone wants to use an upstream operator-sdk with an openshift specific validation they can without our team having to bake it into the upstream. 

So I see a couple things coming out of this EP:
1. define the validation rules needed for an "openshift validator"  but not using the current implementation just the "rules"
2. probably need another EP to design a pluggable or more dynamic validation engine. it **can not** use the kubebuilder style plugins because those are compiled in. 

Maybe something that knows how to git pull validation files from a git repo. that would allow pulling rules data files without having to rebuild operator-sdk and they can be used upstream or downstream.

```
operator-sdk bundle validate --optional-validator github.com/myorg/my-custom-validator
```

camilamacedo86:
Hi @jmrodri and @njhale, 

Thank you for the collab. Regarding the comment, https://github.com/openshift/enhancements/pull/661#issuecomment-796393718, the problem with that solution is that it also not address all concerns. See that our goal here is;

- We have OLM, and it has rules/criteria to work with operators in upstream(k8s) and (ocp) - We offer these standard options in ( operatorhub.io )
- Then, we are proposing here have a validator for OCP such as we have for operatrhub.io 

However, if we follow your suggestion, we ( maintainers ) will not get the standard/rules criteria from OLM upstream/downstream, and we also will need to use SDK as the trust font thrust for OLMs rules.  So, we cannot write any solution that will be able to use that and not have the SDK bin as a dependency. So, IMHO I do not think that it would be the best approach because:

> I am as OF maintainer would like to be able to have an api with the common rules/logic/criteria for OLM so that I can use them in any solution that pleases me without the concern that they will not be == what is used by OLM. 

Then, I still think that the best approach that solves all concerns and is a good case for us starts to think about how we will centralize the specific downstream criteria/logic/rules used by olm to have a downstream api. see that SDK still able to choose if it will show or not thee openshift rules and where. Also, the problem of how we will have diff criteria/rules/logic valid for OLM upstream and downstream shows to be a more valuable problem than solve how people can use SDK CLI to create their validators. We already solved this problem by providing Scorecard.

So, ihmo the best approach and most valuable regarding use case show to use operator-framework/api in downstream. However, a really think is nice we check all possible ideas too.

.markdownlint-cli2.yaml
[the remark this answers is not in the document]
jmrodri: Is there a way to flag some lines as "ignore"? Then we could flag the tables as ignore.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,377 @@
...
+
+**Pros**
+
+- Mainly all pros of option A
+- Solves the OLM team concern about add OpenShift checks in [operator-framework/api][oper-api]
+  
+**Cons**
+
+- Effort required to keep a downstream repository for a downstream component and its releases [operator-framework/api][oper-api]
+- SDK would need to probably use only the downstream import to avoid misleading
```
jmrodri, line 338: SDK upstream would NOT be using a downstream o-f/api. So the con is that anyone that wants to target openshift *MUST* use a downstream SDK. That's actually pretty bad because today it honestly doesn't matter which one you use.

If we must do Option B, then there *MUST* be a mechanism to dynamically load these downstream rules into an upstream at runtime.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,377 @@
...
+The proposal above was defined considering the new Validator be implemented in [operator-framework/api][oper-api] and used by SDK upstream implementation.
+
+**Pros**
+
+- Lower effort/cost since it would be similar to the others as it usage
+- Allow the Validator be used with OKD
+- The Validator can be used by any solution
+- Does not requires the usage/installation of a downstream SDK binary
+
+**NOTE** SDK users and Pipeline uses the upstream binary and shows that we have not a release process for SDK binary from downstream. Just for the images which are useful for Ansible/Helm operators.
```
jmrodri, line 309: That is not true. We release a downstream SDK binary now as of 4.7. 

The binary is hosted on the mirror: https://mirror.openshift.com/pub/openshift-v4/clients/operator-sdk/latest/
We are also building an SDK image, albeit solely for giving access to the binary. https://github.com/openshift/ocp-release-operator-sdk/tree/master/release/sdk

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,377 @@
...
+  
+**Cons**
+
+- Effort required to keep a downstream repository for a downstream component and its releases [operator-framework/api][oper-api]
+- SDK would need to probably use only the downstream import to avoid misleading
+- It could cause delays for the SDK be able to apply the required changes of [operator-framework/api][oper-api] which can result ,for example, in an impact for Pipeline and users to get bug fixes or improvements.
+
+### (Option C) Design and implement Pluggable Validator mechanism for SDK
+
+Instead of we add the OpenshiftValidator to [operator-framework/api][oper-api] we would only provide it via the custom "downstream plugin" and make it available only for SDK binary.
```
jmrodri, line 343:
```suggestion
Instead of adding the OpenshiftValidator to [operator-framework/api][oper-api], we would provide it via the custom "downstream plugin" and make it available only for SDK binary.
```

jmrodri: I don't think it would be "only for SDK binary"

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,377 @@
...
+
+**Pros**
+
+- Address an RFE which allow others create their own validators and use them with SDK.
+- Does not affect the users and Pipeline since they will still able to use the upstream binary and the downstream validator module
+- Address the OLM team concerns about add the validator in the upstream api. Also, address similar concerns, if they exist, for SDK CLI in upstream.
+
+**Cons**
+
+- The Validator cannot be used outside of SDK. So, for example, it cannot be used by [audit proposal][audit-ep] if we decide to not make this solution leverage in SDK CLI (binary).
```
jmrodri, line 353: That isn't true. Why couldn't you reuse the same dynamic loader in another tool like audit? Having dynamic mechanism in SDK does not require it to only be in SDK.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,377 @@
...
+- Brings higher effort to implement since requires the design and implementation of a new feature for SDK which we have none other use case/request to use that so far which justify it.
```
jmrodri, line 354: While there are not specific examples aside from the existing on in this EP, having a flexible mechanism would be a good item since it could be used not only by SDK but whatever tool can import these validators. And all checks can be transformed into these dynamic checks and removed from o-f/api entirely

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,377 @@
...
+- Shows that will be harder for us to keep it maintained and/or receives contributions since it will leave only in [ocp-release-operator-sdk][ocp-release-operator-sdk].
```
jmrodri, line 355: That's also not the case. If this is a dynamic validation system, the system could live in its own repo and USED by both upstream and downstream SDK. Or the system can live solely in the upstream SDK repo.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,377 @@
...
+- We will be unable to promote the feature for the users since it will not available by default.
```
jmrodri, line 356: Not a con, because if it is dynamic then that's a great selling point.

enhancements/operator-framework-api/catalogue-check.md
```diff
@@ -0,0 +1,377 @@
...
+- It does not allow the re-usability of the validator via importing the code implementation to the other projects as the option A or B.
```
jmrodri, line 357: That's because it will be in a dynamic language of some sort that is reusable by loading it and parsing/running.

jmrodri:
So the way I envision this dynamic validator thingy is that all validations live in git repos. Either a single repo with a collection of validations or a one repo per validation, or a combination of both. This means that the validations will no longer be Go code. They will be something like CUE, yaml, json, whatever it needs to be like was suggested by others. 

Then in the SDK (or in other places to). there's a system that knows how to load validations from a repo. We load those validations, parse and run them. The beauty of this is that you do not have to embed these validations into the binaries. So upstream vs downstream goes away entirely. It is super easy to add new validations, you could standup a new repo with your validation in it then point the tools to use it.

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: I just discovered that with you because of another case. I need to update it here. Thank you.

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: How another place would use the SDK Validator plugin? The solution suggested, SDK would define an interface for the CLI to be able to use the validators. Have you another option in mind?

camilamacedo86: I think I saw here https://github.com/openshift/enhancements/pull/661#issuecomment-797588947. 👍 I think it is another option indeed.

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: solved.

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: If the implementation is specified in an SDK plugin available only in the SDK downstream repo as suggested then, it cannot be used without SDK or imported by other projects.

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: See; https://github.com/openshift/enhancements/pull/661#discussion_r598284681

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: It is not the idea that you are suggesting bellow. See; https://github.com/openshift/enhancements/pull/661#discussion_r598284681

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: The ep was clarified. However, not that the idea suggested is not the same added by your here. Your idea/suggestion was added as another alternative solution. See; https://github.com/openshift/enhancements/pull/661#discussion_r598284681

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: The ep was clarified. However, not that the idea suggested is not the same added by your here. Your idea/suggestion was added as another alternative solution. See; https://github.com/openshift/enhancements/pull/661#discussion_r598284681

enhancements/operator-framework-api/catalogue-check.md
camilamacedo86: Could we not import both (downstream and upstream) APIS as suggested here in the SDK repo?

PR #663, opened 2021-02-23
dns: Add configurable-dns-pod-placement enhancement

Miciah:
This enhancement enables cluster administrators to configure the placement of ￼the CoreDNS Pods that provide cluster DNS service.

### DIFF
--- a/enhancements/dns/configurable-dns-pod-placement.md
+++ b/enhancements/dns/configurable-dns-pod-placement.md
@@ -1,0 +1,359 @@
+---
+title: configurable-dns-pod-placement
+authors:
+  - "@Miciah"
+reviewers:
+  - "@candita"
+  - "@danehans"
+  - "@frobware"
+  - "@knobunc"
+  - "@miheer"
+  - "@rfredette"
+  - "@sgreene570"
+approvers:
+  - "@danehans"
+  - "@frobware"
+  - "@knobunc"
+creation-date: 2021-02-23
+last-updated: 2021-03-29
+status: implementable
+see-also: 
+replaces:
+superseded-by:
+---
+
+# Configurable DNS Pod Placement
+
+This enhancement enables cluster administrators to configure the placement of
+the CoreDNS Pods that provide cluster DNS service.
+
+## Release Signoff Checklist
+
+- [X] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [X] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The DNS operator in OpenShift 4.7 and prior versions manages a DaemonSet that
+serves two functions: running CoreDNS and managing node hosts' `/etc/hosts`
+files.  This enhancement, in OpenShift 4.8, replaces this single DaemonSet with
+two: one DaemonSet for CoreDNS and one DaemonSet for managing `/etc/hosts`.
+Additionally, this enhancement adds an API to enable cluster administrators to
+configure the placement of the CoreDNS Pods.
+
+## Motivation
+
+OpenShift 4.7 uses a single DaemonSet for both CoreDNS and for managing node
+hosts' `/etc/hosts` files.  Specifically, this DaemonSet has a container that
+adds an entry for the cluster image registry to `/etc/hosts` to enable the
+container runtime (which does not use the cluster DNS service) to resolve and
+thus pull from the cluster image registry.
+
+Because `/etc/hosts` needs to be managed on every node host, this DaemonSet must
+run on every node host.  Moreover, management of `/etc/hosts` is a critical
+service because the node host may fail to pull images (including those of core
+components) unless `/etc/hosts` has an entry for the cluster image registry.
+Consequently the DaemonSet has a toleration for all taints so that the DNS Pod
+always runs on all nodes.
+
+Some cluster administrators require the ability to configure DNS not to run on
+certain nodes.  For example, security policies may prohibit communication
+between certain pairs of nodes; a DNS query from an arbitrary Pod on some node A
+to the DNS Pod on some other node B might fail if some security policy prohibits
+communication between node A and node B.
+
+Splitting CoreDNS and management of `/etc/hosts` into separate DaemonSets makes
+it possible to remove the blanket toleration for all taints from the CoreDNS
+DaemonSet while keeping the blanket toleration on the DaemonSet that manages
+`/etc/hosts`.  Splitting the DaemonSet also makes it possible to enable use of a
+custom node selector on the CoreDNS DaemonSet.
+
+Another advantage of using separates DaemonSets is that the DaemonSet that
+updates `/etc/hosts` can use the host network to avoid consuming SR-IOV devices
+on nodes that have Smart NICs.
+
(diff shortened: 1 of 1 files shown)


enhancements/dns/configurable-dns-pod-placement.md
```diff
@@ -0,0 +1,306 @@
...
+with type `DNSNodePlacement` to `DNSSpec`:
+
+```go
+// DNSSpec is the specification of the desired behavior of the DNS.
+type DNSSpec struct {
+    // ...
+	
+	// nodePlacement enables explicit control over the scheduling of DNS pods.
+	//
+	// If unset, defaults are used. See nodePlacement for more details.
```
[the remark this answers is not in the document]
Miciah: I think `nodePlacement` is right; the godoc should reference the field by its name, which is `nodePlacement`.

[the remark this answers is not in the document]
sgreene570: Would it make sense to say "refer to x for more info on x" though?
Maybe i'm overthinking this. https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L134 uses the type rather than field name.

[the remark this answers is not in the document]
sgreene570: Either way works for me. Not worth bike shedding over this.

enhancements/dns/configurable-dns-pod-placement.md
```diff
@@ -0,0 +1,306 @@
...
+        node-role.kubernetes.io/infra: ""
+```
+
+### Validation
+
+Omitting `spec.nodePlacement` or its subfields specifies the default behavior.
+
+The API validates that `spec.nodePlacement.nodeSelector`, if specified, is a
+valid node selector and that `spec.nodePlacement.tolerations`, if specified, is
+a list of valid tolerations.
```
sgreene570, line 167: Should the operator validate that `spec.nodePlacement`, if provided

enhancements/dns/configurable-dns-pod-placement.md
```diff
@@ -0,0 +1,306 @@
...
+	//
+	// +optional
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+}
+```
+
+By default, DNS Pods run on untainted Linux nodes.  The `NodePlacement` field
+enables cluster administrators to specify alternative parameters.  For example,
+the following DNS specifies that DNS Pods should run only on "infra" nodes
+(i.e., nodes that have the "node-role.kubernetes.io/infra" label):
```
sgreene570, line 147: Might be worth mentioning that this example assumes infra nodes have no infra node taints.

enhancements/dns/configurable-dns-pod-placement.md
```diff
@@ -0,0 +1,306 @@
...
+1. Enable cluster administrators to control the placement of the DaemonSet that manages `/etc/hosts`.
+2. Enforce security policies.
+
+## Proposal
+
+This enhancement has two distinct parts.  First, the DNS operator, which manages
+the "dns-default" DaemonSet, is modified to manage an additional "node-resolver"
+DaemonSet, and the "dns-node-resolver" container, which manages `/etc/hosts`, is
+moved from the "dns-default" DaemonSet to a new "node-resolver" DaemonSet.  As
+part of this change, the toleration for all taints is removed from the
```
sgreene570, line 90: So, with this enhancement in place, where will dns-default pods run by default? Just on worker nodes?

enhancements/dns/configurable-dns-pod-placement.md
Miciah: Thanks!  I amended the example to include a toleration for any "node-role.kubernetes.io/infra" `NoSchedule` taint.  I don't know how prevalent the use of this taint is, but I gather that it is something that cluster administrators who configure infra nodes often use, and adding the toleration shouldn't cause any problems if the taint doesn't exist, so the example might as well include it.

enhancements/dns/configurable-dns-pod-placement.md
```diff
@@ -0,0 +1,359 @@
...
+
+Omitting `spec.nodePlacement` or its subfields specifies the default behavior.
+
+The API validates that `spec.nodePlacement.nodeSelector`, if specified, is a
+valid node selector and that `spec.nodePlacement.tolerations`, if specified, is
+a list of valid tolerations.
+
+### User Stories
+
+#### As a cluster administrator, I must comply with a security policy that prohibits communication among worker nodes
```
smarterclayton, line 199: Why is solving this use case with explicit placement more appropriate than using service topology?  This is a pretty specific constraint (and doesn’t cover other node to node traffic like monitoring, sdn, control plane traffic, etc).  While constraining DNS to nodes is reasonable on its own right for administrator topology choices, this specific example is fairly vague / not particularly complete.

enhancements/dns/configurable-dns-pod-placement.md
```diff
@@ -0,0 +1,359 @@
...
+"dns-node-resolver" is configured to tolerate all taints (as the "dns-default"
+DaemonSet does in OpenShift 4.7 and earlier) and run on all Linux nodes.
+
+### Risks and Mitigations
+
+A cluster administrator could configure a node selector, or taint all nodes, in
+such a way that DNS Pods could not be scheduled to any node, rendering the DNS
+service unavailable.  Because the DNS service is critical to other cluster
+components including OAuth, fixing misconfigured DNS Pod placement parameters
+could be impossible for the cluster administrator to do.
```
smarterclayton, line 257: Is this only scenario that could block fixing this the user being unable to login?  I’m skeptical that a one time check for placement is the right approach, vs (for example) always ensuring dns runs on the control plane nodes if there are no valid worker nodes in use (or separately as a fallback).  Emulating the scheduler is probably not a great idea (vs detecting zero pods at runtime or always running the previous topology until the new topology is in place, or “failing open” if no pods are detected).  Generally doing any sort of static detection of schedulability adds fragility to a system.

enhancements/dns/configurable-dns-pod-placement.md
Miciah: Service topology is alpha and deprecated, to be superseded by internal traffic policy, which is alpha (and not yet implemented AFAIK).  Users have been strenuously complaining about this for a while—some security policies require prohibiting traffic among worker nodes, so DNS needs to run on non-worker nodes only.

enhancements/dns/configurable-dns-pod-placement.md
Miciah: Yeah, if DNS isn't available to resolve `kubernetes.default.svc.cluster.local`, then presumably that could break OAuth.  If the administrator didn't already have an OAuth token handy, the problem might then be impossible to resolve.  If the user did have an OAuth token, it should be possible to patch the CRD to revert the problematic node-placement parameters.  I agree that trying to infer what the scheduler will do is an imperfect mitigation.  We could add some logic to the reconcile loop to detect if the operator has been reporting degraded for &gt;_n_ minutes and has 0 DNS pods; would it be valid for the operator to revert `spec.nodePlacement`, or does the operator need to maintain some state to track that the current value in `spec.nodePlacement` resulted in having 0 pods and needs to be ignored?

knobunc:
/hold remove

We can address any other concerns in follow-up PRs.

knobunc:
/hold cancel

We can address any other concerns in follow-up PRs.

PR #668, opened 2021-02-25
Add Customize Project Access Roles proposal

jerolimov:
**Issues:**
Epic: https://issues.redhat.com/browse/ODC-4650
Story: https://issues.redhat.com/browse/ODC-5394

Rendered preview: https://github.com/jerolimov/openshift-enhancements/blob/customize-project-roles/enhancements/console/customize-project-roles.md


enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,160 @@
...
+    - name: admin
+      label: Admin
+    - name: edit
+      label: Edit
+    - name: custom-role-name
+      label: "My custom role"
+  ...
+```
+
+- [ ] Should allow the admin to link `Role` or `ClusterRole`? Should this support other namespaces then the current?
```
christianvogt, line 120: I think we should support both `Role` and `ClusterRole`.
@spadgett what exactly did 3.11 support? I want to say that it was a single list that performed a lookup by name in both cluster roles and roles lists.

The namespace will be implied based on the project in which the rolebinding will be created.

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,160 @@
...
+```yaml
+apiVersion: operator.openshift.io/v1
+kind: Console
+metadata:
+  name: cluster
+  ...
+spec:
+  customization:
+    brand: online
+    projectAccessRef:
```
christianvogt, line 133:
```suggestion
    projectAccess:
```

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,162 @@
...
+
+- Introducing a UI to configure this project access roles without a YAML editor.
+
+## Proposal
+
+### Extending the Console CRD
+
+Similar to [Customize Developer Catalog Categories](./catalog-categories.md) we extend the existing `operator.openshift.io/v1` / `Console` CRD. It provides already a `spec.customization` area where we can add a list of roles.
+
+The list of roles should link to `rbac.authorization.k8s.io/v1` / `Role` or `ClusterRole`.
```
spadgett, line 53: I would limit this to `ClusterRole` for now. I'm not sure it's a good idea to allow this to be configured for `Role` since `Roles` with the same names in different namespaces can have totally different permission. The cluster admin can't know what these will be when configuring the option. Also there is no way in the current UI to differentiate between `Role` and `ClusterRole`, so as a project admin you won't know exactly what permission you're granting.

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,160 @@
...
+```yaml
+apiVersion: operator.openshift.io/v1
+kind: Console
+metadata:
+  name: cluster
+  ...
+spec:
+  customization:
+    brand: online
+    projectAccess:
```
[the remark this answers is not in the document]
spadgett: I'd give this even more context like

```yaml
customization:
  developerPerspective:
    clusterRoleOptions:
      - admin
      - edit
      - view
```

Someone better at names might come up something better :) But I think it would be good to make it clear this only applies to the dev perspective project access form, not the general create role binding form.

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,160 @@
...
+
+#### How we handle upgrade
+
+**Mitigation**: We keep the default values as part of the console code, so that without customization the console always shows the default roles. If the default values are changed, the admin needs to merge the new values with their customization manually. The benefit of this is that we will never modify an already customized list.
+
+**Drawback** As soon as the admin customizes the roles, a cluster update will not add new default roles or drop none-existing roles.
+
+## Open Questions
+
+- [ ] Should we provide a way that roles could show a label (visible label in the select or dropdown for the user) instead of just the resource name.
```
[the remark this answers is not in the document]
spadgett: I thought at one point we supported an `openshift.io/description` annotation on the role, which would be a way to do this. We might check if any out of the box `ClusterRoles` have that set.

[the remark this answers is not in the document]
jerolimov: I checked that already and don't find any existing annotation at a `ClusterRole` which could be used here

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,160 @@
...
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The OpenShift Dev Console provides a simplified way to share access the selected project. Users can be assigned to specific roles under the Project menu item on the Project access tab. Currently only the roles "Admin", "Edit" and "View" are selectable.
+
+Cluster admins want give project admins only access to a customized list of predefined roles to force compliance rules or provide more options.
```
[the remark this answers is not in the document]
jerolimov: Reordered motivation (why) and summary (what) to make things clearer and hopefully less repeating.

enhancements/console/customize-project-roles.md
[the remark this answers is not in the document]
jerolimov: Combined both, until we have a better naming:

```yaml
    developerPerspective:
      projectAccessRoles:
```

Or do you prefer `clusterRoleOptions`? My thought here is that its really restrictive to `ClusterRoles`. Which is maybe also fine when we don't want like `Roles` :thinking:

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,162 @@
...
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The OpenShift Developer Console provides a simple way to share access to the selected project. Users can be assigned to specific roles under the Project menu item on the Project access tab. Currently only the roles "Admin", "Edit" and "View" are selectable.
```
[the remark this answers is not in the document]
jerolimov: Added a screenshot and reordered motivation (why) and summary (what). See https://github.com/jerolimov/openshift-enhancements/blob/customize-project-roles/enhancements/console/customize-project-roles.md#motivation

Wdyt?

enhancements/console/customize-project-roles.md
jerolimov: We can show R or CR in the dropdown. But I think ClusterRole is fine because the admin could not predict what project admins allow with their namespaced `Role` -- as you mentioned. So, this is our answer for just a list of strings, right? No label, no kind is required here. Just `projectAccessRoles` or `clusterRoleOptions` or `projectAccess.availableClusterRoles` ...

enhancements/console/customize-project-roles.md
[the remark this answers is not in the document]
christianvogt: @jerolimov sam is right about `openshift.io/description` we may not have it, but we should note to display this value in the UI if available.

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Motivation
+
+Customers with customized roles want to provide easy access for their project admins to restrict access within a project.
+
+OpenShift Console 3.11 allowed customers to customize the list of roles visible in the membership page. This enhancement will provide equivalent functionality in 4.x.
+
+## Summary
+
+Cluster admins want give project admins only access to a customized list of predefined roles to force compliance rules or provide more options.
```
christianvogt, line 34:
```suggestion
Cluster admins want to give project admins only access to a customized list of predefined roles to force compliance rules or provide more options.
```

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,167 @@
...
+apiVersion: operator.openshift.io/v1
+kind: Console
+metadata:
+  name: cluster
+  ...
+spec:
+  customization:
+    brand: online
+    developerPerspective:
+      projectAccessRoles:
```
christianvogt, line 67: I'm not a fan of isolating the config under a `developerPerspective` stanza just because I don't want isolate a feature to one perspective.

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,167 @@
...
+      - admin
+      - edit
+      - view
+  ...
+```
+
+This data will be provided in the `SERVER_FLAGS` as ``.
```
christianvogt, line 74: Not sure this implementation detail is required as part of this doc.

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,167 @@
...
+
+The admin needs to know the current default roles to customize them. We don't want to provide the default roles in the CR so that we can change these defaults later without migrating the customer data.
+
+To provide the default values close to the editor, the console will provide a code snippet in the sidebar of the YAML editor when editing the `Console` operator config.
+
+### User Stories
+
+#### Story 1
+
+As a cluster admin, I want to customize (add, update or remove) some of the Developer Console project access roles. The admin needs to know the current list of roles.
```
christianvogt, line 86:
```suggestion
As a cluster admin, I want to customize (add, update or remove) some of the Developer Console project access roles. The cluster admin needs to know the current list of roles.
```

jerolimov:
@christianvogt @spadgett Updated the doc again. Going with ClusterRoles only for now. The list are only strings now, if we want support other roles later we can also keep the kind and validate that the user uses only `ClusterRole` in the first version.

/cc @jhadvig @rottencandy

And we have another issue with different ideas about the yaml property name. My initial approach was `customization.projectAccess`

Sam recommended a `developerPerspective` separation like `customization.developerPerspective.clusterRoleOptions` because this cluster role options will not be shown in the admin perspective. Btw: Why not?

Christian is against a separation for perspectives. Now we need a decision. I also can provide some other ideas:

1. `customization.projectAdmin.availableClusterRoles` make it also clear that this will affect the project detail page for project admins (in developer perspective) and not cluster admins (in admin perspective)?
2. `customization.projectAdminClusterRoles`
3. `customization.availableProjectClusterRoles`

Wdyt?

enhancements/console/customize-project-roles.md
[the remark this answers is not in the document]
jerolimov: Added to the doc that we will show the annotation `openshift.io/description` if available.

enhancements/console/customize-project-roles.md
jerolimov: Yeah thats true. It's not our "external" API. But it's part of our internal contract between bridge and frontend. Don't know as well.

enhancements/console/customize-project-roles.md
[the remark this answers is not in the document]
jerolimov: What to you think about `projectAdmin.availableClusterRoles` ? It's a more functional description then using the perspective?

enhancements/console/customize-project-roles.md
jerolimov: Removed

enhancements/console/customize-project-roles.md
jhadvig: @christianvogt check https://docs.openshift.com/container-platform/3.11/install_config/web_console_customization.html#changing-the-membership-whitelist which was done by [loading extension scripts](https://docs.openshift.com/container-platform/3.11/install_config/web_console_customization.html#changing-the-membership-whitelist).
Based on the docs, we only supported ClusterRole.

enhancements/console/customize-project-roles.md
jhadvig: Also noticed that the dropdown has `Role` label and also the help-text above is linking to `Role` docs. If we want to just support `ClusterRole` for no we should make that clear, for both mentioned elements.

jerolimov: Good point, then we should also link directly to ClusterRoles and ClusterRoleBindings in the help text above the input. Sam, Christian, @serenamarie125 fine with that?

jhadvig (commented):
@jerolimov thanks for all this, I've added couple of comments, looks really good 👍 
Regarding the structuring of the consoleserver config. I would not necessarily go with `customization. developerPerspective` since that field would also need to contain other fields that are DecConsole specific, like `DeveloperCatalog`. But since we are not separating the perspectives from the beginning, I don't think we should start now, but rather stick to the approach we have.

From my understanding, the customised `ClusterRoles` will be only visible in the DevPerspective -> Project Access page. For that reason I would suggest to use `projectAccess`, instead of `projectAdmin` field name:
```yaml
apiVersion: operator.openshift.io/v1
kind: Console
metadata:
  name: cluster
  ...
spec:
  customization:
    brand: online
    projectAccess:
      availableClusterRoles:
      - admin
      - edit
      - view
```

We should also keep in mind that if in the future we decide to support also `Roles`, we can just add another field which would also contain a specific namespace:
```yaml
apiVersion: operator.openshift.io/v1
kind: Console
metadata:
  name: cluster
  ...
spec:
  customization:
    brand: online
    projectAccess:
      availableClusterRoles:
      - admin
      - edit
      - view
      availableRoles:
        namespace: foo
        roles:
        - ...
```

enhancements/console/customize-project-roles.md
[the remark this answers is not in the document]
jerolimov: "This change will only affect the project page in the developer perspective." + screenshot is there. So I mark this as resolved. :)

christianvogt:
^^ I'm good with the suggestion. Went full circle on `projectAccess` :)
I have a feeling that at some point the "perspectives" will be nothing more than an opinionated nav when we get to allow for custom navigation. In which case there won't be a hard line between what is "admin" vs "dev" perspective.

jerolimov:
Thanks Jakub, I liked `projectAccess` also more then `projectAdmin`, also because this issue used project access always in its title :smile: 

Anything else to-do here? :smirk:

kotarusv:
This feature is very much required for Cisco. This is a kind of show stopper for our migration from 3.11

jhadvig (approved):
@jerolimov I think we should be good to merge this. @spadgett any additional comments?

enhancements/console/customize-project-roles.md
```diff
@@ -0,0 +1,176 @@
...
+
+OpenShift Console 3.11 allowed customers to customize the list of roles visible in the membership page. This enhancement will provide equivalent functionality in 4.x.
+
+## Summary
+
+Cluster admins want to give project admins only access to a customized list of predefined roles to force compliance rules or provide more options.
+
+The OpenShift Developer Console provides a simple way to share access to the currently selected project. This change will only affect the project page in the developer perspective. Users can be assigned here to specific roles on the project access tab. Currently only the roles "Admin", "Edit" and "View" are selectable.
+
+The `openshift.io/description` annotation of a `ClusterRole` can be used as label if it is defined.
```
spadgett, line 39: I would make it `console.openshift.io/description` if existing roles aren't already using an `openshift.io/description` annotation. There's a desire to not avoid new uses of `openshift.io/description` IIRC.

spadgett (approved):
/lgtm

PR #669, opened 2021-02-25
Add Customize Add Page proposal

jerolimov:
**Issues:**
Epic: https://issues.redhat.com/browse/ODC-5013
High level design: https://issues.redhat.com/browse/DTUX-849 on [Google Docs](https://docs.google.com/document/d/1GnLO2Fs0NGmmeHp4oD-jHb_9ZbWn-3r9P_vjyhlFXjw/edit)
Final design: https://issues.redhat.com/browse/PD-818 on [Google Docs](https://docs.google.com/document/d/1-PjCes7J9G-jXK3F6YgTV3QFHqkcwquyOoxaz16dKiQ/edit?usp=sharing)
Story: https://issues.redhat.com/browse/ODC-5412

Rendered preview: https://github.com/jerolimov/openshift-enhancements/blob/customize-add-page/enhancements/console/customize-add-page.md


enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,174 @@
...
+apiVersion: operator.openshift.io/v1
+kind: Console
+metadata:
+  name: cluster
+  ...
+spec:
+  customization:
+    brand: online
+    addPage:
+      hideGroups:
```
christianvogt, line 68: I don't think this is needed at this time as per the requirements. Groups are shown based on whether an action is present or not. I'd rather the admin remove the actions only in case new useful actions are added to a group, since we're taking the approach to configuring an exclusion list.

We can always revisit this later and add extra functiionality if required.

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,174 @@
...
+      - helmcharts
+      - ...
+  ...
+```
+
+Answer:
+
+> Customers requested to hide some unnecessary actions and add their own actions to the add page. Product management expected that it's easier for them to define a negative list, so that new actions will be shown automatically if an operator is installed.
+
+- [ ] If we save a list of shown and hidden actions, its maybe easier later to show an admin the information which new actions are available after an upgrade or when a new operator is installed.
```
christianvogt, line 143: Haven't seen this requirement from UX and PM.

christianvogt (commented):
Overall looks good. I never heard of a requirement to handle groups and PM / UX haven't chimed in yet to any inquiries. I think for the first pass we only support actions unless discussions resume and suggest otherwise.

enhancements/console/customize-add-page.md
jerolimov: removed all hideGroups references. Added this as a question with an answer

> No. Groups are shown based on whether an action is present or not. It's fine that the admin needs to disable new actions after a cluster upgrade.

enhancements/console/customize-add-page.md
jerolimov: Ack! Checked this question and added an answer

> No. This is not required at the moment and we will consider such requirements again when we will provide an UI for the admin in the future.

jerolimov:
@christianvogt See update above. Hope its fine to keep these questions with the answers in the doc.

Anything else?

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Motivation
+
+The OpenShift Developer Console provides a long list options to import applications, components and services. Some customers who uses only a part of the import options should hide unused options.
+
+## Summary
+
+Cluster admins want customize the add page of the OpenShift Developer Console.
```
christianvogt, line 33:
```suggestion
Cluster admins want to customize the add page of the OpenShift Developer Console.
```

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Motivation
+
+The OpenShift Developer Console provides a long list options to import applications, components and services. Some customers who uses only a part of the import options should hide unused options.
```
christianvogt, line 29:
```suggestion
The OpenShift Developer Console provides a long list of options to import applications and services. Some customers who uses only a part of the import options want to hide unused options.
```

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+
+- Introducing a UI to configure this without a yaml editor.
+
+## Proposal
+
+### Extending the Console CRD
+
+Similar to [Customize Developer Catalog Categories](./catalog-categories.md) we extend the existing `operator.openshift.io/v1` / `Console` CRD. It provides already a `spec.customization` area where we can hide actions from the add page.
+
+The cluster admin could disable actions via its id. This ensures that new later installed operators can show new actions without the requirement to enable them manually. To hide a complete group the admin should list all action ids of that group.
```
christianvogt, line 54:
```suggestion
The cluster admin can disable actions by ID. This ensures that later installed operators can show new actions without requiring the admin enable them manually.
```

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+
+A final yaml should look like this one:
```
christianvogt, line 56:
```suggestion
A final YAML should look like this one:
```

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+
+- [x] Should we save a list of enabled actions instead of a list of ids which should not be shown?
+
+For example:
+
+```yaml
+spec:
+  customization:
+    addPage:
+      actions:
```
christianvogt, line 120:
```suggestion
      disabledActions:
```

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+apiVersion: operator.openshift.io/v1
+kind: Console
+metadata:
+  name: cluster
+  ...
+spec:
+  customization:
+    brand: online
+    addPage:
+      hideActions:
```
christianvogt, line 68:
```suggestion
      disabledActions:
```

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+
+- [x] Should we also allow the cluster admin to hide complete groups?
+
+Answer:
+
+> No. Groups are shown based on whether an action is present or not. It's fine that the admin needs to disable new actions after a cluster upgrade.
+
+- [x] Should we save which actions are already configured by the cluster admin?
+
+If we save a list of shown and hidden actions, its maybe easier later to show an admin the information which new actions are available after an upgrade or when a new operator is installed.
```
christianvogt, line 139:
```suggestion
If we save a list of shown and hidden actions, it may be easier later to show an admin the information which new actions are available after an upgrade or when a new operator is installed.
```

christianvogt (commented):
The term `option` and `action` are interchanged in this doc. Could we be consistent with how we name the add page entries.

enhancements/console/customize-add-page.md
jerolimov: thx. done.

enhancements/console/customize-add-page.md
jerolimov: thx. replaced.

enhancements/console/customize-add-page.md
jerolimov: ok

enhancements/console/customize-add-page.md
jerolimov: This is an example for a positive list. Its under the question (Should we save a list of enabled actions instead of a list of ids which should not be shown?). So actions or enabledActions makes is fine here.

Under this example there is an answer that we don't want go this way.

enhancements/console/customize-add-page.md
jerolimov: ok.

enhancements/console/customize-add-page.md
jerolimov: Thanks again :)

jerolimov:
@christianvogt Pushed a commit with changes based on your recommendation. I hope that I don't add to many new issues. Wdyt?

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+      - knative.openshift.io/channel
+      - tekton.dev/pipelines
+  ...
+```
+
+### Provide a code snippet in YAML editor
+
+The admin needs to know the current available actions to hide them. We don't want to provide the defaults in the CR so that we can change these defaults later without migrating the customer data.
+
+To provide the default values close to the editor, the console will provide a code snippets in the sidebar of the YAML editor when editing the `Console` operator config. The snippets should include the ids of all extensions.
```
christianvogt, line 90:
```suggestion
To provide the default values close to the editor, the console will provide a code snippets in the sidebar of the YAML editor when editing the `Console` operator config. The snippets should include the IDs of all actions.
```

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+
+### User Stories
+
+#### Story 1
+
+As an cluster admin, I want to customize the add page. The admin needs to know the current list of available actions. The admin can configure a list of actions which should not be shown on the add page.
```
christianvogt, line 96:
```suggestion
As a cluster admin, I want to customize the add page. The admin needs to know the current list of available actions. The admin can configure a list of actions which should not be shown on the add page.
```

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+
+As a user, I don't want to see any actions on the add page which are hidden by the cluster admin. Empty groups should not be displayed.
+
+### Risks and Mitigations
+
+#### How we handle upgrade
+
+**Mitigation**: The cluster admin can provide a list of actions which should not shown to the users. All features which are not defined in that list will be shown after an upgrade.
+
+**Drawback** If the cluster admin installs a new OpenShift upgrade the customized add page will contain new or unexpected actions.
```
christianvogt, line 108:
```suggestion
**Drawback** If the cluster admin installs a new OpenShift upgrade, the customized add page will contain new or unexpected actions.
```

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Motivation
+
+The OpenShift Developer Console provides many ways to import applications and services. Some customers who uses only a part of this opportunities want to hide unused actions from the add page.
```
christianvogt, line 29: This line read a bit odd. Perhaps something like this instead:
```suggestion
The OpenShift Developer Console provides many ways to import applications and services. Customers who use only some of these action want to hide the unused actions from the add page.
```

jerolimov:
@christianvogt Take all your last type changes 1:1 and rebased this changes to one commit.

enhancements/console/customize-add-page.md
```diff
@@ -0,0 +1,171 @@
...
+      - knative.openshift.io/channel
+      - tekton.dev/pipelines
+  ...
+```
+
+### Provide a code snippet in YAML editor
+
+The admin needs to know the current available actions to hide them. We don't want to provide the defaults in the CR so that we can change these defaults later without migrating the customer data.
+
+To provide the default values close to the editor, the console will provide a code snippets in the sidebar of the YAML editor when editing the `Console` operator config. The snippets should include the IDs of all actions.
```
spadgett, line 90: We probably need more than a snippet to document these, particularly since a lot of admins use the CLI. Have we given thought about where else we can publish these? Maybe in the console customization topic in the OpenShift docs and the API docs for the property.

Have we considered a simple UI for admins to update them (even if not in 4.8)? Maybe a modal similar to what we have for table column customization.

christianvogt: A UI is being considered in future however plans yet in 4.8 or 4.9. This may fall under the umbrella of the same mechanism that will serve customized dashboards.

I agree that we can work with docs to provide guidance on customizing this and other configurations similar to it.

jerolimov: @spadgett @christianvogt Should we discuss this in slack and open a ticket in the epics for this? I also expect that we should do this for all 4 (!) customization options we will have in 4.8.

1. dev catalog categories
2. available cluster roles to assign on the project page (project access)
3. disable actions on the add page
4. disable quick starts

christianvogt: I think we can move forward with this enhancement and provide docs.

PR #671, opened 2021-03-01
WIP: enhancements/monitoring: Describe single replica topology mode

lilic:
As per @simonpasquier proposal doc, this just documents those findings and the minimal we from monitoring team will do do adhere to the new topology mode. cc @openshift/openshift-team-monitoring please have a look, and Simon cc whoever needs to have a look at this. 

The actual implementation is here https://github.com/openshift/cluster-monitoring-operator/pull/1070.

### DIFF
--- a/enhancements/monitoring/integration-with-topology-modes.md
+++ b/enhancements/monitoring/integration-with-topology-modes.md
@@ -28,7 +28,7 @@ see-also:
 
 ## Summary
 
-Monitoring stack needs to react based on the newly added single replica topology mode. Plan is to detect whenever
+The monitoring stack needs to react based on the `infrastructureTopology` field newly added to `infrastructures.config.openshift.io` CRD. Plan is to detect whenever
 cluster is in single replica topology mode and reduce the monitoring stack footprint by adjusting platform Prometheus
 instance to be non Highly Available.
 
--- a/enhancements/monitoring/integration-with-topology-modes.md
+++ b/enhancements/monitoring/integration-with-topology-modes.md
@@ -29,8 +29,8 @@ see-also:
 ## Summary
 
 The monitoring stack needs to react based on the `infrastructureTopology` field newly added to `infrastructures.config.openshift.io` CRD. Plan is to detect whenever
-cluster is in single replica topology mode and reduce the monitoring stack footprint by adjusting platform Prometheus
-instance to be non Highly Available.
+cluster is in single replica topology mode and reduce the monitoring stack footprint by adjusting platform monitoring
+instances to be non Highly Available.
 
 ## Motivation
 
--- a/enhancements/monitoring/integration-with-topology-modes.md
+++ b/enhancements/monitoring/integration-with-topology-modes.md
@@ -1,0 +1,90 @@
+---
+title: monitoring-topology-mode
+authors:
+  - "@lilic"
+reviewers:
+  - "monitoring-team"
+  - "???"
+approvers:
+  - "monitoring-team"
+creation-date: 2021-03-01
+last-updated: 2021-03-01
+status: implementable
+see-also:
+  - "/enhancements/cluster-high-availability-mode-api.md"
+---
+
+# Monitoring Stack on Single Replica Topology Mode
+
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Monitoring stack needs to react based on the newly added single replica topology mode. Plan is to detect whenever
+cluster is in single replica topology mode and reduce the monitoring stack footprint by adjusting platform Prometheus
+instance to be non Highly Available.
+
+## Motivation
+
+In the 4.8 time frame, OpenShift will introduce a new capabilities field in the Infrastructure API that will inform
+operators about the deployment topology. The immediate goal is for operators to adjust the deployment of their operands
+depending on whether the infrastructure is HA (at least 3 nodes) or not (single node).
+
+### Goals
+
+- Reduce monitoring footprint by lowering number of replicas
+
+### Non-Goals
+
+- User workload stack or its Prometheus instance is not adjusted
+- Reduce memory footprint
+- Address CRC requirements (though it might help)
+- Tests from monitoring team
+
+## Proposal
+
+To adjust to the single replica mode, in cluster-monitoring-operator we detect the topology mode based on the "cluster"
+named instance of `Infrastructure`, if that mode is `SingleReplicaTopologyMode`, we only spin up one instance of
+platform Prometheus for that cluster.
+
+### Risks and Mitigations
+
+There are risks that this is not enough to reduce the actual footprint of the monitoring stack. The other thing is that
+as a result of this monitoring is no longer highly available but there is no workaround for that as its a single node
+cluster.
+
+## Design Details
+
+### Test Plan
+
+N/A Its a non goal for monitoring team, apart from manual testing.
+
+### Graduation Criteria
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+Upgrades are a non goal right now.
+
+### Version Skew Strategy
+
+N/A
+
+## Implementation History
+
+## Drawbacks
+
+
+## Alternatives
+
+Skip installing monitoring stack is one alternative, but this is harder to do due to monitoring stack being a core
+component of OpenShift, and our CRDs being used by so many other components.


enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,90 @@
...
+- User workload stack or its Prometheus instance is not adjusted
+- Reduce memory footprint
+- Address CRC requirements (though it might help)
+- Tests from monitoring team
+
+## Proposal
+
+To adjust to the single replica mode, in cluster-monitoring-operator we detect the topology mode based on the "cluster"
+named instance of `Infrastructure`, if that mode is `SingleReplicaTopologyMode`, we only spin up one instance of
+platform Prometheus for that cluster.
```
simonpasquier, line 56: I'd say that CMO will scale all deployment and statefulset resources to only one replica if it detects that the `infrastructureTopology` field is `SingleReplica`.

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,90 @@
...
+### Goals
+
+- Reduce monitoring footprint by lowering number of replicas
+
+### Non-Goals
+
+- User workload stack or its Prometheus instance is not adjusted
+- Reduce memory footprint
+- Address CRC requirements (though it might help)
+- Tests from monitoring team
```
simonpasquier, line 50: Maybe explicitly call out that we assume the field to be immutable (e.g. CMO should retrieve the field once at start time)? @romfreiman is this correct?

simonpasquier (commented):
Thanks!

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Summary
+
+Monitoring stack needs to react based on the newly added single replica topology mode. Plan is to detect whenever
+cluster is in single replica topology mode and reduce the monitoring stack footprint by adjusting platform Prometheus
+instance to be non Highly Available.
+
+## Motivation
+
+In the 4.8 time frame, OpenShift will introduce a new capabilities field in the Infrastructure API that will inform
```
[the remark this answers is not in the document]
lilic: Sounds good, we only care about controlPlaneTopology for Prometheus. Should I say its a non goal we do anything in infraTopology fields?

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,90 @@
...
+operators about the deployment topology. The immediate goal is for operators to adjust the deployment of their operands
+depending on whether the infrastructure is HA (at least 3 nodes) or not (single node).
+
+### Goals
+
+- Reduce monitoring footprint by lowering number of replicas
+
+### Non-Goals
+
+- User workload stack or its Prometheus instance is not adjusted
```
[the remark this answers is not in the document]
lilic: I would say for UWM its the `infrastructureTopology` we would look at, so would be a bit more work but not too much.

enhancements/monitoring/integration-with-topology-modes.md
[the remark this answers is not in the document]
simonpasquier: My understanding is that `controlPlaneTopology` is specifically for components that run on control nodes. Since the monitoring stack (except CMO which is already single replica) is deployed on worker nodes, we should rely on `infrastructureTopology`.
@dhellmann is this correct?

lilic:
Need to adjust the OEP to scale down all the replicas of all our deployments.

enhancements/monitoring/integration-with-topology-modes.md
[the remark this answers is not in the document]
romfreiman: yes

enhancements/monitoring/integration-with-topology-modes.md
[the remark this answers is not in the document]
romfreiman: There are examples that looking at both fields already.

enhancements/monitoring/integration-with-topology-modes.md
romfreiman: Currently, yes. this is the assumption.

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,90 @@
...
+
+To adjust to the single replica mode, in cluster-monitoring-operator we detect the topology mode based on the "cluster"
+named instance of `Infrastructure`, if that mode is `SingleReplicaTopologyMode`, we only spin up one instance of
+platform Prometheus for that cluster.
+
+### Risks and Mitigations
+
+There are risks that this is not enough to reduce the actual footprint of the monitoring stack. The other thing is that
+as a result of this monitoring is no longer highly available but there is no workaround for that as its a single node
+cluster.
```
romfreiman, line 62: Currently by default the metrics disappearing in case of a reboot? Or where there are 2 replicas there are synced between 2 Prometheus instances? In a multi node scenario.
If it's the later, we'll be come complete stateless in case of SNO. Unless the user specify a pvc, correct?

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Design Details
+
+### Test Plan
+
+N/A Its a non goal for monitoring team, apart from manual testing.
```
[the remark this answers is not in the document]
romfreiman: You'll have to adjust here: https://github.com/openshift/origin/blob/master/test/extended/single_node/topology.go#L111
And you have the alerts ticket.
I dont know if you should add more tests specific to your operator behavior.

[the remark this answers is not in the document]
simonpasquier: >  I dont know if you should add more tests specific to your operator behavior.
I don't think we need anything special in the end-to-end tests of the operator since the SingleReplica mode isn't exposed in the operator's configuration.

enhancements/monitoring/integration-with-topology-modes.md
simonpasquier: In HA mode, the metrics aren't synced between Prometheus instances. The assumption is that when one Prometheus instance is down (reboot or pod restart), the other one is still up (we have anti-affinity rules ensuring that Prometheus is scheduled on different nodes). And we use Thanos querier as the main API entrypoint, Thanos pulls/aggregates data from all Prometheus instances.

lilic: Like Simon said for Prometheus this is the case, but different for other components e.g. Alertmanager. Should we include this in the proposal or do we think its known fact?

enhancements/monitoring/integration-with-topology-modes.md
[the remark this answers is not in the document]
dhellmann: `infrastructureTopology` does sound like the right field to observe.

How does the monitoring stack work in the HyperShift case, where the control plane is outside of the cluster but there may be "infrastructure" nodes in the cluster?

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,90 @@
...
+
+The monitoring stack needs to react based on the `infrastructureTopology` field newly added to `infrastructures.config.openshift.io` CRD. Plan is to detect whenever
+cluster is in single replica topology mode and reduce the monitoring stack footprint by adjusting platform monitoring
+instances to be non Highly Available.
+
+## Motivation
+
+In the 4.8 time frame, OpenShift will introduce a new capabilities field in the Infrastructure API that will inform
+operators about the deployment topology. The immediate goal is for operators to adjust the deployment of their operands
+depending on whether the infrastructure is HA (at least 3 nodes) or not (single node).
```
dhellmann, line 39: nit: Do we say `3` for the infrastructure nodes, too? Or just `> 1`?

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Goals
+
+- Reduce monitoring footprint by lowering number of replicas
```
dhellmann, line 43: This feels a bit like it's mixing 2 goals together. We need to be able to run in a non-HA mode because we may not have a place to put a 2nd or 3rd replica. We may also need to reduce the overall footprint, but that may come on top of the reduction to 1 replica.

Since the second part is called out as a non-goal, perhaps this would be better phrased as "Support monitoring stack with a single replica for non-HA configurations" or something like that?

enhancements/monitoring/integration-with-topology-modes.md
dhellmann: Although the field will be static for now, I don't think it's safe to assume that will always be the case. So, unless it represents significantly more work up front, it's better to watch for changes to the infrastructure config resource.

enhancements/monitoring/integration-with-topology-modes.md
dhellmann: > Like Simon said for Prometheus this is the case, but different for other components e.g. Alertmanager. Should we include this in the proposal or do we think its known fact?

Including background information like that helps reviewers who are less familiar with some of the existing implementation details (such as me!). Ideally, this enhancement would address the impact of the change on all of the components managed by the operator.

enhancements/monitoring/integration-with-topology-modes.md
wking: `infrastructureTopology` is not really about number of compute nodes.  You could theoretically set `SingleReplica` and still have hundreds of compute nodes.  It just means you want infrastructure-y components to avoid spending resources for high-availability purposes (e.g. running 1 replica).

enhancements/monitoring/integration-with-topology-modes.md
wking: lowering number of replicas may be sufficient, so it seems fine as an initial goal.  If we need more squeezing afterwards (e.g. by lowering scrape periods?  Reducing stored history?  I don't know what the relevant Prom knobs are), we can do that in follow-up work.  If we know right now that we will need those too, seems like it's worth putting them in as goals.  Or we could make the goal something like "monitoring does whatever they need to do to get under $CPU_QUOTA for $FRACTION of a single-node CI run", and leave the details unspecified in the enhancement.

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,90 @@
...
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+Upgrades are a non goal right now.
+
+### Version Skew Strategy
+
+N/A
```
wking, line 80: The new monitoring code needs to default to its current HA choices when it does not find `infrastructureTopology` on the status, so we don't break existing clusters when they update into the new monitoring operator code.  I think someone may have confirmed that the client injects [the kubebuilder default][1] automatically, but I don't remember clearly enough to feel comfortable with a N/A here.  Can we at least say that we need new monitoring operator code to be compatible with `Infrastructure` where the new topology fields are unset, preserving the implicit HA semantics, even if we don't spell out how that compatibility is going to work at the code level?

[1]: https://github.com/openshift/api/blob/175513fcccbe4eb5223cb45a7246bf697b835c3e/config/v1/types_infrastructure.go#L94

enhancements/monitoring/integration-with-topology-modes.md
simonpasquier: Agreed with @dhellmann. To be clear the scope of the enhancement from the monitoring point of view is strictly to support the new SingleReplica mode.

enhancements/monitoring/integration-with-topology-modes.md
simonpasquier: FWIW we've already validated that the field defaults to `HighlyAvailable`. But agreed that we can even ensure that the cluster monitoring operator will fallback to the default HA deployment if the `infrastructureTopology` field is missing.

lilic:
Closing in favour of Simons PR.

PR #673, opened 2021-03-01
short-circuiting-backoff

mshitrit:
Signed-off-by: mshitrit <mshitrit@redhat.com>

### DIFF
--- a/enhancements/machine-api/short-circuiting-backoff.md
+++ b/enhancements/machine-api/short-circuiting-backoff.md
@@ -40,10 +40,11 @@ By using `MachineHealthChecks` a cluster admin can configure automatic remediati
 The machine healthcheck controller's remediation strategy is deleting the machine, and letting the cloud provider
 create a new one. This isn't the best remediation strategy in all environments.
 
-Any Machine that enters the `Failed` state is remediated immediately, without waiting, by the MHC
-When this occurs, if the error which caused the failure is persistent (spot price too low, configuration error), replacement Machines will also be `Failed`
-As replacement machines start and fail, MHC causes a hot loop of Machine being deleted and recreated
-Hot loop makes it difficult for users to find out why their Machines are failing.
+Any Machine that enters the `Failed` state is remediated immediately, without waiting, by the MHC.
+When this occurs, if the error which caused the failure is persistent (spot price too low, configuration error), replacement Machines will also be `Failed`.
+As replacement machines start and fail, MHC causes a hot loop of Machine being deleted and recreated.
+This hot looping makes it difficult for users to find out why their Machines are failing.
+Another side effect of machines constantly failing, is the risk of hitting the benchmark of machine failures percentage - thus triggering the "short-circuit" mechanism which will prevent all remediations.
 
 With this enhancement we propose a better mechanism.
 In case a machine enters the `Failed` state and does not have a NodeRef or a ProviderID it'll be remediated after a certain time period has passed - thus allowing a manual intervention in order to break to hot loop.
@@ -54,11 +55,11 @@ In case a machine enters the `Failed` state and does not have a NodeRef or a Pro
 
 ### Goals
 
-- Create the ability to define customized remediation for Machine that enters the `Failed` state.
+- Create the opportunity for users to enact custom remediations for Machines that enter the `Failed` state.
 
 ### Non-Goals
 
-TBD
+- This enhancement does not seek to create a pluggable remediation system in the MHC.
 
 ## Proposal
 
@@ -68,15 +69,15 @@ We propose modifying the MachineHealthCheck CRD to support a FailedNodeStartupTi
 
 #### Story 1
 
-As an admin of a hardware based cluster, I would like failed machines to delay before automatically re-provisioning so I'll have a time frame in which to manually analyze and fix them .
+As an admin of a hardware based cluster, I would like failed machines to delay before automatically re-provisioning so I'll have a time frame in which to manually analyze and fix them.
 
 ### Implementation Details/Notes/Constraints
 
-If no value for failedNodeStartupTimeout is defined for the MachineHealthCheck CR, the existing remediation flow
+If no value for `FailedNodeStartupTimeout` is defined for the MachineHealthCheck CR, the existing remediation flow
 is preserved.
 
-In case a machine enters the `Failed` state and does not have a NodeRef or a ProviderID it's remediation will be requeued by failedNodeStartupTimeout.
-After that time has passed if the machine current state remains remediation will be performed.
+In case a machine enters the `Failed` state and does not have a NodeRef or a ProviderID it's remediation will be requeued by `FailedNodeStartupTimeout`.
+After that time has passed if the machine current state remains, remediation will be performed.
 
 
 #### MHC struct enhancement
@@ -86,7 +87,7 @@ After that time has passed if the machine current state remains remediation will
         ...
     
         // +optional
-		FailedNodeStartupTimeout metav1.Duration `json:"failedNodeStartupTimeout,omitempty"`
+        FailedNodeStartupTimeout metav1.Duration `json:"failedNodeStartupTimeout,omitempty"`
     }
 ```
 
@@ -108,13 +109,13 @@ MachineHealthCheck:
 
 ### Risks and Mitigations
 
-No known risks
+No known risks.
 
 ## Design Details
 
 ### Open Questions
 
-See deprecation and upgrade
+See deprecation and upgrade.
 
 ### Test Plan
 
--- a/enhancements/machine-api/short-circuiting-backoff.md
+++ b/enhancements/machine-api/short-circuiting-backoff.md
@@ -1,0 +1,164 @@
+---
+title: short-circuiting backoff
+
+authors:
+  - @mshitrit
+
+reviewers:
+  - @beekhof
+  - @n1r1
+  - @slintes
+
+approvers:
+  - @JoelSpeed
+  - @michaelgugino
+  - @enxebre
+
+creation-date: 2021-03-01
+
+last-updated: 2021-03-01
+
+status: implementable
+
+see-also:
+  - https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191030-machine-health-checking.md
+---
+
+# Support backoff when short-circuiting
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+By using `MachineHealthChecks` a cluster admin can configure automatic remediation of unhealthy machines and nodes.
+The machine healthcheck controller's remediation strategy is deleting the machine, and letting the cloud provider
+create a new one. This isn't the best remediation strategy in all environments.
+
+Any Machine that enters the `Failed` state is remediated immediately, without waiting, by the MHC
+When this occurs, if the error which caused the failure is persistent (spot price too low, configuration error), replacement Machines will also be `Failed`
+As replacement machines start and fail, MHC causes a hot loop of Machine being deleted and recreated
+Hot loop makes it difficult for users to find out why their Machines are failing.
+
+With this enhancement we propose a better mechanism.
+In case a machine enters the `Failed` state and does not have a NodeRef or a ProviderID it'll be remediated after a certain time period has passed - thus allowing a manual intervention in order to break to hot loop.
+
+## Motivation
+
+- Preventing remediation hot loop, in order to allow a manual fix and prevent unnecessary resource usage.
+
+### Goals
+
+- Create the ability to define customized remediation for Machine that enters the `Failed` state.
+
+### Non-Goals
+
+TBD
+
+## Proposal
+
+We propose modifying the MachineHealthCheck CRD to support a FailedNodeStartupTimeout, this is the time period which controls remediation of a machine that enters the `Failed` state.
+
+### User Stories
+
+#### Story 1
+
+As an admin of a hardware based cluster, I would like failed machines to delay before automatically re-provisioning so I'll have a time frame in which to manually analyze and fix them .
+
+### Implementation Details/Notes/Constraints
+
+If no value for failedNodeStartupTimeout is defined for the MachineHealthCheck CR, the existing remediation flow
+is preserved.
+
+In case a machine enters the `Failed` state and does not have a NodeRef or a ProviderID it's remediation will be requeued by failedNodeStartupTimeout.
+After that time has passed if the machine current state remains remediation will be performed.
+
+
+#### MHC struct enhancement
+
+```go
+    type MachineHealthCheckSpec struct {
+        ...
+    
+        // +optional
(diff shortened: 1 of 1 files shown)


elmiko (commented):
added a couple comments, and i have a couple questions.

i see you've added the note about the failed state being paired with the missing noderef and providerid, but i wonder if this feature should be broader in scope? for example, a cluster admin might want to check all mhc remediations, and want to have a blanket timeout for any machine that is going to be remediated. i'm curious if this is something you considered?

the name of this is "short cirtuciting backoff", but i didn't see mention of the short circuit mechanisms. i imagine that there will be cases where the machines that are held in the failed state will cause the mhc to go above its max unhealthy limit, is this the backoff addressed? assuming so, i think it should be more specifically mentioned.

mshitrit:
> i see you've added the note about the failed state being paired with the missing noderef and providerid, but i wonder if this feature should be broader in scope? for example, a cluster admin might want to check all mhc remediations, and want to have a blanket timeout for any machine that is going to be remediated. i'm curious if this is something you considered?

Hmm I could be wrong, but I think what you are referring to is [NodeStartUpTime](https://github.com/openshift/machine-api-operator/blob/49ee07183dbd7b51aa665db74c3c67322a4c869f/pkg/apis/machine/v1beta1/machinehealthcheck_types.go#L82) which already exist.

> the name of this is "short circuiting backoff", but i didn't see mention of the short circuit mechanisms. i imagine that there will be cases where the machines that are held in the failed state will cause the mhc to go above its max unhealthy limit, is this the backoff addressed? assuming so, i think it should be more specifically mentioned.

Good point - Thanks !

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,164 @@
...
+
+- [x] 03/01/2021: Opened enhancement PR
+
+## Drawbacks
+
+no known drawbacks
+
+## Alternatives
+
+- Instead of delaying, canceling the remediation for failed machines.
```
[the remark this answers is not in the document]
mshitrit: Indeed, it's actually part of upstream intended implementation.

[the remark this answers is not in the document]
elmiko: i think you are correct, i was considering an alternative where we would be able to detect the hot looping by examining the metrics generated, and then use that as a signal to alter behavior in the mhc but the more i think about it the less it makes sense to add here. just thinking out loud =)

elmiko:
looking good to me, thanks for the update. @mshitrit is this still wip or are you ready for labels?

mshitrit:
Hi @elmiko 
I'm ready for labels, bring it on :smile:

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,165 @@
...
+
+As an admin of a hardware based cluster, I would like failed machines to delay before automatically re-provisioning so I'll have a time frame in which to manually analyze and fix them .
+
+### Implementation Details/Notes/Constraints
+
+If no value for failedNodeStartupTimeout is defined for the MachineHealthCheck CR, the existing remediation flow
+is preserved.
+
+In case a machine enters the `Failed` state and does not have a NodeRef or a ProviderID it's remediation will be requeued by failedNodeStartupTimeout.
+After that time has passed if the machine current state remains remediation will be performed.
```
elmiko, line 80:
```suggestion
After that time has passed if the machine current state remains, remediation will be performed.
```

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,165 @@
...
+
+
+#### MHC struct enhancement
+
+```go
+    type MachineHealthCheckSpec struct {
+        ...
+    
+        // +optional
+		FailedNodeStartupTimeout metav1.Duration `json:"failedNodeStartupTimeout,omitempty"`
```
elmiko, line 90: i think the spaces/tabs are a little off here

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Design Details
+
+### Open Questions
+
+See deprecation and upgrade
+
+### Test Plan
+
+The existing remediation tests will be reviewed / adapted / extended as needed.
```
elmiko, line 122: i think the main point here is that we will need to add e2e tests to ensure this behaves as we expect.

elmiko (commented):
hey @mshitrit , this generally looks good to me. i added some suggestions that i think help to clarify and a few grammar nits.

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,165 @@
...
+With this enhancement we propose a better mechanism.
+In case a machine enters the `Failed` state and does not have a NodeRef or a ProviderID it'll be remediated after a certain time period has passed - thus allowing a manual intervention in order to break to hot loop.
+
+## Motivation
+
+- Preventing remediation hot loop, in order to allow a manual fix and prevent unnecessary resource usage.
+
+### Goals
+
+- Create the ability to define customized remediation for Machine that enters the `Failed` state.
```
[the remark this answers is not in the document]
mshitrit: Much better :smile:

enhancements/machine-api/short-circuiting-backoff.md
mshitrit: I agree that e2e will have value here, but I think UT are also important here.
IMO with this comment we allow enough flexibility without putting a focus on a specific testing phase.

elmiko (approved):
this is looking good to me. we might remove some of the unused sections, but i'm ok as is.
/lgtm

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,165 @@
...
+### User Stories
+
+#### Story 1
+
+As an admin of a hardware based cluster, I would like failed machines to delay before automatically re-provisioning so I'll have a time frame in which to manually analyze and fix them.
+
+### Implementation Details/Notes/Constraints
+
+If no value for `FailedNodeStartupTimeout` is defined for the MachineHealthCheck CR, the existing remediation flow
+is preserved.
```
JoelSpeed, line 77: In the implementation you've set a default, this will be incompatible with this statement, you won't be able to remove the value. If you want to have no default, that would actually be preferable as this would then preserve existing behaviour for users who upgrade

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,165 @@
...
+    spec:
+      selector:
+        matchLabels: 
+          ...
+      failedNodeStartupTimeout: 48h
+```
+
+### Risks and Mitigations
+
+No known risks.
```
JoelSpeed, line 112: Currently, you're setting a default in the implementation and this would affect existing users, may be worth discussing pros/cons of a default here so we know whether to have one or not

enhancements/machine-api/short-circuiting-backoff.md
JoelSpeed: Worth adding some comments about how you might add an E2E test here

enhancements/machine-api/short-circuiting-backoff.md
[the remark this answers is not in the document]
JoelSpeed: Can you expand on what this means, this is really sparse as an alternative and it's not obvious to me what cancelling the remediation actually means

JoelSpeed (commented):
Please take a look at the comments I've added, there's some problems I think need exploring further before we can merge around defaulting/disabling the new field and also what happens to existing users when they upgrade, will this change the behaviour for them, is that desirable?

enhancements/machine-api/short-circuiting-backoff.md
mshitrit: That's a good point, here is my take on that:
**non default** (pro) - naive users aren't being surprised with a new behavior.
**default** (pro) - naive users do benefit from the new behavior.
I guess the real question here is whether this new behavior benefit all users or not. 
Let me know what you think.

enhancements/machine-api/short-circuiting-backoff.md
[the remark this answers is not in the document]
mshitrit: Sure, I suspect it is due to a poor word choice on my part.
The alternative I was referring is the original solution we've discussed, basically in case that machine state is failed and it has no node not to perform remediation.

enhancements/machine-api/short-circuiting-backoff.md
mshitrit: You are right.
Once we decide what's the best way to proceed regarding default/non default I'll make the proper adjustments.

enhancements/machine-api/short-circuiting-backoff.md
JoelSpeed: I think for now lets keep no default and maintain the existing behaviour, but can we get these pros/cons fleshed out in the risks/mitigations section within the doc?

enhancements/machine-api/short-circuiting-backoff.md
JoelSpeed: Would still like to see some details about the testing we will be able to add here

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,165 @@
...
+TBD
+
+##### Tech Preview -> GA
+
+TBD
+
+##### Removing a deprecated feature
+
+
+### Upgrade / Downgrade Strategy
```
JoelSpeed, line 143: All of this can be removed as it isn't really relevant

enhancements/machine-api/short-circuiting-backoff.md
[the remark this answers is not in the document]
JoelSpeed: Can we flesh this out within the document itself please?

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,165 @@
...
+
+- Instead of delaying, canceling the remediation for failed machines.
+
+## Infrastructure Needed [optional]
+
+Use this section if you need things from the project. Examples include a new
+subproject, repos requested, github details, and/or testing infrastructure.
+
+Listing these here allows the community to get the process for these resources
+started right away.
```
JoelSpeed, line 165: This section can go

JoelSpeed (commented):
No real objections to this, would like to see some extra details fleshed out, I've highlighted where these are in the document

Did we ever start a conversation upstream about this, if so, would be good to link that into this document too

enhancements/machine-api/short-circuiting-backoff.md
mshitrit: Hi, I assume you are not looking for info on Unit Test (which were added) and aiming to E2E tests.
I'm not that familiar with MAO e2e so I don't feel like there is more I can contribute here.
In case there is something specific you had in mind, I'll be happy to include it.

mshitrit:
Hi @JoelSpeed, 
Looks like removing the unused sections is causing the `ci/prow/markdownlint` Job to fail.
Do you want me to revert those changes, or keep them ?
```
enhancements/machine-api/short-circuiting-backoff.md missing "### Graduation Criteria"
enhancements/machine-api/short-circuiting-backoff.md missing "#### Dev Preview -> Tech Preview"
enhancements/machine-api/short-circuiting-backoff.md missing "#### Tech Preview -> GA"
enhancements/machine-api/short-circuiting-backoff.md missing "#### Removing a deprecated feature"
enhancements/machine-api/short-circuiting-backoff.md missing "### Upgrade / Downgrade Strategy"
```

JoelSpeed:
> Do you want me to revert those changes, or keep them ?

We will have to revert as the markdownlint is required, I'll pass this as feedback to the archs though, seems weird to enforce these titles for all enhancements

enhancements/machine-api/short-circuiting-backoff.md
slintes: The default was removed in the implementation, so I guess that was the decision
https://github.com/openshift/machine-api-operator/pull/814/commits/e3d7784a0ac8aededcf47c107e357f7d895f4484

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,154 @@
...
+
+
+#### MHC struct enhancement
+
+```go
+    type MachineHealthCheckSpec struct {
+        ...
+    
+        // +optional
+        FailedNodeStartupTimeout metav1.Duration `json:"failedNodeStartupTimeout,omitempty"`
```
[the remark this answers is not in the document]
slintes: Good question.
From reading this enhancement I'd say the same.
From reading the implemenation I'd say `Startup` makes sense, because the timeout is applied to the machine's creation timestamp. Not sure if that implementation is correct though. Maybe the timeout should be applied to the time when the machine reached the failed phase? (Not sure if that information is available...).
@beekhof @mshitrit WDYT?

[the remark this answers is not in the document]
mshitrit: Per the implementation `failedNodeStartupTimeout` kicks in for machines which their nodes presumably failed to start, so basically I agree with Marc that the name does make sense assuming the implementation is correct.
[Here](https://github.com/openshift/machine-api-operator/pull/814/files#diff-8b3f455c5e13c63eb4eb480e223a91c2051decc3269398fd2f53196196a6033fR764-R770) is the relevant implementation code.

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,154 @@
...
+create a new one. This isn't the best remediation strategy in all environments.
+
+Any Machine that enters the `Failed` state is remediated immediately, without waiting, by the MHC.
+When this occurs, if the error which caused the failure is persistent (spot price too low, configuration error), replacement Machines will also be `Failed`.
+As replacement machines start and fail, MHC causes a hot loop of Machine being deleted and recreated.
+This hot looping makes it difficult for users to find out why their Machines are failing.
+Another side effect of machines constantly failing, is the risk of hitting the benchmark of machine failures percentage - thus triggering the "short-circuit" mechanism which will prevent all remediations.
+
+With this enhancement we propose a better mechanism.
+In case a machine enters the `Failed` state and does not have a NodeRef or a ProviderID it'll be remediated after a certain time period has passed - thus allowing a manual intervention in order to break to hot loop.
```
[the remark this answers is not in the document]
slintes: @mshitrit do you mind merging this change, or give me write permission to your branch? Thx :)

enhancements/machine-api/short-circuiting-backoff.md
```diff
@@ -0,0 +1,154 @@
...
+    spec:
+      selector:
+        matchLabels: 
+          ...
+      failedNodeStartupTimeout: 48h
+```
+
+### Risks and Mitigations
+In case `FailedNodeStartupTimeout` is undefined default behaviour is preserved (i.e. remediation is not postponed).
+The Pro is that naive users aren't being surprised with a new behavior however, the con is that naive users do benefit from the new behavior.
```
elmiko, line 112: i think there might be a small error here
```suggestion
The Pro is that naive users aren't being surprised with a new behavior however, the con is that naive users do not benefit from the new behavior.
```

beekhof:
> Did we propose this feature upstream at all?

My folks didn't, but this idea came from the cloud team so maybe one of your people did?

PR #674, opened 2021-03-01
Service CA Certificate Generation for StatefulSet Pods, first version

mtrmac:
Proposed feature: The Service CA can be used to automatically generate certificates for StatefulSets,
to allow TLS-protected communication between StatefulSet pods.

### DIFF
--- a/enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
+++ b/enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
@@ -99,7 +99,7 @@ Once the secret is generated, users can scale the StatefulSet with no change to 
 
 To ensure that a compromised StatefulSet pod can’t impersonate other pods from the
 same StatefulSet,
-or any other pods attached ot the same headless service
+or any other pods attached to the same headless service
 (if this goal is achievable for the application,
 i.e. if the pods treat the intra-StatefulSet communication as untrusted
 and can plausibly protect against a rogue StatefulSet member):
@@ -132,16 +132,16 @@ Second best is to rely on known properties of the environment,
 i.e. to assume that 127.0.0.1 is localhost-only
 and to use a non-TLS connection on that address.
 
-If a _pod_ had to use TLS for localhost for some reason, the closest to a right way to do that
+If a pod _had_ to use TLS for localhost for some reason, the closest to a right way to do that
 (relying on TLS to verify that only loopback connections happen) is to:
-- Create a temporary CA (specific to the individual pod for)
-- Generate, and have that CA sign a localhost certificate (specific to the individual pod)
+- Create a temporary CA (specific to the individual pod)
+- Generate, and have that CA sign, a localhost certificate (specific to the individual pod)
 - Set up the CA certificate as a trusted root
-- _Irreversibly erase_ the CA private key
+- (Irreversibly erase the CA private key to minimize risk)
 - Set up the pod with the generated localhost certificate+private key.
 - (This doesn’t benefit from service-ca involvement at all.)
 
-Similarly single-purpose CA could be set up for NodeIP services
+Similarly, a single-purpose CA could be set up for NodeIP services
 (if it makes sense to connect to node IP addresses directly at all)
 so that the client connects to the desired NodeIP service
 instead of to any NodeIP service.
@@ -150,9 +150,7 @@ instead of to any NodeIP service.
 
 The only mitigations for the risks below are not to use this feature,
 (and prevent untrusted users from using it — using an admission controller?)
-so that the wildcard certificate is not created at all,
-or perhaps $somehow ensure that any users/code with access to the wildcard certificate
-is not likely to be compromised.
+so that the wildcard certificate is not created at all.
 
 #### Unexpected DNS Collisions in the Future
 
@@ -193,28 +191,9 @@ to include the wildcard subjects, if they are missing.
 
 ### Graduation Criteria
 
-#### Dev Preview -> Tech Preview
+N/A, just a code change to the Service CA.
 
-- Ability to utilize the enhancement end to end
-- End user documentation (TBD: where?)
-- Sufficient test coverage
-- Gather feedback from users rather than just developers
-- Enumerate service level indicators (SLIs), expose SLIs as metrics
-  TBD?
-- Write symptoms-based alerts for the component(s)
-  TBD?
-
-#### Tech Preview -> GA
-
-- More testing (upgrade, downgrade, scale)
-- Sufficient time for feedback
-- Available by default
-- Backhaul SLI telemetry
-  TBD?
-- Document SLOs for the component
-  TBD?
-- Conduct load testing
-  TBD?
+TBD: End-user documentation.
 
 ### Upgrade / Downgrade Strategy
 
@@ -223,17 +202,9 @@ requested for headless services.
 The previously-generated certificates will continue to be valid for their original lifetime,
 and previously-deployed components can continue to use them as before.
 
-If the upgrade to the first version that generates the wildcards
-also included an operator relying on this functionality,
-and the old version of the Service CA were still running,
-only the old version of the secret (with no wildcards) would be generated,
-and that could lead to surprising outcomes for a StatefulSet relying on the wildcard.
-Eventually, after the Service CA is upgrades,
-it will re-generate the secret to include the wildcard subjects.
-
-So, OpenShift components can either not use this feature until x.N+2,
-or design their StatefulSets to fail/wait until the Service CA is upgraded
-and a new wildcard certificate is available.
+No OpenShift components are expected to start using this feature until
+the minor version following the introduction of this feature,
+so version skew is not a current concern.
 
 Previous versions of the Service CA do not check subjects of the generated secrets
 (only that the secrets are signed by the currently-valid service CA root),
@@ -290,37 +261,6 @@ both creating the StatefulSet pods and a secret creation/update,
 which race against each other;
 So, the pods would have to explicitly account for that race,
 by failing, or waiting, if a certificate for their identity is not found.
-
-### Generate Multi-Service Certificates
-So that applications can simply use a single certificate on all listening ports,
-generate a single certificate that signs multiple ~unrelated services as subjects
-(combining the StatefulSet per-pod subjects and ClusterIP services,
-or a ClusterIP service with NodePort service),
-
-In the most general case, this is not possible to do using the current annotation approach,
-because services choose pods by pod labels,
-i.e. the set of services served by a single pod at once (which require a shared certificate)
-can only be determined once such a pod is created —
-and at that point it’s too late to generate a certificate for that pod.
-So, this would require defining a new Custom Resource
-that allows the user to define service combinations to be signed
-(and the user would be resposible for requesting the necessary service combinations).
-
-Generating multi-subject certificates also makes it more difficult
-to reason about security properties and to change service routing:
-once a single certificate for (serviceA+serviceB) exists,
-even if the implementation is later changed
-by moving serviceB to a separate set of pods,
-a compromised pod with access to the earlier combined certificate
-would allow an attacker to continue to impersonate both serviceA and serviceB
-(e.g. because the Service CA does not currently allow revoking certificates).
-
-If the different services are served on different ports,
-they can trivially use different certificates.
-
-Even for a single port exposed via the different services,
-the server should be able to use different certificates based on SNI nowadays
-(e.g. via https://golang.org/pkg/crypto/tls/#Config.Certificates ).
 
 ## Infrastructure Needed [optional]
 
--- a/enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
+++ b/enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
@@ -3,11 +3,14 @@ title: service-ca-cert-generation-for-statefulset-pods
 authors:
   - "@mtrmac"
 reviewers:
-  - TBD
+  - "@marun"
+  - "@sttts"
+  - "@Anandnatraj"
 approvers:
-  - TBD
+  - "@marun"
+  - "@sttts"
 creation-date: 2021-03-01
-last-updated: 2021-03-01
+last-updated: 2021-03-25
 status: implementable
 see-also:
 replaces:
@@ -31,11 +34,12 @@ The Service CA-generated certificates _for headless services_ include
 `*.${service.name}.${service.namespace}.svc` and
 `*.${service.name}.${service.namespace}.svc.cluster.local`
 as subjects,
-to allow TLS-protected communication between StatefulSet pods.
+to allow TLS-protected connections to individual StatefulSet pods.
 
 ## Motivation
 
-StatefulSet-managed pods often need to communicate with each other.
+StatefulSet-managed pods often need to communicate with each other,
+or may be directly accessed by clients.
 This traffic may need to be TLS-protected; because it is cluster-internal,
 and necessary for deploying such a StatefulSet, letting the cluster automatically
 manage the certificates is a natural extension to the same feature already
@@ -43,7 +47,7 @@ provided by the Service CA to allow TLS use with services.
 
 ### Goals
 
-Allow deploying StatefulSets with pods that communicate with each other using TLS,
+Allow connections directly to individual StatefulSets pods using TLS,
 without having to manually generate certificates for these pods.
 
 ### Non-Goals
@@ -95,7 +99,7 @@ it does not provide individual identity.
 
 Once the secret is generated, users can scale the StatefulSet with no change to the usual process.
 
-#### Using TLS to Guarantee Individual Identity of StatefulSet Pods, or StatefulSets
+#### Non-goal: Using TLS to Guarantee Individual Identity of StatefulSet Pods, or StatefulSets
 
 To ensure that a compromised StatefulSet pod can’t impersonate other pods from the
 same StatefulSet,
--- a/enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
+++ b/enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
@@ -18,7 +18,7 @@ superseded-by:
 
 ## Release Signoff Checklist
 
-- [x] Enhancement is `implementable`
+- [ ] Enhancement is `implementable`
 - [ ] Design details are appropriately documented from clear requirements
 - [ ] Test plan is defined
 - [ ] Operational readiness criteria is defined
@@ -27,16 +27,11 @@ superseded-by:
 
 ## Summary
 
-The Service CA can be used to automatically generate certificates for StatefulSets,
+The Service CA-generated certificates _for headless services_ include
+`*.${service.name}.${service.namespace}.svc` and
+`*.${service.name}.${service.namespace}.svc.cluster.local`
+as subjects,
 to allow TLS-protected communication between StatefulSet pods.
-
-The user adds a `service.beta.openshift.io/serving-cert-secret-name: $secretName` annotation,
-and a `$secretName` secret is generated, which can be mounted in the StatefulSet’s pods.
-
-Creating/scaling up a StatefulSet triggers
-both creating the StatefulSet pods and a secret creation/update,
-which race against each other.
-So, the pods **must** fail, or wait, if a certificate for their identity is not found.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


marun (commented):
I appreciate your having transferred your original design document into this enhancement, thank you!

Having reviewed both the [original jira issue](https://issues.redhat.com/browse/API-1120) and this enhancement, I would like to see this enhancement detail the goals outlined in the issue and directly relate those goals to a chosen proposal and alternatives. I found it confusing to try to sort through implementation detail (i.e. how) for which I could not easily discern your intent (i.e. why) in a way that related to the goals from the issue.

On the issue, you commented that the enhancement differed from what was requested as follows:

```
Individual certificates for each pod in a StatefulSet
No localhost/127.0.0.1 certificates
No certificates combining multiple services.
```

If there are good reasons for choosing to deviate from the original request - e.g. because you've had out-of-band conversations with SMEs that suggested that direction - it would be helpful to see the results of those conversations documented in this enhancement so that both other stakeholders and future maintainers can better understand your intent.

Some questions that were still unclear to me after reading the enhancement:

 - why annotate the statefulset instead of the service?
 - why are wildcard certs not an option?
 - why is it preferable to tolerate the potential for races between statefulset and service ca?
    - non-racey alternatives include init containers, wrapper scripts, and wildcard certs
 - how does the proposal to create more certs than necessary guarantee against races during scaling?
 - is statefulset and service ca racing actually a problem given that a pod failing due to cert absence will be restarted?
 - how will a given proposal interact with a statefulset's configured `podManagementPolicy` or `updateStrategy`?

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+
+- Provide each pod in a StatefulSet with a separate secret containing only the certificate+key
+for that single pod:
+
+  This is not currently possible,
+  because the StatefulSet pod template does not have any templating mechanism
+  that would allow a ${podID}-dependent value to be used for a secret name or a volume mount;
+  there’s only StatefulSet.spec.volumeClaimTemplates , which is not practical to use.
+
+  Upstream I could find https://github.com/kubernetes/kubernetes/issues/40651 , which only ends up
```
sttts, line 68: would avoid first person

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+  - Set up the CA certificate as a trusted root
+  - _Irreversibly erase_ the CA private key
+  - Set up the pod with the generated localhost certificate+private key.
+  - (This doesn’t benefit from service-ca involvement at all.)
+  
+  Similarly single-purpose CA could be set up for NodeIP services.
+
+- Generate a single certificate that signs multiple ~unrelated services as subjects
+  (combining the StatefulSet per-pod subjects and ClusterIP services,
+  or a ClusterIP service with NodePort service):
```
sttts, line 111: one could create a separate cert for StatefulSet backed ClusterIP service, right? I.e. one that does has per-pod hostname, but prely the service IP and service DNS name as subjects.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+with a set of individual certificate+key pairs named `{0,1,2,3,…}.{crt,key}`.
+The subject of `$podID.crt` is the pod’s unique DNS identity,
+`${statefulset.name}-${podID}.${service.name}.${service.namespace}.svc` .
+Each pod from a `StatefulSet` can then pick and use its own certificate.
+
+Creating a `StatefulSet` triggers both a `Secret` creation and creating the `StatefulSet` pods,
+which race against each other.
+So, each individual pod **must** fail if a certificate for its individual identity is not found.
+We also generate a few more certificate+key pairs than necessary,
+to allow scaling up without such pod failures.
```
sttts, line 138: how many?

sttts: What happens if we scale too far?

sttts: nvmd about second question, it's answered below

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+
+#### Scaling the StatefulSet down
+
+- An admin decreases the StatefulSet’s requested replica count.
+- The StatefulSet controller starts tearing down the extra pods.
+- Concurrently, the Service CA controller observes the replica count change;
+  it’s unspecified whether that causes the now unnecessary certificates/keys to be removed
+  from the secret, or perhaps even revoked by the CA.
+
+#### Strictly isolating the StatefulSet pods’ certificates
```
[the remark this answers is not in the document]
sttts: the init container solution works, doesn't it? It needs extra work for the pod author, but if isolation matters for the use-case, it's possible. So I think it is fine here.

In other words: the non-goal is to have some automatic way of doing this without init-container.

Also note that "strictly" here in the story is not really "strictly". Would replace "strictly" with "manually".

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+that StatefulSet→Service reference is UID-agnostic,
+and it’s a different set of certificates
+that was not anticipated by the current consumers of the “service UID” certificate extension.
+
+If necessary, we can allocate and define a new “StatefulSet UID” extension,
+or look up a Service UID for StatefulSet.spec.serviceName,
+or maybe even store a StatefulSet UID in the “Service UID”  extension, in the future;
+it seems unnecessary for a first version of this feature.
+
+#### Secret Size Grows with Replicas
```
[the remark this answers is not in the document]
sttts: So this scale to 100'ish nodes before reaching 2 MB object size. I don't think there are many instances of stateful services of that size. And in that case you have probably other problems than using an in-cluster mechanism like service-ca.

mtrmac:
Thanks for the detailed review, I’ll research/update.

> On the issue, you commented that the enhancement differed from what was requested as follows:
> 
> - Individual certificates for each pod in a StatefulSet
> - No localhost/127.0.0.1 certificates
> - No certificates combining multiple services.
>
> If there are good reasons for choosing to deviate from the original request

See the current non-goals section ( https://github.com/openshift/enhancements/pull/674#discussion_r585242709 ). This is the first place that deviation is being discussed, there isn’t any out-of-band consensus on it.

---

> * is statefulset and service ca racing actually a problem given that a pod failing due to cert absence will be restarted?

Yes, I don’t think it’s much of a problem, which is why this design seems viable. There well may be something I’m missing.

> * how does the proposal to create more certs than necessary guarantee against races during scaling?

It doesn’t: if we manage to avoid a race, perfect; if the race does occur, the StatefulSet controller will retry, the secret will be updated, and eventually a pod should be successfully created. Alternatively, we could not create any extra certificates, and just have the race each time, expecting it to eventually resolve to a stable state.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+### User Stories
+
+#### Creating a TLS-protected StatefulSet
+
+- Author of the application, or of the container image wrapping the application,
+  ensures that the pod
+  - Determines its identity $podID (already the case)
+  - Loads `$podID.{crt.key}` on startup from a separate directory
+  - Exits with a failing status if those files are missing
+  - Uses the certificate+private key for accepting intra-StatefulSet connections
```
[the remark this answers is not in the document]
mtrmac: I really meant that the application might need to be modified to be able to use different certificates for different ports/services/host names (e.g. a different certificate for a ClusterIP service used by out-of-StatefulSet clients, and for the intra-StatefulSet connections); such a modification would not be necessary if we were able to provide a single certificate for all of those host names, as was requested in the issue.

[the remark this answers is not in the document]
mtrmac: … I’ll try to clarify.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+
+- An admin increases the StatefulSet’s requested replica count.
+- The StatefulSet controller starts creating the new pods.
+  If the certificate/key pairs already exist in the secret (e.g. if just addding one pod)
+  the pods start up succesfully;
+  if not, the pods fail, and the StatefulSet controller will wait and retry.
+- Concurrently, the Service CA controller observes the replica count increase;
+  if the existing secret does not have enough certificate/key pairs (including a margin),
+  it updates the secret so that it does.
+  (It’s unspecified whether pre-existing certificate/key pairs are reused or replaced.)
```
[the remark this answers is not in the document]
mtrmac: This was intended a a starting point for end-user documentation: we make no commitment on that behavior (and because just updating a secret does not restart pods, it really should not matter to end users). I appreciate that this is not helpful for readers that want to understand what the feature actually does.

The _implementation_ currently always re-generates the whole set of certificates (because that’s simpler to implement, and so that they all expire at a similar time and we avoid secret churn on serial expirations).

Is there a better place for a draft of end-user documentation like this? Or maybe we should just document what the implementation currently does, and commit to that behavior, giving up a little flexibility for an easier-to-understand feature.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
mtrmac: I don’t understand.

[For context, a minimal certificate is basically a tuple (subject public key, subject identity, issuer’s signature of the two)]. Within a certificate, the hostname and the subject are ~the same thing (whether in the subject DN, or in the subjectAltName extension), so I don’t know what having “purely the service DNS name as subject” but a “per-pod host name” would look like.

Yes, one can create a set of separate certificates:
- $statefulSet.Spec.Replicas separate certificates, each with a single per-pod host name
- A single certificate with the host name of ClusterIP $serviceA
- A single certificate with the host name of ClusterIP $serviceB

and a single pod can accept connections for all three of these (by configuring $serviceA and $serviceB selectors to use that pod, opening / mapping ports, etc.); that pod then has to be implemented so that it serves the appropriate certificate for each connection. That’s what this design recommends.

The user request, and something this proposal does not do, is to provide a single certificate that has all of these subjects, so that the pod can just serve a single certificate on all open ports and not have to differentiate.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+and necessary for deploying such a StatefulSet, letting the cluster automatically
+manage the certificates is a natural extension to the same feature already
+provided by the Service CA to allow TLS use with services.
+
+### Goals
+
+Allow deploying StatefulSets with pods that communicate with each other using TLS,
+without having to manually generate certificates for these pods.
+
+### Non-Goals
```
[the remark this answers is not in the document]
mtrmac: Moved the “support individual secrets in StatefulSet” and “generate a single multi-service certificate” discussion to the Alternatives section.

Left the localhost/IP address discussion here, as a thing I don’t think we should provide by the Service CA in any form.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+  Even for a single port exposed via the different services,
+  the server should be able to use different certificates based on SNI nowadays
+  (e.g. via https://golang.org/pkg/crypto/tls/#Config.Certificates ).
+
+## Proposal
+
+Similarly to the existing annotation on services,
+a `service.beta.openshift.io/serving-cert-secret-name: $secretName` annotation on a `StatefulSet`
+requests the `service-ca` controller to generate a single secret `$secretName`
+with a set of individual certificate+key pairs named `{0,1,2,3,…}.{crt,key}`.
```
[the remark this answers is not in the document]
mtrmac: Tried to clarify this (in the “Proposal” section now).

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+- Author of the application, or of the container image wrapping the application,
+  ensures that the pod
+  - Determines its identity $podID (already the case)
+  - Loads `$podID.{crt.key}` on startup from a separate directory
+  - Exits with a failing status if those files are missing
+  - Uses the certificate+private key for accepting intra-StatefulSet connections
+  - Trusts the Service CA for making outgoing intra-StatefulSet connections,
+    using the signed `${statefulset.name}-${podID}.${service.name}.${service.namespace}.svc`
+    (or the same with `….cluster.local`) host names.
+- The StatefulSet for the application
```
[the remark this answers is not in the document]
mtrmac: Like this, just moved within the use case?

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+- After the secret is created/updated, and the StatefulSet controller retries,
+  the new pods start up succesfully.
+
+#### Scaling the StatefulSet down
+
+- An admin decreases the StatefulSet’s requested replica count.
+- The StatefulSet controller starts tearing down the extra pods.
+- Concurrently, the Service CA controller observes the replica count change;
+  it’s unspecified whether that causes the now unnecessary certificates/keys to be removed
+  from the secret, or perhaps even revoked by the CA.
```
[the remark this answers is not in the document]
mtrmac: (The application should be ready to deal with revocation, if it ever did happen.) Dropped to avoid that discussion.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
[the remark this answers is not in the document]
mtrmac: Added a paragraph describing why an application admin might want this.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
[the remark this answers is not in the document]
mtrmac: Right now, the controller code gives up after (hard-coded) 10 failures creating a secret, and I think tries again (another 10 tries) on a complete re-sync that the library-go controller infrastructure causes every 20 minutes.

I’ll update this to use the `serving-cert-generation-error-num` annotation, so that repeated failures creating a secret cause the controller to give up permanently.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+Secrets can now have `originating-StatefulSet-{name,uid}` annotations
+to link back to the StatefulSet;
+this works just like the previous `originating-service-*` annotations.
+
+##### Alpha Annotations Not Used
+
+The StatefulSet controllers only use `service.beta.openshift.io` annotations,
+not the `alpha` annotation which are supported for services.
+This is new code, so supporting an older version seems unnecessary.
+(Or should this feature start out as `alpha`?)
```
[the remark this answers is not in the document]
mtrmac: Note to self: TBD.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+
+### Test Plan
+
+Unit and e2e tests similar to the existing service certificate tests, both in structure and scope.
+
+### Graduation Criteria
+
+TBD?
+
+Given the close (but not 100%) correspondence
```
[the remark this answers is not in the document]
mtrmac: Note to self: TBD.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,344 @@
...
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The Service CA can be used to automatically generate certificates for StatefulSets,
+to allow TLS-protected communication between StatefulSet pods.
+
+Similarly to the existing annotation on services,
```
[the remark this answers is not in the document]
mtrmac: Moved; left just the pod restart requirement.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
mtrmac: > The implementation currently hard-codes creating 5 extra certificates, or 30% more than spec.replicas, whichever is larger.

This is just a guess for how quickly already-operational StatefulSets are likely to scale up, with an assumption that generating new key pairs is cheap enough not to worry about; if that could become a bottleneck (e.g. if a FIPS-mode random number source were noticeably slow), this might require extra tuning or more work to avoid unnecessarily re-generating valid certificates in large batches.

mtrmac:
> Having reviewed both the [original jira issue](https://issues.redhat.com/browse/API-1120) and this enhancement, I would like to see this enhancement detail the goals outlined in the issue and directly relate those goals to a chosen proposal and alternatives.

I’m rather struggling with that because the Jira issue spends much of the space asking for things that seem not _essential_ for the feature to me, or actively unwanted:

> Story 1: Given a user deploys a database that runs as a StatefulSet type resource, also creates two service endpoint: one is ClusterIP type service(servicename1) and another is Headless type Service(servicename2).

Fine…

> There should be a [paraphrasing
> - single certificate with multiple subjects
> - that contains a service wildcard regardless of StatefulSet names or pod identities
> - that contains certificates for non-identities like `127.0.0.1` and `localhost`]

That’s not a user story, that’s an implementation prescription bundling 3-4 design decisions, providing no detailed rationale. (Or is that rationale somehow implied by the references to Etcd or Minio? I know barely anything about either.)

The PR describes why the three above aspects of these design decisions seem not strictly necessary, problematic, or outright unwanted.

The part that _is_ a user story
> The Minio or Etcd StatefulSet needs a Headless service to communicate between the pods of the same StatefulSet. This will require the podname.servicename to be mentioned in the subjectAltNames of the certificate generated to allow that communication to happen. 

is, I think, addressed by this PR.

It’s quite possible that I am missing something, and that the proposed design is not acceptable; if so, let’s please discuss the rationale for those aspects in detail, and record it here.

---

> * why annotate the statefulset instead of the service?
> * why are wildcard certs not an option?

I have added these two to the Alternatives section.

> * why is it preferable to tolerate the potential for races between statefulset and service ca?
>   * non-racey alternatives include init containers, wrapper scripts, and wildcard certs

I can see wildcard certs (see above).

How would init containers / wrapper scripts work? Do you mean that the pod should start,
and observe for changes in the volume mounting the secret, waiting for the certificates to show up?
- (I didn’t realize secrets are updated at container run-time; TO DO: verify)
- I think for scaling up, that would work essentially similarly (except that the delays/re-tries would be manually managed by the container instead of by the StatefulSet controller).
- When initially creating the StatefulSet, the secret may not yet exist if `service-ca` loses the race, so the Kubelet won’t even start the pod; AFAICS any init container / wrapper script would not be relevant in such a situation.

> * how will a given proposal interact with a statefulset's configured `podManagementPolicy` or `updateStrategy`?

Updated in the PR (basically “they don’t matter”).

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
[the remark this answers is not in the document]
marun: I don't think this much detail is suggested. Goals are intended to be succinct, and so are non-goals.

marun:
@mtrmac Much of this proposal seems predicated on the assumption that wildcard certs are not an option. However, the upstream document referenced to support this assertion [1] provides guidelines for dns-based service discovery. It doesn't mention - let alone can be considered authoritative on - the topic of certificate issuance.  

There is also mention of the potential for 'compromised' statefulset pods to impersonate other statefulset pod were wildcard certs to be employed. How likely or problematic this would be would very much depend on how a cluster and its workloads were being employed. I think it desirable to confirm with PM whether this is a problem that needs to be solved. It may be that the customer would not expect compromise (e.g. because they serve workloads for only trusted customers, or are running workloads that do not manage sensitive data). Also, no one implementation is likely to satisfy all use cases, and anticipating unasked requirements can often lead to the addition of unnecessary complexity that will become a maintenance burden going forward. YAGNI

I think requirements need to drive implementation rather than the other way around. Accordingly, I've added these questions to the RFE agenda [3] for further discussion with PM.

/cc @Anandnatraj 

1: https://github.com/kubernetes/dns/blob/master/docs/specification.md
2: https://issues.redhat.com/browse/API-1120
3: https://docs.google.com/document/d/1i90Krpr-QoNIvRCGp7lKcdZW8LYFSUVCvCUE97-9lhU/edit

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,444 @@
...
+  (e.g. because they only communicate using shared storage, or some HW-specific mechanism).
+  So, conceptually it’s a better fit to place the certificate request on the StatefulSet
+  than on the service.
+- Implementation-wise, to generate the
+  `${statefulSet.name}-${podID}.${service.name}.${service.namespace}.svc` certificate subjects,
+  the Service CA controller _must_ observe creation of StatefulSet objects.
+  OTOH it does not actually have to observe Services, because the same value is available as
+  `${statefulSet.name}-${podID}.${statefulSet.spec.serviceName}.${statefulSet.namespace}.svc`.
+
+### Generate Multi-Service Certificates
```
marun, line 415: There was no mention in the RFE of using a single certificate for multiple ports. The ask was for certificates to be issued that could included the SAN of multiple services. This would enable a port securing its traffic with such a certificate to accept traffic from multiple services of potentially different types (i.e. internal clients via clusterip, external clients via nodeip). I think this is an entirely reasonable request.

mtrmac:
> @mtrmac Much of this proposal seems predicated on the assumption that wildcard certs are not an option. However, the upstream document referenced to support this assertion [1] provides guidelines for dns-based service discovery. It doesn't mention - let alone can be considered authoritative on - the topic of certificate issuance.

That document does say that any other DNS names _may_ exist. Assuming the Service CA root is widely trusted within the cluster, creating certificates that may match unknown DNS names complicates security analysis — threat modeling of completely unrelated network traffic using mostly-unrelated host names suddenly has to worry about the right to create and annotate StatefulSets even if that traffic has otherwise nothing to do with StatefulSets.

If we can create a set of certificates with accurate subjects, I think that should be the _default_, rather than creating certificates with wildcard certificates that are known to imprecise (both WRT the generic k8s DNS specification that leaves things unspecified, and WRT the specific case of two StatefulSets named `database` vs. `database-staging`, where certificates for the first one can be used to serve the second one). We can do the precise thing and grant minimal privileges; why do anything else and introduce extra risks?

---

> There is also mention of the potential for 'compromised' statefulset pods to impersonate other statefulset pod were wildcard certs to be employed. How likely or problematic this would be would very much depend on how a cluster and its workloads were being employed.

I agree that this is somewhat obscure; in many StatefulSets a compromised pod might do just as much damage by modifying the data the StatefulSet is responsible for, or the implementation might not be designed to treat intra-StatefulSet traffic as potentially hostile in any case. OTOH it’s not too difficult to imagine that a single compromised pod (+ some way to divert / modify network traffic, which is implied in the threat model of TLS) could do more damage with the ability to impersonate, e.g. by that single pod achieving a quorum within the StatefulSet to “vote in” the attacker’s data changes.

If impersonation within a StatefulSet were a very pressing concern, we really should push for a change to the StatefulSet design so that individual per-pod secrets can be used _very easily_. I agree that this is somewhat unlikely to be a critical security boundary, which is why I think the init container approach could be good enough for those who do care, and do that kind of detailed analysis.

OTOH, to be clear, I have no hard data about the security design of common stateful software either way.

---

>Also, no one implementation is likely to satisfy all use cases, and anticipating unasked requirements can often lead to the addition of unnecessary complexity that will become a maintenance burden going forward. YAGNI

Once a wildcard certificate is issued, we can’t very easily undo that even if we changed our mind in a next OCP version; the Service CA does not support revocation (and that’s fine, revocation is hard if it should not amplify disruptions into downtime).

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
mtrmac: The connection is that if the pod is serving those different services on different ports, it’s trivial to implement the listeners on those ports to serve a different certificate for each port (~= different service), using basically any imaginable TLS server implementation.

Only if two or more services are served on a single port, where the application doesn’t inherently know which service the incoming request is coming for, might the application need extra logic, and a recent TLS server implementation to choose the right certificate (which is easy for _modern_ clients that use SNI; without that, the server might have to guess based on an IP address of the client, or something like that, which is problematic).

The only reasons the user would _need_ a multi-service certificate are 1) exposing a single port as multiple services, or 2) legacy software that can’t be modified (and can’t be hidden behind a service proxy?). The user may _prefer_ a single multi-service certificate so that there are fewer certificates to manage, sure.

The intra-StatefulSet traffic and the traffic for external clients is so conceptually different, that it seems very likely to me that the application implementation is already strongly differentiating between the two, and I’d expect it to already support using different certificates for the two services: If the application is designed to ever be deployed directly over the internet, it already needs to support serving an internet-recognized certificate (e.g. signed by one of the CAs from the browser ecosystem) for external traffic, whereas the intra-StatefulSet traffic is probably firewalled off and using internal host names that can’t be signed by a browser-trusted CA.

---

Anyway, creating combined certificates just based on service/StatefulSet annotations + how selectors of these objects map to pods is impossible in general, as described in the PR.

So we would need to either fall back to brittle heuristics (e.g. only consider pods in service/StatefulSet templates, and ignore any manually-{created,relabeled} pods), or move the secret generation requests out of service/StatefulSet annotations completely, maybe to some CR that contains (secret name to manage, a set of services that should be included in that secret); then the users would be responsible for ensuring that that CR’s set of services matches their actual deployment strategies.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,444 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The Service CA can be used to automatically generate certificates for StatefulSets,
+to allow TLS-protected communication between StatefulSet pods.
+
+The user adds a `service.beta.openshift.io/serving-cert-secret-name: $secretName` annotation,
+and a `$secretName` secret is generated, which can be mounted in the StatefulSet’s pods.
```
jackevans43, line 34: Currently the private key is PKCS#1 which only supports RSA and is no longer supported by some applications (eg. logstash). Can this be upgraded to PKCS#8 please?

jackevans43 (changes requested):

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
mtrmac: @jackevans43 That seems well out of scope of _this_ change. Please file a separate issue at https://github.com/openshift/service-ca-operator/ , or a separate enhancement proposal.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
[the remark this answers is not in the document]
mtrmac: I have moved the details into “Implementation Details/Notes/Constraints”, “constraints” is at least somewhat a better fit than “non-goals”.

mtrmac:
> * (I didn’t realize secrets are updated at container run-time; TO DO: verify)

Tested this, and updated the document:
- Creating a pod when the secret is missing entirely seems to pause the pod creation in ContainerCreating. That could simplify things, the application wouldn’t have to care in the case of _initially_ creating a StatefulSet which triggers creating a secret for the first time.
- Changes to the secret are without much delay visible inside already running pods’ volume mounts. So, on scaling up, the application can either trigger a pod failure and a StatefulSet-driven pod restart, or internally wait for the secret to show up. I don’t know enough to say which one would be preferred — failing seems simpler to me, but I have no idea whether the resulting k8s-visible failure and retry is useful or unwanted by users.

Either way the above does not change the basic point that the application author needs to account for the case of a pod starting when the corresponding certificate is not yet available, in the proposed design.

This would be much more noticeable in the alternative approach of modifying StatefulSets so that we can create individual per-pod secrets: then Kubelet would always pause the pod startup in ContainerCreating until a per-pod secret is available, i.e. the application would not have to specifically check for certificate availability.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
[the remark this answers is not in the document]
mtrmac: I have just copied the criteria from the template, as a starting point — marking things I know nothing about (SLIs, alerts, ) as TBD.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
mtrmac: To link this back to the original issue:

- etcd uses different ports for client-to-service communication (2379) and among-peers communication (2380), and the certificates / keys can be configured separately: https://etcd.io/docs/current/op-guide/security/ `--cert-file` vs. `--peer-cert-file`
- MinIO supports SNI to choose the appropriate certificate depending on what hostname the client is connecting to: https://github.com/minio/operator/blob/570a0a864eb624aa26b5207bf6ae5e4e0cc4c6f9/examples/tenant-with-custom-ca-certs.yaml#L127

So neither of these applications requires generating a single certificate valid for accepting connections for multiple different services.

marun:
I've confirmed with PM that:

- there is no customer requirement precluding the issuance of wildcard certs
- there is a customer requirement to support the issuance of certificates that can be used by multiple services.

Please update this enhancement accordingly, and feel free to reach out on slack to discuss further.

mtrmac:
Updated per discussion to just use add a wildcard subject to headless services, shifting all risks of that to users.

Also expanded on the upgrade / downgrade impact a bit.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,327 @@
...
+(if it makes sense to connect to node IP addresses directly at all)
+so that the client connects to the desired NodeIP service
+instead of to any NodeIP service.
+
+### Risks and Mitigations
+
+The only mitigations for the risks below are not to use this feature,
+(and prevent untrusted users from using it — using an admission controller?)
+so that the wildcard certificate is not created at all,
+or perhaps $somehow ensure that any users/code with access to the wildcard certificate
```
marun, line 154: Maybe just remove this line and the next? Cert security is out of scope for this proposal.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Test Plan
+
+Unit and e2e tests similar to the existing service certificate tests, both in structure and scope.
+
+Ensure that upgrades (or alternatively any controller re-syncs)
+cause re-generation of certificates for headless services
+to include the wildcard subjects, if they are missing.
+
+### Graduation Criteria
```
marun, line 194: Given the reduced scope of implementation, these sections are probably no longer relevant. The existing annotation will just gain support for generating wildcard certificates for headless services.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,327 @@
...
+  TBD?
+
+### Upgrade / Downgrade Strategy
+
+Upgrading to a version that adds the feature will trigger a regeneration of secrets
+requested for headless services.
+The previously-generated certificates will continue to be valid for their original lifetime,
+and previously-deployed components can continue to use them as before.
+
+If the upgrade to the first version that generates the wildcards
```
marun, line 226: I do not believe there are any operators relying on the combination of statefulsets+service ca today, and that is likely to remain true until the proposed feature lands in 4.8. As downgrades are only supported for emergency scenarios caused by failed upgrades (and in the CI testing that validates those scenarios), once a cluster is successfully upgraded to 4.8 any component would be able to safely rely on the proposed feature being available.

marun (commented):
Apologies for delay in reviewing!

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,327 @@
...
+
+## Proposal
+
+Currently, a service annotated with
+a `service.beta.openshift.io/serving-cert-secret-name: $secretName`
+requests the `service-ca` controller to generate a secret `$secretName`
+containing a certificate+key pair for `${service.name}.${service.namespace}.svc`
+(and a `….cluster.local`) as subjects.
+
+For headless services, generate certificates that include two more subjects,
```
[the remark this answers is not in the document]
mtrmac: Compatibility with existing users.

Headless services are, by naive clients, accessible using exactly the same ${service.name}.${service.namespace}.svc host names — the only difference is that for headless services that DNS entry points directly to the pods as multiple A/AAAA records, not to the service’s non-existent ClusterIP.

So, the pre-existing certificate generation can be used both for ClusterIP an headless services in the same way, and reading https://access.redhat.com/documentation/en-us/openshift_container_platform/4.7/html/security_and_compliance/configuring-certificates#add-service-certificate_service-serving-certificate does not suggest that headless services are not expected to work.

---

I could see an argument that it would be cleaner for a service to _either_ be used by the single service hostname (to contact any pod) _or_ to use the StatefulSet-generated individual hostnames for directed communication, but I’m not aware of any implementation restriction that would prevent a single service from being used for both at the same time; e.g. https://github.com/kubernetes/dns/blob/master/docs/specification.md#241---aaaaa-records says both DNS names must work simultaneously.

----

A possible alternative would be to let the user choose (generate a certificate for only the service host name, only the wildcard, or both); that would also mitigate the security risks of wildcards somewhat, in that we would not start generating certificates with completely unnecessary wildcard certificates. OTOH it’s again one more option for the user to deal with.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,327 @@
...
+- **Don’t accept the Service CA as a valid root of trust for connections intended
+  for individual StatefulSet pods**.
+
+#### Certificate Expiration, Manual Secret Regeneration, CA Rotations
+
+No change to existing Service CA use.
+
+### Implementation Details/Notes/Constraints
+
+#### Not Generating Certificates for `127.0.0.1`, `localhost`, or NodeIPs
```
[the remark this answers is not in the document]
mtrmac: I’m afraid I still need to follow up on that, it’s on my list.

In general, this is OK _only_ if the CA is completely single-purpose (only signs certificates for a single service, and clients only trust it for connection to that single service), i.e. no other client could ever be misled by an attacker presenting such a `localhost`/IP certificate.

In other cases, it would be better for the CA to only sign a subject that identifies the service (in this case I guess the `etcd.*` host names). That does not prevent connections from bypassing DNS or the external network and going to 127.0.0.1 — the clients can be configured to make a TCP connection to one host name and to verify the certificate for another, probably by setting `TLSClientConfig.ServerName` (compare https://github.com/golang/go/issues/5829 ).

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
mtrmac: The one thing that can’t just be a part of a service-CA PR is the end-user documentation. I realize that just listing it here won’t make it happen :)

I’ll drop the rest.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
mtrmac: Dropped — the $somehow is honestly a cop-out, and it’s not possible without the admission controller to restrict who can create wildcard certificates anyway.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
mtrmac: Thanks, simplified.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,267 @@
+---
+title: service-ca-cert-generation-for-statefulset-pods
+authors:
+  - "@mtrmac"
+reviewers:
+  - TBD
```
sttts, line 6: @marun myself and @Anandnatraj

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,267 @@
...
+approvers:
+  - TBD
```
sttts, line 8: @marun and myself

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,267 @@
...
+
+StatefulSet-managed pods often need to communicate with each other.
+This traffic may need to be TLS-protected; because it is cluster-internal,
+and necessary for deploying such a StatefulSet, letting the cluster automatically
+manage the certificates is a natural extension to the same feature already
+provided by the Service CA to allow TLS use with services.
+
+### Goals
+
+Allow deploying StatefulSets with pods that communicate with each other using TLS,
```
sttts, line 46: nit: not only with each other, also externally

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,267 @@
...
+  - `${service.name}.${service.namespace}.svc.cluster.local`
+  - `*.${service.name}.${service.namespace}.svc`
+  - `*.${service.name}.${service.namespace}.svc.cluster.local`
+- The user creates/updates a StatefulSet for the application to contain use the generated secret
+
+#### Scaling the StatefulSet Up/Down
+
+Once the secret is generated, users can scale the StatefulSet with no change to the usual process.
+
+#### Using TLS to Guarantee Individual Identity of StatefulSet Pods, or StatefulSets
```
sttts, line 98: maybe write: `Non-goal: Using TLS ...`

sttts:
Some nits.

Overall looks fine.

Have the general question about the wording around "securing traffic beween the StatefulSet pods". Isn't it also for external consumers of the service?

mtrmac:
> Some nits.
Thanks, updated.

> Have the general question about the wording around "securing traffic beween the StatefulSet pods". Isn't it also for external consumers of the service?

Yes, I forgot about that possibility. Updated throughout.

enhancements/authentication/service-ca-cert-generation-for-statefulset-pods.md
```diff
@@ -0,0 +1,272 @@
...
+N/A, just a code change to the Service CA.
+
+TBD: End-user documentation.
+
+### Upgrade / Downgrade Strategy
+
+Upgrading to a version that adds the feature will trigger a regeneration of secrets
+requested for headless services.
+The previously-generated certificates will continue to be valid for their original lifetime,
+and previously-deployed components can continue to use them as before.
```
sttts, line 208: is this supposed to be tested manually by QE?

mtrmac: The upgrade case is unit-tested ( https://github.com/openshift/service-ca-operator/pull/149 `"secret already contains pre-headless cert data, update to headless"` ).

That certificates continue to work until regenerated feels trivial to me (is that mistaken?), and as long as no OpenShift components start relying on the wildcard certificates in the 4.8 cycle, not likely to break anything in practice anyway AFAICS.

xingxingxia: > is this supposed to be tested manually by QE?

@barleyer fyi ^

PR #675, opened 2021-03-02
enhancements: Fix various links in the downstream operator-sdk proposal

timflannagan:
Update the link syntax for links that weren't being properly resolved
when viewing in markdown preview.


jmrodri (approved):
/lgtm

PR #366, opened 2020-06-07
kata containers enhancement proposal

ariel-adam:
Providing the enhancement document for the kata containers project to be reviewed.

PR #677, opened 2021-03-04
sandboxed containers enhancement proposal

fidencio:
This supersedes #366, as it provides fixes for the lint.

The reason a new PR has been opened is because Jens Freimman, who's the person dealing with #366 recently became a father and is out for his paternity leave.

### DIFF
--- a/enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
+++ b/enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
@@ -1,0 +1,370 @@
+---
+title: sandboxed-containers
+authors:
+  - "@ariel-adam"
+  - "@fidencio"
+  - "@jensfr"
+reviewers:
+  - "@cgwalters"
+  - "@mrunalp"
+approvers:
+  - "@cgwalters"
+  - "@mrunalp"
+creation-date: 2020-06-07
+last-updated: 2021-03-12
+status: implemented
+see-also:
+  - "[OpenShift sandboxed containers Operator](https://github.com/openshift/sandboxed-containers-operator)"
+  - "[CoreOS Extensions](https://github.com/openshift/enhancements/pull/317)"
+  - "[QEMU Extension](https://github.com/openshift/machine-config-operator/pull/2376)"
+replaces:
+  - ""
+superseded-by:
+  - ""
+---
+
+# sandboxed containers
+
+Sandboxed containers provides a container runtime using virtual machines,
+and hardware virtualization technologies, providing powerful workload
+isolation compared to the existing container solutions, with the same look
+and feel provided by vanilla containers.
+
+## Release Signoff Checklist
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Sumary
+
+Sandboxed containers will be integrated into OpenShift to provide the ability
+to run kernel isolated containers.  Under the hood, the feature will rely on
+[Kata Containers](https://katacontainers.io/), an Open Source project working
+to provide a stronger workload isolation using hardware virtualization
+technologies as an additional layer of defense.
+
+  
+## Motivation
+
+The main motivation to bring sandboxed containers to OpenShift is to
+provide the ability to handle the following use cases:
+- Run privileged/untrusted workloads.
+- Run workloads requiring administrative privileges above and beyond what is
+  secure in a shared kernel environment (native linux OCI runtime - `runc`)
+- Ensure kernel isolation for each workload.
+- Ensure default resource containment through VM boundaries.
+- Run any workload which requires:
+  - Custom kernel tuning (sysctl, scheduler changes, cache tuning, etc).
+  - Custom kernel modules (out of tree, special arguments, etc).
+  - Exclusive access to hardware.
+  - Root privileges.
+
+### Goals
+
+- Provide a way to enable running user workloads in sandboxed containers on
+  an OpenShift cluster for additional isolation when necessary.
+- Life-cycle management of the underlying sandboxed containers runtime.
+- Provide a security policy to control which users can run what workloads
+  in sandboxed containers in an OpenShift cluster.
+
+### Non Goals
+
+- Running workloads using a different kernel than the one used on the
+  OpenShift node.
+  - Although sandboxed containers are capable of supporting running a
+    different kernel then the one used on the OpenShift node, such features
+    will not be supported in order to simplify the development cycles and
+    testing efforts of this project.
+- Running OpenShift control plane and core operators on Sandboxed
+  Containers.
+  - These will continue to run on the native linux OCI runtime runc.
+
+## Proposal
+
+As part of our proposal, we'll cover a few user stories coming from the Telco industry.
+
+### User Stories
+
+#### Story 1 - Telco CNF deployments
+
+As part of 5G/NFV telco deployments there is a gradual migration from physical
+networking boxes to CNFs (container network functions). Some of these CNFs
+require root access while others could potentially create a threat on the telco
+cloud (as the kernel is shared in that case) by bypassing existing container
+isolation mechanisms. To address these issues, a recommended way is to use a
+sandboxed container runtime.
+
+#### Story 2 - Devops deployments
+
+Telcos are working hard to increase the rate of adding new capabilities to
+their clouds for competing with OTTs (over the top companies such as Netflix,
+Apple, Facebook etc…). This involves huge investments in devops tools and
+processes. In order for devops to deploy new features on production
+environments there is a need for high isolation of the deployed workloads in
+order to control and revert such changes.
+
+#### Story 3 - Hardware Vendors
+
+Many hardware vendors require custom kernel parameters at boot, custom sysctl
+variables, etc. They want to use containers as a convenient packaging format
+for their entire application, but still need access to lower level tuning for
+portions of their applications.
+
+### Implementation Details/Notes/Constraints
+
+For a cluster to be able to run workloads using sandboxed containers, the
+following things have to happen:
+- `kata-containers`, and its dependencies, have to be installed in the
+  appropriate nodes.
+- A runtime class has to be set up, in kubernetes.
+- `CRI-O` has to be configured to use the installed `kata-containers` as
+  the runtime for the newly configured runtime class.
+
+And as a way to perform the lifecycle management (install, upgrade, and
+uninstall) of the items mentioned above, the [OpenShift sandboxed containers
+operator](https://github.com/openshift/sandboxed-containers-operator) was
+created and is a crucial part of our project.  The operator can be used to
+deploy sandboxed containers on all, or on a set, of the worker nodes of a
+cluster.
+
+Described below, you can see how the `KataConfig` CustomResourceDefinition (CRD),
+implemented by the sandboxed containers operator, can be used for deployment:
+
+- Deploying on all the worker nodes:
+  ```yaml=
+  apiVersion: kataconfiguration.openshift.io/v1
+  kind: KataConfig
+  metadata:
+    name: example-kataconfig
+  ```
+- Deploying only on selected nodes:
+  ```yaml=
+  apiVersion: kataconfiguration.openshift.io/v1
+  kind: KataConfig
+  metadata:
+    name: example-kataconfig
+  spec:
+    kataConfigPoolSelector:
+      matchLabels:
+         custom-kata1: test 
+  ```
+
+In the coming subsections, let's dive into the details of how the OpenShift
+sandboxed containers operator covers each one of the points present above.
+
+#### Installation of kata-containers dependencies and artifacts on the node
+
+The following artifacts are installed in each one of the nodes where
+sandboxed containers workloads are supported:
+
+package         | dependency of   | rpm size |  installed size
+----------------|-----------------|----------|-----------------
+ipxe-roms-qemu  | qemu-kiwi       | 1.3M     |   2.58M
+kata-containers |                 |  46M     | 130.72M
+libpmem         | qemu-kiwi       |  79K     |    255K
+pixman          | qemu-kiwi       | 258K     |    682K
+qemu-kiwi       | kata-containers | 2.7M     |  12.08M
+qemu-kvm-common | qemu-kiwi       | 919K     |   2.68M
+seabios-bin     | qemu-kiwi       | 133K     |    384K
+seavgabions-bin | qemu-kiwi       |  43K     |    210K
+sgabios-bin     | qemu-kiwi       |  14K     |      4K
+
+The desired way of installing those in the nodes is taking advantage of
+[CoreOS Extensions](https://github.com/openshift/enhancements/pull/317),
+which adds support for cluster operators (notably the MCO) to deploy
+non-containerized software that is tested and versioned along with the OS,
+but not installed by default.
+
+**Pros:**
+1. There's no extra work related to properly loading QEMU dependencies, as
+   all the packages would be in installed in the same PATH its build was
+   intent for.
+2. QEMU RPMs have been tested / validated using an identical environment.
+3. No additional work with regarding to re-packaging those in another form
+   other than the RPMs which is the way they are consumed by other Layered
+   Products.
+4. We rely on an already existing mechanism to install all of the RPMs -
+   QEMU, QEMU dependencies, and Kata Containers - which are not part of the
+   base image.
+5. There's no chance of hitting issues related to parallel installing
+   packages, as the whole package installation is handled by the MCO.
+6. Updates / Removal become trivial as they are also handled by the MCO.
+
+**Cons:**
+1. 51.5MB extra to be downloaded - coming from QEMU, QEMU dependencies, and
+   Kata Containers - as part of the RHCOS Extensions, for all users relying
+   on extensions; ~150MB installed for those using the sandboxed containers
+   extension.
+
+As an **alternative** to the above option, we could partially rely on [CoreOS
+Extensions](https://github.com/openshift/enhancements/pull/317), but this time
+only for installing `qemu-kiwi` and its dependencies.  The content of the
+`kata-containers` RPM would then be copied to the cluster on
+`/opt/sandboxed-containers/`.
+
+**Pros:**
+1. 5.5MB extra to be download - coming from QEMU and its dependencies - as
+   part of the RHCOS Extensions, for all users relying on extensions; ~20MB
+   installed for those using the QEMU extension.
+2. There's no extra work related to properly loading QEMU dependencies, as
+   all the packages would be in installed in the same PATH its build was
+   intent for.
+3. QEMU RPMs have been tested / validated using an identical environment.
+4. No additional work with regarding to re-packaging those in another form
+   other than the RPMs which is the way they are consumed by other Layered
+   Products.
+5. We rely on an already existing mechanism to install part of the RPMs, QEMU
+   and its dependencies, which are not part of the base image.
+
+**Cons:**
+1. 130MB extra - coming from Kata Containers binaries - copied onto the host,
+   for those who'd be using the Sandboxed Containers,
+2. The sandboxed containers operator still does a work that, in theory,
+   should be restricted to the MCO.
+3. The sandboxed containers operator would have to implement update, upgrade,
+   and downgrade logic, instead of relying on the MCO to do so.
+
+#### Setting up the `kata` runtime class, in Kubernetes
+
+Kubernetes provides support for [Runtime Class](https://kubernetes.io/docs/concepts/containers/runtime-class/),
+a feature for selecting the container runtime configuration, which is used to run a Pod's containers.
+
+A Runtime Class is created, and named as `kata`, in order to ease the
+adoption for those already coming from using Kata Containers with
+Kubernetes.
+
+It's important to note that the Runtime Class support two features which are
+quite important for sandboxed containers:
+- Scheduling, which ensures Pods land only on nodes supporting the specific
+  Runtime Class.
+- Pod Overhead, which allows us to specify overhead resources that are
+  associated to a running pod, such as `QEMU` and `virtiofs` overhead, in our
+  specific case.
+
+Example `kata` Runtime Class that gets created
+```yaml=
+apiVersion: node.k8s.io/v1
+handler: kata
+kind: RuntimeClass
+metadata:
+  name: kata
+overhead:
+  podFixed:
+    cpu: 250m
+    memory: 160Mi
+scheduling:
+  nodeSelector:
+    node-role.kubernetes.io/worker: ""
+```
+
+#### Configuring CRI-O to support `kata` runtime
+
+`CRI-O` supports drop-in configuration files, and we take advantage of that
+feature to drop-in a kata-containers configuration file, which looks like:
+```ini=
+[crio.runtime.runtimes.kata]
+  runtime_path = "/usr/bin/containerd-shim-kata-v2"
+  runtime_type = "vm"
+  runtime_root = "/run/vc"
+  privileged_without_host_devices = true
+```
+
+This configuration is the glue between the installed RPMs and the Runtime
+Class added in Kubernetes, specifying which container runtime that will be
+used to run a workload using `kata` as its Runtime Class.
+
+#### Risks and Mitigations
+
+##### Metrics
+
+While working on the integration of Kata Containers, which is used under the
+hood by sandboxed containers, on to OpenShift, it's been realised that
+metrics won't fully work as expected during the tech preview.
+
+This happens because OpenShift uses `cAdvisor` for its monitoring tools, and
+`cAdvisor` is not capable of collecting the **container** information of a
+Kata Containers, which is running inside the a VM.  **pod** information,
+though, which includes the `containerd-shim-kata-v2`, the  `QEMU`, and the
+`virtiofs` processes, is properly collected.
+
+OpenShift plans to switch  to `CRI stats` in the future, which is part of an
+effort lead by Peter Hunt (@haircommander) and that can be seen
+[here](https://github.com/kubernetes/enhancements/pull/2364).
+
+##### Overhead
+
+There is an overhead involved in the usage of sandboxed containers.
+This is due to the usage of `Qemu` and `virtiofs` to run the sandboxed
+containers.
+
+The default overhead is specified in the `kata` Runtime Class and is described
+below:
+```yaml=
+overhead:
+  podFixed:
+    cpu: 250m
+    memory: 160Mi
+```
+
+## Design Details
+
+### Open Questions
+
+### Test Plan
+
+sandboxed containers team is already in touch with OpenShift QE and is
+currently performing tests related to:
+- The operator features installability.
+- Basic lifecycle of pods using sandboxed containers.
+- Command-line tools integration.
+- Logs, events, and metrics visiblity.
+- Terminal availability.
+- Scaling (both vertical and horizontal).
+- CPU and memory restrictions.
+- Pods behaviour based on node cordon/drain scenarios.
+
+Apart from those, upstream Kata Containers project is already onboarded as part
+of the OpenShift CI, which runs, daily, the `openshift-e2e` tests.  The same
+set of tests, `openshift-e2e`, is planned to run against the downstream
+releases.
+
+### Graduation Criteria
+
+#### Tech Preview -> GA
+
+- Supported upgrade, downgrade, and scale.
+- Sufficient time for feedback.
+- Solid automated process for building sandboxed containers operators and
+  everything it consumes.
+- Solid user-facing documentation.
+- Ability to gate users from using sandboxed containers (or from using native linux containers).
+  - See [this](https://github.com/openshift/enhancements/pull/562) proposal for more details.
+
+#### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature.
+- Deprecate the feature.
+
+### Upgrade / Downgrade Strategy
+
+Considering that `kata-containers` and its dependencies will be part of CoreOS Extensions, the
+upgrade / downgrade will be handled by the MCO.
+
+### Version Skew Strategy
+
+Considering that `kata-containers` and its dependencies will be part of CoreOS Extensions, the
+version skew strategy will be handled by the MCO.
+
+## Implementation History
+
+
+## Drawbacks
+
+
+## Alternatives
+
+
+## Infrastructure Needed [optional


fidencio:
Closing this one as we were able to unblock #366.

enhancements/kata-containers/kata-containers-dev-preview.md
```diff
@@ -0,0 +1,276 @@
...
+**Cons:**
+1. ~20 mb extra, installed in the host, for those who'd be using kata runtime
+2. machine-os-content would still carry this until there is a separate machine-os-content- extensions container
+
+
+
+### KataContainers Operator development
+The goal is to develop a kata operator that can be used to manage the entire lifecycle of kata components in an OpenShift cluster. We will have a controller that watches for a kata **custom resource (CR)** and a daemon-set that acts upon changes to the (CR) and do the required work on the worker nodes (install, uninstall update,...).
+
+Via the CR it will be possible to select a subset of worker nodes.
```
cgwalters, line 138: Why is this out of curiosity?  Why not e.g. enable the feature on all available nodes but use regular `nodeSelector` for steering?

Is it because it fully takes over the crio backend and so hence all containers on that node are kata?  That seems unfortunate if true.

enhancements/kata-containers/kata-containers-dev-preview.md
```diff
@@ -0,0 +1,276 @@
...
+- custom kernel tuning (sysctl, scheduler changes, cache tuning, etc);
+- custom kernel modules (out of tree, special arguments, etc);
+- exclusive access to hardware;
+- root privileges;
+- any other administrative privileges above and beyond what is secure in a shared kernel environment (regular runc).
+
+It should be noted that Kata Containers differ from Kubevirt:
+
+- Kubevirt aims to run **VM images** on OpenShift providing the VM with the look and feel of a virtual machine running as a legacy VM (CRD defining a Virtual Machine in Kubernetes)
+- Kata Containers aim to run **container images** on OpenShift in an isolated manner using virtualization tools (uses Runtime Class resource to choose Kata only for specific workloads which require kernel level isolation)
```
cgwalters, line 37: Semi random but there's actually a 3rd class of thing in here, which is "containers that use virt" - what we do with https://github.com/coreos/coreos-assembler/ overlaps a lot with kata; we want to control qemu, not be inside qemu.

cgwalters (approved):
Overall seems sane to me!

enhancements/kata-containers/kata-containers-dev-preview.md
fidencio: Hey @cgwalters!

> Why is this out of curiosity? 
The default behaviour is to deploy on all available **worker** nodes.  However, it's been decided to leave the option for the admin to select the nodes they want, if they don't want all the worker nodes using Kata Containers.

> Why not e.g. enable the feature on all available nodes but use regular nodeSelector for steering?
That would simplify our code, indeed. Shall we consider removing the option from the operator?

> Is it because it fully takes over the crio backend and so hence all containers on that node are kata? That seems unfortunate if true.
No, that's not the case, at all.  A note capable of doing Kata Containers should not, and will not, lose its ability to do "vanilla" containers.

@cgwalters, have I addressed all your concerns?

enhancements/kata-containers/kata-containers-dev-preview.md
fidencio: That's interesting, and I wonder whether the minimalist QEMU we plan to use for the extensions could also benefit CoreOS Assembler.

Do you mind if we switch this conversation to the Slack, at some point soon, to try to sync on the usages and then, maybe, come up with a full document on how we plan to use QEMU as part of RHCOS, and hopefully agree in the minimal QEMU that could attend yours (CoreOS Assembler) and ours (Kata Containers) needs?

cgwalters: Note that the qemu we use to build (and test) RHCOS (and FCOS) is included in the coreos-assembler container; it has nothing to do with the host, which is the opposite of kata.  It's closer to KubeVirt in a way except this usage is special in that we're not trying to run production VMs that e.g. expose networking.  We don't require any privileges, just KVM available.

I have a blurb on this here https://hackmd.io/8YCMjP_JTzu-vp7PrOjT0g that I need to stick somewhere.

cgwalters: We don't need a minimal qemu really - our container image is already enormous anyways (2.5G+).

What may be good to share though is e.g. Go bindings for interacting with qemu - I touched on that in https://github.com/coreos/coreos-assembler/pull/1336#issuecomment-611759775

enhancements/kata-containers/kata-containers-dev-preview.md
cgwalters: > The default behaviour is to deploy on all available worker nodes. 

Sure, seems reasonable.  In a compact cluster the control plane are workers too so this will apply there.

> That would simplify our code, indeed. Shall we consider removing the option from the operator?

I don't have a really strong opinion on this but if having admins steer the pods with regular selectors works then why not use that?  Less API surface for you to maintain, and feels more native to admins, etc.

enhancements/kata-containers/kata-containers-dev-preview.md
```diff
@@ -0,0 +1,276 @@
...
+
+Via the CR it will be possible to select a subset of worker nodes.
+
+For deploying binaries on the host the current idea is to use a container image that can be mounted on the host level.  
+
+The goal is to be able to install the operator via OperatorHub.
+
+The Operator will:
+- Create a crio drop-in config file via machine config objects;
+- Automatically select the payload image that contains correct version of the kata binaries for the given version of the OpenShift;
```
cgwalters, line 146: In some private discussion it turns out there's a lot more detail behind this bit that we'll want to elaborate on.  From what I can see I lean towards having the MCO install this extra content as an extension (RPM) too.

This also relates to https://github.com/coreos/fedora-coreos-tracker/issues/354 etc.

enhancements/kata-containers/kata-containers-dev-preview.md
```diff
@@ -0,0 +1,276 @@
...
+- Configure a `RuntimeClass` called `kata` which can be used to deploy workload that uses kata runtime;
+- Create a machine-config object to enable the `qemu-kiwi` RHCOS extension.
```
cgwalters, line 148: I'd say instead have single MC object `50-kata-containers-worker` or so that does both this and the crio config to make it more atomic.  Otherwise there's a race where the MCO might try to roll out just the crio change but not the extension, etc.

fidencio:
In what started as a non related conversation with @cgwalters (https://coreos.slack.com/archives/CK1AE4ZCK/p1614974860056400), we agreed that more discussions have to happen, mainly with regards how Kata Containers will be delivered to the OpenShift customers (as in, whether it'll be part of the Extensions, whether it'll be binaries being copied by the Kata Operator, etc).

Based on that, this PR will be updated accordingly, probably by the beginning of the next week.

enhancements/kata-containers/kata-containers-dev-preview.md
```diff
@@ -0,0 +1,276 @@
...
+
+The `RuntimeClass` and payload image names will be visible in the CR status.
+
+#### KataContainers Operator Goals
+- Create an API which supports installation of Kata Runtime on all or selected worker nodes
+- Configure CRI-O to use Kata Runtime on those worker nodes
```
cgwalters, line 154: Is there more to this than just the MC?

enhancements/kata-containers/kata-containers-dev-preview.md
```diff
@@ -0,0 +1,276 @@
...
+- Installation of the runtimeClass on the cluster, as well as of the required components for the runtime to be controlled by the orchestration layer.
+- Updates the Kata runtime
+- Uninstall Kata Runtime and reconfigure CRI-O to not use it.
```
cgwalters, line 157: If done via extensions, this would all be handled by MCO reconciliation.   See also https://blog.verbum.org/2020/08/22/immutable-%E2%86%92-reprovisionable-anti-hysteresis/

I think more broadly, if everything goes via MCO functionality then e.g. node changes will be fully coordinated.  Having more host daemonsets etc. can create synchronization issues.

enhancements/kata-containers/kata-containers-dev-preview.md
```diff
@@ -0,0 +1,276 @@
...
+  runtimeClassName string `json:"runtimeClassName"`
+
+  // +required
+  kataImage string `json:"kataImage"`
+
+  // TotalNodesCounts is the total number of worker nodes targeted by this CR
+  TotalNodesCount int `json:"totalNodesCount"`
+
+  // CompletedNodesCounts is the number of nodes that have successfully completed the given operation
+  CompletedNodesCount int `json:"CompletedNodesCount"`
```
cgwalters, line 200: So if we fully use the MCO, then all this stuff is covered by the MCO objects, specifically `machineconfigpool`.

enhancements/kata-containers/kata-containers-dev-preview.md
fidencio: @cgwalters, I don't fully understand the question.
"Configure CRI-O to use Kata Runtime on those worker nodes" is just a drop-in file, performed by: https://github.com/openshift/sandboxed-containers-operator/blob/fbbc74e21e2fc4f3d0874dd201513a801a1a333c/controllers/openshift_controller.go#L281-L362

As I'm not sure whether this is what you asked, mind to rephrase the question, with more details is possible, so I can try to answer?

(by the way, sorry if I end up missing some details, the person who majoritarily worked on the Operator side is on paternity leave and I'm taking this over from him).

fidencio: Oh, wait, I'm wrong, I'm thinking only about the RPM sides.
"Configure CRI-O to use Kata Runtime" also contains:
* applying RBAC policies;
* defining the CRDs;
* creating the "kata" runtime class;

enhancements/kata-containers/kata-containers-dev-preview.md
fidencio: When you say "everything goes via MCO functionality" you mean ... just forget the operator and let teach the MCO to deal with Kata Containers' needs, entirely?

While I think that's a really good idea, I don't think it's do-able for 4.8 timeframe.

cgwalters: Ah no, I just mean that the operator installs machineconfigs, rather than reaching out to each node individually.  This is a pattern followed by for example the cluster-node-tuning-operator.

fidencio: Oh, that we do. :-)

enhancements/kata-containers/kata-containers-dev-preview.md
fidencio: @cgwalters, I've rewritten this PR, almost entirely, to cover what we current do, what we could do, etc.

Please, whenever you have some cycles, take a look at the changes.
Also, please, help me to loop in all the needed parts that can help to decide whether we could go for having the Kata Containers RPMs **also** as part of the RHCOS extensions (even with those being ~100MB, installed size).

fidencio:
@bpradipt, would you mind to take a careful look on the bits describing the operator?  That's one part of the project I barely touched and I don't confident to ensure the text matches reality.

fidencio:
@zanetworker, I will need your help to ensure this aligns with your vision.

fidencio:
@mrunalp, I really would love your input, mainly as things changed **a lot** from the time when the original PR was created, till what we currently have nowadays.

jackevans43:
@fidencio I'd love to use this feature to isolate less trustworthy / higher risk (eg. Internet facing) workloads - do you think this is worth adding? For these workloads, Firecracker would be a great, could this be added in the future?

I presume NetworkPolicy be applied on the host, outside of the VM (ie. provides a higher level of assurance)?

fidencio:
> @fidencio I'd love to use this feature to isolate less trustworthy / higher risk (eg. Internet facing) workloads - do you think this is worth adding?

I think this is covered by:
```
- any other administrative privileges above and beyond what is secure in a
  shared kernel environment (regular runc).
```
Or would you prefer have it explicitly set there?

> For these workloads, Firecracker would be a great, could this be added in the future?

Right now we prefer to focus on the VMM we have a huge team working on, and maintaining it, which is QEMU.
I'd be interested, really interested, on learning about possible use cases you see Firecracker being used that QEMU couldn't.  This would help the QEMU community to improve, we (as Sandboxed Containers) to see whether we can expand what we cover, and the end users to take a decision.

I don't think this proposal is the right place to discuss this, but maybe https://github.com/openshift/sandboxed-containers-operator/ could be.

I'd suggest opening an issue there, then we can start the discussion and move to the another (more appropriated?) place if needed.

Does this sound reasonable?

> 
> I presume NetworkPolicy be applied on the host, outside of the VM (ie. provides a higher level of assurance)?

Please, open us an issue and we can discuss it there. :-)
By the way, I hope I don't come across as "rude", I just want to be sure that such important conversation gets properly tracked, and properly replied (and here it may be confusing for all the parts involved).

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,437 @@
...
+   installing packages (in case any other operator decides to do the same);
+4. Updates / Removal may be more complicated than having the RPMs as part of
+   the machine-os-content.
+
+
+For additional details about RHCOS extensions in general in: https://github.com/openshift/enhancements/pull/317
+
+##### Use RHCOS extensions for QEMU and its dependencies, copy Kata Containers files onto the host
+
+In this option we completley rely on RHCOS Extensions for the QEMU and its
```
bpradipt, line 221: nit: completley => completely

fidencio: Fixed, thanks!

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,437 @@
...
+
+## Proposal - main development efforts
+
+### Overview
+Kubernetes provides support for RuntimeClasses. RuntimeClass is a feature for
+selecting the container runtime configuration. The container runtime
+configuration is used to run a Pod’s containers.
+
+CRI-O today comes out of the box with a runc as the default runtime. CRI-O also
+supports RuntimeClasses and using this configuration, it will support Sandboxed
```
mrunalp, line 122: s/configuration/feature

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,437 @@
...
+   Products;
+4. We rely on an already existing mechanism to install all of the RPMs - QEMU,
+   QEMU dependencies, and Kata Containers - which are not part of the base
+   image;
+5. There's no chance of hitting issues related to parallel installing packages,
+   as the whole package installation is handled by the MCO;
+6. Updates / Removal become trivial as they are also handled by the MCO.
+
+**Cons:**
+1. 120MB extra - coming from QEMU, QEMU dependencies, and Kata Containers - as
```
mrunalp, line 272: We can capture the rpm size in the payload for qemu-kiwi and kata as it should be smaller like we discussed.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,437 @@
...
+
+If we take telcos for example, there are a number of reasons they require
+isolated workloads:
+
+1. **Telco CNF deployments** - As part of 5G/NFV telco deployments there is a
+   gradual migration from physical networking boxes to CNFs (container network
+   functions). Some of these CNFs require root access while others could
+   potentially create a threat on the operators cloud bypassing existing
+   container isolation mechanisms. For addressing those issues a hardened
+   isolated container solution is required;
```
mrunalp, line 67: nit: replace ; with .

mrunalp (changes requested):
Can the enhancement be rebased to the latest template here - https://github.com/openshift/enhancements/blob/master/guidelines/enhancement_template.md?

enhancements/kata-containers/kata-containers-dev-preview.md
mrunalp: There is scheduling support for runtime classes that we can make use of. The enhancement should mention that.
https://v1-16.docs.kubernetes.io/docs/concepts/containers/runtime-class/#

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+- Kata Containers aims to run **container images** on OpenShift in an
+  isolated manner using virtualization tools (uses Runtime Class resource to
+  choose Kata Containers only for specific workloads which require kernel
+  level isolation).
+  
+## Motivation
+
+The main motivation to bring sandboxed containers to OpenShift is to provide
+an extra layer of isolation for fields that would commonly require that, such
+as Telco companies.
```
mrunalp, line 71: nit: Can be modified to more security conscious users instead.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in
+      [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Sumary
+
+Sandboxed containers will be integrated into OpenShift to provide the ability
+to run kernel isolated containers for any workload which requires:
+- Custom kernel tuning (sysctl, scheduler changes, cache tuning, etc).
```
mrunalp, line 46: nit: I would suggest moving these down to motivation section.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+## Motivation
+
+The main motivation to bring sandboxed containers to OpenShift is to provide
+an extra layer of isolation for fields that would commonly require that, such
+as Telco companies.
+
+### Goals
+
+- Provide a way to enable running user workloads on sandboxed containers on an
+  OpenShift cluster;
```
mrunalp, line 76: nit: replace ; by .

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+- Clusters with Sandboxed Container workloads should support upgrades;
```
mrunalp, line 77: nit: replace ; by .

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+### Non Goals
+
+- Running workloads using a different kernel than the one used on the
+  OpenShift node
+  + Although sandboxed containers are capable of supporting running a
+    different kernel then the one used on the OpenShift node, such features
+    will not be supported in order to simplify the development cycles and
+    testing efforts of this project;
+- Running OpenShift control plane and core operators on Sandboxed Containers
+  + These will continue to run on runc.
```
mrunalp, line 90: native linux OCI runtime runc

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+processes. In order for devops to deploy new features on production
+environments there is a need for high isolation of the deployed workloads in
+order to control and revert such changes.
+
+#### Story 3 - Hardware Vendors
+
+Many hardware vendors require custom kernel parameters at boot, custom sysctl
+variables, etc. They want to use containers as a convenient packaging format
+for their entire application, but still need access to lower level tuning for
+portions of their applications.
```
mrunalp, line 122: nit: Could add a use case for builds that we don't have to tackle right away but is enabled by this work.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+
+### Implementation Details/Notes/Constraints
+
+Kubernetes provides support for [Runtime
+Class](https://kubernetes.io/docs/concepts/containers/runtime-class/), which
+will be used to add, and select, the container runtime to be used for a
+specific workload.
+
+CRI-O, Today, comes out of the box with `runc` as the default runtime, but
```
mrunalp, line 131: s/Today/today

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+
+For addressing the second issue of deploying sandboxed containers and
+configuring the cluster to use it, [OpenShift sandboxed containers
+operator](https://github.com/openshift/sandboxed-containers-operator) has been developed.
+
+#### Risks and Mitigations
+
+While working on the integration of Kata Containers, which is used under the
+hood by sandboxed containers, on to OpenShift, it's been realised that
+metrics won't work as expected during the tech preview.
```
mrunalp, line 156: won't full work

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+
+This happens because OpenShift strongly relies `cAdvisor` for its monitoring
```
mrunalp, line 158: strongly relies -> uses

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+tools, and `cAdvisor` is not capable of collecting the **container**
+information of a Kata Containers, which is running inside the a VM.  **pod**
+information, though, which includes the `containerd-shim-kata-v2`, the
+`QEMU`, and the `virtiofs` processes, is properly collected.
+
+This leads to switching to `CRI stats` in the future, which is part of an
```
mrunalp, line 164: We plan to switch to ..

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+effort being lead by Peter Hunt (@haircommander) and that can be seen
+[here](https://github.com/kubernetes/enhancements/pull/2364).
+
+## Design Details
+
+Described below, you can see the `CustomResourceDefinition (CRD)` implemented by the sandboxed containers operator.
+
+```yaml=
+---
+apiVersion: apiextensions.k8s.io/v1
```
mrunalp, line 174: Instead of this I suggest just showing how the CRD can be used.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+Finally, it's important to mention that altough there's a **strong**
+preference on using CoreOS Extensions for actually installing Kata
+Containers, QEMU, and QEMU dependencies on the worker nodes, it still
+needs an approval from the development team.  A second option, which is
+still doable and should not damage the cluster in any possible way is
+simply copying the kata-containers binaries to the worker nodes, and relying
+on CoreOS Extensions only for QEMU, and QEMU dependencies.  Those options
+are extensively discussed as part of the next section.
+
+### Open Questions
```
mrunalp, line 461: Instead of Open Questions, I suggest creating a sub section for `Installation of kata dependencies and artifacts on the node`

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+       custom-kata1: test 
+```
+
+Thanks to the Kubernetes Runtime Class [scheduling
+support](https://v1-16.docs.kubernetes.io/docs/concepts/containers/runtime-class/#scheduling)
+present since `v1.16.0`, we ensure that when a workload requires a
+specific `RuntimeClass`, it will be scheduled on nodes that
+supports the specific `RuntimeClass`.
+
+Finally, it's important to mention that altough there's a **strong**
```
mrunalp, line 452: I would either incorporate anything here to the `Installation` section suggested below or drop this.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+Containers, QEMU, and QEMU dependencies on the worker nodes, it still
+needs an approval from the development team.  A second option, which is
+still doable and should not damage the cluster in any possible way is
+simply copying the kata-containers binaries to the worker nodes, and relying
+on CoreOS Extensions only for QEMU, and QEMU dependencies.  Those options
+are extensively discussed as part of the next section.
+
+### Open Questions
+
+There's yet no consensus on how the sandboxed containers artifacts will be
```
mrunalp, line 463: I would drop this and instead frame it as extensions as the proposed approach and copying as the alternative.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+   intent for.
+3. QEMU RPMs have been tested / validated using an identical environment.
+4. No additional work with regarding to re-packaging those in another form
+   other than the RPMs which is the way they are consumed by other Layered
+   Products.
+5. We rely on an already existing mechanism to install part of the RPMs, QEMU
+   and its dependencies, which are not part of the base image.
+
+**Cons:**
+1. 5.5MB extra to be download - coming from QEMU and its dependencies - as
```
mrunalp, line 544: I would drop this as a con and move it pros for a 5.5 mb artifact.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+   part of the RHCOS Extensions, for all users relying on extensions; ~20MB
+   installed for those using the QEMU extension.
+2. 130MB extra - coming from Kata Containers binaries - copied onto the host,
+   for those who'd be using the Sandboxed Containers,
+3. The sandboxed containers operator still does a work that, in theory,
+   should be restricted to the MCO.
```
mrunalp, line 550: Mention challenges in upgrade with this model

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+### Test Plan
+
+sandboxed containers team is already in touch with OpenShift QE and is
+currently performing tests related to:
+* the operator installability.
+* basic lifecycle of pods using sandboxed containers.
+* command-line tools integration.
+* logs, events, and metrics visiblity.
+* terminal availability.
+* scalation (both vertical and horizontal).
```
mrunalp, line 573: Scaling

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+* OpenShift: Follow up the PR opened for the `machine-config-operator` to
+  have `sandboxed-containers` added as an extension.  For reference, see:
+  https://github.com/openshift/machine-config-operator/pull/2456
+* sandboxed containers operator: Adapt our operator to deploy those RPMs
+  using the CoreOS Extensions.
+* sandboxed containers operator: Adapt our RPMs to be installed in
+  `/opt/sandboxed-containers/`
+* sandboxed containers operator: Adapt our operator to copy the binaries
+  on to the RHCOS node.
+
```
mrunalp, line 563: We need a section to cover how we will restrict access to kata runtime class. Mention that SCC is being discussed but we may end up using a custom solution for kata.

mrunalp: @haircommander can you help @fidencio with this section?

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+Extensions](https://github.com/openshift/enhancements/pull/317).
+in OpenShift we will present - in detail as part of the next section - the
+current approach, and a few options of what could be done to improve the
+currently have.
+
+For addressing the second issue of deploying sandboxed containers and
+configuring the cluster to use it, [OpenShift sandboxed containers
+operator](https://github.com/openshift/sandboxed-containers-operator) has been developed.
+
+#### Risks and Mitigations
```
mrunalp, line 152: We should also describe any performance related risks here.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,620 @@
...
+
+Finally, it's important to mention that altough there's a **strong**
+preference on using CoreOS Extensions for actually installing Kata
+Containers, QEMU, and QEMU dependencies on the worker nodes, it still
+needs an approval from the development team.  A second option, which is
+still doable and should not damage the cluster in any possible way is
+simply copying the kata-containers binaries to the worker nodes, and relying
+on CoreOS Extensions only for QEMU, and QEMU dependencies.  Those options
+are extensively discussed as part of the next section.
+
```
mrunalp, line 460: Can we cover the pod overhead for kata?

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
zanetworker: I think the statement as makes sense as sandboxed containers provide additional isolation. Also, that would imply that security is not a default (which it is) without it.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
zanetworker: I think that can be left out of the proposal and added later when there is consensus on how that would be implemented given upstream activities. (We can later add the section in a post PR with the details once they are there?)

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,374 @@
...
+Apart from those, there's the plan to onboard the project to the OpenShift
+CI, ensuring then we don't regress over time.  It's important to mention that
+Kata Containers upstream project was already taken onboard of OpenShift CI,
+as part of early stages of this enhancement.
+
+### Graduation Criteria
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
```
mrunalp, line 342: Supported upgrades (instead of just testing)

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,374 @@
...
+
+It's still not decided whether `kata-containers` will be authorized to
+become part of CoreOS Extensions, or if **only** its dependencies (`QEMU`
+and its depencies) will be.
+
+### Test Plan
+
+sandboxed containers team is already in touch with OpenShift QE and is
+currently performing tests related to:
+- The operator installability.
```
mrunalp, line 324: nit: `operator features` or `operator functionality`

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,374 @@
...
+- Basic lifecycle of pods using sandboxed containers.
+- Command-line tools integration.
+- Logs, events, and metrics visiblity.
+- Terminal availability.
+- Scaling (both vertical and horizontal).
+- CPU and memory restrictions.
+- Pods behaviour based on node cordon/drain scenarios.
+
+Apart from those, there's the plan to onboard the project to the OpenShift
```
mrunalp, line 333: Describe the e2e jobs that are/will be run and how often.

enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
```diff
@@ -0,0 +1,374 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Sumary
+
+Sandboxed containers will be integrated into OpenShift to provide the ability
+to run kernel isolated containers.  Under the hood, the feature will rely on
+[Kata Containers](https://katacontainers.io/), an Open Source project working
+to provide a stronger workload isolation using hardware virtualization
+technologies as a second layer of defense.
```
mrunalp, line 47: What about `additional` instead of `second`?

fidencio:
@mrunalp, the last requests have been addressed. Thanks for the review.

PR #682, opened 2021-03-09
alerting as a feature

dofinn:


enhancements/alerting/alerting-as-a-feature.md
```diff
@@ -0,0 +1,302 @@
...
+- [ ] Test plan is defined
+- [x] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Definitions
+
+Caused Based Alert: Describes the direct cause of an issue. Example: etcdMemoryUtilization @ 100%
+
+Symptom Based Alert = Describes a symptom who's source is a cause based alert. Example: etcd connection latency > 200ms. This may be caused by etcdMemoryUtilization (or not?)
```
wking, line 38: How high up the stack you go before a cause-based alert becomes a symptom-based alert depends on your particular SLAs and SLOs.  [The Prom docs][1] link Rob Ewaschuk's excellent [philosophy doc][2], talking about [alerting from the spout][3].  If you're managing etcd, etcd request latency is a reasonable spout.  If you're managing the cluster, where folks can only get at etcd through the Kube API servers, etcd latency becomes a cause-based alert explaining one possible component of your spout Kube API latency alert.

Besides case-based and symptom/spout alerting, I think it's also useful to have diagnostics that look at firing alerts and other in-cluster status and say "ah, the cluster of issues you're seeing look like `$ROOT_CAUSE`".  The [insights folks][4] have [some tools in this space][5], but currently they are internal and don't run in-cluster.

[1]: https://prometheus.io/docs/practices/alerting/
[2]: https://docs.google.com/document/d/199PqyG3UsyXlwieHaqbGiWVa8eMWi8zzAn0YfcApr8Q/edit
[3]: https://docs.google.com/document/d/199PqyG3UsyXlwieHaqbGiWVa8eMWi8zzAn0YfcApr8Q/edit#heading=h.rvrk9gcasjzh
[4]: https://github.com/openshift/insights-operator
[5]: https://www.openshift.com/blog/openshift-insights-for-openshift-cluster-manager

enhancements/alerting/alerting-as-a-feature.md
```diff
@@ -0,0 +1,302 @@
...
+2. Symptom Based alerts provide SRE's with user perspective alerting at scale(worth getting out of bed for). But this perspective is not easily illustrated to component engineering teams. 
+
+So how can can these efforts work together in a cohesive life-cycle? 
+
+### Goals
+
+#### Provide two life cycles that deliver Alerting as a feature. 
+
+* Alerting standards are:
+  * defined by engineering, monitoring team and SREs (SRE assist on critical severity only).
```
wking, line 65: Alerts are aimed at cluster admins, which means SREs + customers.  There should absolutely be SRE and customer feedback on the utility and actionability of all alerts, not just `critical` alerts.  But the `critical` alerts are the loudest, so that's the most impactful place for folks interested in improving alert quality in general but who only have limited time to invest.

enhancements/alerting/alerting-as-a-feature.md
dofinn: > How high up the stack you go before a cause-based alert becomes a symptom-based

Start from as a high up the stack as you can go then. We should be alerting from the perspective of outside in, not inside out.

enhancements/alerting/alerting-as-a-feature.md
```diff
@@ -0,0 +1,302 @@
...
+* Alerting standards are:
+  * defined by engineering, monitoring team and SREs (SRE assist on critical severity only). 
+  * consumed by SREs + customers
+  * Life-cycled by component teams, monitoring teams and SRE. 
+
+* SLOs coupled with symptom based alerts are:
+  * defined by monitoring team and SREs.
+  * Implemented by SRE
+  * Lifecycled by monitoring team and SRE. 
+  * Developed, tested and validated by SRE and eventually committed back to the product.
```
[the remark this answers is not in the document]
dofinn: You have missed the point if you think its a standards vs SLO based thing. They work in paralell. The component experts help drive the standards the specific causes of their components, while SRE helps drive the highest abstractions of symptom based alerts... potentially back into the product.

enhancements/alerting/alerting-as-a-feature.md
```diff
@@ -0,0 +1,302 @@
...
+
+## Motivation
+
+OCP as a product does not have built-in [alerting standards](https://github.com/openshift/enhancements/pull/637), nor built-in SLOs driven by symptom or caused based alerting methodologies. This may be acceptable when managing a small number of clusters but at scale (+100) it breaks down. As an SRE of a cluster, you are given alerts and documentation on how to [manage them](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.6/html/monitoring/managing-alerts). This should come "out of the box" to support a natively support fleets.
+
+There are two current efforts in the space at the moment:
+
+* [alerting standards](https://github.com/openshift/enhancements/pull/637) -> this will have component teams own their own components alerts as they are they subject matter experts (caused based alerts). 
+* SLO driven [symptom based alerting](https://docs.google.com/document/d/199PqyG3UsyXlwieHaqbGiWVa8eMWi8zzAn0YfcApr8Q/edit#heading=h.1upja8jlnnwp) is being explored by OSD/SRE. -> this will enable meaningful alerting from a customer perspective in an SLA environment.
```
RiRa12621, line 51: We need to find a good place for this to live. The OEPs are public and for all of openshift so docs referred to should probably be available publicly as well.

enhancements/alerting/alerting-as-a-feature.md
RiRa12621: +1
This can probably be reworded to incorporate sres generally giving guidance on the alert severity

enhancements/alerting/alerting-as-a-feature.md
bwplotka: Side note, does this separation of alerts matter in the overall proposal? I think it's fair to say that alerting should be a feature of Openshift yes, but the content or alerts could be of a choice of cluster admins really, plus it depends what is being run and with what SLA.

The reason is that cause/symptom alert statements are controversial as noticed. As you mentioned the write up about this from Google SRE here: https://docs.google.com/document/d/199PqyG3UsyXlwieHaqbGiWVa8eMWi8zzAn0YfcApr8Q/edit#heading=h.6ammb5h32uqq is excellent material.
The rule is simple:  `don't write cause-based paging rules for symptoms you can catch otherwise.`

We need to also identify the goal of alerting here. It is not for drilling down (!). It's for notifying you that users are or will be  impacted(!)

> That said, if your debugging dashboards let you move quickly enough from symptom to cause to amelioration, you don't need to spend time on cause-based rules anyway.

enhancements/alerting/alerting-as-a-feature.md
```diff
@@ -0,0 +1,302 @@
...
+
+Alerting as a feature is a holistic composition of [alerting standards](https://github.com/openshift/enhancements/pull/637) coupled with built-in alerting methodologies (none, symptom based and caused based) with an option to deliver SLOs for a single and/or fleet of clusters. 
+
+## Motivation
+
+OCP as a product does not have built-in [alerting standards](https://github.com/openshift/enhancements/pull/637), nor built-in SLOs driven by symptom or caused based alerting methodologies. This may be acceptable when managing a small number of clusters but at scale (+100) it breaks down. As an SRE of a cluster, you are given alerts and documentation on how to [manage them](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.6/html/monitoring/managing-alerts). This should come "out of the box" to support a natively support fleets.
+
+There are two current efforts in the space at the moment:
+
+* [alerting standards](https://github.com/openshift/enhancements/pull/637) -> this will have component teams own their own components alerts as they are they subject matter experts (caused based alerts).
```
bwplotka, line 50: I would be careful here, it sounds like a path to noisy alerts and alert burden. Instead of alerting we might want to invest in tools and practices to drill into such components. As cluster admin, responsible for bigger service (e.g cluster) you should be able to alert on symptoms and one instance of etcd being slow is not a symptom, it might not even impact cluster at all! It is a potentially misleading piece that you should have dashboard for and not screaming alerts 🤔

bwplotka (commented):
Thanks for this proposal, it is a nice starting point, but I am missing technical details on what example intput and output we expect. What APIs you want to have, what data alerts should be sourced from, how we deploy them, who can change them etc.

Do we have plan for this? 🤗

enhancements/alerting/alerting-as-a-feature.md
dofinn: +1 to the above thoughts. Thats why i proposed alerting profiles to enable admins/SRSs to either:
a) configure alerting from scratch
b) configure alerting with caused based alerts (as a starting point)
c) configure alerting with symptom based alerts (as a starting point)

enhancements/alerting/alerting-as-a-feature.md
dofinn: +1. I was trying to illustrate that critical severity involvement should be heavily encouraged, but never as a blocker.

dofinn:
> Thanks for this proposal, it is a nice starting point, but I am missing technical details on what example intput and output we expect. What APIs you want to have, what data alerts should be sourced from, how we deploy them, who can change them etc.
> 
> Do we have plan for this?

Thanks for feedback.

I think i would like to address these finer issues after buy in of the propsed lifecycling and outcomes.

enhancements/alerting/alerting-as-a-feature.md
```diff
@@ -0,0 +1,302 @@
+---
+title: Alerting as a feature
```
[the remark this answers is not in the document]
dofinn: Its built in, its not an after-thought. OCP no longer comes with just alerts, it comes with alerts and out of the box methodology  that can be deployed as easy as a pod for an end user.

enhancements/alerting/alerting-as-a-feature.md
```diff
@@ -0,0 +1,302 @@
...
+
+## Definitions
+
+Caused Based Alert: Describes the direct cause of an issue. Example: etcdMemoryUtilization @ 100%
+
+Symptom Based Alert = Describes a symptom who's source is a cause based alert. Example: etcd connection latency > 200ms. This may be caused by etcdMemoryUtilization (or not?)
+
+## Summary
+
+Alerting as a feature is a holistic composition of [alerting standards](https://github.com/openshift/enhancements/pull/637) coupled with built-in alerting methodologies (none, symptom based and caused based) with an option to deliver SLOs for a single and/or fleet of clusters.
```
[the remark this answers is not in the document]
dofinn: - As a feature of the cluster and extensibly a fleet
- SRE/sAdmins
- What data are you referring to exactly? 
- Built for single,multi and fleet scale.

enhancements/alerting/alerting-as-a-feature.md
dofinn: The alerting standards define how component teams will create alerts for their components. They are the experts in their domains. These alerts should aid in troubleshooting a caused based symptom of which an admin/sre can overlay on the cluster. alerting standards being the caused based alerting layer that will live under symptom. IMO this should be life cycled and developed to reduce MTTR.

PR #683, opened 2021-03-09
Insights Operator pulling and exposing data from the OCM API

tremes:


enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,87 @@
...
+
+Risk: Insights Operator is unable to expose/update the data in the OpenShift API
+
+Mitigation: We can introduce a new state in the Insights Operator (e.g "ObsoleteSubscriptionData"), but a new alert would be probably better.  
+
+## Design Details
+
+### Open questions
+- OCM API authentication - can we use existing "pull-secret" token to access the OCM API as well? If not, what are the alternatives?
+- Where to expose the new data? New secret in the "openshift-config" namespace? Are there some naming conventions for the secret?
```
sbose78, line 79: I would go for a secret named `simple-content-access` in the `openshift-config` namespace.

sbose78: @tremes I would go for the `openshift-config-managed` namespace since human users/admins typically place resources in the  `openshift-config` namespace.

sbose78 (changes requested):

enhancements/insights/pulling-data-from-ocm.md
tremes: Updated

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,95 @@
...
+## Design Details
+
+### Authorization
+
+Insights Operator is able to pull the data from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config` namespace)
+
+### Data in API
+
+The SCA certificate is available via the `simple-content-access-cert` secret in the `openshift-config-managed` namespace
```
sbose78, line 85: Sorry, I'm going to request another change.  Based on https://github.com/openshift/enhancements/blob/master/enhancements/subscription-content/subscription-injection.md

could we use  `etc-pki-entitlement`

tremes: For the name of the secret right? From the link I guess the namespace will be `openshift-config` right? I see there:
```
entitlements:
   - name: etc-pki-entitlement
     namespace: openshift-config
```

tremes: Updated. @sbose78 can you please check?

sbose78: The name would be `etc-pki-entitlement`
The namespace would be `openshift-config-managed` since this configuration would be placed using an operator and not a human admin. 

When the https://github.com/openshift/enhancements/blob/master/enhancements/subscription-content/subscription-injection.md was written, the current approach of placing the secrets on the cluster using the Insights operator wasn't in consideration and hence `openshift-config` was used.

tremes: OK thanks. Let me return it back to `openshift-config-managed` then.

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,95 @@
...
+
+Insights Operator is able to pull the data from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config-managed` namespace)
+
+### Data in API
+
+The SCA certificate is available via the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace
+
+### Open questions
+- How often should the Insights Operator query the OCM API? Every 4 hours? Do we always need to download the full SCA cert data?
```
sbose78, line 88: As of now, we should go with the full cert data and consider enhancing that in future.
Frequency: 8 hours is good enough. As far as I remember from conversations, it takes longer that that to expire.

tremes: OK thanks. I'll update

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,95 @@
...
+
+Mitigation: Introduce a new state in the Insights Operator (e.g "SCADataDegraded") and
+create a new alert based on this new state.
+
+## Design Details
+
+### Authorization
+
+Insights Operator is able to pull the data from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config-managed` namespace)
```
tremes, line 81: I think we can add that IO needs to provide cluster ID. I would expect the OCM API url something like:
`cloud.redhat.com/api/scadata/clusters/<cluster-id>`
Simply that the cluster ID would be part of the URL.

tremes: @sbose78 WDYT pls?

sbose78: .. and cluster-id is something insights already has access to?

tremes: yes

sbose78: > I think we can add that IO needs to provide cluster ID. I would expect the OCM API url something like:
cloud.redhat.com/api/scadata/clusters/<cluster-id>
Simply that the cluster ID would be part of the URL.

Then, yes  👍

tremes: Updated. Thanks.

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,96 @@
...
+### Risks and Mitigations
+
+#### OCM API is down
+
+Risk: OCM API is down or doesn't provide up to date data.
+
+Risk: Insights Operator is unable to expose/update the data in the OpenShift API
+
+Mitigation: Introduce a new state in the Insights Operator (e.g "SCADataDegraded") and
+create a new alert based on this new state.
```
tremes, line 74: I am not sure about the alert, because I don't think it will be possible to take any action on this alert. 

It might make sense to introduce a config option (in the insights operator) that would disable the periodic pulling from the OCM. That might be handy for CI, dev clusters (simply clusters without connection to the OCM)

sbose78: > I am not sure about the alert, because I don't think it will be possible to take any action on this alert.

Would there be any other way to know that pulling is failing? 🤔

sbose78: > It might make sense to introduce a config option (in the insights operator) 

right, needn't be explained in the enhancement proposal though.

tremes: > right, needn't be explained in the enhancement proposal though.

OK

> Would there be any other way to know that pulling is failing? 

Not sure. Could the potential alert be applicable?? I can't think of anyother way of propagating this problem... :-(

sbose78: Let's have the alert. There's always a way to silence it if the admin finds it annoying.

sbose78:
Does insights Operator have access to the org ID and/or cluster Id?

tremes:
@sbose78 The Insights Operator does have access to the cluster ID, but from what I know it doesn't have access to the org ID.

sbose78 (approved):

sbose78:
Thanks Ivan, I think we have all the approvals needed at this stage.

sbose78:
Hi @smarterclayton Would you be able to review this, and potentially merge this ?

tremes:
@sbose78 Thanks for pushing it, but maybe it would be good if someone from the OCM can review too. @abhgupta pls

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,96 @@
+---
+title: pulling-and-exposing-data-from-ocm
+authors:
+  - "@tremes"
+reviewers:
+  - "@sbose78"
+  - "@0sewa0"
```
Sergey1011010, line 7: Marcell doesn't work with us anymore ;)

petli-openshift:
LGTM from my perspective.

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,96 @@
...
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement will enable the Insights Operator to pull the data (SCA certs)
```
dhellmann, line 33: What "data" is being pulled?

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,96 @@
...
+## Motivation
+
+Users could consume RHEL content and container images using the RHEL subscription in the OpenShift 3.x.
+In the OpenShift 4, this is no longer possible because the Red Hat Enterprise Linux Core OS (RHCOS) does not
+provide any attached subscription. This enhancement is to provide users the Simple Content Access (SCA) certs
+from Red Hat Subscription Manager (RHSM).
+
+### Goals
+
+- Extend the Insights Operator config with an OCM API URL to be able to query the data
```
dhellmann, line 47: Why is it the responsibility of Insights Operator to do this?

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,96 @@
...
+#### OCM API is down
+
+Risk: OCM API is down or doesn't provide up to date data.
+
+Risk: Insights Operator is unable to expose/update the data in the OpenShift API
+
+Mitigation: Introduce a new state in the Insights Operator (e.g "SCADataDegraded") and
+create a new alert based on this new state.
+
+## Design Details
```
dhellmann, line 76: The rest of this file doesn't match the modern enhancement template. Please rebase this PR and fill in the missing sections.

enhancements/insights/pulling-data-from-ocm.md
tremes: I can't answer that. I was not present during the original decision. So perhaps @iNecas or @smarterclayton.

enhancements/insights/pulling-data-from-ocm.md
tremes: SCA (Simple Content Access) certificates from RHSM. These certs are required by OpenShift clusters to pull RHEL content for builds (from https://issues.redhat.com/browse/SDB-2197)

dhellmann: How will those certificates make it into a cluster in an environment where OCM is not present?

Does the insights operator use the certificates itself?

tremes: They won't make it into a cluster. No the Insights operator doesn't use or require them.

enhancements/insights/pulling-data-from-ocm.md
tremes: I added more sections. Can you please check?

tremes: Ohh now it's checked by the linter....OK I need to add more

enhancements/insights/pulling-data-from-ocm.md
mfojtik: @smarterclayton FYI :-) I would be interested as well to know why was this embedded into insights operator

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Risks and Mitigations
+
+#### OCM API is down
+
+Risk: OCM API is down or doesn't provide up to date data.
+
+Risk: Insights Operator is unable to expose/update the data in the OpenShift API
+
+Mitigation: Introduce a new state in the Insights Operator (e.g "SCADataDegraded") and
```
mfojtik, line 73: does this mean the clusters won't be able to upgrade if the OCM is down?

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Test Plan
+
+- `insights-operator-e2e-tests` suite can verify the SCA cert data
+  is available
+- Basic test of the validity of the SCA certs. Mount the `etc-pki-entitlement` secret and run e.g `yum install` in the container
+
+### Graduation Criteria
+
+This feature is planned as a technical preview in OCP 4.9 and is planned to go GA in 4.10.
```
mfojtik, line 100: does this mean the feature will be disabled by default and customers can opt-in using some feature gate?

enhancements/insights/pulling-data-from-ocm.md
iNecas: FYI, this effort has been made after consulting with and on direct recommendation from @smarterclayton: If I understood it correctly, the idea is for the insights operator to cover more the "value of RH subscription": where the entitlement certificates fit as well and it seemed there was no better place to put it in. Also, insights-operator already has a trusted connection established with cloud.redhat.com api, where the data are being pulled from.

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,133 @@
...
+### Authorization
+
+The Insights Operator is able to pull the data from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config-managed` namespace).
+
+The Insights Operator must provide a cluster ID as an identifier of the cluster.
+
+### Data in API
+
+The SCA certificate is available via the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace.
```
mfojtik, line 87: how is this certificate going to be injected into a pod?

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Update period
+- Insights Operator query the OCM API every 8 hours and downloads the full data provided
+
+### Test Plan
+
+- `insights-operator-e2e-tests` suite can verify the SCA cert data
+  is available
+- Basic test of the validity of the SCA certs. Mount the `etc-pki-entitlement` secret and run e.g `yum install` in the container
```
mfojtik, line 96: I believe this mostly is affecting builds? (80%+ use cases)... I would expect this to be part of the build API somehow.

enhancements/insights/pulling-data-from-ocm.md
iNecas: Also @dhellmann, @mfojtik  feel free talking to Siamak Sadeghianfar, that is the PM owning this feature AFAIK

mfojtik: @iNecas well, I think openshift controller manager operator might be a better fit for this as this feature will be mainly be used by builds, right? also the way this gets injected into build containers will probably become something in build API?

/cc @adambkaplan

enhancements/insights/pulling-data-from-ocm.md
tremes: No. This would be a new clusteroperator condition and `SCACertsExpired` is probably slightly better name.

enhancements/insights/pulling-data-from-ocm.md
tremes: I am not sure it's the goal of this proposal to answer this question. Do we need to answer it here? @sbose78 is working on "tooling" using this cert.

enhancements/insights/pulling-data-from-ocm.md
tremes: Sorry maybe this is too vaguely defined, but it will be the opposite. It's enabled by default and users/customers can disable it.

enhancements/insights/pulling-data-from-ocm.md
iNecas: @sbose78 is driving this functionality from the building perspective AFAIK (pulling him in).

iNecas: > I think openshift controller manager operator might be a better fit for this as this feature will be mainly be used by builds, right? 

I had the very same question to @smarterclayton when this idea first came and he felt quite convinced IO was the right place compared to all alternatives.

adambkaplan: SCA Certs (and entitlement certs in general) are a Red Hat subscription experience. Insights operator is the only core OCP component we have today that connects OpenShift to the rest of the Red Hat experience. Correct me if I'm wrong, but I don't think the insights operator provides any value/function without a Red Hat subscription. Openshift-controller-manager, on the other hand, does not need a Red Hat subscription to function.

Builds are a consumer of the SCA cert, but it is by no means the only workload that could find itself needing access to subscription content. There are other components in the works which are designed to mount the cluster's SCA cert into any workload, namely the shared resource/projected resource CSI driver.

enhancements/insights/pulling-data-from-ocm.md
adambkaplan: > How will those certificates make it into a cluster in an environment where OCM is not present?

I assume that these clusters would have a Satellite instance where the SCA can be obtained.
For a first pass I think this use case can be considered out of scope - admins can manually obtain the SCA from satellite, add it as a secret in the `openshift-config`, and then create the `Share` object to make it accessible in the cluster.

enhancements/insights/pulling-data-from-ocm.md
adambkaplan: > also the way this gets injected into build containers will probably become something in build API?

Correct, however through mechanisms that are more general purpose:

1. The projected resource CSI driver will allow the SCA cert to be accessed in the cluster through the `Share` custom resource.
2. The build API will be enhanced to allow `CSI` volumes to be mounted in.

enhancements/insights/pulling-data-from-ocm.md
adambkaplan: Keep in mind SRE toil. The operator should use an exponential backoff to retry updating the SCA cert from OCM, and an alert should only fire after a sufficient number of retries.

For comparison - the Samples Operator has an alerting mechanism if imagestreams fail to import. The alert only fires if there are failing imagestream imports after 2 hours, and the status doesn't result in the cluster being marked Degraded.

enhancements/insights/pulling-data-from-ocm.md
adambkaplan: The Insights Operator will also need to do the following:

1. Create a `Share` object for the projected resource CSI driver
2. Create a "viewer" ClusterRole that allows read access to the Share (get, list, watch)

Ideally we aggregate this cluster role - I would recommend aggregation to the k8s admin and edit roles so end users and our default service accounts can access it [1].

Pods will then need to mount the share via a CSI inline volume - see openshift/csi-driver-projected-resource for an example [2].

[1] https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles
[2] https://github.com/openshift/csi-driver-projected-resource/blob/master/docs/simple-example.md

enhancements/insights/pulling-data-from-ocm.md
adambkaplan: Ultimately yes, we'd want to put this in the build suite.

enhancements/insights/pulling-data-from-ocm.md
adambkaplan: Note that a lot of the mechanisms we need to get this all working together (projected resource CSI driver, builds with CSI inline volumes) are going to be true tech preview features - no upgrade, must opt in. I recommend the same - use a feature gate to enable this capability, add it to the `TechPreviewNoUpgrade` feature set.

@deads2k @s-urbaniak I don't think we can get the CSI bits to GA until we have better SCC/admission control around CSI inline volumes.

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,133 @@
...
+This feature is planned as a technical preview in OCP 4.9 and is planned to go GA in 4.10.
+
+#### Dev Preview -> Tech Preview
+- Insights Operator is able to download the data from OCM API and expose it in a cluster API
+- basic functionality is tested
+- this new functionality is documented
+
+#### Tech Preview -> GA
+- ability to distinguish various error states - e.g organization doesn't have SCA allowed versus API returns 404
+- inform a cluster user about the error state (problem with pulling the certificates)
```
adambkaplan, line 109: I'd add admission control for CSI volumes as a GA graduation criteria

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,133 @@
...
+
+There are no other major milestones in the implementation history than the graduation criteria mentioned above.
+
+## Drawbacks
+
+There is no significant drawback.
+
+## Alternatives
+
+Only possible alternative is to implement this functionality in another control plane component/operator.
```
adambkaplan, line 133: Customers today obtain RHEL/SCA certs and add them to worker nodes via MachineConfig objects. We should vet the alternative here and why we don't want to encourage this (namely, SCA certs change regularly, and each update would require nodes to reboot).

enhancements/insights/pulling-data-from-ocm.md
dhellmann: @adambkaplan @iNecas it would be great if the body of this enhancement included the explanation you've given in this thread.

sbose78: Build API changes to support volume mounts
https://github.com/openshift/api/pull/954/files

sbose78: > Builds are a consumer of the SCA cert, but it is by no means the only workload that could find itself needing access to subscription content.

+1 

Subscription content is something that any workload in Openshift should be able to use. Builds happen to be a specialized case.

sbose78: >  Insights operator is the only core OCP component we have today that connects OpenShift to the rest of the Red Hat experience.

Correct.

Additionally, the general intent is to streamline all OCM API communications through this component. In that sense, the term "Insights" isn't quite indicative of what the future of Insights Operator would look like :)

enhancements/insights/pulling-data-from-ocm.md
tremes: Yes good points. There's no plan to mark a cluster as degraded. My idea was to introduce a new clusteroperator condition and create a new alert (with some threshold for number of retries) based on this condition. I think the alert will be a next step.

enhancements/insights/pulling-data-from-ocm.md
tremes: OK thanks. Good point. So it's basically a manual addition requiring node reboot right?

enhancements/insights/pulling-data-from-ocm.md
tremes: OK so I think I would need some help here please. I don't know how a feature gate works or what it exactly means. Also I don't know about the feature set.

enhancements/insights/pulling-data-from-ocm.md
tremes: OK thanks everybody. Let me try to sum the main reasons there.

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,133 @@
...
+
+Mitigation: Introduce a new state in the Insights Operator (e.g "SCADataDegraded") and
+create a new alert based on this new state.
+
+## Design Details
+
+### Authorization
+
+The Insights Operator is able to pull the data from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config-managed` namespace).
```
mfojtik, line 81: who put that secret there? it is managed, so some operator has to sync the secret there :-)

enhancements/insights/pulling-data-from-ocm.md
mfojtik: You make a feature gate in 4.9 and say "this is tech preview, if you want to opt-in, you have to enable this"... then in 4.10 you just make the feature gate on by default (assuming everything will work fine in clusters that have this enabled).

enhancements/insights/pulling-data-from-ocm.md
mfojtik: I think the "insights" operator is kind of confusing name then. I thought the purpose of "insights" is to provide diagnostic data to RH so we can better see cluster issues and we can detect classes of problems and provide mitigations.

Now the insights operator will have second role which is pulling a cert material from RH that enable subscriptions to be used in builds. I'm little concerned about mixing the purposes here.

enhancements/insights/pulling-data-from-ocm.md
adambkaplan: Feature gates are free-form strings that can be added to the `FeatureGate` API object (a custom resource specific to OpenShift). We have facilities to group feature gates into feature sets so cluster admins don't need to know all of them to enable logical groups features.

See https://github.com/openshift/api/blob/master/config/v1/types_feature.go#L113-L121

enhancements/insights/pulling-data-from-ocm.md
adambkaplan: correct - it's a manual addition to the cluster that has disruptive impact, especially at scale.

enhancements/insights/pulling-data-from-ocm.md
adambkaplan: A good question - is this something that the installer does, or does something like the CVO sync the pull secret in `openshift-config` to `openshift-config-managed`?

enhancements/insights/pulling-data-from-ocm.md
tremes: OK thanks. I updated this section.

enhancements/insights/pulling-data-from-ocm.md
tremes: OK thanks. I opened https://github.com/openshift/api/pull/983

enhancements/insights/pulling-data-from-ocm.md
tremes: @adambkaplan Can you please share more details or requirements for the `Share` object and the ClusterRole. I guess the IO has to define a `Share` object where the `backingResource` is the new secret right? And the ClusterRole is then able to read this `Share` object right?

sbose78: Historical note: I had agreed to have de-scope creation of the `Share` object(and related RBAC) since we wanted to make progress with getting the SCA cert into the cluster and provide a path to admins to make it available for users.

enhancements/insights/pulling-data-from-ocm.md
dhellmann: OK, I was asking about how non-OCM clusters do the same thing because it didn't make sense to me for the insights operator to do it just in 1 case. I think the explanation about why the insights operator needs to do it from another thread resolves my question about that. I agree that we can leave other use cases to future enhancements.

I am curious, though, about why we don't make use of OCM features like policy or hive features like syncsets to push certificates down into the cluster instead of making a special case for the operator to pull them. Was that approach explored at all?

adambkaplan: We did not expore OCM or hive. Two thoughts here regarding using OCM or Hive:

- OpenShift Cluster Management (OCM) is an add on that not every OpenShift cluster has. Not as sure about hive other than it is not to my knowledge used by all customers/clusters
- SCA certs change regularly (basically every time the customer makes changes to their subscriptions). Something needs to fetch the certificate and update it automatically - we arrived at the insights operator being the best tool we had lying around.

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,139 @@
...
+  - "@smarterclayton"
+creation-date: 2021-03-04
+last-updated: 2021-03-09
+status: implementable
+see-also:
+replaces:
+superseded-by:
+---
+
+# Insights Operator pulling and exposing data from the OCM API
```
dhellmann, line 21: This is a really generic title. Are we only going to expose the certs, or are there plans to expose other data?

enhancements/insights/pulling-data-from-ocm.md
dhellmann: OK, I was conflating OCM and ACM. Now that I understand they're different things, I see why we need something inside the cluster pulling the certs down.

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,139 @@
...
+available in the `pull-secret` (in the `openshift-config-managed` namespace).
+
+The Insights Operator must provide a cluster ID as an identifier of the cluster.
+
+### Data in API
+
+The SCA certificate is available via the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace.
+
+### Update period
+- Insights Operator query the OCM API every 8 hours and downloads the full data provided
```
dhellmann, line 92: Does that time period need to be adjustable? Does the user need a way to trigger a manual update under any circumstances?

enhancements/insights/pulling-data-from-ocm.md
dhellmann: > @adambkaplan @iNecas it would be great if the body of this enhancement included the explanation you've given in this thread.

I don't see the info from this thread in the doc, yet.

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,139 @@
...
+
+The periodical data pulling can be easily disabled in the cluster configuration. Removing this feature will require updating the Insights operator code base and will remove the `etc-pki-entitlement` secret from the `openshift-config-managed` namespace.
+
+### Upgrade / Downgrade Strategy
+
+There is no upgrade/downgrade strategy needed.
+
+### Version Skew Strategy
+
+There is no Skew strategy needed. This work should have no impact on the upgrade. It doesn't require any coordinated behavior in the control plane. No other components will change.
```
dhellmann, line 125: Does the API used to fetch the data ever change? Does the format of the cert ever change?

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Implementation History
+
+There are no other major milestones in the implementation history than the graduation criteria mentioned above.
+
+## Drawbacks
+
+There is no significant drawback.
```
dhellmann, line 133: What's the expected load impact of a bunch of clusters hitting the API to retrieve the certs?

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Alternatives
+
+- Alternative is to implement this functionality in another control plane component/operator (e.g openshift-controller-manager).
+- Current state, which is the manual addition of the SCA certs to cluster worker nodes. This is not very convenient because the SCA certs change regularly and the change requires node reboot.
```
dhellmann, line 138: Will the automated update somehow avoid those reboots? How often do the certs change and what's the impact of rebooting every OpenShift cluster that often? Remember, large metal clusters can take days to fully reboot.

bparees: the implementation is going to make the token available as a secret in a namespace in the cluster, so it won't go onto the nodes themselves.

Not exactly clear how consumers will access the secret, however, without using something like the (still being developed) csi shared resource driver.

enhancements/insights/pulling-data-from-ocm.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Proposal
+
+### User Stories
+
+#### Consume SCA certs exposed in the API
+
+As an OpenShift user
+I want to consume SCA certs to be able to consume RHEL content and to build
+corresponding container images.
```
bparees, line 65: given the secret is being put into openshift-config-managed, how does an openshift user actually consume the cert?

is that secret going to be readable by all users?  is it expected the admin will update RBAC to make it readable?  or is there some other mechanism envisioned here?

enhancements/insights/pulling-data-from-ocm.md
dhellmann: OK, I don't understand the comment on line 138 about manually adding the certs to worker nodes, then.

bparees: yeah i think that's an incomplete statement.  Today what people do is put the token on the worker node and then configure crio to automount it, i believe.  

As far as i can tell, this EP isn't really a full replacement for that approach because it appears to only address "how do i get the token to my cluster" but not "how does a workload consume the token from the cluster"  (which is fine, i'm not saying this EP has to be a full replacement for that process, but it probably needs to be clearer what this EP does and does not provide, and what is still left on the admin or user to do, once this EP is delivered)

enhancements/insights/pulling-data-from-ocm.md
bparees: > Now the insights operator will have second role which is pulling a cert material from RH that enable subscriptions to be used in builds. I'm little concerned about mixing the purposes here.

I agree it's slightly awkward, but since it's the same team either way, I don't care all that much how they manage the deployment of the stuff they are responsible for (one component with 2 jobs, or 2 components with 2 jobs).  It may make it slightly less obvious for someone to find the right component to debug if things aren't working right, but i dunno that that cost justifies the exercise of defining a whole new operator component to do this one trivial operation.

So as long as the reasoning is included in the EP, i think we can push past this discussion.  If this were one team trying to push code into another team's component(when it clearly didn't belong), i'd feel differently.

enhancements/insights/pulling-data-from-ocm.md
bparees: looking at the code, it looks like it's just using the pull-secret from openshift-config directly, not openshift-config-managed.

so the expectation is that secret is maintained by someone (if it expires, your nodes will have issues pulling images), so i don't think it's the responsibility of this EP to worry about it, other than that this EP should address what happens if it can't talk to OCM because the token expires(or for any other reason), which i think the EP does already.

see: https://github.com/tremes/insights-operator/blob/154641a9fae7c5100f7dce3cf0eddf4b38d56cea/pkg/config/configobserver/configobserver.go#L78

enhancements/insights/pulling-data-from-ocm.md
bparees: > I am not sure it's the goal of this proposal to answer this question.

i think it's worth mentioning the expectations (because those expectations tie into how you intend to secure the content, where it should be stored, as well as demonstrating that there is value in doing this and that it's being done in a way that those downstream consumers will find useful), but i agree that the answer should be general and not get into specific implementation details.

enhancements/insights/pulling-data-from-ocm.md
bparees: this component may be used by builds, but it needs its own tests that pass/fail without any dependency on how builds(or anyone/thing else) attempt to consume the content.

if/when builds add their logic to consume this, they will add their own tests that ensure their consumption logic works, but i don't think that's relevant to this EP.

enhancements/insights/pulling-data-from-ocm.md
bparees: this discussion appears to be resolved, i am marking it so.

enhancements/insights/pulling-data-from-ocm.md
bparees: i find it somewhat weird we would not mark it as degraded.  the definition of degraded is that the operator isn't providing all the function it's supposed to provide, but isn't totally broken.  Adding a new condition that amounts to saying that same thing, rather than using the condition that exists for the purpose seems odd.

It's always possible to upgrade a "Degraded" cluster(via force/override, though I recently learned that OSD is very reluctant to let you do that), but you want to do it with awareness that you're choosing to do so (imagine all your pods that get restarted after upgrade need SCACerts for some init step....wouldn't you want to know before you upgrade that that's going to be a problem?)

i think this is something we can revisit as the feature moves from techpreview to GA, but i do think we should discuss it a bit more.

tremes: OK I didn't know that. My main reason was I didn't want to block an upgrade because of failing SCACerts pull. I think I am fine with changing to "Degraded" condition then. Thanks

enhancements/insights/pulling-data-from-ocm.md
tremes: OK let me change it to mention just the certs. Thanks

enhancements/insights/pulling-data-from-ocm.md
tremes: Good point. The time period is configurable so I can mention it here. For the second question, do you mean an update of the period right? I think I can imagine some situations when an admin might want to update the period. So I can mention it here.

enhancements/insights/pulling-data-from-ocm.md
tremes: Can you help me @petli-openshift to answer this question pls? 
I think the format of the cert is irrelevant from the perspective of this EP (iow it will expose what it receives). Pls correct me if I am wrong.

enhancements/insights/pulling-data-from-ocm.md
tremes: Can you @petli-openshift help me to answer this question pls?

enhancements/insights/pulling-data-from-ocm.md
bparees: @dhellmann are you ok resolving this convo?

enhancements/insights/pulling-data-from-ocm.md
bparees: i think @dhellmann means will the admin have a way to manually trigger the insights operator to pull down a new cert, outside of the scheduled periodic update.  Not just can the admin update the interval.

enhancements/insights/pulling-data-from-ocm.md
bparees: > I'd add admission control for CSI volumes as a GA graduation criteria

why?

Consumption of the content seems outside the scope of this EP.  The GA version of this EP is a component that can pull the cert down and put it in a well-defined location that we are happy with, reliably+securely, that is debuggable, provides useful status info, etc.

enhancements/insights/pulling-data-from-ocm.md
dhellmann: Sounds good. The filename should be updated, too.

enhancements/insights/pulling-data-from-ocm.md
dhellmann: Yes, we should answer both questions.

enhancements/insights/pulling-data-from-ocm.md
dhellmann: It would be good to have a few sentences in the alternatives section explaining that a new operator could have been created but it would require more work from multiple teams (building new images, etc.) and would require more CPU and RAM resources in the cluster. The counter argument is that it makes the cluster operator status more focused, but I think that's not worth the extra work, myself.

sbose78:
I'd like to address a couple of questions I see in the above in multiple discussion threads.

### How does an entity make the SCA cert accessible to service accounts in other namespaces ?

* Make the `Secret`  `etc-pki-entitlement` available for use in other namespaces by creating a cluster-scoped `Share` resource
```
apiVersion: projectedresource.storage.openshift.io/v1alpha1
kind: Share
metadata:
  name: etc-pki-entitlement
spec:
  backingResource:
    kind: Secret
    apiVersion: v1
    name: etc-pki-entitlement
    namespace: openshift-config-managed
```

*  Admin creates a `clusterrolebinding` to allow a service account ( example, `pod-sa` ) access to the above `Share` resource ( [sample](https://github.com/openshift/csi-driver-shared-resource/blob/master/examples/01-clusterrolebinding.yaml) )

### Consumption by Pods

```
kind: Pod
apiVersion: v1
metadata:
  name: user-workload
  namespace: foo
spec:
  serviceAccountName: pod-sa  # <----- Previously a clusterrolebinding was created for this sa
  containers:
    - name: my-frontend
      image: quay.io/quay/busybox
      volumeMounts:
        - mountPath: "/data"
          name: my-csi-volume
      command:
        - sh
        - -c
        - |
          while true
          do
            ls -la /data
            touch /data/foo
            ls -la /data
            echo "calling sleep"
            sleep 120
          done
  volumes:
    - name: my-csi-volume
      csi:
        #readOnly: true
        driver: csi.shared-resources.openshift.io
        volumeAttributes:
          share: etc-pki-entitlement. # <----- name of the cluster-scoped `Share` resource.
```

### Consumption by Builds

After https://issues.redhat.com/browse/BUILD-274 is done, one would be able to do the following: 

```
kind: Build
apiVersion: build.openshift.io/v1
metadata:
  ...
spec:
  source:
     ...
  strategy:
    dockerStrategy:
      volumes:
      - name: etc-pki-entitlement
        type: CSI
        csi: 
          driver: projected-resource.csi.openshift.io
          volumeAttributes:
            share: etc-pki-entitlement
      volumeMounts:
      - name: etc-pki-entitlement
        mountPath: /etc/pki/entitlement
```

As of today, the Build API team has made progress with supporting Secret & ConfigMap volume mounts in Builds as part of https://issues.redhat.com/browse/BUILD-87 . https://issues.redhat.com/browse/BUILD-274 is a candidate for 4.10 .

sbose78:
As Ben mentioned, this EP's sole goal is to ensure the SCA certs are pulled into the cluster and lifecycle'd by the Insights Operator.

Consumption of the same is being driven by work in distinct Openshift components:
* The brand new [OpenShift Projected Resource CSI Driver](https://github.com/openshift/csi-driver-shared-resource/)
* Changes in the Build API to support volume mounts ( specifically, CSI volume mounts )

enhancements/insights/pulling-data-from-ocm.md
tremes: OK why not, but I think Michal was suggesting to move this functionality to `openshift-controller-manager` (This is briefly mentioned there). I can also mention this third option of completely new component/operator.

enhancements/insights/pulling-data-from-ocm.md
tremes: I guess the question was why the new `etc-pki-entitlement` secret is created in the `openshift-config-managed` namespace (rather than `openshift-config` ?). 

This requires a new `Role` & `RoleBinding` definition to be able to create the secret in the `openshift-config-managed` namespace. Is that OK/acceptable? Comparing to extending our existing role to be able to create the secret in the `openshift-config` namespace.

enhancements/insights/pulling-data-from-ocm.md
tremes: Done.

enhancements/insights/pulling-data-from-ocm.md
petli-openshift: The API to fetch the data does not change.

enhancements/insights/pulling-data-from-ocm.md
petli-openshift: The certificates are cached in the API server. We expect the impact would be minimal.

tremes:
I slightly updated the proposal to address some points and cover the part of SCA certs use. Can you @bparees, @dhellmann please take a look at my last commit? Let's discuss potential blockers tomorrow.

enhancements/insights/pulling-data-from-ocm.md
patrickeasters: The cert returned by RHSM is composed of several PEM-encoded blocks of data. The certificate format (what extensions are used, etc) is versioned, but this has not changed in many years. A certificate version change would only impact client tooling like subscription-manager and would be transparent to OCM.

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -84,12 +91,19 @@ available in the `pull-secret` (in the `openshift-config-managed` namespace).
 
 The Insights Operator must provide a cluster ID as an identifier of the cluster.
 
-### Data in API
+### SCA certs in API
 
-The SCA certificate is available via the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace.
+The SCA certificate is available via the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace. The secret will be available for use in other namespaces by creating a cluster-scoped `Share` resource. Cluster admin creates a `clusterrolebinding` to allow a service account access to the `Share` resource.
```
bparees, line 96: link to the csi-shared-resource driver page, especially since this isn't something that just exists on clusters today:

https://github.com/openshift/csi-driver-shared-resource

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,156 @@
...
+### Risks and Mitigations
+
+#### OCM API is down
+
+Risk: OCM API is down or doesn't provide up to date data.
+
+Risk: Insights Operator is unable to expose/update the data in the OpenShift API
+
+Mitigation: Introduce a new state in the Insights Operator (e.g "SCADataDegraded") and
+create a new alert based on this new state.
```
bparees, line 83: this is outdated, right?  i think there was agreement to just use the degraded state.

since it's a techpreview feature, people can't upgrade anyway, but we can use this to gather data about whether the operator is going degraded in customer clusters frequently, and what the implications of that are, before we GA the feature.

We should definitely ensure we've gathered some data on how often this operator is going degraded due to this feature before we GA the feature.

tremes: OK makes sense to me. Perhaps we should explicitly define the HTTP error states when the IO goes degraded (after some threshold/retries), because for example 404 means that the org didn't allow this feature (in OCM or RHSM I guess) so there's no reason to degrade it.

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: Linked at https://github.com/openshift/enhancements/pull/683/files#diff-93dffd2aa59c973a916306675f62907275cff29ff9a16ba77aaf98cfa830249fR100

enhancements/insights/pulling-data-from-ocm.md
tremes: Thanks. I mentioned it at https://github.com/openshift/enhancements/pull/683/files#diff-93dffd2aa59c973a916306675f62907275cff29ff9a16ba77aaf98cfa830249fR148

enhancements/insights/pulling-data-from-ocm.md
tremes: Thank you. I updated the section at https://github.com/openshift/enhancements/pull/683/files#diff-93dffd2aa59c973a916306675f62907275cff29ff9a16ba77aaf98cfa830249fR140

enhancements/insights/pulling-data-from-ocm.md
tremes: I tried to answer both questions in https://github.com/openshift/enhancements/pull/683/files#diff-93dffd2aa59c973a916306675f62907275cff29ff9a16ba77aaf98cfa830249fR102

enhancements/insights/pulling-data-from-ocm.md
sbose78: > I guess the question was why the new etc-pki-entitlement secret is created in the openshift-config-managed namespace (rather than openshift-config ?).

Configuration in `openshift-config` are usually added manually by admins compared, given that `etc-pki-entitlement` is managed by an OpenShift component, Adam and I suggested it be managed in the `openshift-config-managed` namespace.

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,155 @@
...
+### Authorization
+
+The Insights Operator is able to pull the data from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config-managed` namespace).
+
+The Insights Operator must provide a cluster ID as an identifier of the cluster.
+
+### SCA certs in API
+
+The SCA certificate is available via the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace. The secret will be available for use in other namespaces by creating a cluster-scoped `Share` resource. Cluster admin creates a `clusterrolebinding` to allow a service account access to the `Share` resource.
```
sbose78, line 95: I would propose we have the Insights Operator create the `Share` resource as well. This would help tell admins that the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace is a candidate for propagation across namespaces. This was discussed initially but I had put this on hold since the `Share` API was under development as part of an OLM operator. Things have changed since the - we'll have the `Share` resource available on day 0.

I would suggest we add this to this EP and implement the same in parts whenever the `Share` resource becomes available on clusters.

sbose78 (changes requested):
Requesting one change : Create the `Share` resource.

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,155 @@
...
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [x] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement will enable the Insights Operator to pull the data (SCA certs)
```
dhellmann, line 37:
```suggestion
This enhancement will enable the Insights Operator to pull the Simple Content Access certs for using RHEL subscription content
```

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,155 @@
...
+from the OCM (OpenShift Cluster Manager) API. The data will be exposed by the Insights Operator
```
dhellmann, line 38:
```suggestion
from the OCM (OpenShift Cluster Manager) API. The certs will be exposed by the Insights Operator
```

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,155 @@
...
+Users could consume RHEL content and container images using the RHEL subscription in the OpenShift 3.x.
+In the OpenShift 4, this is no longer possible because the Red Hat Enterprise Linux Core OS (RHCOS) does not
+provide any attached subscription. This enhancement is to provide users the Simple Content Access (SCA) certs
+from Red Hat Subscription Manager (RHSM).
+
+### Goals
+
+- Extend the Insights Operator config with an OCM API URL to be able to query the data
+- Periodically pull the data from the OCM API and expose it in the OpenShift API
+- This is an opt-in feature by a cluster user and might be moved to a different OCP component in the future
```
dhellmann, line 53: If we think this might move to a different component in the future, should the config be tied to the insights operator configuration now?

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,155 @@
...
+
+#### OCM API is down
+
+Risk: OCM API is down or doesn't provide up to date data.
+
+Risk: Insights Operator is unable to expose/update the data in the OpenShift API
+
+Mitigation: The Insights Operator is marked as Degraded (in case of HTTP error code other than 404, which means that the organization didn't allow this feature on the OCM side).
+
+## Design Details
```
dhellmann, line 84: The goals mention that this is an opt-in feature. How is it enabled?

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,155 @@
...
+The SCA certificate is available via the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace. The secret will be available for use in other namespaces by creating a cluster-scoped `Share` resource. Cluster admin creates a `clusterrolebinding` to allow a service account access to the `Share` resource.
+
+### Use of the SCA certs
+
+- The SCA certificate can be mounted to a `Pod` as a CSI volume (where the volume attributes will reference the `Share` resource making the secret accessible)
+- The SCA certificate can be mounted to a `Build` strategy as a CSI volume. The CSI driver is described in the [Share Secrets And ConfigMaps Across Namespaces via a CSI Driver](/enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md) enhancement and implemented in [https://github.com/openshift/csi-driver-shared-resource](https://github.com/openshift/csi-driver-shared-resource).
+
+### Update period
+- Insights Operator query the OCM API every 8 hours and downloads the full data provided
+- The time period is configurable and can be changed by the cluster admin. Cluster admin can temporarily set a shorter time period to try to refresh the SCA certs
```
dhellmann, line 104: Is there an API for adjusting the time period, or is the setting in a ConfigMap?

tremes: It can be done via `support` secret. Let me mention it here. Thanks

enhancements/insights/pulling-sca-certs-from-ocm.md
bparees: yes that would be good.

enhancements/insights/pulling-data-from-ocm.md
tremes: The degraded state is mentioned in the EP and will be implemented within https://issues.redhat.com/browse/CCXDEV-5227. Resolving the conversation.

enhancements/insights/pulling-data-from-ocm.md
tremes: This is described in https://github.com/openshift/enhancements/pull/683/files#diff-93dffd2aa59c973a916306675f62907275cff29ff9a16ba77aaf98cfa830249fR97. Resolving

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: I added new section `Configuration` explaining the current situation https://github.com/openshift/enhancements/pull/683/files#diff-93dffd2aa59c973a916306675f62907275cff29ff9a16ba77aaf98cfa830249fR102

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: I also added new section `Configuration` describing all the config options.

enhancements/insights/pulling-data-from-ocm.md
tremes: This is described in the [Use of the SCA certs](https://github.com/openshift/enhancements/pull/683/files#diff-93dffd2aa59c973a916306675f62907275cff29ff9a16ba77aaf98cfa830249fR97) section.

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: I added it to the mitigation description.

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: Based on our discussion, the creation of the `Share` resource object is described in the [### Tech Preview -> GA](https://github.com/openshift/enhancements/pull/683/files#diff-93dffd2aa59c973a916306675f62907275cff29ff9a16ba77aaf98cfa830249fR132)

enhancements/insights/pulling-data-from-ocm.md
mklika: Can we resolve this point @bparees @adambkaplan ?

enhancements/insights/pulling-data-from-ocm.md
mklika: @mfojtik Hi, are you Ok if we resolve this conversation based on the comments above?

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+- The SCA certificate can be mounted to a `Pod` as a CSI volume (where the volume attributes will reference the `Share` resource making the secret accessible)
+- The SCA certificate can be mounted to a `Build` strategy as a CSI volume. The CSI driver is described in the [Share Secrets And ConfigMaps Across Namespaces via a CSI Driver](/enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md) enhancement and implemented in [https://github.com/openshift/csi-driver-shared-resource](https://github.com/openshift/csi-driver-shared-resource).
+
+### Configuration
+- The new configuration is tied to the Insights Operator, because the Insights Operator is currently the only component reading and using the `console.redhat.com` authentication token in the `pull-secret` in the `openshift-config` namespace, so if we decide to move this somewhere else, we must consider this fact.
+- The new OCM API related config attributes are adjustable via `support` secret in the `openshift-config` namespace
+- Following attributes are configurable via the `support` secret:
+  - OCM API endpoint
+  - time period to query the OCM API
+  - flag to disable this functionality
```
wking, line 108: None of these three attributes sound particularly sensitive to me.  Maybe this should be a ConfigMap and not a Secret?  Or am I missing something about why one or all of them is sensitive?

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+### Update period
+- Insights Operator query the OCM API every 8 hours and downloads the full data provided
+- The time period is configurable (via `support` secret) and can be changed by the cluster admin. Cluster admin can temporarily set a shorter time period to try to refresh the SCA certs
+- The documentation will describe the steps how to pull the SCA certs and update the secret manually
+
+### Test Plan
+
+- `insights-operator-e2e-tests` suite can verify the SCA cert data
+  is available
+- Basic test of the validity of the SCA certs. Mount the `etc-pki-entitlement` secret and run e.g `yum install` in the container
```
wking, line 119: Since most consumers will presumably be mounting the Share, maybe this integration test should use that approach instead of shortcutting to use the Secret directly?

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+### Authorization
+
+The Insights Operator is able to pull the certificates from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config-managed` namespace).
+
+The Insights Operator must provide a cluster ID as an identifier of the cluster.
+
+### SCA certs in API
+
+The SCA certificate is available via the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace. The secret will be available for use in other namespaces by creating a cluster-scoped `Share` resource. Cluster admin creates a `clusterrolebinding` to allow a service account access to the `Share` resource.
```
wking, line 95: Not covered in this sentence is the structure of the Secret.  Will a particular well-known key be used?  If so, can we document it here, or is that a low-enough level of detail that it's not worth including in the enhancement proposal?

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+- the use of SCA certs in the `Build` is documented in the Build API documentation
+- the feature might be moved to a different OCP component
+
+#### Removing a deprecated feature
+
+The periodical data pulling can be easily disabled in the cluster configuration. Removing this feature will require updating the Insights operator code base and will remove the `etc-pki-entitlement` secret from the `openshift-config-managed` namespace.
+
+### Upgrade / Downgrade Strategy
+
+There is no upgrade/downgrade strategy needed.
```
wking, line 145: If a cluster is downgraded to a version that does not poll for entitlement updates, will that version of the insights operator (4.8.z?) have logic around to remove the `etc-pki-entitlement` secret and other cruft to keep in-cluster components from trying to consume stale data?  If the insights operator is downgraded before some consumer (builds and/or CSI drivers?), will the higher-version consumers gracefully handle the consumed secret's removal?

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+### Non-Goals
+
+- Insights Operator providing any transformation or post-processing of the SCA certs pulled
+  from the OCM API
+
+## Proposal
+
+### Why is it in the Insights Operator?
+
+The Insights Operator is now the only OCP component that connects an OpenShift cluster to a Red Hat subscription experience (console.redhat.com APIs). The consumers of the SCA certs are not only builds, but also shared resources, such as the CSI driver.
```
adambkaplan, line 64: The ultimate consumers are any workload that needs access to RHEL subscription content. OpenShift Builds are expected to be the largest consumer.

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Risks and Mitigations
+
+#### OCM API is down
+
+Risk: OCM API is down or doesn't provide up to date data.
+
+Risk: Insights Operator is unable to expose/update the data in the OpenShift API
+
+Mitigation: The Insights Operator is marked as Degraded (in case the HTTP code is lower than 200 or greater than 399 and is not equal to 404, because HTTP 404 means that the organization didn't allow this feature).
```
adambkaplan, line 82: Degraded means the cluster can't upgrade. Are we sure we want this?

enhancements/insights/pulling-sca-certs-from-ocm.md
adambkaplan: IIRC the key name can be any `.pem` file. On a RHEL system the subscription cert key names have what appear to be a uuid for the file name.

enhancements/insights/pulling-sca-certs-from-ocm.md
adambkaplan: I think the bits involving the `Share` resource should ultimately live in the OpenShift build suite. Test plan as follows:

1. <Assume> insights operator obtains SCA cert from OCM
2. <Assume> insights operator creates `Share` resource and ClusterRoleBinding
3. ocp build suite creates a build that does a yum install of subscription-only content

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Test Plan
+
+- `insights-operator-e2e-tests` suite can verify the SCA cert data
+  is available
+- Basic test of the validity of the SCA certs. Mount the `etc-pki-entitlement` secret and run e.g `yum install` in the container
+
+### Graduation Criteria
+
+This feature is planned as a technical preview in OCP 4.9 and is planned to go GA in 4.10.
```
adambkaplan, line 123: Note that the `Share` bits are now planned as tech preview for OCP 4.10

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+
+#### Dev Preview -> Tech Preview
+- opt-in feature (called `InsightsOperatorPullingSCA`) enabled with `TechPreviewNoUpgrade` feature set
+- Insights Operator is able to download the certificates from OCM API and expose it in a cluster API
+- Insights Operator is marked as degraded in case of the number of unsuccessful requests to the OCM API exceeds defined threshold
```
adambkaplan, line 128: Ditto my comments above - degraded means no cluster upgrades. Do we want this?

enhancements/insights/pulling-sca-certs-from-ocm.md
bparees: yes, we want it.

1) this is tech preview.  we should absolutely gather data on whether this operator is ending up degraded during the tech preview period before making a final decision for GA

2) degraded doesn't block z-stream upgrades, and can be overridden for y-stream upgrades, the point is it's a conscious choice.  I gave this (somewhat dubious) example in another comment thread on this EP:  suppose you have pods that need this entitlement to do work at startup.  Doing an upgrade is going to trigger all those pods to restart.  You want to know, before you upgrade your cluster, that you don't have valid entitlements on your cluster, or all those pods are going to fail to restart after the node reboots triggered by the upgrade.

fundamentally this operator has a job to do, and if it's not doing the job, it should report degraded(after a reasonable period of time)

bparees: that said, going degraded should only happen after some number of retries or period of time.  not on first failure.

bparees: and needs to take into account disconnected clusters (i.e. it shouldn't be degraded on a cluster w/ no network access)

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+
+#### OCM API is down
+
+Risk: OCM API is down or doesn't provide up to date data.
+
+Risk: Insights Operator is unable to expose/update the data in the OpenShift API
+
+Mitigation: The Insights Operator is marked as Degraded (in case the HTTP code is lower than 200 or greater than 399 and is not equal to 404, because HTTP 404 means that the organization didn't allow this feature).
+
+## Design Details
```
bparees, line 84: is your operator going to respect the cluster proxy configuration so it can reach OCM if the cluster is behind a proxy?

what will it do if the cluster is running disconnected?

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: Yes, Ben is right. The `support` secret is already used for the Insights config stuff (ingress url, gathering period) and I don't know either why a secret is used as the config resource.

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: Maybe we should mention it here. I am using `kubernetes.io/tls` secret with `tls.crt` for the certificate and `tls.key` for the private key.

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: Yes this sounds like an edge case to me and yes there will be a stale secret in such case.

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: The operator should respect the cluster wide proxy configuration and shouldn't go degraded in a disconnected environment. Good points. Thanks. Let me put it there.

tremes: Updated.

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Authorization
+
+The Insights Operator is able to pull the certificates from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config-managed` namespace).
```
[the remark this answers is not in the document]
tremes: Updated. Thanks

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: Updated the sentence. THanks

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: Well maybe because you can configure there also username/password for the basic auth https://github.com/openshift/insights-operator/blob/master/pkg/config/configobserver/configobserver.go#L134.

enhancements/insights/pulling-sca-certs-from-ocm.md
tremes: Does it mean that it will block the graduation criteria from the TP to GA mentioned here? If so then we would need to go GA with your bits...I guess

enhancements/insights/pulling-sca-certs-from-ocm.md
adambkaplan: It appears that prior GH comments we reached consensus that Degraded is a desirable condition. I would like to see this justification reflected in the proposal itself.

enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -0,0 +1,166 @@
...
+- The new configuration is tied to the Insights Operator, because the Insights Operator is currently the only component reading and using the `console.redhat.com` authentication token in the `pull-secret` in the `openshift-config` namespace, so if we decide to move this somewhere else, we must consider this fact.
+- The new OCM API related config attributes are adjustable via `support` secret in the `openshift-config` namespace
+- Following attributes are configurable via the `support` secret:
+  - OCM API endpoint
+  - time period to query the OCM API
+  - flag to disable this functionality
+
+### Update period
+- Insights Operator query the OCM API every 8 hours and downloads the full data provided
+- The time period is configurable (via `support` secret) and can be changed by the cluster admin. Cluster admin can temporarily set a shorter time period to try to refresh the SCA certs
```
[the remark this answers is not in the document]
patrickeasters: They are valid for a year, but will be revoked and regenerated by Candlepin if available content changes.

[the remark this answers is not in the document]
bparees: thanks

bparees:
/approve

looks like there might be a few final items @adambkaplan would like to see explicitly stated in the EP, but in general i think the critical concerns are addressed and/or roadmapped.

tremes:
Thanks @bparees. I added one more commit mentioning the reasons for the decision on the degraded status.

sbose78:
FYI, the consumption APIs are undergoing changes 
https://github.com/openshift/csi-driver-shared-resource/pull/55

PR #686, opened 2021-03-10
NE-530 ingress: global options to enable HSTS

candita:
This enhancement provides the capability to enable global automatic HTTP Strict Transport Security as a default for all routes.

In 3.x and 4.x customers can [provide a per-route annotation to enable HSTS](https://docs.openshift.com/container-platform/4.4/networking/routes/route-configuration.html#nw-enabling-hsts_route-configuration).  
For customers with many routes or regulatory compliance issues, the manual per-route annotation is problematic.

This enhancement extends the Route API to allow cluster administrators to enable HSTS globally, without having to add an annotation to each route.

### DIFF
--- a/enhancements/ingress/global-options-enable-hsts.md
+++ b/enhancements/ingress/global-options-enable-hsts.md
@@ -1,0 +1,465 @@
+---
+title: global-options-enable-hsts 
+authors:
+  - "@cholman"
+
+reviewers:
+  - "@danehans" 
+  - "@frobware"
+  - "@knobunc"
+  - "@Miciah"
+  - "@miheer"
+  - "@rfredette"
+  - "@sgreene570"
+
+approvers:
+  - "@frobware"
+  - "@knobunc"
+
+creation-date: 2021-03-08
+
+last-updated: 2021-03-26
+
+status: implementable
+
+see-also:
+  - https://bugzilla.redhat.com/show_bug.cgi?id=1512759
+  - https://bugzilla.redhat.com/show_bug.cgi?id=1430035
+
+replaces:
+
+superseded-by:
+
+---
+
+# Global Options to Enable HTTP Strict Transport Security (HSTS)
+
+This enhancement provides the capability to enable global automatic HSTS
+as a default for all TLS routes.
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+In 3.x and 4.x customers can [provide a per-route annotation to enable HSTS](
+https://docs.openshift.com/container-platform/4.4/networking/routes/route-configuration.html#nw-enabling-hsts_route-configuration).  For
+customers with many routes or regulatory compliance issues, the manual per-route annotation is
+problematic.
+
+This enhancement extends the `IngressController` API to allow cluster administrators to
+enable HSTS globally, without having to add an annotation to each route.
+
+## Motivation
+
+HSTS ([RFC 6797](https://tools.ietf.org/html/rfc6797)) policy enforces the use
+of HTTPS in client requests to the host, without having to use an HTTP redirect.
+HSTS provides user protection and is concerned with minimizing security threats based on network
+traffic eavesdropping and man-in-the-middle attacks.  Using a response header
+called `Strict-Transport-Security`, the HTTP response informs clients that a website
+can only be accessed via HTTPS.  
+
+Administrators who are tasked with route management and/or regulatory compliance
+issues want an option to automatically and globally enable HSTS. They also want
+to separately apply HSTS globally to specific routers that target applications,
+while retaining the ability to opt-out of HSTS on a per-route basis.
+
+### Goals
+
+This proposal allows administrators to:
+- Continue to enable HSTS per-route (retain the existing feature functionality)
+- Enable HSTS globally, for all TLS routes
+- Enable HSTS per-domain
+- Disable HSTS per-route when it is enabled globally
+
+### Non-Goals
+
+- HSTS cannot be applied to non-TLS routes, even if HSTS is requested for all routes globally.  As
+mentioned in [RFC 6797 Section 8.1](https://tools.ietf.org/html/rfc6797#section-8.1), if an HTTP
+response is received over insecure transport, clients must ignore any present STS headers.
+
+## Proposal
+
+Each goal is addressed in the subsections below.
+
+### Continue to enable HSTS per-route (retain the existing feature functionality)
+The `Strict-Transport-Security` response header is currently applied per-route via an
+annotation called `haproxy.router.openshift.io/hsts_header`.  This annotation has a
+required directive `max-age`, and optional directives `includeSubDomains` and `preload`,
+with the following definitions:
+- `max-age`: delta time range in seconds during which the host is to be regarded as an HSTS host
+  - If set to 0, it negates the effect and the host is no longer regarded as an HSTS host
+  - `max-age` is a time-to-live value, so if a client makes a request for the route's host
+  and this period of time elapses before the client makes another request for the same host, the
+  HSTS policy will expire on that client
+- `includeSubDomains`: if present, applies to any subdomains of the host's domain name, when the
+route admission policy allows wildcards.  If wildcard routes aren't allowed, this directive has no
+effect.
+- `preload`: if present, tells the client to include the route's host in its host preload list so that
+it never needs to do an initial load to get the HSTS header (note that this is not defined in RFC 6797
+and is therefore client implementation-dependent)
+
+HSTS is currently implemented in the HAProxy template and applied to `edge` and
+`reencrypt` routes that have the `haproxy.router.openshift.io/hsts_header` annotation:
+```gotemplate
+{{- /* hsts header in response: */}}
+ {{- $hstsOptionalTokenPattern := `(?:includeSubDomains|preload)` }}
+ {{- $hstsPattern := printf `(?:%[1]s[;])*max-age=(?:\d+|"\d+")(?:[;]%[1]s)*`  $hstsOptionalTokenPattern -}}
+...
+{{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}
+    {{- with $hsts := firstMatch $hstsPattern (index $cfg.Annotations "haproxy.router.openshift.io/hsts_header") }}
+  http-response set-header Strict-Transport-Security {{$hsts}}
+    {{- end }}{{/* hsts header */}}
+  {{- end }}{{/* is "edge" or "reencrypt" */}}
+```
+
+#### Enable HSTS globally, for all TLS routes
+To add the header to all HTTP responses from an `edge` or `reencrypt` route, add a new router environment
+variable called `ROUTER_DEFAULT_HSTS`.
+
+Add a new type `IngressControllerHSTSHeaderPolicy` to the API in
+`openshift/api/blob/master/operator/v1/types_ingress.go`, to capture the configuration of the HSTS Policy.
+```go
+// These are strings instead of enums so that they can be passed in easily from the HAProxy template
+const (
+    AllHSTSHeaderPolicyScope         = "All"
+    LimitedHSTSHeaderPolicyScope     = "Limited"
+)
+
+// HSTSDirectiveType is an optional directive that directs HSTS Policy configuration.
+//+kubebuilder:validation:Enum=preload;includeSubDomains
+type HSTSDirectiveType string
+const(
+    // PreloadHSTSDirective directs the client to include hosts in its host preload list so that
+    // it never needs to do an initial load to get the HSTS header (note that this is not defined
+    // in RFC 6797 and is therefore client implementation-dependent).
+    PreloadHSTSDirective             HSTSDirectiveType = "preload"
+
+    // IncludeSubDomainsHSTSDirective means the HSTS Policy applies to any subdomains of the host's
+    // domain name, when the route admission policy allows wildcards.  If wildcard routes aren't allowed,
+    // this directive has no effect.
+    IncludeSubDomainsHSTSDirective   HSTSDirectiveType = "includeSubDomains"
+)
+
+type IngressControllerHSTSHeaderPolicy {
+	// maxAgeSeconds is the delta time range in seconds during which hosts are regarded as HSTS hosts.
+	// If set to 0, it negates the effect and hosts are no longer regarded as HSTS hosts.
+	// maxAgeSeconds is a time-to-live value, and if this policy is not refreshed on a client, the HSTS
+	// policy will eventually expire on that client.
+	// +required
+    MaxAgeSeconds       int32 `json:"maxAgeSeconds"`
+
+    // HSTSDirectives is a list of optional HSTS directives that direct HSTS Policy configuration.
+    // +kubebuilder:validation:UniqueItems=true
+    // +optional
+    HSTSDirectives      []HSTSDirectiveType `json:"hstsDirectives,omitempty"`
+
+    // scope indicates whether this header policy is enabled for a limited set of routes that have hosts
+    // with domains in the list specified in inScopeDomains (limited scope) or for all TLS routes
+    // (all scope).  If HSTS is desired for all routes, set the scope field to "All". If HSTS is desired
+    // for only a limited set of routes, set the scope field to "Limited" and add a list of subdomains
+    // to the InScopeDomains field.  If scope is omitted, no HSTS header policy will be added.
+    // +optional
+    Scope               string `json:"scope,omitempty"`
+
+    // inScopeDomains is a list of router domains that require HSTS, when not all routes require HSTS.
+    // If scope = All then this field is ignored.  If scope = Limited, HSTS policy is limited to hosts with
+    // domains in the inScopeDomains list.
+    // +optional
+    InScopeDomains      []string `json:"inScopeDomains,omitempty"`
+}
+```
+
+Validation and formatting in the `IngressController`
+would change the fields into a semi-colon delimited string with formatting:
+```yaml
+max-age=<numberofseconds>[;includeSubDomains][;preload]`
+```
+`ROUTER_DEFAULT_HSTS` is detected in the HAProxy configuration file processing, and can be overridden
+by any individual route's annotation.  The global `hstsHeaderPolicy` does not override an existing route
+annotation for `haproxy.router.openshift.io/hsts_header`.
+
+The `hstsHeaderPolicy` variable can be configured by the administrator in the `IngressController`, as shown in this example, where
+the HSTS Policy scope is for all routes, expires after one year, includes subdomains, and requests preload:
+```yaml
+apiVersion: operator.openshift.io/v1
+kind: IngressController
+metadata:
+  name: default
+  namespace: openshift-ingress-operator
+spec:
+  httpHeaders:
+    hstsHeaderPolicy:
+      scope: All
+      maxAgeSeconds: 31536000
+      hstsDirectives:
+      - includeSubDomains
+      - preload
+```
+An alternative option is to use a `hstsHeader` as a single text string so that it matches the existing annotation form.
+Validation in the ingress controller would be different.  I chose to keep it as distinct fields in the `hstsHeaderPolicy`
+for user experience ease.  If we decide not to go that way, the same HSTS Policy would look like this:
+```yaml
+    hstsHeaderPolicy:
+      scope:  All
+      hstsHeader: "max-age=31536000;includeSubDomains;preload"
+```
+
+The `spec.httpHeaders.hstsHeaderPolicy` is formatted and copied to the router environment variable `ROUTER_DEFAULT_HSTS` when the
+`IngressOperator` creates the router.  Another environment variables holds the scope `ROUTER_HSTS_SCOPE`.  Changes are needed in `pkg/operator/controller/ingress/deployment.go`:
+```go
+const (
+	WildcardRouteAdmissionPolicy = "ROUTER_ALLOW_WILDCARD_ROUTES"
+	...
+	RouterHSTSHeaderPolicy       = "ROUTER_DEFAULT_HSTS"
+	RouterHSTSHeaderScope        = "ROUTER_HSTS_SCOPE"
+    ...
+)
+...
+func desiredRouterDeployment(ci *operatorv1.IngressController, ingressControllerImage string, ingressConfig *configv1.Ingress, apiConfig *configv1.APIServer, networkConfig *configv1.Network, proxyNeeded bool) (*appsv1.Deployment, error) {
+...
+    if ci.Spec.HTTPHeaders != nil && ci.Spec.HTTPHeaderPolicy != nil && len(ci.Spec.HSTSHeaderPolicy.MaxAge) != 0 {
+        header := "max-age=" + ci.Spec.HTTPHeaders.HSTSHeaderPolicy.MaxAgeSeconds
+        for _, directive := range ci.Spec.HTTPHeaders.HSTSHeaderPolicy.HSTSDirectives {
+            switch directive {
+                case  PreloadHSTSDirective:
+                    header = header +  ";" + PreloadHSTSDirective
+                case IncludeSubDomainsHSTSDirective:
+                    header = header + ";" + IncludeSubDomainsHSTSDirective
+            }
+        }
+        env = append(env, corev1.EnvVar{Name: RouterHSTSHeaderPolicy, Value: header})
+        switch ci.Spec.HTTPHeaders.HSTSHeaderPolicy.Scope {
+            case LimitedHSTSHeaderPolicyScope: // more details on this later in the EP doc
+                env = append(env, corev1.EnvVar{Name: RouterHSTSHeaderScope, Value: LimitedHSTSHeaderPolicyScope})
+            case AllHSTSHeaderPolicyScope:
+                env = append(env, corev1.EnvVar{Name: RouterHSTSHeaderScope, Value: AllHSTSHeaderPolicyScope})
+        }
+        }
+    }
+...
+```
+
+We don't do any other header validation, but if needed it could go in here:
+https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/controller.go#L441
+
+Accompanying unit and e2e test changes are added to exercise the new header.
+
+A revised HAProxy template is added which will use a new helper method called `hstsHeaderValue`, defined later.
+With the addition of processing for `ROUTER_DEFAULT_HSTS`, the code template snippet above becomes:
+```gotemplate
+{{- $router_global_hsts := env "ROUTER_DEFAULT_HSTS" "" }}
+...
+{{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}
+  {{- /* Get the HSTS header value using the default and the annotations in the cfg */ }}
+  {{- with $hsts := hstsHeaderValue $router_global_hsts "" $cfg "" }}
+    http-response set-header Strict-Transport-Security {{$hsts}}
+  {{- end }}{{/* hstsHeaderValue */}}
+{{- end }}{{/* is "edge" or "reencrypt" */}}
+```
+
+#### Enable HSTS per-domain
+Some of the proposal for this goal is in the section above.  In order to allow the HAProxy template to add HSTS Headers to
+a limited list of domains, it will use the previously mentioned field `InScopeDomains` in the `IngressControllerHSTSHeaderPolicy`
+type.
+
+The details on the processing for limited scope in `desiredRouterDeployment` are completed here.
+Add one more environment variable, `ROUTER_HSTS_LIST`, which will contain the list of domains in `InScopeDomains`:
+```go
+	const RouterHSTSHeaderList        = "ROUTER_HSTS_LIST"
+
+       if scope ==  LimitedHSTSHeaderPolicyScope && len(ci.Spec.HTTPHeaers.HSTSHeaderPolicy.InScopeDomains) > 0 {
+            env = append(env, corev1.EnvVar{Name: RouterHSTSHeaderScope, Value: LimitedHSTSHeaderPolicyScope})
+            env = append(env, corev1.EnvVar{Name: RouterHSTSHeaderList, Value: strings.Join(ci.Spec.HTTPHeaders.HSTSHeaderPolicy.InScopeDomains, ";")})
+       }
+```
+As shown in this example, the HSTS Policy scope is for a limited list of two domains, expires after one year, includes
+subdomains, and requests preload:
+```yaml
+apiVersion: operator.openshift.io/v1
+kind: IngressController
+metadata:
+  name: default
+  namespace: openshift-ingress-operator
+spec:
+  httpHeaders:
+    hstsHeaderPolicy:
+      scope: Limited
+      inScopeDomains:
+        - domainA
+        - domainB
+      maxAgeSeconds: 31536000
+      hstsDirectives:
+        - includeSubDomains
+        - preload
+```
+
+With the addition of processing for `ROUTER_HSTS_SCOPE` of `limited`, and `ROUTER_HSTS_LIST`, the HAProxy template code
+snippet above will gain a few new variables:
+```gotemplate
+{{- $router_global_hsts := env "ROUTER_DEFAULT_HSTS" "" }}
+{{- $router_hsts_scope := env "ROUTER_HSTS_SCOPE" "" }}
+{{- $router_hsts_list := env "ROUTER_HSTS_LIST" }}
+...
+{{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}
+  {{- /* Get the HSTS header value using the default and the annotations in the cfg */ }}
+    {{- with $hsts := hstsHeaderValue $router_global_hsts $router_hsts_scope $cfg $router_hsts_list }}
+      http-response set-header Strict-Transport-Security {{$hsts}}
+    {{- end }}{{/* hstsHeaderValue */}}
+{{- end }}{{/* is "edge" or "reencrypt" */}}
+```
+
+#### Disable HSTS per-route when it is enabled globally
+As mentioned previously, the `Strict-Transport-Security` annotation called
+`haproxy.router.openshift.io/hsts_header` has a required directive `max-age` that determines
+its lifespan.  A per-route override on globally enabled HSTS can be indicated by setting the `max-age`
+value in the route's annotation to zero, as shown in this example:
+```yaml
+metadata:
+  annotations:
+    haproxy.router.openshift.io/hsts_header: max-age=0
+```
+
+The HAProxy template remains the same, and the promised new function `hstsHeaderValue` will be added:
+
+```go
+// If the override annotation string is set, return it - the annotation overrides the default
+// Otherwise, if the scope is limited and the domain is in the list - return the default
+// Finally, if the scope is all - return the default
+// Otherwise, return the empty string
+func hstsHeaderValue(default_header string, hsts_scope string, config ServiceAliasConfig, hsts_list string) string {
+    wantLimited := false
+
+	if override, ok := config.Annotations["haproxy.router.openshift.io/hsts_header"]; ok {
+        return override
+    } else if hsts_scope == LimitedHSTSHeaderPolicyScope {
+        hsts_array := Strings.split(hsts_list, ";")
+        for domain := range hsts_array {
+            if strings.HasSuffix(config.Host, domain) {
+                wantLimited = true
+                break
+            }
+        }
+        if wantLimited {
+            return default_header
+        }
+    } else if hsts_scope == AllHSTSHeaderPolicyScope {
+        return default_header
+    }
+    return ""
+}
+```
+
+### User Stories
+
+#### As a cluster administrator, I want to enable HSTS globally, for all TLS routes
+Update the `IngressController` deployment spec, for example, setting `scope` to `All`, and configuring the rest of the
+HSTS Policy.
+```yaml
+spec:
+  httpHeaders:
+    hstsHeaderPolicy:
+      scope: All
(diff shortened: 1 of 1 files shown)


enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,252 @@
...
+...
+{{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}
+    {{- with $hsts := firstMatch $hstsPattern (index $cfg.Annotations "haproxy.router.openshift.io/hsts_header") }}
+  http-response set-header Strict-Transport-Security {{$hsts}}
+    {{- end }}{{/* hsts header */}}
+  {{- end }}{{/* is "edge" or "reencrypt" */}}
+````
+#### Enable HSTS for all routes globally
+To add the header to all HTTP responses from an `edge` or `reencrypt` route, add a new router environment 
+variable called `ROUTER_DEFAULT_HSTS`, configured by the administrator in the Ingress Operator.  `ROUTER_DEFAULT_HSTS` is detected in the HAProxy configuration file processing, and can be overridden
```
sgreene570, line 126: > add a new router environment  variable called `ROUTER_DEFAULT_HSTS`, configured by the administrator in the Ingress Operator


We need some sort of user-facing API so that a cluster administrator can tell the ingress operator to set the `ROUTER_DEFAULT_HSTS` environment variable for an ingress controller's deployment. In general, we don't want cluster administrators manually modifying router deployments by hand.


I think it would make sense to add an option for requiring HSTS when possible to the Ingress Controller API. This would allow a cluster administrator to trivially enable "global" HSTS for individual ingress controllers, and would allow for different HSTS settings for different ingress controllers. How does that sound to you?

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,252 @@
...
+    - If set to 0, it negates the effect and the host is no longer regarded as an HSTS host
+    - `max-age` decrements over time
+- `includeSubDomains`: if present, applies to any subdomains of the host's domain name
+- `preload`: if present, tells the client to include this host in its host preload list so that 
+it never needs to do an initial load to get the HSTS header (note that this is not defined in RFC 6797 
+and is therefore client implementation-dependent)
+
+HSTS is currently implemented as a pattern in the HAProxy template and applied to `edge` and 
+`reencrypt` routes that match the `haproxy.router.openshift.io/hsts_header` annotation:
+````
```
sgreene570, line 113: I believe you want to use triple backticks for code fencing (as opposed to quadruple backticks).

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,252 @@
...
+  {{- end }}{{/* global_hsts */}}
+  {{- /* Override if the annotation exists.  Set-header replaces any existing header */ }}
+  {{- with $hsts := firstMatch $hstsPattern (index $cfg.Annotations "haproxy.router.openshift.io/hsts_header") }}
+    http-response set-header Strict-Transport-Security {{$hsts}}
+  {{- end }}{{/* hsts header override per-route */}}
+{{- end }}{{/* is "edge" or "reencrypt" */}}
+````
+
+#### Enable HSTS on a node-type basis (infra-only, worker-only, or both)
+(TBD) Clarification has been requested on this requirement.
```
sgreene570, line 153: Being able to enable HSTS on a per ingress controller basis may satisfy the customer's desire to be able to apply different HSTS settings for particular subsets of routes (ie routes that target workloads running on infra nodes, as long as all of those routes are served by the same ingress controller). Based on the [NE-310](https://issues.redhat.com/browse/NE-310) epic description, I get the impression that the customer wants to be able to apply unique HSTS settings for the default ingress controller, but we should try to clarify what exactly their requirements are.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Test Plan
+HAProxy and Router have unit test coverage; for this enhancement, the unit tests are
+expanded to cover the additional four goals.
+
+The operator has end-to-end tests; for this enhancement, add the following tests:
+
+1. Create an IngressController that enables global HSTS
+2. Open a connection to this route and send a request.
+3. Verify that a response is received and that the headers include HSTS
```
sgreene570, line 235: Perhaps an e2e test for this enhancement could also verify that HSTS can be disabled on a per route basis using an annotation as you mentioned.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,252 @@
...
+metadata:
+  annotations:
+    haproxy.router.openshift.io/hsts_header: max-age=0     
+````
+
+### User Stories
+
+#### As a cluster administrator, I want to enable HSTS for all routes globally
+Add an environment variable to the IngressController's deployment spec, for example, setting 
+`ROUTER_DEFAULT_HSTS` to `max-age=31536000`
```
sgreene570, line 194: I think there may be some trailing spaces in this section that are throwing lint errors in the link CI job.

enhancements/ingress/global-options-enable-hsts.md
sgreene570: Yup, something along those lines id imagine.

enhancements/ingress/global-options-enable-hsts.md
sgreene570: Ah I guess this is true (just as long as the entire doc is consistent :grin: )

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,298 @@
...
+the HSTS Policy expires after one year and includes subdomains:
+```yaml
+apiVersion: operator.openshift.io/v1
+kind: IngressController
+metadata:
+  name: default
+  namespace: openshift-ingress-operator
+spec:
+  httpHeaders:
+    hstsHeaderPolicy: "max-age=31536000;includeSubDomains"
```
sgreene570, line 145: We probably want to create a more formal API for specifying HSTS settings for an ingress controller. That way we could perform some validation on user-provided values, and ensure that all of the necessary fields are specified. The ingress-operator could then stringify the API fields into a string to be fed to the router pods via an environment variable (or maybe we could use multiple environment variables to do so). Thoughts?

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,298 @@
...
+{{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}
+    {{- with $hsts := firstMatch $hstsPattern (index $cfg.Annotations "haproxy.router.openshift.io/hsts_header") }}
+  http-response set-header Strict-Transport-Security {{$hsts}}
+    {{- end }}{{/* hsts header */}}
+  {{- end }}{{/* is "edge" or "reencrypt" */}}
+```
+#### Enable HSTS for all routes globally
+To add the header to all HTTP responses from an `edge` or `reencrypt` route, add a new router environment
+variable called `ROUTER_DEFAULT_HSTS`.  To the `IngressController` custom resource definition API defined in
+`manifests/00-custom-resource-definition.yaml`, add a new `httpHeaders` property called `hstsHeaderPolicy` to
```
sgreene570, line 127: The canonical location of the IngressController API is https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,298 @@
...
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+In 3.x and 4.x customers can [provide a per-route annotation to enable HSTS](
+https://docs.openshift.com/container-platform/4.4/networking/routes/route-configuration.html#nw-enabling-hsts_route-configuration).  For
+customers with many routes or regulatory compliance issues, the manual per-route annotation is
+problematic.
+
+This enhancement extends the Route API to allow cluster administrators to
```
sgreene570, line 56:
```suggestion
This enhancement extends the IngressController API to allow cluster administrators to
```

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,298 @@
...
+### Graduation Criteria
+N/A
+
+### Upgrade / Downgrade Strategy
+On upgrade, any previous per-route HSTS configuration remains in effect.
+
+If a cluster administrator applied any of the new HSTS configuration options
+in 4.8, then downgraded to 4.7, the HSTS configuration settings would be ignored
+except for the per-route HSTS configurations.  The administrator would be responsible
+for removing any per-route HSTS configurations that were no longer applicable.
```
sgreene570, line 292: Might be worth mentioning that an end-users browser, depending on the HSTS max-age setting, may still remember to access routes via HTTPS on downgrade (that is, until the HSTS record expires in the user's browser).

enhancements/ingress/global-options-enable-hsts.md
candita: Agreed.  I was working on this today, and decided to leave this version as an alternative.

enhancements/ingress/global-options-enable-hsts.md
candita: In this case, they clarified that they would like it on a per-router basis.  I updated the doc for this case, can you ptal?

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+```yaml
+apiVersion: operator.openshift.io/v1
+kind: IngressController
+metadata:
+  name: default
+  namespace: openshift-ingress-operator
+spec:
+  httpHeaders:
+    hstsHeaderPolicy:
+      scope: all
```
Miciah, line 274: The value for `scope` should be `limited` in this example, right?
```suggestion
      scope: limited
```

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+```gotemplate
+{{- $router_hsts := env "ROUTER_DEFAULT_HSTS" "" }}
+{{- $router_hsts_scope := env "ROUTER_HSTS_SCOPE" "" }}
+{{- $router_hsts_list := env "ROUTER_HSTS_LIST" }}
+...
+{{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}
+  {{- /* First check the environment to see if the header is set globally */}}
+  {{- if (ne "" $router_hsts) }}
+    {{- /* Now check the environment to see if scope is all, or limited but request domain is in the limited list */}}
+    {{- if or (eq "all" $router_hsts_scope) (and (eq "limited" $router_hsts_scope) (eq true (InList req.hdr(Host) $router_hsts_list)) )) }}
```
Miciah, line 294: Everything inside the double-braces (`{{ ... }}`) is Go template, and `req.hdr` is an HAProxy configuration directive, so this will not work properly.  It may be necessary to write out a map file and define an acl (`acl hsts req.hdr(host),lower,map(/var/lib/haproxy/conf/os_hsts.map) -m found`) to use below (`http-response set-header Strict-Transport-Security {{$router_hsts}} if hsts`).

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+variable called `ROUTER_DEFAULT_HSTS`.
+
+Add a new type `IngressControllerHSTSHeaderPolicy` to the API in
+`openshift/api/blob/master/operator/v1/types_ingress.go`, to capture the configuration of the HSTS Policy.
+```go
+type IngressControllerHSTSHeaderPolicy {
+    // max-age is the delta time range in seconds during which the host is regarded as an HSTS host
+	// If set to 0, it negates the effect and the host is no longer regarded as an HSTS host.
+	// max-age` decrements over time.
+    // +required
```
Miciah, line 133: The indentation is wonky here.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+    MaxAge             string `json:"max-age"`
+
+    // scope indicates whether this header policy is enabled on a per-router basis (i.e. some
+    // routers will use HSTS and some will not) or it is global across every router in the cluster
```
Miciah, line 137: I think the terminology is inaccurate here, but I may be misinterpreting.  How's this?
```suggestion
    // scope indicates whether this header policy is enabled for routes that have hosts with domains
    // in the list specified in inScopeRouters or for all routes that have hosts under the cluster
```

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+    // ingress domain.  If per-router HSTS configuration scope is desired, set the scope field to "limited".
+    // If global HSTS is desired, set the scope field to "all".  The default is "all".
+    // +optional
+    Scope               string `json:"scope"`
```
Miciah, line 141: Define a new string type that has a `+kubebuilder:validation:Enum` tag.  Search for "Enum" in this file for examples.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+        if  ok = ci.Spec.HTTPHeaders.HSTSHeaderPolicy.Preload; ok != nil && ok {
+        	header = header +  ";preload"
+        }
+        env = append(env, corev1.EnvVar{Name: RouterHSTSHeaderPolicy, Value: header})
+        scope = ci.Spec.HTTPHeaders.HSTSHeaderPolicy.Scope
+        if scope == "limited" { // more details on this later in the EP doc
+            env = append(env, corev1.EnvVar{Name: RouterHSTSHeaderScope, Value: "limited"})
+        } else {
+            env = append(env, corev1.EnvVar{Name: RouterHSTSHeaderScope, Value: "all"})
+        }
```
Miciah, line 222: Use `switch ci.Spec.HTTPHeaders.HSTSHeaderPolicy.Scope {`.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+    // max-age is the delta time range in seconds during which the host is regarded as an HSTS host
+	// If set to 0, it negates the effect and the host is no longer regarded as an HSTS host.
+	// max-age` decrements over time.
+    // +required
+    MaxAge             string `json:"max-age"`
+
+    // scope indicates whether this header policy is enabled on a per-router basis (i.e. some
+    // routers will use HSTS and some will not) or it is global across every router in the cluster
+    // ingress domain.  If per-router HSTS configuration scope is desired, set the scope field to "limited".
+    // If global HSTS is desired, set the scope field to "all".  The default is "all".
```
Miciah, line 139: Should there be an "off" value?  Hm, I guess `max-age: 0` is the "off" value?  I think it would be clearer to make `scope` a tri-state with off, limited, and all values.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+to separately apply HSTS globally to worker nodes and infrastructure nodes (e.g.
+Prometheus, Grafana, web console hosts), while retaining the ability to opt-out
+of HSTS on a per-route basis.
+
+### Goals
+
+This proposal allows administrators to:
+- Continue to enable HSTS per-route (retain the existing feature functionality)
+- Enable HSTS for all routes globally, based on the cluster's ingress domain
+- Enable HSTS per-router
```
Miciah, line 79: Is it "per-router" or "per domain"?  This requirement is very confusing.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+- Disable HSTS per-route when it is enabled globally
+
+### Non-Goals
+
+- HSTS cannot be applied to non-TLS routes, even if HSTS is requested for all routes globally.  As
+mentioned in [RFC 6797 Section 8.1](https://tools.ietf.org/html/rfc6797#section-8.1), if an HTTP
+response is received over insecure transport, clients must ignore any present STS headers.
+- The current implementation has a `preload` directive for the `Strict-Transport-Security`
+header.  This is not an RFC 6797 directive and therefore its implementation may vary by
+browser.  It can be specified, but support may vary by implementation.
```
Miciah, line 89: Is this a non-goal?  Looks like it belongs under "Risks and Mitigations" or "Implementation Details/Notes/Constraints".

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+Each goal is addressed in the subsections below.
+
+### Continue to enable HSTS per-route (retain the existing feature functionality)
+The `Strict-Transport-Security` response header is currently applied per-route via an
+annotation called `haproxy.router.openshift.io/hsts_header`.  This annotation has a
+required directive `max-age`, and optional directives `includeSubDomains` and `preload`,
+with the following definitions:
+- `max-age`: delta time range in seconds during which the host is regarded as an HSTS host
+  - If set to 0, it negates the effect and the host is no longer regarded as an HSTS host
+  - `max-age` decrements over time
```
Miciah, line 102: The value that HAProxy sends doesn't decrement from one response to the next, does it?  Do you mean to say that `max-age` specifies a timeout period?

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+
+Add a new type `IngressControllerHSTSHeaderPolicy` to the API in
+`openshift/api/blob/master/operator/v1/types_ingress.go`, to capture the configuration of the HSTS Policy.
+```go
+type IngressControllerHSTSHeaderPolicy {
+    // max-age is the delta time range in seconds during which the host is regarded as an HSTS host
+	// If set to 0, it negates the effect and the host is no longer regarded as an HSTS host.
+	// max-age` decrements over time.
+    // +required
+    MaxAge             string `json:"max-age"`
```
Miciah, line 134:
```suggestion
    MaxAge             metav1.Duration `json:"maxAge"`
```
or
```suggestion
    MaxAgeSeconds      int32 `json:"maxAgeSeconds"`
```

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+
+    // includeSubDomains means the HSTS Policy applies to any subdomains of the host's domain name
+    // +optional
+    IncludeSubDomains   bool `json:"includeSubDomains,omitempty"`
+
+    // preload tells the client to include this host in its host preload list so that
+    // it never needs to do an initial load to get the HSTS header (note that this is not defined
+    // in RFC 6797 and is therefore client implementation-dependent).
+    // +optional
+    Preload             bool `json:"preload,omitempty"`"`
```
Miciah, line 157: I'm tempted to suggest making this a list of attributes:
```go
    additionalAttributes []hstsAttribute `json:"additionalAttributes,omitempty"`"`
    // ...
}

// hstsAttribute is an attribute that can be specified on HTTP Strict-Transport-Security headers.
// +kubebuilder:validation:Enum=preload;includeSubDomains
type hstsAttribute string
```
What do you think?  Might want to check with someone with good taste and familiarity with API conventions.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+...
+{{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}
+  {{- /* First check the environment to see if the header is set globally */}}
+  {{- if (ne "" $router_global_hsts) }} 
+    http-response set-header Strict-Transport-Security {{$router_global_hsts}}
+  {{- end }}{{/* global_hsts */}}
+  {{- /* Override if the annotation exists.  Set-header replaces any existing header */ }}
+  {{- with $hsts := firstMatch $hstsPattern (index $cfg.Annotations "haproxy.router.openshift.io/hsts_header") }}
+    http-response set-header Strict-Transport-Security {{$hsts}}
+  {{- end }}{{/* hsts header override per-route */}}
```
Miciah, line 244: It would be preferable to avoid writing two `http-response set-header Strict-Transport-Security` lines if both the global setting and the annotation are set:
```suggestion
  {{- with $hsts := firstMatch $hstsPattern (index $cfg.Annotations "haproxy.router.openshift.io/hsts_header") $router_global_hsts }}
    http-response set-header Strict-Transport-Security {{$hsts}}
  {{- end }}
```

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,411 @@
...
+1. Create an IngressController that enables global HSTS
+2. Open a connection to this route and send a request
+3. Verify that a response is received and that the headers include HSTS as configured in step 1
+
+#### Override global HSTS on an individual route
+1. Create an IngressController that enables global HSTS
+2. Create a route that disables global HSTS with an annotation
+3. Open a connection to this route and send a request
+4. Verify that a response is received and that the headers do not include HSTS as configured in step 2
+
```
Miciah, line 391: How about test coverage for `inScopeRouters`?

Miciah (commented):
The terminology is confusing in a few places.  In particular, I think you wrote "router" few times where you meant "route" (or "route host") with respect to scope.

enhancements/ingress/global-options-enable-hsts.md
candita: Yes, the header specifies an HSTS Policy with a timeout included.  It is a time-to-live that the user agent must implement.  I will update this to make that clear.

enhancements/ingress/global-options-enable-hsts.md
sgreene570: I interpreted the comments in https://issues.redhat.com/browse/NE-310 to suggest that enabling HSTS on a per-ingress controller level would be sufficient. Am I missing something? It's not clear to be if filtering out subdomains of an ingress controller's domain when assigning HSTS settings is strictly necessary to implement this RFE.

enhancements/ingress/global-options-enable-hsts.md
candita: Let me think about that a bit.

enhancements/ingress/global-options-enable-hsts.md
candita: I will change this to per domain, and rename the field from inScopeRouters to inScopeDomains.

They originally requested per-node application of HSTS, so that only infra and application nodes could be selected.  I believe we need to keep per-domain to help with that requirement.  If you think this adds too much to the scope of the EP, let me know and I will ask Marc and Mike if we can split off the work into another RFE.

enhancements/ingress/global-options-enable-hsts.md
candita: I felt wishy-washy that the three header components `max-age`, 'preload`, and 'includeSubDomains` did not belong in the same grouping as scope components `scope` and `inScopeDomains`, so I like the suggestion.  Maybe separating these into `headerDirectives` and `scopeAttributes`.  Would @deads2k be a good person to ask?

enhancements/ingress/global-options-enable-hsts.md
sgreene570: I don't think this adds too much scope to this enhancement. I'm just not sure if I completely understand why being able to filter out domains for HSTS settings on an ingress controller is helpful. If a user wants to use different HSTS settings for 2 different sets of routes, they would have to have more than one ingress controller, right (and therefore, the scope settings in this enhancement wouldn't really be useful for them)? The scope fields proposed in this enhancement would give a user the ability to only enable HSTS in any capacity for a strict subset of routes served by a particular ingress controller, right (which i'm assuming is behavior we definitely want to implement)?

enhancements/ingress/global-options-enable-hsts.md
candita: I have updated the design to use a helper function for the HAproxy template, `hstsHeaderValue`.  Needed to revert the `enums` for scope, back to `string consts.`

enhancements/ingress/global-options-enable-hsts.md
candita: Updated this to use a `hstsDirectives` list

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+const (
+    AllHSTSHeaderPolicyScope         = "All"
+    LimitedHSTSHeaderPolicyScope     = "Limited"
+)
+
+// HSTSDirectiveType is an optional directive that directs HSTS Policy configuration.
+//+kubebuilder:validation:Enum=preload;includeSubDomains
+type HSTSDirectiveType string
+const(
+    // PreloadHSTSDirective directs the client to include this host in its host preload list so that
```
Miciah, line 137: The phrase "this host" doesn't work in this context as this is a global setting that applies to all hosts or to all hosts that match some domain in a list.  It probably isn't helpful to spell that all out here, so maybe just change "this host" to "hosts":
```suggestion
    // PreloadHSTSDirective directs the client to include hosts in its host preload list so that
```

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+
+    // IncludeSubDomainsHSTSDirective means the HSTS Policy applies to any subdomains of the host's domain name
+    IncludeSubDomainsHSTSDirective   HSTSDirectiveType = "includeSubDomains"
+)
+
+type IngressControllerHSTSHeaderPolicy {
+	// maxAgeSeconds is the delta time range in seconds during which the host is regarded as an HSTS host.
+	// If set to 0, it negates the effect and the host is no longer regarded as an HSTS host.
+	// maxAgeSeconds is a time-to-live value, and if it is not refreshed, the HSTS policy for this host
+	// will eventually expire.
```
Miciah, line 150:
```suggestion
	// maxAgeSeconds is the delta time range in seconds during which hosts are regarded as HSTS hosts.
	// If set to 0, it negates the effect and hosts are no longer regarded as HSTS hosts.
	// maxAgeSeconds is a time-to-live value, and if this policy is not refreshed on a client, the HSTS policy
	// will eventually expire on that client.
```

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+	// +required
+    MaxAgeSeconds       int32 `json:"maxAgeSeconds"`
+
+    // HSTSDirectives is a list of optional HSTS directives that direct HSTS Policy configuration.
+    HSTSDirectives      []HSTSDirectiveType `json:"hstsDirectives,omitempty"`
+
+    // scope indicates whether this header policy is enabled for routes that have hosts with domains
+    // in the list specified in inScopeDomains (limited scope) or for all routes that have hosts under the
+    // cluster ingress domain (all scope).  If global HSTS is desired, set the scope field to "all".  The
+    // default is "all".
```
Miciah, line 160: We don't want to start enabling HSTS everywhere when a user upgrades a cluster, so either the default should be some new "None" value, or the default should be "Limited" (which effectively means "None" as `inScopeDomains` defaults to empty).

Miciah, line 160:
```suggestion
    // cluster ingress domain (all scope).  If global HSTS is desired, set the scope field to "All".  The
    // default is "All".
```

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+    // +optional
+    Scope               string `json:"scope,omitempty"`
+
+    // inScopeDomains is a list of router domains that require HSTS, when not all routes require HSTS.
+    // If scope = All then this field is ignored.  If scope = Limited, HSTS policy is limited to the domains
+    // in the inScopeDomains list.  If scope = All, this field is ignored.
+    // +optional
+    InScopeDomains      []string 'json:"inScopeDomains,omitempty"'
```
Miciah, line 168:
```suggestion
    InScopeDomains      []string `json:"inScopeDomains,omitempty"`
```

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+      maxAgeSeconds: 31536000
+      hstsDirectives:
+        - includeSubDomains
+        - preload
+```
+
+With the addition of processing for `ROUTER_HSTS_SCOPE` of `limited`, and `ROUTER_HSTS_LIST`, the HAProxy template code
+snippet above will gain a few new variables:
+```gotemplate
+{{- $router_hsts := env "ROUTER_DEFAULT_HSTS" "" }}
```
Miciah, line 299:
```suggestion
{{- $router_global_hsts := env "ROUTER_DEFAULT_HSTS" "" }}
```
Or rename `$router_global_hsts` everywhere else.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+// Finally, if the scope is all - return the default
+// Otherwise, return the empty string
+func hstsHeaderValue(default_header string, hsts_scope string, config ServiceAliasConfig, hsts_list []string) string {
+    wantLimited := false
+
+	if override, ok := config.Annotations["haproxy.router.openshift.io/hsts_header"]; ok {
+        return override
+    } else if hsts_scope == LimitedHSTSHeaderPolicyScope {
+        for domain := range hsts_list{
+            if strings.Contains(config.Host, domain) {
```
Miciah, line 336: It would be better to parse `ROUTER_HSTS_LIST` into a `sets.String` value in the template plugin setup when the router starts in order to avoid doing a lot of string manipulation or looping when generating `haproxy.config`, but we can work out those implementation details later on.  

Using the host from the `ServiceAliasConfig` is a terrific idea—much simpler than writing a map and using HAProxy acls.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+type.
+
+The details on the processing for limited scope in `desiredRouterDeployment` are completed here.
+Add one more environment variable, `ROUTER_HSTS_LIST`, which will contain the list of domains in `InScopeDomains`:
+```go
+	const RouterHSTSHeaderList        = "ROUTER_HSTS_LIST"
+
+       if scope ==  LimitedHSTSHeaderPolicyScope {
+            env = append(env, corev1.EnvVar{Name: RouterHSTSHeaderScope, Value: LimitedHSTSHeaderPolicyScope})
+            env = append(env, corev1.EnvVar{Name: RouterHSTSHeaderList, Value: ci.Spec.HTTPHeaders.HSTSHeaderPolicy.InScopeDomains})
```
Miciah, line 272: Presumably you will need something like `strings.Join(ci.Spec.HTTPHeaders.HSTSHeaderPolicy.InScopeDomains, ";")` to be able to store it in an environment variable.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+type IngressControllerHSTSHeaderPolicy {
+	// maxAgeSeconds is the delta time range in seconds during which the host is regarded as an HSTS host.
+	// If set to 0, it negates the effect and the host is no longer regarded as an HSTS host.
+	// maxAgeSeconds is a time-to-live value, and if it is not refreshed, the HSTS policy for this host
+	// will eventually expire.
+	// +required
+    MaxAgeSeconds       int32 `json:"maxAgeSeconds"`
+
+    // HSTSDirectives is a list of optional HSTS directives that direct HSTS Policy configuration.
+    HSTSDirectives      []HSTSDirectiveType `json:"hstsDirectives,omitempty"`
```
Miciah, line 155: You can specify that the items need to be unique using the following tag: 
```go
    // +kubebuilder:validation:UniqueItems=true
```

Miciah (commented):
This is looking great!  Minor comments aside, my one big concern is about the default—when the cluster administrator upgrades to a version of OpenShift that includes this enhancement, we don't want to spring a surprise by turning HSTS on everywhere unprompted.  HSTS should be opt-in.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+
+    // scope indicates whether this header policy is enabled for routes that have hosts with domains
+    // in the list specified in inScopeDomains (limited scope) or for all routes that have hosts under the
+    // cluster ingress domain (all scope).  If global HSTS is desired, set the scope field to "all".  The
```
sgreene570, line 159: Would `All` scope enable the given HSTS settings for all of the routes served by the given Ingress Controller?

Miciah:
BTW, please add "ingress:" to the title of the PR.  (Someone asked me to do that on some of my ingress enhancement PRs a while back.)

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+of HTTPS in client requests to the host, without having to use an HTTP redirect.
+HSTS provides user protection and is concerned with minimizing security threats based on network
+traffic eavesdropping and man-in-the-middle attacks.  Using a response header
+called `Strict-Transport-Security`, the HTTP response informs clients that a website
+can only be accessed via HTTPS.  
+
+Administrators who are tasked with route management and/or regulatory compliance
+issues want an option to automatically and globally enable HSTS. They also want
+to separately apply HSTS globally to worker nodes and infrastructure nodes (e.g.
+Prometheus, Grafana, web console hosts), while retaining the ability to opt-out
```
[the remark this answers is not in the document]
candita: This is the Motivation section, so I was describing the requester's motivation.  I suppose that could be confusing.  I can tweak it while I'm making other changes.

enhancements/ingress/global-options-enable-hsts.md
candita: Yes, it would add the HSTS header to all http responses for edge and reencrypt routes under the cluster ingress domain.  I added some more to clarify:

If HSTS is desired for all routes, set the scope field to "All". If HSTS is desired for only a limited set of routes, set the scope field to "Limited" and add a list of subdomains to the InScopeDomains field.

enhancements/ingress/global-options-enable-hsts.md
candita: You are right.  The default will be none if Scope is not specified.

enhancements/ingress/global-options-enable-hsts.md
sgreene570: > Yes, it would add the HSTS header to all http responses for edge and reencrypt routes under the cluster ingress domain. I added some more to clarify:

What if an Ingress Controller is serving routes with hosts that aren't under the cluster's ingress domain? Would `all` exclude those routes?

enhancements/ingress/global-options-enable-hsts.md
candita: I'll keep it a semi-colon delimited string to be passed in as an env variable, and do the parsing in hstsHeaderValue.
Thanks for all your comments!

enhancements/ingress/global-options-enable-hsts.md
candita: I reworked this.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,457 @@
...
+
+### Continue to enable HSTS per-route (retain the existing feature functionality)
+The `Strict-Transport-Security` response header is currently applied per-route via an
+annotation called `haproxy.router.openshift.io/hsts_header`.  This annotation has a
+required directive `max-age`, and optional directives `includeSubDomains` and `preload`,
+with the following definitions:
+- `max-age`: delta time range in seconds during which the host is to be regarded as an HSTS host
+  - If set to 0, it negates the effect and the host is no longer regarded as an HSTS host
+  - `maxAgeSeconds` is a time-to-live value, and if it is not refreshed, the HSTS policy for this host will
+  eventually expire
```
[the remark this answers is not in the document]
candita: I changed the comments so that scope is clearly opt-in, and reworked some of the code.

enhancements/ingress/global-options-enable-hsts.md
candita: No, if scope is "All" then all routes will get the configured HSTS Policy header.  I re-wrote this, ptal.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,466 @@
...
+Administrators who are tasked with route management and/or regulatory compliance
+issues want an option to automatically and globally enable HSTS. They also want
+to separately apply HSTS globally to specific routers that target applications,
+while retaining the ability to opt-out of HSTS on a per-route basis.
+
+### Goals
+
+This proposal allows administrators to:
+- Continue to enable HSTS per-route (retain the existing feature functionality)
+- Enable HSTS for all routes globally, based on the cluster's ingress domain
```
Miciah, line 77:
```suggestion
- Enable HSTS for all TLS routes
```

enhancements/ingress/global-options-enable-hsts.md
Miciah: It still says "for all routes that have hosts under the cluster ingress domain (all scope)".  Also, the second goal is, "Enable HSTS for all routes globally, based on the cluster's ingress domain".  I think in both cases that should be "all TLS routes" without any reference to the cluster's ingress domain, right?

Miciah: Also, the heading "Enable HSTS for all routes globally, based on the cluster's ingress domain" should omit "based on the cluster's ingress domain" if "All" means includes hosts that aren't under the ingress domain.

enhancements/ingress/global-options-enable-hsts.md
```diff
@@ -0,0 +1,466 @@
...
+    // domains in the inScopeDomains list.
+    // +optional
+    InScopeDomains      []string `json:"inScopeDomains,omitempty"`
+}
+```
+
+Validation and formatting in the `IngressController`
+would change the fields into a semi-colon delimited string with formatting:
+```yaml
+max-age=<numberofseconds>[;includeSubDomains[;preload]]`
```
Miciah, line 183: `preload` could be specified without `includeSubDomains`, right?
```suggestion
max-age=<numberofseconds>[;includeSubDomains][;preload]`
```
(That suggestion still isn't exactly right because the ordering doesn't really matter, but it's close enough.)

enhancements/ingress/global-options-enable-hsts.md
candita: Fixed.

PR #687, opened 2021-03-10
Add AWS EFS CSI driver operator

jsafrane:
AWS EFS CSI driver (+ corresponding operator) is an optional OCP component. It should be installed through OLM, when
users opts-in.

This document describes existing (unsupported) solution and how to turn it into a supported one, both from code (and
build and shipment), and from user perspective.


enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+## Proposal
+High level only, see design details below.
+
+1. Rewrite current aws-efs-operator to use [CSI driver installation functions of library-go](https://github.com/openshift/library-go/tree/master/pkg/operator/csi/csicontrollerset).
+   * This allows us to share well tested code to install CSI drivers and implement bugfixes / new features at a single
+     place.
+   * It will add common features of CSI driver operators which are missing in the current community operator, such as
+     proxy injection and restarting the drivers when cloud credentials change.
+
+2. Update (rewrite) current aws-efs-operator to watch `ClusterCSIDriver` CR named `efs.csi.aws.com`.
```
2uasimojo, line 108: Will aws-efs-operator ever *write* (status, spec, labels, annotations, ...) to the CR? (Concerns about one CRD having multiple "owners".)

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+4. Ship the operator + CSI driver through ART pipeline.
+   * Most of the operator just installs / removes CSI driver, i.e. manages driver's RBAC rules, CredentialsRequest,
+     DaemonSet, Deployment, CSIDriver and StorageClasses. See existing
+     [AWS EBS operator](https://github.com/openshift/aws-ebs-csi-driver-operator/tree/master/assets) for objects it
+     manages, EFS should be very similar in this regard.
+   * `SharedVolume` processing on top of it is very trivial, it just creates PV + PVC.
+
+5. Ship [efs-utils](https://github.com/aws/efs-utils) as a base image for the EFS CSI driver image.
+
+6. Update installer to delete Access Points and EFS volumes that have `owner` tag when destroying a cluster.
```
2uasimojo, line 127: This feels wrong. We've talked about the barriers (networking, permissions, etc.) that would make this difficult for the operator to manage. IMO for those reasons -- and for symmetry -- all provisioning/deprovisioning on the AWS side should be outside of the purview of the operator.

(To be clear, I agree CI must have the ability to provision and deprovision -- more on this below -- but I don't think this tagging approach is the way to go.)

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Implementation Details/Notes/Constraints [optional]
+
+### Risks and Mitigations
+
+Currently, all cloud-based CSI drivers shipped by OCP are installed when a cluster is installed on the particular cloud.
+AWS EFS will not be installed by default. We will include a note about opt-in installation in docs.
+
+After installation of the new operators, either as a fresh install or upgrading the existing one, users must create the
+CR when the operator is installed on a new cluster. This is something that was not necessary in the previous releases of
```
2uasimojo, line 140: Need to see proposed changes to the CRD, if any, and an example of what a CR will look like to (de)activate the EFS CSI driver. (Not sure if that level of detail is appropriate for this document, but you included dockerfile snippets so.... :) )

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+## Design Details
+
+### Rewrite to library-go & ClusterCSIDriver CRD
+
+This should be pretty straightforward. We already ship number of CSI drivers that use library-go. Example usage:
+[AWS EBS CSI driver](https://github.com/openshift/aws-ebs-csi-driver-operator/blob/383a7638b37a4b9a9831c7747e8c499eedcf030f/pkg/operator/starter.go#L67).
+EFS CSI driver operator should not be much more complicated.
+
+Library-go does not support un-installation of CSI driver (and static resources created by `StaticResourcesController`),
+this must be implemented there.
```
2uasimojo, line 152: What happens to existing PV, PVC, and SharedVolume resources when the driver is uninstalled? Do they have finalizers that block the deletion until they are removed manually?

(Also it is worth noting that "uninstall" could happen one of two ways: CR edit as assumed here, or uninstall of the operator via OLM. For the latter case, https://github.com/operator-framework/enhancements/pull/46 will be of interest.)

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+### `SharedVolume` CRD
+
+Since `SharedVolume` is trivial to handle, just creates PV + PVC, we leave handling of the CRD as a implementation
+detail. We may try to copy code from the community operator, but if it's hard to combine library-go and
```
2uasimojo, line 157: > copy code from the community operator

I'm confused. The rest of this document makes it sound like this feature is accomplished by modifying the community operator rather than starting over somewhere else. However, I got the opposite impression in our previous talks. Which is it?

But to the technical point: I don't know anything about library-go, but aws-efs-operator's handling of SharedVolume requires nothing more complex than k8s primitives (CRUD and Watch). IOW if it's possible to port *anything* from controller-runtime to library-go, it should be possible to port SharedVolume handling.

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+controller-runtime in a single binary, we may as well reimplement the handling to library-go style controller.
+
+### Upgrade from the community operator
+
+Prerequisites:
+* The community aws-efs-operator is installed in OCP 4.7 cluster in `openshift-operators` namespace. The CSI driver runs
+in the same namespace too.
+  * The operator `Subscription` contains the community catalog as the operator source.
+
+Expected workflow (confirmed with OLM team):
```
2uasimojo, line 167: How does this flow play with the OCM UI around OperatorHub?

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+   to the community source. The "old" operator is still running.
+2. User edits the `Subscription` and updates the `source:` to the Red Hat certified catalog.
+   (Alternatively, user may un-install the community operator and install the supported one.)
+3. OLM reconciles this change, checks the new catalog and if there is a newer operator there, it will install it as an
+   update. In other words, we need to ship a newer version in the Red Hat catalog than in the community one.
+4. The "new" CSI operator starts. It does not see `ClusterCSIDriver` CR for EFS and only emits Info level alert that
+   there is no CR and does nothing. This means the old CSI driver is still running (installed by the "old" operator)
+   and any apps that use it are still running.
+5. User reads release notes, documentation or alert and creates th CR.
+6. The "new" operator "adopts" the CSI driver objects. Namely, it updates DaemonSet of the CSI driver, which then does
```
2uasimojo, line 179: Doesn't this entail moving several of these objects (the namespace-scoped ones) to a different namespace? If so, not sure there's really a requirement to preserve the same names.

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+   ```
+
+It would be nice if we did not ship `ose-aws-efs-utils-base` anywhere, so people can consume the EFS utils only through
+the EFS operator. **We do not want to maintain it for everyone.**
+
+### Code organization
+
+* Reuse existing `github.com/openshift/aws-efs-operator` for the operator.
+  * Move the controller-runtime based operator one to a separate branch (`release-4.7`, even though it contains all
+    previous releases?)
```
2uasimojo, line 216: Or `release-pre-4.8`, or `release-community` or...

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+  * Continue with library-go based one in `master`.
+* Fork the CSI driver into `github.com/openshift/aws-efs-csi-driver`.
+* Fork `github.com/aws/efs-utils` into `github.com/openshift/aws-efs-utils`.
+
+### Open Questions
+
+* How to graduate community operator into a supported one?
```
2uasimojo, line 223: Meaning, importantly, how to prevent someone from installing the community operator into 4.8? Yeah, that's a tough one. Perhaps the supported operator checks for that, refuses to function, and throws an alert?

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Test Plan
+
+* Automatic jobs:
+  * Run CSI certification tests with the CSI driver:
+     1. Install OCP on AWS.
+     2. Install the operator.
+     3. Create AWS EFS share (this may require extra privileges in CI and figure out the networking!)
+     4. Set up dynamic provisioning in the driver (by creating a StorageClass). This assumes dynamic provisioning
+        is implemented (released) upstream before we start testing.
```
2uasimojo, line 239: I recommend not tying the test plan to dynamic provisioning -- or in fact any direct AWS interaction -- in the actual driver or operator. The test environment itself can be configured with appropriate access to the AWS side; and the test cases can do the provisioning/deprovisioning as separate steps from the operator intsallation and configuration.

We had to do plenty of this while developing the community operator. We never quite got to automating tests end to end, but I wrote a [utility](https://github.com/2uasimojo/aws-efs-manage) that does the AWS networking, security, and provisioning setup/teardown. You may be able to reuse pieces of it here.

As for permissions, I'm almost certain it is already possible to specify AWS creds in CI today. That's all you should need.

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+
+#### Removing a deprecated feature
+
+TODO: shall we deprecate `SharedVolume`?
```
2uasimojo, line 274: Probably. You convinced me that it carries a security risk that we can't count on users understanding.

...and if we're going to do this, doing it on the boundary of community=>supported would probably make the most sense. Happy to discuss strategy here.

2uasimojo (commented):
This is cool. It is a bit divergent from what we've discussed previously -- but in a direction I like :)

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+   FROM ose-aws-efs-utils-base
+   COPY --from=builder <the driver binary> /usr/bin/
+   ```
+
+It would be nice if we did not ship `ose-aws-efs-utils-base` anywhere, so people can consume the EFS utils only through
+the EFS operator. **We do not want to maintain it for everyone.**
+
+### Code organization
+
+* Reuse existing `github.com/openshift/aws-efs-operator` for the operator.
```
bertinatto, line 214: nit: add `repository` to not confuse with code

bertinatto: IMO we should rename it to `aws-efs-csi-driver-operator` to be consistent with the other repositories.

Consistent naming is the only requirement for creating new CSI driver/operator repositories (we'd get away with that because this repository already exists, but it'd be nice to make it consistent).

enhancements/storage/csi-efs-operator.md
bertinatto: +1 for something else than `release-4.7`, since the operator wasn't part of OCP 4.7

enhancements/storage/csi-efs-operator.md
bertinatto: Wouldn't it be enough if we just remove it from OperatorHub?

enhancements/storage/csi-efs-operator.md
bertinatto: I'd love to have the functionality provided by `SharedVolume` implemented in the CSI driver itself. Maybe we can focus on that rather than supporting this CR?

enhancements/storage/csi-efs-operator.md
bertinatto: Here's the CR the cluster-storage-operator creates for EBS: https://github.com/openshift/cluster-storage-operator/blob/master/assets/csidriveroperators/aws-ebs/10_cr.yaml

enhancements/storage/csi-efs-operator.md
jsafrane: Yes, it will write status there. The CRD is watched by several operators, but each operator watches its own instance with explicitly defined name and does not touch other CRs.
https://github.com/openshift/api/blob/4b79815405ec40f1d72c3a74bae0ae7da543e435/operator/v1/0000_90_cluster_csi_driver_01_config.crd.yaml#L39

enhancements/storage/csi-efs-operator.md
jsafrane: Problem is that it's not possible to destroy a cluster if it has EFS volume. Installer will get stuck at destroying networks, because AWS does not allow to delete those used by EFS volumes. Since it's cluster admin who must create EFS share, it's up to them either to add `owner` tag (for automatic deletion on cluster teardown) or preserve it, but deal with the consequences.

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+* AWS EFS volumes do not have any size enforcement. They can hold as much data as apps put there and someone will have to
+  pay for the storage. OCP offer standard metrics for amount of data on PVCs, it's up to the cluster admin to create
+  alerts if the consumption gets too high.
+* It uses [efs-utils](https://github.com/aws/efs-utils) to provide encryption in transport, which is not included in
+  RHEL. We will need to package it, either as an image or RPM package.
+
+## Proposal
+High level only, see design details below.
+
+1. Rewrite current aws-efs-operator to use [CSI driver installation functions of library-go](https://github.com/openshift/library-go/tree/master/pkg/operator/csi/csicontrollerset).
```
jmrodri, line 102: Since you are already using controller-runtime, have you considered making this an [operator-sdk](https://github.com/operator-framework/operator-sdk/) project? We have [integration with OLM](https://master.sdk.operatorframework.io/docs/olm-integration/) like `operator-sdk run bundle` which will help you test out your bundle with OLM, among other things. 

There's also a supported operator-sdk offering beginning with OpenShift 4.7:
https://mirror.openshift.com/pub/openshift-v4/clients/operator-sdk/latest/

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Open Questions
+
+* How to graduate community operator into a supported one?
+  * The community one does not go through registry.redhat.io.
+    * The "community" team is willing to do anything that's necessary to help with the transition.
+  * The community one has only a single channel, `stable`.
+  * We want the community one working in 4.7 and older and the supported one in 4.8 and newer.
+  * It should be possible to ship new community updates in 4.7 channel/branch to fix potential bugs there and/or
+    make the transition as seamless as possible.
```
jmrodri, line 229: Will the community operator work in 4.8 as well? or are you sunsetting the community operator? Will the community operator also migrate to library-go?

jmrodri:
@jsafrane I'd like to understand more about the shift from controller-runtime to library-go. What were some of the deficiencies you found with controller-runtime? Have you looked into operator-sdk at all? (see previous comment for urls). We can speak offline about controller-runtime as to not derail this EP.

enhancements/storage/csi-efs-operator.md
jsafrane: I tried to cover *why* in Design Details below. This is almost complete CSI driver operator using library-go code:

https://github.com/openshift/aws-ebs-csi-driver-operator/blob/383a7638b37a4b9a9831c7747e8c499eedcf030f/pkg/operator/starter.go

Basically configure a library and run it. It's so comfortable. We would need to re-implement it in operator-sdk and then we would have two code bases that we need to maintain and sync.

enhancements/storage/csi-efs-operator.md
jsafrane: Added more detailed note about installer behavior and opt-in nature of the tag.

enhancements/storage/csi-efs-operator.md
jsafrane: Added a note about this.

enhancements/storage/csi-efs-operator.md
jsafrane: Added link to EBS CR.

enhancements/storage/csi-efs-operator.md
jsafrane: added a note about what happens when the operator is removed via OLM (driver works, but is not managed, PVs are usable) and when the operand is removed via CR deletion (driver is removed, PVs are not usable).

enhancements/storage/csi-efs-operator.md
jsafrane: Elsewhere in the document is written that the operator will be re-written. Therefore, the SharedVolume code may be carried over. SharedVolume is super simple, so re-implemenation may be actually easier than reusing the existing code.

enhancements/storage/csi-efs-operator.md
jsafrane: I don't know what GUI is going to show - some experiments are needed.

enhancements/storage/csi-efs-operator.md
jsafrane: I considered moving the driver to `openshift-cluster-csi-drivers` namespace, where AWS EBS CSI driver runs, especially if the operator was installed during cluster installation (having a single predictable namespace helps a lot). We moved to OLM, so using the OLM-native way of managing namespaces is better here. And helps with the migration.

@bertinatto, what do you think? Are we painting ourselves into a corner again?

Added a note that we won't change the namespace.

enhancements/storage/csi-efs-operator.md
jsafrane: If it was a new repo, then aws-efs-csi-driver-operator would be great, now that it already exists, I think it's better to leave it as it is - who knows where is it referenced.

enhancements/storage/csi-efs-operator.md
jsafrane: using `release-community`.

enhancements/storage/csi-efs-operator.md
jsafrane: Checking OLM, it supports `minKubeVersion` in CSV files - we may hide the new operator in 4.7 and earlier. Not sure if there is an easy way how to hide the old operator in 4.8 and newer. Added as an extra open item.

enhancements/storage/csi-efs-operator.md
jsafrane: Ideally, I'd like OperatorHub to offer the community operator in OCP <= 4.7 and the supported one in OCP >= 4.8. Having them both could be confusing.

As written above, code the current operator will be still available in a branch and new community updates can be released from there, without any code change.

jsafrane:
@jmrodri @2uasimojo, thanks for your comments! I updated / rephrased some parts based on your feedback, available as a separate commit for easier tracking.

> I'd like to understand more about the shift from controller-runtime to library-go. What were some of the deficiencies you found with controller-runtime? Have you looked into operator-sdk at all? (see previous comment for urls). We can speak offline about controller-runtime as to not derail this EP.

I looked at operator-sdk a long time ago, we even had our initial cluster-storage-operator written in it. But then we got bitten by a operator-sdk version bump and it was the same effort to move to library-go or to newer operator-sdk / controller-runtime.
In addition, library-go made writing OCP-style operators much easier, with ManagementState, LogLevel, creating static resources, updating DaemonSet/Deployments using `metadata.Generation` tracking, typed clients and much more! It's not really well usable outside of OCP though.

BTW, I need to check operator-sdk cmdline to generate CSVs and other stuff. That's definitely helpful.

enhancements/storage/csi-efs-operator.md
2uasimojo: > Wouldn't it be enough if we just remove it from OperatorHub?

That would prevent pre-4.8 clusters from installing it too, right?

enhancements/storage/csi-efs-operator.md
2uasimojo: Actually, a fork from aws-efs-operator to aws-efs-csi-driver-operator could make a lot of sense. That way the old community operator doesn't need to undergo what's essentially a full rewrite, or deal with the branching issue; and the new operator can have its preferred name but still have the benefit of preserving whatever code may be useful.

enhancements/storage/csi-efs-operator.md
2uasimojo: I'll just point out that aws-efs-operator *is* written against operator-sdk (but 0.16.0, so no `run bundle`). Jan has been talking about porting to library-go for $reasons. To editorialize shamelessly, I feel like it's because it's what he's used to: osdk/controller-runtime is likewise "so comfortable" for me :)

enhancements/storage/csi-efs-operator.md
2uasimojo: I also feel like this topic could use more discussion.

enhancements/storage/csi-efs-operator.md
2uasimojo: I think there's still a disconnect here.

I'm not on board with the operator trying to talk to AWS. This will lead to confusion and complexity for negligible benefit.

It still requires the consumer (human or CI) to properly configure AWS access credentials. If they don't, the volume removal will fail, and they won't necessarily find out about it *except* in the case of a blocked cluster uninstall -- and even then, the cause may not be obvious.

It increases your test matrix substantially.

And it's asymmetrical. If the operator can handle deletion, it can easily handle provisioning as well. And that would make a lot more sense to the user.

I don't understand the objection to the CI test case/harness cleaning up explicitly.

enhancements/storage/csi-efs-operator.md
2uasimojo: > I'd love to have the functionality provided by `SharedVolume` implemented in the CSI driver itself.

Can you explain this a bit further, please? Today all `SharedVolume` does is accept a FS and AP ID and create a PV/PVC pair for it. It's basically just a shorthand way to create dozens of lines worth of resource definition with two fields. And it's also an RBAC back door -- which was deliberate, but isn't worth the security in my (new and improved) opinion.

enhancements/storage/csi-efs-operator.md
2uasimojo: Having looked at the [library-go port WIP](https://github.com/openshift/aws-efs-operator/pull/31), which has nothing in common with the original, I don't see a reason to try to reuse that repository.

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,341 @@
...
+2. Update (rewrite) current aws-efs-operator to watch `ClusterCSIDriver` CR named `efs.csi.aws.com`.
+   * `ClusterCSIDriver` CRD is already provided by OCP and is used to install all other CSI drivers that OCP ships.
+     * Every CSI driver operator watches only its CR, with [a defined name](https://github.com/openshift/api/blob/4b79815405ec40f1d72c3a74bae0ae7da543e435/operator/v1/0000_90_cluster_csi_driver_01_config.crd.yaml#L39).
+       They don't act on or modify CRs of other CSI drivers.
+   * Users must explicitly create the CR to install the CSI driver to allow users to uninstall the driver if they don't
+     like it.
+
+3. Keep existing functionality:
+   * `SharedVolume` CRD in `aws-efs.managed.openshift.io/v1alpha1` API group keeps working as it is now.
+   * Update from the community operator to the supported operator should be as seamless as possible.
```
qinpingli, line 117: `Update from the community operator to the supported operator` means uninstall the community operator and reinstall the supported operator?
Do we support `update from the supported operator to the community operator` too?

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,341 @@
...
+     At least CI should use this approach to ensure there are no leftovers.
+
+### Implementation Details/Notes/Constraints [optional]
+
+### Risks and Mitigations
+
+Currently, all cloud-based CSI drivers shipped by OCP are installed when a cluster is installed on the particular cloud.
+AWS EFS will not be installed by default. We will include a note about opt-in installation in docs.
+
+After installation of the new operators, either as a fresh install or upgrading the existing one, users must create the
```
qinpingli, line 143: In one cluster, we should only support installing an efs csi driver operator?  Regardless it's the community one or the supported one.

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,341 @@
...
+[AWS EBS CSI driver](https://github.com/openshift/aws-ebs-csi-driver-operator/blob/383a7638b37a4b9a9831c7747e8c499eedcf030f/pkg/operator/starter.go#L67).
+EFS CSI driver operator should not be much more complicated.
+
+Library-go does not support un-installation of CSI driver (and static resources created by `StaticResourcesController`),
+this must be implemented there.
+
+Note: When the *operator* is un-installed, nothing happens to the CRD, CR, the installed CSI driver, existing PVs and
+PVCs. Everything is working, just not managed by the operator. When the *operand* is un-instaled by removing the
+operator CR, the driver will be removed. Existing PVs, PVCs and SharedVolumes will be still present in the API server,
+however, they won't be really usable. Nobody can mount / unmount EFS volumes without the driver.
```
qinpingli, line 163: I have a question here, if the `operator` is un-installed, when we delete the operator CR, the efs csi driver will not be removed? IIUC, the `operator` will watch the CR and process the driver install or un-install logic.

enhancements/storage/csi-efs-operator.md
qinpingli: A question about collecting logs with must-gather, is there any difference for putting efs csi driver and operator in different namespaces. If we keep the operator and driver in the old namespace, do we need to create a special image to collect logs for efs csi driver and operator?

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,341 @@
...
+  * As consequence, user cannot un-install the CSI driver easily.
+* By default, it is installed in `openshift-operators` namespace and the CSI driver runs there too.
+* It does not create EFS volumes in AWS! It's up to the cluster admin to create EFS volume in AWS, figure in which VPC
+  it should be available and set correct security group. See [this KB article](https://access.redhat.com/articles/5025181)
+  for suggested procedure.
+* It offers `SharedVolume` CRD in `aws-efs.managed.openshift.io/v1alpha1` API group to create PV + PVC for an EFS
+  volume. The EFS volume still needs to be created manually, see the previous bullet!
+  * All that `SharedVolume` really does is that it creates PV with proper settings + a PVC. This allows unprivileged
+    users to create PVs. Cluster admins need to be careful to give permissions to use this CRD only to trusted users.
+  * Using `SharedVolume` is purely optional! Cluster admin may create PVs for EFS volumes manually, it's not harder
```
qinpingli, line 77: In the default, the cluster admin and the storage admin have the create access right for the `SharedVolume`? The regular user can get/list the `SharedVoume` CRs?

2uasimojo: Hm, this is an excellent question -- I actually have no idea how this works. I want to say that, while I was testing, I never had to create specific RBAC for SharedVolumes. Which of course makes the security issue far worse.

[Later] Confirmed. I just installed the community operator and my (GitHub IDP) dedicated-admins user was able to create a SharedVolume.

qinpingli: Thanks! We'll check this in our test too.

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,339 @@
...
+[aws-efs-operator](https://github.com/openshift/aws-efs-operator) uses upstream CSI driver, upstream CSI sidecars and
+[the operator itself](https://quay.io/repository/app-sre/aws-efs-operator?tab=tags) is not released through
+registry.redhat.com and therefore cannot be supported by us.
+
+### Goals
+
+* Allow users to use AWS EFS volumes in a supported way.
+* Allow existing users of community-supported aws-efs-operator to update to Red Hat supported solution.
+* Keep existing community community-supported aws-efs-operator features (namely, `SharedVolume` in
+  `aws-efs.managed.openshift.io` API group). See below.
```
2uasimojo, line 54: IMO this should be a non-goal. Problems with SharedVolume in the existing operator:
- It is insecure (TL;DR: anyone with permissions to create SharedVolume can attach to any access point)
- Customers don't like the automatic creation of PVCs, and particularly the strict naming convention.
- No support for dynamic provisioning.

In order to keep SharedVolume, you would need to change it so much that it might as well be a different CR, or none at all.

Another soft concern is customers not understanding whether their SharedVolume is being managed by the old thing (insecure, unsupported) or the new thing (secure, supported). If they think it's the latter but it's really the former, they could end up pretty upset.

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,339 @@
...
+  * Using `SharedVolume` is purely optional! Cluster admin may create PVs for EFS volumes manually, it's not harder
+    than using `SharedVolume`.
+
+### AWS EFS CSI driver
+
+AWS EFS CSI driver is different from the other CSI drivers we ship. Listing the main differences here for completeness.
+
+* It implements dynamic provisioning in non-standard way. Cluster admin must manually create an EFS share in the cloud
+  (and figure out related networking and firewalls). The EFS CSI driver will provision volumes out of it as
+  *subdirectories* of it. The CSI driver cannot create EFS shares on its own.
```
2uasimojo, line 86: This is no longer true. The upstream driver [supports dynamic provisioning](https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/examples/kubernetes/dynamic_provisioning) as of version 1.2.0. The implementation looks a little weird to me, so you might consider rethinking this design with that in mind.

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,339 @@
...
+
+2. Update (rewrite) current aws-efs-operator to watch `ClusterCSIDriver` CR named `efs.csi.aws.com`.
+   * `ClusterCSIDriver` CRD is already provided by OCP and is used to install all other CSI drivers that OCP ships.
+     * Every CSI driver operator watches only its CR, with [a defined name](https://github.com/openshift/api/blob/4b79815405ec40f1d72c3a74bae0ae7da543e435/operator/v1/0000_90_cluster_csi_driver_01_config.crd.yaml#L39).
+       They don't act on or modify CRs of other CSI drivers.
+   * Users must explicitly create the CR to install the CSI driver to allow users to uninstall the driver if they don't
+     like it.
+
+3. Keep existing functionality:
+   * `SharedVolume` CRD in `aws-efs.managed.openshift.io/v1alpha1` API group keeps working as it is now.
```
2uasimojo, line 114: Recommend nixing this, as noted above.

enhancements/storage/csi-efs-operator.md
2uasimojo: Given the complexity of managing "upgrades" (or downgrades) and the low usage of the community operator, is it worth considering *only* providing documentation for such moves instead?

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,339 @@
...
+Prerequisites:
+* The community aws-efs-operator is installed in OCP 4.8 cluster in `openshift-operators` namespace. The CSI driver runs
+in the same namespace too.
+  * The operator `Subscription` contains the community catalog as the operator source.
+
+Expected workflow (confirmed with OLM team):
+
+1. User upgrades to 4.9. Nothing interesting happens at this point, because the operator `Subscription` still points
+   to the community source. The "old" operator is still running.
+2. User edits the `Subscription` and updates the `source:` to the Red Hat certified catalog.
```
2uasimojo, line 180: IMO these should be entirely separate products. Separate subscriptions etc. As noted above, I think it's acceptable for the upgrade process to be manual; but if it's going to be semi-automated, the new thing can watch & wait for the old version of the driver to disappear before setting up the new one.

enhancements/storage/csi-efs-operator.md
bertinatto: Thanks for bringing this up, I am +1 for dropping this requirement. Besides making things easier for us, the current `SharedVolume` feels a bit "out of place" compared to other solutions.

enhancements/storage/csi-efs-operator.md
bertinatto: I meant the CSI driver should support dynamic provisioning, which has been implemented recently (as you pointed out above).

enhancements/storage/csi-efs-operator.md
jsafrane: Removed SharedVolume as a separate commit.

enhancements/storage/csi-efs-operator.md
jsafrane: The implementation really creates subdirs in an existing EFS share. The cluster admin still must create an initial share first.

enhancements/storage/csi-efs-operator.md
jsafrane: Done

enhancements/storage/csi-efs-operator.md
jsafrane: Reworked the upgrade - user must un-install the old operator + driver and install the new one as a separate step. No automatic upgrade.

enhancements/storage/csi-efs-operator.md
2uasimojo: Still seeing this bullet in the latest commit.

2uasimojo (commented):
This is looking solid to me. Thanks for the updates.

enhancements/storage/csi-efs-operator.md
jsafrane: I think it's correct now. We won't remove the community-supported aws-efs-operator from Operator Hub. Users may choose the old one or not to upgrade to the new one, if they depend on `SharedVolume`.

2uasimojo: Okay, the wording was confusing to me, made it still sound like the *new* solution was going to support `SharedVolume`. It's made clear further down, so no big deal.

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,332 @@
...
+* It's written using `controller-runtime`.
+* Once the operator is installed by OLM, the operator automatically installs the CSI driver, without any CRD/CR.
+  * As consequence, user cannot un-install the CSI driver easily.
+* By default, it is installed in `openshift-operators` namespace and the CSI driver runs there too.
+* It does not create EFS volumes in AWS! It's up to the cluster admin to create EFS volume in AWS, figure in which VPC
+  it should be available and set correct security group. See [this KB article](https://access.redhat.com/articles/5025181)
+  for suggested procedure.
+* It offers `SharedVolume` CRD in `aws-efs.managed.openshift.io/v1alpha1` API group to create PV + PVC for an EFS
+  volume. The EFS volume still needs to be created manually, see the previous bullet!
+  * All that `SharedVolume` really does is that it creates PV with proper settings + a PVC. This allows unprivileged
```
jimmyjones2, line 75: I found this very unsual and different to other storage providers I've used before. For example I'm installing a helm chart which creates a RWX PVC - I'd expect this to be dynamically provisioned and work. SharedVolume breaks this workflow.

2uasimojo: This has been discussed several times and each time weighs the scales more heavily in favor of not carrying the `SharedVolume` CR/concept forward into this solution.

(However, I'll point out that, even if `SharedVolume` _is_ implemented, it wouldn't be required as part of the workflow. You could ignore it and create your PVs and PVCs directly in the same way you're used to.)

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,315 @@
...
+   * `ClusterCSIDriver` CRD is already provided by OCP and is used to install all other CSI drivers that OCP ships.
+     * Every CSI driver operator watches only its CR, with [a defined name](https://github.com/openshift/api/blob/4b79815405ec40f1d72c3a74bae0ae7da543e435/operator/v1/0000_90_cluster_csi_driver_01_config.crd.yaml#L39).
+       They don't act on or modify CRs of other CSI drivers.
+   * Users must explicitly create the CR to install the CSI driver to allow users to uninstall the driver if they don't
+     like it.
+
+3. Remove `SharedVolume` CRD.
+   * The CRD may be insecure - it allows anyone who has permission to create CRs to "steal" data from any
+     EFS share by crafting `SharedVolume` pointing to it.
+   * The CR just creates a PV+PVC, cluster admin may grant permissions to create PVs to trusted people, which si more
```
bertinatto, line 120: nit: `s/si/is/g`

enhancements/storage/csi-efs-operator.md
```diff
@@ -0,0 +1,315 @@
...
+  for older OCP releases.
+* Fork the CSI driver into `github.com/openshift/aws-efs-csi-driver`.
+* Fork `github.com/aws/efs-utils` into `github.com/openshift/aws-efs-utils`.
+
+### Open Questions
+
+**How to actually hide the community operator in 4.9 and newer operator catalogs?** We don't want to have two operators
+in the catalog there. We still want to offer the community operator in 4.8 and earlier though.
+
+* Right now it seems there will be two operators available in OLM and it's not that bad.
```
bertinatto, line 253: Is it possible to tag the community operator as deprecated?

bertinatto (commented):
/lgtm
At this point we've implemented all the functionality and CI jobs are on the works (see this [Jira card](https://issues.redhat.com/browse/STOR-553)), so it makes sense to merge this now.
We can make adjustments as follow-ups.

enhancements/storage/csi-efs-operator.md
jsafrane: Fixed

bertinatto:
@jsafrane should we update the status of this enhancement and get it merged?

jsafrane:
Thanks for poking. I marked it as implemented. @bertinatto, PTAL.

jsafrane:
/assign @knobunc
for approval

PR #688, opened 2021-03-10
enhancements/monitoring: Describe single replica topology mode

simonpasquier:
Taking over the work that @lilic started in #671. 

cc @dhellmann @romfreiman @wking


enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,169 @@
...
+operators honor the `SingleReplica` value (see
+[openshift/origin#25812](https://github.com/openshift/origin/pull/25812) and
+[openshift/origin/pull/25885](https://github.com/openshift/origin/pull/25885)).
+
+As soon as the implementation in CMO is merged, we will adjust the single-node
+topology
+[tests](https://github.com/openshift/origin/blob/master/test/extended/single_node/topology.go)
+to remove the exception in place for the monitoring components.
+
+Because the topology mode isn't exposed into the CMO configuration, we don't plan
```
dhellmann, line 138: I'm not sure I understand this statement. Does it refer to the configuration of the CMO itself, rather than the operands?

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Implementation History
+
+## Drawbacks
+
+N/A
+
+## Alternatives
+
+* Do nothing. It means that the monitoring stack would use more resources than is actually needed.
```
dhellmann, line 164: I think, given the anti-affinity rules, it also means some of the operands would always show a Pod failing to schedule, right?

dhellmann (commented):
This draft looks good. I have a couple of clarifying points inline, but nothing worth holding up merging it.

/approve

enhancements/monitoring/integration-with-topology-modes.md
simonpasquier: yes good point

enhancements/monitoring/integration-with-topology-modes.md
simonpasquier: I will clarify but it refers to the configuration that CMO exposes to the cluster admins via the `openshift-monitoring/cluster-monitoring-operator` configmap (for "historical" reasons, CMO doesn't expose CRDs for this). It allows to configure both the operator (for example to enable user-workload monitoring) and the operands (for example to set resource requests/limits on Prometheus).

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,176 @@
...
+## Implementation History
+
+## Drawbacks
+
+N/A
+
+## Alternatives
+
+* Do nothing. CMO would deploy multiple replicas for Prometheus Alertmanager,
+* Thanos querier, Thanos ruler and prometheus-adapter. As a result, the
```
dhellmann, line 170: The formatting here looks suspicious. I think lines 170-171 shouldn't start with `*`.

dhellmann (commented):
I have one formatting nit, but the updated content is clearer and I think this is ready.

dhellmann:
The latest draft looks good.

Who needs to approve this before we merge it?

/approve

simonpasquier:
I'd say @s-urbaniak as the monitoring team lead (though he's missing from the OWNERS file) and @bparees.

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,176 @@
...
+
+As described in the [Cluster High-availability Mode API](https://github.com/openshift/enhancements/blob/master/enhancements/single-node/cluster-high-availability-mode-api.md)
+enhancement and implemened in [OpenShift API](https://github.com/openshift/api/pull/827),
+OpenShift 4.8 introduces 2 new status fields in the Infrastructure API that
+informs operators about the cluster deployment topology.
+
+Depending on whether an operand is deployed on the control nodes or not, the
+operator should look at either `controlPlaneTopology` and
+`infrastructureTopology`. If the field value is `HighlyAvailable`, the operator
+should configure the operand for high-avaible operation. If the field value is
```
bparees, line 49:
```suggestion
should configure the operand for high-availability operation. If the field value is
```

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,176 @@
...
+`SingleReplica`, it should *not* configure the operand for high-avaible
```
bparees, line 50:
```suggestion
`SingleReplica`, it should *not* configure the operand for high-availability
```

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,176 @@
...
+### Non-Goals
+
+- Implement additional enhancements to reduce CPU and memory footprints.
+- Allow cluster admins to control the number of replicas per operand.
+- Address CodeReady Containers requirements (though it might help).
+- Add tests for `SingleReplica` to the existing end-to-end test suite in the CMO repository.
+
+## Proposal
+
+Whenever CMO needs to reconcile, it retrieves the
```
bparees, line 68: does this mean it won't watch the topology resouce directly for changes?  It seems like it should, so we aren't dependent on a reconcile cycle triggered by something else to pick up a change.

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,176 @@
...
+* openshift-user-workload-monitoring/prometheus-user-workload (StatefulSet): 2 replicas
+* openshift-user-workload-monitoring/thanos-ruler (StatefulSet): 2 replicas
+
+By default, CMO deploys no operand on control nodes thus it should not look at
+the `controlPlaneTopology` field.
+
+### Risks and Mitigations
+
+In HA mode, the Prometheus instances are spread on different nodes using
+soft anti-affinity and they work completely indepently one from another (there's no
```
bparees, line 97:
```suggestion
soft anti-affinity and they work completely independently one from another (there's no
```

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,176 @@
...
+redundancy in term of metrics scraping, rule evaluations and querying as long
+as one of the Prometheus instances is still up.
+
+The Thanos ruler instances of user workload monitoring work the same as
+Prometheus except they only deal with rule evaluations.
+
+The 3 Alertmanager instances receive alerts from both Prometheus instances and
+as long as one instance is up and running, the notifications should flow to the
+receivers. Alertmanager instances also replicate silences and notification logs
+between themselves to prevent duplicate notifications.
```
bparees, line 110: unrelated to this EP but why do we run 3 of them then?  2 seems sufficient for HA.

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,176 @@
...
+don't plan to add any test specific to the single-node mode to the CMO test
+suite.
+
+### Graduation Criteria
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+Upgrades are a non goal right now.
```
bparees, line 154: what exactly does this mean?  that upgrades in a single node toplogy are not supported?  or something else?

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Version Skew Strategy
+
+If CMO reads a value that is neither equal to `HighAvailable` nor
+`SingleReplica`, it should consider that the topology is `HighAvailable`.
```
bparees, line 159: what about if it cannot find/read the resource at all?

bparees (changes requested):
couple nits but generally looks good to me.

enhancements/monitoring/integration-with-topology-modes.md
simonpasquier: Yes in-place upgrades in single node topology aren't supported for now (from https://github.com/openshift/enhancements/blob/master/enhancements/single-node/production-deployment-approach.md#upgrade--downgrade-strategy).

enhancements/monitoring/integration-with-topology-modes.md
simonpasquier: good point!

enhancements/monitoring/integration-with-topology-modes.md
simonpasquier: Alertmanager is a lightweight component compared to Prometheus and with 3 replicas, we have a more resilient system even without persistent storage. More context [here](https://github.com/openshift/cluster-monitoring-operator/issues/309) but we could definitely reconsider this setting.

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,182 @@
...
+
+As stated in the [Single-node Production Deployment Approach](https://github.com/openshift/enhancements/blob/master/enhancements/single-node/production-deployment-approach.md)
+proposal, in-place upgrades are a non-goal right now.
+
+### Version Skew Strategy
+
+If CMO reads a value that is neither equal to `HighAvailable` nor
+`SingleReplica`, it should consider that the topology is `HighAvailable`.
+
+If CMO gets an error while reading the resource, it errors out and retries.
```
bparees, line 165: what decision does it make about the number of replicas to configure in this case, though?

simonpasquier: I'd say:
* if it gets a "not found" error, it assumes `HighAvailable`.
* if it gets any other error, it should do nothing and retry reading the resource. 
What do you think?

bparees: seems reasonable to me i think.

dhellmann: Do we have functional configurations in which the Infrastructure config resource isn't present at all? That didn't come up in the enhancement describing the API https://github.com/openshift/enhancements/blob/master/enhancements/single-node/cluster-high-availability-mode-api.md

enhancements/monitoring/integration-with-topology-modes.md
simonpasquier: I've updated the text to make it clearer.

enhancements/monitoring/integration-with-topology-modes.md
simonpasquier: Agree that the `infrastructure/cluster` resource should always be there but I don't think it hurts to cover this case. @dhellmann do you think that CMO should bail out when the resource isn't found?

dhellmann: Assuming a default value for the API seems reasonable, but we should be consistent with other operators. What do other operators do when they can't find that particular resource?

simonpasquier: I've searched within the openshift org and the only thing that pops up is the cluster ingress operator [1] which bails out on any failure to retrieve the cluster resource.

[1] https://github.com/openshift/cluster-ingress-operator/blob/a5525f387106974730ccb9f422e05115498cbc47/pkg/operator/operator.go#L179-L182

dhellmann: OK, that would put the pod into a crash-loop, until the resource is created. That seems like a reasonable approach, too.

CMO only runs on OpenShift, right? You don't need to worry about valid environments without that CR on other distributions?

bparees: surely we can do better than a crashloop....

dhellmann: Well, not "crash loop" I guess. But the reconcile loop would error and retry, right? That seems like the correct behavior.

bparees: yes, retry (preferably on a delay, not in a hotloop) on error is reasonable.

simonpasquier: Makes sense, I've updated the text accordingly.

dhellmann:
The latest draft looks good to me.

/approve

simonpasquier:
@bparees friendly ping to reevaluate the updated proposal :)

enhancements/monitoring/integration-with-topology-modes.md
```diff
@@ -0,0 +1,184 @@
...
+proposal, in-place upgrades are a non-goal right now.
+
+### Version Skew Strategy
+
+If CMO reads a value from the `infrastructures.config.openshift.io/cluster`
+resource that is neither equal to `HighAvailable` nor `SingleReplica`, it
+considers that the topology is `HighAvailable`.
+
+If CMO gets any error while reading the resource, it errors out and retries
+with a delay (as for any failure to synchronize the operands).
```
bparees, line 167: just a little more clarification here:

If it has never been able to read the resource/see a topology config (due to error or missing value) it defaults to highavailable.  If it previously determined the topology, but *now* can't read the topology(i.e. during a resync/relist), it remains on whatever topology it last was able to determine.

PR #689, opened 2021-03-10
Add Enhancement for Installing to Azure Stack Hub

patrickdillon:
Create an enhancement to outline work for installing and supporting OpenShift on Azure Stack Hub.

cc @openshift/openshift-team-installer 
cc @staebler 

/label priority/important-soon

### DIFF
--- a/enhancements/installer/azurestack.md
+++ b/enhancements/installer/azurestack.md
@@ -1,0 +1,221 @@
+---
+title: azurestack
+authors:
+  - "@patrickdillon"
+reviewers:
+  - "@staebler"
+  - "@jhixson74"
+approvers:
+  - "@staebler"
+creation-date: 2021-03-10
+last-updated: 2021-03-10
+status: implementable
+see-also:
+  - "/enhancements/installer/azure-support-known-cloud-environments.md"
+---
+
+## Release Signoff Checklist
+
+- [X] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement covers adding support to install OpenShift clusters to Azure Stack Hub (ASH).
+Azure Stack Hub looks and feels like standard Azure, but differs significantly in terms of
+implemenetation and technical details. Azure Stack Hub shares characteristics from [Azure cloud
+environments](azure-support-known-cloud-environments.md) and on-prem platforms. This enhancement
+covers details for the Installer to create infrastructure and run an OpenShift cluster on ASH.
+
+## Motivation
+
+Currently, Azure Stack Hub is not supported but is a significant on-prem platform.
+
+### Goals
+
+1. Installer creates infrastructure and installs OpenShift on Azure Stack Hub.
+2. All standard support that entails, such as destroy.
+
+### Non-Goals
+
+1. ASH will not necessarily support all Azure public cloud features.
+## Proposal
+
+
+### User Stories
+
+#### Story 1
+As a cluster admin, I want to be able to use `openshift-install` to create and destroy a cluster on
+Azure Stack Hub.
+
+#### Story 2
+As an OpenShift developer, I want Azure Stack Hub to be managed and released like other platforms and covered by CI tests.
+
+### Implementation Details/Notes/Constraints [optional]
+
+Primary ways in which Azure Stack differs from Azure:
+- users must specify API endpoint(s)
+- utilizes different API versions than public Azure
+- requires different rhcos image
+- limited instance metadata service (IMDS) for VMs
+- does not support private DNS zones
+- limited subset of Azure infrastructure available (ergo different Terraform provider)
+
+### Risks and Mitigations
+
+The risk of adding ASH support is somewhere between adding a new Azure environment and an entirely new platform.
+
+## Design Details
+
+### Infrastructure Global Configuration
+
+#### API
+ASH will need to be added to the known [Azure Cloud Environments in the API](https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L339-L355):
+
+```go
+// AzureStackCloud is the Azure cloud environment used on-premises.
+AzureStackCloud AzureCloudEnvironment = "AzureStackCloud"
+```
+
+#### Cluster Config Operator
+Cluster config operator will [need to be updated](https://github.com/openshift/cluster-config-operator/blob/master/pkg/operator/kube_cloud_config/azure.go#L23) and the new API vendored in.
+
+#### API endpoints
+API endpoints vary between Azure Stack deployments and, therefore, must be provided to the installer by the user or
+constructed by the installer based on user input.
+
+The primary API endpoint is the Management Endpoint, which follows a naming pattern:
+
+```shell
+https://management.{region}.{Azure Stack Hub domain}
+```
+Once this management endpoint is determined, it can be used to determine other endpoints of the stack instance.
+For example, with the `PPE3` region and `stackpoc.com` domain we can [query the metadata endpoint for endpoints](https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-rest-api-use?view=azs-2008):
+
+```shell
+curl 'https://management.ppe3.stackpoc.com/metadata/endpoints?api-version=2015-01-01' | jq
+{
+  "galleryEndpoint": "https://providers.ppe3.local:30016/",
+  "graphEndpoint": "https://graph.windows.net/",
+  "portalEndpoint": "https://portal.ppe3.stackpoc.com/",
+  "authentication": {
+    "loginEndpoint": "https://login.microsoftonline.com/",
+    "audiences": [
+      "https://management.stackpoc.com/81c9b804-ec9e-4b5a-8845-1d197268b1f5"
+    ]
+  }
+}
+
+```
+
+##### Kubelet, Cloud-provider endpoints and Azure/go-autorest
+
+ When the cloud environment is set to
+`AZURESTACKCLOUD`, the Kubelet, utilizing [Azure/go-autorest](https://github.com/Azure/go-autorest/blob/562d376/autorest/azure/environments.go#L207-L225) expects the environment variable `AZURE_ENVIRONMENT_FILEPATH` to point to a [json configuration file](https://kubernetes-sigs.github.io/cloud-provider-azure/install/configs/#azure-stack-configuration), which is [typically located at `/etc/kuberentes/azurestackcloud.json`](https://github.com/kubernetes-sigs/cloud-provider-azure/issues/151).
+
+The installer can create this JSON file based on the metadata described above, and lay down the configuration file
+and Kubelet environment variable with machineconfigs.
+
+##### Operators
+Many operators will also need to set the resource management (and perhaps other) endpoints. In order to provide the
+endpoints to operators, the installer should add the endpoints file to the
+[cloud provider configmap](https://github.com/openshift/installer/blob/master/pkg/asset/manifests/cloudproviderconfig.go#L126-L141):
+
+```go
+  cloudProviderEndpointsKey = "azurestackcloud.json"
+
+  azureStackEndpoints, err := azure.stackEndpoints{
+			resourceManagerEndpoint:          resourceManagerEndpoint,
+			serviceManagementEndpoint:        serviceManagementEnpoint,
+			...
+		}.JSON()
+		if err != nil {
+			return errors.Wrap(err, "could not create cloud provider config endpoints")
+		}
+		cm.Data[cloudProviderEndpointsKey] = azureStackEndpoints
+
+```
+
+Operators can then [create clients](https://github.com/Azure/azure-sdk-for-go/blob/master/services/attestation/mgmt/2020-10-01/attestation/client.go#L45-L53) using the resource manager endpoint and subscription ID from the cloud provider
+configs.
+#### Installer Manifests & Cloud Provider Config
+ASH should be able to reuse Azure machine manifests with proper validation.
+
+The [Azure cloud-provider-config](https://github.com/openshift/installer/blob/master/pkg/asset/manifests/azure/cloudproviderconfig.go) will need to be slightly adapted. `LoadBalancerSku` for ASH will be `basic`. `UseInstanceMetadata` should probably be `false`. Service principal credentials will be included here [(see next section)](#identity-management).
+
+The cluster [DNS manifest contains the private DNS zone](https://github.com/openshift/installer/blob/master/pkg/asset/manifests/dns.go#L110-L112). At a minimum this should be removed, but there is an [open question](#open-question) about how this should be handled.
+
+### Cloud Provider Authentication (Identity Management)
+Azure Stack does not support user-assigned identities. In public Azure, the Installer uses a user-assigned identity to
+authenticate to the cloud provider. To overcome this lack of user-assigned identities, we can passthrough the service
+principal used for Installer authentication to the cloud-provider config with `aadClientID` & `aadClientSecret`. We
+must ensure that we are able to rotate the service principal secret in the cloud-provider config and that doing so
+will cause the updated service principal to be used (presumably by writing a new machine config & causing a reboot).
+
+### User Authentication
+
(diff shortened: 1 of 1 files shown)


enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,160 @@
...
+#### Story 1
+As a cluster admin, I want to be able to use `openshift-install` to create and destroy a cluster on
+Azure Stack Hub.
+
+#### Story 2
+As an OpenShift developer, I want Azure Stack Hub to be managed and released like other platforms and covered by CI tests.
+
+### Implementation Details/Notes/Constraints [optional]
+
+Primary ways in which Azure Stack differs from Azure:
```
staebler, line 63: I would also list the need to specify the API endpoints as a way in which ASH differs. That difference is significant.

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,160 @@
...
+
+#### API endpoints
+ASH endpoints are user-provided and, therefore, the Azure SDK treats ASH endpoints differently than the already-known endpoints of other Azure environments. When the cloud environment is set to
+`AZURESTACKCLOUD` the SDK expects the environment variable `AZURE_ENVIRONMENT_FILEPATH` to point to a [json configuration file](https://kubernetes-sigs.github.io/cloud-provider-azure/install/configs/#azure-stack-configuration), which is [typically located at `/etc/kuberentes/azurestackcloud.json`](https://github.com/kubernetes-sigs/cloud-provider-azure/issues/151).
+
+The Installer will need to construct this json configuration file from user input and include the file as part of the cloud-provider-config configmap. The file will also need to be present on nodes for the kubelet. `AZURE_ENVIRONMENT_FILEPATH` will need to be set programmatically for ASH on the kubelet, presumably through the kubelet's systemd unit [(see open questions)](#open-questions).
+
+Operators will set the `AZURE_ENVIRONMENT_FILEPATH` and mount the endpoints JSON file from the cloud-provider-config configmap. All operators using the Azure SDK will need to do this.
```
staebler, line 94: Is there a way for operators to configure the Azure client directly with the custom endpoints rather than relying on a json file? It would be cumbersome for an operator with a static manifest to do this with mounts, given that the mounted ConfigMap must be copied over to the namespace. The mounting option is feasible for controllers that the operator starts but not easily for operators directly.

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,160 @@
...
+Cluster config operator will [need to be updated](https://github.com/openshift/cluster-config-operator/blob/master/pkg/operator/kube_cloud_config/azure.go#L23) and the new API vendored in.
+
+#### API endpoints
+ASH endpoints are user-provided and, therefore, the Azure SDK treats ASH endpoints differently than the already-known endpoints of other Azure environments. When the cloud environment is set to
+`AZURESTACKCLOUD` the SDK expects the environment variable `AZURE_ENVIRONMENT_FILEPATH` to point to a [json configuration file](https://kubernetes-sigs.github.io/cloud-provider-azure/install/configs/#azure-stack-configuration), which is [typically located at `/etc/kuberentes/azurestackcloud.json`](https://github.com/kubernetes-sigs/cloud-provider-azure/issues/151).
+
+The Installer will need to construct this json configuration file from user input and include the file as part of the cloud-provider-config configmap. The file will also need to be present on nodes for the kubelet. `AZURE_ENVIRONMENT_FILEPATH` will need to be set programmatically for ASH on the kubelet, presumably through the kubelet's systemd unit [(see open questions)](#open-questions).
```
staebler, line 92: Where will this user input come from? Is the expectation that they will have already set up an `AZURE_ENVIRONMENT_FILEPATH` on the install machine, given that the installer will presumably need to know the endpoints for its accesses to the Azure API?

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,160 @@
...
+
+Primary ways in which Azure Stack differs from Azure:
+- requires a different rhcos image
+- limited instance metadata service (IMDS) for VMs
+- does not support private DNS zones
+- limited subset of Azure infrastructure available (ergo different Terraform provider)
+
+### Risks and Mitigations
+
+Adding ASH support is somewhere in between adding a new Azure environment and an entirely new platform.
```
staebler, line 71: Can you give a brief pros and cons of having ASH as a separate platform versus having ASH as a separate cloud environment under the Azure platform? Given that we cannot re-use the terraform and a number of fields in the install config are either not supported in ASH or not relevant to public Azure, I do not see much benefit from the installer side for keeping ASH under the Azure platform. I assume that having a separate platform would, however, have a lot more impact to in-cluster operators that do not see much divergence between ASH and public Azure.

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,160 @@
...
+This provider lacks the ability to create a service principal for the resource group and assign a contributor role, which is required by the Ingress controller.
+
+These actions are achievable through the CLI, if necessary these commands could be run as a Terraform post hook.
+
+The Terraform Azure Active Directory provider does not state support for Azure Stack. It could be tested for AzureStack support, in which case a service principal could be created by that provider and the role assignment would be created with an ARM template executed by the ASH provider.
+
+### Open Questions
+
+1. We need to explore how to pass the JSON endpoints file to the Kubelet:
+    1. Can the endpoints file be simply passed from the installer through ignition? Does this
```
staebler, line 127: Could this be MachineConfigs that the installer adds for master and workers?

Does the bootstrap machine need this as well? I am assuming not since you have been able to run kubelet successfully on the bootstrap machine with the Azure platform.

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,160 @@
...
+  require further MCO work?
+    1. Can we pass the env var by programatically populating one of the [already existing Kubelet
+  unit's `EnvironmentFile`s](https://github.com/openshift/machine-config-operator/blob/master/templates/master/01-master-kubelet/_base/units/kubelet.service.yaml#L19). If so, which & how
```
staebler, line 130: This could be done as a drop-in in the same MachineConfigs that are used for adding the json file.

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,160 @@
...
+  (ignition or MCO)?
+1. Should ASH produce a cluster DNS manifest or follow the baremetal approach and not create one?
+1. Suggestions for determining all operators that need to adapt for the ASH config.
```
staebler, line 133: https://sourcegraph.com/search?q=repo:github.com/openshift/+rev:master+file:vendor/github.com/Azure/azure-sdk-for-go/version/version.go+archived:no+count:1000+fork:yes&patternType=literal

enhancements/installer/azurestack.md
staebler: Follow up on this. You can definitely set the endpoint programmatically when creating the client.
For example, see the following.
https://github.com/staebler/cluster-ingress-operator/blob/3bb8ed560892ae648806e6133d18803da9ec341b/pkg/dns/azure/client/client.go#L88

enhancements/installer/azurestack.md
patrickdillon, line 130: `EnvironmentFile` does not work for this purpose but `Environment` does, e.g.:

```
[Service]
Type=notify
ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
Environment="AZURE_ENVIRONMENT_FILEPATH=/etc/kubernetes/azurestackcloud.json"
```

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,160 @@
...
+
+The Installer will need to construct this json configuration file from user input and include the file as part of the cloud-provider-config configmap. The file will also need to be present on nodes for the kubelet. `AZURE_ENVIRONMENT_FILEPATH` will need to be set programmatically for ASH on the kubelet, presumably through the kubelet's systemd unit [(see open questions)](#open-questions).
+
+Operators will set the `AZURE_ENVIRONMENT_FILEPATH` and mount the endpoints JSON file from the cloud-provider-config configmap. All operators using the Azure SDK will need to do this.
+
+#### Installer Manifests & Cloud Provider Config
+ASH should be able to reuse Azure machine manifests with proper validation.
+
+The [Azure cloud-provider-config](https://github.com/openshift/installer/blob/master/pkg/asset/manifests/azure/cloudproviderconfig.go) will need to be slightly adapted. `LoadBalancerSku` for ASH will be `basic`. `UseInstanceMetadata` should probably be `false`.
```
patrickdillon, line 99: `useManagedIdentityExtension` must be `false`. Need to explore further whether we an alternative method for auth is needed.

enhancements/installer/azurestack.md
patrickdillon: @staebler I have updated this and I think these revisions make more sense.

enhancements/installer/azurestack.md
patrickdillon: I think the update should clarify this.

patrickdillon:
@gnufied brought up CSI support for Azure Stack https://github.com/openshift/kubernetes/pull/643#issuecomment-811594245 

We do not have a good background on storage, so I'm glad you reached out. Some quick googling led me to this: https://github.com/kubernetes-sigs/azuredisk-csi-driver 

Can we discuss what more needs to be done to add storage support? We can move it over to JIRA too which might be more appropriate.

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,213 @@
...
+Currently, Azure Stack Hub is not supported but is a significant on-prem platform.
+
+### Goals
+
+1. Installer creates infrastructure and installs OpenShift on Azure Stack Hub.
+2. All standard support that entails, such as destroy.
+
+### Non-Goals
+
+1. ASH will not "inherit" established Azure
```
staebler, line 46: It is easy enough to support these features. Since the installer is using the same platform for Azure Stack, the code already exists for support. The installer will use different terraform code, but that code is based largely on the existing terraform code for public Azure. It is easier to retain support than to try to remove support from the terraform code and add validation in the installer to prevent the user from using the feature. I agree that it should not be a **goal** to retain the full suite of feature, but we should not be explicitly disallowing features.

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,213 @@
...
+As an OpenShift developer, I want Azure Stack Hub to be managed and released like other platforms and covered by CI tests.
+
+### Implementation Details/Notes/Constraints [optional]
+
+Primary ways in which Azure Stack differs from Azure:
+- users must specify API endpoint(s)
+- requires different rhcos image
+- limited instance metadata service (IMDS) for VMs
+- does not support private DNS zones
+- limited subset of Azure infrastructure available (ergo different Terraform provider)
```
staebler, line 68: One other aspect that may fall under this category of limiited subset of infrastructure is limited API versions available. We need to be careful that in the future we are cognizant of API version selection and are specific about which API versions an Azure Stack deployment must support in order to run OpenShift.

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,213 @@
...
+The installer can create this JSON file based on the metadata described above, and lay down the configuration file
+and Kubelet environment variable with machineconfigs.
+
+##### Operators
+Many operators will also need to set the resource management (and perhaps other) endpoints. In order to provide the
+endpoints to operators, the installer should add the endpoints file to the
+[cloud provider configmap](https://github.com/openshift/installer/blob/master/pkg/asset/manifests/cloudproviderconfig.go#L126-L141):
+
+```go
+  cloudProviderEndpointsKey = "endpoints"
```
staebler, line 131:
```suggestion
  cloudProviderEndpointsKey = "endpoints.json"
```

enhancements/installer/azurestack.md
```diff
@@ -0,0 +1,213 @@
...
+be edited in the manifests. If we believe having users supply only the domain is more ergonomic, then this does seem like
+a viable option that would capture most use cases.
+
+Another option to consider, would be adding the ability to [supply all endpoints, similar to AWS](https://github.com/openshift/installer/blob/master/pkg/types/aws/platform.go#L39).
+### Terraform
+Azure Stack has [a stand-alone Terraform provider](https://registry.terraform.io/providers/hashicorp/azurestack/latest/docs).
+
+This provider lacks the ability to create a service principal for the resource group and assign a contributor role, which is required by the Ingress controller.
+
+These actions are achievable through the CLI, if necessary these commands could be run as a Terraform post hook.
```
staebler, line 175: Is this conflating the service principal used by the ingress operator with the user-assigned identity attached to the VMs, which is used by the kube-controller-manager? For public Azure, the terraform creates a user-assigned identity but does not create any service principals. Azure Stack does not support user-assigned identities, even through the CLI.

patrickdillon: My statement in the enhancement is confused. The issue is, as you say, the lack of user-assigned identities--not service principals, but it looks like the "main" identity is scoped to the resource group. Here is the code from [IPI](https://github.com/openshift/installer/blame/619d0b28f377ab0c397b1a8b169dac5429599a38/data/data/azure/main.tf#L96-L107). The  [UPI](https://github.com/openshift/installer/blob/master/docs/user/azure/install_upi.md#grant-access-to-the-identity) doc states:

> Grant the Contributor role to the Azure identity so that the Ingress Operator can create a public IP and its load balancer.

I'm not sure I have a firm grasp on the problem. From reading the docs it _seems_ we need to create this identity, to give operators access to create resources in the resource group. The machine API operator would probably need this as well, but perhaps it is not mentioned as these are UPI docs.

Really what I am missing is why/how isn't this handled by the CCO? If we're using passthrough I assume this isn't a problem at all because those creds clearly have perms to create resources in the resource group.

Let me know what you think and I'll update the doc accordingly.

staebler: This is handled by the CCO. All of the in-cluster operators use credentials that are granted by the CCO (assuming mint mode). The UPI docs are a bit misleading when they say that the Contributor role is needed by the Ingress operator. The Ingress operator is creating loadbalancer Services, but it is the kube-controller-manager that is creating the cloud resources for those Services.

In public Azure, the kube-controller-manager uses the user-assigned identity assigned to the VM. Azure Stack does not support user-assigned identities. Presumably, we could allow Azure Stack to create system-assigned identities for each VM. The installer would then need to assign the Contributor role to each system-assigned identity **after** each VM was created. The kubelet also uses the managed identity assigned to the VM, so the managed identity is need for worker machines too. However, the installer does not create the worker VMs so cannot assign the Contributor role to the system-assigned identities for the worker VMs. As far as I am aware, the machine-api does not have a feature for assigning roles to the system-assigned identities of the VMs that the machine-api creates.

A possible solution is to add that feature to the machine-api. Another solution is to supply a Service Principal in the cloud config that the kubelet and kube-controller-manager would use when accessing the Azure API. The latter solution is what I think you are using currently in your UPI work. There may be problems with utilizing that solution long-term due to how the user would replace that Service Principal. I don't think that simply changing the Service Principal in the cloud-provider-config ConfigMap would cause the replacement Service Principal to roll out to all of the kubelets and kube-controller-managers.

patrickdillon:
I get it now. The piece I was missing is that the main identity is assigned
to the VMs in TF and the machinesets. Then you explained the rest; thank
you for that!

If we want to manage the roles after creation, then VM extensions may
provide this:
https://registry.terraform.io/providers/hashicorp/azurestack/latest/docs/resources/virtual_machine_extension


But it seems the MAO may be more involved. Yes the UPI solution writes the
cloud provider config with the service principal ID & secret. So I guess
the question is: does updating the cloud provider config configmap in
cluster cause a new machine config to be written? Does a new machine config
cause a reboot, in which case the new creds would be picked up by the
kubelet.

enhancements/installer/azurestack.md
patrickdillon: I get it now. The piece I was missing is that the main identity is assigned
to the VMs in TF and the machinesets. Then you explained the rest; thank
you for that!

If we want to manage the roles after creation, then VM extensions may
provide this:
https://registry.terraform.io/providers/hashicorp/azurestack/latest/docs/resources/virtual_machine_extension


But it seems the MAO may be more involved. Yes the UPI solution writes the
cloud provider config with the service principal ID & secret. So I guess
the question is: does updating the cloud provider config configmap in
cluster cause a new machine config to be written? Does a new machine config
cause a reboot, in which case the new creds would be picked up by the
kubelet.

patrickdillon:
@staebler This has been revised. PTAL

staebler (approved):
/lgtm

PR #690, opened 2021-03-10
Add repositoryMirrors spec to ImageContentSourcePolicy

QiWang19:
Request for enhancement: https://issues.redhat.com/browse/RFE-1608
Epic: https://issues.redhat.com/browse/OCPNODE-521
Ref: https://github.com/openshift/api/issues/636

Registry mirror set by ImageContentSourcePolicy only will be used if the image
is referenced by digest because the mirror-by-digest of /etc/containers/registries.conf is set to true.
This causes issue since there are use cases where no digests are available.

In ImageContentSourcePolicy, add repositoryMirrors spec, so user can
easily configure it in the same spot they set post-installation mirror configuration.

### DIFF
--- a/enhancements/api-review/add-repositoryMirrors-spec.md
+++ b/enhancements/api-review/add-repositoryMirrors-spec.md
@@ -1,0 +1,189 @@
+---
+title: add-repositoryMirrors-spec
+authors:
+  - "@QiWang19"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2021-03-10
+last-updated: 2021-03-10
+status: implementable
+---
+
+# Add repositoryMirrors spec to cluster wide ImageContentSourcePolicy
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Today, the ImageContentSourcePolicy object sets up mirror with `mirror-by-digest-only` property set to true. Requires images be pulled by digest only. This enhancement plans to add new spec to ImageContentSourcePolicy with configurable field `allowMirrorByTags`, so that `mirror-by-digest-only` can be configured by ImageContentSourcePolicy.
+
+## Motivation
+
+Today, the ImageContentSourcePolicy object sets up mirror configuration with `mirror-by-digest-only` property set to true, which leads to using mirrors only when images are referenced by digests. However, when working with disconnected environments, sometimes multiple ImageContentSourcePolicies are needed, some of them are used by apps/manifests that don't use digests when pulling the images.
+
+Adding the `repositoryMirrors` spec to ImageContentSourcePolicy to configure mirrors will make it possible to pull images from mirror using tags without requiring image digest reference.
+
+### Goals
+
+- Enable the user to pull images through ImageContentSourcePolicy configured mirrors by tags.
+
+### Non-Goals
+
+- This proposal does not recommend using by-tag references for OpenShift release images. Those should still be referenced by digest, regardless of whether `allowMirrorByTags` is enabled for repositories where the release images are mirrored
+
+## Proposal
+
+The new spec `repositoryMirrors` will be added to ImageContentSourcePolicy. The `allowMirrorByTags` property will allow for mirroring by-tag images through configured mirrors.
+
+```go
+// RepositoryMirrors holds cluster-wide information about how to handle mirrors in the registries config.
+// Note: this is different from the RepositoryDigestMirrors that the mirrors only 
+// work when pulling the images that are referenced by their digests.
+type RepositoryMirrors struct {
+    // source is the repository that users refer to, e.g. in image pull specifications.
+    // +required
+    Source string `json:"source"`
+    // If true, mirrors will only be used for digest pulls. Pulling images by
+    // tag can potentially yield different images, depending on which endpoint
+    // we pull from.  Forcing digest-pulls for mirrors avoids that issue.
+    // +optional
+    AllowMirrorByTags bool `json:"allowMirrorByTags"`
+    // mirrors is one or more repositories that may also contain the same images.
+    // The order of mirrors in this list is treated as the user's desired priority, while source
+    // is by default considered lower priority than all mirrors. Other cluster configuration,
+    // including (but not limited to) other repositoryMirrors objects,
+    // may impact the exact order mirrors are contacted in, or some mirrors may be contacted
+    // in parallel, so this should be considered a preference rather than a guarantee of ordering.
+    // +optional
+    Mirrors []string `json:"mirrors"`
+}
+```
+
+The ImageContentSourcePolicy CRD has the cluster-wide information about handle the mirrors only when pulling the images that are referenced by their digests.
+It will now watch for mirrors configured through spec `repositoryMirrors` to not require digest reference.
+
+An example ImageContentSourcePolicy file will look like:
+```yaml
+apiVersion: operator.openshift.io/v1alpha1
+kind: ImageContentSourcePolicy
+metadata:
+  name: ubi8repo
+spec:
+  repositoryMirrors:
+  - mirrors:
+    - example.io/example/ubi-minimal 
+    source: registry.access.redhat.com/ubi8/ubi-minimal
+    allowMirrorByTags: false # or equivalently, by leaving allowMirrorByTags unset
+```
+
+### User Stories
+
+#### As a user, I would like to pull images from mirror using tag, without digests reference
+The user need to define multiple ImageContentSourcePolicies, used by apps/manifests that don't use digests to pull
+the images when working with disconnected environments or pulling from registries that act as transparent pull-through proxy cache.
+The user can pull image without digest by configuring `repositoryMirrors` spec in the ImageContentSourcePolicy file.
+And create the ImageContentSourcePolicy project. Once this is done, the images can be pulled from the mirrors without the digest referenced.
+
+### Implementation Details/Notes/Constraints [optional]
+
+Implementing this enhancement requires changes in:
+- openshift/api
+- openshift/machine-config-operator
+
+This is an example of the ImageContentSourcePolicy file:
+
+```yaml
+apiVersion: operator.openshift.io/v1alpha1
+kind: ImageContentSourcePolicy
+metadata:
+  name: ubi8repo
+spec:
+  repositoryMirrors:
+  - mirrors:
+    - example.io/example/ubi-minimal 
+    source: registry.access.redhat.com/ubi8/ubi-minimal 
+```
+
+The ImageContentSourcePolicy file will create a drop-in file at `/etc/containers/registries.conf` that looks like:
+
+```toml
+unqualified-search-registries = ["registry.access.redhat.com", "docker.io"]
+[[registry]]
+  location = "registry.access.redhat.com/ubi8/"
+  insecure = false
+  blocked = false
+  # No mirror-by-digest-only = true configured
+  prefix = ""
+
+  [[registry.mirror]]
+    location = "example.io/example/ubi8-minimal"
+    insecure = false
+```
+
+### Risks and Mitigations
+
+Pulling images from mirror registries without the digest specifications could lead to returning different image version from different registry if the image tag mapping is out of sync. But the OpenShift core required image using digests to avoid different versions won't consume this feature at all, so it is not exposed to the risks that anyone who actually uses the feature will be exposed to.
+
+## Design Details
+
+### Test Plan
+
+Update the tests that are currently in the MCO to verify that `mirror-by-digest-only` not have been set when spec repositoryMirrors of the ImageContentSourcePolicy is created.
+
+### Graduation Criteria
+
+GA in openshift will be looked at to
+determine graduation.
+
+#### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+- Enumerate service level indicators (SLIs), expose SLIs as metrics
+- Write symptoms-based alerts for the component(s)
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+
+#### Removing a deprecated feature
+
+- With `repositoryMirrors` getting in, the current `repositoryDigestMirrors` can be deprecated since its functionality will also be fully satisfied by `repositoryMirrors` and we don't have to keep duplicated APIs.
+- In operator.openshift.io/v1alpha1, gently announce to the users the `repositoryDigestMirrors` will be deprecated. In operator.openshift.io/v1alpha2, the implementation of `repositoryDigestMirrors` will be removed and replaced by `repositoryMirrors`.
+
+### Upgrade / Downgrade Strategy
+
+Upgrades should not be affect. But the downgrade will be affected.
+Users upgraded and configured the `repositoryMirrors` spec will presumably have their CRI-O configurations clobbered and break their tag-mirroring-dependent workflows after they downgrade to a version that lacks support for the `repositoryMirrors` spec.
+We will not change current default behavior of `repositoryDigestMirrors`. We are just add new spec so users can configure it through the ImageContentSourcePolicy after the upgrade.
+
+### Version Skew Strategy
+
+Upgrade skew will not impact this feature. The MCO does not require skew check. CRI-O with n-2 OpenShift skew will still be able to handle the new property.
+
+## Implementation History
+
+Major milestones in the life cycle of a proposal should be tracked in `Implementation
+History`.
+
+## Drawbacks
+
+## Alternatives
+
+## Infrastructure Needed [optional]


QiWang19:
@umohnani8 PTAL

dmesser:
@QiWang19 Another impactful use case outside of disconnected environments is the use of registries that act as transparent pull-through proxy cache for upstream registries. Especially in light of some public registries introducing severe pull rate limits.

enhancements/api-review/add-mirror-by-digest-only.md
```diff
@@ -0,0 +1,226 @@
...
+  - mirrors:
+    - example.io/example/ubi-minimal 
+    source: registry.access.redhat.com/ubi8/ubi-minimal
+    mirror-by-digest-only: false
+```
+
+### User Stories
+
+#### As a user, I would like to pull images from mirror using tag, without digests reference
+The user can specify `mirror-by-digest-only: false` for the `source` under `repositoryDigestMirrors` in the ImageContentSourcePolicy file. And create the ImageContentSourcePolicy project. Once this is done, the images can be pulled from the mirrors without the digest referenced.
```
wking, line 77: This use-case should probably be expanded to explain why anybody would want to do this.  Unless you also have a way to mirror signatures (which sadly are still not something that gets represented in the image's blog graph, as far as I'm aware), it exposes you to man-in-the-middle attacks from anywhere between the image-build location and your image-consuming cluster.  For example, the image is vulnerable to replacement with a malicious image while transitioning between the build system and the mirror repository, while it is at rest on the mirror repository, and while it is transitioning between the mirror repository and the node about to launch a container.  Some of those transitions might be covered by TLS, but still, you need a lot of trust in all the steps in the chain of custody to be able to do this with any confidence.  Is transitioning consumers to by-digest pullspecs really that difficult?

enhancements/api-review/add-mirror-by-digest-only.md
```diff
@@ -0,0 +1,226 @@
...
+end to end tests.**
+
+#### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
+
+### Upgrade / Downgrade Strategy
+
+Not applicable. Upgrades and Downgrades should not be affected. The `mirror-by-digest-only` field already exists in the `/etc/containers/registries.conf` file. We will not change current default behavior. We are just exposing this option to the user so that they can configure it through the ImageContentSourcePolicy.
```
wking, line 201: Users with the new configuration who downgrade to a version that lacks support for the new feature will presumably have their CRI-O configurations clobbered and break their tag-mirroring-dependent workflows.

enhancements/api-review/add-mirror-by-digest-only.md
```diff
@@ -0,0 +1,226 @@
...
+
+How will the component handle version skew with other components?
+What are the guarantees? Make sure this is in the test plan.
+
+Consider the following in developing a version skew strategy for this
+enhancement:
+- During an upgrade, we will always have skew among components, how will this impact your work?
+- Does this enhancement involve coordinating behavior in the control plane and
+  in the kubelet? How does an n-2 kubelet without this feature available behave
+  when this feature is used?
```
wking, line 213: Probably worth saying something about the age and stability of the `mirror-by-digest-only` CRI-O configuration property, to demonstrate CRI-O with n-2 OpenShift skew will still be able to handle the new property.  Or maybe it's just the existing property being removed, in which case it should be even easier to make that case.

enhancements/api-review/add-mirror-by-digest-only.md
```diff
@@ -0,0 +1,226 @@
...
+- Document SLOs for the component
+- Conduct load testing
+
+**For non-optional features moving to GA, the graduation criteria must include
+end to end tests.**
+
+#### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
```
wking, line 197: Seems like the body of this section should be replaced with "Not applicable, because we are not deprecating any features." or similar

enhancements/api-review/add-mirror-by-digest-only.md
```diff
@@ -0,0 +1,226 @@
...
+No need to outline all of the test cases, just the general strategy. Anything
+that would count as tricky in the implementation and anything particularly
+challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations).
+
+### Graduation Criteria
+
+**Note:** *Section not required until targeted at a release.*
```
wking, line 146: Looks like you have a lot of boilerplate template content in this section.  If you're going to be born GA, probably just say that?

enhancements/api-review/add-mirror-by-digest-only.md
```diff
@@ -0,0 +1,226 @@
...
+To pull images from mirror registries without the digest specifications, set the MirrorByDigestOnly false,
+this could lead to returning different image from different registry if the image tag mapping is out of sync.
+
+## Design Details
+
+### Test Plan
+
+Update the tests that are currently in the MCO to verify that `mirror-by-digest-only` not have been set when the ImageContentSourcePolicy is created to drop it.
+
+**Note:** *Section not required until targeted at a release.*
```
wking, line 129: This and later in this section look like template boilerplate that can be dropped.

enhancements/api-review/add-mirror-by-digest-only.md
```diff
@@ -0,0 +1,226 @@
...
+apiVersion: operator.openshift.io/v1alpha1
+kind: ImageContentSourcePolicy
+metadata:
+  name: ubi8repo
+spec:
+  repositoryDigestMirrors:
+  - mirrors:
+    - example.io/example/ubi-minimal 
+    source: registry.access.redhat.com/ubi8/ubi-minimal
+    mirror-by-digest-only: false
```
wking, line 71: Having this false in a `repositoryDigestMirrors` entry feels strange to me.  Maybe it would be better to have `ImageContentSourcePolicy` grow a new `spec` property like `repositoryMirrors` that is less opinionated about digest-ness?  See also [this text][1] in the `repositoryDigestMirrors` godocs that would be incompatible with allowing by-tag mirroring.

[1]: https://github.com/openshift/api/blob/a99ffa1cac6709edf8f502b16890b16f9a557e00/operator/v1alpha1/types_image_content_source_policy.go#L23-L29

QiWang19: Sounds good to me.

QiWang19:
@wking PTAL

QiWang19:
@mrunalp PTAL

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,175 @@
...
+
+```go
+// Registry represents a registry.
+type Registry struct {
+  // ...
+
+	// If true, mirrors will only be used for digest pulls. Pulling images by
+	// tag can potentially yield different images, depending on which endpoint
+	// we pull from.  Forcing digest-pulls for mirrors avoids that issue.
+	MirrorByDigestOnly bool `toml:"mirror-by-digest-only,omitempty"`
```
wking, line 53: Defaulting to the unsafe `false` is riskier than I'd like.  If this is a true boolean in `registries.conf` (I don't know), I'd prefer `AllowTags` or some such here to default safe.

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,175 @@
...
+  blocked = false
+  # No mirror-by-digest-only = true configured
+  prefix = ""
+
+  [[registry.mirror]]
+    location = "example.io/example/ubi8-minimal"
+    insecure = false
+```
+
+If both `repositoryMirrors` and `repositoryDigestMirrors` are configured for same “source” repository, the ImageContentSourcePolicy will be rejected.
```
wking, line 117: With a property for digest-ness under `repositoryMirrors`, maybe time to go v1alpha2 and drop `repositoryDigestMirrors`?

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Risks and Mitigations
+
+To pull images from mirror registries without the digest specifications, this could lead to returning different image version from different registry if the image tag mapping is out of sync. This does not bring risks for ICSP usage since it does not influence the openshift deployments, which required image using digests to avoid different versions.
```
wking, line 121: I think the "This does not bring risks..." sentence is a bit off base.  It would be more clear to say the OpenShift core won't consume this feature at all, so it is not exposed to the risks that anyone who actually uses the feature will be exposed to.

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Goals
+
+- Enable the user to pull images through ImageContentSourcePolicy configured mirrors by tags.
+
+### Non-Goals
+
+## Proposal
+
+The Image API has an optional `MirrorByDigestOnly` filed with type `bool` to `Registry`.
```
adambkaplan, line 43: Which Image API? Are we talking `registries.conf` API for containers/image, or the OpenShift ImageStream APIs?

enhancements/api-review/add-repositoryMirrors-spec.md
adambkaplan: Isn't mirror by digest the default behavior today? If that is the case then then new API needs to be opt-in, as suggested with `AllowMirrorByTags`

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,175 @@
...
+The ImageContentSourcePolicy file will create a drop-in file at `/etc/containers/registries.conf` that looks like:
+
+```toml
+unqualified-search-registries = ["registry.access.redhat.com", "docker.io"]
+[[registry]]
+  location = "registry.access.redhat.com/ubi8/"
+  insecure = false
+  blocked = false
+  # No mirror-by-digest-only = true configured
+  prefix = ""
```
adambkaplan, line 110: Echoing my prior comment about having this behavior be opt-in, ICSP should expose an API to allow opt in:

```suggestion
    source: registry.access.redhat.com/ubi8/ubi-minimal 
    allowMirrorByTag: true
\`\`\`

The ImageContentSourcePolicy file will create a drop-in file at `/etc/containers/registries.conf` that looks like:

\`\`\`toml
unqualified-search-registries = ["registry.access.redhat.com", "docker.io"]
[[registry]]
  location = "registry.access.redhat.com/ubi8/"
  insecure = false
  blocked = false
  allowMirrorByTag = true
  prefix = ""
```

enhancements/api-review/add-repositoryMirrors-spec.md
QiWang19: If the repositoryDigestMirrors dropped, the digest requirement will not exist. Is it possible the customer still wants to limit the use of mirrors? This looks like a very rare case, I am not sure.

enhancements/api-review/add-repositoryMirrors-spec.md
QiWang19: The above is /etc/containers/registries.conf, it already have mirror-by-digest-only. We don't plan to change it in this enhancement.

enhancements/api-review/add-repositoryMirrors-spec.md
QiWang19: These lines of API definition is current `registries.conf`. Not the ImageCententSourcePolicy that this enhancement try to implement.

enhancements/api-review/add-repositoryMirrors-spec.md
QiWang19: The image API is registries.conf

enhancements/api-review/add-repositoryMirrors-spec.md
QiWang19: > Isn't mirror by digest the default behavior today? If that is the case then then new API needs to be opt-in, as suggested with `AllowMirrorByTags`

I have updated this enhancement to add a new spec of `repositoryMirrros` per comment https://github.com/openshift/enhancements/pull/690#discussion_r632822088

enhancements/api-review/add-repositoryMirrors-spec.md
adambkaplan: Please clarify then - there is a separate OpenShift Image API which makes the sentence ambiguous.

QiWang19:
The comment has disagreements about https://issues.redhat.com/browse/RFE-1608. Adding a new spec `RepositoryMirrors` to `ImageContentSourcePolicySpec`vs exposing a new bool field to users. What is our condense?

enhancements/api-review/add-repositoryMirrors-spec.md
wking: `mirror-by-digest-only: true` is sane, and protects you from malicious/compromised registries by forcing you to use by-digest pullspecs.  But you can ask for one if those with:

```yaml
apiVersion: operator.openshift.io/v1alpha1
kind: ImageContentSourcePolicy
metadata:
  name: ubi8repo
spec:
  repositoryMirrors:
  - mirrors:
    - example.io/example/ubi-minimal 
    source: registry.access.redhat.com/ubi8/ubi-minimal
  allowMirrorByTags: false
```

without needing a `repositoryDigestMirrors` entry.

enhancements/api-review/add-repositoryMirrors-spec.md
QiWang19: Fixed by adding allowMirrorByTag as opt-in.

QiWang19:
@wking @adambkaplan PTAL.

QiWang19:
@mrunalp PTAL

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,186 @@
...
+
+Today, the ImageContentSourcePolicy object sets up mirror configuration with `mirror-by-digest-only` property set to true, which leads to using mirrors only when images are referenced by digests. However, when working with disconnected environments, sometimes multiple ImageContentSourcePolicies are needed, some of them are used by apps/manifests that don't use digests when pulling the images.
+
+Adding the `repositoryMirrors` spec to ImageContentSourcePolicy to configure mirrors will make it possible to pull images from mirror using tags without requiring image digest reference.
+
+### Goals
+
+- Enable the user to pull images through ImageContentSourcePolicy configured mirrors by tags.
+
+### Non-Goals
```
mrunalp, line 39: Add a bullet that we won't enable this setting for OpenShift payload images.

enhancements/api-review/add-repositoryMirrors-spec.md
QiWang19: > With a property for digest-ness under `repositoryMirrors`, maybe time to go v1alpha2 and drop `repositoryDigestMirrors`?

@wking  How we do the API version change and the drop for `repositoryDigestMirrors`? Do we add the new `repositoryMirrors` and drop the old one at the same time in openshift/api?

Should this enhancement mention we are going to drop the `repositoryDigestMirrors`?

QiWang19:
@wking @mrunalp @adambkaplan PTAL. can we merge this enhancement?

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,189 @@
...
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Today, the ImageContentSourcePolicy object sets up mirror with `mirror-by-digest-only` property set to true. Requires images be pulled by digest only. This enhancement plans to add new spec to ImageContentSourcePolicy with configurable field `allowMirrorByTags`, so that mirror-by-digest-only` can be configured by ImageContentSourcePolicy.
```
wking, line 27: nit: unbalanced backticks around the second `mirror-by-digest-only`.

enhancements/api-review/add-repositoryMirrors-spec.md
wking: done in the current iteration.  Can we mark this thread resolved?

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,189 @@
...
+
+Adding the `repositoryMirrors` spec to ImageContentSourcePolicy to configure mirrors will make it possible to pull images from mirror using tags without requiring image digest reference.
+
+### Goals
+
+- Enable the user to pull images through ImageContentSourcePolicy configured mirrors by tags.
+
+### Non-Goals
+
+- The `allowMirrorByTags` will not be configurable or applied to OpenShift payload images.
```
wking, line 41: This wording seems to have been inspired by [this comment][1], but I think the wording could be more clear.  Maybe you're aiming for something like:

> - This proposal does not recommend using by-tag references for OpenShift release images.  Those should still be referenced by digest, regardless of whether `allowMirrorByTags` is enabled for repositories where the release images are mirrored.

[1]: https://github.com/openshift/enhancements/pull/690#discussion_r655659846

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Proposal
+
+The new spec `repositoryMirrors` will be added to ImageContentSourcePolicy. The `allowMirrorByTags` will be the configurable filed can allow using mirrors without requiring digest reference.
```
wking, line 45: Maybe the second sentence could be rephrased to:

> The `allowMirrorByTags` property will allow for mirroring by-tag images through configured mirrors.

or some such.

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,189 @@
...
+
+```go
+// RepositoryMirrors holds cluster-wide information about how to handle mirrors in the registries config.
+// Note: this is different from the RepositoryDigestMirrors that the mirrors only 
+// work when pulling the images that are referenced by their digests.
+type RepositoryMirrors struct {
+	// source is the repository that users refer to, e.g. in image pull specifications.
+	// +required
+	Source string `json:"source"`
+  // If true, mirrors will only be used for digest pulls. Pulling images by
```
wking, line 55: nit: inconsistent indenting flipping between tabs and spaces.

QiWang19: Fixed.

mrunalp:
/lgtm
/approve

We can fix the nits in follow-on PRs.

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,189 @@
...
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+
+#### Removing a deprecated feature
+
+- With `repositoryMirrors` getting in, the current `repositoryDigestMirrors` can be deprecated since its functionality will also be fully satisfied by `repositoryMirrors` and we don't have to keep duplicated APIs.
```
smarterclayton, line 167: This isn't a very strong argument, changing behavior without understanding the difference in behavior is a bug.

So I would probably put a hold on this feature with this as is without stronger justification.

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,189 @@
...
+    // +required
+    Source string `json:"source"`
+    // If true, mirrors will only be used for digest pulls. Pulling images by
+    // tag can potentially yield different images, depending on which endpoint
+    // we pull from.  Forcing digest-pulls for mirrors avoids that issue.
+    // +optional
+    AllowMirrorByTags bool `json:"allowMirrorByTags"`
+    // mirrors is one or more repositories that may also contain the same images.
+    // The order of mirrors in this list is treated as the user's desired priority, while source
+    // is by default considered lower priority than all mirrors. Other cluster configuration,
```
smarterclayton, line 62: Specifically "source being last" only makes sense for digest, it DOES not make sense for tag in all use cases.  Tag is fundamentally dependent on hoping you get the most up to date version of a tag during upgrade scenarios.  

Both image stream import and crio can use repository digest order consistently.  However, I'm not sure without an understanding of image import tag mirroring what we do here.

Is there another enhancement that covers ImageStream import behavior with repository tag design?  We should not be introducing a V1 api without a full design covering how import will interpret this, because mirroring of tags may actually need these feels to be designed.  If there is no additional design, I'd like to see one before I withdraw my hold on the implementation.

enhancements/api-review/add-repositoryMirrors-spec.md
```diff
@@ -0,0 +1,189 @@
...
+kind: ImageContentSourcePolicy
+metadata:
+  name: ubi8repo
+spec:
+  repositoryMirrors:
+  - mirrors:
+    - example.io/example/ubi-minimal 
+    source: registry.access.redhat.com/ubi8/ubi-minimal
+    allowMirrorByTags: false # or equivalently, by leaving allowMirrorByTags unset
+```
```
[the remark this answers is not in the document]
smarterclayton: I would probably recommend defining composition of mirrors in name order (ICSP A mirrors go before ICSP B mirrors)

PR #693, opened 2021-03-13
CONVENTIONS: Add section on limits

smarterclayton:
Some scenarios have been described where a limit is appropriate.
Add additional guidance.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -290,6 +290,24 @@ Resource request review history:
 * [BZ 1920159](https://bugzilla.redhat.com/show_bug.cgi?id=1920159) --
   for tracking changes in 4.7/4.8 for single-node RAN
 
+##### Allowed use of limits
+
+There is an exception process for adding limits to payload workloads. In the following
+scenarios workloads may be given permission to set limits:
+
+* Workloads that are scale-invariant and have a fixed memory or CPU cost no matter what the scale of the cluster or workload
+  * The workload must be demonstrated to use fixed memory or CPU, and the component must have a plan to detect whether the limit causes impact to end user workloads
+  * Workloads that have a dynamic range of usage are not a good candidate for fixed limits.
+  * Memory limits must be set to 25-50% above the observed maximum in standard e2e runs to ensure additional headroom
+  * Examples:
+    * A health check sidecar container that performs a very simple request
+    * A controller that looks at a fixed set of objects that cannot vary based on cluster size, workload scale, or end user action
+
+In general, per-operator dynamic calculation of limits is discouraged, and instead workloads are expected to regulate their own consumption where possible. In cases where the workload has a very large dynamic range, dynamic sharding or scaling using standard (or per operator workload) is to be preferred over dynamic limit setting.
+
+To get approval for an exception, contact a release architect and an exception will be recorded via code in the standard e2e suite which will allow changes to merge. Non-payload operators are encouraged to seek approval as well to better document use cases - other solutions may be suggested or problems in the platform may be uncovered.
+
+
 #### Taints and Tolerations
 
 An operator deployed by the CVO should run on control plane nodes and therefore should


smarterclayton:
/assign @dhellmann @markmc @derekwaynecarr 

https://github.com/openshift/origin/pull/25959 ratches in enforcement.

CONVENTIONS.md
```diff
@@ -290,6 +290,24 @@ Resource request review history:
 * [BZ 1920159](https://bugzilla.redhat.com/show_bug.cgi?id=1920159) --
   for tracking changes in 4.7/4.8 for single-node RAN
 
+##### Allowed use of limits
+
+There is an exception process for adding limits to payload workloads. In the following
+scenarios workloads may be given permission to set limits:
```
dhellmann, line 296: You could mention that there are exceptions on line 165, referring down to this section.

Gaining that permission is going to be extra work. As a developer, why would I do that? What are the benefits to me, users, the product, etc.?

CONVENTIONS.md
```diff
@@ -290,6 +290,24 @@ Resource request review history:
...
+
+* Workloads that are scale-invariant and have a fixed memory or CPU cost no matter what the scale of the cluster or workload
```
dhellmann, line 298: I'm not sure what "scale-invariant" means. Is that the same as "have a fixed memory or CPU cost..."?

CONVENTIONS.md
```diff
@@ -290,6 +290,24 @@ Resource request review history:
...
+  * The workload must be demonstrated to use fixed memory or CPU, and the component must have a plan to detect whether the limit causes impact to end user workloads
```
dhellmann, line 299: What sort of proof would be needed? I'm not sure how much detail we want to go into here, but if we can say that a type of test scenario is run, that metrics from CCX show steady consumption, or whatever, then that helps make the reader understand what they would have to do.

CONVENTIONS.md
```diff
@@ -290,6 +290,24 @@ Resource request review history:
...
+  * Workloads that have a dynamic range of usage are not a good candidate for fixed limits.
```
dhellmann, line 300: This is a reason not to grant an exception, but the previous 2 items are conditions under which an exception would be granted. Is this different from the statement on line 298?

CONVENTIONS.md
```diff
@@ -290,6 +290,24 @@ Resource request review history:
...
+  * Memory limits must be set to 25-50% above the observed maximum in standard e2e runs to ensure additional headroom
```
dhellmann, line 301: Is this a condition for granting an exception, or something that needs to happen after the exception is granted?

Are there similar guidelines for CPU limits?

CONVENTIONS.md
```diff
@@ -290,6 +290,24 @@ Resource request review history:
...
+  * Examples:
+    * A health check sidecar container that performs a very simple request
+    * A controller that looks at a fixed set of objects that cannot vary based on cluster size, workload scale, or end user action
+
+In general, per-operator dynamic calculation of limits is discouraged, and instead workloads are expected to regulate their own consumption where possible. In cases where the workload has a very large dynamic range, dynamic sharding or scaling using standard (or per operator workload) is to be preferred over dynamic limit setting.
```
dhellmann, line 306: What is "dynamic sharding"?

I think the implication from the start of this paragraph is that if we're going to set limits, they need to be static regardless of the cluster profile or topology. Is that correct? Why? Maybe something like

> Workloads are expected to regular their own resource consumption instead of using explicit limits, where possible, because this ensures optimal performance in all configurations. In cases where the workload has a very large dynamic range, dynamic sharding or scaling using standard (or per operator workload) should be used instead of limit setting because it takes advantage of the scheduler's ability to distribute work across the cluster. If an exception for the use of limits is approved by a release architect, static limits are preferred over dynamic calculations of limits because changing the limits of a Pod requires restarting a component, which disrupts service, adds load to the cluster, and risks scheduling failure.

I've guessed at some reasons, but maybe you had other things in mind. Once we work out the wording, I think this paragraph would make a good introduction to this sub-section and could move up to around line ~294.

CONVENTIONS.md
```diff
@@ -290,6 +290,24 @@ Resource request review history:
...
+
+To get approval for an exception, contact a release architect and an exception will be recorded via code in the standard e2e suite which will allow changes to merge. Non-payload operators are encouraged to seek approval as well to better document use cases - other solutions may be suggested or problems in the platform may be uncovered.
```
dhellmann, line 308: Is there a team page or other reference for "release architect"? If I'm new to the org, how do I find one?

The second sentence here feels like a different train of thought, so I suggest creating a new paragraph for it.

CONVENTIONS.md
smarterclayton: Yeah, would probably add a benefit for each one.  For the current one, it would be:

```
Setting a limit in these cases will ensure that bugs introduced that cause memory leaks will fail in more predictable ways, and more quickly identify the failure to the responsible component.
```

but that's the only advantage (I think)

CONVENTIONS.md
smarterclayton: It is, I could just say "scale-invariant (have fixed memory or CPU usage regardless of the size of cluster, size of node, or size of workload and usage)"

CONVENTIONS.md
smarterclayton: Probably not

CONVENTIONS.md
smarterclayton: This is probably just addition to the advice up above about how to pick your limit, so it could go up there with a pointer back to "limits need a problem"

CONVENTIONS.md
smarterclayton: Hrm, it's how to get an exception, the second sentence is "who to get it from"

CONVENTIONS.md
dhellmann: That advantage seems also to come with the downsides mentioned up above. :-) If there's a specific example of a component that would benefit from memory limits, it might help make it easier for the reader to decide whether their thing fits the pattern for even worrying about limits.

CONVENTIONS.md
dhellmann: I try to leave jargon out unless the word is an important aspect of what is being explained or the text is less readable (because you have to repeat a longer phrase over and over). In this case, I would cut "scale-invariant" in favor of the simpler terms.

CONVENTIONS.md
dhellmann: Maybe combine it with the first point, to reinforce the meaning.

CONVENTIONS.md
dhellmann: It seems like this new section breaks down into a couple of points

* You mostly don't want to use limits, but sometimes you do. (An intro that reinforces what was said in the parent section.)
* If you think you want limits, here are the criteria that have to be met for them to be allowed. (The 2 items on lines 298-299. Are there more?)
* After those criteria are met, here's how you pick the limits safely. (Line 301 covers it for memory, do we have similar advice for other limits?)
* After you have limits, here's the process for having them approved. (The stuff currently on line 308.)

So, I think you want this here, and maybe want to expand it, but it should be its own paragraph instead of a part of the list that it's currently in.

CONVENTIONS.md
dhellmann: I wasn't clear in my feedback.

> To get approval for an exception, contact a release architect and an exception will be recorded via code in the standard e2e suite which will allow changes to merge. 

This sentence is doing a lot of work and could be expanded a bit to provide more context for the reader.

1. It introduces the idea that there is a separate, required process to gain approval for using limits. Is this section of this document meant to be the full description of that process? If so, is there a formal process or is it "start a conversation with a release architect"? If it is more complicated than we want to lay out in full here, is there another doc we could link to?
2. It says that using limits requires review by a specific group of people, rather than normal code review by the team that owns the component being limited. Maybe that's the only "process"? How does someone gain the attention of that group? Who is in the group (how do I find them if I want to talk to them)?
3. It also says that we use a CI job to enforce that most components do not have limits. Is that a check of the YAML files or is there some characteristic of the e2e test suite that prevents limits from working as a side-effect? We may not want to mention which CI job enforces the rule, since the name may change over time, but can we link to a list of CI jobs with their purposes so that someone could find the job definition to see how it works?

> Non-payload operators are encouraged to seek approval as well to better document use cases - other solutions may be suggested or problems in the platform may be uncovered.

This sentence suggests that things that are not part of the release payload should also follow these rules, which explains the scope of the rules discussed above but isn't really part of the process of applying them. Maybe we should talk about the scope at the start of the new section? Something like "OpenShift components, including those in the release payload and those delivered to customers separately, should not have resource limits applied in most circumstances. <updated content from line 306 here>"

dhellmann:
@smarterclayton are you still working on this?

PR #694, opened 2021-03-15
Amend egress router CNI controller

danielmellado:
This PR amends the job done for the egress router CNI controller to the
existing EP.


russellb:
an update to reflect what was implemented

/approve
/lgtm

PR #695, opened 2021-03-15
move CVO upgrades doc to enhancements

deads2k:
Moves CVO docs to enhancements.  The tombstone in the original location is here: https://github.com/openshift/cluster-version-operator/pull/631 .  These docs cover integration with other components, what those components need to do to participate, what impact they have on each other, and how a cluster upgrades.

See https://github.com/openshift/enhancements/#is-my-thing-an-enhancement for criteria.


enhancements/cluster-version-operator/upgrades.md
```diff
@@ -0,0 +1,83 @@
+# Upgrades and order
```
wking, line 1: Can this go under the existing `enhancements/update/` directory?

enhancements/cluster-version-operator/upgrades.md
```diff
@@ -0,0 +1,83 @@
...
+
+In the CVO, upgrades are performed in the order described in the payload (lexographic), with operators only running in parallel if they share the same file prefix `0000_NN_` and differ by the next chunk `OPERATOR`, i.e. `0000_70_authentication-operator_*` and `0000_70_image-registry-operator_*` are run in parallel.
```
wking, line 3: They are parallel graph-node entries (it might be worth bringing over [this graph][1] to clarify graph nodes).  That means they could run in parallel, but they might not; it depends on how many workers are available.

Also, let's use some modern examples, like:

```console
$ oc adm release extract --to=manifests quay.io/openshift-release-dev/ocp-release:4.7.2-x86_64
Extracted release payload from digest sha256:83fca12e93240b503f88ec192be5ff0d6dfe750f81e8b5ef71af991337d7c584 created at 2021-03-10T11:32:57Z
$ ls manifests/0000_70_* | cat
manifests/0000_70_cluster-network-operator_00_namespace.yaml
...
manifests/0000_70_cluster-network-operator_05_clusteroperator.yaml
manifests/0000_70_console-operator.crd.yaml
manifests/0000_70_dns-operator_00-cluster-role.yaml
...
manifests/0000_70_dns-operator_03-cluster-operator.yaml
```

So:

>  ..., i.e. `0000_70_cluster-network-operator_*` and `0000_70_dns-operator_*` are parallel graph nodes, and may be run in parallel.

[1]: https://github.com/openshift/cluster-version-operator/blob/69099669668f533fa6659a85c26d607505d8d558/docs/user/reconciliation.md#manifest-graph

enhancements/cluster-version-operator/upgrades.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+Which in practice can be described in runlevels:
+
+```
+0000_10_*: config-operator
+0000_20_*: kube-apiserver
+0000_25_*: kube scheduler and controller manager
+0000_30_*: other apiservers: openshift and machine
+0000_40_*: reserved
```
wking, line 53: Is this really reserved?  Looks like `kube-storage-version-migrator-operator` is running there in 4.7.2, although I'm not sure what fed into that decision.

enhancements/cluster-version-operator/upgrades.md
```diff
@@ -0,0 +1,83 @@
...
+## Generalized ordering
+
+The following rough order is defined for how upgrades should proceed:
+
+```
+config-operator
+
+kube-apiserver
+
+kcm/ksched
```
wking, line 26: For an enhancement-level doc, I think it's probably worth dropping the code block here and making this whole section an enumerated list with English entries:

> 1. config-operator
> 1. Kubernetes API-server
> 1. Kubernetes controller manager and scheduler
> ...

enhancements/cluster-version-operator/upgrades.md
deads2k: > Is this really reserved? Looks like `kube-storage-version-migrator-operator` is running there in 4.7.2, although I'm not sure what fed into that decision.

probably a lack of this doc existing... :)

enhancements/cluster-version-operator/upgrades.md
deads2k: > They are parallel graph-node entries (it might be worth bringing over [this graph](https://github.com/openshift/cluster-version-operator/blob/69099669668f533fa6659a85c26d607505d8d558/docs/user/reconciliation.md#manifest-graph) to clarify graph nodes). That means they could run in parallel, but they might not; it depends on how many workers are available.

How much of that doc should move over?  Read through, it actually answers question I've had.

wking: Also, if the expectation is covering the API around operator-provided manifests, we should include the 0000 defaulting stuff [from `oc`][1]:

> Operators are expected to host the config they need to be installed to a cluster in the `/manifests` directory in their image. This command iterates over a set of operator images and extracts those manifests into a single, ordered list of Kubernetes objects that can then be iteratively updated on a cluster by the cluster version operator when it is time to perform an update. Manifest files are renamed to `0000_70_<image_name>_<filename>` by default, and an operator author that needs to provide a global-ordered file (before or after other operators) should prepend `0000_NN_<component>_` to their filename, which instructs the release builder to not assign a component prefix. Only images in the input that have the image label `io.openshift.release.operator=true` will have manifests loaded.

Because folks who don't care about their order in the payload can reduce some mental overhead by leaving `0000_NN_<component>_` off and relying on `oc` to inject it for them.

[1]: https://github.com/openshift/oc/blob/bdd8f89997ce224cb6f186716c18bee872d3b318/pkg/cli/admin/release/new.go#L73-L82

deads2k:
I'd rather move the collateral we already have an tidy from there (unless the information is wrong). I'm happy to move whatever other CVO docs make sense to start though.  https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/operators.md also looks prime for moving to me.

enhancements/cluster-version-operator/upgrades.md
```diff
@@ -0,0 +1,83 @@
...
+
+Priorities during upgrade:
+
+1. Simplify the problems that can occur
+2. Ensure applications aren't disrupted during upgrade
+3. Complete within a reasonable time period (30m-1h for control plane)
+
+During upgrade we bias towards predictable ordering for operators that lack sophistication about detecting their prerequisites. Over time, operators should be better at detecting their prerequisites without overcomplicating or risking the predictability of upgrades.
+
+Currently, upgrades proceed in operator order without distinguishing between node and control plane components. Future improvements may allow nodes to upgrade independently and at different schedules to reduce production impact. This in turn complicates the investment operator teams must make in testing and understanding how to version their control plane components independently of node infrastructure.
```
wking, line 13: Worth including [this blurb][1]:

> Ordering is only applied during upgrades, where some components rely on another component being updated first. As a convenience, the CVO guarantees that components at an earlier run level will be created or updated before your component is invoked. Note however that components without `ClusterOperator` objects defined may not be fully deployed when your component is executed, so always ensure your prerequisites know that they must correctly obey the `ClusterOperator` protocol to be available. More sophisticated components should observe the prerequisite `ClusterOperator`s directly and use the `versions` field to enforce safety.

What about [the `Kind` hints][2]?  [The extension restrictions][3]?  [The `create-only` annotation][4]?  The `image-references docs][5]?   Probably also worth linking [the profile enhancement][6], because the CVO will ignore manifests that don't declare themselves to be members of the current profile.

[1]: https://github.com/openshift/cluster-version-operator/blame/69099669668f533fa6659a85c26d607505d8d558/docs/dev/operators.md#L32
[2]: https://github.com/openshift/cluster-version-operator/blob/69099669668f533fa6659a85c26d607505d8d558/docs/dev/operators.md#what-do-i-put-in-manifests
[3]: https://github.com/openshift/cluster-version-operator/blame/69099669668f533fa6659a85c26d607505d8d558/docs/dev/operators.md#L98
[4]: https://github.com/openshift/cluster-version-operator/blame/69099669668f533fa6659a85c26d607505d8d558/docs/dev/operators.md#L98
[5]: https://github.com/openshift/cluster-version-operator/blob/69099669668f533fa6659a85c26d607505d8d558/docs/dev/operators.md#how-do-i-ensure-the-right-images-get-used-by-my-manifests
[6]: https://github.com/openshift/enhancements/blob/ffe55d0c2bb63b52b421fb567de66a62e567a04f/enhancements/update/cluster-profiles.md

enhancements/cluster-version-operator/upgrades.md
wking: I've just filed a PR to get the storage-version-migrator into the default runlevel: openshift/cluster-kube-storage-version-migrator-operator#46, so we can keep `reserved` here and mark this thread resolved.

enhancements/cluster-version-operator/upgrades.md
wking: > How much of that doc should move over? Read through, it actually answers question I've had.

That doc is aimed at users trying to understand how the CVO works.  That's distinct from, but not completely orthogonal to the stuff that operator maintainers need to understand to set up their manifests.  Not clear to me whether you want to attempt a single doc aimed at both, or you're scoping this to operator maintainers, or scoping this to just admins, or what.

enhancements/cluster-version-operator/operators.md
```diff
@@ -0,0 +1,161 @@
...
+
+- 00-04 - CVO
+- 05 - cluster-config-operator
+- 07 - Network operator
+- 08 - DNS operator
+- 09 - Service certificate authority and machine approver
+- 10-29 - Kubernetes operators (master team)
+- 30-39 - Machine API
+- 50-59 - Operator-lifecycle manager
+- 60-69 - OpenShift core operators (master team)
```
wking, line 127: This list is stale vs. the one in `upgrades.md`.  Can we drop it?

LalatenduMohanty (approved):
/lgtm

sdodson:
/approve
Cause it makes sense and this was agreed upon.
/lgtm cancel
Because it doesn't pass the linter, need to update to lass the linter.

sdodson (approved):

deads2k:
serve the machine!  linter is now happy, reapplying lgtm

PR #697, opened 2021-03-17
ingress: Add transition-ingress-from-beta-to-stable enhancement

Miciah:
This enhancement updates the [ingress-to-route](https://github.com/openshift/openshift-controller-manager/blob/master/pkg/route/ingress/ingress.go) controller to use the stable `networking.k8s.io/v1` version of the Ingress API.  Although the API server already supports the v1 API version without this enhancement, the ingress-to-route controller requires changes to transition from using the v1beta1 client to using the v1 client and to support new v1 features.  These new features include the `spec.pathType` and `spec.ingressClassType` fields as well as the associated IngressClass API.  This enhancement does *not* extend the Route API to accommodate new features in the Ingress API; these are only supported to the extent that they are compatible with the Route API.

---

Implementation PRs:

- [ ] [openshift/cluster-ingress-operator#574 WIP: ingressclass: New controller.](https://github.com/openshift/cluster-ingress-operator/pull/574)
- [ ] [openshift/openshift-controller-manager#172 WIP: ingress: Migrate to networking.k8s.io/v1.](https://github.com/openshift/openshift-controller-manager/pull/172)
- [ ] [openshift/api#873 route: Define the ingress-to-route controller name](https://github.com/openshift/api/pull/873)
- [ ] ~TODO openshift/cluster-monitoring-operator PR for alerts.~

### DIFF
--- a/enhancements/ingress/transition-ingress-from-beta-to-stable.md
+++ b/enhancements/ingress/transition-ingress-from-beta-to-stable.md
@@ -1,0 +1,319 @@
+---
+title: transition-ingress-from-beta-to-stable
+authors:
+  - "@Miciah"
+reviewers:
+  - "@candita"
+  - "@danehans"
+  - "@frobware"
+  - "@knobunc"
+  - "@miheer"
+  - "@rfredette"
+  - "@sgreene570"
+approvers:
+  - "@danehans"
+  - "@frobware"
+  - "@knobunc"
+creation-date: 2021-03-09
+last-updated: 2021-03-09
+status: provisional
+see-also:
+replaces:
+superseded-by:
+---
+
+# Transitioning Ingress API from Beta to Stable
+
+This enhancement transitions OpenShift fully onto the stable
+`networking.k8s.io/v1` API version of the Ingress API.
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement updates the
+[ingress-to-route](https://github.com/openshift/openshift-controller-manager/blob/master/pkg/route/ingress/ingress.go)
+controller to use the stable `networking.k8s.io/v1` version of the Ingress API.
+Although the API server already supports the v1 API version without this
+enhancement, the ingress-to-route controller requires changes to transition from
+using the v1beta1 client to using the v1 client and to support new v1 features.
+These new features include the `spec.pathType` and `spec.ingressClassType`
+fields as well as the associated IngressClass API.  This enhancement does *not*
+extend the Route API to accommodate new features in the Ingress API; these are
+only supported to the extent that they are compatible with the Route API.
+
+## Motivation
+
+Some users want to use new features in the `networking.k8s.io/v1` API version,
+and Kubernetes 1.22 will remove the `networking.k8s.io/v1beta1` API version.
+
+The IngressClass API is of particular utility.  With this API, an administrator
+can specify which Ingresses OpenShift should publish and which Ingresses
+OpenShift should ignore, on the assumption that some third-party ingress
+controller publishes them.  This is particularly important with respect to
+status updates, as the Ingress API does not provide a way to differentiate
+status published by one ingress controller from status published by another
+ingress controller, which has the consequence that OpenShift's updates to an
+Ingress's status can interfere with the operation of third-party ingress
+controllers or other controllers that read the Ingress's status.  Thus
+supporting this API is important for accommodating third-party ingress
+controllers in OpenShift clusters.
+
+The following table lists new API fields in Ingress v1 and the status of their
+support with the implementation of this enhancement:
+
+| Field                                          | Supported     |
+|------------------------------------------------|---------------|
+| `spec.defaultBackend`                          | No            |
+| `spec.ingressClassName`                        | Yes[^1]       |
+| `spec.rules[*].http.paths[*].backend.service`  | Yes           |
+| `spec.rules[*].http.paths[*].backend.resource` | No            |
+| `spec.rules[*].http.paths[*].pathType`         | Partially[^2] |
+
+
+### Goals
+
+1. Honor `spec.ingressClassName`.
+2. Prepare for the removal of the v1beta1 API version.
+
+### Non-Goals
+
+1. Extend the Route API to support new Ingress features.
+2. Implement `spec.rules[*].http.paths[*].pathType: Exact`.
+3. Implement new functionality (such as regular expression matching) for `spec.rules[*].http.paths[*].pathType: ImplementationSpecific`.
+4. Implement `spec.defaultBackend`.
+5. Enable Ingresses to use backends that are not Services.
+
+## Proposal
+
+First, the ingress-to-route controller is updated to use the v1 client libraries
+and API types.
+
+Next, the ingress-to-route controller is extended to check for the
+`kubernetes.io/ingress.class` annotation and `spec.ingressClassName` field on
+Ingresses.  If either of these is set, the ingress-to-route controller checks if
+the specified IngressClass exists and, if it does, what its `spec.controller`
+field value is.  If the Ingress specifies an IngressClass that does not exist or
+that does not specify `openshift.io/ingress-to-route` for `spec.controller`, the
+ingress-to-route controller ignores the Ingress.
+
+Additionally, the ingress-to-route controller is extended to check the value of
+`spec.rules[*].http.paths[*].pathType` on Ingresses.  If an Ingress rule
+specifies the value "Exact" for this field, the ingress-to-route controller
+ignores the rule.
+
+Finally, the ingress operator is extended with a new controller that creates an
+IngressClass for each IngressController.  When creating the IngressClass for the
+default IngressController, the operator annotates the IngressClass with the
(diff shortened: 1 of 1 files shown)


enhancements/ingress/transition-ingress-from-beta-to-stable.md
```diff
@@ -0,0 +1,302 @@
...
+The ingress operator is also modified to list all Ingresses and Routes in the
+cluster and publish a metric for Routes that were created for Ingresses that
+OpenShift no longer manages.  This metric is used in one of the alerting rules
+described below.
+
+Two alerting rules are added to cluster-monitoring-operator (see Risks and
+Mitigations for the purpose of these alerts):
+
+* An alert for Routes that were created from Ingresses that OpenShift is no longer managing.
+* An alert for Ingresses older than 1 day that do not specify `spec.ingressClassName`.
```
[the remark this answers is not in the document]
Miciah: I don't know what the use case for leaving `spec.ingressClassName` is, but I'm assuming one exists, or else the field would be required for new ingresses.  My thinking was that the use case probably involved a developer creating an incomplete ingress and then updating it to make it complete, in which case a high-tenancy cluster with a lot of churn could continuously have _some_ incomplete ingress at any point in time over a 1-day period, so it makes more sense to alert when a specific ingress has been incomplete for &gt;1 day.  This thinking could be entirely wrong though, and I don't really know why `spec.ingressClassName` is optional.

enhancements/ingress/transition-ingress-from-beta-to-stable.md
```diff
@@ -0,0 +1,302 @@
...
+
+The ingress operator is modified to create and manage an IngressClass for each
+IngressController.
+
+The ingress operator is also modified to list all Ingresses and Routes in the
+cluster and publish a metric for Routes that were created for Ingresses that
+OpenShift no longer manages.  This metric is used in one of the alerting rules
+described below.
+
+Two alerting rules are added to cluster-monitoring-operator (see Risks and
```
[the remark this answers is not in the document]
Miciah: Oh, I forgot we could add rules in https://github.com/openshift/cluster-ingress-operator/blob/master/manifests/0000_90_ingress-operator_03_prometheusrules.yaml.  Yeah, let's do that.

enhancements/ingress/transition-ingress-from-beta-to-stable.md
```diff
@@ -0,0 +1,302 @@
...
+these Ingresses before this enhancement, and moreover, the Ingress API
+specification explicitly states that ingress controllers "should" (not "must")
+ignore Ingresses that do not specify `spec.ingressClassName`.  Thus this risk is
+somewhat mitigated by long-standing circumstances and existing documentation.
+Furthermore, if the cluster administrator does not want OpenShift to expose
+Ingresses that do not explicitly specify an OpenShift-owned IngressClass, then
+the administrator has the option of creating a custom IngressClass and
+annotating it with `ingressclass.kubernetes.io/is-default-class=true`.
+
+Finally, it is possible that a user could have created an Ingress, _did_ specify
```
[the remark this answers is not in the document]
Miciah: I mean it as "could have created [...], [and] _did_ specify [...], and nevertheless intended [...]".  I can try to rewrite it to be clearer.

enhancements/ingress/transition-ingress-from-beta-to-stable.md
```diff
@@ -0,0 +1,302 @@
...
+This enhancement updates the
+[ingress-to-route](https://github.com/openshift/openshift-controller-manager/blob/master/pkg/route/ingress/ingress.go)
+controller to use the stable `networking.k8s.io/v1` version of the Ingress API.
+Although the API server already supports the v1 API version without this
+enhancement, the ingress-to-route controller requires changes to transition from
+using the v1beta1 client to using the v1 client and to support new v1 features.
+These new features include the `spec.pathType` and `spec.ingressClassType`
+fields as well as the associated IngressClass API.  This enhancement does *not*
+extend the Route API to accommodate new features in the Ingress API; these are
+only supported to the extent that they are compatible with the Route API.
```
candita, line 49: Does "these" in L48 refer to the "new features in the Ingress API" ?  If so, it would be useful to have a table of those that are compatible with the existing Route API and those that are not.  This would help describe goal #1 below better.

enhancements/ingress/transition-ingress-from-beta-to-stable.md
```diff
@@ -0,0 +1,302 @@
...
+status published by one ingress controller from status published by another
+ingress controller, which has the consequence that OpenShift's updates to an
+Ingress's status can interfere with the operation of third-party ingress
+controllers or other controllers that read the Ingress's status.  Thus
+supporting this API is important for accommodating third-party ingress
+controllers in OpenShift clusters.
+
+### Goals
+
+1. Enable users to use new Ingress v1 API features where feasible.
```
candita, line 70: Which features are actually expected to be feasible?  If we are not able to enumerate those features, it is best to explicitly point out the ones we tested and will support.

enhancements/ingress/transition-ingress-from-beta-to-stable.md
```diff
@@ -0,0 +1,302 @@
...
+Ingresses.  If either of these is set, the ingress-to-route controller checks if
+the specified IngressClass exists and, if it does, what its `spec.controller`
+field value is.  If the Ingress specifies an IngressClass that does not exist or
+that does not specify `openshift.io/ingress-to-route` for `spec.controller`, the
+ingress-to-route controller ignores the Ingress.
+
+Additionally, the ingress-to-route controller is extended to check the value of
+`spec.rules[*].http.paths[*].pathType` on Ingresses.  If an Ingress rule
+specifies the value "Exact" for this field, the ingress-to-route controller
+ignores the rule.
```
candita, line 98: To me it seems counterintuitive that "Exact" means to ignore the rule.  Is this just one of the infeasible features?

enhancements/ingress/transition-ingress-from-beta-to-stable.md
```diff
@@ -0,0 +1,302 @@
...
+
+Finally, the ingress operator is extended with a new controller that creates an
+IngressClass for each IngressController.  When creating the IngressClass for the
+default IngressController, the operator annotates the IngressClass with the
+`ingressclass.kubernetes.io/is-default-class` annotation if no other
+IngressClass already has that annotation.  Note that the operator does not add
+the annotation when reconciling already created IngressClass objects as the
+administrator may intentionally have configured the cluster to have no default
+IngressClass.
```
candita, line 107: Perhaps there should be no automatic "is-default-class" annotation if we want to allow the cluster to have no default.  If the admin may intentionally want no default, it doesn't make sense for them to have to go and remove it after we add it.

enhancements/ingress/transition-ingress-from-beta-to-stable.md
```diff
@@ -0,0 +1,302 @@
...
+The ingress-to-route controller is modified to use the `networking.k8s.io/v1`
+client and API definitions and to respect `spec.ingressClassName` as described
+above.
+
+The ingress operator is modified to create and manage an IngressClass for each
+IngressController.
+
+As follow-up work, we are considering modifying the ingress operator to list all
+Ingresses and Routes in the cluster and publish a metric for Routes that were
+created for Ingresses that OpenShift no longer manages.  This metric could used
```
candita, line 169: "could used" -> "could be used"

enhancements/ingress/transition-ingress-from-beta-to-stable.md
```diff
@@ -0,0 +1,302 @@
...
+
+An alternative to exposing Ingresses that *do not* specify
+`spec.ingressClassName` would be not to expose them but instead to rely on a
+release note to prompt cluster administrators to verify that Ingresses specified
+`spec.ingressClassName` as needed prior to upgrade.  This alternative would pose
+an unknown but likely high risk of breaking existing users.
+
+An alternative to exposing Ingresses that *do* specify `spec.ingressClassName`
+would be not to expose them (i.e., delete existing associated Routes).  This
+alternative would pose an unknown but likely moderate risk to high of breaking
```
candita, line 295: "moderate risk to high" -> "moderate to high risk" ?

enhancements/ingress/transition-ingress-from-beta-to-stable.md
Miciah: It does.  I'll add a table.  Thanks for the suggestion!

enhancements/ingress/transition-ingress-from-beta-to-stable.md
Miciah: For new features, it's pretty much just `spec.ingressClassName`.  I'll remove this list item because it is vague and doesn't add anything.

enhancements/ingress/transition-ingress-from-beta-to-stable.md
Miciah: Right, it is infeasible because the Route API does not have a corresponding way to express "match this exact path" (as opposed to "match this prefix").

enhancements/ingress/transition-ingress-from-beta-to-stable.md
Miciah: I don't have a clear understanding of why an administrator _wouldn't_ want a default ingressclass; I just didn't want to preclude the possibility, so I added this description of how to configure it.  Having no default ingressclass means that creating an ingress without specifying an ingressclass does not expose the ingress, which is a change in behavior from before this enhancement.  Having the default ingresscontroller's ingressclass be the default one eliminates a configuration step for administrators who want it to be default, which I expect is the most common and least confusing configuration for administrators and users.

sgreene570:
Looks good.

/lgtm

PR #699, opened 2021-03-17
Updated for schema->index change

alanconway:
- replaced pipeline.schema with elasticsearch output.index
- added outputDefaults to apply settings without redefining the default output
- updated examples

/cc @vimalk78
/cc @sichvoge
/cc @jcantril

### DIFF
--- a/enhancements/cluster-logging/forwarding-json-structured-logs.md
+++ b/enhancements/cluster-logging/forwarding-json-structured-logs.md
@@ -52,7 +52,8 @@ The current output record looks like:
 }
 ```
 
-This proposal enables an alternate form of output record including a `structured` object field for JSON log entries and user-defined `schema` name to identify the format of the entry.
+This proposal enables an alternate form of output record including a `structured` object field for JSON log entries.
+For Elasticsearch outputs, new configuration is added to determine the index to use.
 
 ```json
 {
@@ -60,7 +61,6 @@ This proposal enables an alternate form of output record including a `structured
     "name":"fred",
     "home":"bedrock",
   }",
-  "schema": "flintstones-schema",
   "kubernetes":{"labels":{"app.kubernetes.io/name": "flintstones", ...}},
   ...
 }
@@ -68,8 +68,9 @@ This proposal enables an alternate form of output record including a `structured
 
 This proposal describes
 
-* extensions to the logging [data model][data_model] - `structured` and `schema` fields.
+* extensions to the logging [data model][data_model] - `structured` field.
 * extensions to the `ClusterLogForwarder` API to configure JSON parsing and forwarding.
+* extensions to the `ClusterLogForwarder` API to select an index.
 * indexing structured records in current and future Elasticsearch stores.
 * replacing the [defunct MERGE_JSON_LOG][defunct_merge] feature.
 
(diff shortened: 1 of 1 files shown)


enhancements/cluster-logging/forwarding-json-structured-logs.md
```diff
@@ -84,155 +85,138 @@ Logging terminology can have overlapping meanings, this document uses the follow
...
+
+For each log entry; if `parse: json` is set _and_ the entry is valid JSON, the output record will include a `structured` field _equivalent_ to the JSON entry.
+It may differ in field order and use of white-space.
+
+All record will have a `message` field for backwards compatibility.
+
+#### Output  configuration for type:elasticsearch
+
+New fields for output type `elasticsearch`
+* `indexKey`: (string, optional) Use value of meta-data key as `index` value, if present.
```
sichvoge, line 136: Is there a reason why we have `indexKey` and `indexName` and not just `index`? Index could be a string and we could recognize if it starts and ends with `"` or is one of the valid selectors (`kubernetes.namespaceName` or `kubernetes.labels.<key>`).

With that we could save to test for all the possible combinations if one is missing, etc.

enhancements/cluster-logging/forwarding-json-structured-logs.md
```diff
@@ -84,155 +85,138 @@ Logging terminology can have overlapping meanings, this document uses the follow
...
   These keys are supported:
-  - `kubernetes.namespaceName`: Use the namespace name as the schema name.
+  - `kubernetes.namespaceName`: Use the namespace name as the index name.
   - `kubernetes.labels.<key>`: Use the string value of kubernetes label with key `<key>`.
+  - `openshift.labels.<key>`: Use the string value of an openshift label with key `<key>` (see [](forwarder-tagging.md))
```
sichvoge, line 140: What is this?

enhancements/cluster-logging/forwarding-json-structured-logs.md
```diff
@@ -84,155 +85,138 @@ Logging terminology can have overlapping meanings, this document uses the follow
...
 
-### User Stories
+#### Default output configuration
 
-#### I want Elasticsearch to index structured logs with an index-per-schema
+To allow index settings on the default elasticsearch output, there is a new section in the ClusterLogForwarder spec:
 
-Setting the `schema` field in outgoing JSON records will send them to different
-Elasticsearch indices. See the [Elasticsearch notes](#es-notes) for more details.
+* outputDefaults (type map, optional):
```
sichvoge, line 152: I guess we only want to support this for when people use JSON right? cc/ @jcantrill

sichvoge (changes requested):
Added comments.

enhancements/cluster-logging/forwarding-json-structured-logs.md
alanconway: It is very unusual to require users to include '"' in a string value, I think that would be more confusing than helpful. 
The set of valid selectors is not closed, we might add others in future so, it is risky to guess whether the user intends a given string to be interpreted literally or as a key - the meaning might change accidentally if we include other valid keys in future. 
It's a common pattern, you'll find it in our syslog plugin for example.

enhancements/cluster-logging/forwarding-json-structured-logs.md
alanconway: Its a markdown link to the forwarder-tagging.md proposal. If you render the markdown as HTML (e.g. view it on github) it becomes a real link.

enhancements/cluster-logging/forwarding-json-structured-logs.md
alanconway: I suggest we implement it for all output types. It is as simple as saving a JSON snippet per output type and applying it before applying config from the API. It is potentially useful in its own right (e.g. I want to create 50 syslog outputs all with the same config except for the target URL, so I create a default output and 50 syslog outputs with just a URL) 
I think it would be harder to clearly document a restricted version of this (and deal with all the users who are surprised when it doesn't work outside those restrictions) than to just implement it properly.

alanconway:
Can I get an LGTM?
@vimalk78 @jcantrill

vimalk78 (approved):
/lgtm

PR #700, opened 2021-03-18
Add initial OEP for lso cleanup

gnufied:
cc @openshift/sig-storage

### DIFF
--- a/enhancements/storage/cleanup-lso-symlink.md
+++ b/enhancements/storage/cleanup-lso-symlink.md
@@ -41,6 +41,9 @@ This design fixes that problem and ensures that both PVs and symlinks are remove
 * Do not allow deletion of `LocalVolumeSet` object if it has bound PVCs.
 * Remove unbound PVs and remove symlinks on the host if `LocalVolume` or `LocalVolumeSet` objects are removed.
 
+Note: We already do not permit deletion of `LocalVolume` objects via finalizer if it has bound PVCs. This was difficult to implement correctly for `LocalVolumeSet` objects because local-static-provisoner was being used as a daemon and tracking of individual PVs was not possible. This is discussed in more detail in - https://github.com/openshift/local-storage-operator/pull/150#discussion_r492828334
+
+
 ### Non-Goals
 
 * There is another use-case that this design does not solve is - editing of `LocalVolume` and `LocalVolumeSet` objects. If user edits LSO CRs in such a manner that - a node which was previously selected is no longer included via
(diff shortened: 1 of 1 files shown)


enhancements/storage/cleanup-lso-symlink.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Motivation
+
+Currently when a user deletes `LocalVolume` or `LocalVolumeSet` objects - we do not automatically delete the PVs created from those objects and neither do we remove the symlinks created on the node. This has been documented as a limitation because removal of symlinks and PVs can be sometimes race-prone.
+
+This design fixes that problem and ensures that both PVs and symlinks are removed when `LocalVolume` and `LocalVolumeSet` objects are removed.
+
+### Goals
+
+* Do not allow deletion of `LocalVolumeSet` object if it has bound PVCs.
```
jsafrane, line 41: Seeing it written like this: why do we allow deletion of LocalVolumes then? Please add a note that LocalVolumes can't be deleted already.

enhancements/storage/cleanup-lso-symlink.md
```diff
@@ -0,0 +1,66 @@
...
+
+* There is another use-case that this design does not solve is - editing of `LocalVolume` and `LocalVolumeSet` objects. If user edits LSO CRs in such a manner that - a node which was previously selected is no longer included via
+node-selector or if CR excludes a device which was previously used for local-storage then - created PV is not automatically removed and corresponding symlink on the node is not removed either.
+
+## Proposal
+
+1. We propose that when `LocalVolume` or `LocalVolumeSet` object is created - they both contain a finalizer (`LocalVolume` object already has a finalizer). In addition we propose that - when user deletes these objects then finalizer can only be removed:
+   a. If there are no bound PVs that are created by this LSO CR.
+   b. None of the existing PVs(that match LSO CR) should have reclaim policy of `Retain.`
+   c. If there are unbound PVs with reclaimPolicy of `Delete` then finalizer should not be removed until all such PVs are deleted.
```
jsafrane, line 54: I admit that the conditions here (and below) are hard to read and put together.
In a. it allows deletion when there are unbound PVs and in b. and c. it does not allow to delete those.

In addition, why should ReclaimPolicy matter? Unound PV was not used, so there is no data to retain on them.

Perhaps it would be better to use PV phase:

* LSO removes the finalizer on LV/LVS only when all PVs associated with it are deleted.
* LSO automatically deletes Pending and Available PVS for LV/LVS with a deletion timestamp, as we expect there is no data on them. See below how LSO deletes them avoiding potential races with PV controller.
* Released PVs with `Delete` policy are deleted according to the policy.
* Bound PVs or Released PVs with `Retain` policy actively block LSO from removing the finalizer. An event is sent to tell users which PV blocks the deletion.
  * For Bound PVs, user should delete the corresponding PVC to get the PV Released.
  * For Released PVs with reclaim policy "Retain", user should back up any important data from the volume and delete the PV.

enhancements/storage/cleanup-lso-symlink.md
```diff
@@ -0,0 +1,66 @@
...
+
+2. If `LocalVolume` or `LocalVolumeSet` object can not be deleted because of #1.a or #1.b - then appropriate event should be logged and user should be informed. LSO is not going to perform any cleanup of bound PVs or PVs with `Retain` reclaim policy.
+3. Before we can remove finalizer from LSO managed CRs, we should ensure deletion of created unbound PVs. To do that:
```
jsafrane, line 57: This should be probably more separated from the points above, maybe a sub-chapter / paragraph how to safely remove a PV without races?

enhancements/storage/cleanup-lso-symlink.md
```diff
@@ -0,0 +1,66 @@
...
+   b. After making sure all unbound PVs for a deleted LSO CR are released, control-plane of LSO will wait for all such PVs to be deleted before removing the finalizer.
+4. On the node if a `LocalVolume` or `LocalVolumeSet` objects is being deleted (i.e has `deletionTimestamp`) then - control-loop running on the node will stop "resyncing" the PV objects on the node, so as control-plane can mark them as released.
+5. On the node if a `LocalVolume` or `LocalVolumeSet` object is being deleted and its PVs are in `Released` status then - node side control-loop should scrub the device, remove the symlink on the node and then delete the PV.
+6. Once LSO's control-plane detects that all PVs created by LSO CR is deleted, it would remove the finalizer and related objects.
+
+## Drawbacks
+
+One drawback is - if user wants to keep one or more PVs around (in case they have data on it) and still delete the `LocalVolume` or `LocalVolumeSet` object - this design will not allow such flow. Also this change would result in data being wiped off from previously created PVs, which could surprise some users.
```
jsafrane, line 66: User can use `Retain` policy on PV and force-remove the `LocalVolume`.

enhancements/storage/cleanup-lso-symlink.md
gnufied: done.

enhancements/storage/cleanup-lso-symlink.md
gnufied: I reworded this section.

enhancements/storage/cleanup-lso-symlink.md
gnufied: divided this into sections.

enhancements/storage/cleanup-lso-symlink.md
gnufied: added

PR #701, opened 2021-03-18
metrics scraping with local authentication and authorization

deads2k:
Monitoring needs to be reliable and is the very useful when trying to debug clusters in an already degraded state.  We want to ensure that metrics scraping can always work if the scraper can reach the target, even if the kube-apiserver is unavailable or unreachable. To do this, we will combine a local authorizer (already merged in many binaries and the rbac-proxy) and client-cert based authentication to have a fully local authentication and authorization path for scraper targets.

If networking (or part of networking) is down and a scraper target cannot reach the kube-apiserver to verify a token and a subjectaccessreview, then the metrics scraper can be rejected. The subjectaccessreview (authorization) is already largely addressed, but service account tokens are still used for scraping targets. Tokens require an external network call that we can avoid by using client certificates. Gathering metrics, especially client metrics, from partially functionally clusters helps narrow the search area between kube-apiserver, etcd, kubelet, and SDN considerably.

In addition, this will significantly reduce the load on the kube-apiserver. We have observed in the CI cluster that token and subject access reviews are a significant percentage of all kube-apiserver traffic. 

/assign @s-urbaniak 
/assign @sym3tri

### DIFF
--- a/enhancements/monitoring/client-cert-scraping.md
+++ b/enhancements/monitoring/client-cert-scraping.md
@@ -1,0 +1,132 @@
+---
+title: client-cert-base-metric-scraping
+authors:
+  - "@deads2k"
+reviewers:
+  - "@sur"
+approvers:
+  - "@sur"
+creation-date: 2021-03-18
+last-updated: 2021-03-18
+status: implementable
+see-also:
+replaces:
+superseded-by:
+---
+
+# Client Cert Base Metrics Scraping
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Monitoring needs to be reliable and is the very useful when trying to debug clusters in an already degraded state.
+We want to ensure that metrics scraping can always work if the scraper can reach the target, even if the kube-apiserver
+is unavailable or unreachable.
+To do this, we will combine a local authorizer (already merged in many binaries and the rbac-proxy) and client-cert based
+authentication to have a fully local authentication and authorization path for scraper targets.
+
+## Motivation
+
+If networking (or part of networking) is down and a scraper target cannot reach the kube-apiserver to verify a token
+and a subjectaccessreview, then the metrics scraper can be rejected.
+The subjectaccessreview (authorization) is already largely addressed, but service account tokens are still used
+for scraping targets.
+Tokens require an external network call that we can avoid by using client certificates.
+Gathering metrics, especially client metrics, from partially functionally clusters helps narrow the search area between
+kube-apiserver, etcd, kubelet, and SDN considerably.
+
+In addition, this will significantly reduce the load on the kube-apiserver.
+We have observed in the CI cluster that token and subject access reviews are a significant percentage of all kube-apiserver traffic.
+
+### Goals
+
+1. Gather all possible metrics even when targets cannot reach the kube-apiserver.
+2. Avoid sending valid kube-apiserver credentials to everyone who can create a servicemonitor.
+
+### Non-Goals
+
+1. Force all servicemonitor targets to change at the same time.
+
+## Proposal
+
+1. The monitoriting operator creates a client cert/key pair valid against the kube-apiserver.
+   There is a CSR signer name specifically for this.
+2. The kube-controller-manager operator runs an approver that allows monitoring operator SA to get signed certificate for
+   The existing scraper identity.
+3. The monitoring operator puts the client cert/key pair into a secret.
+4. The monitoring operator keeps these credentials up to date (the signer is only good for 30 days or so).
+5. The metrics scraper mounts the client cert/key pair and ensures that it is hot-reloadable or builds a process suicider.
+   There is library-go code to suicide on file changes.
+6. A participating metrics scraper target, can read the in-cluster client cert CA bundle and create an authenticator
+   using it.
+   The default delegated authentication stack does this and on openshift, the configmap is exposed to system:authenticated.
+7. A servicemonitor grows a way to specify using this client cert.
+   Technically this is optional.
+   If you find it too difficult, just *always* use the client-cert for scraping since the secrets never leave disk.
+   Those targets which don't support it will simply ignore the client-cert.
+8. Profit.
+   Now we can collect metrics for at least
+   1. openshift-apiserver
+   2. openshift-controller-manager
+   3. kube-scheduler
+   4. all kubelets
+   5. node-exporter
+   6. all control plane operators
+   7. kube-controller-manager
+   8. etcd
+   These are the ones I've updated already and we'll be able to get more.
+
+### User Stories
+
+#### Story 1
+
+I want developers to quickly solve my problems.
+
+View the metrics that will now be collected.
+
+#### Story 2
+
+### Implementation Details/Notes/Constraints [optional]
+
+### Risks and Mitigations
+
+## Design Details
+
+### Open Questions [optional]
+
+### Test Plan
+
+### Upgrade / Downgrade Strategy
+
+The CSR API has been at v1 and able to support our needs for multiple releases.
+The servicemonitors don't actually have to change in 4.9, they could change in 4.10 and we could still use this new capability in 4.9.
+
+### Version Skew Strategy
+
+Since service monitor's don't have to change, we don't need risk destabilizing changes.
+
+## Drawbacks
+
+I have to stop here to make dinner.
+
+## Alternatives
+
+Similar to the `Drawbacks` section the `Alternatives` section is used to
+highlight and record other possible approaches to delivering the value proposed
+by an enhancement.
+
+## Infrastructure Needed [optional]
+
+Use this section if you need things from the project. Examples include a new
+subproject, repos requested, github details, and/or testing infrastructure.
+
+Listing these here allows the community to get the process for these resources
+started right away.


deads2k:
In just the past week we have seen significant difficulties that would alleviated by this

1. no metrics when the openstack service load balancer for kubernetes.default.svc is broken makes correlating the connection difficulties hard/impossible
2. network checkers have no way to report failures during network outages because they cannot validate the metrics scraper to give results and cannot provide results to the kube-apiserver
3. a significant portion of all kube-apiserver traffic is for token and subject accessreviews, making us more vulnerable to overload and rejection of requests

Given the relatively small scope and risk since, we should prioritize this highly for our coming releases.  Monitoring is extremely valuable when things are not behaving, but certain classes of failures have nearly complete monitoring blackouts at this time.

enhancements/monitoring/client-cert-scraping.md
```diff
@@ -0,0 +1,132 @@
...
+The CSR API has been at v1 and able to support our needs for multiple releases.
+The servicemonitors don't actually have to change in 4.9, they could change in 4.10 and we could still use this new capability in 4.9. 
+
+### Version Skew Strategy
+
+Since service monitor's don't have to change, we don't need risk destabilizing changes.
+
+## Drawbacks
+
+I have to stop here to make dinner.
```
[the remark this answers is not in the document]
sttts: 🍕

enhancements/monitoring/client-cert-scraping.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Non-Goals
+
+1. Force all servicemonitor targets to change at the same time.
+
+## Proposal
+
+1. The monitoriting operator creates a client cert/key pair valid against the kube-apiserver.
+   There is a CSR signer name specifically for this.
+2. The kube-controller-manager operator runs an approver that allows monitoring operator SA to get signed certificate for
```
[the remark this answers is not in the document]
deads2k: > Is the approver decision going to be based solely on exact username match (and maybe groups match?) or is there going to be any additional mechanism (like RBAC or something custom)?

I think exact user match makes sense.  The service-account for the monitoring operator can create a client cert for the metrics scraper.  I think that permission is reasonable and happen to know exactly what identity that is.

enhancements/monitoring/client-cert-scraping.md
```diff
@@ -0,0 +1,132 @@
...
+4. The monitoring operator keeps these credentials up to date (the signer is only good for 30 days or so).
+5. The metrics scraper mounts the client cert/key pair and ensures that it is hot-reloadable or builds a process suicider.
+   There is library-go code to suicide on file changes.
+6. A participating metrics scraper target, can read the in-cluster client cert CA bundle and create an authenticator
+   using it.
+   The default delegated authentication stack does this and on openshift, the configmap is exposed to system:authenticated.
+7. A servicemonitor grows a way to specify using this client cert.
+   Technically this is optional.
+   If you find it too difficult, just *always* use the client-cert for scraping since the secrets never leave disk.
+   Those targets which don't support it will simply ignore the client-cert.
```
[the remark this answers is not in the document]
deads2k: > note: it's a shame the less-secure way is the default one, that means you have to opt-in your service to not receiving to the scraper SA's credentials. I wish we could simply default to the cert.

An e2e test trying to find offenders could be used to close gaps.  But I don't see that as essential to building out the initial capability.

enhancements/monitoring/client-cert-scraping.md
```diff
@@ -0,0 +1,132 @@
...
+1. Gather all possible metrics even when targets cannot reach the kube-apiserver.
+2. Avoid sending valid kube-apiserver credentials to everyone who can create a servicemonitor.
+
+### Non-Goals
+
+1. Force all servicemonitor targets to change at the same time.
+
+## Proposal
+
+1. The monitoriting operator creates a client cert/key pair valid against the kube-apiserver.
```
s-urbaniak, line 60: s/monitoriting/monitoring

enhancements/monitoring/client-cert-scraping.md
```diff
@@ -0,0 +1,132 @@
...
+To do this, we will combine a local authorizer (already merged in many binaries and the rbac-proxy) and client-cert based
+authentication to have a fully local authentication and authorization path for scraper targets.
+
+## Motivation
+
+If networking (or part of networking) is down and a scraper target cannot reach the kube-apiserver to verify a token
+and a subjectaccessreview, then the metrics scraper can be rejected.
+The subjectaccessreview (authorization) is already largely addressed, but service account tokens are still used
+for scraping targets.
+Tokens require an external network call that we can avoid by using client certificates.
```
s-urbaniak, line 42: question: as tokens are signed, wouldn't a local signature check suffer?

s-urbaniak: as an after thought, potentially a signature-only check in addition to scoped tokens would potentially suffice?
At least the advantage would be that client-side no additional cert management would be necessary.

enhancements/monitoring/client-cert-scraping.md
```diff
@@ -0,0 +1,132 @@
...
+2. Avoid sending valid kube-apiserver credentials to everyone who can create a servicemonitor.
+
+### Non-Goals
+
+1. Force all servicemonitor targets to change at the same time.
+
+## Proposal
+
+1. The monitoriting operator creates a client cert/key pair valid against the kube-apiserver.
+   There is a CSR signer name specifically for this.
```
s-urbaniak, line 61: which one is it specifically? is there an existing one or do we have to implement a new?

enhancements/monitoring/client-cert-scraping.md
```diff
@@ -0,0 +1,132 @@
...
+2. The kube-controller-manager operator runs an approver that allows monitoring operator SA to get signed certificate for
+   The existing scraper identity.
+3. The monitoring operator puts the client cert/key pair into a secret.
+4. The monitoring operator keeps these credentials up to date (the signer is only good for 30 days or so).
+5. The metrics scraper mounts the client cert/key pair and ensures that it is hot-reloadable or builds a process suicider.
```
simonpasquier, line 66: Prometheus will automatically reload the certificate/key files if they change.

enhancements/monitoring/client-cert-scraping.md
```diff
@@ -0,0 +1,132 @@
...
+   Now we can collect metrics for at least
+   1. openshift-apiserver
+   2. openshift-controller-manager
+   3. kube-scheduler
+   4. all kubelets
+   5. node-exporter
+   6. all control plane operators
+   7. kube-controller-manager
+   8. etcd
+   These are the ones I've updated already and we'll be able to get more.
```
simonpasquier, line 85: All monitoring components (except Prometheus, Alertmanager and Thanos querier) also use kube-rbac-proxy to secure their metrics endpoint. The 3 others got scraped via oauth-proxy (because historical reasons I guess).

enhancements/monitoring/client-cert-scraping.md
```diff
@@ -0,0 +1,132 @@
...
+2. The kube-controller-manager operator runs an approver that allows monitoring operator SA to get signed certificate for
+   The existing scraper identity.
+3. The monitoring operator puts the client cert/key pair into a secret.
+4. The monitoring operator keeps these credentials up to date (the signer is only good for 30 days or so).
+5. The metrics scraper mounts the client cert/key pair and ensures that it is hot-reloadable or builds a process suicider.
+   There is library-go code to suicide on file changes.
+6. A participating metrics scraper target, can read the in-cluster client cert CA bundle and create an authenticator
+   using it.
+   The default delegated authentication stack does this and on openshift, the configmap is exposed to system:authenticated.
+7. A servicemonitor grows a way to specify using this client cert.
```
simonpasquier, line 71: Not sure what you mean here. The service monitor allows to specify a client certificate/key either by file path in the Prometheus container or reference to a secret's key.

https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#tlsconfig

enhancements/monitoring/client-cert-scraping.md
deads2k: > question: as tokens are signed, wouldn't a local signature check suffice?

No. In general tokens are not signed.  ServiceAccount tokens in particular are, *but* if you only check locally, then the serviceaccount tokens become irrevocable.  It is important to be able to revoke tokens.

enhancements/monitoring/client-cert-scraping.md
deads2k: > which one is it specifically? is there an existing one or do we have to implement a new?

`kubernetes.io/kube-apiserver-client` from https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/ .  Can update the doc.

enhancements/monitoring/client-cert-scraping.md
deads2k: > Not sure what you mean here. The service monitor allows to specify a client certificate/key either by file path in the Prometheus container or reference to a secret's key.

That appears to be a reference to a secret in the namespace containing the target.  If that were the case, then every target would have to have client cert credentials for scanning /metrics.  We want to avoid this level of permissions for every component.

Is is possible to specify a client-cert that exists in the metric scraping namespace?

simonpasquier: The requirement is that the secret exists in the same namespace as the service monitor. The other option today would be to use the `certFile` and `keyFile` files but it assumes that the target knows the location of these files in the Prometheus container. It's not worse than what we do now for bearer [tokens](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/07a71fcac595fd9dd9ab2a7a79e817df6cdf1675/manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml#L53) and [CA](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/07a71fcac595fd9dd9ab2a7a79e817df6cdf1675/manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml#L67) but it's deprecated and we'll probably remove it eventually.

simonpasquier: And I'm not saying that we can't evolve things upstream :)

deads2k: > The other option today would be to use the `certFile` and `keyFile` files but it assumes that the target knows the location of these files in the Prometheus container.

This option is more reasonable for the effort we're trying to complete here.  Having every namespace with the power to scrape all metrics is not desireable.

s-urbaniak: > The requirement is that the secret exists in the same namespace as the service monitor. 
@simonpasquier i think referencing the secret via reference rather than file would be desirable. Do I understand correctly you think about losening the restriction in prometheus-operator to be able to reference a secret from a different namespace?

s-urbaniak:
@deads2k i think we reached consensus generally :+1: do you mind to address the nits above so we can merge? I think we can add specifics as a follow-up?

enhancements/monitoring/client-cert-scraping.md
```diff
@@ -0,0 +1,132 @@
...
+
+1. The monitoriting operator creates a client cert/key pair valid against the kube-apiserver.
+   There is a CSR signer name specifically for this.
+2. The kube-controller-manager operator runs an approver that allows monitoring operator SA to get signed certificate for
+   The existing scraper identity.
+3. The monitoring operator puts the client cert/key pair into a secret.
+4. The monitoring operator keeps these credentials up to date (the signer is only good for 30 days or so).
+5. The metrics scraper mounts the client cert/key pair and ensures that it is hot-reloadable or builds a process suicider.
+   There is library-go code to suicide on file changes.
+6. A participating metrics scraper target, can read the in-cluster client cert CA bundle and create an authenticator
```
s-urbaniak, line 68: is it expected for the participating metrics scraper target to be able to hot-reload the client cert CA as well? (asking, as kube-rbac-proxy doesn't right now)

s-urbaniak:
@deads2k how about we wrap this enhancement up? i guess we reached consensus generally :tada:

enhancements/monitoring/client-cert-scraping.md
deads2k: > > The requirement is that the secret exists in the same namespace as the service monitor.
> 
> @simonpasquier i think referencing the secret via reference rather than file would be desirable. Do I understand correctly you think about losening the restriction in prometheus-operator to be able to reference a secret from a different namespace?

I think an ideal way would be to say, "send me a client cert that identifies you as the scraper" without being able to specify exactly what.  The exact mechanism can be niggled later.  For the purposes of OCP, we can even hardcode it as a default since key material never leaves the scraper pod.

enhancements/monitoring/client-cert-scraping.md
s-urbaniak: i was about to ask if there's things to reuse, thanks!

enhancements/monitoring/client-cert-scraping.md
s-urbaniak, line 62: We concretely ended up to be in the KCM https://github.com/openshift/cluster-policy-controller sidecar.

deads2k:
It's hard to overstate the degree to which I do not care whether markdown lints cleanly or not :(

deads2k:
> It's hard to overstate the degree to which I do not care whether markdown lints cleanly or not :(

wow, and to run it locally I need node-js.  Seriously, who cares

PR #703, opened 2021-03-19
management workload partitioning

dhellmann:
This enhancement presents an alternative approach to replace #628.
It describes an approach to allow us to isolate the control plane services
to run on a restricted set of CPUs. This will be especially useful
for resource-constrained enviornments, such as single-node production
deployments, where the user wants to reserve most of the CPU resources
for their own workloads and needs to configure OpenShift to run on a
fixed number of CPUs within the host.

### DIFF
--- a/enhancements/management-workload-partitioning.md
+++ b/enhancements/management-workload-partitioning.md
@@ -74,7 +74,7 @@ management and normal workloads there.
   know what constraints we might be given in the future so we need a
   solution that is not tied to the current known limit.
 * We want a general approach, that can be applied to all OpenShift
-  control plane components.
+  control plane and per-node components.
 
 ### Non-Goals
 
@@ -89,13 +89,14 @@ management and normal workloads there.
   single-node developer deployments, multi-node production
   deployments, and Internet-of-things devices, those use cases are not
   addressed directly in this document.
-* This enhancement does not address reducing actual utilization. There
-  is no expectation that a cluster configured to use a small number of
+* This enhancement does not address reducing actual utilization,
+  beyond providing a way to have a predictable upper-bounds. There is
+  no expectation that a cluster configured to use a small number of
   cores for management services would offer exactly the same
   performance as the default. It must be stable and continue to
   operate reliably, but may respond more slowly.
-* This enhancement assumes that the configuration of management
-  services is done as part of installing the cluster, and cannot be
+* This enhancement assumes that the configuration of a management CPU
+  pool is done as part of installing the cluster, and cannot be
   changed later. Future enhancements may address enabling or
   reconfiguring the feature described here on an existing cluster.
 * This enhancement describes partitioning concepts that could be
@@ -127,9 +128,16 @@ customer or by vendors who are not our partners.
 The basic proposal is to provide a way to identify management
 workloads at runtime and to use CRI-O to run them on a user-selected
 set of CPUs, while other workloads will be prevented from running
-there. This effectively gives 3 pools of CPUs (shared, dedicated, and
-management) and means the shared CPU pool will need other CPUs to
-support burstable or best-effort workloads.
+there. This effectively gives 3 pools of CPUs
+
+* **management** -- a user-defined cpuset that restricts where
+  management components run
+* **dedicated** -- CPUs used for exclusive workloads
+* **shared** -- CPUs in the set `management + !dedicate`
+
+Because the management CPU pool is restricted to management workloads,
+the shared CPU pool will need other CPUs to support burstable or
+best-effort workloads that are not identified as management workloads.
 
 We want the isolation to be in place from the initial buildout of the
 cluster, to ensure predictable behavior. Therefore, the feature must
@@ -140,47 +148,55 @@ disabled.
 
 We generally want components to opt-in to being considered management
 workloads. Therefore, for a regular pod to be considered to contain a
-management workload it must be labeled with `io.openshift.management:
-true`.
+management workload it must be labeled with a *workload label*,
+`workload.openshift.io/{pool_name}`. For now, we will focus on
+`workload.openshift.io/management`, but this syntax supports other
+types of workloads that may be defined in future enhancements.
 
 We want to treat all OpenShift components as management workloads,
 including those that run the control plane. Therefore, kubelet will be
-modified to treat all static pods as management workloads, when the
-feature is enabled. We will update operator manifests and
+modified to treat static pods with the label as management workloads,
+when the feature is enabled. We will update operator manifests and
 implementations to label all OpenShift components not running in
-static pods with the `io.openshift.management: true` label, and add a
-CI job to require that label be present in all workloads created from
-the release payload.
+static pods with the label, and add a CI job to require that the label
+be present in all workloads created from the release payload.
 
 We need kubelet to know when the feature is enabled, but we cannot
-change the configuration schema that kubelet gets from
+change the configuration schema that kubelet inherits from
 upstream. Therefore we will have kubelet look for a new configuration
 file on startup, and the feature will only be enabled if that file is
 found.
 
 We want to give cluster administrators control over which workloads
 are run on the management CPUs. Therefore, only pods in namespaces
-labeled with `io.openshift.management: true` will be subject to
-special handling. Normal users cannot add a label to a namespace
-without the right RBAC permissions.
+labeled with a workload label will be subject to special
+handling. Normal users cannot add a label to a namespace without the
+right RBAC permissions.
 
 We want to continue to use the scheduler for placing management
 workloads, but we cannot rely on the CPU requests of those workloads
 to accurately reflect the constrained environment in which they are
 expected to run. Instead of scaling those CPU request values, we will
 change them to request a new [extended
-resource](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/),
-"management cores." We will modify kubelet to advertise management
-cores as an extended resource when the management workload
+resource](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/)
+called `cpu.workload.openshift.io/{pool_name}`
+(`cpu.workload.openshift.io/management`). We will modify kubelet to
+advertise the same new extended resource when the management workload
 partitioning feature is enabled, using a value equivalent to the CPU
 resources for the entire host. This large value should allow the
 scheduler to always be able to place workloads, while still accurately
-accounting for those requests.
+accounting for those requests. The naming convention of the resource
+will allow us to support other CPU pools in future enhancements.
 
 We need to ensure fair allocation of CPU time between different
-management workloads. Therefore, we will copy the original CPU
-requests for management workload pods into an annotation that CRI-O
-can use to configure the CPU shares when running the containers.
+management workloads. CRI-O does not receive information about
+extended resource reservations when a container is started, but it
+does receive the content of annotations on the pod. Therefore, we will
+copy the original CPU requests for containers in management workload
+pods into a annotations that CRI-O can use to configure the CPU shares
+when running the containers. We will use the naming convention
+`cpu.workload.openshift.io/{container-name}` and create an annotation
+for each container in the pod.
 
 We need to change pod definitions as each pod is created, so that the
 scheduler, kubelet, and CRI-O all see a consistently updated version
@@ -191,20 +207,23 @@ introduced with traditional admission webhooks or controllers.
 Therefore, we will build an admission hook into the [kubernetes API
 server in
 OpenShift](https://github.com/openshift/kubernetes/tree/master/openshift-kube-apiserver/admission),
-to intercept API requests to create pods.
+to intercept API requests to create pods. We will update kubelet to
+make the same changes to static pods.
 
 The API server requires a configuration resource as input, rather than
 a ConfigMap or command line flag.  Therefore, we need an API-driven
 way to enable management workload partitioning in the admission
-hook. We will extend the Infrastructure configuration resource with a
-new status field to indicate whether the feature is on or off.
+hook. We will add a [feature
+set](https://github.com/openshift/api/blob/master/config/v1/types_feature.go#L102-L119)
+with a feature gate to control whether the feature is on or off.
 
-*TBD: API field name, see open questions*
+*We probably need more details about the feature gate.*
 
-Some pods used to run OpenShift control plane components are started
-before the API server. Therefore, we will have to manually add extra
-metadata to those pod definitions, instead of relying on the admission
-hook to do it.
+Some pods used to run OpenShift control plane components
+(kubeapiserver, kube-controller-manager, scheduler, and etcd) are
+started before the API server. Therefore, we will have to manually add
+extra metadata to those pod definitions, instead of relying on the
+admission hook to do it.
 
 Some of the workloads that will be classified as management are
 outside of the OpenShift release payload. Therefore we will need to
@@ -218,19 +237,22 @@ labels ourselves, starting with
 * ptp-operator
 * local-storage-operator
 * cluster-logging (fluentd)
-* klusterlet (RHACM)
+* klusterlet and klusterlet add-ons (RHACM)
+* kubevirt
+* N3000
 
 ### User Stories
 
 #### Radio Access Network (RAN) Use Case
 
-In the context of telcocommunications service providers' 5G Radio Access
-Networks, it is increasingly common to see "cloud native" implementations
-of the 5G Distributed Unit (DU) component. Due to latency constraints,
-this DU component needs to be deployed very close to the radio antenna for
-which it is responsible. In practice, this can mean running this
-component on anything from a single server at the base of a remote cell
-tower or in a datacenter-like environment serving several base stations.
+In the context of telcocommunications service providers' 5G Radio
+Access Networks, it is increasingly common to see "cloud native"
+implementations of the 5G Distributed Unit (DU) component. Due to
+latency constraints, this DU component needs to be deployed very close
+to the radio antenna for which it is responsible. In practice, this
+can mean running this component on anything from a single server at
+the base of a remote cell tower or in a datacenter-like environment
+serving several base stations.
 
 A hypothetical DU example is an unusually resource-intensive workload,
 requiring 20 dedicated cores, 24 GiB of RAM consumed as huge pages,
@@ -263,39 +285,74 @@ correctly scheduled to run on the management CPU pool.
 
 1. User sits down at their computer.
 2. The user creates their `install-config.yaml`, including extra
-   values in the `bootstrapInPlace` section (since that enables
-   deploying a single-node).
-   * `managementCPUIDs` is a CPU set specifier for the CPUs to add to
-     the isolated set used for management components. The default is
-     empty, and when the value is empty the management workload
-     partitioning feature is disabled. The rest of the steps below
-     assume it is enabled, unless otherwise stated.
+   values in the specifying the CPUs to include in the management CPU
+   set. (See below for details.)
+   * Does the user also need to explicitly enable the feature flag?
 3. The user runs the installer.
-4. The installer uses `managementCPUIDs` to generate an extra machine
-   config manifest to configure CRI-O to process management workloads
-   in a special way.
+4. The installer uses the management CPU pool settings to generate an
+   extra machine config manifest to configure CRI-O to process
+   management workloads in a special way.
 5. The installer creates a machine config manifest to write a
    configuration file for kubelet. The file should only be readable by
    the kubelet.
-6. The installer updates the status fields of the Infrastructure
-   config resource to enable the feature for the admission hook.
+6. The installer enables the feature set for the feature flag
+   controlling management workload partitioning.
 7. The kubelet starts up and finds the configuration file enabling the
    new feature.
-8. The kubelet reads static pod definitions. It replaces the `cpu`
-   requests with `management-cpu` requests of the same value and adds
-   an annotation for CRI-O with the same value.
-9. Something schedules a regular pod with the
-   `io.openshift.management: true` label in a namespace with the
-   `io.openshift.management: true` label.
-10. The admission hook modifies the pod, replacing the CPU requests
-    with management CPU requests and adding an annotation for CRI-O.
-11. The scheduler sees the new pod and finds available management CPU
-    resources on the node. The scheduler places the pod on the node.
-12. Repeat steps 9-11 until all pods are running.
-13. Single-node deployment comes up with management components
+8. The kubelet advertises `cpu.workload.openshift.io/management`
+   extended resources on the node based on the number of CPUs in the
+   host.
+9. The kubelet reads static pod definitions. It replaces the `cpus`
+   requests with `cpu.workload.openshift.io/management` requests of
+   the same value and adds the
+   `cpu.workload.openshift.io/{container-name}` annotations for CRI-O
+   with the same values.
+10. Something schedules a regular pod with the
+    `workload.openshift.io/management` label in a namespace with the
+    `workload.openshift.io/management` label.
+11. The admission hook modifies the pod, replacing the CPU requests
+    with `cpu.workload.openshift.io/management` requests and adding
+    the `cpu.workload.openshift.io/{container-name}` annotations for
+    CRI-O.
+12. The scheduler sees the new pod and finds available
+    `cpu.workload.openshift.io/management` resources on the node. The
+    scheduler places the pod on the node.
+13. Repeat steps 10-12 until all pods are running.
+14. Single-node deployment comes up with management components
     constrained to subset of available CPUs
 
+#### Pod mutation
+
+The kubelet and API admission hook will change pods labeled with
+`workload.openshift.io/management` so the CPU requests are replaced
+with management CPU requests and an annotation is added with the same
+value.
+
+```yaml
+requests:
+  cpu:
+    400m
+```
+
+becomes
+
+```yaml
+requests:
+  cpu.workload.openshift.io/management: 400m
+```
+
+and
+
+```yaml
+annotations:
+  cpu.workload.openshift.io/container-name: 400m
+```
+
+
 #### CRI-O Changes
+
+*NOTE: This section needs to be updated to deal with the generalized
+CPU pools.*
 
 CRI-O will be updated to support a new configuration value that a user
 can specify, `mgmt_ctr_cpuset`.
@@ -315,13 +372,12 @@ CRI-O will be configured to support a new annotation on pods,
 [crio.runtime.runtimes.runc]
   runtime_path = "/usr/bin/runc"
   runtime_type = "oci"
-  allowed_annotations = ["io.openshift.management.cores"]
+  allowed_annotations = ["cpu.workload.openshift.io"]
 ```
 
-Pods that have the `io.openshift.management.cores` annotation will
-have their cpuset configured to the value in `mgmt_ctr_cpuset`, as
-well as have their CPU shares configured to the value of the
-annotation.
+Pods that have the `cpu.workload.openshift.io` annotation will have
+their cpuset configured to the value in `mgmt_ctr_cpuset`, as well as
+have their CPU shares configured to the value of the annotation.
 
 Note that this field does not conflict with the `infra_ctr_cpuset`
 config option, as the infra container will still be put in that
@@ -333,13 +389,13 @@ container should also be considered to be managed.
 A new admission hook in the kubernetes API server within OpenShift
 will mutate pods when they are created to make 2 changes.
 
-1. It will move CPU requests to "management core" requests, so that
+1. It will move CPU requests to workload CPU requests, so that
    the scheduler can successfully place the pod on the node, even
    though the sum of the CPU requests for all management workloads may
    exceed the actual CPU capacity of the management CPU pool.
-2. It will add an annotation, `io.openshift.management.cores` with a
-   value equal to the original CPU requests, so that CRI-O can use the
-   value to configure the CPU shares for the container.
+2. It will add annotations with for each container with the value
+   equal to the original CPU requests, so that CRI-O can use the value
+   to configure the CPU shares for the container.
 
 We will not change pods in a way that changes their quality-of-service
 class. So, we would not strip CPU requests unless they also have
@@ -363,14 +419,34 @@ it is mutated in a way similar to the API server admission hook will
 mutate regular pods.
 
 Kubelet will be changed so that when the feature is enabled, it
-advertises a new extended resource of "management cores", representing
-all of the CPU capacity of the host (not just the management CPU
-pool).
+advertises a new extended resource of
+`cpu.workload.openshift.io/management`, representing all of the CPU
+capacity of the host (not just the management CPU pool).
 
 #### Installer Changes
 
-The installer will be changed to accept the new `managementCPUIDs`
-configuration input value.
+The installer will be changed to accept a new `workloadCPUPools`
+parameter with a list of CPU pools to treat as separate from the
+standard shared pool. (This name may change.)
+
+```yaml
+workloadCPUPools:
+  - name: management
+    cpuIDs: 0-1
+```
+
+The default is empty, and when the value is empty the management
+workload partitioning feature is disabled.
+
+For the first version, we will only support 1 pool with the name
+"management". That restriction may be lifted in the future.
+
+The `cpuIDs` value is a CPU set specifier for the CPUs to add to the
+isolated set used for management components, using the same cpuset
+syntax used elsewhere in kubernetes. Asking for explicit IDs instead
+of simply a count gives the user control in situations where they need
+specific CPUs to be available because they're close to accelerators,
+NICs, or other special hardware.
 
 The installer will be changed to generate an extra machine config
 manifest to configure CRI-O so that containers from pods with the
@@ -382,10 +458,8 @@ write the `/etc/kubernetes/management-pinning` configuration file for
 kubelet. The file will have SELinux settings configured so that it is
 only readable by the kubelet.
 
-The installer will be changed to update the
-`status.managementWorkloadPartitioning` field of the Infrastructure
-config resource to `enabled` so the admission hook will know when to
-mutate pods.
+The installer will be changed to support the feature flag for enabling
+management workload partitioning.
 
 ### Risks and Mitigations
 
@@ -413,8 +487,11 @@ to components that we try to keep in sync with upstream versions. This
 design isolates those patches more fully than the previous approaches
 did, which should make carrying the patches easier.  This design also
 provides an opportunity to upstream the feature in a way that other
-approaches did not, which may mean not having to carry the patches
-indefinitely.
+approaches did not, which may mean we can avoid carrying the patches
+indefinitely. Even if the approach is not accepted upstream exactly as
+it is described here, it will be informative to share the experience
+as we believe workload isolation is a common problem for other
+kubernetes users.
 
 ## Design Details
 
@@ -426,6 +503,23 @@ indefinitely.
    or something similar, and say for now that it only applies when the
    single-node deployment approach is used?
 2. What should the new field in the Infrastructure CRD be?
+3. Should we worry about heterogeneous control plane hardware in
+   multi-node clusters? The install-config schema may need to support
+   specifying the same pool as using different CPUs on different nodes
+   in that case. For example
+
+   ```yaml
+   workloadCPUPools:
+     - name: management
+       nodes:
+         - master-0
+       cpuIDs: 0-1
+     - name: management
+       nodes:
+         - master-1
+         - master-2
+       cpuIDs: 23-24
+   ```
 
 ### Test Plan
 
@@ -504,8 +598,13 @@ N/A
 
 ## Implementation History
 
-Major milestones in the life cycle of a proposal should be tracked in `Implementation
-History`.
+A CI job to identify pods that need a workload label and ensure it is
+present: https://github.com/openshift/origin/pull/25992
+
+Kubelet changes: https://github.com/openshift/kubernetes/pull/627
+
+The first set of CRI-O changes to refactor annotation handling:
+https://github.com/cri-o/cri-o/pull/4680
 
 ## Drawbacks
 
@@ -515,6 +614,24 @@ pool" feature may be acceptable upstream, and we could reimplement
 management workload partitioning to use that new implementation.
 
 ## Alternatives
+
+### Force pods into the management CPU pool without opt-in
+
+We could force all pods of a certain priority into the management CPU
+pool. There are two issues with that approach:
+
+1. There are no RBAC restrictions on selecting the priority for a
+   pod. We could mitigate that by still using the namespace label to
+   opt-in.
+2. By not requiring opt-in, we make it harder for component authors to
+   predict and understand how their code is run. The pods created from
+   deployments managed by operators would not match the deployment
+   settings, and it would not be clear why. By requiring opt-in,
+   component authors have a chance to learn about and understand the
+   feature.
+
+Opt-in also gives us the opportunity to expand to more special-purpose
+CPU pools in the future.
 
 ### Use PAO to configure this behavior day 2
 
--- a/enhancements/management-workload-partitioning.md
+++ b/enhancements/management-workload-partitioning.md
@@ -346,9 +346,12 @@ and
 
 ```yaml
 annotations:
-  cpu.workload.openshift.io/container-name: 400m
+  io.openshift.workload.management.cpu/container-name: 400m
 ```
 
+As a special case, kubelet and the API admission hook should strip any
+annotations from pods trying to set the "cpuset" resource. See the
+next section for more details.
 
 #### CRI-O Changes
 
@@ -359,24 +362,45 @@ CRI-O will be updated to support new configuration settings for workload types.
 
 ```ini
 [crio.runtime.workloads.{workload-type}]
-  cpu_set = "0-1"
   label = "workload.openshift.io/target={workload-type}"
+  resource_annotation_prefix = "io.openshift.workload.{workload-type}"
+  resources = {
+    "cpu": "",
+    "cpuset": "0-1",
+  }
 ```
 
-The `cpu_set` field describes the CPU set that workloads will be
-configured to use, based on their type.
+The `label` field is used to match pods that should be treated as
+having the workload type. The label key on the pod is compared for an
+exact match against the value specified in the configuration file.
 
-The `annotation` field gives the annotation prefix used to match pods
-to the workload type. Specifying this in the configuration file makes
-it easier to change later and keeps OpenShift-specific values out of
-CRI-O.
+The `resource_annotation_prefix` is the start of the annotation key
+used to pass settings down to CRI-O.
+
+The `resources` map associates annotation suffixes with default
+values. CRI-O will define a well-known set of resources and other
+values will not be allowed. In OpenShift, we do not want to allow
+arbitrary pods to set their own `cpuset` but other CRI-O users may
+want that ability.
+
+To pass a setting into CRI-O, the pod should have an annotation made
+by combining the `resource_annotation_prefix`, the key from
+`resources`, and the container name, like this:
+
+```
+io.openshift.workload.management.cpu/container_name = {value}
+```
 
 In the management workload case, we will configure it with values like
 
 ```ini
 [crio.runtime.workloads.management]
-  cpu_set = "0-1"
   label = "workload.openshift.io/target=mangement"
+  resource_annotation_prefix = "io.openshift.workload.management"
+  resources = {
+    "cpu": "",
+    "cpuset": "0-1",
+  }
 ```
 
 CRI-O will be configured to support a new annotation on pods,
--- a/enhancements/management-workload-partitioning.md
+++ b/enhancements/management-workload-partitioning.md
@@ -495,10 +495,10 @@ CRI-O will be configured to support a new annotation on pods,
   runtime_type = "oci"
 ```
 
-Pods that have the `io.openshift.workload.management.cpu` annotation
-will have their cpuset configured to the value from the appropriate
-workload configuration, as well as have their CPU shares configured to
-the value of the annotation.
+Pods that have the `io.openshift.workload.management.cpushares`
+annotation will have their cpuset configured to the value from the
+appropriate workload configuration, as well as have their CPU shares
+configured to the value of the annotation.
 
 Note that this field does not conflict with the `infra_ctr_cpuset`
 config option, as the infra container will still be put in that
--- a/enhancements/management-workload-partitioning.md
+++ b/enhancements/management-workload-partitioning.md
@@ -149,8 +149,8 @@ disabled.
 We generally want components to opt-in to being considered management
 workloads. Therefore, for a regular pod to be considered to contain a
 management workload it must be labeled with a *workload label*,
-`workload.openshift.io/{pool_name}`. For now, we will focus on
-`workload.openshift.io/management`, but this syntax supports other
+`workload.openshift.io/target={workload_type}`. For now, we will focus on
+`workload.openshift.io/target=management`, but this syntax supports other
 types of workloads that may be defined in future enhancements.
 
 We want to treat all OpenShift components as management workloads,
@@ -169,9 +169,10 @@ found.
 
 We want to give cluster administrators control over which workloads
 are run on the management CPUs. Therefore, only pods in namespaces
-labeled with a workload label will be subject to special
-handling. Normal users cannot add a label to a namespace without the
-right RBAC permissions.
+labeled with a label
+`workload.openshift.io/allowed={comma_separated_list_of_type_names}`
+will be subject to special handling. Normal users cannot add a label
+to a namespace without the right RBAC permissions.
 
 We want to continue to use the scheduler for placing management
 workloads, but we cannot rely on the CPU requests of those workloads
@@ -179,14 +180,15 @@ to accurately reflect the constrained environment in which they are
 expected to run. Instead of scaling those CPU request values, we will
 change them to request a new [extended
 resource](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/)
-called `cpu.workload.openshift.io/{pool_name}`
-(`cpu.workload.openshift.io/management`). We will modify kubelet to
-advertise the same new extended resource when the management workload
-partitioning feature is enabled, using a value equivalent to the CPU
-resources for the entire host. This large value should allow the
-scheduler to always be able to place workloads, while still accurately
-accounting for those requests. The naming convention of the resource
-will allow us to support other CPU pools in future enhancements.
+called `cpu.workload.openshift.io/{workload_type}`.  We will modify
+kubelet to advertise the `cpu.workload.openshift.io/management`
+extended resource when the management workload partitioning feature is
+enabled, using a value equivalent to the CPU resources for the entire
+host. This large value should allow the scheduler to always be able to
+place workloads, while still accurately accounting for those
+requests. The naming convention of the resource will allow us to
+support other CPU pools in future enhancements. We may change the
+formula for the advertised resources in future enhancements.
 
 We need to ensure fair allocation of CPU time between different
 management workloads. CRI-O does not receive information about
@@ -287,7 +289,6 @@ correctly scheduled to run on the management CPU pool.
 2. The user creates their `install-config.yaml`, including extra
    values in the specifying the CPUs to include in the management CPU
    set. (See below for details.)
-   * Does the user also need to explicitly enable the feature flag?
 3. The user runs the installer.
 4. The installer uses the management CPU pool settings to generate an
    extra machine config manifest to configure CRI-O to process
@@ -302,14 +303,14 @@ correctly scheduled to run on the management CPU pool.
 8. The kubelet advertises `cpu.workload.openshift.io/management`
    extended resources on the node based on the number of CPUs in the
    host.
-9. The kubelet reads static pod definitions. It replaces the `cpus`
+9. The kubelet reads static pod definitions. It replaces the `cpu`
    requests with `cpu.workload.openshift.io/management` requests of
    the same value and adds the
    `cpu.workload.openshift.io/{container-name}` annotations for CRI-O
    with the same values.
 10. Something schedules a regular pod with the
-    `workload.openshift.io/management` label in a namespace with the
-    `workload.openshift.io/management` label.
+    `workload.openshift.io/target=management` label in a namespace
+    with the `workload.openshift.io/allowed=management` label.
 11. The admission hook modifies the pod, replacing the CPU requests
     with `cpu.workload.openshift.io/management` requests and adding
     the `cpu.workload.openshift.io/{container-name}` annotations for
@@ -354,19 +355,32 @@ annotations:
 *NOTE: This section needs to be updated to deal with the generalized
 CPU pools.*
 
-CRI-O will be updated to support a new configuration value that a user
-can specify, `mgmt_ctr_cpuset`.
+CRI-O will be updated to support new configuration settings for workload types.
 
 ```ini
-[crio.runtime]
-mgmt_ctr_cpuset = “0-1”
+[crio.runtime.workloads.{workload-type}]
+  cpu_set = "0-1"
+  label = "workload.openshift.io/target={workload-type}"
 ```
 
-This field describes the CPU set that management workloads will be
-configured to use.
+The `cpu_set` field describes the CPU set that workloads will be
+configured to use, based on their type.
+
+The `annotation` field gives the annotation prefix used to match pods
+to the workload type. Specifying this in the configuration file makes
+it easier to change later and keeps OpenShift-specific values out of
+CRI-O.
+
+In the management workload case, we will configure it with values like
+
+```ini
+[crio.runtime.workloads.management]
+  cpu_set = "0-1"
+  label = "workload.openshift.io/target=mangement"
+```
 
 CRI-O will be configured to support a new annotation on pods,
-`io.openshift.management.cores`.
+`cpu.workload.openshift.io/{container-name}`.
 
 ```ini
 [crio.runtime.runtimes.runc]
@@ -376,8 +390,9 @@ CRI-O will be configured to support a new annotation on pods,
 ```
 
 Pods that have the `cpu.workload.openshift.io` annotation will have
-their cpuset configured to the value in `mgmt_ctr_cpuset`, as well as
-have their CPU shares configured to the value of the annotation.
+their cpuset configured to the value from the appropriate workload
+configuration, as well as have their CPU shares configured to the
+value of the annotation.
 
 Note that this field does not conflict with the `infra_ctr_cpuset`
 config option, as the infra container will still be put in that
@@ -425,12 +440,14 @@ capacity of the host (not just the management CPU pool).
 
 #### Installer Changes
 
-The installer will be changed to accept a new `workloadCPUPools`
-parameter with a list of CPU pools to treat as separate from the
-standard shared pool. (This name may change.)
+The installer will be changed to accept a new `workloadSettings`
+parameter with a list of workload types and their settings. For now,
+the only setting will be the name and the CPU sets to treat as
+separate from the standard shared pool. The name `workloadSettings`
+may change, based on discussion on the installer code changes.
 
 ```yaml
-workloadCPUPools:
+workloadSettings:
   - name: management
     cpuIDs: 0-1
 ```
@@ -438,8 +455,8 @@ workloadCPUPools:
 The default is empty, and when the value is empty the management
 workload partitioning feature is disabled.
 
-For the first version, we will only support 1 pool with the name
-"management". That restriction may be lifted in the future.
+For the first version, we will only support 1 workload type with the
+name "management". That restriction may be lifted in the future.
 
 The `cpuIDs` value is a CPU set specifier for the CPUs to add to the
 isolated set used for management components, using the same cpuset
@@ -450,8 +467,8 @@ NICs, or other special hardware.
 
 The installer will be changed to generate an extra machine config
 manifest to configure CRI-O so that containers from pods with the
-`io.openshift.management.cores` annotation are run on the
-`mgmt_ctr_cpuset`.
+`cpu.workload.openshift.io/management` annotation are run on the
+CPU set specified by the management workload settings.
 
 The installer will be changed to create a machine config manifest to
 write the `/etc/kubernetes/management-pinning` configuration file for
@@ -520,6 +537,9 @@ kubernetes users.
          - master-2
        cpuIDs: 23-24
    ```
+4. Is the feature flag enabled automatically when the user specifies
+   the workload settings in the install-config, or do they need to
+   enable it explicitly.
 
 ### Test Plan
 
--- a/enhancements/management-workload-partitioning.md
+++ b/enhancements/management-workload-partitioning.md
@@ -156,7 +156,7 @@ disabled.
 We generally want components to opt-in to workload partitioning and
 especially to being considered management workloads. Therefore, for a
 regular pod to be considered to contain a management workload it must
-be labeled with an annotation configuring its *workload type*,
+have an annotation configuring its *workload type*,
 `workload.openshift.io/{workload-type}`. For now, we will focus on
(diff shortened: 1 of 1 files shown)


enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+  clusters to run with management workloads on a restricted set of
+  CPUs.
+* Clusters built in this way should pass the same Kubernetes and
+  OpenShift conformance and functional end-to-end tests as single-node
+  deployments that are not isolating the management workloads.
+* We have a goal of running on 4 hyperthreads today, but we do not
+  know what constraints we might be given in the future so we need a
+  solution that is not tied to the current known limit.
+* We want a general approach, that can be applied to all OpenShift
+  control plane components.
```
smarterclayton, line 77: control plane AND per-node components

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+  deployments, and Internet-of-things devices, those use cases are not
+  addressed directly in this document.
+* This enhancement does not address reducing actual utilization. There
+  is no expectation that a cluster configured to use a small number of
+  cores for management services would offer exactly the same
+  performance as the default. It must be stable and continue to
+  operate reliably, but may respond more slowly.
+* This enhancement assumes that the configuration of management
+  services is done as part of installing the cluster, and cannot be
+  changed later. Future enhancements may address enabling or
```
smarterclayton, line 99: This confuses me, but I'll comment on the details.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+We want the isolation to be in place from the initial buildout of the
+cluster, to ensure predictable behavior. Therefore, the feature must
+be enabled during installation. To enable the feature, the user will
+specify the set of CPUs to run all management workloads. When the
+management CPU set is not defined, the feature will be completely
+disabled.
+
+We generally want components to opt-in to being considered management
+workloads. Therefore, for a regular pod to be considered to contain a
+management workload it must be labeled with `io.openshift.management:
```
smarterclayton, line 143: The label should probably be `workload.openshift.io/management = ""` which means "This is considered part of the management workload of the cluster".

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+true`.
+
+We want to treat all OpenShift components as management workloads,
+including those that run the control plane. Therefore, kubelet will be
+modified to treat all static pods as management workloads, when the
```
smarterclayton, line 148: Why?  Needs some justification vs just having the operators also write the label into static pods.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+feature is enabled. We will update operator manifests and
+implementations to label all OpenShift components not running in
+static pods with the `io.openshift.management: true` label, and add a
+CI job to require that label be present in all workloads created from
+the release payload.
+
+We need kubelet to know when the feature is enabled, but we cannot
+change the configuration schema that kubelet gets from
+upstream. Therefore we will have kubelet look for a new configuration
+file on startup, and the feature will only be enabled if that file is
```
smarterclayton, line 158: Can't we just make this part of how we name cpu pools?  I.e. if you have a pool with a specific name, we enable this?

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+found.
+
+We want to give cluster administrators control over which workloads
+are run on the management CPUs. Therefore, only pods in namespaces
+labeled with `io.openshift.management: true` will be subject to
```
smarterclayton, line 163: Namespaces aren't sufficient.  We will have namespaces that have mixed management/not-management workloads and it's not appropriate for us to be forced to split those into different namespaces.

The RBAC behavior can be addressed simply by requiring high/admin privilege to set this on a pod, possibly by the same component that does the cpu request -> workload transition.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+We need to change pod definitions as each pod is created, so that the
+scheduler, kubelet, and CRI-O all see a consistently updated version
+of the pod and do not need to make independent decisions about whether
+it should be treated as a management workload. We need to intercept
+pod creation for *all* pods, without race conditions that might be
+introduced with traditional admission webhooks or controllers.
+Therefore, we will build an admission hook into the [kubernetes API
+server in
+OpenShift](https://github.com/openshift/kubernetes/tree/master/openshift-kube-apiserver/admission),
+to intercept API requests to create pods.
```
smarterclayton, line 194: This admission plugin can be the one that requires you have clutser admin to set this label (or some suitably high permission)

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+cores as an extended resource when the management workload
+partitioning feature is enabled, using a value equivalent to the CPU
+resources for the entire host. This large value should allow the
+scheduler to always be able to place workloads, while still accurately
+accounting for those requests.
+
+We need to ensure fair allocation of CPU time between different
+management workloads. Therefore, we will copy the original CPU
+requests for management workload pods into an annotation that CRI-O
+can use to configure the CPU shares when running the containers.
```
smarterclayton, line 183: Why would we need an annotation?  I would expect converting:

```
requests:
  cpu: 400m
```

to

```
requests:
  management.openshift.io/cpu: 400m
```

to be sufficient to convey that

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+Therefore, we will build an admission hook into the [kubernetes API
+server in
+OpenShift](https://github.com/openshift/kubernetes/tree/master/openshift-kube-apiserver/admission),
+to intercept API requests to create pods.
+
+The API server requires a configuration resource as input, rather than
+a ConfigMap or command line flag.  Therefore, we need an API-driven
+way to enable management workload partitioning in the admission
+hook. We will extend the Infrastructure configuration resource with a
+new status field to indicate whether the feature is on or off.
```
smarterclayton, line 200: This feels more like a feature gate - why not add new options to featuregate and just treat this as a generic feature?

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+
+This section outlines an end-to-end workflow for deploying a cluster
+with management-workload-partitioning enabled and how pods are
+correctly scheduled to run on the management CPU pool.
+
+1. User sits down at their computer.
+2. The user creates their `install-config.yaml`, including extra
+   values in the `bootstrapInPlace` section (since that enables
+   deploying a single-node).
+   * `managementCPUIDs` is a CPU set specifier for the CPUs to add to
```
smarterclayton, line 268: Before we approve this I kind of expect the answer about why this isn't just a specially named cpu pool

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+```ini
+[crio.runtime]
+mgmt_ctr_cpuset = “0-1”
+```
+
+This field describes the CPU set that management workloads will be
+configured to use.
+
+CRI-O will be configured to support a new annotation on pods,
+`io.openshift.management.cores`.
```
smarterclayton, line 312: I would expect just to use the resources

enhancements/management-workload-partitioning.md
mrunalp: We'll check if that comes down to CRI-O cc: @haircommander @rphillips

enhancements/management-workload-partitioning.md
mrunalp: cc: @deads2k

enhancements/management-workload-partitioning.md
mrunalp: kubelet will filter the pods with the label

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+  changed later. Future enhancements may address enabling or
+  reconfiguring the feature described here on an existing cluster.
+* This enhancement describes partitioning concepts that could be
+  expanded to be used for other purposes. Use cases for partitioning
+  workloads for other purposes may be addressed by future
+  enhancements.
+
+## Proposal
+
+[Previous attempts](https://github.com/openshift/enhancements/pull/62)
```
markmc, line 108:
```suggestion
[Previous attempts](https://github.com/openshift/enhancements/pull/628)
```

enhancements/management-workload-partitioning.md
haircommander: note: taking the `/` means we can't specify a per-container field idiomatically (like how apparmor does: `container.apparmor.security.beta.kubernetes.io/<container_name>`), which we were talking about doing for the cpushares

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+We want to continue to use the scheduler for placing management
+workloads, but we cannot rely on the CPU requests of those workloads
+to accurately reflect the constrained environment in which they are
+expected to run. Instead of scaling those CPU request values, we will
+change them to request a new [extended
+resource](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/),
+"management cores." We will modify kubelet to advertise management
+cores as an extended resource when the management workload
+partitioning feature is enabled, using a value equivalent to the CPU
+resources for the entire host. This large value should allow the
```
markmc, line 176: We're using want to advertise the full host's CPU availability just because it's a large number? In a typical case, this could easily be ~10 times the actual capacity of management cores. Are we saying this resource should essentially have unlimited availability?

If so, why do "we want to continue to use the scheduler"? What purpose does having a request for a finite amount of management core serve if the capacity is infinite?

markmc: Oh, is it *just* for the CPU shares as described in the next paragraph? Or is the scheduler related purpose too?

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+also identify add-on components that we know our initial customers
+will use and to which we contribute, and make the changes to add the
+labels ourselves, starting with
+
+* performance-addon-operator
+* sriov-network-operator
+* ptp-operator
+* local-storage-operator
+* cluster-logging (fluentd)
+* klusterlet (RHACM)
```
markmc, line 221: Consider kubevirt and N3000 operator too

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+
+#### High-level End-to-end Workflow
+
+This section outlines an end-to-end workflow for deploying a cluster
+with management-workload-partitioning enabled and how pods are
+correctly scheduled to run on the management CPU pool.
+
+1. User sits down at their computer.
+2. The user creates their `install-config.yaml`, including extra
+   values in the `bootstrapInPlace` section (since that enables
```
markmc, line 266: Are we saying this would go in the bootstrapInPlace section? Since/if we believe this won't be single-node specific forever, nor is it specific to bootstrap-in-place (i.e. single-node doesn't have to be bootstrap-in-place), it should be elsewhere

markmc (commented):
Awesome to see this proposal fleshed out so quickly, thanks!

enhancements/management-workload-partitioning.md
dhellmann: @derekwaynecarr emphasized that doing it this way made the performance of the cluster predictable from the outside. I think some of the rest of us liked it because it reduced the complexity and scope. If we don't have to support turning this on day 2, we simplify the first version and can ignore some edge cases, for now.

enhancements/management-workload-partitioning.md
dhellmann: Was this a belt-and-braces solution, in case we missed a static pod? Or is this a vestigial idea from before we decided we would have to update all of the operators anyway?

enhancements/management-workload-partitioning.md
dhellmann: I think the latest version of this doesn't use a separate pool in CRI-O. @mrunalp or @haircommander you may understand this point better than I do, do you want to respond?

enhancements/management-workload-partitioning.md
dhellmann: Is it possible to say that someone needs a specific role in order to add a specific label to a pod?

What namespaces do we have with mixed workloads?

enhancements/management-workload-partitioning.md
derekwaynecarr: filter on kubelet side stops mirrors pods from inaccurately reporting resources back to kube-apiserver.

enhancements/management-workload-partitioning.md
dhellmann: We will use the annotation on the pod to set CPU shares, but we wanted to maintain the scheduler use as well. I think that had to do with future work related to multi-node clusters, where only control plane or infrastructure nodes would advertise management cores as a resource, so management workloads would always be scheduled there. We may have other ways to achieve that, though.

Or perhaps this is another vestigial detail from the early version that talked about scaling the resources. @mrunalp or @derekwaynecarr do you remember?

enhancements/management-workload-partitioning.md
dhellmann: What is the distinction?

I think we picked that resource because the API server was already looking at it, but when we discussed it there wasn't a particularly firm reason to use it over something else. IIRC, @deads2k suggested another resource as an alternative, but I don't see that in the notes.

enhancements/management-workload-partitioning.md
dhellmann: Right, I wrote it this way to start so we would have something concrete, but there's an open question about the best placement.

enhancements/management-workload-partitioning.md
dhellmann: IIRC, the CRI-O team didn't want to add a field to all pools listing the CPUs and we wanted to reserve the ability to use runtime classes for other purposes. Maybe I'm mixing up concepts, though? @mrunalp or @haircommander ?

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+true`.
+
+We want to treat all OpenShift components as management workloads,
+including those that run the control plane. Therefore, kubelet will be
+modified to treat all static pods as management workloads, when the
+feature is enabled. We will update operator manifests and
+implementations to label all OpenShift components not running in
+static pods with the `io.openshift.management: true` label, and add a
+CI job to require that label be present in all workloads created from
+the release payload.
```
dhellmann, line 153: Example test from @deads2k : https://github.com/openshift/origin/pull/25992

enhancements/management-workload-partitioning.md
derekwaynecarr: when running SNO, management components implicitly include per node and control components.
when running compact clusters (schedulable control hosts), it likely means the same.
when running non-compact, management in this context means per node only (and not control).

dhellmann: Is it safe to assume that in a non-compact cluster the user is going to want to dedicate an entire host (or 3) to the management workloads?

enhancements/management-workload-partitioning.md
dhellmann: Ah, right. We need to do the same mutation that we do for normal pods, to change the resource requests, because otherwise the static pods will appear to consume CPU resources on the host.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+  likely to need different approaches.
+* This enhancement does not address non-compressible resource
+  requests, such as for memory.
+* This enhancement does not address ways to disable operators or
+  operands entirely.
+* Although the results of this enhancement may be useful for
+  single-node developer deployments, multi-node production
+  deployments, and Internet-of-things devices, those use cases are not
+  addressed directly in this document.
+* This enhancement does not address reducing actual utilization. There
```
derekwaynecarr, line 92: It does not address need, but it does put a cap on actual utilization, and the utilization ceiling is predictable at installation before admitting end user workload.

enhancements/management-workload-partitioning.md
dhellmann: How would that work for pods created from deployments managed by a cluster operator? Would all cluster operators need the extra permission?

enhancements/management-workload-partitioning.md
dhellmann: Good point, I'll add that when I make the next update pass.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+purposes of this document, "management workloads" include all
+OpenShift core components necessary to run the cluster, any add-on
+operators necessary to make up the "platform" as defined by telco
+customers, and operators or other components from third-party vendors
+that the customer deems as management rather than operational. It is
+important to note that not all of those components will be delivered
+as part of the OpenShift payload and some may be written by the
+customer or by vendors who are not our partners.
+
+The basic proposal is to provide a way to identify management
```
derekwaynecarr, line 127: it may be useful to define the 3 pools like follows:

management: cpuset that restricts management components
shared: management + !(dedicated)
dedicated: exclusive

the management pools is not exclusive, but its restricted from the outset to not use the full shared pool whose size may fluctuate.  this also keeps the capability compatible with the cpu manager component.

enhancements/management-workload-partitioning.md
derekwaynecarr: i am not sure i follow @haircommander .  i do prefer @smarterclayton syntax in that the label is classifying workload as either 'management' or 'workload'.

enhancements/management-workload-partitioning.md
derekwaynecarr: my view is if you deploy a static pod to a node, its a management component in the sense that its not bound to scheduling constraint, starts independent of an apiserver, and requires root permission on the node to even lay down.

enhancements/management-workload-partitioning.md
derekwaynecarr: i am not following where the definition of named pools is enumerated on the node?

my view was we have cpu-manager,topology-manager all enabled and we have a system/kube-reserved setting enabled that is integral, and we are effectively saying 'run management components in that reserved cpuset'.

enhancements/management-workload-partitioning.md
derekwaynecarr: i agree.
openshift monitoring is a good example.
it has per-node components and per infra host components (prometheus).
in standalone deployments with workers, we only want to compress per-node components.

enhancements/management-workload-partitioning.md
derekwaynecarr: opaque resources may only be visible as annotation at cri level, we do need to confirm.
we need a naming scheme for the annotation (if required) so please provide.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+
+The API server requires a configuration resource as input, rather than
+a ConfigMap or command line flag.  Therefore, we need an API-driven
+way to enable management workload partitioning in the admission
+hook. We will extend the Infrastructure configuration resource with a
+new status field to indicate whether the feature is on or off.
+
+*TBD: API field name, see open questions*
+
+Some pods used to run OpenShift control plane components are started
```
derekwaynecarr, line 204: can we enumerate?  static pod was the example i had in mind which is why kubelet itself needs to know how to do this conversion when reading static pods from disk prior to mirroring the pod back to apiserver.  this is why the kubelet carry is needed on static pod file parsing where it applies defaults today.

enhancements/management-workload-partitioning.md
haircommander: I have misunderstood the scope of this label. ignore that comment

enhancements/management-workload-partitioning.md
derekwaynecarr: kubelet itself knows two pools:

shared / dedicated

all we are doing is potentially subdividing the shared pool, but its not exclusive.
i see no way to make it exclusive without confusing cpu manager.
i also prefer that we keep the number of potential pools closely aligned to system/kube-reservation.

maybe we have 

```
- system-cpu-pools
  - name: management
    cpuIDs: 0-1
  - name: networking
    cpuIDs: 2-3
``

but all of these named pools are a subset of the kubelet understood shared pool and must fit within the reservation for cpu on the node.

Comment (derekwaynecarr) [reply enhancements/management-workload-partitioning.md:268]: issue with the above is if we have `workload.io.openshift=management,networking` as the type of label.

i would probably say enumerating it as a list, but we just support one name for now may be safest as potentially we may want to move log collection or networking on a different set in future?

File: enhancements/management-workload-partitioning.md (hunk 5 above)
Comment (dhellmann) [reply enhancements/management-workload-partitioning.md:158]: Are we talking about making the feature more general, instead of focusing on the management use case for now?

File: enhancements/management-workload-partitioning.md [hunk 20]
```diff
@@ -0,0 +1,549 @@
...
+document that the configuration only applies to pods created after it
+is set, so if an operator is installed before the namespace setting is
+changed the operator or its operands may need to be re-installed or
+restarted.
+
+The schedule for delivering this feature is very aggressive. We have
+tried to minimize the number of places that need complex changes to
+make it more likely that we can meet the deadline.
+
+Many implementation details proposed here require us to carry patches
```
derekwaynecarr, line 411: i have desired the ability to let the kubelet manage and launch pods that should have run in the reservation resource set.  for static pods in particular, this capability is very sensible to pursue upstream imo.  the pattern of cpu->management conversion if not applicable upstream is something we may want to publish for others in community to adopt similar to intel and nokia and others who have all explored cpu management policies outside of core k8s.  this has a benefit of being entirely complementary to the in-tree cpu manager.

enhancements/management-workload-partitioning.md
dhellmann: I'm not sure that what you're saying is incompatible with what is being proposed.

The idea with the pod label is to opt-in to being put on the management pool. The namespace label is about security, and giving the cluster admin control over what gets to use management pools at all.

Are we saying that we can't use the namespace because that gives anyone in that namespace the ability to run on the special pool?

enhancements/management-workload-partitioning.md
smarterclayton: maybe.  maybe not.  i think it’s orthogonal and if you create a management cpu pool you get all management workloads in there.  if we want to get clever later we come back with a more clever proposal

enhancements/management-workload-partitioning.md
dhellmann: For now we want to make sure everything is scheduled. In the future, we may tune the value to something more realistic so that the scheduler distributes work in a multi-node cluster.

enhancements/management-workload-partitioning.md
smarterclayton: i don’t agree.  i think there should be no magic, just the single label and the single path.  have label, go to cores.

enhancements/management-workload-partitioning.md
smarterclayton: i expected pools to be done exactly like existing topology, but maybe i don’t know how cores are assigned to topologies today.  i merely expected a well known topology name to map to this use case, not create a new config type.

can someone link to how i carve off cpus today?

enhancements/management-workload-partitioning.md
smarterclayton: we already do perms checks in pod admission, this is no different.

enhancements/management-workload-partitioning.md
smarterclayton: yeah,  we can define a virtual permission if necessary, or reuse a perm that is equivalent (like being able to delete a node) or similar (like being able to run privileged pods)

enhancements/management-workload-partitioning.md
smarterclayton: if we have to add a global config field i’d rather add the ability to enable specific features or put it behind an existing feature gate while alpha

enhancements/management-workload-partitioning.md
smarterclayton: we would have a label cpu.workload.openshift.io/management like node role probably,

we need to make a case whether a workload classification is exclusive or inclusive.  i’m pretty sure we would have more labels if we needed them since something can be both management and another classification, but i’m not convinced we need it yet

enhancements/management-workload-partitioning.md
dhellmann: The ones we identified are kubeapiserver, kube-controller-manager, scheduler, etcd. I will update the doc at this point to add the list.

enhancements/management-workload-partitioning.md
smarterclayton: also we have to prevent the same set of users from directly setting (create or update) the management core resource

enhancements/management-workload-partitioning.md
mrunalp: We can make everything generic up and down the stack so if needed we can support other pools in the future.

enhancements/management-workload-partitioning.md
derekwaynecarr: @smarterclayton there is no getting around carrying the filter patch in the kubelet admission of a static pod to strip/convert.  if your argument is 'only do that when the label is present' , i am fine with that.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+mutate regular pods.
+
+Kubelet will be changed so that when the feature is enabled, it
+advertises a new extended resource of "management cores", representing
+all of the CPU capacity of the host (not just the management CPU
+pool).
+
+#### Installer Changes
+
+The installer will be changed to accept the new `managementCPUIDs`
```
mdelder, line 372: How are the `managementCPUIDs` predicted? I saw a reference to integer counts earlier; is that the only component of the ID (e.g. some logical integer count for the available CPUs?) Would the same set of logical IDs (e.g. `1-2`) be appropriate for 10 different clusters configured via the same `install-config.yaml`?

enhancements/management-workload-partitioning.md
derekwaynecarr: ok, all of those should have the label, and the kubelet admission check to strip/convert on static pods would cover the gap i believe.

enhancements/management-workload-partitioning.md
derekwaynecarr: the enumerated ids may vary if hyperthreading is on/off , and if system is multi or single numa.  customer would know where they want us isolated.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+
+## Design Details
+
+### Open Questions [optional]
+
+1. If we think this feature is eventually going to be useful in
+   regular clusters, do we want the settings in the `bootstrapInPlace`
+   section? Should we add a `managementWorkloadPartitioning` section,
+   or something similar, and say for now that it only applies when the
+   single-node deployment approach is used?
```
staebler, line 427: We should not put the field under `bootstrapInPlace`, as long-term management workload partitioning  is orthogonal to whether the cluster uses bootstrap-in-place. The installer can still reject the install config when the field is set for a multi-node cluster or even for a non-bootstrap-in-place cluster. And we can easily relax that restriction in the future when we support other cluster configurations.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+
+We need to change pod definitions as each pod is created, so that the
+scheduler, kubelet, and CRI-O all see a consistently updated version
+of the pod and do not need to make independent decisions about whether
+it should be treated as a management workload. We need to intercept
+pod creation for *all* pods, without race conditions that might be
+introduced with traditional admission webhooks or controllers.
+Therefore, we will build an admission hook into the [kubernetes API
+server in
+OpenShift](https://github.com/openshift/kubernetes/tree/master/openshift-kube-apiserver/admission),
```
sttts, line 193: does this mean we mutate the requests to specify extended resources? Does this conflict with any reconcile function of existing controllers when the pod look different after creation? @soltysh

soltysh: That shouldn't be a problem, we already have a few mutating admissions (eg. Image injection from ImageStream) so that should work fine.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+annotation.
+
+Note that this field does not conflict with the `infra_ctr_cpuset`
+config option, as the infra container will still be put in that
+cpuset.  They can be configured as the same value if the infra
+container should also be considered to be managed.
+
+#### API Server Admission Hook
+
+A new admission hook in the kubernetes API server within OpenShift
```
soltysh, line 333: I'm curious why we prefer admission hook vs being explicit for specific set of pods?

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+support burstable or best-effort workloads.
+
+We want the isolation to be in place from the initial buildout of the
+cluster, to ensure predictable behavior. Therefore, the feature must
+be enabled during installation. To enable the feature, the user will
+specify the set of CPUs to run all management workloads. When the
+management CPU set is not defined, the feature will be completely
+disabled.
+
+We generally want components to opt-in to being considered management
```
romfreiman, line 141: why do we want to go with opt-in approach again? It's really had not to loose track of it, and requires high maintenance. Also, some yaml are templates and being rendered, others using knative, and more interesting approaches. Also, there are OLM operators - sriov, pao and others. Besides that, there are operatnds that are deployed only in certain cases (for example, registry in case of BM setup will not be deployed unless storage is configured). So it would be hard to get hermetic coverage in the CI.

Cannot we at least say the CVO does all the heavy lifting? If it recognizes a pod, it adds that annotation? It will solve 80% of the effort?

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+## Design Details
+
+### Open Questions [optional]
+
+1. If we think this feature is eventually going to be useful in
+   regular clusters, do we want the settings in the `bootstrapInPlace`
+   section? Should we add a `managementWorkloadPartitioning` section,
+   or something similar, and say for now that it only applies when the
+   single-node deployment approach is used?
+2. What should the new field in the Infrastructure CRD be?
```
dhellmann, line 428: 3. Should we worry about heterogeneous control plane hardware in multi-node clusters? The install-config schema may need to support specifying the same pool as using different CPUs on different nodes in that case.

enhancements/management-workload-partitioning.md
dhellmann: Ack. When we talked today, we agreed to move this out of the BiP section and come up with a better name.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,549 @@
...
+Enabling the feature after installation is not supported, so we do not
+need to address what happens if an older cluster upgrades and then the
+feature is turned on.
+
+### Version Skew Strategy
+
+N/A
+
+## Implementation History
+
```
dhellmann, line 506: https://github.com/openshift/kubernetes/pull/627 has the kubelet changes

https://github.com/cri-o/cri-o/pull/4680 has the first set of CRI-O changes

enhancements/management-workload-partitioning.md
dhellmann: The text changed enough that this comment thread is detached from the diff view, but I don't think we can resolve it until we hear from @smarterclayton

enhancements/management-workload-partitioning.md
dhellmann: I think this is basically the same comment being discussed in the thread starting with https://github.com/openshift/enhancements/pull/703#discussion_r597867142 so I am going to resolve this one.

dhellmann:
I have updated the draft document with changes based on the comments and feedback so far. Please take another look. If you want to see only what has changed today, you can look at the commit https://github.com/openshift/enhancements/pull/703/commits/66dee2dd4c01075d0607b579f079596ced3c9373

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,666 @@
...
+#### CRI-O Changes
+
+*NOTE: This section needs to be updated to deal with the generalized
+CPU pools.*
+
+CRI-O will be updated to support a new configuration value that a user
+can specify, `mgmt_ctr_cpuset`.
+
+```ini
+[crio.runtime]
```
mrunalp, line 361: We can make this work similar to how we specify runtime handlers today. e.g.
```
[crio.runtime.workloads.{workload-type}]
  cpu_set = "0-1"
```

For the management case:
```
[crio.runtime.workloads.management]
  cpu_set = "0-1"
```

In the future the configuration could be:
```
[crio.runtime.workloads.management]
  cpu_set = "0-1"

[crio.runtime.workloads.networking]
  cpu_set = "2-3"

```

cc: @haircommander

mrunalp: This can also handle other settings besides cpu_set for workloads in the future.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,666 @@
...
+capacity of the host (not just the management CPU pool).
+
+#### Installer Changes
+
+The installer will be changed to accept a new `workloadCPUPools`
+parameter with a list of CPU pools to treat as separate from the
+standard shared pool. (This name may change.)
+
+```yaml
+workloadCPUPools:
```
mrunalp, line 433: Now based on my CRI-O setting suggestion above, I am wondering if we should make this more general as well so if we have to handle things beyond CPU sets for workload types, we can do that in the future. With that, the top level could be called `workloadSettings` or `workload` and then list below could have more than just `cpuIDs` in the future.

enhancements/management-workload-partitioning.md
mrunalp: We can also add the `label` for activating this under the same setting and the `cpu_shares_annotation` to avoid any hard coding.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,666 @@
...
+cluster, to ensure predictable behavior. Therefore, the feature must
+be enabled during installation. To enable the feature, the user will
+specify the set of CPUs to run all management workloads. When the
+management CPU set is not defined, the feature will be completely
+disabled.
+
+We generally want components to opt-in to being considered management
+workloads. Therefore, for a regular pod to be considered to contain a
+management workload it must be labeled with a *workload label*,
+`workload.openshift.io/{pool_name}`. For now, we will focus on
```
sttts, line 152: what is the API support level of these?

I would prefer a very explicit "unsupported" keyword in these labels.

sttts: `workload.internal.openshfit.io/pool-name`

dhellmann: I don't know about "unsupported". The feature will be supported, right? And "internal" implies that it shouldn't be used by components that aren't part of the core release payload, but we must support add-on operators using it.

If the feature is only enabled with a feature gate, is that sufficient to communicate that it is an alpha feature?

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,666 @@
...
+   new feature.
+8. The kubelet advertises `cpu.workload.openshift.io/management`
+   extended resources on the node based on the number of CPUs in the
+   host.
+9. The kubelet reads static pod definitions. It replaces the `cpus`
+   requests with `cpu.workload.openshift.io/management` requests of
+   the same value and adds the
+   `cpu.workload.openshift.io/{container-name}` annotations for CRI-O
+   with the same values.
+10. Something schedules a regular pod with the
```
deads2k, line 310: I think we're likely to want to allow some pods in a namespace to target CPUs group/X and some pods in the same namespace to target CPUs group/Y.  In order to do this and maintain control, the namespace level label is like something like `workload.openhift.io/target-cpu-sets-allowed=management,in,comma,separate,list`.  And the pod level label is something like `workload.openshift.io/target-core-set=management`.

That is the solution I described in the e2e test that catches namespaces missing their labels: https://github.com/openshift/origin/pull/25992/files

dhellmann: Is it better/more consistent to use 1 label with multiple values than to use multiple labels? I don't have a good sense of what's "usual" for that. Nodes can have multiple role labels, for example, and this feels similar to that.

enhancements/management-workload-partitioning.md
dhellmann: I agree we should change it. I like `workloadSettings` but don't feel that strongly so I'll defer to the installer team.

enhancements/management-workload-partitioning.md
dhellmann: > We can also add the `label` for activating this under the same setting and the `cpu_shares_annotation` to avoid any hard coding.

So, the installer would set up a machine config resource to configure CRI-O with a new workloads section, including the cpu_set and the name of the label (or annotation?) that will be used to enable use of those workload settings. That restricts all of the openshift-specific stuff to the installer, and let's CRI-O work from its config file. Is that right?

mrunalp: Yes, right.

enhancements/management-workload-partitioning.md
dhellmann: We discussed this a bit more in slack and I think I understand the reasoning for doing it this way. I'm not sure if we want the spelling exactly as proposed. We talk about CPU sets in this doc, but including that in the label implies that's the only aspect of workload isolation that we might ever do (I could see networking being used, for example).

How about `workload.openshift.io/allowed=comma,separated,list` for the permission label on namespaces and `workload.openshift.io/target=single-name` for the label on pods indicating the intent?

dhellmann: I think the latest draft incorporates this comment thread. See 340d2dd62c1e669d68a59c76c8cc989386514340

enhancements/management-workload-partitioning.md
dhellmann: I have tried to capture this in the latest iteration. See 340d2dd62c1e669d68a59c76c8cc989386514340

enhancements/management-workload-partitioning.md
dhellmann: I captured this in 340d2dd62c1e669d68a59c76c8cc989386514340

dhellmann:
The latest draft incorporates feedback from @deads2k and @mrunalp about the label names and CRI-O schema. See 340d2dd62c1e669d68a59c76c8cc989386514340 for the delta.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,686 @@
...
+
+and
+
+```yaml
+annotations:
+  cpu.workload.openshift.io/container-name: 400m
+```
+
+
+#### CRI-O Changes
```
dhellmann, line 353: Some of the comments in https://github.com/openshift/enhancements/commit/340d2dd62c1e669d68a59c76c8cc989386514340#r48621955 aren't showing up in this section when I look at the full PR.

dhellmann: @haircommander @mrunalp @browsell I tried to capture what we just discussed in the latest patch. Please review this section.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,710 @@
...
+[crio.runtime.workloads.management]
+  label = "workload.openshift.io/target=mangement"
+  resource_annotation_prefix = "io.openshift.workload.management"
+  resources = {
+    "cpu": "",
+    "cpuset": "0-1",
+  }
+```
+
+CRI-O will be configured to support a new annotation on pods,
```
dhellmann, line 406: @haircommander @mrunalp @browsell This example doesn't seem to match what we have above. Is it still right? Do we still need `allowed_annotations`? Should it have a different value in the example?

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,710 @@
...
+The installer will be changed to accept a new `workloadSettings`
+parameter with a list of workload types and their settings. For now,
+the only setting will be the name and the CPU sets to treat as
+separate from the standard shared pool. The name `workloadSettings`
+may change, based on discussion on the installer code changes.
+
+```yaml
+workloadSettings:
+  - name: management
+    cpuIDs: 0-1
```
cynepco3hahue, line 476: Who will be responsible to remove the specified CPU set from the CPU manager?

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,710 @@
...
+```yaml
+requests:
+  cpu.workload.openshift.io/management: 400m
+```
+
+and
+
+```yaml
+annotations:
+  io.openshift.workload.management.cpu/container-name: 400m
```
cynepco3hahue, line 349: should be updated to `cpu.workload.openshift.io/{container-name}`

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,710 @@
...
+
+*NOTE: This section needs to be updated to deal with the generalized
+CPU pools.*
+
+CRI-O will be updated to support new configuration settings for workload types.
+
+```ini
+[crio.runtime.workloads.{workload-type}]
+  label = "workload.openshift.io/target={workload-type}"
+  resource_annotation_prefix = "io.openshift.workload.{workload-type}"
```
cynepco3hahue, line 366: should be updated to `cpu.workload.openshift.io`

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,710 @@
...
+values will not be allowed. In OpenShift, we do not want to allow
+arbitrary pods to set their own `cpuset` but other CRI-O users may
+want that ability.
+
+To pass a setting into CRI-O, the pod should have an annotation made
+by combining the `resource_annotation_prefix`, the key from
+`resources`, and the container name, like this:
+
+```
+io.openshift.workload.management.cpu/container_name = {value}
```
cynepco3hahue, line 391: should be updated to `cpu.workload.openshift.io/container_name = {value}`

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,710 @@
...
+```
+
+In the management workload case, we will configure it with values like
+
+```ini
+[crio.runtime.workloads.management]
+  label = "workload.openshift.io/target=mangement"
+  resource_annotation_prefix = "io.openshift.workload.management"
```
cynepco3hahue, line 399: should be updated to `cpu.workload.openshift.io`

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,710 @@
...
+```
+
+The `label` field is used to match pods that should be treated as
+having the workload type. The label key on the pod is compared for an
+exact match against the value specified in the configuration file.
+
+The `resource_annotation_prefix` is the start of the annotation key
+used to pass settings down to CRI-O.
+
+The `resources` map associates annotation suffixes with default
```
cynepco3hahue, line 380: Can you please clarify the difference between `cpu` and `cpuset`?
IIUC we should specify the cpuset that we will pin the relevant group of pods to it. But how will it work with the CPU shares?

mrunalp: We are using the CPU requests specified in the pod manifest for the containers and converting those to CPU shares. See the kubelet PoC PR.

enhancements/management-workload-partitioning.md
derekwaynecarr: @smarterclayton 

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/cpumanager/policy_static.go#L53

I was mapping this feature to what the static cpu manager defines as RESERVED.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,710 @@
...
+This section outlines an end-to-end workflow for deploying a cluster
+with management-workload-partitioning enabled and how pods are
+correctly scheduled to run on the management CPU pool.
+
+1. User sits down at their computer.
+2. The user creates their `install-config.yaml`, including extra
+   values in the specifying the CPUs to include in the management CPU
+   set. (See below for details.)
+3. The user runs the installer.
+4. The installer uses the management CPU pool settings to generate an
```
staebler, line 293: Is the expectation that the management CPU pool is specified for both the master machineconfigpool and the worker machineconfigpool? If so, is the same cpuset going to be universally applicable to both machineconfigpools?

enhancements/management-workload-partitioning.md
dhellmann: The next draft will include a comment about using PAO on line 246 and in the GA graduation criteria.

enhancements/management-workload-partitioning.md
dhellmann: The next draft includes an updated schema in the next section. The wording in this section is trying to explain how the management CPU pool will be configured as a specific case, so I would like to keep the wording as-is here.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,736 @@
...
+The `cpuIDs` value is a CPU set specifier for the CPUs to add to the
+isolated set used for management components, using the same cpuset
+syntax used elsewhere in kubernetes. Asking for explicit IDs instead
+of simply a count gives the user control in situations where they need
+specific CPUs to be available because they're close to accelerators,
+NICs, or other special hardware.
+
+The installer will be changed to generate an extra machine config
+manifest to configure CRI-O so that containers from pods with the
+`cpu.workload.openshift.io/management` annotation are run on the
```
smarterclayton, line 514: I expected to see "partition" in this (if partition is the name we settle on).

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,736 @@
...
+mutate regular pods.
+
+Kubelet will be changed so that when the feature is enabled, it
+advertises a new extended resource of
+`cpu.workload.openshift.io/management`, representing all of the CPU
+capacity of the host (not just the management CPU pool).
+
+#### Installer Changes
+
+The installer will be changed to accept a new `workload.partitions`
```
smarterclayton, line 476: Whenever we introduce new terminology to the entire project we need to be careful to select the naming.  I'd expect at minimum a paragraph here justifying the choice of both words and the alternatives considered and why.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,736 @@
...
+parameter for each [machine
+pool](https://github.com/openshift/installer/blob/master/docs/user/customization.md#machine-pools)
+with a list of workload types and their settings. For now, the only
+setting will be the name and the CPU sets to treat as separate from
+the standard shared pool. The name `workload.partitions` may change,
+based on discussion on the installer code changes.
```
smarterclayton, line 482: Final name needs to be in the enhancement (as I mentioned before, we are adding a top level concept to openshift, that comes with a high expectation of due dilligence on naming)

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,736 @@
...
+
+```yaml
+controlPlane:
+  name: master
+  workload:
+    partitions:
```
smarterclayton, line 488: So looking at this in context - this is two INCREDIBLY generic terms, and then you use `cpuIDs` which is incredibly specific.  Generally we use workload to mean end user application - it's not unreasonable to have a management workload though (especially when there is shared responsibility between project and user for some components).  So "management" is reasonable.

I think it's partitions which is too generic.  What other things go into workload?  How many things besides "partitions"?  I'd like to see 2-3 examples of "concepts we might add into this structure" (i'm not against future proofing with structs, but it's gotta have justification).

For instance, make an argument why we wouldn't just say `workloadPartitions`, and make an argument for, and do it here in this doc.

dhellmann: @crawford did you have anything in mind for other settings under `workload`?

Do we want another term instead of `partition`? We've used "CPU pools" elsewhere in describing what we're doing, do we want to use `workload.cpuPools`? Or is that overloading the term "pool"?

enhancements/management-workload-partitioning.md
dhellmann: @smarterclayton so would that be `cpu.partition.workload.openshift.io/management` (assuming we stick with "partition")?

enhancements/management-workload-partitioning.md
mrunalp: It should match the latest iteration of what we discussed today for the management stanza.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+requests:
+  cpu.workload.openshift.io/management: 400
+```
+
+and
+
+```yaml
+annotations:
+  io.openshift.workload.management.cpu/{container-name}: 400
+```
```
mrunalp, line 424: We can have a table of conversions here. E.g. 400m -> 400, 0.3 -> 300 1 -> 1000 and so on.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+   exceed the actual CPU capacity of the management CPU pool.
+2. It will strip all container-specific workload annotations from new
+   pods. This both prevents a user from configuring a pod incorrectly
+   and prevents unprivileged users from running their pods in a CPU
+   pool they should not have access to.
+3. It will add annotations with for each container with the value
+   equal to the original CPU requests * 1000, so that CRI-O can use
+   the value to configure the CPU shares for the container.
+
+We will not change pods in a way that changes their quality-of-service
```
mrunalp, line 525: nit: Can use the upstream terminology here: `Quality of Service (QoS)`. See https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+memory requests, because if we mutate the pod so that it has no CPU or
+memory requests the quality-of-service class of the pod would be
+changed automatically. Any labeled pod that is already BestEffort
+would be annotated using `0` as the value so that CRI-O will have an
+indicator to configure the CPU shares as BestEffort.
+
+#### Kubelet Changes
+
+Kubelet will be changed to look for a configuration file,
+`/etc/kubernetes/management-pinning`, to enable the management
```
mrunalp, line 536: We are using `/etc/kubernetes/workload-pinning` in the kubelet PR.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+specifier for the CPUs making up each CPU pool.
+
+Kubelet will be changed so that when the feature is enabled, [when a
+static pod definition is
+read](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/config/file_linux.go)
+it is mutated in a way similar to the API server admission hook will
+mutate regular pods.
+
+Kubelet will be changed so that when the feature is enabled, it
+advertises a new extended resource of `cpu.workload.openshift.io/{CPU
```
mrunalp, line 547: {workload-type}.workload.openshift.io/cores

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+### Implementation Details/Notes/Constraints
+
+#### High-level End-to-end Workflow
+
+This section outlines an end-to-end workflow for deploying a cluster
+with workload partitioning enabled and how pods are correctly
+scheduled to run on the management CPU pool.
+
+1. User sits down at their computer.
+2. The user creates a machine config manifest to configure CRI-O to
```
mrunalp, line 341: We need to show the manifests somewhere below in the details/implentation sections.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+`workload.openshift.io/allowed={comma_separated_list_of_type_names}`
+will be subject to special handling.
+
+We want to continue to use the scheduler for placing management
+workloads, but we cannot rely on the CPU requests of those workloads
+to accurately reflect the constrained environment in which they are
+expected to run. Instead of scaling those CPU request values, we will
+change them to request a new [extended
+resource](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/)
+called `cpu.workload.openshift.io/{target}` (for example,
```
browsell, line 198: {workload}.workload.openshift.io/cores

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+`cpu.workload.openshift.io/management`).  We will modify kubelet to
```
browsell, line 199: management.workload.openshift.io/cores

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+* local-storage-operator
+* cluster-logging (fluentd)
+* klusterlet and klusterlet add-ons (RHACM)
+* kubevirt
+* N3000
+
+Due to time constraints, the first iteration of this work will also
+require the
+[performance-addon-operators](https://github.com/openshift-kni/performance-addon-operators)
+to configure the [kube-reserved and system-reserved CPU sets for
```
browsell, line 292: configure reserved-cpus

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+2. The user creates a machine config manifest to configure CRI-O to
+   partition management workloads.
+3. The user creates a machine config manifest to write a configuration
+   file for kubelet to enable the same workload partition. The file
+   should only be readable by the kubelet.
+4. The user runs the installer to create the standard manifests, adds
+   their extra manifests from steps 2 and 3, then creates the cluster.
+5. The kubelet starts up and finds the configuration file enabling the
+   new feature.
+6. The kubelet advertises `cpu.workload.openshift.io/management`
```
browsell, line 350: management.workload.openshift.io/cores

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+7. The kubelet reads static pod definitions. It replaces the `cpu`
+   requests with `cpu.workload.openshift.io/management` requests of
+   the same value and adds the
+   `io.openshift.workload.management.cpu/{container-name}` annotations
+   for CRI-O with the same values.
+8. Something schedules a regular pod with the
+   `workload.openshift.io/management` annotation in a namespace with
+   the `workload.openshift.io/allowed=management` label.
+9. The admission hook modifies the pod, replacing the CPU requests
+   with `cpu.workload.openshift.io/management` requests and adding the
```
browsell, line 362: management.workload.openshift.io/cores

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+   `io.openshift.workload.management.cpu/{container-name}` annotations
+   for CRI-O.
+10. The scheduler sees the new pod and finds available
+    `cpu.workload.openshift.io/management` resources on the node. The
```
browsell, line 366: management.workload.openshift.io/cores

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,787 @@
...
+requests:
+  cpu:
+    400m
+```
+
+becomes
+
+```yaml
+requests:
+  cpu.workload.openshift.io/management: 400
```
browsell, line 416: management.workload.openshift.io/cores

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,788 @@
...
+CRI-O will be configured to support a new annotation on pods,
+`io.openshift.workload.management.cpushares/{container-name}`.
+
+```ini
+[crio.runtime.runtimes.runc]
+  runtime_path = "/usr/bin/runc"
+  runtime_type = "oci"
+```
+
+Pods that have the `io.openshift.workload.management.cpushares`
```
mrunalp, line 498: Should be `workload.openshift.io/management`

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,788 @@
...
+annotation will have their cpuset configured to the value from the
+appropriate workload configuration, as well as have their CPU shares
+configured to the value of the annotation.
```
mrunalp, line 501: of the annotation `io.openshift.workload.management.cpushares`

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,788 @@
...
+
+## Implementation History
+
+A CI job to identify pods that need a workload label and ensure it is
+present: https://github.com/openshift/origin/pull/25992
+
+Kubelet changes: https://github.com/openshift/kubernetes/pull/627
+
+The first set of CRI-O changes to refactor annotation handling:
+https://github.com/cri-o/cri-o/pull/4680
```
mrunalp, line 654: https://github.com/cri-o/cri-o/pull/4725 included all the changes

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,852 @@
...
+```text
+io.openshift.workload.management.cpushares/container_name = {value}
+```
+
+In the management workload case, we will configure it with values like
+
+```ini
+[crio.runtime.workloads.management]
+  activation_annotation = "workload.openshift.io/management"
+  resource_annotation_prefix = "io.openshift.workload.management"
```
haircommander, line 482: did this change from `annotation_prefix` ?

dhellmann: Fixed

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,852 @@
...
+  resources = {
+    "cpushares": "",
+    "cpuset": "0-1",
+  }
```
haircommander, line 486: nit: for us to process this it has to be on one line

haircommander: also, we have to set the key as '=':
`resources = { "cpushares" =  "$cpushares", "cpuset" = "$cpuset", }`

dhellmann: oh, no, I see you did that with the cpuset, too, so I guess that's just an example?

dhellmann: @haircommander I think I've corrected this.

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,851 @@
...
+change the configuration schema that kubelet inherits from
+upstream. Therefore we will have kubelet look for a new configuration
+file on startup, and the feature will only be enabled if that file is
+found.
+
+We want to give cluster administrators control over which workloads
+are run on the management CPUs. Normal users cannot add a label to a
+namespace without the right RBAC permissions. Therefore, only pods in
+namespaces labeled with a label
+`workload.openshift.io/allowed={comma_separated_list_of_type_names}`
```
lack, line 190: I think this is not an annotation? `workload.openshift.io/allowed: "management"`

lack: Line 359 says it's an annotation, not a label.

lack (changes requested):

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,851 @@
...
+io.openshift.workload.management.cpushares/container_name = {value}
+```
+
+In the management workload case, we will configure it with values like
+
+```ini
+[crio.runtime.workloads.management]
+  activation_annotation = "workload.openshift.io/management"
+  annotation_prefix = "io.openshift.workload.management"
+  resources = { "cpushares" = "", "cpuset" = "0-1" }
```
lack, line 483: Line 450 says "cpu", not "cpushares".  Which one is correct?

markmc:
Awesome work, all. It looks like there aren't any significant open questions or issues.

I'm still bothered by "a value equivalent to the CPU resources for the entire host" - it seems like an arbitrary over-subscription ratio, but I guess we don't ever expect to actually exhaust this resource?

I find the CRI-O section pretty confusing, but happy to put that down to a lack of knowledge on my part.

It's disappointing that we won't have installer support this go-round, but understandable. We should track this part as a GA requirement for a future release.

Thanks!

/approve

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,848 @@
...
+memory requests, because if we mutate the pod so that it has no CPU or
+memory requests the quality-of-service class of the pod would be
+changed automatically. Any pod that is already BestEffort
+would be annotated using `0` as the value so that CRI-O will have an
+indicator to configure the CPU shares as BestEffort.
+
+#### Kubelet Changes
+
+Kubelet will be changed to look for a configuration file,
+`/etc/kubernetes/workload-pinning`, to enable the management
```
derekwaynecarr, line 534: minor nit: can we name this file with an openshift domain prefix or suffix?

`/etc/kubernetes/openshift-workload-pinning`

derekwaynecarr (approved):
great work.

/approve
/lgtm

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,848 @@
...
+   should only be readable by the kubelet.
+4. The user runs the installer to create the standard manifests, adds
+   their extra manifests from steps 2 and 3, then creates the cluster.
+5. The kubelet starts up and finds the configuration file enabling the
+   new feature.
+6. The kubelet advertises `management.workload.openshift.io/cores`
+   extended resources on the node based on the number of CPUs in the
+   host.
+7. The kubelet reads static pod definitions. It replaces the `cpu`
+   requests with `cpu.workload.openshift.io/management` requests of
```
cynepco3hahue, line 354: `cpu.workload.openshift.io/management` -> `management.workload.openshift.io/cores`

enhancements/management-workload-partitioning.md
dhellmann: > It's disappointing that we won't have installer support this go-round, but understandable. We should track this part as a GA requirement for a future release.

https://github.com/openshift/enhancements/pull/727

dhellmann:
> It's disappointing that we won't have installer support this go-round, but understandable. We should track this part as a GA requirement for a future release.

https://github.com/openshift/enhancements/pull/727

enhancements/management-workload-partitioning.md
dhellmann: #726 updates the name

enhancements/management-workload-partitioning.md
dhellmann: Thanks, #728 fixes this

enhancements/management-workload-partitioning.md
```diff
@@ -0,0 +1,848 @@
...
+
+The admission hook will reject pods with multiple annotations, for
+now. Future work may allow multiple workload types with different
+priorities to support clusters with different types of configurations.
+
+#### Pod mutation
+
+The kubelet and API admission hook will change pods annotated with
+`workload.openshift.io/management` so the CPU requests are replaced
+with management CPU requests and an annotation is added with the same
```
wking, line 403: The cluster-version operator's reconciliation currently [requires in-cluster `requests` to `DeepEqual` the manifest's requests][1], so it's going to be hotlooping on this trying to stomp out the admission hook's change.  We have one CVO-hotloop bug where [one potential approach is server-side apply][2], which should work around that (unless I'm misunderstanding server-side apply).  But looks like server-side apply [is still beta][3].  If we need this to work before we have stable server-side apply, do folks have recommendations for updating the CVO's `requests`-reconciliation logic?  Do we need to teach it about what the admission hook might do, so it can ignore that?  Something less invasive?  I expect this is going to be an issue for every operator which manages operand workloads, not just for the CVO.

[1]: https://github.com/openshift/cluster-version-operator/blob/master/lib/resourcemerge/core.go#L522-L530
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1881484#c1
[3]: https://kubernetes.io/docs/reference/using-api/server-side-apply/

wking: @dhellmann pointed out that the admission hook will only mutate pods, not Deployments, etc., so the CVO won't notice.  I dunno how the Deployment and DaemonSet controllers handle reconciling pod requests, but that's not my code to worry about ;)

dhellmann: According to @derekwaynecarr those controllers allow mutations like this. That's how service mesh layers work, for example.

PR #449, opened 2020-08-21
Add Tunable Router Buffer Sizes EP

sgreene570:
Enhancement Proposal to support https://issues.redhat.com/browse/NE-354.

PR #708, opened 2021-03-25
Ingress: HAProxy thread tuning

rfredette:
This proposal is to allow the cluster administrator to configure the number of connection handling threads within ingress controller pods.

### DIFF
--- a/enhancements/ingress/haproxy-thread-tuning.md
+++ b/enhancements/ingress/haproxy-thread-tuning.md
@@ -6,7 +6,7 @@ reviewers:
   - "@Miciah"
   - "@danehans"
   - "@frobware"
-  - "@sgreene"
+  - "@sgreene570"
   - "@knobunc"
   - "@miheer"
   - "@candita"
@@ -16,9 +16,9 @@ approvers:
   - "@frobware"
   - "@danehans"
 creation-date: 2021-03-24
-last-updated: 2021-03-25
+last-updated: 2021-03-26
 status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced|informational
-see-also:
+see-also: [Tunalbe Router Buffer Sizes](https://github.com/openshift/enhancements/pull/449)
 replaces:
 superseded-by:
 ---
@@ -37,26 +37,27 @@ superseded-by:
 ## Summary
 
 This proposal is to allow the cluster administrator to configure the number of
-connection handling threads within router pods.
+connection handling threads within ingress controller pods.
 
 ## Motivation
 
 As the number of routes and the volume of traffic through a cluster increases,
-eventually the router pod will reach a limit to how many connections it can
-handle. In order to increase the maximum number of connections, administrators
-can split the routes into multiple shards, each shard being handled by a
-separate router.  More router pods can also be deployed without sharding,
-allowing more connections to be handled, but this comes with more overhead, as
-each router must separately verify the health of its backends.  These redundant
-health checks increase the volume of traffic passing through the cluster, and
-can cancel out much of the performance gain from deploying more router pods.
+eventually the ingress controller pod will reach a limit to how many
+connections it can handle. In order to increase the maximum number of
+connections, administrators can split the routes into multiple shards, each
+shard being handled by a separate ingress controller.  More ingress controller
+pods can also be deployed without sharding, allowing more connections to be
+handled, but this comes with more overhead, as each ingress controller must
+separately verify the health of its backends.  These redundant health checks
+increase the volume of traffic passing through the cluster, and can cancel out
+much of the performance gain from deploying more ingress controller pods.
 
 In OpenShift 3.x, users could configure routers to increase the number of
 threads that HAProxy spawns, allowing a single router to handle more
 connections without additional health check overhead. This document proposes to
 allow users to employ a similar strategy, and provide a field within the
-IngressController API to specify the number of threads allocated within router
-pods
+IngressController API to specify the number of threads allocated within ingress
+controller pods
 
 ### Goals
 
@@ -69,8 +70,8 @@ Expose additional performance tuning parameters available within HAProxy
 
 ## Proposal
 
-Add the subresource `threading` to the IngressController API. It will currently
-contain one field, `count`.
+Add the field `threading` to the IngressController API. It will currently
+contain one subfield, `count`.
 
 ```go
 type IngressControllerSpec struct {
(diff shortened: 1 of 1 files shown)


enhancements/ingress/haproxy-thread-tuning.md
```diff
@@ -0,0 +1,190 @@
...
+	//
+	// +optional
+	Threading IngressControllerThreading `json:"threading,omitempty"`
+}
+
+type IngressControllerThreading struct {
+	// count defines the number of threads created per router pod. Creating
+	// more threads allows each router pod to handle more connections, at the
+	// cost of more system resources used. If this field is empty, the
+	// IngressController will use a default value of 4 threads.
```
frobware, line 93: Add a note that the value is "4" because that's the current default in OCP 4.7 and we don't want to perturb existing behaviour.

enhancements/ingress/haproxy-thread-tuning.md
```diff
@@ -0,0 +1,190 @@
...
+$ oc patch ingresscontroller/<controller-name> --type=merge -p '{"spec": {"threading": {"count": <new-thread-count>} } }'
+```
+
+New router pods will be rolled out with the updated thread count.
+
+### Risks and Mitigations
+
+#### Increased Resource Usage
+
+It's possible that an administrator could configure routers to have enough
```
frobware, line 135: nit: wording; "to over specify the number of threads"? "to overcommit on the the number of threads"?

enhancements/ingress/haproxy-thread-tuning.md
```diff
@@ -0,0 +1,190 @@
...
+	Count int `json:"count,omitempty"`
+}
+```
+
+When unset, the ingress operator will provision routers with 4 threads,
+matching the existing behavior in OpenShift 4.7.
+
+It would be possible to implement this field as a simple integer within the
+`IngressControllerSpec` as `spec.threadCount` or something similar, however
+there are other tunable threading performance options that HAProxy exposes, and
```
frobware, line 107: Perhaps cite "cpu pinning" as a concrete example of something we may want to consider in the future.

enhancements/ingress/haproxy-thread-tuning.md
```diff
@@ -0,0 +1,190 @@
...
+preferable to include those fields within a single subresource. In order to
+make that potential upgrade less painful, the `spec.threading` subresource will
+be added with only one field, `spec.threading.count`.
+
+### User Stories
+
+#### Story 1
+
+> As a cluster administrator, I want to increase the amount of incoming
+> connections my cluster can handle without configuring IngressController
```
frobware, line 119: I think another reason is that I have a big box that is under utilised w.r.t. routing (e.g., 256GB RAM, 100 cores).

frobware (commented):
Overall LGTM but leaving for additional reviews/comments.

enhancements/ingress/haproxy-thread-tuning.md
Miciah: In short: "scale up, not out".  The API lets you have more threads per pod replica, which means you may not need to create more pod replicas to accommodate load.

enhancements/ingress/haproxy-thread-tuning.md
```diff
@@ -0,0 +1,190 @@
...
+Provide an API for configuring the number of threads handling connections in
+router pods
+
+### Non-Goals
+
+Expose additional performance tuning parameters available within HAProxy
+
+## Proposal
+
+Add the subresource `threading` to the IngressController API. It will currently
```
Miciah, line 72: "subresource" → "subfield" (or just "field").

enhancements/ingress/haproxy-thread-tuning.md
```diff
@@ -0,0 +1,190 @@
...
+}
+
+type IngressControllerThreading struct {
+	// count defines the number of threads created per router pod. Creating
+	// more threads allows each router pod to handle more connections, at the
+	// cost of more system resources used. If this field is empty, the
+	// IngressController will use a default value of 4 threads.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=4
```
Miciah, line 96: Might as well make the minimum 1.  (I remember we discussed this, but was there some objection to going with 1?)

Miciah:
BTW, please put "ingress:" in the title of the PR, and add a description to the PR (you can copy and paste the summary from the enhancement proposal).

enhancements/ingress/haproxy-thread-tuning.md
```diff
@@ -0,0 +1,190 @@
...
+  - "@candita"
+approvers:
+  - "@knobunc"
+  - "@miciah"
+  - "@frobware"
+  - "@danehans"
+creation-date: 2021-03-24
+last-updated: 2021-03-25
+status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced|informational
+see-also:
```
sgreene570, line 21: May want to include https://github.com/openshift/enhancements/pull/449 (Tunable router buffers) in "see-also".

sgreene570 (commented):
Looks great so far! Some very minor suggestions.

enhancements/ingress/haproxy-thread-tuning.md
```diff
@@ -0,0 +1,190 @@
+---
+title: haproxy-thread-tuning
+authors:
+  - "@rfredette"
+reviewers:
+  - "@Miciah"
+  - "@danehans"
+  - "@frobware"
+  - "@sgreene"
```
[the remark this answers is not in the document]
rfredette: apologies, I've fixed it

enhancements/ingress/haproxy-thread-tuning.md
candita: Does the minimum stay at 4 or change to 1?

rfredette: Good catch. I meant to change it to 1

enhancements/ingress/haproxy-thread-tuning.md
```diff
@@ -0,0 +1,216 @@
...
+  - "@candita"
+approvers:
+  - "@knobunc"
+  - "@miciah"
+  - "@frobware"
+  - "@danehans"
+creation-date: 2021-03-24
+last-updated: 2021-03-26
+status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced|informational
+see-also: [Tunalbe Router Buffer Sizes](https://github.com/openshift/enhancements/pull/449)
```
Miciah, line 21: I think these references are supposed to be formatted like this:
```suggestion
see-also:
- "/enhancements/ingress/tunable-router-buffer-sizes.md"
```
Looks like there is little consistency among existing enhancements though.  Let's sort this out separately; no need to hold up the PR for this.

enhancements/ingress/haproxy-thread-tuning.md
```diff
@@ -0,0 +1,216 @@
...
+  - "@miheer"
+  - "@candita"
+approvers:
+  - "@knobunc"
+  - "@miciah"
+  - "@frobware"
+  - "@danehans"
+creation-date: 2021-03-24
+last-updated: 2021-03-26
+status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced|informational
```
Miciah, line 20:
```suggestion
status: implementable
```
...But let's get this in a follow-up, when we can change the status to "implemented".  * grin *.

Miciah (commented):
Looks great!  Thanks!  
/approve
/lgtm

PR #709, opened 2021-03-26
Add proposal for credentials management outside the openshift cluster for new platforms

akhil-rane:
/cc @dgoodwin @abhinavdahiya @staebler @joelddiaz @sdodson @derekwaynecarr

### DIFF
--- a/enhancements/installer/credentials-management-outside-openshift-cluster.md
+++ b/enhancements/installer/credentials-management-outside-openshift-cluster.md
@@ -1,0 +1,134 @@
+---
+title: Credentials Management outside OpenShift Cluster
+authors:
+  - "@akhil-rane"
+reviewers:
+  - "@abhinavdahiya"
+  - "@dgoodwin"
+  - "@joelddiaz"
+approvers:
+  - "@derekwaynecarr"
+  - "@sdodson"
+creation-date: 2021-03-25
+last-updated: 2021-03-25
+status: provisional
+---
+
+# Credentials Management outside OpenShift Cluster
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `partially implemented`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The intent of this enhancement is to take the process of credentials management outside the OpenShift cluster for new 
+platforms. We propose to make *manual* mode as default for clusters on new platforms, optionally supported by CLI tooling
+that follows the best practices of the hosting infrastructure at that time.
+
+## Motivation
+
+The main motivation behind this enhancement is to satisfy the customer requirement to follow the best security practices 
+by not storing admin credentials inside the cluster
+
+### Goals
+
+As part of this enhancement we plan to do the following:
+* Guide new platforms to start with Manual mode rather than Mint mode.
+* Provide optional tooling (ccoctl) for administrators to manage their credentials outside the cluster. (prep for install, reconcile, prep for upgrade)
+* Establish precedent for future direction of preferring *Manual* over *Mint*.
+
+### Non-Goals
+
+* Removing Mint mode for any currently supported cloud. (AWS, Azure, GCP)
+* Changing the default for any provider currently defaulting to Mint mode, though we may revisit in the future.
+
+## Proposal
+
+Currently, *mint* is a default credentials mode for OpenShift. In this mode we run OpenShift installer with an admin 
+level cloud credential. The admin credential is stored in kube-system namespace and then used by the cloud credential 
+operator to process the CredentialRequests in the cluster and create new users with fine-grained permissions.
+The customers have reported that the need to store admin credentials inside the cluster is a major disadvantage. Based 
+on the feedback, we propose to run credentials related setup outside a cluster and then start installation process in a 
+*manual* mode without providing admin credentials to the installer.
+
+### User Stories
+
+#### Story 1
+As an OpenShift cluster administrator I should be able to extract the CredentialsRequest manifests from the release image 
+and create required manifests/infrastructure to satisfy all in-cluster component's CredentialsRequests. This has been 
+already [implemented](https://github.com/openshift/cloud-credential-operator/blob/master/docs/ccoctl.md).
+
+#### Story 2
+As an OpenShift cluster administrator, I should be able to reconcile my credentials manually using local CLI tooling on 
+an on-going basis. This is in [backlog](https://issues.redhat.com/browse/CCO-106) and targeted for 4.9.
+
+#### Story 3
+As an OpenShift cluster administrator, I should be able to prep credentials for a cluster upgrade. This is in [backlog](https://issues.redhat.com/browse/CCO-106) 
+and targeted for 4.9.
+
+### Risks and Mitigations
+
+## Design Details
+
+We intend to build an optional tool **ccoctl** which will handle credentials management of the cluster in *manual* mode. 
+The following is the set of requirements for the current prototype (for AWS). We can have something similar for other 
+platforms.
+
+* ccoctl should be able to setup Identity Provider to authenticate OpenShift components
+* ccoctl should be able to take a list of CredentialsRequests from the release image and create/update 
+  Roles/Users/ServicePrincipals/ServiceAccounts with appropriate permissions.
+* ccoctl should be able to take list of CredentialsRequests for a release image, and the Identity Provider URL to generate 
+  the objects that need to be passed to the installer for successful installation
+* ccoctl should be able to delete all the resources that it had created in the cloud
+
+We envision **ccoctl** as a recommended tool to setup credentials on new platforms but customers are free to use other 
+tools like Terraform/AWS CloudFormation to do the above mentioned tasks. Read more about AWS implementation details [here](https://github.com/openshift/cloud-credential-operator/blob/master/docs/ccoctl.md)
+
+We also plan to have a detailed documentation in place to guide new cloud providers to implement *manual* mode.
+
+To upgrade a cluster we need to execute following steps:
+* Examine the CredentialsRequests in the new OpenShift release. Check if permissions in the existing CredentialsRequest
+  have changed.
+* Create/update credentials in the underlying cloud provider, and also create/update Kubernetes Secrets in the correct
+  namespaces to satisfy all CredentialsRequests in the new release.
+* CCO sets `Upgradeable=False` when in manual mode. Set an appropriate annotation `cloudcredential.openshift.io/upgradeable-to` 
+  to a new upgradable version to allow upgrade.
+
+All the above task will also be executable by *ccoctl*. Details in [this](https://issues.redhat.com/browse/CCO-84) card.
+
+### Open Questions
+
+### Test Plan
+
+We plan to have a e2e test that will externally set up a credentials management infrastructure and then kickstart 
+install in a *manual* mode.
+
+### Graduation Criteria
+
+### Upgrade / Downgrade Strategy
+
+### Version Skew Strategy
+
+## Implementation History
+
+We currently have a work-in-progress CLI tool [ccoctl](https://github.com/openshift/cloud-credential-operator/blob/master/docs/ccoctl.md) 
+to create and manage cloud credentials outside the cluster for AWS cloud. The design details of this tool is discussed above.
+
+## Drawbacks
+
+* Taking the credentials management outside the cluster will create additional overhead for the customer to make sure all
+  the required infrastructure is in place before starting the installation process. Current tooling we have only supports
+  the AWS cloud, we do not have anything planned for other cloud providers.
+* Push-button upgrades will not work in *manual* mode as the cluster no longer has the admin credentials to mint credentials 
+  (with fine-grained permissions) for in-cluster components.
+
+## Alternatives
+
+## Infrastructure Needed [optional]
+
--- a/enhancements/installer/credentials-management-outside-openshift-cluster.md
+++ b/enhancements/installer/credentials-management-outside-openshift-cluster.md
@@ -27,13 +27,13 @@ status: provisional
 
 ## Summary
 
-The intent of this enhancement is to take the process of credentials management outside the OpenShift cluster for new 
+The intent of this enhancement is to take the process of credentials management outside the OpenShift cluster for new
 platforms. We propose to make *manual* mode as default for clusters on new platforms, optionally supported by CLI tooling
 that follows the best practices of the hosting infrastructure at that time.
 
 ## Motivation
 
-The main motivation behind this enhancement is to satisfy the customer requirement to follow the best security practices 
+The main motivation behind this enhancement is to satisfy the customer requirement to follow the best security practices
 by not storing admin credentials inside the cluster
 
 ### Goals
@@ -50,44 +50,44 @@ As part of this enhancement we plan to do the following:
 
 ## Proposal
 
-Currently, *mint* is a default credentials mode for OpenShift. In this mode we run OpenShift installer with an admin 
-level cloud credential. The admin credential is stored in kube-system namespace and then used by the cloud credential 
+Currently, *mint* is a default credentials mode for OpenShift. In this mode we run OpenShift installer with an admin
+level cloud credential. The admin credential is stored in kube-system namespace and then used by the cloud credential
 operator to process the CredentialRequests in the cluster and create new users with fine-grained permissions.
-The customers have reported that the need to store admin credentials inside the cluster is a major disadvantage. Based 
-on the feedback, we propose to run credentials related setup outside a cluster and then start installation process in a 
+The customers have reported that the need to store admin credentials inside the cluster is a major disadvantage. Based
+on the feedback, we propose to run credentials related setup outside a cluster and then start installation process in a
 *manual* mode without providing admin credentials to the installer.
 
 ### User Stories
 
 #### Story 1
-As an OpenShift cluster administrator I should be able to extract the CredentialsRequest manifests from the release image 
-and create required manifests/infrastructure to satisfy all in-cluster component's CredentialsRequests. This has been 
+As an OpenShift cluster administrator I should be able to extract the CredentialsRequest manifests from the release image
+and create required manifests/infrastructure to satisfy all in-cluster component's CredentialsRequests. This has been
 already [implemented](https://github.com/openshift/cloud-credential-operator/blob/master/docs/ccoctl.md).
 
 #### Story 2
-As an OpenShift cluster administrator, I should be able to reconcile my credentials manually using local CLI tooling on 
+As an OpenShift cluster administrator, I should be able to reconcile my credentials manually using local CLI tooling on
 an on-going basis. This is in [backlog](https://issues.redhat.com/browse/CCO-106) and targeted for 4.9.
 
 #### Story 3
-As an OpenShift cluster administrator, I should be able to prep credentials for a cluster upgrade. This is in [backlog](https://issues.redhat.com/browse/CCO-106) 
+As an OpenShift cluster administrator, I should be able to prep credentials for a cluster upgrade. This is in [backlog](https://issues.redhat.com/browse/CCO-106)
 and targeted for 4.9.
 
 ### Risks and Mitigations
 
 ## Design Details
 
-We intend to build an optional tool **ccoctl** which will handle credentials management of the cluster in *manual* mode. 
-The following is the set of requirements for the current prototype (for AWS). We can have something similar for other 
+We intend to build an optional tool **ccoctl** which will handle credentials management of the cluster in *manual* mode.
+The following is the set of requirements for the current prototype (for AWS). We can have something similar for other
 platforms.
 
 * ccoctl should be able to setup Identity Provider to authenticate OpenShift components
-* ccoctl should be able to take a list of CredentialsRequests from the release image and create/update 
+* ccoctl should be able to take a list of CredentialsRequests from the release image and create/update
   Roles/Users/ServicePrincipals/ServiceAccounts with appropriate permissions.
-* ccoctl should be able to take list of CredentialsRequests for a release image, and the Identity Provider URL to generate 
+* ccoctl should be able to take list of CredentialsRequests for a release image, and the Identity Provider URL to generate
   the objects that need to be passed to the installer for successful installation
 * ccoctl should be able to delete all the resources that it had created in the cloud
 
-We envision **ccoctl** as a recommended tool to setup credentials on new platforms but customers are free to use other 
+We envision **ccoctl** as a recommended tool to setup credentials on new platforms but customers are free to use other
 tools like Terraform/AWS CloudFormation to do the above mentioned tasks. Read more about AWS implementation details [here](https://github.com/openshift/cloud-credential-operator/blob/master/docs/ccoctl.md)
 
 We also plan to have a detailed documentation in place to guide new cloud providers to implement *manual* mode.
@@ -97,7 +97,7 @@ To upgrade a cluster we need to execute following steps:
   have changed.
 * Create/update credentials in the underlying cloud provider, and also create/update Kubernetes Secrets in the correct
   namespaces to satisfy all CredentialsRequests in the new release.
-* CCO sets `Upgradeable=False` when in manual mode. Set an appropriate annotation `cloudcredential.openshift.io/upgradeable-to` 
+* CCO sets `Upgradeable=False` when in manual mode. Set an appropriate annotation `cloudcredential.openshift.io/upgradeable-to`
   to a new upgradable version to allow upgrade.
 
 All the above task will also be executable by *ccoctl*. Details in [this](https://issues.redhat.com/browse/CCO-84) card.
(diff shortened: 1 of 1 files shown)


enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,129 @@
...
+
+### User Stories
+
+#### Story 1
+As a OpenShift cluster administrator I should be able to extract the CredentialsRequest manifests from the release image
+and create credentials in the underlying cloud provider. 
+
+#### Story 2
+As a OpenShift cluster administrator I should be able to extract the CredentialsRequest manifests from the release image 
+and create Kubernetes Secrets in to satisfy all CredentialsRequests.
```
joelddiaz, line 66:
```suggestion
and create Kubernetes Secrets in the cluster to satisfy all in-cluster component's CredentialsRequests.
```

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,129 @@
...
+### Risks and Mitigations
+
+## Design Details
+
+We intend to build an optional tool **ccoctl** which will handle credentials management of the cluster in *manual* mode. 
+The following is the set of requirements for the current prototype (for AWS). We can have something similar for other 
+platforms.
+
+* ccoctl should be able to setup Identity Provider to authenticate OpenShift components
+* ccoctl should be to take a list of CredentialsRequests from the release image and create/update Roles with
```
joelddiaz, line 81:
```suggestion
* ccoctl should be to take a list of CredentialsRequests from the release image and create/update Roles/Users/ServicePrincipals/ServiceAccounts with 
```

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,129 @@
...
+## Design Details
+
+We intend to build an optional tool **ccoctl** which will handle credentials management of the cluster in *manual* mode. 
+The following is the set of requirements for the current prototype (for AWS). We can have something similar for other 
+platforms.
+
+* ccoctl should be able to setup Identity Provider to authenticate OpenShift components
+* ccoctl should be to take a list of CredentialsRequests from the release image and create/update Roles with 
+  appropriate permissions.
+* ccoctl should be to take list of CredentialsRequests for a release image, and the Identity Provider URL to generate
```
joelddiaz, line 83:
```suggestion
* ccoctl should be able to take list of CredentialsRequests for a release image, and the Identity Provider URL to generate 
```

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,129 @@
...
+
+We currently have a work-in-progress CLI tool [ccoctl](https://github.com/openshift/cloud-credential-operator/blob/master/docs/ccoctl.md) 
+to create and manage cloud credentials outside the cluster for AWS cloud. The design details of this tool is discussed above.
+
+## Drawbacks
+
+* Taking the credentials management outside the cluster will create additional overhead for the customer to make sure all
+  the required infrastructure is in place before starting the installation process. Current tooling we have only supports
+  the AWS cloud, we do not have anything planned for other cloud providers.
+* Push-button upgrades will not work in *manual* mode.
```
joelddiaz, line 124:
```suggestion
* Push-button upgrades will not work in *manual* mode as the cluster no longer has the powerful credentials to self-manage credentials. 
```

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The intent of this enhancement is to take the process of credentials management outside the OpenShift cluster for new 
+platforms. We propose to make *manual* mode as default for clusters on new platforms.
```
dgoodwin, line 31: Optionally supported by CLI tooling. (ccoctl)

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+
+The main motivation behind this enhancement is to satisfy the customer requirement to follow the best security practices 
+by not storing admin credentials inside the cluster
+
+### Goals
+
+As part of this enhancement we plan to do the following:
+* Set *manual* as default/preferred credentials mode for OpenShift
+* Not store admin level cloud credentials inside a cluster
+* Provide a tooling design for credentials management outside the cluster
```
dgoodwin, line 43: Might need to tighten these up a little, I would restate the goals as:

  * Guide new platforms to start with Manual mode rather than Mint mode.
  * Provide optional tooling (ccoctl) for administrators to manage their credentials outside the cluster. (prep for install, reconcile, prep for upgrade)
  * Establish precedent for future direction of prefering Manual over Mint.

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Non-Goals
+
+* We do not plan to go into design details on how this tool will be ported to cloud providers other than AWS.
```
dgoodwin, line 47: Removing Mint mode for any currently supported cloud. (AWS, Azure, GCP)
Changing the default for any provider currently defaulting to Mint mode, though we may revisit in the future.

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+As a OpenShift cluster administrator I should be able to extract the CredentialsRequest manifests from the release image
+and create credentials in the underlying cloud provider. 
+
+#### Story 2
+As a OpenShift cluster administrator I should be able to extract the CredentialsRequest manifests from the release image 
+and create Kubernetes Secrets in to satisfy all in-cluster component's CredentialsRequests.
+
+#### Story 3
+As a OpenShift cluster administrator I should be able to install a OpenShift cluster, without providing admin credentials
+to the installer, by injecting precreated Kubernetes Secrets.
```
dgoodwin, line 70: As an OpenShift cluster administrator, I should be able to reconcile my credentials manually using local CLI tooling on an on-going basis.
As an OpenShift cluster administrator, I should be able to prep credentials for a cluster upgrade.

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+### Open Questions
+
+### Test Plan
+
+We plan to have a e2e test that will externally set up a credentials management infrastructure and then kickstart 
+install in a *manual* mode.
+
+### Graduation Criteria
+
+### Upgrade / Downgrade Strategy
```
dgoodwin, line 101: I think this section is more about how do we roll out this new functionality. In this case it would roll out smooth as we're not changing any behavior on an existing cluster. This is more about new platforms, and future direction of CCO, should we start changing the default behavior and recommended best practice.

The notes below are relevant though and maybe could jump up to the design section.

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+As a OpenShift cluster administrator I should be able to install a OpenShift cluster, without providing admin credentials
+to the installer, by injecting precreated Kubernetes Secrets.
+
+### Risks and Mitigations
+
+## Design Details
+
+We intend to build an optional tool **ccoctl** which will handle credentials management of the cluster in *manual* mode. 
+The following is the set of requirements for the current prototype (for AWS). We can have something similar for other 
+platforms.
```
dgoodwin, line 78: I think we need a plan here for how ccoctl is distributed. I am really wondering if this should be exposed in oc adm.

sjenning:
This doesn't have a lot of detail, but I'm on board with the general idea of pre-creating creds outside the cluster and the cluster not having highly privileged AWS access i.e. IAM. :+1:

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+Currently, *mint* is a default credentials mode for OpenShift. In this mode we run OpenShift installer with an admin 
+level cloud credential. The admin credential is stored in kube-system namespace and then used by the cloud credential 
+operator to process the CredentialRequests in the cluster and create new users with fine-grained permissions.
+The customers have reported that the need to store admin credentials inside the cluster is a major disadvantage. Based 
+on the feedback, we propose to run credentials related setup outside a cluster and then start installation process in a 
+*manual* mode without providing admin credentials to the installer.
+
+### User Stories
+
+#### Story 1
```
staebler, line 60: Are these really three separate stories? They seem more like steps of the same story. In other words, is it meaningful for a user to do Story 1 by itself?

enhancements/installer/credentials-management-outside-openshift-cluster.md
staebler, line 70: How do we envision a user doing this injection? Is it going to be through custom manifests added manually by the user to {install-dir}/manifests? I am concerned about the complexity of our recommended flow. Is it practical for the installer to do the job of the `ccoctl` tool?

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+As a OpenShift cluster administrator I should be able to extract the CredentialsRequest manifests from the release image
```
staebler, line 61: Is the `ccoctl` tool going to be tied to a release image? Or will the user need to know the release image?

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+
+We currently have a work-in-progress CLI tool [ccoctl](https://github.com/openshift/cloud-credential-operator/blob/master/docs/ccoctl.md) 
+to create and manage cloud credentials outside the cluster for AWS cloud. The design details of this tool is discussed above.
+
+## Drawbacks
+
+* Taking the credentials management outside the cluster will create additional overhead for the customer to make sure all
+  the required infrastructure is in place before starting the installation process. Current tooling we have only supports
+  the AWS cloud, we do not have anything planned for other cloud providers.
+* Push-button upgrades will not work in *manual* mode as the cluster no longer has the admin credentials to mint credentials
```
staebler, line 124: Is there a way that push-button upgrades could work for upgrades where there are no changes in CredentialsRequests?

staebler, line 124: I'm sure that this has been discussed at length, but this is a huge drawback. Going forward we would be saying that, by default, we do not support push-button upgrades.

enhancements/installer/credentials-management-outside-openshift-cluster.md
dgoodwin: We expect that it will be tied to a release image and extracted with oc release commands. I don't think it will be tightly coupled in that it would often work with future releases fine.

enhancements/installer/credentials-management-outside-openshift-cluster.md
dgoodwin: Yes manual mode today involved injecting via custom manifests, but ccoctl will generate everything for you and drop the files if desired.

Integrating right into openshift-install is interesting. If that is on the table that could be very useful, but we'd still want a ccoctl (or oc adm?) entrypoint for on-going maintenance I think.

enhancements/installer/credentials-management-outside-openshift-cluster.md
dgoodwin: As designed today, no, you will always need to indicate that your cluster has been prepped for the next upgrade. We wanted users who opt into Manual mode to be familiar with this flow so it was decided this would be better for now.

enhancements/installer/credentials-management-outside-openshift-cluster.md
staebler: That doesn't quite answer my (poorly phrased) question. When using `ccoctl`, is the expectation that the user would need to specify the release image from which to get the CredentialsRequests?

enhancements/installer/credentials-management-outside-openshift-cluster.md
staebler: Yes, for on-going maintenance. But it seems like a much better user experience to have the installer do the ccoctl work during install, rather than the preferred installation path requiring multiple steps with multiple tools.

enhancements/installer/credentials-management-outside-openshift-cluster.md
staebler: Are the "users who opt into Manual mode" the users with existing clusters?

staebler: That part that I am tripping up on is that the second sentence of the summary of this enhancement is "We propose to make *manual* mode as default for clusters on new platforms." That takes away the "opt-in" aspect of Manual mode.

enhancements/installer/credentials-management-outside-openshift-cluster.md
dgoodwin: Ah yes, for install prep yes you will need to refer to the release image with the cred requests you will need. I am hoping we can run day 2 and work against the CredRequests in a live cluster/current kube context to help users reconcile.

enhancements/installer/credentials-management-outside-openshift-cluster.md
dgoodwin: We would not change anything automatically on anyone, but yes I believe a current cluster admin could flip their cluster to manual mode if they took explicit action to do so.

dgoodwin: That's a good point, this is what we are proposing for a new cloud provider we expect to be added. All clusters on that platform would initially be in this mode. Mint may never happen on that platform, we're not sure.

Many providers today are passthrough only which is technically masking some of the problems we've tried to solve with manual mode, where nothing is actively preventing you from upgrading when your passthrough shared creds are lacking some new permission. So we're not forcing them to take action for upgrades but at the cost of potential breakage. I think these cloud providers tend to use pretty broad permissions so I don't recall seeing it bite anyone yet.

So I guess here, for new platforms, we are proposing that this is now the default, you do not have automated 4.x to 4.x+1 upgrades, you'll need to take action manually to clarify that you've prepared for it.

It should be noted that this does not apply to .z upgrades, 4.7.1 to 4.7.2 is automated without user interaction.

joelddiaz: I don't know if this helps: the push-button upgrades are available as long as the cluster admin sets the appropriate annotation. Once the annotation is in place, the upgrade button will be available (whether the user has actually completed the necessary work will show up during/post upgrade).

The new requirement for doing some pre-upgrade steps is a direct consequence of the decision to move cloud/platform credentials management out of the cluster. Given this new directive, we're left with something needing to tell the cluster that is okay to perform an upgrade (since it can no longer self-service). We could encode this in the release image so that CCO/CVO can peek into the next version to know about the upgradeability...but I'm just brainstorming here and I don't think we've spent much time considering how to improve this new upgrade experience.

enhancements/installer/credentials-management-outside-openshift-cluster.md
derekwaynecarr: @dgoodwin +1

also add "that follows the best practices of the hosting infrastructure at that time."

enhancements/installer/credentials-management-outside-openshift-cluster.md
derekwaynecarr: 100% agree with above.

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+
+### User Stories
+
+#### Story 1
+As a OpenShift cluster administrator I should be able to extract the CredentialsRequest manifests from the release image
+and create credentials in the underlying cloud provider. 
+
+#### Story 2
+As a OpenShift cluster administrator I should be able to extract the CredentialsRequest manifests from the release image 
+and create Kubernetes Secrets in to satisfy all in-cluster component's CredentialsRequests.
```
derekwaynecarr, line 66: required for a specific desired installation topology.

I anticipate we evolve https://github.com/openshift/enhancements/blob/master/enhancements/single-node/cluster-high-availability-mode-api.md#cluster-operators

where `controlPlaneTopology` will have an option for "External"

in that case the permissions are only pertinent for what runs within the cluster itself.

enhancements/installer/credentials-management-outside-openshift-cluster.md
derekwaynecarr: Agree with above for credentials that are scoped to a single cluster.

enhancements/installer/credentials-management-outside-openshift-cluster.md
derekwaynecarr: i assume distribution of `ccoctl` is specific to an enhancement around that component.  this enhancement is more focused on 'should each cluster be empowered to mint scoped cloud credentials' as a default posture moving forward.

enhancements/installer/credentials-management-outside-openshift-cluster.md
derekwaynecarr: can we separate patch release upgrades from minor release upgrades in this flow?

it is a _good thing_ that we are clear with our users on the permissions we require when they differ across releases.

is it possible to integrate whatever checking is required direct in the cvo itself when it evaluates the payload for other characteristics?

enhancements/installer/credentials-management-outside-openshift-cluster.md
dgoodwin: @staebler do we want to outline how that could work in this enhancement or leave it for later? If so, how could it work? :) Are you thinking new openshift-install subcommand, something integrated into installconfig, or other?

staebler: I would think that it would be the default behavior of the installer. There would not be a sub-command. If the user wanted to opt out of having the installer creates the roles and secret manifests, then there would be an option in the install config to tell the installer to not create them. Could that be a new value for `credentialsMode`?

akhil-rane: IIUC same thing is happening when we set `credentialsMode` to manual. We might need to add a different (default) mode where in installer does ccoctl work as part of install.

enhancements/installer/credentials-management-outside-openshift-cluster.md
akhil-rane: @LalatenduMohanty @wking WDYT about including this logic in CVO?

enhancements/installer/credentials-management-outside-openshift-cluster.md
wking: +1, I agree that this seems like a single "extract and locally reconcile" use-case.  Probably also need an "and destroy when I'm done with them" tail for that use case.

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+
+We also plan to have a detailed documentation in place to guide new cloud providers to implement *manual* mode.
+
+To upgrade a cluster we need to execute following steps:
+* Examine the CredentialsRequests in the new OpenShift release. Check if permissions in the existing CredentialsRequest
+  have changed.
+* Create/update credentials in the underlying cloud provider, and also create/update Kubernetes Secrets in the correct
+  namespaces to satisfy all CredentialsRequests in the new release.
+* Set an appropriate annotation `cloudcredential.openshift.io/upgradeable-to` to a new upgradable version.
+
```
LalatenduMohanty, line 98: Can we add the exact steps for installation and upgrades (keeping the POC for AWS in mind) here?

LalatenduMohanty: It will give more clarity around how the user experience is changing because of this.

enhancements/installer/credentials-management-outside-openshift-cluster.md
LalatenduMohanty: I understand the need for manual mode from customers. But are we sure we want to make it the default? Why not keeping mint mode defult but manual mode for customer who want this? May be this has been discussed before.  I am asking automatic upgrade was part of OCP/OTA plan and making the manual mode as default is a big change. I hope product management has already signed-off on this. 

I agree that we should separate Y stream updates from Z stream updates. We should keep the automatic updates enabled for Z stream updates if that is possible by changing CVO  to validate if the cluster has right permissions, I am fine with the idea. 

We would like to know more about the checks we need to add to CVO and then we can say more clearly if it will be feasible with respect to CVO's architecture.

enhancements/installer/credentials-management-outside-openshift-cluster.md
joelddiaz: there is no requirement to use `ccoctl`, but with the tool it would look like:
oc adm release extract --credentials-requests --cloud=aws --to=/path/to/extract (for the release image being upgraded to)
ccoctl aws create-iam-roles --region=us-east-1 --credentials-requests-dir=/path/to/extract --name "some-name-here-used-to-name-iam-roles" --identity-provider-arn=ARN:for_AWS_IAM_IDENTITY_PROVIDER_HERE --output-dir=/path/to/manifests
oc apply -f (for each file in /path/to/manifests)
oc patch (to apply the upgradeable-to annotation)

enhancements/installer/credentials-management-outside-openshift-cluster.md
joelddiaz: The idea is that the OpenShift product will get out of the business of self-managing cloud permissions/credentials. Mint mode would not be implemented today based on this directive. And based on this directive, supporting AWS Role-based/Pod Identity with short lived tokens (STS) is done outside of the cluster as CCO was not extended to support minting IAM Roles from inside the cloud-credential-operator.

This does not mean that we will take away Mint mode for the platforms where it exists today.

wking: In the absence of #363, the currently recommended pattern is for information about 4.y to be back ported to 4.(y-1) operators, so they can intelligently say "ah, the current cluster state is not compatible with the information I've been taught about 4.y" and set `Upgradeable=False`.  I'm not in favor about moving that responsibility into the CVO.  For example, under some circumstances, the cred operator might be able to list permissions attached to existing credentials, see that they met the expanded 4.y criteria, and set `Upgradeable=True`.  The CVO will never be able to perform that in-cluster cred permissions listing.  I do really appreciate the just-works nature of mint and passthrough mode.  Currently manual mode is a very small fraction of Telemetry-submitting clusters.  But I also understand that handing clusters admin-level creds is a big ask.  If cluster admins want to manually create expanded creds as needed (and ideally reduce perms if we reduce our asks), I am completely fine with that.  But I would prefer the cred-operator have [read-only perms sufficient to audit the creds vs. the cred-requests][1], regardless of who is granting the creds.  I don't see any reason why cluster admins would be uncomfortable with such a read-only cred, but if and only if they are, the cred operator could fall back to the `upgradeable-to` flow, with `Upgradeable=True` and a reason like `BlindManualMode` with a message like `I have no audit creds, so whether you have credentials in place for 4.(y+1) is entirely on you, cluster admins`.  Even then, the cred operator could go `Upgradeable=False` if a cred request it knew 4.(y+1) would ask for had no associated Secret at all.

[1]: https://github.com/openshift/cloud-credential-operator/blame/517f71efb6c5babddc7207b5e6331f51316f5880/README.md#L148

wking: For cases where the cred-operator lacks a sufficient cred to audit CredRequest Secrets, it would also be good to have CI around each CredRequest to show that the consuming operator complains clearly if the Secret it has is insufficient to perform a task.  That would make life easier on admins in the completely manual mode to identify and recover if/when they flub a Secret.

enhancements/installer/credentials-management-outside-openshift-cluster.md
```diff
@@ -0,0 +1,130 @@
...
+tools like Terraform/AWS CloudFormation to do the above mentioned tasks. Read more about AWS implementation details [here](https://github.com/openshift/cloud-credential-operator/blob/master/docs/ccoctl.md)
+
+We also plan to have a detailed documentation in place to guide new cloud providers to implement *manual* mode.
+
+To upgrade a cluster we need to execute following steps:
+* Examine the CredentialsRequests in the new OpenShift release. Check if permissions in the existing CredentialsRequest
+  have changed.
+* Create/update credentials in the underlying cloud provider, and also create/update Kubernetes Secrets in the correct
+  namespaces to satisfy all CredentialsRequests in the new release.
+* Set an appropriate annotation `cloudcredential.openshift.io/upgradeable-to` to a new upgradable version.
```
sdodson, line 97: Can we please add how this requirement is enforced? CCO sets `Upgradeable=False` whenever the next minor isn't included in the value?

sdodson:
Given this has already been implemented I've asked @akhil-rane to read it over and reconcile and differences between the enhancement and the implementation and then I'll merge this.

enhancements/installer/credentials-management-outside-openshift-cluster.md
akhil-rane: CCO sets `Upgradeable=False` by default in Manual mode. When user is ready to upgrade (after making necessary changes to satisfy credentials requests in next version), they can `cloudcredential.openshift.io/upgradeable-to:<version>` to upgrade

sdodson:
/lgtm cancel
@akhil-rane please see the lint test failures, the linting jobs may be new since this PR was last updated, sorry.

akhil-rane:
@sdodson I think we need LGTM here

PR #710, opened 2021-03-29
New enhancment proposal: forward-to-Loki

alanconway:
Proposal to add loki forwarding to the ClusterLogForwarder.
First draft, see Open Questions and TODO markers.

/assign @jcantrill
/cc @periklis

### DIFF
--- a/enhancements/cluster-logging/forward-to-loki.md
+++ b/enhancements/cluster-logging/forward-to-loki.md
@@ -1,0 +1,255 @@
+# Forward to Loki
+
+## Release Sign-off Checklist
+
+- [X] Enhancement is `implementable`
+- [X] Design details are appropriately documented from clear requirements
+- [X] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Add a new output type to the `ClusterLogForwarder` API to forward logs to a Loki instance.
+An important part of using Loki is choosing the right _Loki labels_ to define log streams.
+We present a default set of Loki labels, and explain how the choice of labels affect performance and queries.
+
+## Motivation
+
+Loki will become our default log store in the near future.
+There are also several use cases for forwarding to an external Loki instance.
+
+### Goals
+
+- Forward logs to a Loki instance
+- Choose a good default set of Loki labels.
+- Provide configuration to control the choice of Loki labels
+- Describe how expected query patterns correspond to Loki labels and JSON payload
+
+### Non-Goals
+
+The following may be covered in future proposals, but are out of scope here:
+
+- Query tools.
+- Optimizing/reducing the JSON payload format.
+- Alternate payload formats.
+- Use of collectors other than `fluentd` (e.g. Promtail, FluentBit)
+
+## Proposal
+
+### Summary of Loki labels
+
+Loki and Elasticsearch take different approaches to indexing and search.
+Elasticsearch does content parsing and indexing _during ingest_ to optimize the _query_ path.
+Loki optimizes the _ingest_ path by using simple key-value `labels` for initial indexing,
+and deferring content parsing until _query_ time.
+
+The benefit is faster ingest, which reduces the risk of dropping logs.
+The trade-off is that need to carefully choose what to use as labels.
+
+For more details:
+- [Guide to labels in Loki](https://grafana.com/blog/2020/08/27/the-concise-guide-to-labels-in-loki/)
+- [Out-of-order streams issue]([https://github.com/grafana/loki/issues/1544)
+
+To summarize the key points:
+- Each unique combination of labels defines an ordered Loki _stream_.
+  - Too many streams degrades performance.
+  - _Cardinality_ refers to the number of unique label combinations.
+- The cardinality of `kubernetes.pod_name` was found to be
+  - too high in clusters with _very_ rapid turn-over of short-lived pods.
+  - acceptable in other (possibly more typical) clusters with high log data rates.
+- Minimize the _total number_ of labels per stream
+  - There is a limit of 30 max, fewer labels == better performance.
+- Log streams must be _ordered by collection timestamp_.
+  - Must not allow distinct nodes with separate clocks to contribute to a single stream.
+
+All logging meta-data will still be included in log records.
+Labels do not need to _identify the source of logs_, only to _partition the search space_.
+At query time labels reduce the search space, then Loki uses log content to complete the search.
+
+**Note**: Loki [label names][] must match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*`.
+Meta-data names are translated to Loki labels by replacing illegal characters with '_'.
+
+Nested JSON objects are referenced by converting their JSON path to a label name.
+For example, if `kubernetes.namespace_name` is a label, then this JSON log record:
+
+``` json
+{"kubernetes.namespace_name":"foo","kubernetes":{"labels":{"foo":"bar"}}}
+```
+would match this LogQL filter:
+
+``` logql
+{ kubernetes_namespace_name="foo" } | json | kubernetes_labels_foo == "bar"
+```
+### Default Loki Labels
+
+The default label set should be suitable for most deployments.
+The user can configure a different set for specific needs.
+
+The default label set is:
+
+`log_type`: One of `application`, `infrastructure`, `audit`. Basic log categorization.
+
+`cluster`: The cluster name, i.e. the DNS authority part of the cluster master URL.\
+This name is DNS-unique, human readable, and known to all k8s API clients.
+It is recommend as a cluster ID by [this discussion](https://github.com/kubernetes/kubernetes/issues/2292).
+To print the name of the connected cluster:
+```bash
+oc config view -o jsonpath='{.clusters[].name}{"\n"}
+```
+
+`kubernetes.namespace_name`: namespace where the log originated.
+
+`kubernetes.pod_name`: name of the pod where the log originated.
+
+`kubernetes.host`: Host name of the cluster node where the log record originated.\
+This is *always included* to guarantee ordered streams, even if the user configures a label set without it.
+
+**Note:** `container_name` and `image_name` are *not* included in the defaults. They are not high-cardinality by themselves, but they are multipliers of `namespace/pod`, which is the highest cardinality we want to allow by default.
+
+**Note:** We use names rather than UID for cluster, namespace and pod.
+Names are more likely to be known and easier to use in queries.
+Using both names *and* UIDs is redundant, they partition the data in approximately the same way.
+
+UIDs (and all other meta-data) are still available in the log payload for filtering in queries.
+
+### Proposed API
+
+Add a new `loki` output type to the `ClusterLogForwarder` API:
+
+``` yaml
+- name: myLokiOutput
+  type: loki
+  url: ...
+  secret: ...
+```
+
+`url` and `secret` have the usual meaning with regard to TLS and certificates.
+The `secret` may also contain `username` and `password` fields for Loki.
+
+The following optional output fields are Loki-specific:
+
+- `labelKeys`: ([]string, default=_see [Default Loki Labels](#default-loki-labels))_ \
+  A list of meta-data keys to replace the default labels.\
+  Keys are translated to [label names][] as described in [Summary of Loki Labels](#summary-of-loki-labels)
+  Example: `kubernetes.labels.foo` => `kubernetes_labels_foo`.\
+  **Note**: `kubernetes.host` is *always* be included, even if not requested.
+  It is required to ensure ordered label streams.
+- `tenantKey`: (string, default=`"kubernetes.namespaceName"`) \
+  Use the value of this meta-data key as the Loki tenant ID.\
+  At least these keys are supported:
+  - `kubernetes.namespaceName`: Use the namespace name as the tenant ID.
+  - `kubernetes.labels.<key>`: Use the string value of kubernetes label with key `<key>`.
+  - `openshift.labels.<key>`: use the value of a label attached by the forwarder.
+
+The full set of meta-data keys is listed in [data model][].
+
+### Implementation Details
+
+The output will be implemented using this fluentd plugin: https://grafana.com/docs/loki/latest/clients/fluentd/
+
+Notes on plug-in configuration:
+
+- Security: configured by the `output.secret` as usual.
+- K8s labels as Loki labels: supported by the plug-in.
+- Always include `kubernetes.host` to avoid avoid out-of-order streams.
+- Tenant set from `output.tenantKey`
+- Output format is `json`, serializes the fluentd record like other outputs.
+- Static labels set as `extra_labels` to avoid extracting from each record.
+
+### User Stories
+
+#### Treat each namespace as a separate Loki tenant
+
+I want logs from each namespace to be directed to separate Loki tenants, using Loki's "soft tenancy" model:
+
+``` yaml
+- name: myLokiOutput
+  type: loki
+  url: ...
+  secret: ...
+  tenantKey: kubernetes.namespace_name
+```
+
+#### Query all logs from a namespace
+
+``` logql
+{ kubernetes_namespace_name="mynamespace"}
+```
+
+#### Query logs from a specific container in a named Pod
+
+``` logql
+{ kubernetes_namespace_name="mynamespace"" } |= kubernetes_pod_name == "mypod" |= kubernetes_container_name="mycontainer
+```
+
+#### Query logs from a labeled application
+
+Using the default configuration:
+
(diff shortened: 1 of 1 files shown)


enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,221 @@
...
+```
+### Default Loki Labels
+
+The proposed default label set is:
+
+- `log_type`: one of "application", "infrastructure", "audit"
+- `cluster`: for muti-cluster deployments (see Open Questions).
+- `kubernetes.host`: aka node, ensures ordered streams.
+- `kubernetes.namespace_name`: likely to be most common known factor in a search.
+- `kubernetes.container_name`: low cardinality, usually well-known
```
eranra, line 93: @alanconway This label is strange to me ... I understand that this is low cardinality ... but on the flip side if there is no indexing to pod name ... why the container name is intersting .... in any event the code need to scan to find all pod names ... so it doesn't sound expensive to me to look for container_name

eranra: @alanconway there are two labels that I was expecting and I do not see in the default list::

(1) `app` which is very common and might allow fast focus on specific application ... this is similar in sense to namespace because this is "logical" label but the granularity of application (i.e. cardinality) is higher. On the other hand the cardinality is limited to the applications in the cluster so no reason to expect huge number of those
 
(2) `tier` same as (1) with a bit higher cardinality maybe a bit similar to `container_name` but in the logical space and not in the physical space

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,221 @@
...
+  - Must not allow distinct nodes with separate clocks to contribute to a single stream.
+
+All logging meta-data will still be included in log records.
+Labels do not need to _identify the source of logs_, only to _partition the search space_.
+At query time labels reduce the search space, then Loki uses log content to complete the search.
+
+**Note**: Loki [label names][] must match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*`.
+Meta-data names are translated to Loki labels by replacing illegal characters with '_'.
+
+Nested JSON objects are referenced by converting their JSON path to a label names.
```
syedriko, line 74: to label names (spelling)

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,221 @@
...
+``` logql
+{ kubernetes_namespace_name="foo" } | json | kubernetes_labels_foo == "bar"
+```
+### Default Loki Labels
+
+The proposed default label set is:
+
+- `log_type`: one of "application", "infrastructure", "audit"
+- `cluster`: for muti-cluster deployments (see Open Questions).
+- `kubernetes.host`: aka node, ensures ordered streams.
```
syedriko, line 91: I'm not sure I see how it does that

syedriko:
@alanconway Could you fix the spelling of the title and the syntax so there's no red background?

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,221 @@
...
+``` logql
+{ } |= kubernetes_labels_app == "myapp""
+```
+
+By adding `labelKeys: [ kubernetes.labels.app ]` to the Loki output configuration this
+becomes a faster label query:
+
+``` logql
+{ kubernetes_labels_app="myapp" }
+```
```
syedriko, line 151: I'm not seeing how assigning labels to a Loki output assigns them to a particular application.

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,221 @@
...
+- Describe how expected query patterns correspond to Loki labels and JSON payload
+
+### Non-Goals
+
+The following may be covered in future proposals, but are out of scope here:
+
+- Query tools.
+- Optimizing/reducing the JSON payload format.
+- Alternate payload formats.
+- Use of collectors other than `fluentd` (e.g. promtail, fluentbit)
```
syedriko, line 37: I'm figuring we're talking about https://grafana.com/docs/loki/latest/clients/fluentd/, but may be it's worth mentioning explicitly.

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,221 @@
...
+To summarize the key points:
+- Each unique combination of labels defines an ordered Loki _stream_.
+  - Too many streams degrades performance.
+  - _Cardinality_ refers to the number of unique label combinations.
+- The cardinality of `kubernetes.pod_name` was found to be
+  - too high in clusters with _very_ rapid turn-over of short-lived pods.
+  - acceptable in other (possibly more typical) clusters with high log data rates.
+- Minimize the _total number_ of labels per stream
+  - There is a limit of 30 max, fewer labels == better performance.
+- Log streams must be _ordered by collection timestamp_.
```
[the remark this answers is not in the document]
cyriltovena: 👋 there !

Loki contributor here :)

When timestamp are equal Loki keep the natural order of insertion.

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,221 @@
...
+  - Must not allow distinct nodes with separate clocks to contribute to a single stream.
```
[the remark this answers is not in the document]
cyriltovena: I think that's a correct assumption !

enhancements/cluster-logging/forward-to-loki.md
alanconway: > @alanconway This label is strange to me ... I understand that this is low cardinality ... but on the flip side if there is no indexing to pod name ... why the container name is intersting .... in any event the code need to scan to find all pod names ... so it doesn't sound expensive to me to look for container_name

Good question: I'm uncertain about it. The idea is that container name is a constant for pods created by deployments, even when the pod names are randomized. So it might be useful to locate pods that are part of a multi-namespace application. I think it needs a bit more justification though based on looking at real clusters.

alanconway: > @alanconway there are two labels that I was expecting and I do not see in the default list::
> 
> (1) `app` which is very common and might allow fast focus on specific application ... this is similar in sense to namespace because this is "logical" label but the granularity of application (i.e. cardinality) is higher. On the other hand the cardinality is limited to the applications in the cluster so no reason to expect huge number of those

Again you spot my every uncertainty! First draft of this did include kubernetes.labels.app and kubernetes.labels.deploymentconfig in the defaults. However they can be enabled using `labelKeys` so I took them out.
I can probably be argued into putting "app" back  into defaults, as it is a crucial one for cross-namespace apps.

> (2) `tier` same as (1) with a bit higher cardinality maybe a bit similar to `container_name` but in the logical space and not in the physical space

What's "tier"?

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,221 @@
...
+The API below allows the user to add labels that are relevant for their application.
+
+### Proposed API
+
+Add a new `loki` output type to the `ClusterLogForwarder` API:
+
+``` yaml
+- name: myLokiOutput
+  type: loki
+  url: ...
```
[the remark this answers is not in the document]
alanconway: I disagree. The syntax of an "endpoint" in a configuration file is maddeningly vague, some want "host:port", some want {host: ..., port: ...} some will accept ipv6 literals without square brackets, some won't etc. Any attempt to be "flexible" about the syntax will run aground somewhere (been there, done that) By clearly stating that we want a URL, there is a well known public standard specifying exactly what that means.

enhancements/cluster-logging/forward-to-loki.md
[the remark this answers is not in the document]
alanconway: Yup - the problem here is only if we send records with _different_ timestamps that are out of order. @cyril-ui-developer please correct me but as I understand it these are timestamps added by the _collector_  - not timestamps in the original log text. If I read it all right, as long as we have a collecter-per-node, and we don't have any streams that receive logs from 2 separate nodes, we are OK.

enhancements/cluster-logging/forward-to-loki.md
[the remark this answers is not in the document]
alanconway: A loki stream is defined as a unique combination of label values - it's a "stream" on the database side, not on the collector side.  That is why we need to include a node-id or node hostname in our label set (if we don't include a pod name/id), to ensure that we cannot have Loki streams that get contributions from multiple nodes.

enhancements/cluster-logging/forward-to-loki.md
alanconway: It ensures that there is no combination of label:value pairs that is the same for 2 different nodes. If it were then we potentially get mis-ordered timestamps when they are aggregated at the database.

enhancements/cluster-logging/forward-to-loki.md
eranra: @alanconway my mistake ... I should have used the correct label which is `app.kubernetes.io/component` but actually I assume your answer to that and thoughts will be similar that this can be part of default but maybe better to get more experiance and for now can be added as part of the configuration

enhancements/cluster-logging/forward-to-loki.md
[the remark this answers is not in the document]
eranra: ok

enhancements/cluster-logging/forward-to-loki.md
periklis: @alanconway What do you think of putting the pod name in `labelKeys` per default, i.e. `labelKeys = []string{"kubernetes_pod_name"}`. Users will basically have it in default installation and have the choice to remove it later, if they have more flaky workloads, disruptive environment or more batch-like workloads (like DPTP).

alanconway: @periklis that's the other thing I'm  uncertain of. Loki guys also suggested this. Main considerations I see:

1. We know it was a problem for DPTP, we _believe_ DPTP is unusual. Loki folks claim to have included it without problems but we don't have details of those clusters.
2. When searching for a named pod, you clearly know the namespace, so you're only searching per namespace.
3. For post-mortem bug-hunts, a responsive search for specific pods is important.
4. For routine cluster monitoring, specific pod search is not important at all,  you want big cluster-wide searches for specific log types, severities or contents.

So: do we let people opt in to improve performance, or tell them to opt out after a disaster? I think you have the best grasp of the trade-off, but I'd love to have numbers. In particular I guess that including pod name is good for 3. but is it bad for 4.? What are your thoughts?

enhancements/cluster-logging/forward-to-loki.md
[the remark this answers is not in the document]
alanconway: Also node is a nice way to evenly partition the search space. It is not something a user will normally know when constructing a query, but it _is_ something we know when we try to correlating logs with metrics (and maybe traces, haven't got that far) 

@bwplotka -  FYI, you may want to review this also.

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,221 @@
...
+  - too high in clusters with _very_ rapid turn-over of short-lived pods.
+  - acceptable in other (possibly more typical) clusters with high log data rates.
+- Minimize the _total number_ of labels per stream
+  - There is a limit of 30 max, fewer labels == better performance.
+- Log streams must be _ordered by collection timestamp_.
+  - Must not allow distinct nodes with separate clocks to contribute to a single stream.
+
+All logging meta-data will still be included in log records.
+Labels do not need to _identify the source of logs_, only to _partition the search space_.
+At query time labels reduce the search space, then Loki uses log content to complete the search.
```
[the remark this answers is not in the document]
alanconway: Added tenancy controls to the API, would like your opinin @periklis

alanconway:
@periklis see https://github.com/openshift/enhancements/pull/710#discussion_r605919288 for discussion on pod name. I'm still undecided, and I'm hoping you can help me decide.

enhancements/cluster-logging/forward-to-loki.md
cyriltovena: I would prefer that people opt out. The problem DPTP had is something we'll address over time.

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,235 @@
...
+  url: ...
+  secret: ...
+```
+
+`url` and `secret` have the usual meaning with regard to TLS and certificates.
+The `secret` may also contain `username` and `password` fields for Loki.
+
+The following optional fields are Loki-specific:
+
+- `labelKeys`: ([]string, optional)  A list of meta-data keys to add to the label set.\
```
eranra, line 116: @alanconway taking into account that we do not know exactly the correct defaults etc ... why there is just an option to "add" keys to the list ... maybe we also need to support "remove" default keys from the list ... actually ,... you can change the definition of "lablekeys" to overwrite the list of kesy to custom list and allow users to do whatever they want with the keys

enhancements/cluster-logging/forward-to-loki.md
periklis: @alanconway I am pro "opt-out" without having numbers to prove this. The pod name sets from a user experience a high expectation. The user could achieve the same with the cluster node and the container name assuming bin packing is biting (having 2 replicas of the container on the same node). 

@cyriltovena Would you mind to share some insights from your experience if possible, e.g.
- The amount of pod name labels used in various use cases (.g. 0.5TB/day, 1TB/day, 2TB/day ingestion)
- The impact of this pod name label volumes on botldb index size?

enhancements/cluster-logging/forward-to-loki.md
periklis: @eranra I believe the schema here is:
- The Cluster Logging stack decides a bare minimum set of labels to be always present to support use cases in the OCP console, e.g. pod detail log view.
- The user can add more labels to aggregate them in the index if he desires so. Removing labels from his/her configuration is as simple as dropping them from `labelKeys` again.

eranra: @periklis hence my point ... do we really need to force customers to that required list of labels ... if so, the list should really be bare minimum set of labels... and I argue that we want to set as default the recommended set of labels and not minimum required … so we want to allow customers to change labels that are recommended and not required

periklis: @eranra I don't believe we will ever expose our minimum set of labels in `labelKeys`. Latter is only to add additional labels from scraped kubernetes labels field. We will certainly not recommend anything here except the pod name.

alanconway: @eranra  I have re-worded so that the user can _replace_ the default label set (e.g. not include pod name, or use UIDs instead of names) with the exception of the host name, which is mandatory to ensure ordered streams.

alanconway:
Updated to include pod name by default, remove container name (see text for explanation) and allow the user to replace the default label set instead of  just add to it.
Sombody gimme an LGTM :)

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,249 @@
...
+At query time labels reduce the search space, then Loki uses log content to complete the search.
+
+**Note**: Loki [label names][] must match the regex `[a-zA-Z_:][a-zA-Z0-9_:]*`.
+Meta-data names are translated to Loki labels by replacing illegal characters with '_'.
+
+Nested JSON objects are referenced by converting their JSON path to a label name.
+For example, if `kubernetes.namespace_name` is a label, then this JSON log record:
+
+``` json
+{"kubernetes.namespace_name":"foo","kubernetes":{"labels":{"foo":"bar"}}
```
Red-GV, line 78:
```suggestion
{"kubernetes.namespace_name":"foo","kubernetes":{"labels":{"foo":"bar"}}}
```

(Prevents the readme from having a bunch of red highlights everywhere on GitHub)

enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -0,0 +1,249 @@
...
+
+_**TODO**: Review Loki client configuration for any other fields._
+
+### Implementation Details
+
+The output will be implemented using this fluentd plugin: https://grafana.com/docs/loki/latest/clients/fluentd/
+
+### User Stories
+
+_**TODO**: Examples of setting tennant IDs._
```
Red-GV, line 152:
```suggestion
_**TODO**: Examples of setting tenant IDs._
```

enhancements/cluster-logging/forward-to-loki.md
cyriltovena: I have a cluster that supports 13k pod names (consistent) for 10TB per day. The stream churns for that cluster over a day is 2millions. (stream added + stream removed).

Again we are working at improving everyday on that matter.

alanconway: It has been updated to be opt-out.

PR #711, opened 2021-03-30
olm: add "fail open" risk/mitigation

njhale:
Add a risk/mitigation and the decision around it as informed by @bparees ' discussion with stakeholders (OTA).

### DIFF
--- a/enhancements/olm/max-openshift-versions-for-operators.md
+++ b/enhancements/olm/max-openshift-versions-for-operators.md
@@ -9,7 +9,7 @@ approvers:
   - "@ecordell"
   - "@spadgett"
 creation-date: 2021-01-11
-last-updated: 2021-01-19
+last-updated: 2021-03-29
 status: provisional
 see-also:
   -  N/A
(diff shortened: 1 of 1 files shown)


enhancements/olm/max-openshift-versions-for-operators.md
```diff
@@ -186,8 +186,11 @@ As the index maintainer, I want to be able to dynamically change which `{major}.
...
-In extreme cases, cluster admins can override the CVO's upgrade checks via means documented elsewhere.
+  - This risk is mitigated given that admins will still be able to apply security fixes with patch updates whether or not OLM is blocking Minor OpenShift Upgrades. This gives them a safe window to either remove the minor-blocking operator or update it to a version that is compatible with the next minor OpenShift version.
+  - In extreme cases, cluster admins can override the CVO's upgrade checks via means documented elsewhere.
+- "Failing open", by assuming operators that don't specify their `maxOpenShiftVersion` are compatible with the next minor release of OpenShift, runs the risk of breaking installed operators
+  - ~~This could be mitigated by surfacing warnings for installed operators with unspecified compatability~~
+  - A mitigation was deemed unnecessary by stakeholders (OTA team)
```
bparees, line 193: this also closes out the "open question":

```
When a CSV is present that does not include a maxOpenShiftVersion annotation, how will this information be surfaced by the UI and CLI?
```

guess we're saying it won't be surfaced, other than via the status condition message text.

njhale:
We can add this to a larger EP update.

PR #712, opened 2021-03-30
[cluster-logging] Add proposal for loki as an alternative log store

periklis:
This PR provides the enhancement proposal to add a new API to manage Loki as an alternative log store in cluster logging.

/cc @ewolinetz @blockloop @alanconway @jcantrill @lukas-vlcek @Red-GV @sasagarw @sichvoge @bbrowning 
/assign @ewolinetz @alanconway

### DIFF
--- a/enhancements/cluster-logging/loki-storage.md
+++ b/enhancements/cluster-logging/loki-storage.md
@@ -1,0 +1,623 @@
+---
+title: Cluster Logging Loki as an alternative Log Store
+authors:
+  - "@periklis"
+reviewers:
+  - "@ewolinetz"
+  - "@blockloop"
+  - "@lukas-vlcek"
+  - "@sichvoge"
+approvers:
+  - "@ewolinetz"
+  - "@alanconway"
+  - "@jcantril"
+creation-date: 2021-03-29
+last-updated: 2021-03-29
+status: provisional
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Cluster Logging: Loki as an alternative Log Store
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The purpose of this is to provide an alternative low-retention log store based on [Loki](https://github.com/grafana/loki) for the cluster logging stack. This aims to provide a cloud-native first solution specialized in storing logs and in turn lessen the operational cost running cluster logging.
+
+Currently our main log store, i.e. Elasticsearch, is a more general-purpose ingestion, indexing and querying cluster-based solution for text search. It is a powerful solution with fine-grained control on how to scale log storage, indexing and querying.
+
+In general it serves its purpose quite well but it is a constant source of labor regarding operations. Its powerfulness and complexity requires a constant toll on the OpenShift cluster administrator side to understand how to individually scale the Elasticsearch cluster as a log store.
+
+Representative actions that an OpenShift cluster administrator needs to figure out besides controlled operations by [elasticsearch operator](https://github.com/openshift/elasticsearch-operator):
+- Regular evaluation and balance of the CPU/Memory requests and limits because configuration for ingestion and query paths is not individual configurable.
+- Regular evaluation and adaption of redundancy effects on memory and persistent volume usage.
+- Usage of `managementState: Unmanaged` to unblock the cluster in situations of bugs, e.g. locked indices, shard allocation.
+- Usage of the Elasticsearch cluster API to tune the cluster for options not supported in the Elasticearch custom resource.
+
+On the other hand and based on Openshift Logging team's experience operating Loki as an internal central log aggregation solution, we found that a more specialized solution preserves the same log ingestion/querying user experience while lowering the human operational burden. Representative characteristics that showcase the improvements:
+- All components are stateless and can be scaled out individually based on load scenarios.
+- Separate scale out for the ingestion and querying path possible.
+- Includes rate limiting for fine grained ingestion and query control.
+- Persists logs on object storage solutions only, regardless redundancy setting.
+- No individual control of any cluster API to tune any operations.
+- Provides consistent schema-less model compatible with the rest of our observability tools, i.e. Prometheus, Thanos.
+
+In summary offering an alternative low-retention store based on Loki in cluster logging provides many new benefits in lifting the whole stack to a more cloud-native first experience. This has the potential to lessen the operational and the support burden by a lot.
+
+## Motivation
+
+Currently cluster logging is restricted to providing support for Elasticsearch as a log store. This proposal outlines the API in form of a custom resource definition (CRD) for an alternative log store based on Loki and managed by the [loki-operator](https://github.com/viaq/loki-operator).
+
+In addition it describes how to integrate the custom resource definition into the ClusterLogging custom resource that serves as the single entry point for installing the whole cluster logging stack.
+
+### Goals
+
+The specific goals of this proposals are:
+
+* Outline the API to provide a declarative CRD for Loki deployments managed by the [loki-operator](https://github.com/viaq/loki-operator).
+
+* Propose API support into the ClusterLogging CRD for single entry point deployments.
+
+We will be successful when:
+
+* Users are able to declare a Loki instance via the CRD and the [loki-operator](https://github.com/viaq/loki-operator) rolls out the Loki deployment.
+
+* Users are able to declare Loki as alternative store in the Cluster Logging custom resource.
+
+* Users are able to operate Loki in Cluster Logging without the need to tune every single bit via the proposed CRD.
+
+### Non-Goals
+
+* We will not provide a general purpose API to install Loki at the field. The API is dedicated to cluster logging only.
+
+* We will not expose or implement all possible Loki configuration knobs in the API besides the most crucial to operate in cluster logging.
+
+* We will not expose Loki as a general log store for other ingestion sources than the ones supported and managed by cluster logging.
+
+* We will not offer support for custom labeling of logs other than the defaults provided by the cluster logging collector.
+
+## Proposal
+
+To offer a seamless cluster logging experience with Loki as an alternative store, we propose to limit the API support to the following key features:
+
+* Deploy Loki **only** in microservices mode, i.e. each component is deployed and managed as an individual deployment/statefulset. This enables individual capacity planning and resource utilization per path ingestion/querying.
+
+* Multi-tenancy is enabled by default and requires the collector to apply the required [tenant identification](https://grafana.com/docs/loki/latest/operations/multi-tenancy/) on ingestion time. No user can disable this setting. In turn queries require the same identification information to be successful. Furthermore there is no support for querying across tenants.
+
+* Loki is operated only in a zero-dependency setup mode as described [here](https://grafana.com/docs/loki/latest/configuration/examples/#almost-zero-dependencies-setup). The only externally managed requirement is the availability of an S3-compatible API for object storage to persist logs.
+
+* Support a limited set of Loki scale outs, which are battle-tested on our internally managed Loki deployment. The target scale outs are defined using the format `Nx.{extra-small, small, medium}` where `N` the number of Loki deployments:
+  * **1x.extra-small**: Dedicated scale for development and demo purposes only! Not a valid configuration for cluster logging production environments.
+  * **1x.small**: Dedicated scale for low-retention log storage without replication factor support. All components are deployed with HA by running at least two replicas.
+  * **1x.medium**: Dedicated scale for log-retention log storage with single replication factor support. All components are deployed with HA by running at least two replicas. The only exception is the ingester that requires at minimum three replicas to allow failover when single replication factor enabled.
+
+* Support tuning of CPU/Memory resource and limits for each individual component beyond the selected scale-out size.
+
+* Support adding kubernetes tolerations and node selectors for all Loki components individually. This enables operating the Loki deployment on dedicated nodes.
+
+* Support a limited set of control options from Loki's [limit_config](https://grafana.com/docs/loki/latest/configuration/#limits_config) for the ingestion and query paths. The set is an opinionated list of options to enable tuning of object storage usage. The list of options is offered globally across the whole Loki deployment or per tenant.
+
+* Support for replication factor of two for `1x.small` and `1x.medium` by default.
+
+* Support auto-compaction of the index and chunks stored on the object storage when replication factor of two enabled.
+
+For **1x.extra-small** instances only:
+
+* Support for replication factor of one only.
+
+
+### User Stories
+
+* As a cluster admin, I want to select a different scale for my Loki deployment to enable single replication factor and auto-compaction.
+
+* As a cluster admin, I want to expand resources/limits for individual Loki components to support load scenarios not declared by the cluster t-shirt size selected.
+
+* As a cluster admin, I want to configure scheduling of individual Loki components to custom selected nodes on my cluster to separate resource usage of the cluster logging from other workloads on my cluster.
+
+* As a cluster admin, I want to tune the ingestion limits globally or per tenant to support custom ingestion characteristics (e.g. ingenstion rate, label cardinality).
+
+* As a cluster admin, I want to tune the query limits globally or per tenant to support custom query characteristics (e.g. chunks per series returned).
+
+### Implementation Details
+
+#### Assumptions
+
+* The proposed API will be owned and managed exclusively as a custom resource definition by the [loki-operator](https://github.com/viaq/loki-operator).
+
+* The [cluster-logging-operator](https://github.com/openshift/cluster-logging-operator) will only create a stanza of this CRD and in addition manage an owner reference on it.
+
+#### Security
+
+The cluster logging admin is responsible for creating and maintaining the secret in the `openshift-logging` namespace to permit access for the Loki deployment to securely connect to an object storage endpoint.
+
+The secret is required to provide all of the following key-value pairs:
+- `bucketnames`: Comma separated list of bucket names to evenly distribute chunks over.
+- `endpoint`: S3-compatible API Endpoint to connect to.
+- `access_key_id`: Access Key ID.
+- `access_key_secret`: Secret Access Key.
+
+Optional:
+- `region`: Region to use.
+
+### Risks and Mitigations
+
+In general the key risks to provide the proposed API in form of a custom resource definition are:
+
+**Risk:** Providing too many configuration knobs and in turn a brittle API for managing Loki deployments.
+
+**Mitigation:** Limit by maintaining a strong opinionated API for Loki deployments that encompasses only features around dependencies (e.g. object storage), scheduling (e.g. replicas, tolerations/node-selectors) and no more.
+
+**Risk:** Providing poor user experience for day 1 and day 2, e.g. user needs to figure out requests/limits for each component.
+
+**Mitigation:** Provide opinionated and battle-tested scale out formats in form of a T-Shirt size (e.g. `1x.medium`). Leaving room for adding more formats later (e.g. `2x.medium` for hard multi-tenancy) and knobs (e.g. replicas) to adapt the format to live environments.
+
+## Design Details
+
+The discussion about the proposed API is split in two sections. The first section `lokistack_types.go` describes the custom resource definition for the [loki-operator](https://github.com/viaq/loki-operator) to reconcile a Loki deployment.
+
+The API in `lokistack_types.go` includes all required and optional configuration options to support the [proposal](#proposal) above. Furthermore the status struct in `lokistack_types.go` declares conditions for the Loki deployment. The Loki deployment can be under various situations in a degraded state resulting in no deployment.
+
+The second section `clusterlogging_types.go` describes the minimal stanza for translating the `logStore` spec into a `LokiStackSpec`.
+
+### lokistack_types.go
+
+#### Types and constants
+
+A `LokiStack` custom resource can declare one of the following three sizes as values of the `LokiStackSpec`'s field `Size`. Each value has impact to all of the following per Loki component:
+
+1. `Replicas`: The number of replica pods per Loki component. Besides two exceptions all components are deployed with at minimum two replicas. First exception is the `1x.extra-small` resulting in a single replica per component. Second exception is the Loki compactor that is always deployed with a single replica, because multiple replicas are not supported.
+2. `Requests/Limits`: The number of requested CPU (in millicores) and Memory (in Bytes) and their limits per component.
+3. `PodDisruptionBudgets`: The number of minimum available pods per component to ensure normal operations under selected settings, e.g. `ReplicationFactor = 1`.
+
+```go
+// ManagementStateType defines the type for CR management states.
+//
+// +kubebuilder:validation:Enum=Managed;Unmanaged
+type ManagementStateType string
+
+const (
+    // ManagementStateManaged when the LokiStack custom resource should be
+    // reconciled by the operator.
+    ManagementStateManaged ManagementStateType = "Managed"
+
+    // ManagementStateUnmanaged when the LokiStack custom resource should not be
+    // reconciled by the operator.
+    ManagementStateUnmanaged ManagementStateType = "Unmanaged"
+)
+
+// LokiStackSizeType declares the type for Loki deployment scale outs.
+//
+// +kubebuilder:validation:Enum=OneXExtraSmallSize;OneXSmall;OneXMedium
+type LokiStackSizeType string
+
+const (
+    // OneXExtraSmallSize defines the size of a single Loki deployment instance
+    // with extra small resources/limits requirements and without HA support.
+    // This size is ultimately dedicated for development and demo purposes.
+    // DO NOT USE THIS IN PRODUCTION!
+    OneXExtraSmallSize LokiStackSizeType = "1x.extra-small"
+
+    // OneXSmall defines the size of a single Loki deployment instance
+    // with small resources/limits requirements and HA support for all
+    // Loki components. This size is dedicated for setup **without** the
+    // requirement for replication factor of two and auto-compaction.
+    OneXSmall LokiStackSizeType = "1x.small"
+
+    // OneXMedium defines the size of a single Loki deployment instance
+    // with small resources/limits requirements and HA support for all
+    // Loki components. This size is dedicated for setup **with** the
+    // requirement for replication factor of two and auto-compaction.
+    OneXMedium LokiStackSizeType = "1x.medium"
+)
+
+```
+
+#### Mandatory: Loki deployment specification
+
+The mandatory fields of a `LokiStack` custom resource declaration are:
+- `Size`: Selection of one of the supported t-shirt size formats (see [Types and constants](#types-and-constants)) to reconcile the `CPU/Memory Requests`, `CPU/Memory Limits`, `Replicas` and `PodDisruptionBudgets` per Loki component.
+- `Storage`: A value of `ObjectStorageSpec` to declare the S3 compatible-API endpoint in a `Secret` located in `openshift-logging` by a OpenShift administrator.
+- `ReplicationFactor`: The replication factor declares how many Loki ingester replicas should process each log stream in parallel. This does not always result in replicated data on the object storage. It is a safety measure to allow failover of ingester replicas because of unexpected disruption (e.g. node rescheduling, cluster upgrades, etc.). In turn `1` can result in data loss.
+
+__Note__: Redundancy of the persistent object storage is configured by the administator per bucket based on the object storage provider policies available (e.g. multi-zone/region buckets).
+
+The optional parts are `Limits` (see discussion in section [Optional: Global/Per-Tenant Limits](#optional-globalper-tenant-limits)) and `Template` (see details in section [Optional: Loki component configuration](#optional-Loki-component-configuration)).
+
+__Note__: The `Template` field if left empty will be populated by the [loki-operator](https://github.com/viaq/loki-operator) on the first reconciliation by translating the `Size` selection. This automatic update is to inform the user about the opinionated defaults for each supported scale out beyond documentation. The user can use the `Template` field to override these defaults.
+
+```go
+// ObjectStorageSecretSpec is a secret reference containing name only, no namespace.
+type ObjectStorageSecretSpec struct {
+    // Name of a secret in the namespace configured for object storage secrets.
+    //
+    // +required
+    Name string `json:"name"`
+}
+
+// ObjectStorageSpec defines the requirements to access the object
+// storage bucket to persist logs by the ingester component.
+type ObjectStorageSpec struct {
+    // Secret for object storage authentication.
+    // Name of a secret in the same namespace as the cluster logging operator.
+    //
+    // +required
+    Secret *ObjectStorageSecretSpec `json:"secret,omitempty"`
+}
+
+// LokiStackSpec defines the desired state of LokiStack
+type LokiStackSpec struct {
+
+    // ManagementState defines if the CR should be managed by the operator or not.
+    // Default is managed.
+    //
+    // +required
+    // +kubebuilder:default:=Managed
+    // +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Management State"
+    ManagementState ManagementStateType `json:"managementState,omitempty"`
+
+    // Size defines one of the support Loki deployment scale out sizes.
+    //
+    // +required
+    Size LokiStackSizeType `json:"size,omitempty"`
+
+    // Storage defines the spec for the object storage endpoint to store logs.
+    //
+    // +required
+    Storage ObjectStorageSpec `json:"storage,omitempty"`
+
+    // ReplicationFactor defines the policy for log stream replication.
+    //
+    // +required
+    ReplicationFactor int32 `json:"replicationFactor,omitempty"`
+
+    // Limits defines the limits to be applied to log stream processing.
+    //
+    // +optional
+    Limits *LimitsSpec `json:"limits,omitempty"`
+
+    // Template defines the resource/limits/tolerations/nodeselectors per component
+    //
+    // +optional
+    Template *LokiTemplateSpec `json:"template,omitempty"`
+}
+```
+
+#### Optional: Global/Per-Tenant limits
+
+In general the `LimitsSpec` supports declaring limits for the ingestion and/or the query path individually. The settings can be applied either globally or per tenant. One exception to this rule is the ingestion limit `MaxGlobalStreamsPerTenant` which is applied on ingestion globally.
+
+```go
+// LimitsTemplateSpec defines the limits  applied at ingestion or query path.
+type LimitsTemplateSpec struct {
+    // IngestionLimits defines the limits applied on ingested log streams.
+    //
+    // +optional
+    IngestionLimits IngestionLimitSpec `json:"ingestion,omitempty"`
+
+    // QueryLimits defines the limit applied on querying log streams.
+    //
+    // +optional
+    QueryLimits QueryLimitSpec `json:"queries,omitempty"`
+}
+
+// LimitsSpec defines the spec for limits applied at ingestion or query
+// path across the cluster or per tenant.
+type LimitsSpec struct {
+
+    // Global defines the limits applied globally across the cluster.
+    //
+    // +optional
+    Global LimitsTemplateSpec `json:"global,omitempty"`
+
+    // Tenants defines the limits applied per tenant.
+    //
+    // +optional
+    Tenants map[string]LimitsSpec `json:"tenants,omitempty"`
+}
+```
+
+##### Query Limits
+
+The global/per-tenant query limits ensure sane default sizing of query result sets. In general users of [Loki HTTP API's](https://grafana.com/docs/loki/latest/api/) query capabilities can provide a `limit` and/or `start`, `end` parameters to control the query result set. However, if none given the limits below reflect sane defaults to keep the query path from unbound operations.
+
+In addition a sane default for `MaxChunksPerQuery` limits the required object storage operations and in turn provides sanity in object storage per-operation cost.
+
+```go
+// QueryLimitSpec defines the limits applies at the query path.
+type QueryLimitSpec struct {
+
+    // MaxEntriesLimitPerQuery defines the aximum number of log entries
+    // that will be returned for a query.
+    //
+    // +optional
+    MaxEntriesLimitPerQuery int32 `json:"maxEntriesLimitPerQuery,omitempty"`
+
+    // MaxChunksPerQuery defines the maximum number of chunks
+    // that can be fetched by a single query.
+    //
+    // +optional
+    MaxChunksPerQuery int32 `json:"maxChunksPerQuery,omitempty"`
+
+    // MaxQuerySeries defines the the maximum of unique series
+    // that is returned by a metric query.
+    //
+    // + optional
+    MaxQuerySeries int32 `json:"maxQuerySeries,omitempty"`
+}
+```
+
+##### Ingestion Limits
+
+The global/per-tenant ingestion limits are typical rate limiting control options. The administrator can control the following options:
+
+1. Log volume rate and burst size to omit unbound memory operations, i.e. by setting `IngestionRate` and `IngestionBurstSize`.
+2. Label key and value lenghts to ensure faster log stream lookups, i.e. by setting `MaxLabelLength`, `MaxLabelValueLength`, `MaxLabelNamessPerSeries`.
+3. Log stream limits to ensure better object storage utilization, i.e. by setting `MaxStreamsPerTenant`, `MaxGlobalStreamsPerTenant`.
+
+From the above the most significant for healthy Loki operations are the the log stream limits. Since Logs are aggregated per tenant and unique set of labels to log streams, a high number of streams has a negative impact on object storage usage.
+
+This translates in general into the pattern that too many unique streams result in too many small chunks. In turn smaller chunks require more object storage operations to download/upload them. The general advice is to control the stream limits as per settings below and ensure a low churn on the label side by controlling cardinality.
+
+```go
+// IngestionLimitSpec defines the limits applied at the ingestion path.
+type IngestionLimitSpec struct {
+
+    // IngestionRate defines the sample size per second. Units MB.
+    //
+    // +optional
+    IngestionRate int32 `json:"ingestionRate,omitempty"`
+
+    // IngestionBurstSize defines the local rate-limited sample size per
+    // distributor replica. It should be set to the set at least to the
+    // maximum logs size expected in a single push request.
+    //
+    // +optional
+    IngestionBurstSize int32 `json:"ingestionBurstSize,omitempty"`
+
+    // MaxLabelLength defines the maximum number of characters allowed
+    // for label keys in log streams.
+    //
+    // +optional
+    MaxLabelLength int32 `json:"maxLabelLength,omitempty"`
+
+    // MaxLabelValueLength defines the maximum number of characters allowed
+    // for label values in log streams.
+    //
+    // +optional
+    MaxLabelValueLength int32 `json:"maxLabelValueLength,omitempty"`
+
+    // MaxLabelNamessPerSeries defines the maximum number of label names per series
+    // in each log stream.
+    //
+    // +optional
+    MaxLabelNamessPerSeries int32 `json:"maxLabelNamesPerSeries,omitempty"`
+
+    // MaxStreamsPerTenant defines the maximum number of active streams
+    // per tenant, per ingester.
+    //
+    // +optional
+    MaxStreamsPerTenant int32 `json:"maxStreamsPerTenant,omitempty"`
+
+    // MaxGlobalStreamsPerTenant defines the maximum number of active streams
+    // per tenant, across the cluster.
+    //
+    // +optional
+    MaxGlobalStreamsPerTenant int32 `json:"maxGlobalStreamsPerTenant,omitempty"`
+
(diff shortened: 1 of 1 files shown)


enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The purpose of this is to provide an alternative low-retention log store based on [Loki](https://github.com/grafana/loki) for the cluster logging stack. This aims to provide a cloud-native first solution specialized in storing logs and in turn lessen the operational cost running cluster logging.
+
+Currently our main log store, i.e. Elasticsearch, is a more general-purpose ingestion, indexing and querying cluster-based solution for text search. It is a powerful solution with fine-grained control on how to scale log storage, indexing and querying.
+
+In general it serves its purpose quite well but it is a constant source of labor regarding operations. Its powerfulness and complexity requires a constant toll on the human operator side to understand how to individually scale the cluster as a log store. Representative actions that a human operator needs to figure out besides controlled operations by [elasticsearch operator](https://github.com/openshift/elasticsearch-operator):
```
syedriko, line 39: "the Elasticsearch(?) cluster as a log store". Cluster is very overloaded term.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+- Regular evaluation and balance of the CPU/Memory requests and limits because configuration for ingestion and query paths is not individual configurable.
+- Regular evaluation and adaption of redundancy effects on memory and persistent volume usage.
+- Usage of `managementState: Unmanaged` to unblock the cluster in situations of bugs, e.g. locked indices, shard allocation.
+- Usage of the Elasticsearch cluster API to tune the cluster for options not supported in the Elasticearch custom resource.
```
syedriko, line 43: Elasticsearch cluster

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+
+On the other hand and based on our experience operating Loki as an internal central log aggregation solution, we found that a more specialized solution preserves the same log ingestion/querying user experience while lowering the human operational burden. Representative characteristics that showcase the improvements:
```
syedriko, line 45: Who is we/our?

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+- All components are stateless and can be scaled out individually based on load scenarios.
```
syedriko, line 46: Loki is a log *store*, so *all* the components can't be stateless.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+
+* Loki is operated only in a zero-dependency setup mode as described [here](https://grafana.com/docs/loki/latest/configuration/examples/#almost-zero-dependencies-setup). The only externally managed requirement is the availability of an S3-compatible API for object storage to persist logs.
+
+* Support a limited set of Loki scale outs, which are battle-tested on our internally managed Loki deployment. The target scale outs are defined using the format `Nx.{extra-small, small, medium}` where `N` the number of Loki deployments:
+  * **1x.extra-small**: Dedicated scale for development and demo purposes only! Not a valid configuration for cluster logging production environments.
+  * **1x.small**: Dedicated scale for low-retention log storage without redundancy support. All components are deployed with HA by running at least two replicas.
+  * **1x.medium**: Dedicated scale for log-retention log storage with single redundancy support. All components with HA by running at least two replicas except the ingester requiring at minimum three replicas to allow failover.
```
syedriko, line 99: "components **are deployed?** with HA"

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+
+* Support optionally single redundancy for log store on the target object storage.
+
+* Support auto-compaction of the index and chunks stored on the object storage when single redundancy enabled.
+
+### User Stories
+
+* As a cluster logging admin, I want to be able to configure Loki as an alternative log store in the ClusterLogging custom resource.
+
+* As a cluster logging admin, I want to be able to split logs to Loki as an alterantive log store in the ClusterLogForwarder custom resources to be able to support custom tenants.
```
syedriko, line 117: What does "split logs to Loki" mean here? Forward as in Cluster Log Forwarding?

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+    // OneXExtraSmallSize defines the size of a single Loki deployment instance
+    // with extra small resources/limits requirements and without HA support.
+    // This size is ultimately dedicated for development and demo purposes.
+    // DO NOT USE THIS IN PRODUCTION!
+    OneXExtraSmallSize LokiStackSizeType = "1x.extra-small"
+
+    // OneXSmall defines the size of a single Loki deployment instance
+    // with small resources/limits requirements and HA support for all
+    // Loki components. This size is dedicated for setup **without** the
+    // requirement for single redudancy and auto-compaction.
```
syedriko, line 196: redundancy (spelling)

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+    OneXSmall LokiStackSizeType = "1x.small"
+
+    // OneXMedium defines the size of a single Loki deployment instance
+    // with small resources/limits requirements and HA support for all
+    // Loki components. This size is dedicated for setup **with** the
+    // requirement for single redudancy and auto-compaction.
```
syedriko, line 202: redundancy

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+
+The integration of the `LokiStack` CRD into `ClusterLogging/v1` custom resources is following the present pattern to first into the latter.
+
+```go
+type LogStoreType string
+
+const (
+    LogStoreTypeElasticsearch LogStoreType = "elasticsearch"
+
+    LogStoreTypeLoki  LogStoreType = "Loki"
```
syedriko, line 543: Lowercase "loki", in keeping with the rest of "types" in the CL CRD,  type: "elasticsearch", type: "kibana", type: "fluentd"?

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+
+Testing the integration of the `LokiStack` custom resource in the `ClusterLogging` custom resource will be based upon this technique.
+
+#### E2E Testing
+
+Finally E2E testing requiring a running cluster in form of [kind](https://kind.sigs.k8s.io/) or [OpenShift](https://github.com/openshift/) is dedicated later for:
+
+1. Upgrade `LokiStack` scenarios.
+2. Compatibility testing with various S3-compatible object-storage providers (e.g. Openshift Storage).
+3. Integration with various caching providers, e.g. Memcached, Redis.
```
syedriko, line 574: This is a bit sudden. Where do these fit into the picture, how are they deployed, controlled etc.?

syedriko:
@periklis An overall comment is that this proposal does not mention that Loki comes with Grafana for a web UI. Are we looking to add 
```
visualization:
    type: "grafana"
```
to the CL CRD?

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+    // with the select cluster size.
+    ReasonInvalidRedundancyConfiguration LokiStackConditionReason = "InvalidRedundancyConfiguration"
+)
+
+// LokiStackStatus defines the observed state of LokiStack
+type LokiStackStatus struct {
+    // Conditions of the Loki deployment health.
+    //
+    // +optional
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
```
alanconway, line 529: The operator-sdk status.Conditions type makes managing conditions easier: 
https://pkg.go.dev/github.com/operator-framework/operator-sdk/pkg/status

enhancements/cluster-logging/loki-storage.md
periklis: Nope `custom resource` is correct here. The sentence is dedicated to user poking directly on Elastricsearch via its Cluster API because we don't expose all knobs in the Elasticearch custom resource definition.

enhancements/cluster-logging/loki-storage.md
periklis: WE == OpenShift Logging Team. Good point, will amend.

enhancements/cluster-logging/loki-storage.md
periklis: Loki is run stateless aka all it's components can be simple deployments w/o PVCs. The statefulness part is offloaded to the object storage, e.g. S3.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+
+* Loki is operated only in a zero-dependency setup mode as described [here](https://grafana.com/docs/loki/latest/configuration/examples/#almost-zero-dependencies-setup). The only externally managed requirement is the availability of an S3-compatible API for object storage to persist logs.
+
+* Support a limited set of Loki scale outs, which are battle-tested on our internally managed Loki deployment. The target scale outs are defined using the format `Nx.{extra-small, small, medium}` where `N` the number of Loki deployments:
+  * **1x.extra-small**: Dedicated scale for development and demo purposes only! Not a valid configuration for cluster logging production environments.
+  * **1x.small**: Dedicated scale for low-retention log storage without redundancy support. All components are deployed with HA by running at least two replicas.
```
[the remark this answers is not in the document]
periklis: It sound like a contradiction but in fact it is not. Redundancy does not refer to replicas here, but more to data processing redundancy. For example "single redundancy" means for Loki, that each log stream is processed on at least two ingesters. On the other hand you can still choose to have two ingester replicas but use "Zero Redundancy". From a traffic serving point you have two replicas always, so HA. However from a data processing point, you might loose data here, if a replica is killed and the handover is not finished to the other replica.

enhancements/cluster-logging/loki-storage.md
periklis: Split Logs refers here to `ClusterLogForwarder`'s capabilities to split log by type, i.e. `app`, `infra`, `audit`.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+  * **1x.medium**: Dedicated scale for log-retention log storage with single redundancy support. All components with HA by running at least two replicas except the ingester requiring at minimum three replicas to allow failover.
+
```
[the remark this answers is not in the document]
periklis: N > 1 is an option in this scheme to buy some freedom of choice later. There are two aspects that might require to use N=2 for Loki in future, which essentially means two loki clusters:
- Loki supports only a soft multi-tenancy model, i.e. all tenants are stored on the same object storage bucket. In addition tenants are fighting on query time for the same querier/query-frontend resources, where a busy/noise tenant might make other slower. What if we want to have hard tenancy for such specific use cases?
- Retention time is configured on the object storage bucket level. What if we want to support retetion times per tenant?

Having the option to scale with another Loki cluster - "cluster as a cattle" model - is a option here.

enhancements/cluster-logging/loki-storage.md
periklis: These are optional performance-oriented dependencies for query performance. The path forward here is to support them via their dedicated operators, i.e. loki-operator will create eventually a Memcached CR or a Redis CR.

periklis:
> @periklis An overall comment is that this proposal does not mention that Loki comes with Grafana for a web UI. Are we looking to add
> 
> ```
> visualization:
>     type: "grafana"
> ```
> 
> to the CL CRD?

This proposal is dedicated to the log store. We do not indent to introduce Grafana as a managed component by cluster-logging, as there is an operator out there for this job. We eventually will offer a route to a gateway in front of loki (for authn/authz) to allow users pick their visualization of choice. The only default we are going to have for Loki is integration in the OpenShift console.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+**Mitigation:** Limit by maintaining a strong opinionated API for Loki deployments that encompasses only features around dependencies (e.g. object storage), scheduling (e.g. requests/limits, tolerations/node-selectors) and no more.
+
+**Risk:** Providing poor user experience for day 1 and day 2, e.g. user needs to figure out requests/limits for each component.
+
+**Mitigation:** Provide opinionated and battle-tested scale out formats in form of a T-Shirt size (e.g. `1x.medium`). Leaving room for adding more formats later (e.g. `2x.medium` for hard multi-tenancy) and knobs (e.g. requests/limits) to adapt the format to live environments.
+
+## Design Details
+
+The discussion about the proposed API is split in two sections. The first section `lokistack_types.go` describes the custom resource definition for the [loki-operator](https://github.com/viaq/loki-operator) to reconcile a Loki deployment.
```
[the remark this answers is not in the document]
periklis: This is an interesting point, thanks for raising.

The discussion in the log exploration team to use `LokiStack` instead of `Loki` is based on two observations:
- To operate Loki for various scenarios (e.g. authn/authz), we might introduce auxiliary components, like the [observatorium-api](https://github.com/observatorium/observatorium). Thus we cannot speak anymore about Loki, but more of a stack that includes Loki.
- The term `LokiStack` is also a little bit derived from Grafana's [helm charts](https://grafana.com/docs/loki/latest/installation/helm/). They refer to a `LokiStack` every time you do install Loki plus other things, like Grafana/Prometheus/Promtail. We don't intend to install any of those, but for our Gateway proxy case this sound like being well aligned.

WDYT?

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+    // at minimum by two separate ingester instances.
+    SingleRedundancy RedundancyPolicyType = "SingleRedundancy"
+)
+```
+
+#### Mandatory: Loki deployment specification
+
+The mandatory fields of a `LokiStack` custom resource declaration are:
+- `Size`: Selection of one of the supported t-shirt size formats (see [Types and constants](#types-and-constants)) to reconcile the `CPU/Memory Requests`, `CPU/Memory Limits`, `Replicas` and `PodDisruptionBudgets` per Loki component.
+- `Storage`: A value of `ObjectStorageSpec` to declare the S3 compatible-API `URL` endpoint and a `Secret` located in `openshift-logging` by a human operator. If the `Secret` provides a key-value pair for `endpoint` then the `URL` is overridden.
```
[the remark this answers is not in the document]
periklis: Good point! I am a bit biased on how we handle URL+Secret in the `ClusterLogForwarder` case, but you are right. I missed the hidden bad endpoint case that masks a good URL.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+    // Template defines the resource/limits/tolerations/nodeselectors per component
+    //
+    // +optional
+    Template *LokiTemplateSpec `json:"template,omitempty"`
+}
+```
+
+#### Optional: Global/Per-Tenant limits
+
+In general the `LimitsSpec` supports declaring limits for the ingestion and/or the query path individually. The settings can be applied either globally or per tenant. One exception to this rule is the ingestion limit `MaxGlobalStreamsPerUser` which is applied on ingestion globally.
```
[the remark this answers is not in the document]
periklis: Also a good point to discuss. This is an naming inconsistency from Loki's [limit_config](https://grafana.com/docs/loki/latest/configuration/#limits_config), they write `User` and mean `Tenant`.

enhancements/cluster-logging/loki-storage.md
periklis: One exception to the rule is the ingester only if you run it with the Write-Ahead-Log feature enabled.

enhancements/cluster-logging/loki-storage.md
periklis: This pkg is not available in operator-sdk 1.x anymore.

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
periklis: I have rephased this to:
```
If the `Secret` provides a key-value pair for `endpoint` then the `URL` needs to be empty or else the CR is reported as invalid.
```

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,630 @@
...
+
+* Propose API support into the ClusterLogging CRD for single entry point deployments.
+
+We will be successful when:
+
+* Users are able to declare a Loki instance via the CRD and the [loki-operator](https://github.com/viaq/loki-operator) rolls out the Loki deployment.
+
+* Users are able to declare Loki as alternative store in the Cluster Logging custom resource.
+
+* Users are able to operate Loki in Cluster Logging without the need to tune every single bit via the proposed CRD.
```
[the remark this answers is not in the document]
periklis: The general guidance we would like to give users is 14-30days. It is only a guidance because they need to configure it on the object storage side.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,630 @@
...
+
+* We will not offer support for custom labeling of logs other than the defaults provided by the cluster logging collector.
+
+## Proposal
+
+To offer a seamless cluster logging experience with Loki as an alternative store, we propose to limit the API support to the following key features:
+
+* Deploy Loki **only** in microservices mode, i.e. each component is deployed and managed as an individual deployment/statefulset. This enables individual capacity planning and resource utilization per path ingestion/querying.
+
+* Multi-tenancy is enabled per default and requires the collector to apply the required [tenant identification](https://grafana.com/docs/loki/latest/operations/multi-tenancy/) on ingestion time. No user can disable this setting. In turn queries require the same identification information to be successful. Furthermore there is no support for querying across tenants.
```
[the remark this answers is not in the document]
periklis: AFAIK, there is no issue upstream. The key message here is that we don't include any GW like observatorium-api in the proposed API here. Such GW would solve some of the cross-tenant query work in OCP as per need to be aware of namespaces/rbac for in-cluster logging.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,630 @@
...
+
+* Loki is operated only in a zero-dependency setup mode as described [here](https://grafana.com/docs/loki/latest/configuration/examples/#almost-zero-dependencies-setup). The only externally managed requirement is the availability of an S3-compatible API for object storage to persist logs.
```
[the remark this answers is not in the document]
periklis: I am not aware of the support for such in loki. I assume yes.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,630 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The purpose of this is to provide an alternative low-retention log store based on [Loki](https://github.com/grafana/loki) for the cluster logging stack. This aims to provide a cloud-native first solution specialized in storing logs and in turn lessen the operational cost running cluster logging.
+
+Currently our main log store, i.e. Elasticsearch, is a more general-purpose ingestion, indexing and querying cluster-based solution for text search. It is a powerful solution with fine-grained control on how to scale log storage, indexing and querying.
+
+In general it serves its purpose quite well but it is a constant source of labor regarding operations. Its powerfulness and complexity requires a constant toll on the OpenShift cluster administrator side to understand how to individually scale the Elasticsearch cluster as a log store.
```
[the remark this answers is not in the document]
periklis: We don't need to make the comparison encompassing all the angles. I selected the most representative issues that we have on production systems today. Yes, virtual/physical hardware cost is important but it will remain high for Loki, too. Only local disk space will get a huge relief, but the cost for object storage will eliminate this IMHO.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,630 @@
+---
+title: Cluster Logging Loki as an alternative Log Store
```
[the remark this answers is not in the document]
periklis: The where here is "Cluster Logging". I tend to think that our default packaging model for in-cluster solutions is the operator. Thus we don't need to repeat ourselves. WDYT?

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
blockloop: I think we need to stop using the word "redundancy," because it's confusing. Loki doesn't mention the word "redundancy" in its code or documents that I can find. In Loki, Cortex, Thanos, etc they use "replication factor" which means how many replicas exist - which can be anything from one to infinity. When I hear "redundancy" I have to stop and think for a second. Can we please stick to the convention that other projects in our space use and avoid this confusion altogether?

enhancements/cluster-logging/loki-storage.md
blockloop: @syedriko Loki is not a "log store." By its own definition, it is a "log aggregation system." The storage is separate from Loki itself as Peri mentioned. The short-term statefulness of Loki can be enabled for efficiency, but it is not required by any means since the storage is offloaded to S3.

enhancements/cluster-logging/loki-storage.md
blockloop: I disagree. I don't think Loki will create any CR. We should ask the user for endpoints to an existing system. If we want to enable caching then that would live in CLO. Loki Operator is not an operator of operators it's just an operator of Loki.

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
blockloop: Stack refers to the components that make up the whole. Each component cannot be useful on its own. It is a delineation between the single-store instance of Loki - i.e. the all-in-one system which is intended for development and testing.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,628 @@
...
+- `RedundancyPolicy`: A selection of the supported redundancy policies (see [Types and constants](#types-and-constants)). All selections except `ZeroRedundancy` require a valid `Size` selection or the deployment will not proceed and reported as degraded state in the `LokiStackStatus`, i.e:
+  - `SingleRedundancy` requires at minimum `1x.medium`.
+
+The optional parts are `Limits` (see discussion in section [Optional: Global/Per-Tenant Limits](#optional-globalper-tenant-limits)) and `Template` (see details in section [Optional: Loki component configuration](#optional-Loki-component-configuration)).
+
+__Note__: The `Template` field if left empty will be populated by the [loki-operator](https://github.com/viaq/loki-operator) on the first reconciliation by translating the `Size` selection. This automatic update is to inform the user about the opinionated defaults for each supported scale out beyond documentation. The user can use the `Template` field to override these defaults.
+
+```go
+// ObjectStorageSecretSpec is a secret reference containing name only, no namespace.
```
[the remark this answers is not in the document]
blockloop: I think it's Kubernetes convention to put secrets in the namespace with the things that will use them. I would not suggest changing this to allow the secrets to exist anywhere. If the idea is to prevent access then RBAC is the solution.

enhancements/cluster-logging/loki-storage.md
syedriko: @blockloop Loki is log store from the point of view of the user of the cluster logging stack, in the sense that you say 
 logStore:
    type: "loki" 
flip the power switch off, flip it back on and expect to find your logs where you last saw them.
My point being, Loki vs Elasticsearch it is not a fair comparison, the comparison needs to be between Loki + S3 backend vs Elasticsearch. On public cloud, administering the S3 backend can be made someone else's problem - for a price, but that's not all the cases out there.

enhancements/cluster-logging/loki-storage.md
syedriko: I too think operator composition needs to be left to a higher-level operator, like CLO.

enhancements/cluster-logging/loki-storage.md
blockloop: I'm not sure I understand the reason for the distinction in the language. Your original comment appeared to be aimed at misrepresenting the statelessness of the stack when, in fact, the stack itself is stateless - save for any outside dependencies. The comparison between Elasticsearch and Loki is more superficial than the storage device itself. Ad absurdum, the comparison can technically delve into PVC storage engine vs S3 which doesn't provide anything to the conversation this enhancement aims to have. At least I don't think so. I think the point of this paragraph is to compare the two objective realities of the stacks. That being said, the Loki stack that _we control_ via the operator is entirely stateless and can be deployed as such, regardless of the syntactical differences we admit in calling it a store or an aggregation engine.

syedriko: @blockloop The very title of this document calls Loki a log store. We better come to some agreement on what it is.
I'm referring to the paragraphs just above that describe all the trouble one has to go to to administer Elastic as well as describing Loki as almost trouble free. This is what I think is not a fair comparison. In both cases it is expected that logs will be stored in highly available storage. In case of Elastic, it provides that highly available store, which is where many of the complexities of its administration originate. Loki does not provide its own highly available store but rather relies on one behind an S3 interface. But that object storage needs provisioning and administration which is likely non-trivial. It can be outsourced, it can be amortized over many consumers, but to completely eliminate it from the picture I think is incorrect. Let's say we're deploying OCP on bare metal and want to install a Loki based cluster logging solution on top of that. Would we need to deploy an S3-compatible store (Ceph?) even if we don't need it for anything else?

syedriko: I think I got it - look at this not from the perspective of writing an operator to run Loki but from the perspective of providing log storage for the Openshift cluster logging stack. Remove Elastic from the current stack and replace it with Loki - will there be a gap left?

periklis: Folks thank you first of all for this lively discussion about the details. IC where @syedriko drives the conversation. However, I would like to remind a couple of things to drive this to something useful like amending this section (please propose your amendments):
1. This is a summary section, it is not meant to be an exhaustive fair comparison between Loki and Elasticsearch. The effort here is to make visible the administrative burden on the one hand while proposing the mission to relief on the other hand.
2. The most important keyword is not `object storage` per se but `scale out`. For sure you can scale Elasticsearch horizontally. Its statefulness to do so is something one needs to master. The persistent volume attached is only one visible part of this statefulness. Other less visible to the non-expert are the shard allocation and the impact on memory for keeping in memory those shards to answer queries.
3. This section aims to highlight that you can scale the two paths ingestion and querying individually. This seems to be easier to achieve with dedicated individually scaled components like in Loki rather than role-based cluster nodes like in Elasticsearch. The separation of concerns is an indicator pro Loki here, but experience will teach us best.
4. This proposal is not encompassing all the details about object storage support by purpose. The message here is that the API supports to select the S3 storage of your choice. IMHO the support matrix of Loki and various object storage providers (S3, GCS, OpenShift Storage, etc.) belongs to a separate enhancement proposal that aims to close the loose ends from an Cluster-Logging API perspective. Our `ClusterLogging` API is unfortunately a bit biased using fields like `logStore` but this is something to be addressed there and not here. This proposal make the trade-off to re-use `store` as wording for Loki for the sake of simplicity.

enhancements/cluster-logging/loki-storage.md
periklis: Agreed! We can keep composition the ClusterLogging API and to cluster-logging-operator to drive the CRs for caching, etc.

Besides that this subsection discusses the E2E testing use cases for Loki. One of them is to test Loki with caching. Should we mention that we need CLO for such an E2E test? I believe not because:
- Testing Loki with caching requires available caching instance only not CLO creating caching CRs.
- Limiting the amount of dependencies (e.g. operators) one needs to run the tests is by experience a good path toward limiting flakes to happen.

@syedriko @blockloop WDYT? How should this sentence be improved? Mentioning CLO explicitly as a dependency for testing or not would help the reader?

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
periklis: I have reworked Redundancy into ReplicationFactor. PTAL on the new wording.

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
periklis: I suggest for the start to keep things simple and the operator should live and deploy its operands on `openshift-logging`. To keep RBAC management to a minimum, we should start simple and inform the administrator to create a secret in `openshift-logging`.

enhancements/cluster-logging/loki-storage.md
syedriko: @periklis As far as this text goes, it may be best to just drop the entire Summary section. What it describes clearly does not fit every OCP target environment and is, frankly, light on detail and not convincing. As far as motivation, we're adding Loki for those use cases where it is appropriate. I would also suggest dropping "alternative". From the point of view of the cluster logging stack, it's a log store. There may be others in the future. What I would add is how log collectors are going to talk to Loki (I'm assuming we're keeping fluentd and adding a Loki plugin to it) and how the UI is going to be exposed (I'm assuming Kibana is coupled with Elastic).

enhancements/cluster-logging/loki-storage.md
syedriko: @periklis I would keep testing of caching to CLO

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,605 @@
...
+
+Currently our main log store, i.e. Elasticsearch, is a more general-purpose ingestion, indexing and querying cluster-based solution for text search. It is a powerful solution with fine-grained control on how to scale log storage, indexing and querying.
+
+In general it serves its purpose quite well but it is a constant source of labor regarding operations. Its powerfulness and complexity requires a constant toll on the OpenShift cluster administrator side to understand how to individually scale the Elasticsearch cluster as a log store.
+
+Representative actions that an OpenShift cluster administrator needs to figure out besides controlled operations by [elasticsearch operator](https://github.com/openshift/elasticsearch-operator):
+- Regular evaluation and balance of the CPU/Memory requests and limits because configuration for ingestion and query paths is not individual configurable.
+- Regular evaluation and adaption of redundancy effects on memory and persistent volume usage.
+- Usage of `managementState: Unmanaged` to unblock the cluster in situations of bugs, e.g. locked indices, shard allocation.
```
ewolinetz, line 44: we can't guarantee this won't happen in Loki either, can we? Or is our escape hatch in cases to just roll a component pod?

syedriko: No we can't. One of the reasons I'm suggesting to drop the Summary altogether.

enhancements/cluster-logging/loki-storage.md
ewolinetz: I think there is a slight disconnect with verbiage happening...

The purpose of this proposal is to talk about the API for the Loki Operator of which it provides a Loki cluster as a service to anything else that would use it. CLO is one of those customers however it is not the sole one.

To the Loki Operator it is a log aggregation system. There are multiple components that make up the Loki solution that it is responsible for managing and keeping healthy so that any of the clients that interact with Loki just have a simple end point to push entries to or query from. It allows us to give you the appearance it is a place to send your logs to or query them from, however it is not a simple "logstore".

It seems like this may be something we need to add to the description to clarify things.
Maybe something like "what Loki is/isn't" and we can resolve this thread?

syedriko: @ewolinetz I think we need to clarify the scope, if it is just about the API for the Loki operator or if is it about plugging in a log store based on that Loki operator into the larger logging stack.

periklis: @syedriko I really appreciate that you widen the horizon on this topic with interesting questions, e.g. drop alternative wording (I assume you mean we keep both stores supported. That's also a question for @sichvoge), UI exposure. In general I agree these topics need some sort of resolution. However, the summary block in this doc has some important constraints.

First of all the section is mandatory for our enhancement proposal process. It cannot be simply dropped. The summary block is not meant to be a management summary for a whole project or epic. The purpose is clearly to provide a entrypoint/context to the text below according to the [template](https://github.com/openshift/enhancements/blob/master/guidelines/enhancement_template.md#summary). It should not be an exhaustive description of what a team wants to achieve across topics that require multiple enhancement proposal docs (e.g. UI expose aka replacing Kibana maybe?)

I have to admit that I have a hard time reading this thread to identify amendments that clearly shape the document in a better form. The summary provided aims to provide an entrypoint on why we want to provide a alternative store regardless if the alternative means replacement or not. The agenda to keep/drop Elasticsearch is a topic for another proposal or arch meeting. In addition the summary uses a light comparative approach to highlight differences, e.g. statefulness/statelessness, individual/combined scale out of ingestion and querying. It does neither compare in detail all the aspects nor tries to put Elasticsearch in a dark corner. The reader should be informed by this section on how to calibrate his lenses on a highly overloaded term as `logStore`.

Finally, the document is scoped to a `LokiStack` API that is provided by the `loki-operator`. For the sake of a useful bridge to the `ClusterLogging` API, this proposal also describes how to achieve an integration with present means and no more.

enhancements/cluster-logging/loki-storage.md
periklis: No we cannot guarantee this won't happen with the `loki-operator`. The `managementState` field is always meant to be a stop signal for the operator **not** the operand. IMHO it is a bad smell having in the API for an operand (describes the desired state of the operand) a field meant for the operator. Why not simply `oc scale deploy loki-operator --replicas=0` when someone needs to stop the reconciliation?

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,605 @@
...
+
+* We will not offer support for custom labeling of logs other than the defaults provided by the cluster logging collector.
+
+## Proposal
+
+To offer a seamless cluster logging experience with Loki as an alternative store, we propose to limit the API support to the following key features:
+
+* Deploy Loki **only** in microservices mode, i.e. each component is deployed and managed as an individual deployment/statefulset. This enables individual capacity planning and resource utilization per path ingestion/querying.
+
+* Multi-tenancy is enabled by default and requires the collector to apply the required [tenant identification](https://grafana.com/docs/loki/latest/operations/multi-tenancy/) on ingestion time. No user can disable this setting. In turn queries require the same identification information to be successful. Furthermore there is no support for querying across tenants.
```
Red-GV, line 95: Asking out of my own ignorance here, is this a limit by Loki itself or are we imposing that? Also, is there some scenario in which a customer for some reason has would have multiple tenants assigned under it and thus might need such a feature?

alanconway: The current collector proposal allows the user to set a fixed tenant ID, use the namespace or use a label value.  Do you think we should default to tenant==namespace, or enforce it as the only option?

periklis: @alanconway The default case should be `tenant == namespace` to keep the multi-tenancy model aligned with what we support with ES right now.

enhancements/cluster-logging/loki-storage.md
alanconway: @syedriko would this be clear/complete?:
- The scope here is the API to Loki operator
-  Logging stack users that want Loki as a drop-in replacement to ES would just set `logStore: loki` in ClusterLogging, the operators take care of the rest (do we need to expose some S3 info in ClusterLogging for this to work??)
-  Logging stack users that want more control/a different Loki instance see https://github.com/openshift/enhancements/pull/710#pullrequestreview-631355181

We can add a note to each of the proposals to that effect.

alanconway: I guess @syedriko has a point that somewhere we have to describe the changes to ClusterLogging to enable Loki. I'm inclined to add that to https://github.com/openshift/enhancements/pull/710#pullrequestreview-631355181 rather than here
- it's not worth creating another proposal just for that
- its healthy for this to describe a Loki operator independently of how CLO will use it.
- #710 is already about configuring the CLO,  makes sense to discuss CL alongisde CLF

enhancements/cluster-logging/loki-storage.md
ewolinetz: > Why not simply `oc scale deploy loki-operator --replicas=0` when someone needs to stop the reconciliation?

can you guarantee that you will be the only one impacted by that? that is to say... will you be the only one using the operator? the `Unmanaged` paradigm was to allow the operator to continue operating for other CRs.

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
ewolinetz: In general, retention is also going to be determined by the s3 configuration. we will try to scale up to serve requests as best we can and make recommendations if necessary. but that should mostly be something hidden to the customer

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,605 @@
...
+
+* Support adding kubernetes tolerations and node selectors for all Loki components individually. This enables operating the Loki deployment on dedicated nodes.
+
+* Support a limited set of control options from Loki's [limit_config](https://grafana.com/docs/loki/latest/configuration/#limits_config) for the ingestion and query paths. The set is an opinionated list of options to enable tuning of object storage usage. The list of options is offered globally across the whole Loki deployment or per tenant.
+
+For **1x.medium** instances only:
+
+* Support optionally replication factor of two for log store on the target object storage.
+
+* Support auto-compaction of the index and chunks stored on the object storage when replication factor of two enabled.
```
ewolinetz, line 114: didn't we discuss we would have this for anything with RF2, so compaction should also apply for `1x.small` right?

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,605 @@
...
+
+* Support tuning of CPU/Memory resource and limits for each individual component beyond the selected scale-out size.
+
+* Support adding kubernetes tolerations and node selectors for all Loki components individually. This enables operating the Loki deployment on dedicated nodes.
+
+* Support a limited set of control options from Loki's [limit_config](https://grafana.com/docs/loki/latest/configuration/#limits_config) for the ingestion and query paths. The set is an opinionated list of options to enable tuning of object storage usage. The list of options is offered globally across the whole Loki deployment or per tenant.
+
+For **1x.medium** instances only:
+
+* Support optionally replication factor of two for log store on the target object storage.
```
ewolinetz, line 112: similar question regarding RF2 and `small`

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,605 @@
...
+
+* Support auto-compaction of the index and chunks stored on the object storage when replication factor of two enabled.
+
+### User Stories
+
+* As a cluster logging admin, I want to be able to configure Loki as an alternative log store in the `ClusterLogging` custom resource.
+
+* As a cluster logging admin, I want to be able to split logs to Loki as an alterantive log store in the `ClusterLogForwarder` custom resources to be able to support custom tenants.
```
ewolinetz, line 120: i think we should drop the cluster logging use cases for this proposal and focus on what we will expose from the Loki Operator stand-point.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,605 @@
...
+
+* As a cluster logging admin, I want to select a different scale for my Loki deployment to enable single replication factor and auto-compaction.
+
+* As a cluster logging admin, I want to expand resources/limits for individual Loki components to support load scenarios not declared by the cluster t-shirt size selected.
+
+* As a cluster logging admin, I want to configure scheduling of individual Loki components to custom selected nodes on my cluster to separate resource usage of the cluster logging from other workloads on my cluster.
+
+* As a Cluster Logging admin, I want to tune the ingestion limits globally or per tenant to support custom ingestion characteristics (e.g. ingenstion rate, label cardinality).
+
+* As a Cluster Logging admin, I want to tune the query limits globally or per tenant to support custom query characteristics (e.g. chunks per series returned).
```
ewolinetz, line 130: to go along with my prior comment we could update this to something along the lines of "As a Loki Operator consumer"

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
alanconway: Still seems redundant. Loki is _always_ composed of distributed components. Even if you choose to "distribute" them all into a single container in single-store mode, they are still a "stack". So it seems to me that the word "Loki" already means "set of distributed components", and there's no need to add "Stack". But this is a nit, not something to hold up the PR.

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
alanconway: +1  I'd prefer not to have a separate place for the endpoint at all (unless you have endpoint(s) that can't be expressed as a URL, like the kafka client-side load-balancing endpoint-list)
What is the syntax of your "endpoint" in the Secret if not a URL? My preference for URLs is based on much past pain, a complete and unambigous endpoint syntax is trickier than it looks, and _documenting_ a non-standard syntax ... well - URL is just easier, whatever the warts. If you tell me an endpoint can be written simply as "host:port" I will slap you.

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
alanconway: Agreed. Needs some research to justify the extra flexibility, and maybe even the research isn't justified if this is common practice and nobody has complained to date.

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
alanconway: I'd advocate we stick to tenant unless we can clearly document distinct meanings for "User" and "Tenant" from our own perspective. I see the argument for following Loki terms, but a user of our API should normally not need to go down to loki config files, and "user/tenant" is already a minefield of confusion in the industry in general.

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
alanconway: Also this sounds like a counter argument: "The term LokiStack is also a little bit derived from Grafana's helm charts. They refer to a LokiStack every time you do install Loki plus other things, like Grafana/Prometheus/Promtail." 
We explicitly do not want this operator to manage collectors or visualization, so it sounds like LokiStack is a name with built-in scope-creep.

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
lukas-vlcek: @bwplotka To be really honest, when I saw first iteration of @periklis's HW proposal for Loki integration I thought to myself "If only we ever gave ES cluster so much resources (especially CPU...)". BTW did you know OCP default budget for single ES node was 1/2 CPU? (Why this was happening is a different story...) Going forward, if we will want to make any claims about how much Loki is doing better than ES we should really compare 1-1 on the resources level.

enhancements/cluster-logging/loki-storage.md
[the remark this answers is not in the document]
periklis: As per discussion above we will stick with `LokiStack` for now to keep the option to add auxiliary components for Loki in the stack.

enhancements/cluster-logging/loki-storage.md
periklis: As per discussion, we will add a `ManagementState` field with default value `Managed`. Thus users don't need to add this field per default.

alanconway:
/approve
I'll let someone else add the LGTM

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,623 @@
...
+    // +optional
+    Tenants map[string]LimitsSpec `json:"tenants,omitempty"`
+}
+```
+
+##### Query Limits
+
+The global/per-tenant query limits ensure sane default sizing of query result sets. In general users of [Loki HTTP API's](https://grafana.com/docs/loki/latest/api/) query capabilities can provide a `limit` and/or `start`, `end` parameters to control the query result set. However, if none given the limits below reflect sane defaults to keep the query path from unbound operations.
+
+In addition a sane default for `MaxChunksPerQuery` limits the required object storage operations and in turn provides sanity in object storage per-operation cost.
```
ewolinetz, line 333: i forget, did we talk about having something like a "profile" sort of sizing for the limits on the current api proposal or in a follow up? (to give a facade to the different knobs that loki currently exposes but may go away)

periklis: No not really, we have not investigated or discussed this topic too much. There is always an implication that selecting a `Size: [1x.extra-small|1x.small|1x.medium]` will apply some sane defaults on the limits.

enhancements/cluster-logging/loki-storage.md
```diff
@@ -0,0 +1,623 @@
...
+    ManagementStateManaged ManagementStateType = "Managed"
+
+    // ManagementStateUnmanaged when the LokiStack custom resource should not be
+    // reconciled by the operator.
+    ManagementStateUnmanaged ManagementStateType = "Unmanaged"
+)
+
+// LokiStackSizeType declares the type for Loki deployment scale outs.
+//
+// +kubebuilder:validation:Enum=OneXExtraSmallSize;OneXSmall;OneXMedium
```
vishnoisuresh, line 200: I wonder If it's intentional that `OneXExtraSmallSize` has **size** but not others `OneXSmall;OneXMedium` fields.

periklis: @vishnoisuresh  This is a typo! Thanks for watching out! As per source code in enhancement proposal being more for demonstration, what matters is on the operator source code. AFAICS, we have this fixed there:
```
// +kubebuilder:validation:Enum=SizeOneXExtraSmall;SizeOneXSmall;SizeOneXMedium
```

PR #713, opened 2021-03-30
Ingress: Post-implementation fixups for a couple of EPs

sgreene570:
Move `gcp-lb-global-access` & `tunable-router-buffer-sizes` to implemented. Touch up the API example in `tunable-router-buffer-sizes` to match what was merged in openshift/api.


russellb:
Thanks a bunch for these updates!  I think most enhancements don't get this type of cleanup.

/approve
/lgtm

PR #714, opened 2021-03-30
README: clarify that new components require an enhancement

markmc:
A recent request to add a new bugzilla component received the response
"new components typically require architect signoff".

This is better expressed as "requires an enhancement", and "user-facing
component" is a good rule-of-thumb for "will need a new bugzilla component".


russellb:
/approve

makes sense to me

PR #717, opened 2021-03-31
Create pre-assigning-nodes-to-machine-config-pools.md

beekhof:
Ensure that the MCO's understanding of the "correct" config pool for a node is consistent with the config pool it was provisioned with via Ignition


beekhof:
Closing.  MCO team to recreate and complete if necessary

PR #555, opened 2020-12-02
Cluster High-availability Mode API

dhellmann:
As we add more deployment topologies for OpenShift, we need a way to
communicate expectations to the operators managing the cluster. This
enhancement describes an API for telling operators how hard they
should work to provide high-availability.

PR #718, opened 2021-03-31
Introduce HA conventions for platform components

ravisantoshgudimetla:
This PR introduces HA conventions for platform specific operators/operands

cc @smarterclayton

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -126,6 +126,24 @@ it is used to manage all elements of the cluster.
 * Every component must remain available to consumers without disruption during upgrades and reconfiguration
   * See the next section for more details
 
+#### High Availability
+
+We focus on minimizing the impact of a failure of individual nodes in a cluster by ensuring operators or operands are spread across multiple nodes.
+When OpenShift runs in [cluster high availability mode](https://github.com/openshift/enhancements/pull/555), the supported cluster topologies are 3 control-plane nodes and 2 workers, or 3 control-plane nodes that also run workloads (compact clusters).  The following scenarios are intended to support the minimum 2 worker node configuration.  
+Please note that in case of [Single Node OpenShift](https://github.com/openshift/enhancements/blob/master/enhancements/single-node/production-deployment-approach.md), since the replicas needed are always 1, there is no need to have affinities set.
+
+
+
+* If the operator or operand wishes to be highly available and can tolerate the loss of one replica, the default configuration for it should be
+  * Two replicas
+  * Hard pod [anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node) on hostname (no two pods on the same node)
+  * Use the maxUnavailable rollout strategy on deployments (prefer 25% by default for a value)
+  * This is the recommended approach for all the components unless the component needs 3 replicas
+* If the operator or operand requires >= 3 replicas and should be running on worker nodes
+  * Set soft pod anti-affinity on the hostname so that pods prefer to land on separate nodes (will be violated on two node clusters)
+  * Use maxSurge for deployments if possible since the spreading rules are soft
+
+In future, when we include the descheduler into OpenShift by default, it will periodically rebalance the cluster to ensure spreading for operand or operator is honored. At that time we will remove hard anti-affinity constraints, and recommend components move to a surge model during upgrades.
 
 #### Upgrade and Reconfiguration
 


CONVENTIONS.md
```diff
@@ -126,6 +126,19 @@ it is used to manage all elements of the cluster.
...
 
+#### High Availability
+
+We focus on minimizing the impact of single node failures by ensuring operators or operands are spread across multiple nodes while maintaining backwards compatibility.
+
+* If the operator or operand wishes to be highly available and can tolerate the loss of one replica, the default configuration for it should be
+  * Two replicas
+  * Hard pod [anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node) on hostname (no two pods on the same node)
+  * Use the maxUnavailable rollout strategy on deployments (prefer 25% by default for a value)
+* If the operator or operand requires >= 3 replicas
```
simonpasquier, line 137: Should it be mentioned that the minimum requirement is 2 worker nodes which explains why replicas >= 3 mandate soft affinity?

ravisantoshgudimetla: I've mentioned it as maintaining backwards compatibility. I can expand on it a bit further if it clarifies things.

CONVENTIONS.md
```diff
@@ -126,6 +126,20 @@ it is used to manage all elements of the cluster.
 * Every component must remain available to consumers without disruption during upgrades and reconfiguration
   * See the next section for more details
 
+#### High Availability
+
+We focus on minimizing the impact of single node failures by ensuring operators or operands are spread across multiple nodes while maintaining backwards compatibility. Currently, we mandate 3 control-plane nodes and 2 worker nodes in OpenShift, so the following scenarios account for components that are running on both control-plane and worker nodes.
```
smarterclayton, line 131: Second sentence needs to describe compact clusters (which is equivalent to 3+3)

CONVENTIONS.md
```diff
@@ -126,6 +126,20 @@ it is used to manage all elements of the cluster.
...
+* If the operator or operand wishes to be highly available and can tolerate the loss of one replica, the default configuration for it should be
+  * Two replicas
+  * Hard pod [anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node) on hostname (no two pods on the same node)
+  * Use the maxUnavailable rollout strategy on deployments (prefer 25% by default for a value)
+  * This is the recommended approach for all the components unless the component needs 3 replicas
+* If the operator or operand requires >= 3 replicas and should be running on worker nodes
+  * Set soft pod anti-affinity on the hostname so that pods prefer to land on separate nodes (will be violated on two node clusters)
+  * Use maxSurge for deployments if possible since the spreading rules are soft
+
+In future, when we include descheduler in our product, it will periodically rebalance the cluster to ensure spreading for operand or operator is honored. At that time we will remove hard anti-affinity constraints, and recommend components move to a surge model during upgrades.
```
smarterclayton, line 142: Slight tweak, we don't call openshift a product in the conventions / enhancements repo

> In the future, when we include the descheduler into OpenShift by default, ...

CONVENTIONS.md
dhellmann, line 131: Since we use "single node" to describe a deployment topology, this sentence might be clearer as something like

```suggestion
We focus on minimizing the impact of a failure of individual nodes in a cluster by ensuring operators or operands are spread across multiple nodes while maintaining backwards compatibility. Currently, we mandate 3 control-plane nodes and 2 worker nodes in OpenShift, so the following scenarios account for components that are running on both control-plane and worker nodes. 
```

dhellmann: It would be ideal if this new section was organized around the 3 supported deployment topologies (single-node, compact, and standard), with the conventions and expectations for each discussed separately.

CONVENTIONS.md
ravisantoshgudimetla: Done

ravisantoshgudimetla:
@smarterclayton and @dhellmann - I incorporated the changes suggested by you. PTAL

CONVENTIONS.md
```diff
@@ -126,6 +126,22 @@ it is used to manage all elements of the cluster.
 * Every component must remain available to consumers without disruption during upgrades and reconfiguration
   * See the next section for more details
 
+#### High Availability
+
+We focus on minimizing the impact of a failure of individual nodes in a cluster by ensuring operators or operands are spread across multiple nodes while maintaining backwards compatibility.
```
Miciah, line 131: What does "while maintaining backwards compatibility" have to do with these guidelines?  Is this just a reminder that having multiple pod replicas may necessitate extra precautions related to version skew?

CONVENTIONS.md
```diff
@@ -126,6 +126,22 @@ it is used to manage all elements of the cluster.
...
+
+
+* If the operator or operand wishes to be highly available and can tolerate the loss of one replica, the default configuration for it should be
+  * Two replicas
+  * Hard pod [anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node) on hostname (no two pods on the same node)
+  * Use the maxUnavailable rollout strategy on deployments (prefer 25% by default for a value)
+  * This is the recommended approach for all the components unless the component needs 3 replicas
+* If the operator or operand requires >= 3 replicas and should be running on worker nodes
+  * Set soft pod anti-affinity on the hostname so that pods prefer to land on separate nodes (will be violated on two node clusters)
+  * Use maxSurge for deployments if possible since the spreading rules are soft
```
Miciah, line 142: Why use maxSurge important when using soft affinity?  Is there a reason not to use maxSurge with hard affinity?

Miciah (commented):
Are zones out of scope for these guidelines?

CONVENTIONS.md
```diff
@@ -126,6 +126,22 @@ it is used to manage all elements of the cluster.
 * Every component must remain available to consumers without disruption during upgrades and reconfiguration
   * See the next section for more details
 
+#### High Availability
+
+We focus on minimizing the impact of a failure of individual nodes in a cluster by ensuring operators or operands are spread across multiple nodes while maintaining backwards compatibility.
+The smallest supported configuration for OpenShift is 3 control-plane nodes and 2 workers, or 3 control-plane nodes that also run workloads (compact clusters), so the following scenarios are intended to support the minimum 2 worker node configuration
```
dhellmann, line 132: This is no longer true. The smallest supported configuration is 1 node. See https://github.com/openshift/enhancements/blob/master/enhancements/single-node/production-deployment-approach.md

As a developer considering what to do about HA, I need to understand all of the topologies for OpenShift deployments (1, 3, and 3+2 nodes) so that I can make my component respond correctly in all of those situations. This text really does need to cover all 3 topologies so that new contributors are aware of them and aware that even if a component wants to provide HA services in some configurations it must not assume that it always can.

CONVENTIONS.md
ravisantoshgudimetla: maxSurge with hard affinity would just allow pods to be created but because of node limit the pods will be in pending state not consuming actual resources on the node whereas with soft affinity since the constraints are ok to be violated pods will be created on the nodes after scheduling stage

CONVENTIONS.md
ravisantoshgudimetla: backwards compatibility refers to 2 worker nodes requirement supported.

CONVENTIONS.md
ravisantoshgudimetla: IIUC, in the context of SNO, I think there won't be any soft-affinities or hard-affnities set since we would like to have 1 replica. Would it be ok to call that out explicitly?  Since we are going to reduce replicas and replica requirement is called in the [proposal](https://github.com/openshift/enhancements/blob/master/enhancements/single-node/production-deployment-approach.md), I think it is safe to HA is in the context of 3 Master + 3 Worker nodes or compact clusters. 

How about this:

1``
When OpenShift runs in [cluster high availability mode](https://github.com/openshift/enhancements/pull/555), the supported cluster topologies are 3 control-plane nodes and 2 workers, or 3 control-plane nodes that also run workloads (compact clusters), the following scenarios are intended to support the minimum 2 worker node configuration. In case of SNO, please look at the replica requirement [specified](https://github.com/openshift/enhancements/blob/master/enhancements/single-node/production-deployment-approach.md).
```

Comment (dhellmann) [reply CONVENTIONS.md:132]: Yes, you're right that affinities shouldn't be used and only 1 replica should be used for single node deployments. I also agree that stating that explicitly, here, makes sense.

If the reader of this document is someone trying to understand what to do for HA settings in their new component, then it's better not to send them off to read another document to learn about 1 of the cases they need to handle. If we summarize all of the information they need here, they will be able to understand all of the cases together and do their work more effectively. It's OK to link to other documents for more detail to supplement that summary.

File: CONVENTIONS.md (hunk 4 above)
Comment (Miciah) [reply CONVENTIONS.md:131]: Are you saying there was a version of OpenShift where the minimum was 3 worker nodes, and the focus must include maintaining compatibility with those clusters?  Isn't that a vacuous requirement?  I mean, if these guidelines avoid breaking 2-worker clusters, is extra effort required not to break 3-worker clusters?

File: CONVENTIONS.md (hunk 5 above)
Comment (Miciah) [reply CONVENTIONS.md:142]: Ah, I see.  Thanks for explaining!  The way we get around that with the ingress operator is that it creates a deployment with a label that has a hash of the deployment and a hard pod anti-affinity rule that selects on that label, so that the hard anti-affinity rule does not apply to changes to the deployment that cause rolling updates.  Maybe it isn't worth going into all that detail here though.

File: CONVENTIONS.md (hunk 4 above)
Comment (ravisantoshgudimetla) [reply CONVENTIONS.md:131]: Nope, we are concerned about 2-worker clusters which we expanded in the later section. In order to avoid confusion, I am dropping this line altogether.

**Comment by ravisantoshgudimetla** [comment]
> Are zones out of scope for these guidelines?

Yes, usually the zones or nodes are failure domains, we are spreading across nodes as they are the cause of single point of failure at the application level. You can always add zones to the deployment spec if it helps in your case.

**Comment by ravisantoshgudimetla** [comment]
@dhellmann @Miciah - Updated with another version. PTAL

File: CONVENTIONS.md [hunk 7]
```diff
@@ -126,6 +126,24 @@ it is used to manage all elements of the cluster.
 * Every component must remain available to consumers without disruption during upgrades and reconfiguration
   * See the next section for more details
 
+#### High Availability
+
+We focus on minimizing the impact of a failure of individual nodes in a cluster by ensuring operators or operands are spread across multiple nodes.
+When OpenShift runs in [cluster high availability mode](https://github.com/openshift/enhancements/pull/555), the supported cluster topologies are 3 control-plane nodes and 2 workers, or 3 control-plane nodes that also run workloads (compact clusters), the following scenarios are intended to support the minimum 2 worker node configuration.  
+Please note that in case of [Single Node OpenShift]((https://github.com/openshift/enhancements/blob/master/enhancements/single-node/production-deployment-approach.md).), since the replicas needed are always 1, there is no need to have affinities set.
```
Miciah, line 133: Small markup fix:
```suggestion
Please note that in case of [Single Node OpenShift](https://github.com/openshift/enhancements/blob/master/enhancements/single-node/production-deployment-approach.md), since the replicas needed are always 1, there is no need to have affinities set.
```

CONVENTIONS.md
```diff
@@ -126,6 +126,24 @@ it is used to manage all elements of the cluster.
 * Every component must remain available to consumers without disruption during upgrades and reconfiguration
   * See the next section for more details
 
+#### High Availability
+
+We focus on minimizing the impact of a failure of individual nodes in a cluster by ensuring operators or operands are spread across multiple nodes.
+When OpenShift runs in [cluster high availability mode](https://github.com/openshift/enhancements/pull/555), the supported cluster topologies are 3 control-plane nodes and 2 workers, or 3 control-plane nodes that also run workloads (compact clusters), the following scenarios are intended to support the minimum 2 worker node configuration.
```
Miciah, line 132:
```suggestion
When OpenShift runs in [cluster high availability mode](https://github.com/openshift/enhancements/pull/555), the supported cluster topologies are 3 control-plane nodes and 2 workers, or 3 control-plane nodes that also run workloads (compact clusters).  The following scenarios are intended to support the minimum 2 worker node configuration.  
```

CONVENTIONS.md
ravisantoshgudimetla: Done. I have committed your suggestion. Thank you @Miciah

ravisantoshgudimetla:
@dhellmann - Can you PTAL, so that we can get this merged.

PR #719, opened 2021-03-31
Create change-ocp-install-config-values-after-install-completes

hhockett:
Initial enhancement created for change-ocp-install-config-values-after-install-completes

This PR introduces a way to change install-config values including cluster name, base domain name, and other networking configuration after OpenShift install is complete.


russellb:
/hold

This seems to be missing any detail on how this would be accomplished.  Is this intended as a feature request?  Or is this still a work-in-progress proposal?

hhockett:
Closing now with RFE opened as new feature request, thank you!

PR #722, opened 2021-04-06
Add "Build OKD for ppc64le" proposal

mjturek:
Add proposal to build and publish ppc64le payloads alongside x86_64 builds.

### DIFF
--- a/enhancements/multi-arch/okd-ppc64le-builds.md
+++ b/enhancements/multi-arch/okd-ppc64le-builds.md
@@ -49,16 +49,6 @@ done should be extensible enough that adding other architectures would be simple
 - Build and release OKD payloads for ppc64le alongside x86_64.
 - Update documentation to reflect ppc64le availability
 
-### Risks and Mitigations
-
-What are the risks of this proposal and how do we mitigate. Think broadly. For
-example, consider both security and how this will impact the larger OKD
-ecosystem.
-
-How will security be reviewed and by whom? How will UX be reviewed and by whom?
-
-Consider including folks that also work outside your immediate sub-project.
-
 ## Design Details
 
 ### Test Plan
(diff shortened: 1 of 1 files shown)


mjturek:
I know this is a bit bare bones at the moment but I'd like to start a conversation about how we can start building and publishing for ppc64le. If there's something that you think should be detailed or discussed, please bring it up here!

LorbusChris:
Thanks for working on this @mjturek!

I'd suggest we approach Fedora and/or CentOS infra folks to discuss the resource requirements with them.

Some additional details on how the builds will be done would be great, too.
For OKD on x86, FCOS is built on an external Jenkins cluster, while container builds are done on Prow (the majority of OKD containers are mirrored/promoted OCP CI builds, the rest are also built on Prow and directly tagged into the `origin` namespace.)

How do you envision the process be for ppc64le?

clnperez:
/remove-lifecycle stale

We've been working on a prow cluster on Power inside of Power VS (part of IBM Cloud). Once we have that working we can extend what we've learned there to OKD. It's just taking a bit longer as there are no upstream images for prow on Power (yet).

LorbusChris:
We might still want to do this eventually. We're working on ARM64 CI and OKD builds now, which should serve as a blueprint for further archs.

/remove-lifecycle stale

clnperez:
@LorbusChris we're getting a bit more interest from customers. You'd mentioned getting with the Fedora and/or CentOS infra folks. Is there a process there? Do you know who to ping?

LorbusChris:
Hi @clnperez,
I think that specific comment is now obsolete.
OKD is built on OpenShift's multi-cluster Prow CI system, so there'd have to be at least one ppc64le build farm cluster set up and added to Prow.
I think work for that is already on its way, but I'm not involved in that effort. Once that's set up though, and there are per-component CI builds and CI release payloads, it'll be possible to also create OKD release payloads for that arch.

clnperez:
Yep, that's underway (and I've been keeping informed). Sounds like what we do need is fcos builds first -- and that's a conversation that's happening (kind of) here: https://github.com/coreos/fedora-coreos-tracker/issues/987 & https://github.com/coreos/coreos-assembler/issues/2473

PR #724, opened 2021-04-07
Update internationalization enhancement for how to handle login template translation strings

sg00dwin:
https://issues.redhat.com/browse/CONSOLE-2535

@spadgett @jhadvig

### DIFF
--- a/enhancements/console/internationalization.md
+++ b/enhancements/console/internationalization.md
@@ -64,6 +64,11 @@ Localization will not be included in the CLI at this time, and bidirectional (ri
 
 ## Proposal
 
+
+### User Stories
+
+As an OpenShift user, I want to access to the web console information in my preferred language.
+
 [react-i18next framework](https://react.i18next.com/) an internationalization framework for React based on [i18next](http://i18next.com/).
 It provides standard internationalization features such as plurals, context, interpolation, and format,
 as well as language detection, loading, caching, and extensions through additional plugins.
@@ -110,6 +115,15 @@ The desired fallback experience if not all pieces/parts of the UI are translated
 
 We will also modify our existing file conversion tools to handle translation files spread across multiple packages so we can easily export/import translation files.
 
+Translation of the login page will be handled by [oauth-server](https://github.com/openshift/oauth-server), which will check for request `Accept-Language` header. If the header is missing in the request, `oauth-server` will default to English language for localization.
+
+[oauth-server](https://github.com/openshift/oauth-server) contains the default OKD login templates. Custom templates are stored in the [oauth-templates](https://github.com/openshift/oauth-templates) repository and will also need to be updated for the other brands like the OCP and OpenShift Dedicated login pages.
+In case of any changes in these repositories, Console-Team will be responsible for syncing those changes between them.
+
+Since there are only a small number of strings that rarely change, the localized strings for each of the supported languages will be directly hardcoded into the `oauth-server`, so we don't have to load them during runtime.
+These strings are to be included in the console repository and the i18next-parser will add them into the files sent to the Globalization team for translation. Afterwards the console team will contribute them back to the [oauth-server](https://github.com/openshift/oauth-server).
+
+
 ### Risks and Mitigations
 
 **Risk**: All text is not translated.
@@ -141,15 +155,33 @@ We will have unit and integration tests. We will also work with the Globalizatio
 
 Initially we will just mark strings internally. Translations and testing will be added once marking is complete.
 
+#### Dev Preview -> Tech Preview
+
+None
+
+#### Tech Preview -> GA
+
+None
+
+#### Removing a deprecated feature
+
+None
+
 ### Upgrade / Downgrade Strategy
 
 We will need to make sure that new components with hard-coded strings are marked for translation.
+
+### Version Skew Strategy
+
+None
+
 
 ## Implementation History
 
 * 2020-08-03 - [PR against console](https://github.com/openshift/console/pull/6194) opened
 * 2020-04-23 - [UXD exploration of i18next](https://github.com/jschuler/console/pull/1) opened
 * 2020-04-13 - [UXD analysis and comparison of libraries](https://github.com/patternfly/patternfly-react/issues/3952#issuecomment-613645855) posted
+* 2021-03-25 - [PR against oauth-server](https://github.com/openshift/oauth-server/pull/71) opened
 
 ## Drawbacks
 


enhancements/console/internationalization.md
```diff
@@ -110,6 +110,8 @@ The desired fallback experience if not all pieces/parts of the UI are translated
 
 We will also modify our existing file conversion tools to handle translation files spread across multiple packages so we can easily export/import translation files.
 
+Translation of the login templates within [oauth-templates](https://github.com/openshift/oauth-templates) and [oauth-server](https://github.com/openshift/oauth-server) repositories will be manually updated by the console team since their are only a small number of strings that rarely change. These strings are to be included in the console repository so that the i18next-parser will add them in the files sent to the Globalization team for translation.
```
spadgett, line 113: We should add more detail here like how we are using the `Accept-Language` header and optional `lng` query parameter to determine what language to show on the backend. We might want @jhadvig to contribute that detail since he did the backend work. We should also discuss if/how we plan to handle the language user preference for the login page. (See discussion in https://github.com/openshift/oauth-server/pull/71)

sg00dwin: @spadgett The text has been updated with additional detail provided by @jhadvig

jhadvig:
@stlaz PTAL

enhancements/console/internationalization.md
```diff
@@ -110,6 +110,15 @@ The desired fallback experience if not all pieces/parts of the UI are translated
 
 We will also modify our existing file conversion tools to handle translation files spread across multiple packages so we can easily export/import translation files.
 
+Translation of the login page will be handled by [oauth-server](https://github.com/openshift/oauth-server), which will check firstly the request URL for `lng` query. If the query is missing, `oauth-server` will check for request `Accept-Language` header. If both the query and header are missing in the request, `oauth-server` will differ to english language for localization.
```
spadgett, line 113: s/differ/default

enhancements/console/internationalization.md
```diff
@@ -110,6 +110,15 @@ The desired fallback experience if not all pieces/parts of the UI are translated
...
+
+[oauth-server](https://github.com/openshift/oauth-server) contains default login templates. Custom templates are stored in [oauth-templates](https://github.com/openshift/oauth-templates) repository.
```
spadgett, line 115:
```suggestion
[oauth-server](https://github.com/openshift/oauth-server) contains the default OKD login templates. Custom templates are stored in the [oauth-templates](https://github.com/openshift/oauth-templates) repository and will also need to be updated for the other brands like the OCP and OpenShift Dedicated login pages.
```

enhancements/console/internationalization.md
```diff
@@ -110,6 +110,15 @@ The desired fallback experience if not all pieces/parts of the UI are translated
...
+
+Since there are only a small number of strings, that rarely change, the localized strings for each of the supported languages will be directly part of the `oauth-server`, so we don't have to load them during runtime.
```
spadgett, line 117: I'm a little unclear what this is saying.

spadgett (commented):
Thanks @sg00dwin and @jhadvig. We should get approval from @stlaz

enhancements/console/internationalization.md
jhadvig: @spadgett it's referring to https://github.com/openshift/oauth-server/blob/60b81ecc54f02b6f2fd4e9bba36c06888bee0272/pkg/server/login/locales.go#L41-L50

Maybe to make it clear we could say:
```suggestion
Since there are only a small number of strings, that rarely change, the localized strings for each of the supported languages will be directly hardcoded into the `oauth-server`, so we don't have to load them during runtime.
```

enhancements/console/internationalization.md
sttts: Why `lng` query **and** header? Why is the latter not enough?

stlaz (changes requested):

enhancements/console/internationalization.md
spadgett: > Why `lng` query **and** header? Why is the latter not enough?

@sttts We need to be able to override the system setting since we have a user preference for selecting a language in console. If there is a standard query parameter, we can definitely use that instead.

I wouldn't drop the `Accept-Language` header because we want the default to be the system language when the user hasn't explicitly selected something different as a language preference in console.

enhancements/console/internationalization.md
spadgett: We use https://www.i18next.com/, but this seems heavy weight for a handful of strings in a static HTML page. I'm not sure how this would work with login page customization as well, since those pages would need to include the script. It's something we could explore though.

enhancements/console/internationalization.md
stlaz: > I wouldn't drop the Accept-Language header because we want the default to be the system language when the user hasn't explicitly selected something different as a language preference in console.

That's fair. I would also expect the `ui_locales` to configure the session if set but I see no cookie handling here. Also, I am missing a description of the UI element that allows configuring this on the oauth-server side.

enhancements/console/internationalization.md
stlaz: What's the plan going forward, then? This only localizes the console, but the information about the cluster state (operator statuses, for example) are still going to be in English. Is that ok or do you expect these to be translated eventually, too?

If they _are_ going to be translated, we'd need a bit more modularity in those translations. But I guess for now we can go with what you're proposing.

jhadvig: Agreed, since there are just a handful of strings I would prefer to put them directly into a go pkg, as is currently proposed.

enhancements/console/internationalization.md
jhadvig: Thats a good question. I think for now it has to be manual.

enhancements/console/internationalization.md
spadgett: > What's the plan going forward, then? This only localizes the console, but the information about the cluster state (operator statuses, for example) are still going to be in English. Is that ok or do you expect these to be translated eventually, too?

We don't currently have any plan to translate data coming from k8s resources, only the messages in console itself.

enhancements/console/internationalization.md
spadgett: I think console would store the last selected language in browser `localStorage` and then add `ui_locales` to the request when redirecting the user to login page. I haven't looked closely at the spec, though. Is this a query parameter?

jhadvig: @spadgett thoughts?

jhadvig:
@stlaz @sttts @spadgett is there anything else needed? Or can we merge this one ?

enhancements/console/internationalization.md
```diff
@@ -110,6 +110,14 @@ The desired fallback experience if not all pieces/parts of the UI are translated
 
 We will also modify our existing file conversion tools to handle translation files spread across multiple packages so we can easily export/import translation files.
 
+Translation of the login page will be handled by [oauth-server](https://github.com/openshift/oauth-server), which will check firstly the request URL for `ui_locales` query. If the query is missing, `oauth-server` will check for request `Accept-Language` header. If both the query and header are missing in the request, `oauth-server` will default to English language for localization.
+
+[oauth-server](https://github.com/openshift/oauth-server) contains the default OKD login templates. Custom templates are stored in the [oauth-templates](https://github.com/openshift/oauth-templates) repository and will also need to be updated for the other brands like the OCP and OpenShift Dedicated login pages.
```
sttts, line 115: do we need a bump of oauth-server when the templates change? I assume the console team will do those?

jhadvig: yes, if any of the templates change we will make needed changes.
Will incorporate that into the enhancement

jhadvig:
@sg00dwin based on chat with @stlaz and @sttts we agreed that we will drop the `ui_locales` param check. Also we should add to the enhancement's text that the ConsoleTeam is reponsible for updating the oauth-server repo for any i18n related changes.

enhancements/console/internationalization.md
```diff
@@ -110,6 +110,14 @@ The desired fallback experience if not all pieces/parts of the UI are translated
 
 We will also modify our existing file conversion tools to handle translation files spread across multiple packages so we can easily export/import translation files.
 
+Translation of the login page will be handled by [oauth-server](https://github.com/openshift/oauth-server), which will check firstly the request URL for `ui_locales` query. If the query is missing, `oauth-server` will check for request `Accept-Language` header. If both the query and header are missing in the request, `oauth-server` will default to English language for localization.
```
jhadvig, line 113:
```suggestion
Translation of the login page will be handled by [oauth-server](https://github.com/openshift/oauth-server), which will check for request `Accept-Language` header. If the header is missing in the request, `oauth-server` will default to English language for localization.
```

enhancements/console/internationalization.md
jhadvig: In case of any changes in these repositories, Console-Team will be responsible for syncing those changes between them.

jhadvig (changes requested):
@sg00dwin added two comments with changes that are needed.

enhancements/console/internationalization.md
sg00dwin: updated text

PR #733, opened 2021-04-15
build: Support Mounted Resource Volumes

adambkaplan:
Proposal to support mounts for Secrets and ConfigMaps in builds. This
augments prior work which let Secrets and ConfigMaps be used as sources
in builds. Unlike the current approach, this enhancement leverages
buildah's volume mount feature to let content be available only at build
time, and not at runtime. This is useful for builds that need to access
private artifact repositories or RHEL subscription content.

This current proposal builds on top of Ben Parees's original volume
mount proposal. Supported volume types will be gated at the API level to
ensure the volume type does not pose a security risk and is correctly
lifecycled. To distinguish Kubernetes volumes+mounts from buildah's
volume mount mechanism, the terms "volume content," "input volume,"
"buildah volume mount," and "buildah runtime environment" were
introduced. The proposed API uses these terms to distinguish build
volume mounts from pod volume mounts. Documentation requirements for
this feature were added in the "Drawbacks" section of the proposal.

Because the build controller uses the privileged security context
constraint, build pods to bypass most security features in OpenShift.
Therefore, future volume mounts in build pods need to be tightly
controlled. Open-ended volume types like CSI volumes could be an attack
vector if a developer uses an insecure CSI driver implementation. This
proposal establishes graduation criteria for adding new volume source
types to builds, including security, testing, lifecycle concerns,
failure modes, and feature gating.

### DIFF
--- a/enhancements/builds/volume-mounted-resources.md
+++ b/enhancements/builds/volume-mounted-resources.md
@@ -1,0 +1,395 @@
+---
+title: volume-mounted-resources
+authors:
+  - @bparees
+  - @adambkaplan
+reviewers:
+  - @derekwaynecarr
+  - @smarterclayton
+  - @deads2k
+approvers:
+  - @bparees
+  - @deads2k
+  - @sttts
+creation-date: 2020-01-08
+last-updated: 2021-05-05
+status: implementable
+---
+
+# Volume Mounted Resources
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [x] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Today builds support getting source input from configmaps and secrets, hereafter referred to as "resources".
+When users utilize this feature, the resource is volume-mounted into the build pod and placed in the "build context" within the build's execution environment, alongside other build sources like git source code.
+The next steps depend on whether it is an s2i or dockerfile build.
+
+For s2i builds, the generated Dockerfile contains commands to `ADD` the content at a path specified by the user, the assemble script is invoked, and then the injected content is zeroed out prior to committing the image via a `RUN rm` command added to the Dockerfile.
+
+For dockerfile builds, the user is instructed to add appropriate `ADD` and `RUN rm` commands to their dockerfile to inject the content that is available in the build's working directory (along with their application source, where applicable).
+
+There are a few undesirable aspects to this:
+
+1. In the dockerfile case, the content can still be found in lower layers of the image unless a layer squashing option is selected.
+2. Requires extra work by the user in the Dockerfile, so each Dockerfile must be customized
+
+This enhancement proposes to introduce an option to use buildah's capability to mount a volume at build time.
+The content mounted into the build pod would be then mounted into the container processing the Dockerfile, making that content available within the container so Dockerfile commands could reference it.
+No explicit `ADD` would be required, and since mounted content is not committed to the resulting image, no `RUN rm` equivalent is required to clean up the injected content.
+
+To avoid security and lifecycle concerns, the following volume types will be supported initially:
+
+1. Secrets
+2. ConfigMaps
+
+## Motivation
+
+### Goals
+
+* Simplify how users consume secret + configmap content in builds
+* Increase the security of protected content being injected to images
+* Simplify use cases that require consuming credentials during the build, but need to ensure those credentials do not end up in the output image.
+* Eventually extend this api to allow the mounting of other volumes (such as those backed by persistent storage)
+
+### Non-Goals
+
+* This enhancement should not result in a change of behavior for users of the existing secret/configmap injection api.
+* Provide immediate support for persistent volume claims in builds.
+  This is a long term goal that will be addressed in a future enhancement proposal.
+* Provide support to mount Secrets and ConfigMaps shared by the [projected resource CSI driver](/enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md).
+  This is a long term goal that will be addressed in a future enhancement proposal.
+
+## Proposal
+
+### User Stories [optional]
+
+The enabled use cases are essentially identical to what can be done with the configmap/secret input api in builds today, but with a better user experience and security as discussed above.
+It does not enable a new use case that is not already possible today, except that layer squashing will not be required.
+
+Future extensions to this enhancement could enable additional use cases, such as:
+
+- Persistent volumes that cache content
+- Shared Secrets and ConfigMaps, such a Simple Content Access certificate used to download RHEL content.
+
+These will be discussed in respective future enhancements.
+
+### Definitions
+
+The folllowing terms will be used in the remainder of the document to clarify the behavior of volume mounts in builds:
+
+- *Volume content*: contents used within a build that are not indended to be present in the resulting container image.
+- *Input Volume*: a Kubernetes Volume that is added to the build Pod, with the intent of being used as a transient bind mount within a buildah build.
+  See buildah's `--volume` [option](https://github.com/containers/buildah/blob/master/docs/buildah-bud.md#options).
+- *Buildah volume mount*: a directory that is added to Buildah's runtime environment as a transient bind mount.
+- *Container volume mount*: a volume that is mounted into a container within a Kubernetes pod.
+- *Buildah runtime environment*: the process that runs the buildah build within the build pod's main container.
+  In OCP 4.6 and higher, buildah uses OCI isolation and effectively runs as a sub-container within the build pod's main container.
+
+### Implementation Details/Notes/Constraints [optional]
+
+We will need to introduce a new mechanism in the build api which allows the user to indicate that they want to inject *volume content* into the build.
+Unlike source content, *volume content* is not intended to be included in the container image produced by an OpenShift build.
+Initially the only allowed volume types will be ConfigMaps and Secrets.
+The API will otherwise be similar to the existing secret/configmap injection api in which users identify the configmap/secret and the target path for injection.
+
+#### Build/BuildConfig API
+
+Volume content can be declared via a new `[]BuildVolume` field that is added to the Source and Docker strategy structs.
+These will define the *input volumes* that are appended to other volumes in the build pod.
+Items in the `[]BuildVolume` array will be translated directly into a pod `Volume`, using Kubernetes mechanisms to set up the volume for consumption by the build pod.
+
+Unlike Kubernetes, the types of volume sources that can be defined will be restricted to those explictly supported by Builds.
+Build pods are created with the privileged security context constraint, which allows a pod to mount any volume type with no SELinux restrictions.
+This SCC is inherited from the build controller.
+Restricting supported volume sources in the API ensures that builds only allow safe volume types to be mounted.
+
+The subset of supported volume sources will be defined by the `BuildVolumeSource` object.
+This will behave like the Kubernetes `VolumeSource` object, with the addition of a type discriminator that is set on Build/BuildConfig admission (if not specified).
+The OpenAPI schema must ensure that one of the volume sources are set.
+Initially only `Secret` and `ConfigMap` will be supported as valid Volume sources.
+The volume sources themselves will use Kubernetes volume source types, as the values in these volume sources will be passed directly to the build pod.
+
+The destination *buildah volume mount* will be controlled by the new `[]BuildVolumeMount` field, which is a subset of the Kubernetes `VolumeMount` API.
+The options in `BuildVolumeMount` will be constrained to those allowed in the *buildah volume mount*:
+
+- `MountPath` - the destination directory to place the *buildah volume mount* within buildah's runtime environment.
+- `ReadOnly` - whether or not the volume mount within buildah's runtime environment is read-only.
+
+Other options like `SubPath` and `MountPropagation` may be considered for support in the future.
+
+When the build pod is constructed, all *input volumes* will be mounted in a subdirectory of a fixed location in the build pod's main container - `/var/run/openshift.io/volumes`.
+The OpenShift builder process will infer the destination directory of these volumes inside the *buildah runtime environment* from the `Build` API object (already available via an environment variable, serialized as JSON).
+The logic that invokes buildah will then pass the buildah volume mounts as transient bind mount arguments.
+
+#### API Example
+
+```go
+// SourceBuildStrategy defines input parameters specific to an Source build.
+type SourceBuildStrategy struct {
+  
+  // existing API
+  ...
+  // Note that these fields will also be added to the DockerBuildStrategy API
+
+  // Volumes is a list of input volumes that can be mounted into buildah's runtime environment.
+  // Only a subset of Kubernetes Volume sources are supported by builds.
+  // More info: https://kubernetes.io/docs/concepts/storage/volumes
+  Volumes []BuildVolume
+
+  // VolumeMounts are volumes to mount into buildah's runtime environment at the specified mount path.
+  VolumeMounts []BuildVolumeMount
+}
+```
+
+```go
+// BuildVolume describes a volume that is made available to build pods, such that it can be mounted into buildah's runtime environment.
+// Only a subset of Kubernetes Volume sources are supported by builds.
+type BuildVolume struct {
+
+  // Volume's name.
+	// Must be a DNS_LABEL, unique within the pod, and cannot collide with volumes that are always added by the build controller.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+	Name string 
+
+	// BuildVolumeSource represents the location and type of the mounted volume.
+	BuildVolumeSource `json:",inline" protobuf:"bytes,2,opt,name=volumeSource"`
+}
+
+type BuildVolumeSourceType string
+
+const (
+  BuildVolumeSourceTypeSecret = "Secret"
+  BuildVolumeSourceTypeConfigMap = "ConfigMap"
+)
+
+// Represents the source of a volume to mount.
+// Only one of its members may be specified.
+// This is a subset of the core Kubernetes VolumeSource.
+type BuildVolumeSource struct {
+
+  // Type is the type for the volume source.
+  // Type must match the populated volume source, and if not specified will be inferred on admission.
+  // Only one type of volume source can be specified.
+  Type BuildVolumeSourceType
+
+  // Secret represents a secret that should populate this volume.
+  // More info: https://kubernetes.io/docs/concepts/storage/volumes#secret
+  // +optional
+  Secret *kapi.SecretVolumeSource
+
+  // ConfigMap represents a configMap that should populate this volume
+  // +optional
+  ConfigMap *kapi.ConfigMapVolumeSource
+}
+```
+
+```go
+// BuildVolumeMount describes a mounting of a Volume within buildah's runtime environment.
+type BuildVolumeMount struct {
+  // This must match the Name of a BuildVolume, and cannot collide with volumes that are always added to the build pod by the build controller.
+  Name string
+  // Mounted read-only if true, read-write otherwise (false or unspecified).
+  // Defaults to false.
+  // +optional
+  ReadOnly bool
+  // Path within the build runtime environment at which the volume should be mounted.
+  // Must not contain ':', and cannot collide with a destination path generated by the builder process.
+  MountPath string
+}
+```
+
+Example usage (use of the existing secret/configmap injection api is included for comparison, it is not changing):
+
+```yaml
+apiVersion: v1
+items:
+- apiVersion: build.openshift.io/v1
+  kind: BuildConfig
+  metadata:
+    name: mybuild
+    namespace: p1
+  spec:
+    strategy:
+      sourceStrategy:
+        from:
+          kind: ImageStreamTag
+          name: nodejs:10-SCL
+          namespace: openshift
+        volumes:
+        - name: secret
+          secret:
+            secretName: somesecret
+        - name: config
+          configMap:
+            name: someconfigmap
+            items:
+            - key: somekey
+              path: volume/path/value.txt
+        volumeMounts:
+        - name: config
+          mountPath: /tmp/config
+        - name: secret
+          mountPath: /tmp/secret
+      type: Source
+    source:
+      secrets:
+      - secret: 
+          name: myOtherSecret
+        destinationDir: /tmp/othersecret
+      configMaps:
+      - configMap:
+          name: myOtherConfigMap
+        destinationDir: /tmp/otherconfig
+```
+
+#### Volume Name and Destination Collisions
+
+Builds today have several container volume mounts that are provided by the build controller, including:
+
+- Node pull secrets (host)
+- System configuration (ConfigMap)
+- Certificate authorities for the cluster proxy and internal registry (ConfigMap)
+- Blob metadata cache (host)
+- RHEL entitlements (host via cri-o configuration)
+
+In addition, the builder process generates the following buildah volume mounts:
+
+- RHEL entitlements (destination is `/run/secrets/etc-pki-entitlement`, `/run/secrets/redhat.repo`, and `/run/secrets/rhsm`)
+- Custom PKI trust bundle - destination is `/etc/pki/ca-trust`, added via the Build/BuildConfig's `mountTrustedCA` option.
+
+If an input volume's name collides with a volume created by the build controller, or implied by OpenShift's cri-o configuration, the build should fail to run.
+Automatically avoiding volume name collisions - for example, by adding randomized suffixes - can be addressed in a future enhancement.
+
+If a volume mount path collides with a buildah volume mount generated by the builder process, the build should fail to run.
+This is unavoidable - we cannot have two mount points share the same destination in the build.
+
+#### Failure behavior
+
+A BuildConfig object can reference a Secret or ConfigMap in a volume that does not exist in the BuildConfig's namespace (yet).
+If a Build is generated from a BuildConfig and a referenced Secret or ConfigMap does not exist, the build controller should report a failure status and not create a build pod.
+This behavior should also apply to Secrets or ConfigMaps referenced in the build source array.
+
+### Risks and Mitigations
+
+**Risk:** Build mounts can alter the behavior of the build itself (a security risk).
+
+*Mitigations:*
+
+- Supported volume types are gated by the API.
+  For example, `HostPath` volume mounts are not supported.
+- Input volumes are mounted in a fixed location within the build pod's container.
+  The container volume is created by the build controller and cannot be changed by an end user.
+- Builds fail to run if a volume name collides with a volume generated by the build controller.
+- Builds fail if the a volume mount path collides with a destination generated by the builder process.
+
+**Risk:** Supporting arbitrary volume sources can lead to privilege escalations.
+This is a particular concern for builds since build pods run privileged.
+
+*Mitigations:*
+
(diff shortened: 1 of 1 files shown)


enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,298 @@
...
+The enabled use cases are essentially identical to what can be done with the configmap/secret input api in builds today, but with a better user experience and security as discussed above
+It does not enable a new use case that is not already possible today, except that layer squashing will not be required.
+
+Future extensions to this enhancement could enable the use case of providing build input content from a persistent volume and allowing the build to store/cache content for future builds on such a volume.
+Those will be discussed in the future enhancement at that time.
+
+### Implementation Details/Notes/Constraints [optional]
+
+We will need to introduce a new mechanism in the build api which allows the user to indicate that they want to inject "volume" content into the build.
+Initially the only allowed volume types will be configmaps and secrets.
```
bparees, line 89: and CSI volumes

bparees (commented):
one nit, otherwise looks as good as the day i wrote it :)

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,298 @@
...
+
+  // VolumeMounts are volumes to mount into the build
+  VolumeMounts []VolumeMount
+}
+```
+
+```go
+// BuildVolume describes a volume that is made available to build pods, such that it can be mounted into the build execution environment.
+// Only a subset of Kubernetes Volume sources are supported by builds.
+type BuildVolume struct {
```
bparees, line 153: doing this, instead of reusing the k8s volume type and adding validation/constraints that control which volumetypes are allowed may be a sticking point for people like @deads2k or @smarterclayton 

personally i'm ok with it (though it does carry a slightly higher maintenance burden for you), but i'd like to hear from them just in case they are fundamentally opposed.

adambkaplan: I decided to go this route because the real meat of the k8s Volume API is in the `VolumeSource` object, which to my understanding is a bag of pointers. We get the added maintenance burden of validating one and only one source type is set, which we don't _have_ to do on admission (though this is probably a good idea).

adambkaplan: On the flip side, we benefit by gating volume types in the API - invalid volume types fail fast.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+
+
+## Motivation
+
+### Goals
+
+* Simplify how users consume secret + configmap content in builds
+* Increase the security of protected content being injected to images
+* Simplify use cases that require consuming credentials during the build, but need to ensure those credentials do not end up in the output image.
+* Eventually extend this api to allow the mounting of traditional volumes (such as those backed by persistent storage)
```
gabemontero, line 66: should some reference to @bparees 's original volume mount proposal that you noted in the PR description be made here?

I'm assuming that is at least part of the future being referenced here

I'm also assuming there is no superceding or replacement of @bparees 's original EPs in this space.  Just some "augmentation"

But some brief text describing the relationships, overlapping, non-overlapping elements, even some sort of roadmap-ish flow seems useful and possibly even necessary here

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+
+```go
+// VolumeMount describes a mounting of a Volume within the build execution environment.
+type VolumeMount struct {
+  // This must match the Name of a Volume.
+  Name string
+  // Mounted read-only if true, read-write otherwise (false or unspecified).
+  // Defaults to false.
+  // +optional
+  ReadOnly bool
```
gabemontero, line 191: At least with the current projected resource / share / csi implementation we do not support read only because we have to copy the data from memory (the controller cache) to the mount location after the mount happens... we are not mounting from one existing fs to another like can occur in the more general volume cases

And this is on initial creation, not just updating later on if the secret or configmap changes and the controller receives an event for it.

So minimally this godoc, and the progression of things in general, would have to note that caveat for our new driver

Or we talk about further features and complications to https://github.com/openshift/csi-driver-projected-resource to manage multiple tmpfs  ... one we can write to / update and the one that is used for the volume the driver returns back to the kubelet

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+  // Path within the build environment at which the volume should be mounted.  Must
+  // not contain ':'.
+  MountPath string
```
gabemontero, line 194: So I almost posed this question above, but chose to do it here with the literal struct golang definition.

I'll copy the text from above for reference:

```
These fields will be wired, via the build controller, directly to the build pod that is constructed.
In addition all mounts into the pod will be done at a path of our choosing, not the VolumeMount path specified, to ensure the user cannot overwrite critical function inside the build pod and use it as an escalation pathway.
Example - all volumes are mounted under `/var/run/openshift.io/volumes` in the build pod containers.
```

So I'm trying to reconcile those words above, including `mounts into the pod will be done at a path of our choosing, not the VolumeMount path specified` with this presence of this field.

Is this a relative path off of `/var/run/openshift.io/volumes` then?  Or are we talking pod file system vs. build container context blah blah blah sort of thing, and this means transient mounts when we call buildah from openshift/builder ?

If so, use of `relative path` and `build context`  in the godoc above seems needed to me to make this clear.

If it is something other then relative path, I need more explanation in general.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+        - name: content-access
+          csi:
+            driver: csi-driver-projected-resource.openshift.io
+            volumeAttributes:
+              share: etc-pki-entitlement
+        volumeMounts:
+        - name: config
+          mountPath: /tmp/config
+        - name: secret
+          mountPath: /tmp/secret
```
gabemontero, line 234: yeah these don't seem like relative paths .... or again maybe it is a Pod FS vs. build context sort of thing

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+As additional types are requested, we will need to determine it is safe to add them.
+Gating will also ensure that we address any volume lifecycle concerns (for example - binding of PersistentVolumeClaims).
+
+## Design Details
+
+### Test Plan
+
+This feature will need new e2e tests that leverage the new api option.
+We have existing tests for configmap+secret injection would should be able to be copied+adapted to testing this feature relatively easily.
+Tests for the CSI volume may have to be done independently of the core OpenShift test suite at outset.
```
gabemontero, line 267: Yep until the move into the CVO via the storage operators happens, once openshift/api and openshift/client-go are updated with the new fields, we can vendor them into https://github.com/openshift/csi-driver-projected-resource and build upon the existing e2e's there to create BC's that use the new stuff.

gabemontero (changes requested):
Took a first pass at this @bparees @adambkaplan 

Ranges from some clarification on verbiage in godoc and proposal doc ... though I won't argue too too much if you all think I'm being overly precise.

The read only bit and some of the internal details of how the projected resource driver currently work probably will engender more discussion and possibly more changes to the driver at some point.

enhancements/builds/volume-mounted-resources.md
bparees: >  Or are we talking pod file system vs. build container context blah blah blah sort of thing, and this means transient mounts when we call buildah from openshift/builder ?

yeah, it's making that distinction.  We will choose the mount location within the build pod.  The user will choose the mount location within the buildah spawned container (the one that's executing the Dockerfile commands).

enhancements/builds/volume-mounted-resources.md
sttts: "validating one and only one source type is set" can be done purely in OpenAPI, no need for admission.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+	// +optional
+	Secret *kapi.SecretVolumeSource `json:"secret,omitempty" protobuf:"bytes,1,opt,name=secret"`
+	
+	// ConfigMap represents a configMap that should populate this volume
+	// +optional
+	ConfigMap *kapi.ConfigMapVolumeSource `json:"configMap,omitempty" protobuf:"bytes,2,opt,name=configMap"`
+
+	// CSI (Container Storage Interface) represents ephemeral storage that is handled by certain external CSI drivers (Beta feature).
+	// +optional
+	CSI *kapi.CSIVolumeSource `json:"csi,omitempty" protobuf:"bytes,3,opt,name=csi"`
```
sttts, line 179: OpenAPI to restrict to exactly one:
```
oneOf:
- required: ["secret"]
- required: ["configMap"]
- required: ["csi"]
```
This can be added to the CRD OpenAPI schema using a yaml patch (kubebuilder has no marker for that).

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+
+Proposed api/structs:
+Note:  DockerBuildStrategy will be updated in the same way.
+
+```go
+// SourceBuildStrategy defines input parameters specific to an Source build.
+type SourceBuildStrategy struct {
+  // From is reference to an DockerImage, ImageStream, ImageStreamTag, or ImageStreamImage from which
+  // the docker image should be pulled
+  From kapi.ObjectReference
```
deads2k, line 120: don't use object reference.  Create your own type to match.  See doc on https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5366-L5381

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+  PullSecret *kapi.LocalObjectReference
+
+  // Env contains additional environment variables you want to pass into a builder container.
+  Env []kapi.EnvVar
+
+  // Scripts is the location of Source scripts
+  Scripts string
+
+  // Incremental flag forces the Source build to do incremental builds if true.
+  Incremental *bool
```
deads2k, line 134: will this ever be defaulted to true?  Basically we learned over time that nothing is a bool and it's better to have an enum with something like `MakeStrategy` with values `{Full, Incremental}`

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+
+  // ForcePull describes if the builder should pull the images from registry prior to building.
+  ForcePull bool
```
deads2k, line 137: Is this really an `ImagePrePullPolicy` with values `Always`, `IfNotPresent`, and `Never` like on pods?

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+// SourceBuildStrategy defines input parameters specific to an Source build.
+type SourceBuildStrategy struct {
+  // From is reference to an DockerImage, ImageStream, ImageStreamTag, or ImageStreamImage from which
+  // the docker image should be pulled
+  From kapi.ObjectReference
+
+  // PullSecret is the name of a Secret that would be used for setting up
+  // the authentication for pulling the Docker images from the private Docker
+  // registries
+  PullSecret *kapi.LocalObjectReference
```
deads2k, line 125: Same comment about reusing the object reference types.

enhancements/builds/volume-mounted-resources.md
deads2k: > doing this, instead of reusing the k8s volume type and adding validation/constraints that control which volumetypes are allowed may be a sticking point for people like @deads2k or @smarterclayton

Having the local help and the constrained choice and avoiding automatic additions are all good properties you get with this approach.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+
+	// BuildVolumeSource represents the location and type of the mounted volume.
+	BuildVolumeSource `json:",inline" protobuf:"bytes,2,opt,name=volumeSource"`
+}
+
+
+// Represents the source of a volume to mount.
+// Only one of its members may be specified.
+// This is a subset of the core Kubernetes VolumeSource.
+type BuildVolumeSource struct {
```
deads2k, line 167: This looks like a discriminated union.  In order to add somethign safely in the future with an old controller running, you need a `VolumeType string` field.  That way the old controller can read the volumetype and realize he doesn't support the one chosen.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+	// Secret represents a secret that should populate this volume.
+	// More info: https://kubernetes.io/docs/concepts/storage/volumes#secret
+	// +optional
+	Secret *kapi.SecretVolumeSource `json:"secret,omitempty" protobuf:"bytes,1,opt,name=secret"`
```
deads2k, line 171: Be sure you really intend to keep up with this. Someone added specific keys to path mappings.  Will you support them?  If so, this is fine.  If this is a blind pass through, that could be ok as long as you're sure you don't accidentally change the mounting rules for users.

enhancements/builds/volume-mounted-resources.md
deads2k, line 179: How will this be controlled with respect to SCCs allowed for users?  I'm actually unsure today how CSI volumes are controlled via SCC

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,297 @@
...
+
+	// CSI (Container Storage Interface) represents ephemeral storage that is handled by certain external CSI drivers (Beta feature).
+	// +optional
+	CSI *kapi.CSIVolumeSource `json:"csi,omitempty" protobuf:"bytes,3,opt,name=csi"`
+}
+```
+
+```go
+// VolumeMount describes a mounting of a Volume within the build execution environment.
+type VolumeMount struct {
```
deads2k, line 185: For this struct, please get a review from the storage team.  I'm not as intimately familiar with volume mount problems like path traversals and readonly as they are.

enhancements/builds/volume-mounted-resources.md
adambkaplan: It is worth clarifying that this is the mount from the build container into buildah's sub-container. From that perspective we don't need to worry if the CSI driver does a writable volume mount, we can tell buildah to mount a directory into the build sub-container as read-only.

enhancements/builds/volume-mounted-resources.md
adambkaplan: This is a true co-authoriship - a big chunk of the text is Ben's original proposal. That never merged, I linked the two via GitHub comments.

enhancements/builds/volume-mounted-resources.md
adambkaplan: This is our existing API, I don't think we can change it without breaking things.

enhancements/builds/volume-mounted-resources.md
adambkaplan: Existing API, can't change without breaking users.

deads2k:
I like the way this API mirrors volumes.  I want to be sure

1. we don't accidentally escalate users by adding this feature - this is the dev-ex team's responsibility
2. that SCC can actually handle CSI volumes in some reasonable way - this the apiserver or storage team's responsibility

enhancements/builds/volume-mounted-resources.md
adambkaplan: removed CSI volumes :(

enhancements/builds/volume-mounted-resources.md
adambkaplan: updated API docs and proposal to distinguish "pod volume mount" from "buildah bind mount"

enhancements/builds/volume-mounted-resources.md
adambkaplan: Proposal and API doc updated to distinguish "pod volume mount" from "buildah volume mount"

enhancements/builds/volume-mounted-resources.md
adambkaplan: Existing API in Builds

enhancements/builds/volume-mounted-resources.md
adambkaplan: Existing API in builds

enhancements/builds/volume-mounted-resources.md
adambkaplan: Added the discriminator field. My understanding is that we should set/validate this field on admission, in the same way we do with the build strategy type discriminator. Is this correct?

enhancements/builds/volume-mounted-resources.md
adambkaplan: Due to concerns with SCCs, CSI volume sources were dropped from this EP. The build controller uses the privileged SCC to do "safe" host path mounts in build pods, and even with builds in user namespaces we will likely keep these host path mounts.

CSI volume sources can be reintroduced in the future. I think we can gate the allowed CSI driver for these volume types in one of two ways (perhaps not mutually exclusive):

- Enhance SCCs to allowlist drivers for `CSI` volume sources.
- Enhance the cluster Build controller config to allowlist drivers for CSI volume mounts.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,374 @@
...
+
+  // ConfigMap represents a configMap that should populate this volume
+  // +optional
+  ConfigMap *kapi.ConfigMapVolumeSource
+}
+```
+
+```go
+// BuildVolumeMount describes a mounting of a Volume within buildah's runtime environment.
+type BuildVolumeMount struct {
```
adambkaplan, line 194: /cc @nalind 

For mount options supported by buildah.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,374 @@
...
+## Design Details
+
+## Open Questions [optional]
+
+1. Can we make this the default or even only behavior for builds?
+   No, need to make it opt-in to avoid potentially breaking existing buildconfigs.
+2. What happens if a user overrides the default volume mounts used by Builds today?
+   We prevent the build from running.
+   Overriding default volume mounts can alter build behavior, which can be a security risk.
+3. Can cluster admins alter the default `mounts.conf` configuration for cri-o?
```
adambkaplan, line 313: /cc @mrunalp 

I believe we use `mounts.conf` to get RHEL entitlements into workloads from the host, which is relevant for a) RHEL7 nodes, and b) clusters that use MacineConfig to add RHEL entitlements to nodes.

In both situations, these are work-arounds for our lack of entitlements support in RHCOS.

adambkaplan:
> I want to be sure
> 
> 1. we don't accidentally escalate users by adding this feature - this is the dev-ex team's responsibility
> 2. that SCC can actually handle CSI volumes in some reasonable way - this the apiserver or storage team's responsibility

Based on our conversation in slack and my own investigation, we need additional build controller configuration to gate CSI volume mounts. I think this is worth pursuing in a separate enhancement proposal, since ephemeral CSI drivers will be a net-new thing for OpenShift.

Enhancing SCCs probably needs to be added to the original EP for the projected resource CSI driver. We can only allow an SCC to allow all or no CSI volume mounts, which is not sufficient protection.

enhancements/builds/volume-mounted-resources.md
bparees: so if CSI volumes are being dropped from the enhancement, how does this move us towards the shared entitlement cred use case?

(don't get me wrong, it's still a useful evolution of the api and configmap/secret volume mounting has existing use cases/demand)

adambkaplan: We don't get the benefit of shared entitlements, but we remove developer pain with the experience of adding secrets. Especially for the Docker strategy, where developers are on their own to remove the secret and squash layers.

Adding CSI volumes need to be added in a follow-up EP, where we will need to carefully consider how we secure builds with CSI volumes.

adambkaplan: (note - we still get the foundation of an API to add volumes. From a coding perspective this will be 80% of what we need to do to support the shared entitlement use case).

bparees: got it.

enhancements/builds/volume-mounted-resources.md
nalind: The library accepts options for controlling propagation ("shared", "rshared", "private", "rprivate", "slave", "rslave"), regular options ("rw, "ro", "dev"/"nodev", "suid"/"nosuid", "exec"/"noexec"), and SELinux labeling controls that I think end up being ignored since we're already in a container ("Z", "z").
There are also a couple of custom ones: "U" to have the ownership of the volume's contents changed to match the current USER value, and "O" which causes the volume source to be used as the lower layer for an overlay mount at the mount point, using a per-run temporary directory as the upper (write) layer.  I don't know that either of those is particularly useful in builds, and of course their functionality willl be limited in the future by the permissions that the builder binary has.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,377 @@
...
+- Blob metadata cache (host)
+- RHEL entitlements (host via cri-o configuration)
+
+In addition, the builder process generates the following buildah volume mounts:
+
+- RHEL entitlements (destination is `/run/secrets/etc-pki-entitlement`, `/run/secrets/redhat.repo`, and `/run/secrets/rhsm`)
+- Custom PKI trust bundle - destination is `/etc/pki/ca-trust`, added via the Build/BuildConfig's `mountTrustedCA` option.
+
+If an input volume's name collides with a volume created by the build controller, or implied by OpenShift's cri-o configuration, the build should fail to run.
+If a volume mount path collides with a buildah volume mount generated by the builder process, the build should fail to run.
```
nalind, line 277: Currently, when multiple volumes are set to be mounted at a given location, we choose one of them and proceed.  This will require some patching.

adambkaplan: I think this is something we can do in openshift/builder or in the build controller. Buildah itself doesn't have to fail here (unless you believe this is buggy behavior).

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,377 @@
...
+
+// Represents the source of a volume to mount.
+// Only one of its members may be specified.
+// This is a subset of the core Kubernetes VolumeSource.
+type BuildVolumeSource struct {
+
+  // Type is the type for the volume source.
+  // Type must match the populated volume source, and if not specified will be inferred on admission.
+  // Only one type of volume source can be specified.
+  Type BuildVolumeSourceType
```
coreydaley, line 182: @adambkaplan Do we really need this since we can infer the type from whether or not `Source` or `ConfigMap` are nil? Then we can just kick it back if both are not nil? (or maybe use a validation?)

adambkaplan: Yes - this is a type discriminator that helps us address controller version skew. As we add more supported volume types, we will run into a situation on upgrade where the apiserver will support new volume types, but the controller won't. This field will need to be set by openshift-apiserver if it is empty, in the same way we set the strategy type.

enhancements/builds/volume-mounted-resources.md
mrunalp: @adambkaplan Today we don't allow cluster admins to modify the mounts.conf through ContaineRuntineConfig CRD. They can of course directly modify any such file using the MCO but that is unsupported.

enhancements/builds/volume-mounted-resources.md
adambkaplan: Good to know - we will start with ReadOnly `ro` as the only supported option.

adambkaplan:
@bparees @deads2k @sttts @mrunalp all comments have been addressed.

Final call for review - otherwise I think this is good to merge.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,391 @@
...
+          configMap:
+            name: someconfigmap
+            items:
+            - key: somekey
+              path: volume/path/value.txt
+        - name: content-access
+          csi:
+            driver: csi-driver-projected-resource.openshift.io
+            volumeAttributes:
+              share: etc-pki-entitlement
```
bparees, line 241: this is no longer valid/applicable since csi volume types aren't supported.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,391 @@
...
+        volumeMounts:
+        - name: config
+          mountPath: /tmp/config
+        - name: secret
+          mountPath: /tmp/secret
+        - name: content-access
+          mountPath: /etc/pki/entitlement
```
bparees, line 248: remove this as well

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,391 @@
...
+- Certificate authorities for the cluster proxy and internal registry (ConfigMap)
+- Blob metadata cache (host)
+- RHEL entitlements (host via cri-o configuration)
+
+In addition, the builder process generates the following buildah volume mounts:
+
+- RHEL entitlements (destination is `/run/secrets/etc-pki-entitlement`, `/run/secrets/redhat.repo`, and `/run/secrets/rhsm`)
+- Custom PKI trust bundle - destination is `/etc/pki/ca-trust`, added via the Build/BuildConfig's `mountTrustedCA` option.
+
+If an input volume's name collides with a volume created by the build controller, or implied by OpenShift's cri-o configuration, the build should fail to run.
```
bparees, line 276: we should be able to generate our own volume names(both on the pod and in buildah) such that we avoid collisions.  We don't have to use the volume name the user provides, when setting up volumes.

enhancements/builds/volume-mounted-resources.md
```diff
@@ -0,0 +1,391 @@
...
+If a volume mount path collides with a buildah volume mount generated by the builder process, the build should fail to run.
+
+#### Failure behavior
+
+A BuildConfig object can reference a Secret or ConfigMap in a volume that does not exist in the BuildConfig's namespace (yet).
+If a Build is generated from a BuildConfig and a referenced Secret or ConfigMap does not exist, the build controller should report a failure status and not create a build pod.
+This behavior should also apply to Secrets or ConfigMaps referenced in the build source array.
+
+### Risks and Mitigations
```
bparees, line 285: is there a risk that a user can create a build that mounts a secret or configmap that they themselves don't have permission to view?

i think we mitigate that in the existing secret injection logic, though i don't recall how.  We need to ensure it is mitigated here as well.

bparees (changes requested):

enhancements/builds/volume-mounted-resources.md
adambkaplan: I don't see any such logic in the build controller. Secrets and ConfigMap volume mounts are local object references in the pod spec. My understanding is that if a Secret exists in a namespace, it can be used by any service account in said namespace.

bparees: i guess it is constrained by "you can't create builds if you can't create pods"

enhancements/builds/volume-mounted-resources.md
adambkaplan: True about creating our own volume names, though we get into fun situations where we need to validate/truncate superficially long volume names.

adambkaplan: @bparees I made this a "we can change this in a future enhancement" given the short window we have to land this in 4.8

bparees: btw i wasn't suggesting random suffixes.  you don't have to use any part of the user given name, you can just call it "vol1", "vol2", etc.  So no issue w/ truncation.  Again we already do this when mounting the secrets/configmaps into the build pod.  (generate our own volume names)

PR #738, opened 2021-04-19
Installing OCP on ARM-Based Smart NICs

danwinship:
Pushing this to start some initial discussion

----

### Summary

As servers have greater and greater networking needs, hardware vendors have created increasingly more complex and powerful network cards. The latest iteration of this is "Smart NICs" that include an entire ARM system running Linux on the NIC. With such a NIC, we can offload nearly all network processing from the host computer onto the NIC "DPU" (Data Processing Unit), freeing up host CPU resources for customer workloads (and allowing faster data transfer by ensuring the network stack doesn't need to compete for CPU with the customer workload).

However, doing this requires installing and managing an appropriate software stack on all of the NICs in the cluster. This software will need to be kept in sync with the software on the host PC (eg, as new
functionality is added to OCP). The obvious answer to this problem is to run OpenShift on the NIC as well as on the host.

----

cc @zshi-redhat @dcbw @russellb

### DIFF
--- a/enhancements/network/dpu/ocp-on-dpu.md
+++ b/enhancements/network/dpu/ocp-on-dpu.md
@@ -1,0 +1,311 @@
+---
+title: DPU Support in OCP
+authors:
+  - "@danwinship"
+reviewers:
+  - "@derekwaynecarr"
+  - "@dcbw"
+  - "@zshi-redhat"
+approvers:
+  - "@derekwaynecarr"
+  - "@dcbw"
+creation-date: 2021-09-16
+last-updated: 2021-04-15
+status: provisional
+---
+
+# DPU Support in OCP
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+See [DPU Support in OCP (Overview)]. This document describes the
+overall architecture, installation, and upgrades.
+
+[DPU Support in OCP (Overview)]: ./overview.md
+
+## Motivation
+
+See Overview doc.
+
+### Goals
+
+- Create a "Developer Preview" version of the feature for OCP 4.10
+  with the following limitations:
+
+  - This will be difficult to install as there will be no explicit
+    support anywhere for installing the infra and tenant clusters
+    as a unit.
+
+  - Upgrades to the infra cluster will result in network outages in
+    the tenant cluster, as there will be no coordination between
+    infra cluster and tenant cluster reboots.
+
+  - The infra cluster will require separate ARM master nodes. We may
+    support running these as VMs on an x86 host.
+
+  - There will be no support for running customer workloads of any
+    sort in the infra cluster.
+
+  - It will not necessarily be possible to upgrade from Dev Preview
+    to Tech Preview without reinstalling one or both clusters.
+
+- Create further Tech Preview and GA versions of the feature with
+  improved functionality.
+
+### Non-Goals
+
+- ...
+
+## Proposal
+
+### User Stories
+
+See Overview doc.
+
+### API Extensions
+
+None for now
+
+### Risks and Mitigations
+
+#### Hardware Availability
+
+We are currently developing using pre-GA BlueField-2 hardware, with a
+plan based on official descriptions of how the final GA BlueField-2
+hardware will work. So far (presumably due to the standard global chip
+supply chain problems?) we have been unable to get GA cards (either
+for free or by paying for them). If we are unable to get GA cards
+soon, it may not be possible to ship the Dev Preview as designed for
+OCP 4.10.
+
+#### OCP on ARM
+
+Some parts of OCP-on-bare-metal are not currently available for ARM.
+In particular, I think we don't have "ironic" images, which are needed
+for an IPI install.
+
+## Design Details
+
+### Design Details - Dev Preview
+
+The initial dev preview will be a very limited implementation, to
+demonstrate the overall architecture.
+
+#### Overall Cluster Architecture
+
+FIXME: fill this in:
+
+- DPU Host and DPU NIC management interfaces
+- How to connect the DPU to the network
+  - How infra cluster and tenant cluster traffic is distinguished;
+    VLANs? Parallel subnets?
+- Difference between network setup for provisioning vs network setup
+  at runtime?
+
+#### Installation
+
+Installing a cluster with BF-2 cards is potentially tricky, since the
+cards can be configured in multiple modes, and in some modes, may not
+allow their host to access the network by default.
+
+To install the cluster then, it will be necessary to:
+
+- Install the tenant cluster masters and at least one worker node that
+  is not a DPU Host. Install the SR-IOV Operator in the tenant
+  cluster, with the proper configuration (FIXME: what is the proper
+  configuration?).
+
+- Create a MachineConfigPool in the tenant cluster for the DPU Host
+  workers, and update the Cluster Network Operator configuration to
+  indicate that this pool is for DPU Host workers.
+
+- Power on, but don't yet install, the DPU Hosts in the tenant
+  cluster (so that the BF-2s themselves will be powered on).
+
+- Install the infra cluster.
+
+- Create a MachineConfigPool in the infra cluster for the DPU NIC
+  workers, and update the Cluster Network Operator configuration to
+  indicate that this pool is for DPU NIC workers.
+
+- Install the DPU Operator (see below) onto the infra cluster,
+  providing it with an appropriate tenant-cluster kubeconfig.
+  Initially, this will just ensure that each of the cards is
+  configured in "passthrough" mode, allowing the tenant hosts
+  unrestricted network access.
+
+- FIXME something must ensure that standalone kube-proxy runs on infra
+  cluster BF-2 workers to provide service resolution. (Maybe CNO does
+  this based on the infra DPU NIC MCP?)
+
+- Add the DPU Host workers to the tenant cluster. The combination of
+  the CNO and SR-IOV Operator config in the tenant cluster and the DPU
+  Operator in the infra cluster will cause each DPU Host worker to be
+  configured to run ovn-kubernetes in "`dpu-host`" mode with its
+  corresponding DPU NIC worker running ovn-kubernetes in "`dpu`" mode.
+
+  Setting this up correctly requires that DPU NICs can match
+  themselves with their corresponding DPU Hosts. We have not yet
+  determined how this will happen. It may require explicit
+  configuration in Dev Preview.
+
+#### Upgrade
+
+When upgrading to a new OCP release, each BF-2 node in the infra
+cluster will need to reboot at some point. This will cause a temporary
+(~ 2 minute) network outage on the corresponding tenant cluster host.
+
+In the GA product, we will need to coordinate the upgrades between the
+two clusters, so that each NIC reboots just before its corresponding
+host, after the host has already been drained of pods. However, this
+functionality will not be implemented in the Dev Preview; upgrades
+between the two clusters will be uncoordinated, resulting in
+networking outages.
+
+#### OCP Modifications in the Tenant Cluster
+
+##### ovn-kubernetes (DPU Host mode)
+
+The modifications to ovn-kubernetes itself to support "DPU Host" mode
+have already been made upstream.
+
+##### cluster-network-operator
+
+CNO will need to know when to configure ovn-kubernetes in "DPU Host"
+mode. For Dev Preview, we are not yet ready to add proper
+`Network.operator.openshift.io` API for this, so we will instead have
+CNO react to an optional ConfigMap.
+
+##### Special Handling of BF-2 Host Nodes
+
+When running in "DPU Host" mode, ovn-kubernetes will configure all
+pod-network pods on a node to have SR-IOV interfaces rather than veths
+as their primary network interfaces. Because of how SR-IOV currently
+works in Kubernetes, this requires setting certain fields in the
+`spec` of any pods that will be scheduled on those nodes. This means
+that "ordinary" SR-IOV-unaware pod-network pods cannot be scheduled to
+BF-2 host nodes.
+
+While this will eventually need to be fixed, this means that for Dev
+Preview, we need to ensure that OCP does not attempt to run any of its
+own pod-network pods on BF-2 host nodes. (eg, `dns-default` pods
+cannot run on these nodes). This may involve some combination of
+special configuration, and tainting of the nodes in the BF-2 host
+MachineConfigPool.
+
+#### OCP Modifications in the Infra Cluster
+
+##### ovn-kubernetes (DPU NIC mode)
+
+As with the tenant side, the changes to ovn-kubernetes itself to
+support "DPU NIC" mode have already been made upstream.
+
+Infra cluster masters and non-DPU worker nodes will run ovn-kubernetes
+(in "`full`" mode) as their CNI plugin. However, DPU NIC nodes will
+not be able to do this, since it is not currently possible to run
+ovn-kubernetes in both "`full`" mode and "`dpu-nic`" mode on the same
+host.
+
+The current plan for Dev Preview is that DPU NIC worker nodes will not
+run any CNI plugin, and will only support host-network pods.
+(Actually, they need _some_ CNI in order for the node to become Ready
+at all, so they will run a dummy CNI plugin that indicates to
+Kubelet/CRI-O that node networking is ready, but then fails every
+`CNI_ADD` call.)
+
+##### DPU Operator
+
+We will write a "[DPU Operator]" to handle coordination between the
+infra and tenant clusters, and setting up the special behavior for the
+BF-2 cards in the infra cluster. This will be available via OLM and
+installable on "day 2". See that enhancement for more details.
+
+[DPU Operator]: https://github.com/openshift/enhancements/pull/890
+
+##### Special Handling of BF-2 NIC Nodes
+
+Similarly to the tenant cluster, we will need to prevent OCP-internal
+pod-network pods from being scheduled to BF-2 NIC nodes, since there
+will not be a functioning CNI plugin / pod network on those nodes.
+Again, this may involve some combination of configuration and
+tainting.
+
+### Design Details - Tech Preview
+
+TBD
+
+### Design Details - GA
+
+TBD
+
+### Open Questions
+
+...
+
+### Test Plan
+
+...
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+
+#### Tech Preview -> GA
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+When downgrading to a version of OCP that does not support this
+feature, the administrator must first disable the feature and return
+the cluster to fully-non-offloaded operation.
+
+### Version Skew Strategy
+
+For Dev Preview, we do not promise to support y-stream version skew
+between DPU NICs and DPU Hosts (which implies that it may not be
+possible to update a cluster using the Dev Preview feature to the next
+y-stream release, since the clusters cannot be upgraded perfectly in
+sync). Z-stream version skew should hopefully not be a problem.
+
+In the final product, we will need to ensure we can handle some skew
+between the clusters at least during upgrades.
+
+### Operational Aspects of API Extensions
+
+N/A
+
+#### Failure Modes
+
+N/A
+
+#### Support Procedures
+
+N/A
+
+## Implementation History
+
+- 2021-04-19: Initial proposal
+- 2021-09-14: Split out overview and made this mostly just about Dev Preview for now.
+
+## Drawbacks
+
+## Alternatives
+
+## Infrastructure Needed
+
+Infrastructure both for initial development and for later CI will be
+provided as part of the Hardware Enablement project of the Network
+Services team. QE may eventually need some hardware of their own as
+well.
--- a/enhancements/network/dpu/overview.md
+++ b/enhancements/network/dpu/overview.md
@@ -1,0 +1,280 @@
+---
+title: DPU Support in OCP (Overview)
+authors:
+  - "@danwinship"
+reviewers:
+  - "@derekwaynecarr"
+  - "@dcbw"
+  - "@zshi-redhat"
+approvers:
+  - "@derekwaynecarr"
+  - "@dcbw"
+creation-date: 2021-04-15
+last-updated: 2021-09-15
+status: provisional
+---
+
+# DPU Support in OCP (Overview)
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+As servers have greater and greater networking needs, hardware vendors
+have created increasingly more complex and powerful network cards. The
+latest iteration of this is "DPUs" (Data Processing Units) that
+include an entire ARM system running Linux on the NIC. With such a
+NIC, we can offload nearly all network processing from the host
+computer onto the DPU, freeing up host CPU resources for customer
+workloads (and allowing faster data transfer by ensuring the network
+stack doesn't need to compete for CPU with the customer workload).
+
+However, doing this requires installing and managing an appropriate
+software stack on all of the NICs in the cluster. This software will
+need to be kept in sync with the software on the host PC (eg, as new
+functionality is added to OCP). The obvious answer to this problem is
+to run OpenShift on the NIC as well as on the host.
+
+The initial target of this work is the [NVIDIA BlueField-2 DPU].
+However, similar products are being worked on by other vendors which
+we will eventually want to support, so the solution should not be too
+specialized to this one product.
+
+[NVIDIA BlueField-2 DPU]: https://www.nvidia.com/en-us/networking/products/data-processing-unit/
+
+## Glossary
+
+- **Smart NIC** - a vaguely-defined term for an enterprise-class NIC
+  with extensive offload capabilities, but not necessarily with the
+  ability to run arbitrary user software. For example, the NVIDIA
+  Mellanox ConnectX-5 and ConnectX-6.
+
+- **DPU** / **Data Processing Unit** - a Smart NIC with a full CPU,
+  RAM, storage, etc, running a full operating system, able to offload
+  network processing and potentially other tasks from its host system.
+  For example, the NVIDIA BlueField-2.
+
+  (Some vendors other than NVIDIA are using the terms "Smart NIC" and
+  "DPU" slightly differently, but for now at least, this is the way we
+  are defining these terms.)
+
+- **DPU NIC** - refers to the DPU and its CPU, OS, etc (as opposed to
+  the CPU, OS, etc, of the x86 server that the DPU is installed in). A
+  **DPU NIC node** or **DPU NIC worker** is an OCP worker node
+  running on a DPU NIC.
+
+- **DPU Host** - refers to an x86 server (and its CPU, OS, etc) which
+  contains a DPU. A **DPU Host node** or **DPU Host worker** is an OCP
+  worker node running on a DPU Host.
+
+- **Two-Cluster Design** - any architecture for deploying an OCP
+  cluster with (some) nodes containing DPUs, where the DPUs are nodes
+  in a second OCP cluster. (For purposes of this term, the
+  architecture still counts as "two-cluster" even if HyperShift is
+  involved and there are actually three clusters.)
+
+- **Infra Cluster** - in a Two-Cluster Design, the OCP cluster which
+  contains the DPU NIC workers. (It may also include master nodes
+  and/or non-DPU workers, depending on the particular details of the
+  two-cluster design.)
+
+- **Tenant Cluster** - in a Two-Cluster Design, the OCP cluster
+  containing the DPU Host workers. (It may also include master nodes
+  and/or workers without DPUs.)
+
+## Motivation
+
+### User Stories
+
+#### Network Offload
+
+As a cluster administrator, I would like to offload network processing
+from the host CPU to the NIC, to increase throughput and to free up
+host CPU cycles for workloads rather than network processing.
+
+In particular:
+
+- We can move OpenFlow processing and (eventually) IPsec encryption
+  from the DPU Host CPU to the DPU. (These would be fully offloaded to
+  custom silicon in the DPU, not simply run as processes on the
+  ARM-based DPU.)
+
+- We can move all of the node-side OVS, OVN, and OVN Kubernetes
+  daemons from the DPU Host CPU to the DPU NIC CPU. (The host side
+  would retain only a CNI plugin binary.)
+
+- Is there interest in moving DNS or routers off the host CPU?
+
+- ...
+
+#### Security Offload
+
+As a cluster administrator, I would like to use a DPU to monitor its
+host node. The NIC would run in an isolated mode, where the host OS
+has no control over it. Possibly even the cluster administrators of
+the host's OCP cluster would have no control over the NIC.
+
+#### Storage Offload
+
+(FIXME: Details? NVIDIA talks about this case in their marketing
+materials but I haven't heard as much about it in an OCP context and
+maybe this isn't an important case for us?)
+
+The BlueField-2 cards have M.2 NVMe ports and can have directly
+attached SSDs.
+
+### Goals
+
+- Design a system for provisioning and maintaining a RHEL-based
+  software stack on DPUs to support features such as "[Smart NIC
+  OVN Offload]".
+
+  - Although this enhancement is currently written to assume that
+    this means running OCP on the NICs, this is not yet cast in
+    stone. Some other approaches are discussed under "Alternatives"
+    and one of those could potentially replace the OCP approach.
+
+  - Although not part of the initial design, we assume some
+    customers may be interested taking advantage of the NIC/host
+    separation to run additional security and monitoring
+    functionality on the NICs in the future. Thus, we are interested
+    in designs with more isolation between the NIC and the host,
+    even though that is not necessarily interesting for the primary
+    network-offload use case.
+
+- Design an appropriate architecture for an environment containing a
+  mix of x86_64 hosts with "dumb" NICs, x86_64 hosts with DPUs, and
+  the ARM-based DPUs themselves. (As discussed below, such an
+  environment may involve one or more OCP clusters.)
+
+- Define how OCP installs and upgrades will work in such an
+  environment.
+
+[Smart NIC OVN Offload]: https://github.com/openshift/enhancements/pull/732
+
+### Non-Goals
+
+- Support for network plugins other than OVN Kubernetes. The
+  architecture designed here may not be generically applicable and is
+  not guaranteed to be usable by other network plugins even if they
+  independently implement their own BlueField, etc, support.
+
+- Support for NICs that are called "Smart NICs" or "DPUs" but which
+  don't meet our requirements (below).
+
+### Supported Hardware
+
+We want to support DPUs that are vaguely "BlueField-2-ish". eg:
+
+  1. The NIC has a sufficiently-powerful ARM processor (or,
+     theoretically, a processor of some other architecture that RHCOS
+     supports), and sufficient RAM, storage, etc, to run as an OCP
+     node.
+
+       - Current BlueField-2 specs mention "Up to 8 Armv8 A72 cores"
+         and "8GB / 16GB / 32GB of on-board DDR4". It is possible we
+         would only support a subset of SKUs.
+
+       - The terms "Smart NIC" and "DPU" are also being used to refer
+         to NICs that just have programmable ASICs or FPGAs, rather
+         than a full computing environment. These are out of scope.
+
+  2. The NIC has a Baseboard Management Controller that supports IPMI,
+     Redfish, or some other protocol [supported by the
+     baremetal-operator], to control and provision it independently of
+     its host.
+
+  3. The NIC implements SR-IOV in such a way that we can connect
+     pods on the host directly to an OVN bridge on the NIC.
+
+  4. All of the above can be done without running any proprietary
+     drivers or third-party software on either the host's or the NIC's
+     RHCOS installation.
+
+[supported by the baremetal-operator]: https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#bmc
+
+## Proposal
+
+The details will be figured out in further enhancements.
+
+### Risks and Mitigations
+
+...
+
+## Design Details
+
+### Cluster Architecture
+
+The expected (tenant) cluster architecture at this time is:
+
+- Three master nodes, which do not contain DPUs.
+- Some number of worker nodes (maybe 0, but probably not) that do not contain DPUs.
+- A medium-to-large number of workers that do contain DPUs.
+
+(The initial Developer Preview will require at least 1 non-DPU worker
+node.)
+
+## Open Questions
+
+### API Extensions
+
+### Test Plan
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+
+#### Tech Preview -> GA
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+### Version Skew Strategy
+
+### Operational Aspects of API Extensions
+
+#### Failure Modes
+
+#### Support Procedures
+
+## Implementation History
+
+- 2021-04-19: Initial proposal
+- 2021-09-14: Simplified to be just an overview
+
+## Drawbacks
+
+## Alternatives
+
+### Running RHCOS But Not OCP on the NIC
+
+Rather than running a full OpenShift installation on the DPU, we could
+instead provision a simple RHCOS system from the node, and have each
+node manage its own DPU's software and configuration directly.
+(Indeed, this is how the proof of concept for OVN offload was
+implemented.)
+
+However, we expect customers to be interested in the greater degree of
+security provided by having the DPU NIC software be entirely isolated
+from the DPU Host software. Eventually, this can evolve into a system
+like AWS's "Nitro", where not just network processing, but also other
+security and cluster monitoring functionality is handled by the DPU,
+entirely outside the control of the host CPU.
+
+In a HyperShift environment, it is possible that the DPU NICs could be
+provisioned (with plain RHCOS) by something outside the control of the
+"tenant" cluster. (As far as I can tell, the complete architecture for
+bare-metal HyperShift has not been worked out yet, so it is not clear
+exactly how provisioning even of normal nodes will work there.)
+
+## Infrastructure


enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+
+(While the exact architectural constraints may change in the future,
+it is likely that some customers will always want the possibility of
+only having Smart NICs on a subset of nodes.)
+
+### Overall Cluster Architecture
+
+There are several possible overall architectures:
+
+  - The Smart NICs could be nodes within the same OCP cluster as their
```
derekwaynecarr, line 113: is there a motivation for this topology?  in prior discussions, this broke one of the intended benefits of ensuring strong security isolation between infrastructure and workload.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+    separate from the masters and workers.) This may require changes
+    to tolerations and node selectors on existing DaemonSets, both
+    within OCP and in third-party operators or customer workloads (and
+    it may turn out that this approach is unworkable for this reason).
+
+  - The Smart NICs could be part of a second cluster running in
+    parallel with the primary cluster, with the control plane of the
+    NIC cluster running as pods within the primary cluster.
+
+  - The Smart NICs could be part of a second cluster which is
```
derekwaynecarr, line 126: i had hoped to avoid mixed architectures in standalone topologies.

danwinship: I mean, that would make things simpler technically for us, but it seems like it would be _much_ more complicated administratively for the customer if they are not already using rackmount ARM servers.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
+---
+title: Installing OCP on ARM-Based Smart NICs
+authors:
+  - "@danwinship"
+reviewers:
+  - "@russellb"
+  - "@dcbw"
+  - "@zshi-redhat"
+approvers:
```
russellb, line 9: Should list @derekwaynecarr as the approver on this one

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+  the BlueField-2 where:
+
+    1. The NIC is capable of running OpenShift.
+
+    2. The NIC has an IPMI-addressable Baseboard Management Controller
+       that can be used to control and provision it independently of
+       its host.
+
+    3. The NIC implements SR-IOV in such a way that we can connect
+       pods on the host directly to an OVN bridge on the NIC.
```
erig0, line 67: Worth a blurb about third-party CNIs other than ovn-kubernetes.. even if that blurb is "they're on their own."

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+monitoring functionality from the CPU to the DPU in the future.
+
+[Smart NIC OVN Offload]: https://github.com/openshift/enhancements/pull/732
+
+### Non-Goals
+
+- Support for ASIC- or FPGA-based Smart NICs.
+
+- Support for ARM-based Smart NICs where we would not be able to run
+  RHCOS on the NIC (eg, because the NIC requires proprietary drivers).
```
erig0, line 90: This also includes MOFED, IMO.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+Although we will eventually want to use IPMI to control and provision
+the NICs, just like we do for the hosts with bare-metal IPI, this will
+not work with the pre-GA BlueField-2 cards that we currently have
+access to. Instead, we will have to communicate with the NICs over PCI
+from the host. The most future-compatible approach is probably to
+create an IPMI proxy that can run on each host, claiming its own
+separate IP address on the provisioning network, and then proxying
+commands to the NIC via the PCI "RShim" interface. (Thus, we will have
+to first install OCP on the node, then deploy a DaemonSet to the node
+to run the IPMI proxy, then install OCP on the NIC via the proxy.)
```
erig0, line 154: I think this is a nice idea. Keep in mind that whatever runs on the OCP host also needs access to device nodes to access the rshim console.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+IPMI directly to the NIC's Baseboard Management Controller connected
+directly via the out-of-band BMC network port on the card.
+
+(One unresolved problem with the "RShim" approach is that based on the
+proof of concept prototypes, it does not seem to be possible to do a
+fully-unsupervised installation via that interface. At one point, user
+intervention over a serial console is required to pick the correct
+option from a boot menu whose contents cannot be reliably predicted in
+advance. We may have to just have the user ssh in at some point during
+the install process to deal with this.)
```
erig0, line 166: IMO, an alternative to all this provisioning is to assume that the BF-2 will PXE boot every time it boots. Then "provisioning" is just a matter of configuring the BF-2 to PXE boot by default. This has other implication though, e.g. root filesystem in RAM or via NFS. But I would hope with CoreOS this would be relatively easy.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+We cannot commit to shipping this feature in any specific OCP release
+at this time, since it depends on when NVIDIA releases the GA version
+of the BlueField-2 card.
+
+We are currently developing using pre-GA hardware, with a plan based
+on official descriptions of how the final GA BlueField-2 hardware will
+work. In the event that the product is changed before GA, some or all
+of our plans may need to be revisited.
+
+At the time of writing, the network driver for the out-of-band BMC
```
erig0, line 188: s/BMC/management/

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+network port has been posted to LKML and gone through some review, but
+has not yet merged to the Linux kernel. We will need to have this
+driver in the (ARM) RHCOS kernel to be able to do an OCP install over
+the BMC port. While we assume that this driver will merge upstream
+before the product goes GA, it is possible that it won't.
```
erig0, line 193: IMO it's a hard requirement.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+## Test Plan
+
+...
+
+## Graduation Criteria
+
+### Dev Preview -> Tech Preview
+
+- We are able to do a fully-unsupervised install to the NIC over the
+  OOB BMC port. (That is, the temporary hacks discussed above under
```
erig0, line 208: s/BMC/management/

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+    to tolerations and node selectors on existing DaemonSets, both
+    within OCP and in third-party operators or customer workloads (and
+    it may turn out that this approach is unworkable for this reason).
+
+  - The Smart NICs could be part of a second cluster running in
+    parallel with the primary cluster, with the control plane of the
+    NIC cluster running as pods within the primary cluster.
+
+  - The Smart NICs could be part of a second cluster which is
+    completely separate from the primary cluster, with its own master
```
zshi-redhat, line 127: It is expected that there will be some kinds of control plane communication between two separate clusters to exchange necessary information for deployment and running workload. It would not be totally separated and we perhaps need to provide the management of such info exchanges.

enhancements/network/ocp-on-smart-nics.md
zshi-redhat: BlueField2 supports three different modes, with each has different level of security isolation.
The default mode is not the strongest isolation mode, I assume we may start from the default mode and then allow switching of the modes as user may have different requests on the security isolation.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+  and the ARM-based Smart NICs themselves. (As discussed below, such
+  an environment may involve one or more OCP clusters.)
+
+- Define how OCP installs and upgrades will work in such an
+  environment.
+
+This enhancement does not cover what we actually _do_ with the NICs
+after they are provisioned. The primary use case of OVN offload is
+covered by "[Smart NIC OVN Offload]". It is possible that we may also
+want to allow customers to move additional network, security, and
```
zshi-redhat, line 80: and storage functionality.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+
+(One unresolved problem with the "RShim" approach is that based on the
+proof of concept prototypes, it does not seem to be possible to do a
+fully-unsupervised installation via that interface. At one point, user
+intervention over a serial console is required to pick the correct
+option from a boot menu whose contents cannot be reliably predicted in
+advance. We may have to just have the user ssh in at some point during
+the install process to deal with this.)
+
+This would explicitly be a temporary hack which would not ever be part
```
zshi-redhat, line 168: Rshim may not be suitable for running productized deployment, I'd still consider it as a userful tool to be added for debugging. wdyt? For example, either via rhcos image or included in the debug container.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+of a released version of OCP.
+
+### ...
+
+...
+
+## Risks and Mitigations
+
+### External Schedule/Feature Changes
+
```
zshi-redhat, line 178: Shall we also mention the readiness of openshift on aarch64 as a prerequisite?

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+Rather than running a full OpenShift installation on the NIC, we could
+instead provision a simple OCP system from the node, and have each
+node manage its own NIC's software and configuration directly.
+(Indeed, this is how the proof of concept for OVN offload was
+implemented.)
+
+However, we expect customers to be interested in the greater degree of
+security provided by having the NIC software be entirely isolated from
+the host software. Eventually, this can evolve into a system like
```
hardys, line 267: Is this security advantage offset by having a much larger attack surface when running the full OCP stack on the NIC?

It seems to me that this kind of workload is ideal for some kind of minimal/immutable "appliance" image based on RHCOS, deployed by the host that owns the NIC.

Also should the Alternatives contain the option of just running kubelet on the NIC, and the reasons for rejecting that approach?

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+  - The Smart NICs could be nodes within the same OCP cluster as their
+    hosts. We would need to use some combination of labels and taints
+    to distinguish "real" nodes from Smart NIC nodes. (Among other
+    things, the Smart NICs would be their own `MachineConfigPool`,
+    separate from the masters and workers.) This may require changes
+    to tolerations and node selectors on existing DaemonSets, both
+    within OCP and in third-party operators or customer workloads (and
+    it may turn out that this approach is unworkable for this reason).
+
+  - The Smart NICs could be part of a second cluster running in
```
enxebre, line 122: how/what is orchestrating upgrades and ocp version skew in the topologies with more than one cluster? Is this proposal intending to provide impl details for this?

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+(While the exact architectural constraints may change in the future,
+it is likely that some customers will always want the possibility of
+only having Smart NICs on a subset of nodes.)
+
+### Overall Cluster Architecture
+
+There are several possible overall architectures:
+
+  - The Smart NICs could be nodes within the same OCP cluster as their
+    hosts. We would need to use some combination of labels and taints
```
enxebre, line 114: How/What is meant to set this labels/taints? Is this intending to rely on MCO? If so I don't think it supports explicit labels/taints a node creation via kubelet flags other than master/worker.

enhancements/network/ocp-on-smart-nics.md
danwinship: So, I guess the big advantage of this is that you don't have to do upgrades; you just change the image served by the PXE server. Right?

You can't PXE-boot to a full running OCP install, so this would imply a non-OCP solution as discussed in some of the other comments.

Although this would save us from having to deal with upgrading the NICs, it would imply that the provisioning infrastructure would now be a production part of the cluster and would need to be managed, upgraded, etc, which I believe we do not currently do.

enhancements/network/ocp-on-smart-nics.md
danwinship: Sorry, "the product" in that sentence was referring to BlueField-2, not OCP. ie, It's possible that NVIDIA will ship the GA cards before they've finished upstreaming this driver, in which case we have to figure out what we want to do.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,276 @@
...
+"DPU" (Data Processing Unit), freeing up host CPU resources for
+customer workloads (and allowing faster data transfer by ensuring the
+network stack doesn't need to compete for CPU with the customer
+workload).
+
+However, doing this requires installing and managing an appropriate
+software stack on all of the NICs in the cluster. This software will
+need to be kept in sync with the software on the host PC (eg, as new
+functionality is added to OCP). The obvious answer to this problem is
+to run OpenShift on the NIC as well as on the host.
```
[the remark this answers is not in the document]
danwinship: > This answer is not obvious to me.

Yeah, the more I wrote the less obvious it became. :slightly_smiling_face:

This decision had already been made before I got involved, and I don't yet know everything that has been discussed, and I'm discovering now that this has apparently already been discussed more with other OpenShift teams that I hadn't even heard about. So this WIP enhancement is partly about trying to gather all this info...

> Will we have the same remote-worker-node issue that resulted in creating single-node deployments?

Not sure what that refers to

> Do we need to support these things being part of multi-node as well as single-node deployments?

I'm not sure from your questions if you're talking about the "host" cluster being multi/single-node or the NIC cluster. The expected architecture for the host cluster is that you'll have 3 masters, some non-0 number of dumb workers, and a medium to large number of smart workers with smart NICs. The NICs would either all be part of the same OCP cluster as the hosts, or else would all be part of some other OCP cluster. I don't think there's any interest in running a single-node cluster on a NIC. I'll clarify that.

> Will these NICs have the RAM and CPU capacity to run as an OpenShift node?

Yes. (BF-2 has "Up to 8 Armv8 A72 cores" and "8GB / 16GB / 32GB of on-board DDR4", and we could say we only supported a subset of configurations if we needed to.) I should make this more explicit in the "Goals"/"Non-Goals".

> Do we know for certain that we will be able to "own" the OS on the NIC? For all vendors? Will all of the drivers we need be open sourced?

For BF-2, yes. For other vendors, not necessarily. The "Goals"/"Non-Goals" section is supposed to be trying to carve out the set of NICs we might support. (eg, there's a Non-Goal of "`Support for ARM-based Smart NICs where we would not be able to run RHCOS on the NIC (eg, because the NIC requires proprietary drivers).`")

> Do we need to support running software on the NICs that is not containerized?

No more so than on any other OCP node. (eg, we'd run OVS as a systemd service, but ovn-kubernetes in pods.) In particular, the ARM OS doesn't need to run any support software for the NIC.

enhancements/network/ocp-on-smart-nics.md
danwinship: I talk a bit about that in the "Upgrade / Downgrade Strategy" and "Version Skew" sections, but this enhancement is still a work in progress

enhancements/network/ocp-on-smart-nics.md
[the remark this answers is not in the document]
dhellmann: > > This answer is not obvious to me.
> 
> Yeah, the more I wrote the less obvious it became. 🙂
> 
> This decision had already been made before I got involved, and I don't yet know everything that has been discussed, and I'm discovering now that this has apparently already been discussed more with other OpenShift teams that I hadn't even heard about. So this WIP enhancement is partly about trying to gather all this info...

+1 to collecting it all in one place

> > Will we have the same remote-worker-node issue that resulted in creating single-node deployments?
> 
> Not sure what that refers to

If a node reboots and can't connect to its control plane, it doesn't start any workloads. Since we can't ensure that all RAN nodes would always be able to see their control plane, we couldn't ensure that they would bring their workloads back online after a power loss. That resulted in [a series of enhancements](https://github.com/openshift/enhancements/tree/master/enhancements/single-node) to run the entire OpenShift stack on a single node.

> > Do we need to support these things being part of multi-node as well as single-node deployments?
> 
> I'm not sure from your questions if you're talking about the "host" cluster being multi/single-node or the NIC cluster. The expected architecture for the host cluster is that you'll have 3 masters, some non-0 number of dumb workers, and a medium to large number of smart workers with smart NICs. The NICs would either all be part of the same OCP cluster as the hosts, or else would all be part of some other OCP cluster. I don't think there's any interest in running a single-node cluster on a NIC. I'll clarify that.

It seems like these NICs could have the same connectivity issues, especially if they are used in one of the RAN servers on a cell tower. As you say, the configurations described in the enhancement assume the NIC is either a member of the same cluster its host is or the member of a different multi-node cluster. The first approach doesn't seem likely to work for the telco cases at all because it would result in a 2 node cluster, with all of the reliability issues that brings with it. The second approach would only work if the NIC is running its own single-node deployment.

If our telco customers are not looking at these NICs, then maybe that limitation doesn't matter, for now.

> > Will these NICs have the RAM and CPU capacity to run as an OpenShift node?
> 
> Yes. (BF-2 has "Up to 8 Armv8 A72 cores" and "8GB / 16GB / 32GB of on-board DDR4", and we could say we only supported a subset of configurations if we needed to.) I should make this more explicit in the "Goals"/"Non-Goals".

Nice. Can I get one to use as a development server? ;-)

enhancements/network/ocp-on-smart-nics.md
danwinship: The motivation is "it's way easier", so if there are customers interested in the increased network throughput who don't care as much about the isolation stuff, then it may be a good approach as a first version. If there aren't enough customers who would be interested in that, then it wouldn't make sense.

(In a two-cluster architecture we need to either have one ovn-kubernetes master programming two sets of OVN databases, or else two ovn-kubernetes masters, one of which talks to two different apiservers. Either way, there's going to be a bunch of rearchitecting.)

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,313 @@
...
+    control planes for both clusters are running in a different
+    cluster, ROKS-style.
+
+The latter two approaches allow for more isolation between the hosts
+and the NICs, since potentially even the cluster administrators of the
+primary cluster might not be able to influence the NIC cluster. This
+would be especially important if additional primary-cluster security
+and monitoring functions were offloaded to the NICs in the future.
+
+### Initial Development With Beta BlueField-2 Hardware
```
Billy99, line 175: My understanding of the BF-2 cards is that will be GA Hardware with a BMC network port and GA Hardware without a BMC network port (this was from a NVIDIA engineer working on the tiger team). Someone can correct me if I'm wrong about this. So if this is true, we as Red Hat need to decide if will support both versions of hardware. Or support the hardware with BMC network port upfront and add support for non-BMC hardware later.

enhancements/network/ocp-on-smart-nics.md
erig0: > So, I guess the big advantage of this is that you don't have to do upgrades; you just change the image served by the PXE server. Right?

Correct. upgrade == reboot

> You can't PXE-boot to a full running OCP install, so this would imply a non-OCP solution as discussed in some of the other comments.

I was unaware of that. Why can't it boot a full OCP install?

enhancements/network/ocp-on-smart-nics.md
dcbw: @hardys In a number of these scenarios the host is *untrusted*. So it can't be deployed by the host because then you lose the isolation benefits of walling off the actual node from a network perspective.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,313 @@
...
+
+  4. All of the above can be done without running any proprietary
+     drivers or third-party software on either the host's or the NIC's
+     RHCOS installation.
+
+## Design Details
+
+### Clusters Will Contain a Mix of Smart and "Dumb" NICs
+
+OVN Kubernetes is not currently capable of mixing offloaded and
```
dcbw, line 128: @danwinship I think we should try to change this. It's not clear to me (though I haven't read deeply) why a pod would care that it was offloaded or not. So sure, pods could request being scheduled to an offloaded node, but by default I'd presume they'd land somewhere and if that happened to be a smartnic node, so be it.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,313 @@
...
+non-offloaded networking on the same node, so every pod on a node
+containing a Smart NIC must have offloaded networking, which, in the
+current architecture, means it must explicitly _request_ offloaded
```
dcbw, line 131: @danwinship I think we should try to change this. It's not clear to me (though I haven't read deeply) why a pod would care that it was offloaded or not. So sure, pods could request being scheduled to an offloaded node, but by default I'd presume they'd land somewhere and if that happened to be a smartnic node, so be it.

enhancements/network/ocp-on-smart-nics.md
dcbw: @derekwaynecarr all our stuff is containerized already; the same stuff that runs on the host today (OVS, ovn-controller, ovnkube, network metrics, etc) moves to the NIC.  It makes little sense to write a container lifecycle manager since we already have one: kubelet. 

For the security angle, it may not be worth having a completely separate control plane just for SmartNIC control, unless you're already going to have that for other reasons. The point is that the hosts are untrusted and since all networking through the NIC, workloads that somehow compromise the host can be detected, monitored, and isolated (I guess?). The NIC wouldn't run regular workloads, just control-plane stuff.

enhancements/network/ocp-on-smart-nics.md
dcbw: I presume the taints would be added by MCO, CNO, SRIOV operator, etc that has visibility into the hardware of each node. It could also be added by ovn-kubernetes running on the NIC since it clearly knows the IP of its host.

I'm not sure the taint needs to happen at node creation time necessarily, just before untrusted (non control plane?) workloads can be scheduled on the machine.

enhancements/network/ocp-on-smart-nics.md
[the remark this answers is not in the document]
romfreiman: > The first approach doesn't seem likely to work for the telco cases at all because it would result in a 2 node cluster,

Maybe, in case of a SNO,  we can have it as a worker (host - master, card - worker). It's not a real SNO of course, but a master + worker topology (that is doesnt yet exist of course).

enhancements/network/ocp-on-smart-nics.md
danwinship: I talked about this a little with @zshi-redhat ; it seems that the problem is that we need to use the sr-iov device plugin to allocate a VF, but kubelet will only ask it to do that if the pod has the right annotations, and we can't just annotate it via an admission plugin, because the admission plugin runs before the pod gets scheduled. But I think he agreed it would be possible to change things to have the offloaded-ovn-kube cni plugin do the VF allocation itself

zshi-redhat: Yes, pod can request being scheduled to an offloaded node by adding sriov extended resource request in pod spec, on the other hand, pod that doesn't request extended resource can also be scheduled to the offloaded node unless we explicitly set affinity or labels or taints. 
The problem now is there are several cluster pods or infra pods such as pods in openshift-monitoring, openshift-ingress namespaces etc which don't have affinity or labels or taints that allow them to not be scheduled on the offloaded node. So if the scheduling of these pods to offloaded node happens, then ovn-k8s is not capable of creating network for these pods which results in failure of pod creation.
There are several ways to solve this:
1) run these infra pods with hostnetwork:true  which doesn't need to run ovn-k8s CNI
2) add extended sriov resource request to these pods that *will* be scheduled on the offloaded node (which may not be able to know upfront before k8s-scheduler runs)
3) set anti-affinity to the daemonsets/deployments so that their pods won't be scheduled on offloaded node.
All above three would require changing the manifests in bindata for various openshift cluster operators.

enhancements/network/ocp-on-smart-nics.md
zshi-redhat: There are some details on [where](https://github.com/openshift/enhancements/pull/732/files#diff-9cb730b0bcd44036b71650e25bd939ec20d549835f76ff5b1e964c6d20d6e574R142-R144) the label is defined and [who](https://github.com/openshift/enhancements/pull/732/files#diff-9cb730b0bcd44036b71650e25bd939ec20d549835f76ff5b1e964c6d20d6e574R203-R207) applies the label in [this](https://github.com/openshift/enhancements/pull/732) enhancement.
In short, the label is applied to nodes in the same MCP by SR-IOV Operator and the offloading of ovn components to SmartNIC is triggered while the label is applied. The current idea is to first install the node as a regular openshift node, then customize it to a SmartNIC node.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,313 @@
...
+
+We want to support Smart NICs that are vaguely "BlueField-2-ish". eg:
+
+  1. The NIC has a sufficiently-powerful ARM processor (or,
+     theoretically, a processor of some other architecture that RHCOS
+     supports), and sufficient RAM, SSD, etc, to run as an OCP node.
+
+       - Current BlueField-2 specs mention "Up to 8 Armv8 A72 cores"
+         and "8GB / 16GB / 32GB of on-board DDR4". It is possible we
+         would only support a subset of SKUs.
```
s1061123, line 107: When I look the BF2 spec, BF2 seems not to have storage (e.g. ssd/hdd) hence BF2 should run kernel with ramdisk. This means BF2's RAM will be consumed not only by process working memory also by ramdisk for RHCOS/container images/logging data from container).

Currently OSP does not have strict storage management as far as I know. Is there any idea for that?

danwinship: It's got 32G of eMMC (ie, flash-card-on-a-chip), which is supposed to be divided between two installs (so, eg, you're running from partition 0 and install an upgrade into partition 1, then reboot into partition 1, and if the watchdog timer goes off before the card comes back up, it will automatically reboot back into partition 0 for you).

The [OCP requirements for bare metal](https://docs.openshift.com/container-platform/4.6/installing/installing_bare_metal/installing-bare-metal.html#minimum-resource-requirements_installing-bare-metal) say workers need at least 8G of storage, so we should be OK.

(The cards also have M.2 NVMe ports so you can attach SSDs to them.)

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,313 @@
...
+It may turn out that it doesn't make sense for us to worry about these
+cards.]
+
+Although we will eventually want to use IPMI to control and provision
+the NICs, just like we do for the hosts with bare-metal IPI, this will
+not work with the pre-GA BlueField-2 cards that we currently have
+access to. Instead, we will have to communicate with the NICs over PCI
+from the host. The most future-compatible approach is probably to
+create an IPMI proxy that can run on each host, claiming its own
+separate IP address on the provisioning network, and then proxying
```
pliurh, line 187: Is this `provisioning network` dedicated to smart NIC provisioning?

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,313 @@
...
+commands to the NIC via the PCI "RShim" interface. (Thus, we will have
+to first install OCP on the node, then deploy a (highly-privileged)
+DaemonSet to the node to run the IPMI proxy, then install OCP on the
+NIC via the proxy.)
+
+Later on, it should be possible to remove the proxy and instead speak
+IPMI directly to the NIC's Baseboard Management Controller connected
+directly via the out-of-band management network port on the card.
```
pliurh, line 195: According to https://docs.mellanox.com/display/BlueField2DPUENUG/NVIDIA+BlueField-2+Ethernet+DPU+User+Guide, BF-2 cards only have 2 regular ports plus 1 OOB port.

If we use the OOB port for provisioning, then we have to use one regular port for the smart nic cluster network. It means only 1 port can be used by the host cluster, thus users cannot use bond anymore.

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,313 @@
...
+    nodes. (In some external discussion of this approach, it was
+    assumed that the masters in this scenario would have to be on ARM
+    hosts, but this isn't required; you can have a mix of x86_64
+    master nodes and ARM Smart NIC "worker" nodes.)
+
+  - The Smart NICs could be part of a second cluster, where the
+    control planes for both clusters are running in a different
+    cluster, ROKS-style.
+
+The latter two approaches allow for more isolation between the hosts
```
pliurh, line 169: For the 2 clusters approach, do we also want to use ovn-kube as the CNI of the smart NIC cluster? Is that possible to have OVS managed by 2 OVN instances?

enhancements/network/ocp-on-smart-nics.md
```diff
@@ -0,0 +1,313 @@
...
+Although we will eventually want to use IPMI to control and provision
+the NICs, just like we do for the hosts with bare-metal IPI, this will
+not work with the pre-GA BlueField-2 cards that we currently have
+access to. Instead, we will have to communicate with the NICs over PCI
+from the host. The most future-compatible approach is probably to
+create an IPMI proxy that can run on each host, claiming its own
+separate IP address on the provisioning network, and then proxying
+commands to the NIC via the PCI "RShim" interface. (Thus, we will have
+to first install OCP on the node, then deploy a (highly-privileged)
+DaemonSet to the node to run the IPMI proxy, then install OCP on the
```
pliurh, line 190: When provisioning the smart NIC the host will lose cluster network connection. It will be hard to debug if the IPMI proxy pod doesn't work as expected.

enhancements/network/ocp-on-smart-nics.md
hardys: @dcbw ack - is that dependent on embedded vs separated host mode on the NIC?  Or just that there isn't any USB connectivity to the host to enable reprogramming via rshim?  May be worth calling out explicitly to clarify that in this scenario the host will not have full control over the connected hardware.

enhancements/network/ocp-on-smart-nics.md
danwinship: No, there's already a provisioning network.

For non-BlueField bare-metal installs, the expectation is that your host has a spiffy PCI NIC that does 10-100G ethernet and offload and stuff, which is what's used for real traffic, but there's also another wimpier ethernet port soldered to the motherboard that lets you do IPMI, etc, to manage the host. So we install the node via that management port, and then transition to using the "real" NIC for everything else after that. The assumption is that for security reasons they'd be connected to different ethernet networks, though that isn't actually required.

For real (future) BlueField installs, the expectation is the same, except that in addition to the management port on the host motherboard, there's _another_ management port on the BF-2, so you'll have three ethernet cables: one for host management/provisioning, one for NIC management/provisioning, and one for real cluster traffic. (Or maybe two for traffic; there are actually 2 100G ports on the BF-2 in addition to the 1G management port.) I imagine the host management port and the NIC management port would be connected to the same network as each other (which is still separate from the real network), because otherwise you'd need two sets of PXE servers and stuff...

For the "IPMI proxy" scenario, we can't use the management port on the BF-2, so instead we have the host claim a second IP address on its own management port's network, and pretend that that's the BF-2 management port.

enhancements/network/ocp-on-smart-nics.md
romfreiman: @zshi-redhat can you please elaborate what's the difference between openshift-monitoring, ingress and others? What they expected to have? Can you please provide an example. If it's a minor change, I guess we can ask relevant teams to push it forward.

enhancements/network/ocp-on-smart-nics.md
romfreiman: To customize how? Automatically by the operator based on hw discovery or we expect the user to add such annotaions?

zshi-redhat: @romfreiman It is expected that user or cluster admin will first create a custom MCP and add SmartNIC x86 nodes to the pool,  then run the following steps to customize the nodes in the custom MCP:
-  (manual) Specify node labels in SR-IOV Operator API to match on the custom MCP
-  (auto) SR-IOV Operator will apply pre-defined label to the nodes in custom MCP (The label is used by Cluster Network Operator to deploy/remove ovn-kube daemonsets to/from the nodes in custom MCP)
-  (auto) SR-IOV Operator will apply pre-defined MachineConfigs to the nodes in custom MCP via MCO API
-  (auto) Nodes in custom MCP reboot
-  (auto) SR-IOV Operator configures SmartNIC device and expose SR-IOV devices as k8s extended resource
-  (auto) Nodes become ready and serving SR-IOV pod creation

enhancements/network/ocp-on-smart-nics.md
zshi-redhat: @romfreiman With current OVN-Kubernetes design, it is only supported to run SR-IOV pods on the SmartNIC x86 node. Regular pods whose networks require virtual **veth** interfaces cannot be plugged to OVN network because there will be no Open vSwitch running on the SmartNIC x86 node. Some pods in openshift-monitoring, openshift-ingress or other namespaces are the pods whose networks require virtual **veth** interfaces. So this is a limitation that current SmartNIC x86 nodes cannot host regular **veth** pods.

The example of setting pod to use host network is as below:
```
apiVersion: v1
kind: Pod
metadata:
  name: testpod
spec:
  hostNetwork: true       <======
  containers:
  - name: container0
    image: centos
    command: [ "/bin/bash", "-c", "--" ]
    args: [ "while true; do sleep 300000; done;" ]
```
It needs change in pod or daemonset or deployment manifests.
This problem is also described [here](https://github.com/openshift/enhancements/pull/732/files#diff-9cb730b0bcd44036b71650e25bd939ec20d549835f76ff5b1e964c6d20d6e574R277-R340).

enhancements/network/ocp-on-smart-nics.md
danwinship: Because there's no image corresponding to "fully-configured OCP node" that it could boot.

The on-disk state depends on the local cluster configuration, so we can't just ship a generic one-size-fits-all disk image as part of the OCP release. And because of the RHCOS immutable-filesystem-image thing, you can't boot an almost-right image and make a few small fixups to it; you have to boot an almost-right image, write out an updated ostree commit, and reboot into the updated immutable filesystem.

danwinship: (I guess in theory some operator could create a bootable fully-configured Smart NIC worker image, but there's nothing that does that currently.)

eranco74: Full OCP install requires a bootstrap node and an orchestrator (the OpenShift installer). 
You can PXE-bot a single node OpenShift.

enhancements/network/ocp-on-smart-nics.md
s1061123: At least 4.7, worker node requires 120G...
4.6 doc (you mentioined) seems broken (vCPU 2 goes RAM field).

https://docs.openshift.com/container-platform/4.7/installing/installing_bare_metal/installing-bare-metal.html#minimum-resource-requirements_installing-bare-metal

danwinship: Ah... well, immediately post-install, a bare-metal worker is using 8G of its disk (masters use a bit more), so the "120G required" seems to be more about assuming how much space the customer is _eventually_ going to need for their own workloads.

Still, we aren't going to have _lots_ of extra space (especially if we stick with two partitions).

s1061123: I do not only mention that initial required space also the space the process/container uses (at least, to keep the container image, cri-o consumes the space and many containers/processes uses the space for logging). 

I believe you also consider these stuff and conclude you can. I have concern so I just share it with you.

Note: please not add me into reviewer because I want not to be the one.

danwinship:
@zshi-redhat @pliurh I've repushed with the enhancement split up into two docs, one with specific details of what we plan to do in Dev Preview, and then another that's just a broad overview (and glossary) that's a parent to the Dev Preview enhancement, and to #732, and #890. Can you re-review? (And did I get the glossary terms right?)

enhancements/network/dpu/ocp-on-dpu.md
```diff
@@ -0,0 +1,260 @@
...
+  cluster (so that the BF-2s themselves will be powered on).
+
+- Install the infra cluster.
+
+- Install the [DPU Operator] onto the infra cluster, providing it with
+  an appropriate tenant-cluster kubeconfig. Initially, this will just
+  ensure that each of the cards is configured in "passthrough" mode,
+  allowing the tenant hosts unrestricted network access.
+
+- Add the BF-2 workers to the tenant cluster. The combination of the
```
zshi-redhat, line 123: SR-IOV Operator is missing here.
The combination of the CNO and SR-IOV Operator config in the tenant cluster and the DPU Operator in the infra cluster (SR-IOV Operator initializes the VF0 used as ovn-k8s-mp0 port).

enhancements/network/dpu/ocp-on-dpu.md
```diff
@@ -0,0 +1,260 @@
...
+As with the tenant side, the changes to ovn-kubernetes itself to
+support "DPU NIC" mode have already been made upstream.
+
+Infra cluster masters and non-BF-2 worker nodes will run
+ovn-kubernetes (in "normal" mode) as their CNI plugin. However, BF-2
+NIC nodes will not be able to do this, since it is not currently
+possible to run ovn-kubernetes in both "normal" mode and "DPU NIC"
+mode on the same host.
+
+The current plan for Dev Preview is that BF-2 worker nodes will not
```
zshi-redhat, line 188: The term `BF-2 worker(s)` is used for both tenant workers and infra workers in this proposal. We shall change the naming for either infra or tenant cluster.

enhancements/network/dpu/ocp-on-dpu.md
```diff
@@ -0,0 +1,260 @@
...
+run any CNI plugin, and will only support host-network pods.
+(Actually, they need _some_ CNI in order for the node to become Ready
```
zshi-redhat, line 190: _some_ CNI, not just for node to become Ready, but also for cluster pods (such as MCD, DNS) to function properly.  It is worth mentioning the use of kube-proxy in infra cluster as well.

enhancements/network/dpu/ocp-on-dpu.md
```diff
@@ -0,0 +1,260 @@
...
+  ensure that each of the cards is configured in "passthrough" mode,
+  allowing the tenant hosts unrestricted network access.
+
+- Add the BF-2 workers to the tenant cluster. The combination of the
+  CNO config in the tenant cluster and the DPU Operator in the infra
+  cluster will cause each BF-2 worker to be configured to run
+  ovn-kubernetes in "DPU Host" mode with its corresponding NIC running
+  ovn-kubernetes in "DPU NIC" mode.
+
+[DPU Operator]: https://github.com/openshift/enhancements/pull/890
```
zshi-redhat, line 129: We have the link to DPU Operator in the below `DPU Opeator` section. Maybe remove this or simply refers to the latter.

enhancements/network/dpu/ocp-on-dpu.md
```diff
@@ -0,0 +1,260 @@
...
+- Install the [DPU Operator] onto the infra cluster, providing it with
+  an appropriate tenant-cluster kubeconfig. Initially, this will just
+  ensure that each of the cards is configured in "passthrough" mode,
+  allowing the tenant hosts unrestricted network access.
+
+- Add the BF-2 workers to the tenant cluster. The combination of the
+  CNO config in the tenant cluster and the DPU Operator in the infra
+  cluster will cause each BF-2 worker to be configured to run
+  ovn-kubernetes in "DPU Host" mode with its corresponding NIC running
+  ovn-kubernetes in "DPU NIC" mode.
```
zshi-redhat, line 127: We use "DPU" mode instead of "DPU NIC" in ovn-kubernetes.

enhancements/network/dpu/overview.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Goals
+
+- Design a system for provisioning and maintaining a RHEL-based
+  software stack on Smart NICs to support features such as "[Smart NIC
+  OVN Offload]".
+
+  - Although this enhancement is currently written to assume that
+    this means running OCP on the NICs, this is not yet cast in
+    stone. Some other approaches are discussed under "Alternatives"
```
zshi-redhat, line 129: I don't see other approaches discussed in `Alternatives`.

enhancements/network/dpu/overview.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Risks and Mitigations
+
+...
+
+## Design Details
+
+### Cluster Architecture
+
+The expected cluster architecture at this time is:
```
zshi-redhat, line 203: expected `tenant` cluster architecuture.

enhancements/network/dpu/overview.md
```diff
@@ -0,0 +1,239 @@
...
+### Goals
+
+- Design a system for provisioning and maintaining a RHEL-based
+  software stack on Smart NICs to support features such as "[Smart NIC
+  OVN Offload]".
+
+  - Although this enhancement is currently written to assume that
+    this means running OCP on the NICs, this is not yet cast in
+    stone. Some other approaches are discussed under "Alternatives"
+    and one of those could potentially replace the OCP approach.
```
zshi-redhat, line 130: Do we really mean `replace` or they will co-exist in different setups? This sounds like a different path forward, would like to hear what is the other approach.

enhancements/network/dpu/ocp-on-dpu.md
```diff
@@ -0,0 +1,260 @@
...
+  node.
+
+- Create a MachineConfigPool in the tenant cluster for the BF-2
+  workers, and update the Cluster Network Operator configuration to
+  indicate that this pool is for BF-2 workers.
+
+- Power on, but don't yet install, the BF-2 hosts in the tenant
+  cluster (so that the BF-2s themselves will be powered on).
+
+- Install the infra cluster.
```
pliurh, line 116: Currently, both clusters share the same VLAN for the API network. We disconnected the DHCP server in each cluster manually during the cluster installation. And we do some manual IP assignments. I think we need to consider how to resolve this in the dev-preview.

enhancements/network/dpu/ocp-on-dpu.md
danwinship: Hm... how about "DPU NIC workers" and "DPU Host workers" ?

enhancements/network/dpu/ocp-on-dpu.md
danwinship: machine-config-daemon is hostNetwork. I thought we had decided we'd just not run CoreDNS on either the DPU Host workers or the DPU NIC workers for Dev Preview?

enhancements/network/dpu/overview.md
danwinship: I don't think we would want to support multiple very-different architectures. We're going to want to settle on one architecture that's flexible enough to handle all of our use cases.

enhancements/network/dpu/overview.md
```diff
@@ -0,0 +1,239 @@
...
+
+The details will be figured out in further enhancements.
+
+### Risks and Mitigations
+
+...
+
+## Design Details
+
+### Cluster Architecture
```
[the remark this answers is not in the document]
danwinship: This isn't really talking about "two-cluster" design; it's about the computing environment as a whole, regardless of how we implement DPU support in it.

eg, in particular, we are not currently considering Single Node OpenShift as a use case.

enhancements/network/dpu/ocp-on-dpu.md
danwinship: "disconnected the DHCP server" seems more like a problem with our particular network setup than a problem with the feature. (Customers will normally not try to install OCP on a network that has a DHCP server that they don't want to use.)

But I agree about VLANs, etc

enhancements/network/dpu/ocp-on-dpu.md
zshi-redhat: Right about DNS.  ok I thought when we say CNI, it would cover the k8s service networking like accessing from hostnetwork pod to api server, thus the kube-proxy, but if the CNI here is just about pod networking, it shall be fine to mention kube-proxy.

enhancements/network/dpu/ocp-on-dpu.md
zshi-redhat: That sounds good!
I think these could be descriptive names, we use DPU and DPU Host in the ovn-k8s code.

enhancements/network/dpu/ocp-on-dpu.md
Billy99: Currently, the infra and tenant clusters are in two different subnets, but using the same interface (hence the temporary disconnect of DHCP server). Discussion last week on whether it should be one subnet or not. Not sure if we came to a conclusion.

enhancements/network/dpu/overview.md
```diff
@@ -0,0 +1,269 @@
...
+from the host CPU to the NIC, to increase throughput and to free up
+host CPU cycles for workloads rather than network processing.
+
+In particular:
+
+- We can move all of the node-side OVS, OVN, and OVN Kubernetes
+  daemons from the host CPU to the NIC CPU. (The host side would
+  retain only a CNI plugin binary.)
+
+- We can move IPsec encryption from the host CPU to the NIC CPU
```
bn222, line 108: Since NIC CPU is not defined, I assume NIC CPU refers to the CPU on the NIC. It's important to note that it is much slower than the host CPU. At the same time, this section mentions an increase in throughput as being one of the goals. For that reason, it makes sense to explicitly add that the real goal of moving IPsec encryption to the NIC CPU is to, in turn, offloaded it to the NIC itself.

PR #739, opened 2021-04-19
management-workload-partitioning: improve namespacing in annotation names

dhellmann:
Modify the annotation names used in the management-workload-partitioning
enhancement to use better namespacing, based on comments on
https://github.com/openshift/kubernetes/pull/632

/cc @mrunalp @browsell @deads2k

PR #740, opened 2021-04-19
workload-partitioning: change crio annotation name

dhellmann:
Change the annotation name used by crio to find the configuration of
each container so that the name matches the scheme used for other
annotations, instead of having the parts of the name appear in the
reverse order.

Based on comments from https://github.com/openshift/kubernetes/pull/632
and changes to this design doc in #739

/cc @mrunalp @browsell @haircommander

### DIFF
--- a/enhancements/management-workload-partitioning.md
+++ b/enhancements/management-workload-partitioning.md
@@ -215,7 +215,7 @@ does receive the content of annotations on the pod. Therefore, we will
 copy the original CPU requests for containers in management workload
 pods into a annotations that CRI-O can use to configure the CPU shares
 when running the containers. We will use the naming convention
-`io.openshift.workload.{workload-type}.cpushares/{container-name}` and
+`cpushares.{workload-type}.workload.openshift.io/{container-name}` and
 create an annotation for each container in the pod. See below for
 details about this choice of name.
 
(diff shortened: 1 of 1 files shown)


dhellmann:
We only want this or #741

enhancements/management-workload-partitioning.md
```diff
@@ -215,7 +215,7 @@ does receive the content of annotations on the pod. Therefore, we will
 copy the original CPU requests for containers in management workload
 pods into a annotations that CRI-O can use to configure the CPU shares
 when running the containers. We will use the naming convention
-`io.openshift.workload.{workload-type}.cpushares/{container-name}` and
+`cpushares.{workload-type}.workload.openshift.io/{container-name}` and
```
deads2k, line 218: {workload-type}.cpushares.workload.openshift.io/{container-name}` probably makes more sense so the workload type won't think it claims all of foo.workload.openshfit.io

dhellmann: We didn't want CRI-O to have to know about the annotation scheme, since the scheme is specific to OpenShift and CRI-O is not. So, we wanted the "resource type" (cpushares, whatever) to be tacked onto the front or end of the annotation. That originally led to the name being reversed.

I think I like #741 better, since it makes the resource type handling more transparent, but I want to hear what @haircommander thinks before we choose.

dhellmann:
We have consensus around #741 so I am closing this PR.

PR #741, opened 2021-04-19
workload-partitioning: update cri-o annotation handling

dhellmann:
Update the annotation name, configuration examples, and the way
resource values are passed to CRI-O.

/cc @mrunalp @haircommander @browsell

### DIFF
--- a/enhancements/management-workload-partitioning.md
+++ b/enhancements/management-workload-partitioning.md
@@ -423,12 +423,6 @@ annotations:
   resources.workload.openshift.io/{container-name}: {"cpushares": 400}
 ```
 
-The annotation name used to set the resource requests is reversed
-(when compared to the other workload annotation or the resource
-request name) so that CRI-O can be configured with a prefix value to
-simplify the way it processes the strings without having to understand
-their content.
-
 The new request value and annotation value are scaled up by 1000 from
 the original CPU request input because opaque resources do not support
 units or fractional values.
--- a/enhancements/management-workload-partitioning.md
+++ b/enhancements/management-workload-partitioning.md
@@ -353,14 +353,14 @@ scheduled to run on the management CPU pool.
 7. The kubelet reads static pod definitions. It replaces the `cpu`
    requests with `management.workload.openshift.io/cores` requests of
    the same value and adds the
-   `resources.workload.openshift.io/{container-name}: {"cpushares": "400"}`
+   `resources.workload.openshift.io/{container-name}: {"cpushares": 400}`
    annotations for CRI-O with the same values.
 8. Something schedules a regular pod with the
    `target.workload.openshift.io/management` annotation in a namespace with
    the `workload.openshift.io/allowed: management` annotation.
 9. The admission hook modifies the pod, replacing the CPU requests
    with `management.workload.openshift.io/cores` requests and adding
-   the `resources.workload.openshift.io/{container-name}: {"cpushares": "400"}`
+   the `resources.workload.openshift.io/{container-name}: {"cpushares": 400}`
    annotations for CRI-O.
 10. The scheduler sees the new pod and finds available
     `management.workload.openshift.io/cores` resources on the node. The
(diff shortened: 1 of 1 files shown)


dhellmann:
We only want this or #740

deads2k:
> We only want this or #740

I prefer 741 if it works

mrunalp:
I prefer this over #740

enhancements/management-workload-partitioning.md
```diff
@@ -353,14 +353,14 @@ scheduled to run on the management CPU pool.
 7. The kubelet reads static pod definitions. It replaces the `cpu`
    requests with `management.workload.openshift.io/cores` requests of
    the same value and adds the
-   `io.openshift.workload.management.cpushares/{container-name}`
+   `resources.workload.openshift.io/{container-name}: {"cpushares": N}`
```
cynepco3hahue, line 356: I am unsure regarding `N`, should it be the integer or the string `"N"`?

cynepco3hahue: I think it better to make it string, in case if we will support additional parameters in the future(for example cpuset that will have value 0-3 and can be the integer)

dhellmann: Sorry, that should be an integer value, I just didn't have a specific number to put there. I'll expand the example to include a number.

cynepco3hahue: @dhellmann  I saw the example below, so it should be ok.
Do you know if will we always have only CPU shares or does it possible we will have an additional resource in the future(CPUSet for example). Something like `{"cpushares": 256, "cpuset": "0-3"}`.

mrunalp:
Talked with @haircommander and @rphillips and we agreed on this one.

haircommander:
I'll give a tenative LGTM here. The approach and strategy makes sense, but I'd like to code it out before giving a final :+1:

rphillips:
It does drop management, which is the name of the workload... are we removing that functionality?

mrunalp:
@rphillips The workload is still specified in the toml section and the selector. This is only for the resources specified.

cynepco3hahue:
> I'll give a tenative LGTM here. The approach and strategy makes sense, but I'd like to code it out before giving a final

@haircommander Please let me know, what format do you prefer to have under `resources.workload.openshift.io/{container-name}: {"cpushares": N}`, if N should be string or int(important during unmarshall)

haircommander:
it should be a string!

https://github.com/cri-o/cri-o/pull/4787

enhancements/management-workload-partitioning.md
dhellmann: I updated these examples to use real numbers for clarity.

enhancements/management-workload-partitioning.md
```diff
@@ -464,24 +458,27 @@ arbitrary pods to set their own `cpuset` but other CRI-O users may
...
 ```
 
 In the management workload case, we will configure it with values like
 
 ```ini
 [crio.runtime.workloads.management]
   activation_annotation = "target.workload.openshift.io/management"
-  annotation_prefix = "io.openshift.workload.management"
+  annotation_prefix = "resources.workload.openshift.io"
   resources = { "cpushares" = "", "cpuset" = "0-1" }
```
cynepco3hahue, line 477: `"cpushares": "" -> "cpushares": 0`

enhancements/management-workload-partitioning.md
```diff
@@ -591,7 +588,7 @@ deployment):
 ```ini
 [crio.runtime.workloads.management]
 activation_annotation = "target.workload.openshift.io/management"
-annotation_prefix = "io.openshift.workload.management"
+annotation_prefix = "resources.workload.openshift.io"
 resources = { "cpushares" = "", "cpuset" = "0-1,52-53" }
```
cynepco3hahue, line 592: `"cpushares": "" -> "cpushares": 0`

PR #742, opened 2021-04-19
Update projected resource CSI driver proposal

adambkaplan:
- Install via OCP payload and cluster storage operator
- Add delivery via OLM on a developer preview basis
- Condense and clarify existing proposal in light of current
  CSI driver implementation.

### DIFF
--- a/enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
+++ b/enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
@@ -69,7 +69,7 @@ There are numerous use cases where information should be defined only once on th
 ### Goals
 
 - Provide easy to use sharing of `Secrets` and `ConfigMaps` defined in one namespace by pods in other namespaces.
-- When a shared `Secret` or `ConfigMap` changes, the container filesystem content the CSI driver injects into the Pod needs to reflect those changes.
+- [reach] When a shared `Secret` or `ConfigMap` changes, the container filesystem content the CSI driver injects into the Pod needs to reflect those changes.
 - When a cluster-admin revokes access to shared `Secrets` or `ConfigMaps` that was previously granted, that data should no longer be accessible.
 
 ### Non-Goals
@@ -98,18 +98,18 @@ the credentials, I want to stop updating content within the existing pods consum
 
 #### Install
 
-For the developer preview release, the CSI driver will be installable via OLM.
-This release channel is meant to be a preview only option and allow the CSI driver to be installed on OpenShift outside of the normal release cycle.
+For the developer preview release, the CSI driver will be installable via YAML manifests available on GitHub.
+This is meant to be a preview only option and allow the CSI driver to be installed on OpenShift outside of the normal release cycle.
 
 When the CSI driver is ready for general availability, it will be installed on OpenShift via the cluster storage operator and an associated delivery operator - see the [CSI driver install proposal](/enhancements/storage/csi-driver-install.md).
-The delivery operator and OLM operator can be one and the same, with appropriate modifications depending on the context.
 
 #### CSI starting points
 
 The projected resource CSI driver will provision ephemeral volumes and take advantage of upstream's ephemeral CSI storage mechanisms.
-Ephemeral storage allows the CSI driver to use tmpfs for storage while providing appropriate protection via SELinux labels.
-Each pod will have its own copy of the shared resource that the CSI driver mounts in, and there will be a 1:1 correspondence between the pod and the SELinux label applied to the mounted files.
+Ephemeral storage allows the CSI driver to create tmpfs filessytems for each pod, where the Secret/ConfigMap content will be copied.
+The tmpfs filesystems will automatically have appropriate SELinux labeling to ensure content does not leak between pods - the driver does not need to provision or manage these labels.
 The process will be similar to what exists in core Kubernetes when Secrets and ConfigMaps are mounted into a pod.
+Acutal implementation can borrow a significant amount of logic from the [secrets-store-csi-driver](https://github.com/kubernetes-sigs/secrets-store-csi-driver), which is used to mount secrets from external "sealed" secret providers like [Vault](https://learn.hashicorp.com/tutorials/vault/kubernetes-secret-store-driver).
 
 #### End to end flow (from creation, to opt-in, to validation, and then injections)
 
@@ -202,16 +202,18 @@ The CSI Driver, aside from all the CSI and filesystem bits (see below for detail
 
 With all this necessary metadata, the CSI driver then:
 
-- reads of the `Secret` or `ConfigMap` noted in the `ProjectedResource`, storing it on disk on initial reference to said `Secret` or `ConfigMap`.
-- creation of a tmpfs filesystem for each requesting `ServiceAcount` that contains the `Secret` / `ConfigMap` data.
-- creation of k8s watches `ProjectedResources`, and for `Secret` / `ConfigMaps` in the namespaces listed in each discovered `ProjectedResources`.
-- final returning of a volume that is based off of the `ServiceAccount` specific tmpfs for the data noted in the `volumeAttributes`.
+- Performs a subject access review (SAR) check to verify the pod's service account has permission to GET the referenced Share resource.
+- If the SAR check succeeds, reads of the `Secret` or `ConfigMap` noted in the `ProjectedResource`, storing it on disk on initial reference to said `Secret` or `ConfigMap`.
+- Creates a tmpfs filesystem for each requesting `ServiceAcount` that contains the `Secret` / `ConfigMap` data.
+- Creates k8s watches `ProjectedResources`, and for `Secret` / `ConfigMaps` in the namespaces listed in each discovered `ProjectedResources`.
+- Returns a volume that is based off of the `ServiceAccount` specific tmpfs for the data noted in the `volumeAttributes`.
 
 After initial set up of things, steady state CSI driver activities include:
 
-- as updates of the watched `Secrets` and `ConfigMaps` come in, SAR checks will be done for each of the previous requesting `ServiceAccounts` being tracked against the updated `Secret` and `ConfigMap`.
-- for any `ServiceAccounts` that no longer have access, their tmpfs filesystems are unmounted, effectively removing their access to the data (though in prototyping and implementation, we'll need focused testing that this scales and mount propagation does not become a gating factor).
-- the SAR checks and if need be removal of access to data can facilitate cluster admin scenarios noted in the `User Stories` section like credential rotation.
+- As updates of the watched `Secrets` and `ConfigMaps` come in, SAR checks will be done for each of the previous requesting `ServiceAccounts` being tracked against the updated `Secret` and `ConfigMap`.
+  SAR checks are also done at regular relist intervals, which will pick up permission changes to associated cluster roles/role bindings.
+- For any `ServiceAccounts` that no longer have access, contents in their tmpfs filesystems are removed. Unmounting any volume in a running pod is not feasible (can result in `EBUSY` errors).
+- The SAR checks and if need be removal of access to data can facilitate cluster admin scenarios noted in the `User Stories` section like credential rotation.
 - But for `ServiceAccounts` that pass the SAR check, the tmpfs file with the data for the `Secret` / `ConfigMap` is updated
 
 #### An overview of CSI and the main K8S hook points
@@ -247,43 +249,55 @@ Research details and notes:
 - And then the [CSI plugin to the Kubelet](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#ephemeral-inline-volume-proposal), performs the necessary steps to engage the CSI driver.
 - the [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md) details the CSI architecture, concepts, and protocol.
 
-#### Current thoughts on implementation, how to inject/consume content on every host/node in an OpenShift cluster
+#### Ephemeral CSI driver implementation
 
-Implementation time approach:
-
-- Two example CSI plugins to use as a starting point are the [secrets
- store csi
- plugin](https://github.com/kubernetes-sigs/secrets-store-csi-driver),
- which includes a `SecretProviderClass` CSI plug point and a simpler
- [single node host path
- plugin](https://github.com/kubernetes-csi/csi-driver-host-path/),
- though for our read only purposes, prototyping has shown we can make
- it multi-node.
-- Both provide PV and ephemeral options.  We only need the ephemeral option.
-- The host path is simpler, but generally considered "POC" level
-- The secret store employs a `SecretProviderClass`, which we do not need.
-- So we will create a new plugin repo where we pick elements from either of these samples as it suits our purposes.
-- And delete all the PV related stuff we do not need.
-- Our new CSI plugin would be hosted as a `DaemonSet` on the k8s cluster, to get per node/host granularity
-- and that `DaemonSet` will have privileged containers.  See the
-  [csi-node-driver-registrar](https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/kubernetes-1.17/hostpath/csi-hostpath-plugin.yaml#L48-L52)
-  container and the main [csi plugin
-  container](https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/kubernetes-1.17/hostpath/csi-hostpath-plugin.yaml#L82-L83)
-  for the hostpath example and [secrets store
-  container](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/deploy/secrets-store-csi-driver.yaml#L61)
-  in the secrets store example.
-- When provisioning volumes, the plugin can access the node/host
-  filesystem, like for example the directory, corresponding to the
-  Pod's mount point, where the secret contents will be stored (though
-  we have some choices on exactly which OS level file system features
-  we employ to store the content ... see the Open Questions)
-- All the reference implementations reviewed to date leverage https://github.com/kubernetes/utils/blob/master/mount/mount.go
+The CSI driver will be deployed as a DaemonSet so that shared Secrets/ConfigMaps can be added to pods running on the node.
+Pods for this DaemonSet will need to run as privileged to access host paths on the node.
+The CSI driver will only need to support "ephemeral" volumes that are specified by the `csi:` volume type on Pods.
+There is no requirement for this CSI driver to provision persistent volumes.
 
 Some details around inline ephemeral volumes:
 - The employ a subset of the CSI specification flows, specifically only the `NodePublish` and `NodeUnpublish` CSI steps
 - the underlying k8s api object `CSIVolumeSource` has the same `volumeAttributes` for storing metadata to be passed to the CSI driver
 - there is only a single `LocalObjectReference` for a `Secret` reference in the same namespace as the consuming pod.  So that aspect does not help us, and will not factor into the solution.
 - The `volumeAttributes` field for metadata gives us enough capabilities for expressing both the name, namespace, and type of sharable API object to be consumed.  And who is requesting access.
+
+#### Enhancements to Security Context Constraints
+
+OpenShift uses Security Context Constraints (SCCs) to regulate the capabilities granted to Pods, based on the user creating the pod and the service account assigned to the pod.
+Amongst other things, SCCs determine if a pod can mount a specific volume type.
+SCCs deny mounting a volume unless its type is explicitly allowed.
+The most restrictive SCC only allows volume types that are known to be isolated from the host filesystem (`secret`, `configMap`, `persistentVolumeClaim`, etc.).
+
+SCCs as designed today have an "all or nothing" approach when it comes securing CSI volume mounts.
+This is not sufficient as cluster admins are free to install third party CSI drivers that support `csi` volume mounts.
+Because we want to deliver this CSI driver in the payload, we need ensure that this CSI driver is usable while ensuring third party CSI drivers can be restricted.
+
+The following API extends SCCs to restrict the allowed CSI drivers used for ephemeral CSI volumes, in the same manner that Flexvolumes are restricted:
+
+```go
+type SecurityContextConstraints struct {
+  // existing API
+  ...
+  //
+
+  Volumes []FSType
+
+  // AllowedCSIVolumes is an allowlist of permitted ephemeral CSI volumes.
+  // Empty or nil indicates that all CSI drivers may be used.
+  // This parameter is effective only when the usage of CSI volumes is allowed in the "Volumes" field.
+	// +optional
+  AllowedCSIVolumes []AllowedCSIVolume
+}
+
+// AllowedCSIVolume represents a single CSI volume driver that is allowed to be used.
+type AllowedCSIVolume struct {
+  // Driver is the name of the CSI driver that supports ephemeral CSI volumes.
+  Driver string
+}
+```
+
+With this API, the default SCCs delivered via OpenShift can be updated to include the projected resource CSI driver as an allowed driver.
 
 #### Some details around prototype efforts to date
 
@@ -329,8 +343,7 @@ anyway.
 
 *Mitigations:*
 
-- OLM version of the CSI driver operator will only be made available as a community operator
-- Users will be informed that this operator will need to be uninstalled prior to upgrading to the version of OpenShift that deploys this CSI driver via the OCP payload.
+- There will be no OLM operator at outset, as the intent is to distribute this driver via the OCP payload. 
 
 **Risk:** CSI driver will not scale
 
--- a/enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
+++ b/enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
@@ -23,7 +23,7 @@ replaces:
 
 creation-date: 2020-03-17
 
-last-updated: 2020-09-01
+last-updated: 2021-04-19
 
 <!-- status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced -->
 status: implementable
@@ -50,53 +50,33 @@ status: implementable
 >unmount associated tmpfs for the SA in questions (assuming tmpfs per SA scales) so the data is no longer available from
 >that location.  But since it could have copied the data elsewhere, do we need to kill the Pod?  If so, would the CSI
 >driver even be allowed to initiate killing a Pod?  If so, what is required and what approach should be employed?
-> 2. CVO vs. OLM for delivery.  May need @derekwaynecarr to cast the deciding vote.  Latest pro-con list below in the
->implementation details install section.  Top level question to answer is do we want this feature installed on
->all OCP clusters, or have it be an optional component.  If optional, OLM is the defacto answer.  If it always needs to
->be there, does that mandate CVO, or as some reviewers have speculated, *perhaps not in the near future*.
 
 ## Summary
 
-This proposal will describe a [CSI driver](https://github.com/container-storage-interface/spec/blob/master/spec.md) based
-means for OpenShift cluster administrators to share `Secrets` / `ConfigMaps` defined in one namespace with
-other namespaces.
+This proposal will describe a [CSI driver](https://github.com/container-storage-interface/spec/blob/master/spec.md) based means for OpenShift cluster administrators to share `Secrets` / `ConfigMaps` defined in one namespace with other namespaces.
 
-Any owners/authors/users of Pods in those namespaces can opt into the consumption of the content of those shared `Secrets` /
- `ConfigMaps` during their Pod's execution.
-
-This proposal was broken off from [a proposal on injecting RHEL entitlements/subscriptions](https://github.com/openshift/enhancements/pull/214)
-**EDITORIAL NOTE:** if/when PR 214 merges the reference link will be adjusted to point to the merged document hosted on
-https://github.com/openshift/enhancements.  Or if that PR is replaced by another proposal, or this proposal, this proposal will
-point that out.
-
+Any owners/authors/users of Pods in those namespaces can opt into the consumption of the content of those shared `Secrets` / `ConfigMaps` during their Pod's execution.
 
 ## Motivation
 
 Simplify provision and consumption of `Secrets` and `ConfigMaps` defined in one namespace by pods in other namespaces.
+There are numerous use cases where information should be defined only once on the cluster and shared broadly, such as:
+
+- Distribution of custom PKI certificate authorities, such as corporate self-signed certificates.
+- Simple Content Access certificates used to access RHEL subscription content.
+  See also the [Subscription Content Access](/enhancements/subscription-content/subscription-content/access.md) proposal.
 
 ### Goals
 
 - Provide easy to use sharing of `Secrets` and `ConfigMaps` defined in one namespace by pods in other namespaces.
-
-- When a shared `Secret` or `ConfigMap` changes, the container filesystem content the CSI driver injects into the Pod
-needs to reflect those changes.
-
-- When a cluster-admin revokes access to shared `Secrets` or `ConfigMaps` that was previously granted, that data should
-no longer be accessible.
+- When a shared `Secret` or `ConfigMap` changes, the container filesystem content the CSI driver injects into the Pod needs to reflect those changes.
+- When a cluster-admin revokes access to shared `Secrets` or `ConfigMaps` that was previously granted, that data should no longer be accessible.
 
 ### Non-Goals
 
-1. This proposal assumes the sensitive content encapsulated in the `Secret` or `ConfigMap` is valid and correct for the user's purposes.
-No validation of the content for specific usage within a Pod will be done.  That said, this proposal provides
-for watching for updates to those objects.
-So if the cluster administrator realized on their own they have provided bad data, and then corrects it, this proposal
-will watch for, discover, and propagate the update when new Pods consuming the content are provisioned.
-
-2. Protecting this content from being viewed/copied by users: if it's available for the Pod to use, the user who can
-create a Pod that can utilize this feature can also read / view / exfiltrate the Secret/ConfigMap contents. It is technically
-impossible to avoid that.
-
-3. This proposal for now assumes linux only for the underlying host/node.  No windows, etc.
+- Validate the content within a shared `Secret` or `ConfigMap`.
+- Prevent content in shared `Secrets` or `ConfigMaps` from being copied.
+- Support for Windows nodes/workloads.
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -118,53 +98,23 @@ the credentials, I want to stop updating content within the existing pods consum
...
-- this is where using ephemeral CSI storage comes in
-- using ephemeral based storage allows for use of a tmpfs (virtual memory filesystem) for storage
-- tmpfs is fast and has negligible overhead
-- it will have to reconstructed if the Pod moves
-- but how it eases dealing with SELinux protection is the bonus we are taking
-- so each pod gets its own tmpfs
-- the kubelet leverages tmpfs in a similar fashion already
+The projected resource CSI driver will provision ephemeral volumes and take advantage of upstream's ephemeral CSI storage mechanisms.
+Ephemeral storage allows the CSI driver to use tmpfs for storage while providing appropriate protection via SELinux labels.
+Each pod will have its own copy of the shared resource that the CSI driver mounts in, and there will be a 1:1 correspondence between the pod and the SELinux label applied to the mounted files.
```
gabemontero, line 111: "The kubelet manages SELinux labelling for us" was a pertinent phrase I remember from the conversations with @jsafrane, and is a result of the choice made with the first three bullets here

If we are reworking the phrasing here, I say we add that as a way of tying together *why* we are doing it this way .. it also leads into the last bullet below IMO

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -213,118 +163,94 @@ the underlying, associated Pod references CSI volumes using `Shares`
...
-reference to said `Secret` or `ConfigMap`
-- creation of a tmpfs filesystem for each requesting `ServiceAcount` that contains the `Secret` / `ConfigMap` data
-- creation of k8s watches `ProjectedResources`, and for `Secret` / `ConfigMaps` in the namespaces listed in each discovered
-`ProjectedResources`
-- final returning of a volume that is based off of the `ServiceAccount` specific tmpfs for the data noted in the
-`volumeAttributes`
+
+With all this necessary metadata, the CSI driver then:
+
+- reads of the `Secret` or `ConfigMap` noted in the `ProjectedResource`, storing it on disk on initial reference to said `Secret` or `ConfigMap`.
```
gabemontero, line 205: a failure in the prior form of the EP @adambkaplan that we can correct since you are reworking.... SAR checks are performed on the initial provisioning as well

Some brief detail on the SAR check seems useful as well, and should be included:  namely the SAR check sees if the serviceaccount for the Pod to be provisioned has permission to get the referenced Share

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -213,118 +163,94 @@ the underlying, associated Pod references CSI volumes using `Shares`
...
-- as updates of the watched `Secrets` and `ConfigMaps` come in, SAR checks will be done for each of the previous requesting
-`ServiceAccounts` being tracked against the updated `Secret` and `ConfigMap`
-- for any `ServiceAccounts` that no longer have access, their tmpfs filesystems are unmounted, effectively removing their access to
-the data (though in prototyping and implementation, we'll need focused testing that this scales and mount propagation
-does not become a gating factor)
-- the SAR checks and if need be removal of access to data can facilitate cluster admin scenarios noted in the `User Stories`
-section like credential rotation.
-- But for `ServiceAccounts` that pass the SAR check, the tmpfs file with the data for the `Secret` / `ConfigMap` is updated
 
+- as updates of the watched `Secrets` and `ConfigMaps` come in, SAR checks will be done for each of the previous requesting `ServiceAccounts` being tracked against the updated `Secret` and `ConfigMap`.
```
gabemontero, line 212: - steady state also performs SARs when the relist for the existing Shares occurs in the controller for any active pods using those shares ... if a Pod loses permissions, the tmpfs is unmounted

- and certainly if a Share/ Share's referenced ConfigMap/ referenced Secret are removed, those also result in the tmpfs getting unmounted

gabemontero (changes requested):
some minor feedback @adambkaplan 

as much from gaps in the existing EP that your modifications highlighted for me as anything

but also a suggested tweak to some of your consolidation of text around the "why tmpfs per pod selinux labelling" portion of changes

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -213,118 +163,94 @@ the underlying, associated Pod references CSI volumes using `Shares`
...
+- for any `ServiceAccounts` that no longer have access, their tmpfs filesystems are unmounted, effectively removing their access to the data (though in prototyping and implementation, we'll need focused testing that this scales and mount propagation does not become a gating factor).
```
jsafrane, line 213: I don't think you can unmount a filesystem from under a running pod (you may get `EBUSY`). Clearing the tmpfs (`rm -rf`) would be probably better.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -213,118 +163,94 @@ the underlying, associated Pod references CSI volumes using `Shares`
...
-- creation of k8s watches `ProjectedResources`, and for `Secret` / `ConfigMaps` in the namespaces listed in each discovered
-`ProjectedResources`
-- final returning of a volume that is based off of the `ServiceAccount` specific tmpfs for the data noted in the
-`volumeAttributes`
+
+With all this necessary metadata, the CSI driver then:
+
+- reads of the `Secret` or `ConfigMap` noted in the `ProjectedResource`, storing it on disk on initial reference to said `Secret` or `ConfigMap`.
+- creation of a tmpfs filesystem for each requesting `ServiceAcount` that contains the `Secret` / `ConfigMap` data.
+- creation of k8s watches `ProjectedResources`, and for `Secret` / `ConfigMaps` in the namespaces listed in each discovered `ProjectedResources`.
```
jsafrane, line 207: How many watches per cluster do you expect? This can explode easily.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -50,53 +50,33 @@ status: implementable
...
 ### Goals
 
 - Provide easy to use sharing of `Secrets` and `ConfigMaps` defined in one namespace by pods in other namespaces.
-
-- When a shared `Secret` or `ConfigMap` changes, the container filesystem content the CSI driver injects into the Pod
-needs to reflect those changes.
-
-- When a cluster-admin revokes access to shared `Secrets` or `ConfigMaps` that was previously granted, that data should
-no longer be accessible.
+- When a shared `Secret` or `ConfigMap` changes, the container filesystem content the CSI driver injects into the Pod needs to reflect those changes.
```
jsafrane, line 72: Kubernetes did the mistake that Secrets and ConfigMap are automatically refreshed. With hundreds of thousands of pods in a cluster, watching and refreshing the ConfigMaps and Secrets takes significant amount of resources. Later on, Kubernetes introduces opt-out from updating content of these volumes.

IMO it would make sense either to start with opt-in for the refresh of the volumes (but it might create confusion) or follow Kubernetes and have opt-out.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -118,53 +98,23 @@ the credentials, I want to stop updating content within the existing pods consum
...
-- using ephemeral based storage allows for use of a tmpfs (virtual memory filesystem) for storage
-- tmpfs is fast and has negligible overhead
-- it will have to reconstructed if the Pod moves
-- but how it eases dealing with SELinux protection is the bonus we are taking
-- so each pod gets its own tmpfs
-- the kubelet leverages tmpfs in a similar fashion already
+The projected resource CSI driver will provision ephemeral volumes and take advantage of upstream's ephemeral CSI storage mechanisms.
+Ephemeral storage allows the CSI driver to use tmpfs for storage while providing appropriate protection via SELinux labels.
+Each pod will have its own copy of the shared resource that the CSI driver mounts in, and there will be a 1:1 correspondence between the pod and the SELinux label applied to the mounted files.
+The process will be similar to what exists in core Kubernetes when Secrets and ConfigMaps are mounted into a pod.
```
jsafrane, line 112: In Kubernetes we have some tricks to make sure content of the volume is changed in an atomic way, so the Pod gets always correct content when a secret is updated in the middle of Pod's `read()`. Applying something like that here would be beneficial. See https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/atomic_writer.go

(btw, as consequence, we made Secrets and ConfigMap volumes as read-only in the pods, because users could fool the atomic writer to access the host filesystem by rogue symlinks).

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -399,14 +325,18 @@ anyway.
...
+*Mitigations:*
 
+- OLM version of the CSI driver operator will only be made available as a community operator
+- Users will be informed that this operator will need to be uninstalled prior to upgrading to the version of OpenShift that deploys this CSI driver via the OCP payload.
+
+**Risk:** CSI driver will not scale
+
+*Mitigations:*
+
+- Minimize local file system writes to when the Secret actually changes.
```
jsafrane, line 339: Please include some scale tests. How many ProjectedResources  do you expect? How many namespaces / pods that use them do you expect?

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -118,53 +98,23 @@ the credentials, I want to stop updating content within the existing pods consum
...
-- Using OLM avoids unexpected hiccups in the feature's installation impacting initial cluster install or cluster upgrade.
-
-#### CSI starting points
-
-**WE ARE GOING WITH EPHEMERAL MODE**
+For the developer preview release, the CSI driver will be installable via OLM.
+This release channel is meant to be a preview only option and allow the CSI driver to be installed on OpenShift outside of the normal release cycle.
 
-**NO SecretProvideClass IS NEEDED**
+When the CSI driver is ready for general availability, it will be installed on OpenShift via the cluster storage operator and an associated delivery operator - see the [CSI driver install proposal](/enhancements/storage/csi-driver-install.md).
```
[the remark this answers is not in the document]
gabemontero: yeah this is fluid at the moment ... based on some slack conversations with the apiserver team around https://github.com/openshift/enhancements/pull/733 (both in that EP and slack), and based on a conversation I just had with @adambkaplan I think he is moving off of the dev preview via OLM piece and will just deliver this in the OCP payload via the CSO stuff now as part of 4.9

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: Yep this needs to be corrected @adambkaplan 

And yep @jsafrane that is in fact what we are doing.  

- SAR / permission check:  https://github.com/openshift/csi-driver-projected-resource/blob/master/pkg/hostpath/hostpath.go#L302-L316
- which translates to https://github.com/openshift/csi-driver-projected-resource/blob/master/pkg/hostpath/hostpath.go#L344-L356
- and the `rm -rf` equivalent at https://github.com/openshift/csi-driver-projected-resource/blob/master/pkg/hostpath/hostpath.go#L225-L239

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: Yep I share this concern @jsafrane and also have asked for scheduled scale testing per you comment below

The driver already employs a form of namespace filtering to control which namespaces we watch, but it needs to be "productised" before we bring things into the payload and claim GA status, with ways to configure and finer grained control, so that the watches for configmaps and secrets can be on different namespaces if desired.  @adambkaplan has Jira / Epics recorded for this work.

But some mention of that here or wherever appropriate in the EP should be done @adambkaplan ... assuming you want to take that on with this PR.  Or let me know if you want me to follow up with a subsequent PR.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: I'm good with adding that feature as well.

@adambkaplan if you agree as well, can you add?

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: Yep we have scale testing as part of the GA / productization epic/stories I mentioned above.  I believe I can add that as part of the existing e2e's I have (albeit as optional at least initially).

But yes let's make noted of that here @adambkaplan

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: Yeah I looked at the atomic_writer.go when originally crafting things. but I did not use it explicitly because of the k8s magic linking stuff, where I did not want to follow that path, IIRC.  And then at the time I did not copy/paste/simplify that logic per se moving forward.

But in hindsight yeah I know see what you are getting at.  And agree we should revisit.

There is also the read only stuff.  At the moment, the driver does not support read only, but between this discussion and other discussion in https://github.com/openshift/enhancements/pull/733 I think that needs to be added to the productization epic.

So @adambkaplan if you agree, I think the new questions/features to note in the EP and track are 
- do we want to update after creation at all
- if so, opt into updating or opt out of updating
- approximate / reuse as much as possible k8s/k8s atomic_writer.go.  Though using it as is means vendoring in https://github.com/kubernetes/kubernetes which at first blush I do not want to take on, so approximate/copy seems the thing.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: yeah I was trying to capture "work out of the box" with my last comment @jsafrane 

I'm fine with adding that verbiage instead as well @adambkaplan 

But I think some indication that we want to defer to default k8s mechanisms for handling SELinux is a useful thing to do.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
jsafrane: With minor refactoring it could be possible to move the atomic writer to `kubernetes/utils`, which has almost no deps and is easily consumable. If we can satisfy "_Used by multiple Kubernetes repositories_" criteria.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
adambkaplan: Would it make sense for us to have a separate controller that labels Secrets and ConfigMaps that are shared? That way the watches of the CSI driver daemons only act on secrets that have a specific label key, whose value maps to the appropriate share instance. Unclear if this would reduce the volume of watch events sent to the driver.

This separate controller introduces its own set of issues, such as leader election. Unlike the storage driver, we only want one active controller that labels Secrets and ConfigMaps based on which resources have been shared.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
adambkaplan: I can see two forms of scale testing needed:

1. Number of total shares in the cluster - how many can a cluster handle?
2. Number of pods that can share a single share. RHEL entitlements come to mind - what if 1000 pods mounted the same simple content access cert? What happens if that SCA is rotated and 1000 pods consume it?

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
adambkaplan: Since we are already concerned with scaling, I would prefer we allow opting in for refreshes. For the Build use case we won't need entitlements to be refreshed.

gabemontero: sure thing .... so aside from any updates to this EP, either add this to our existing "productization" jira or create a new jira to get this added

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -213,158 +163,148 @@ the underlying, associated Pod references CSI volumes using `Shares`
...
+
+// AllowedCSIVolume represents a single CSI volume driver that is allowed to be used.
+type AllowedCSIVolume struct {
+  // Driver is the name of the CSI driver that supports ephemeral CSI volumes.
+  Driver string
+}
+```
+
+With this API, the default SCCs delivered via OpenShift can be updated to include the projected resource CSI driver as an allowed driver.
+
```
adambkaplan, line 301: @deads2k per our discussion on Slack, I've added a section here to extend SCCs to support these CSI drivers. I borrowed the API used to restrict flex volumes - is this still a valid approach?

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -465,22 +429,43 @@ There are several drawbacks with these approaches:
...
+  - So the cluster admin sets that SELinux label on the `HostPath` volume and namespace where it will be consumed.
+    Admin would need to create a new `SecurityContextConstraint` that is based off of the `hostpath-anyuid` SCC (which is defined on install) and includes the SELinux label can be created.
+  - However, the cluster administrator has to be careful with the mount provided.
+    Mounts at certain parts of the filesystem can basically root the node.
+- Host injection via the MCO requires a node reboot for the updates to take effect.
+  That is generally deemedundesirable for this feature.
+- The MCO API has not been intuitive for the type of users to date who are interested in introducing and then consuming small collections of files on a host.
+  The RHCOS team has experienced challenges in coaching consumers who want to put entitlements on 4.x nodes via the MCO.
+
+### Implement a provider for the Secret Store CSI Driver
```
adambkaplan, line 443: @jsafrane @gabemontero @deads2k @derekwaynecarr @smarterclayton 

This is an alternative approach that we should think about, especially if we want to present something upstream.
At minimum we should have an answer as to why we _don't_ want to pursue this approach.

In many respects this approach can work as long as we get the RBAC story right.
The secret store CSI driver (at present) restricts use on the namespace level.
We could propose SARs on the `SecretProvierClass` object as an opt-in feature.
The cluster-wide `Share` object then becomes a way to control which content is globally available.

We'd also want a means to configure the secret store providers via the Cluster Storage Operator.
There are some secret store providers that are cloud provider specific, as well as HashiCorp's Vault provider that maintained by a third party.

gabemontero: based on our conversation this morning @adambkaplan do you think we've properly captured the disadvantages of this approach?  both in the EP text and comments here?

the entire driver/api is still alpha
secrets per namespace in addition to SecretProviderClass per namespace (IIRC)I f
I feel like I'm forgetting an item or 2 ....

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: My concerns also included
3. Number of namespaces to monitor for secrets/configmapas
4. Number of total secrets/configmaps as a result of the number of namespaces we watch

1. is something as well in my mind, but not as big a deal for me as 3 and 4 ... I think in practice the number of shares will be relatively smaller
2.  is something that we should pay due diligence on, but is the least of my concerns based on a conversation with @derekwaynecarr during a recent on top pillar arch call .... bottom line, this boils down the tmpfs mounts, of which clusters already have a bunch of since that is how secrets work ... he expressed confidence to me this would be "OK"

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -213,158 +163,154 @@ the underlying, associated Pod references CSI volumes using `Shares`
...
-- the underlying k8s api object `CSIVolumeSource` has the same `volumeAttributes` for storing metadata to be passed to the CSI driver
-- there is only a single `LocalObjectReference` for a `Secret` reference in the same namespace as the consuming pod.  So that aspect does not help us, and will not factor into the solution.
-- The `volumeAttributes` field for metadata gives us enough capabilities for expressing both the name, namespace, and type of sharable API object to be consumed.  And who is requesting access.
+- The underlying k8s api object `CSIVolumeSource` has the same `volumeAttributes` for storing metadata to be passed to the CSI driver
+- There is only a single `LocalObjectReference` for a `Secret` reference in the same namespace as the consuming pod.
+  So that aspect does not help us, and will not factor into the solution.
+- The `volumeAttributes` field for metadata gives us enough capabilities for expressing both the name, namespace, and type of sharable API object to be consumed.
+  Using pod info on mount lets the driver see which service account is requesting access.
```
gabemontero, line 265: this also provides for us
- pod name
- pod namespace 
- pod uid
- the fact we are using ephemeral

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -213,158 +163,154 @@ the underlying, associated Pod references CSI volumes using `Shares`
...
-`ServiceAccounts` being tracked against the updated `Secret` and `ConfigMap`
-- for any `ServiceAccounts` that no longer have access, their tmpfs filesystems are unmounted, effectively removing their access to
-the data (though in prototyping and implementation, we'll need focused testing that this scales and mount propagation
-does not become a gating factor)
-- the SAR checks and if need be removal of access to data can facilitate cluster admin scenarios noted in the `User Stories`
-section like credential rotation.
-- But for `ServiceAccounts` that pass the SAR check, the tmpfs file with the data for the `Secret` / `ConfigMap` is updated
 
+- As updates of the watched `Secrets` and `ConfigMaps` come in, SAR checks will be done for each of the previous requesting `ServiceAccounts` being tracked against the updated `Secret` and `ConfigMap`.
+  SAR checks are also done at regular relist intervals, which will pick up permission changes to associated cluster roles/role bindings.
```
gabemontero, line 214: relist of the Secrets, ConfigMaps, and Shares all result in reexecuting the SARs

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -199,214 +153,90 @@ rules:
...
-- Pods with this annotation are mutated, adding the CSI driver volume specification,
-- if the Pod `ServiceAccount` is allowed sufficient
-access to the requested `ProjectedResource` (i.e. a SAR check for the SA getting the requested `ProjectedResource` passes)
-- if a Pod already happens to have the CSI Volume specified with reference to a `ProjectedResource`, again the plugin will see if the
-associated `ServiceAccount` for the Pod has sufficient access to the `ProjectedResource`
-- if sufficient access does not exist, the admission plugin marks the Pod as invalid/forbidden.
-- if sufficient access does exists, the admission plugin adds a CSI volume into the Pod's `volume` array, to match the
-example noted above.
-
+The cluster admin is then responsible for assigning this cluster role to appropriate serivce accounts that consume the share (either directly or via role aggregation).
```
adambkaplan, line 156: Shortened the RBAC requirements for brevity. We assume a cluster admin knows how to do this.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -199,214 +153,90 @@ rules:
...
+      volumeMounts:
+      - name: cluster-share
+        mountPath: /path/to/shared-item-data
+    volumes:
+    - name: cluster-share
+      csi:
+        driver: projectedresources.storage.openshift.io
+        volumeAttributes:
+          share: the-projected-resource
+  ```
```
adambkaplan, line 181: Cleaned up the example for mounting the CSI volume.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -199,214 +153,90 @@ rules:
...
-- But for `ServiceAccounts` that pass the SAR check, the tmpfs file with the data for the `Secret` / `ConfigMap` is updated
+With all this necessary metadata, the CSI driver then:
 
+- Performs a subject access review (SAR) check to verify the pod's service account has permission to GET the referenced Share resource.
+- If the SAR check succeeds, reads of the `Secret` or `ConfigMap` noted in the `Share`, storing it on disk on initial reference to said `Secret` or `ConfigMap`.
+- Creates a tmpfs filesystem for each requesting `ServiceAcount` that contains the `Secret` / `ConfigMap` data.
+- Creates k8s watches `Shares`, and for `Secret` / `ConfigMaps` in the namespaces listed in each discovered `Share`.
+- Returns a volume that is based off of the `ServiceAccount` specific tmpfs for the data noted in the `volumeAttributes`.
 
-#### An overview of CSI and the main K8S hook points
```
adambkaplan, line 291: Moved this section to an appendix with further details

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
adambkaplan: Do we need this level of detail here?

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -199,214 +153,90 @@ rules:
...
+- The SAR checks and if need be removal of access to data can facilitate cluster admin scenarios noted in the `User Stories` section like credential rotation.
+- But for `ServiceAccounts` that pass the SAR check, the tmpfs file with the data for the `Secret` / `ConfigMap` is updated
 
-```
+#### Ephemeral CSI driver implementation
 
-- And then the [CSI plugin to the Kubelet](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#ephemeral-inline-volume-proposal), performs the necessary steps to engage the CSI driver.
-- the [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md) details the CSI architecture, concepts, and protocol
-
-#### Current thoughts on implementation, how to inject/consume content on every host/node in an OpenShift cluster
```
adambkaplan, line 325: Moved this to the appendix

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -431,12 +261,27 @@ Some links related to that testing:
 
 ### Graduation Criteria
 
-We will not want to GA this EP as is until the related work noted in the Summary is complete.
+#### Developer Preview
 
+While the CSI driver is in developer preview state, cluster admins should be able to install via YAML manifests defined in GitHub.
+
+#### Tech Preview
```
adambkaplan, line 268: Since we are delivering CSI drivers on a tech preview basis, there is no issue IMO providing this new driver as tech preview in the payload.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -448,15 +293,17 @@ N/A
 
 ## Drawbacks
 
-N/A
+TODO
```
adambkaplan, line 296: Probably need to remark the complexity of implementing a CSI driver, delivering a driver via the storage operator.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -512,4 +380,77 @@ And some details from our PV based prototyping with the hostpath examples (again
 
 ## Infrastructure Needed
 
-N/A
+- GitHub repositories for the CSI driver and operator
+- Add the CSI driver and operator to the OpenShift release payload (ART)
+
+## Appendix
```
adambkaplan, line 386: I copied the deeper R&D and proof of concept results here. I think we should keep this information for future reference, however it gets in the way of the proposal if left in higher sections.

adambkaplan (commented):
@gabemontero @jsafrane incorporated a bit more of your feedback here. Please take a look regarding releasing as tech preview.

I think the only thing to complete is the drawbacks section.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: > So I read this as:
> 
>     1. If we only sync content at volume creation, said volume can be writeable. We should support read-only as an option.
> 

Not 100% sure I'm parsing your item 1)  correctly @adambkaplan but I *think* you are on the right path.  I'm going to tweak your wording a bit, to better reconcile with what the current code does, and how code changes might manifest.

Today, since we always send Secret/ConfigMap updates from the controller watch to the tmpfs mount that backs the CSI volume, we always create that mount with read-write permissions.

However, if we introduce a mode where we never send Secret/ConfigMap updates after the initial provisioning of the CSI volume and its underlying tmpfs mount, we could set things up where the tmpfs mount that baacks the CSI volume is read only.

Implementation detail: I'll have to refresh my memory when I try to implement this, but IIRC, we will need to create a writeable tmpfs that we initially seed with the in memory ConfigMap / Secret contents from the controller, then bind mount that to a second read only tmpfs mount, where that second read only tmpfs mount is what is provided to the Pod for its CSI volume.  We cannot just create a read only tmpfs mount, then write data to it.

>     2. If we do sync content, we should ensure the volume is read-only.
> 

Less sure about this statement @adambkaplan 

So I'm assuming here you are separating the fact that he CSI Pod volume is marked read-only from whether the tmpfs mount that backs the CSI volume is read-write, because if it was read-only, we cannot write the new ConfigMap/Secret data as part of "syncing" content.

If you were separating the read write nature of the pod volume from the read write nature of the underlying tmpfs mount, I'm not sure it works that way.  By that, I mean I'm not sure if there is a mechanism other than OS level file permissions that prevents logic running in the Pod from writing to the volume.

If there is such a mechanism I'm unaware of it.

Or as an alternative, perhaps it is possible to do "the sync"  by a) unmount the existing tmpfs volume with the old ConfigMap/Secret contents (where it is a bind mount of another tmpfs volume we can update ... a "side care writeable tmpfs mount"), b) update our "side car writeable tmpfs mount", then c) bind mount a new read only tmpfs from the "side care writeable tmpfs" with the new ConfigMap/Secret contents to the same path as the tmpfs we unmounted in a)

^^ seems doable assuming it is not a gross abuse of tmpfs and it does not violate the k8s volume contract in some what I am unaware of 

If you were referring to such a method @adambkaplan, then "OK", though it took some reading between the lines to get there.

> 
> @gabemontero @jsafrane is this correct?

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -118,53 +101,24 @@ the credentials, I want to stop updating content within the existing pods consum
...
-- tmpfs is fast and has negligible overhead
-- it will have to reconstructed if the Pod moves
-- but how it eases dealing with SELinux protection is the bonus we are taking
-- so each pod gets its own tmpfs
-- the kubelet leverages tmpfs in a similar fashion already
+The projected resource CSI driver will provision ephemeral volumes and take advantage of upstream's ephemeral CSI storage mechanisms.
+Ephemeral storage allows the CSI driver to create tmpfs filessytems for each pod, where the Secret/ConfigMap content will be copied.
+The tmpfs filesystems will automatically have appropriate SELinux labeling to ensure content does not leak between pods - the driver does not need to provision or manage these labels.
+The process will be similar to what exists in core Kubernetes when Secrets and ConfigMaps are mounted into a pod.
+Acutal implementation can borrow a significant amount of logic from the [secrets-store-csi-driver](https://github.com/kubernetes-sigs/secrets-store-csi-driver), which is used to mount secrets from external "sealed" secret providers like [Vault](https://learn.hashicorp.com/tutorials/vault/kubernetes-secret-store-driver).
```
gabemontero, line 116: This is a bit of a leap, though certainly that implementation leverages tmpfs mounts.  

And I think your simplification has removed the key points that were the original motivation / intent of the content you've replaced.
- tmpfs mounts per csi pod volume gets us "SELinux compliance" for free
- tmpfs mounts are fast, cheap resource wise
- tmpfs mounts == k8s ephemeral based storage volumes

I'm fine if you want to distill my verbose list into something more concise, but I think some form of these 3 bullets is what is needed here

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -199,214 +153,90 @@ rules:
...
+  spec:
+    containers:
+      ...
+      volumeMounts:
+      - name: cluster-share
+        mountPath: /path/to/shared-item-data
+    volumes:
+    - name: cluster-share
+      csi:
+        driver: projectedresources.storage.openshift.io
```
gabemontero, line 178: at least at the moment, the driver name is `csi-driver-projected-resource.openshift.io`  .... per https://github.com/openshift/csi-driver-projected-resource/blob/master/examples/csi-app.yaml#L18

But of course if we envision changing the name as part of our path to the CVO, that is good as well.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: LGTM

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: SGTM

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: IMO I think the details are relevant in that implication of how it works affects the behavior perceived by users.

Specifically, if permissions are removed, the Share content is *eventually* removed from the Pods affected by the permission change, but not immediately.  There is a time gap between whenever the permission change occurs and the next relist (default 10 or 15 minutes IIRC).

An alternate approach, but one that does not exist yet, aside from *maybe* copying a bunch of internal SCC related code which is not exposed for public consumption, is to listen to changes to clusterroles/clusterrolebindings etc., then reverse engineer whether a clusterrole/clusterrolebinding combination is what allowed a given Pod SA to access a given share.  If we could do this, there would be no perceived delay between permission removal and content removal.

If anything, I think we keep the content here, and perhaps better emphasize the current potential delay here, and this list some form of the nirvana potential future I quickly noted above in the alternatives section.

And to be fair, my original EP that merged last year lacked any exposure to this IMO.

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
gabemontero: SGTM

gabemontero:
> @gabemontero @jsafrane incorporated a bit more of your feedback here. Please take a look regarding releasing as tech preview.
> 
> I think the only thing to complete is the drawbacks section.

OK finished my pass at your latest updates @adambkaplan 

a fair amount of sgtm / plus ones along with a few opinions on questions you posed, and a couple of counters to changes you made

thanks

enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
jsafrane: Unmounting existing tmpfs *is* gross. There is some time where an application does not see any config / secret file and it may break the app in unexpected ways.

You should ensure that whatever you mount inside NodePublish is read-only. You can mount a read-write tmpfs *somewhere*, prepare the projected files there, and then bind-mount it read-only to the directory given to you in `NodePublish`. Thinking about it, it becomes hard to maintain, because you need keep track of these read-write mounts and they must be able to survive CSI driver pod stop + restart (e.g. during CSI driver update).

gabemontero: Yeah @jsafrane  I was worried about the grossness of the unmount / mount to get updates.  Also thanks for the confirm on the 2 tmpfs / bind mount 1 to the other for read only notion I mentioned earlier in this thread.

I do not disagree on the maintenance wrt driver restart, but I have in fact dove into the CSI driver pod stop / restart scenario, persisting state related to live pods, etc.  I do think this is a doable progression.

To bubble back up though @adambkaplan , IMO your original item 2) needs to be changed.

If we do sync content, the volume needs to be read-write.

As we do not want to go down the unmount / remount path.

But that does mean we need some form of k8s atomic writer equivalent.

Which circles back to considering turning off the sync content feature for now, as builds does not need it, and then drill down on the need of it for the other scenarios like cert rotation, etc. which we solve later, with more runway.

gabemontero: OK so I've been able to experiment with things in this space a bit and have some finding to reports, including some corrections to things I stated in this threads earlier.

- the 2 tmpfs with the "top level" bind mount to the pod destination approach does work
- by writing to the "lower tmpfs", we can change content at that layer, but still have pod destination be read only
- I even added some `touch /data/foo` bash invocations in my pod to confirm that failed when the pod volume was marked read only, but it succeeded when the pod volume was read write
- my prior statements on the direct correlation between tmp fs read/write perms and the pods volume read/write perms was incorrect because I forgot that the kubelet add its own mount on top of things (`/var/kubelet/....` or some such)

So moving forward, we can definitely support a mode where our driver refreshes the content (to support cert rotation scenarios for example), but we do *NOT* allow pod volumes based on our driver to be provisioned unless the `readOnly: true`.

I was able to validate both elements today with pre-merge branch I have.

I'll next move onto updating / validating the e2e's I have around changing data, restarting the driver / pod, etc..  But at first blush, I'm reasonably confident we can maintain this additional element of state, and bypass the need for a k8s/k8s atomic writer equivalent.

gabemontero: One amendment to ^^

We don't have to worry about multiple writers, but IIRC there is still the concern of a Pod trying to *read* its mounted content while we are updating it.

gabemontero: So my restart the driver tests have uncovered that more than I hoped for will be involved with adding the second tmpfs layer, but still being able to change the data for the Pod after the driver restart.

I'm still experimenting but I suspect the "gross" unmount/remount of tmpfs options previously discussed above might be needed.

I'll report back when I land on something definitive.

But falling back on the previously discussed staging plan where we support the build scenario with full read only and push out support for rotating certs, etc. which require update may be where we land short term.

PR #745, opened 2021-04-20
Security Profiles Operator integration in OpenShift

JAORMX:
The intent is to propose a deeper integration of the [Security Profiles Operator](https://github.com/kubernetes-sigs/security-profiles-operator)
in order to increase the usability of runtime security for users of the platform.

### DIFF
--- a/enhancements/security/security-profiles-operator.md
+++ b/enhancements/security/security-profiles-operator.md
@@ -49,6 +49,8 @@ as part of the OpenShift platform in order to allow folks to:
 
 * Enable users to automatically generate profiles for their workloads
 
+At some later point:
+
 * Provide predefined profiles for existing default workloads in OpenShift
 
 
@@ -78,15 +80,22 @@ the need to run privileged containers
 
 ### Goals
 
-* Integrate the Security Profiles Operator into OpenShift by having it
-  installed by default via the Cluster Version Operator (CVO).
+* Enable application developers to ship their applications with
+  relevant security profiles
 
-* Integrate the SPO into OpenShift's SCCs.
+* Enable application developers to easily record the security profiles
+  for their application.
 
-* Enable application developers to ship their operators with
-  relevant security profiles (OLM integration)
+* Integrate the SPO into OpenShift's SCCs if possible. It is not strictly
+  required, but would increase the usefulness of SCCs and would be a good
+  security hardening in general. The backup plan is to ship some example
+  custom SCCs and add documentation for users.
 
 ### Non-Goals
+
+* Integrate the Security Profiles Operator into OpenShift by having it
+  installed by default via the Cluster Version Operator (CVO). While this is
+  a long-term goal, it is not what we're trying to do in the first iteration.
 
 * Developing of profiles for all OpenShift components. While having the OpenShift
   components locked down is a long-term goal and SPO might come with several profiles
@@ -95,10 +104,11 @@ the need to run privileged containers
 
 ## Proposal
 
-The Security Profiles Operator should be installed by default in OpenShift
-via the CVO as any other core component. As it is, its main function is to
-install security policies and ensure that they're attached to the appropriate
-pods which require them.
+The Security Profiles Operator should be available for installation as an
+add-on. Its main function is to install security policies and ensure that
+they're attached to the appropriate pods which require them. Additionally,
+the Security Profiles Operator enables recording of SELinux and seccomp
+profiles, which can be used e.g. in CI runs.
 
 The installation is managed by a DaemonSet workload (Security Profiles Operator Daemon
 or SPOD) that listens for security profiles (either `SeccompProfile` or
@@ -272,14 +282,13 @@ with the privileges that the application needs.
 
 The `SeccompProfile` objects rely on the [OCI Seccomp BPF Hook](https://github.com/containers/oci-seccomp-bpf-hook)
 to do the actual recording. So we'd need to ensure this hook is available in RHCOS
-before enabling this feature.
+before enabling this feature. Failing that, SPO allow to record `SeccompProfile`
+objects using its built-in BPF recorder or by tailing `audit.log`.
 
-There isn't currently support for `SelinuxProfile` recording, but
-the work is scheduled to start in the 4.9 cycle, and the intent
-is to leverage [Udica](https://github.com/containers/udica) to generate
-the profile.
+To record SELinux profiles, SPO tails `audit.log`, looking for AVC messages
+and transforms them into a `SelinuxProfile` objects.
 
-### SCC integration
+### SCC integration (non-blocking)
 
 Security Context Constraints allow a user to restrict the security settings
 that a workload may have. Given that the profile settings are in the pod's
@@ -387,6 +396,10 @@ This will prevent workloads trying to:
 
 * Maliciously gain privileges by using a profile installed from another
   namespace.
+
+NOTE: This was filed as [RFE-2204](https://issues.redhat.com/browse/RFE-2204)
+but so far not accepted. The backup plan is to just include an example
+SCC using the existing SCC capabilities and provide documentation to users.
 
 ### OLM integration
 
--- a/enhancements/security/security-profiles-operator.md
+++ b/enhancements/security/security-profiles-operator.md
@@ -1,0 +1,636 @@
+---
+title: security-profiles-operator
+authors:
+  - "@jaormx"
+  - "@jhrozek"
+  - "@saschagrunert"
+reviewers:
+  - "@ashcrow"
+  - "@mrunalp"
+  - "@sttts"
+  - "@mrunalp"
+  - "@shawn-hurley"
+  - "@dmesser"
+approvers:
+  - TBD
+creation-date: yyyy-mm-dd
+last-updated: yyyy-mm-dd
+status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced|informational
+see-also:
+  - "/enhancements/this-other-neat-thing.md"
+replaces:
+  - [selinux-operator proposal](https://github.com/openshift/enhancements/pull/327)
+superseded-by:
+  - "/enhancements/our-past-effort.md"
+---
+
+# Security Profiles Operator Integration in OpenShift
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Let's integrate the [Security Profiles Operator](https://github.com/kubernetes-sigs/security-profiles-operator)
+as part of the OpenShift platform in order to allow folks to:
+
+* Easily install security profiles to secure their workloads
+  (both SELinux and Seccomp)
+
+* Manage the lifecycle of the aforementioned profiles (make sure
+  that they get attached to the appropriate container as upgrades
+  happen)
+
+* Enable users to automatically generate profiles for their workloads
+
+* Provide predefined profiles for existing default workloads in OpenShift
+
+
+## Motivation
+
+OpenShift already has several security controls enabled by default. Amongst them
+are the following:
+
+* SELinux to restrict what resources can containers interact with
+
+* Seccomp to restrict the system calls that containers are allowed to do
+
+* Security Context Constraints to control what security knobs and handles are
+  service accounts able to tweak in their workloads.
+
+The default SELinux and Seccomp profiles work great for restricted workloads.
+However, once the workload needs a little more permissions, setting these
+security settings becomes non-trivial. In face of this, developers often
+respond by turning off these controls and running their workloads as privileged
+containers.
+
+The [Security Profiles Operator (SPO)](https://github.com/kubernetes-sigs/security-profiles-operator)
+is an upstream project aiming to ease installation and development of security
+profiles in Kubernetes. By using these profiles, developers would have a more
+granular way of setting permissions for their workloads, hopefully eliminating
+the need to run privileged containers
+
+### Goals
+
+* Integrate the Security Profiles Operator into OpenShift by having it
+  installed by default via the Cluster Version Operator (CVO).
+
+* Integrate the SPO into OpenShift's SCCs.
+
+* Enable application developers to ship their operators with
+  relevant security profiles (OLM integration)
+
+### Non-Goals
+
+* Developing of profiles for all OpenShift components. While having the OpenShift
+  components locked down is a long-term goal and SPO might come with several profiles
+  for some selected workloads, the intent is not to ship a very wide library of profiles
+  with the initial OpenShift inclusion
+
+## Proposal
+
+The Security Profiles Operator should be installed by default in OpenShift
+via the CVO as any other core component. As it is, its main function is to
+install security policies and ensure that they're attached to the appropriate
+pods which require them.
+
+The installation is managed by a DaemonSet workload (Security Profiles Operator Daemon
+or SPOD) that listens for security profiles (either `SeccompProfile` or
+`SelinuxProfile`) and ensures that these are installed on the nodes.
+
+The SPOD instance is configurable through a CRD which allows us to set where
+it's scheduled and the capabilities that are enabled/disabled. For OpenShift,
+SELinux would be enabled by default and the DaemonSet would be scheduled
+everywhere.
+
+`SeccompProfiles` are handled by simply persisting a file on a directory that
+the Kubelet has access to. A `SeccompProfile` looks as follows:
+
+```yaml
+---
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
+kind: SeccompProfile
+metadata:
+  name: profile-complain-block-high-risk
+  annotations:
+    description: "Enables complain mode whilst blocking high-risk syscalls. Some essential syscalls are allowed to decrease log noise."
+spec:
+  defaultAction: SCMP_ACT_LOG
+  architectures:
+  - SCMP_ARCH_X86_64
+  syscalls:
+  - action: SCMP_ACT_ALLOW
+    names:
+    - exit
+    - exit_group
+    - futex
+    - nanosleep
+
+  - action: SCMP_ACT_ERRNO
+    names:
+    - acct
+    - add_key
+    - bpf
+    - clock_adjtime
+    - clock_settime
+    - create_module
+    - delete_module
+    - finit_module
+    - get_kernel_syms
+    - get_mempolicy
+    - init_module
+    - ioperm
+    - iopl
+    - kcmp
+    - kexec_file_load
+    - kexec_load
+    - keyctl
+    - lookup_dcookie
+    - mbind
+    - mount
+    - move_pages
+    - name_to_handle_at
+    - nfsservctl
+    - open_by_handle_at
+    - perf_event_open
+    - personality
+    - pivot_root
+    - process_vm_readv
+    - process_vm_writev
+    - ptrace
+    - query_module
+    - quotactl
+    - reboot
+    - request_key
+    - set_mempolicy
+    - setns
+    - settimeofday
+    - stime
+    - swapoff
+    - swapon
+    - _sysctl
+    - sysfs
+    - umount2
+    - umount
+    - unshare
+    - uselib
+    - userfaultfd
+    - ustat
+    - vm86old
+    - vm86
+```
+
+As seen, the object is simply a wrapper around the JSON formatting, which
+allows us to do extra validations on the objects, thus ensuring easily that
+it's appropriately formatted.
+
+`SelinuxProfiles` require extra work, as to install a SELinux policy
+one has to parse the policy, compile it to the main system's policy
+and then install it in the system (which keeps it in memory). To do this
+an extra component called [selinuxd](https://github.com/JAORMX/selinuxd) has
+been developed. The intent is that selinuxd allows one to simply drop a file
+on a pre-determined directory to install a SELinux policy, and that policy will
+get picked up and installed by the daemon. Thus, the SPOD instance simply
+places the policy into a directory, the selinuxd instance picks it up
+and handles the installation and tracking of policies.
+
+A `SelinuxProfile` object looks as follows:
+
+```yaml
+---
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
+kind: SelinuxProfile
+metadata:
+  name: errorlogger
+  namespace: my-namespace
+spec:
+  policy: |
+    (blockinherit container)
+    (allow process var_log_t ( dir ( open read getattr lock search ioctl add_name remove_name write ))) 
+    (allow process var_log_t ( file ( getattr read write append ioctl lock map open create  ))) 
+    (allow process var_log_t ( sock_file ( getattr read write append open  )))
+```
+
+It is mostly a wrapper around [SELinux's CIL language](https://github.com/SELinuxProject/cil)
+with the main block and naming handled by the operator. Naming of policies
+as implemented by the daemon is as follows `<object name>_<object namespace>`. Thus,
+we're able to distinguish between policies installed in different namespaces.
+
+To ease usability and allow for developers to easily bind profiles to workloads,
+an object called `ProfileBinding` was created. The intent is for the binding to
+declare what profile goes to a ceratain image. The binding is then enforced by a
+webhook which mutates the pod definition to add the appropriate profile
+to its security context.
+
+The `ProfileBinding` object looks as follows:
+
+```yaml
+---
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
+kind: ProfileBinding
+metadata:
+  name: profile-binding
+spec:
+  profileRef:
+    kind: SeccompProfile
+    name: profile-allow-unsafe
+  image: nginx:1.19.1
+```
+
+This would help manage security profiles and match them to workloads
+accross the application's lifecycle. Where one could upgrade their
+workload, and create a binding that matches the new application's image
+tag.
+
+Finally, to aide developers in writing profiles, there is a `ProfileRecording`
+object. The object would look as follows:
+
+```yaml
+---
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
+kind: ProfileRecording
+metadata:
+  # The name of the Recording is the same as the resulting `SeccompProfile` CRD
+  # after reconciliation.
+  name: test-recording
+spec:
+  kind: SeccompProfile
+  podSelector:
+    matchLabels:
+      app: alpine
+```
+
+The operator automates the recording of profiles, and, when the
+workload is done, it's able to output a ready-to-use profile
+with the privileges that the application needs.
+
+The `SeccompProfile` objects rely on the [OCI Seccomp BPF Hook](https://github.com/containers/oci-seccomp-bpf-hook)
+to do the actual recording. So we'd need to ensure this hook is available in RHCOS
+before enabling this feature.
+
+There isn't currently support for `SelinuxProfile` recording, but
+the work is scheduled to start in the 4.9 cycle, and the intent
+is to leverage [Udica](https://github.com/containers/udica) to generate
+the profile.
+
+### SCC integration
+
+Security Context Constraints allow a user to restrict the security settings
+that a workload may have. Given that the profile settings are in the pod's
+`securityContext` it is imperative that we appropriately integrate with SCCs.
+
+Currently, the only SCC that allows users to set the Seccomp and SELinux
+settings for a pod is the privileged one. It looks as follows:
+
+```bash
+Name:                                           privileged
+Priority:                                       <none>
+Access:                                         
+  Users:                                        system:admin,system:serviceaccount:openshift-infra:build-controller
+  Groups:                                       system:cluster-admins,system:nodes,system:masters
+Settings:                                       
+  Allow Privileged:                             true
+  Allow Privilege Escalation:                   true
+  Default Add Capabilities:                     <none>
+  Required Drop Capabilities:                   <none>
+  Allowed Capabilities:                         *
+  Allowed Seccomp Profiles:                     *
+  Allowed Volume Types:                         *
+  Allowed Flexvolumes:                          <all>
+  Allowed Unsafe Sysctls:                       *
+  Forbidden Sysctls:                            <none>
+  Allow Host Network:                           true
+  Allow Host Ports:                             true
+  Allow Host PID:                               true
+  Allow Host IPC:                               true
+  Read Only Root Filesystem:                    false
+  Run As User Strategy: RunAsAny                
+    UID:                                        <none>
+    UID Range Min:                              <none>
+    UID Range Max:                              <none>
+  SELinux Context Strategy: RunAsAny            
+    User:                                       <none>
+    Role:                                       <none>
+    Type:                                       <none>
+    Level:                                      <none>
+  FSGroup Strategy: RunAsAny                    
+    Ranges:                                     <none>
+  Supplemental Groups Strategy: RunAsAny        
+    Ranges:                                     <none>
+```
+
+This basicallly allows for setting all security settings for a pod, which
+might not be what a user intends to allow.
+
+To enable integration with the SPO and promote better security in workloads,
+we could add another SCC called `measured` that would look similar to this:
+
+```bash
+Name:                                           measured
+Priority:                                       <none>
+Access:                                         
+  Users:                                        <none>
+  Groups:                                       <none>
+Settings:                                       
+  Allow Privileged:                             false
+  Allow Privilege Escalation:                   false
+  Default Add Capabilities:                     <none>
+  Required Drop Capabilities:                   <none>
+  Allowed Capabilities:                         *
+  Allowed Seccomp Profiles:                     *
+  Allowed Volume Types:                         *
+  Allowed Flexvolumes:                          <all>
+  Allowed Unsafe Sysctls:                       *
+  Forbidden Sysctls:                            <none>
+  Allow Host Network:                           true
+  Allow Host Ports:                             true
+  Allow Host PID:                               true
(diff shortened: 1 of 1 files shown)


enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,627 @@
+---
+title: security-profiles-operator
+authors:
+  - "@jaormx"
+  - "@jhrozek"
+  - "@sascha"
```
saschagrunert, line 6:
```suggestion
  - "@saschagrunert"
```

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,627 @@
...
+
+* Easily install security profiles to secure their workloads
+  (both SELinux and Seccomp)
+
+* Manage the lifecycle of the aforementioned profiles (make sure
+  that they get attached to the appropriate container as upgrades
+  happen)
+
+* Get auto-generated security profiles that ease the barrier of
+  entry of securing workloads.
```
saschagrunert, line 48: I would split-up this point into:
- enable users to automatically generate profiles for their workloads
- provide predefined profiles for existing default workloads in OpenShift

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,627 @@
...
+      app: alpine
+```
+
+The operator automates the recording of profiles, and, when the
+workload is done, it's able to output a ready-to-use profile
+with the privileges that the application needs.
+
+The `SeccompProfile` objects rely on the [OCI Seccomp BPF Hook](https://github.com/containers/oci-seccomp-bpf-hook)
+to do the actual recording. So we'd need to ensure this hook is available in RHCOS
+before enabling this feature.
```
saschagrunert, line 266: Let's mention that we need CRI-O v1.21.0 for this to work properly beside the OCI seccomp BPF hook. Let's also mention that the operator feature will not work in case both conditions are not met, but this has no influence on the overall functionality of the operator. This means the operator can also be used on previous OpenShift versions.

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,627 @@
...
+Where SELinux and Seccomp profile definitions are linked to a
+Service Account, similar to other objects in the CSV.
+
+### User Stories
+
+The personas it aims to help are documented in [the project's documentation](
+https://github.com/kubernetes-sigs/security-profiles-operator/blob/master/doc/personas.md).
+
+The user stories that are being covered are [also appropriately documented](
+https://github.com/kubernetes-sigs/security-profiles-operator/blob/master/doc/user-stories.md)
```
saschagrunert, line 436:
```suggestion
https://github.com/kubernetes-sigs/security-profiles-operator/blob/master/doc/user-stories.md).
```

saschagrunert:
cc @mrunalp

enhancements/security/security-profiles-operator.md
JAORMX: @saschagrunert the CRI-O version depends on the OpenShift version. The intent is to target this towards OpenShift 4.10 and not any older version. CRI-O 1.21 IIRC is available in 4.9, so we're good on that side.

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,629 @@
...
+it's scheduled and the capabilities that are enabled/disabled. For OpenShift,
+SELinux would be enabled by default and the DaemonSet would be scheduled
+everywhere.
+
+`SeccompProfiles` are handled by simply persisting a file on a directory that
+the Kubelet has access to. A `SeccompProfile` looks as follows:
+
+```yaml
+---
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
```
sttts, line 110: we don't want to ship alpha APIs, do we?

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,629 @@
...
+#### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+- Enumerate service level indicators (SLIs), expose SLIs as metrics
+- Write symptoms-based alerts for the component(s)
+
+#### Tech Preview -> GA
```
sttts, line 538: v1 API

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,629 @@
...
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+**Examples**: These are generalized examples to consider, in addition
+to the aforementioned [maturity levels][maturity-levels].
+
+#### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- End user documentation, relative API stability
```
sttts, line 532: no alpha API

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,629 @@
...
+
+To enable integration with the SPO and promote better security in workloads,
+we could add another SCC called `measured` that would look similar to this:
+
+```bash
+Name:                                           measured
+Priority:                                       <none>
+Access:                                         
+  Users:                                        TBD
+  Groups:                                       TBD
```
sttts, line 332: none?

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,629 @@
...
+  (both SELinux and Seccomp)
+
+* Manage the lifecycle of the aforementioned profiles (make sure
+  that they get attached to the appropriate container as upgrades
+  happen)
+
+* Enable users to automatically generate profiles for their workloads
+
+* Provide predefined profiles for existing default workloads in OpenShift
+
```
sttts, line 50: I miss a very clear statement about dev-preview/tech-preview/GA. The API is alpha upstream. What is the timeline for a non-alpha API ? As alpha I think this can be a dev-preview at best.

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,629 @@
...
+authors:
+  - "@jaormx"
+  - "@jhrozek"
+  - "@saschagrunert"
+reviewers:
+  - "@ashcrow"
+  - "@mrunalp"
+approvers:
+  - TBD
+  - "@oscardoe"
```
sttts, line 12: this needs the auth team as approver as they own SCCs. Also OLM team must be here.

enhancements/security/security-profiles-operator.md
JAORMX: @sttts that's why this is marked as WIP in the title. Stuff like this needs to be determined. having it as dev-preview in 4.10 sounds acceptable.

enhancements/security/security-profiles-operator.md
JAORMX: TBD == To Be Done. We need to give some thought to this.

enhancements/security/security-profiles-operator.md
JAORMX: we don't indeed

enhancements/security/security-profiles-operator.md
JAORMX: will add, thanks!

JAORMX:
@sttts thanks for the review! I'll address the comments

enhancements/security/security-profiles-operator.md
saschagrunert: All our API's are alpha for now (planned for the upcoming v0.3.0 release). We may consider graduating them and changing the release versioning model (otherwise the graduation would not make that much sense together with semver.

Created an issue about this topic in https://github.com/kubernetes-sigs/security-profiles-operator/issues/407

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,629 @@
...
+  Allow Privilege Escalation:                   false
+  Default Add Capabilities:                     <none>
+  Required Drop Capabilities:                   <none>
+  Allowed Capabilities:                         *
+  Allowed Seccomp Profiles:                     *
+  Allowed Volume Types:                         *
+  Allowed Flexvolumes:                          <all>
+  Allowed Unsafe Sysctls:                       *
+  Forbidden Sysctls:                            <none>
+  Allow Host Network:                           false
```
jhrozek, line 344: The SCC should probably allow the host resources. In general I think that the SCC should be pretty much the privileged one minus the capability to run privileged containers.

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+everywhere.
+
+`SeccompProfiles` are handled by simply persisting a file on a directory that
+the Kubelet has access to. A `SeccompProfile` looks as follows:
+
+```yaml
+---
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
+kind: SeccompProfile
+metadata:
```
stlaz, line 119: IIRC there was a lot of indecision upstream as to how to deal with seccomp profiles and what should be the default, it's nice to see that you're trying to work with that your own way.

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+  podSelector:
+    matchLabels:
+      app: alpine
+```
+
+The operator automates the recording of profiles, and, when the
+workload is done, it's able to output a ready-to-use profile
+with the privileges that the application needs.
+
+The `SeccompProfile` objects rely on the [OCI Seccomp BPF Hook](https://github.com/containers/oci-seccomp-bpf-hook)
```
stlaz, line 273: Which containers would be affected by this syscall monitoring? What's the performance impact?

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+to do the actual recording. So we'd need to ensure this hook is available in RHCOS
+before enabling this feature.
+
+There isn't currently support for `SelinuxProfile` recording, but
+the work is scheduled to start in the 4.9 cycle, and the intent
```
stlaz, line 278: Do you have an enhancement we can read?

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+          - apiGroups:
+            - ""
+            resources:
+            - nodes
+            verbs:
+            - get
+            - list
+            - watch
+          serviceAccountName: some-sa
+      securityProfiles:
```
kevinrizza, line 415: OLM is intentionally trying to move away from the cluster service version and eventually in a direction where arbitrary kubernetes manifests are dropped into an operator bundle that OLM applies as the installation step. We already have support for a number of other manifests. Is there a reason why we couldn't simply add support in OLM for these manifest types?

JAORMX: I was not aware of the OLM moving away from CSVs. Is there an enhancement or doc about this? I can change this proposal to reflect that.

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Let's integrate the [Security Profiles Operator](https://github.com/kubernetes-sigs/security-profiles-operator)
+as part of the OpenShift platform in order to allow folks to:
```
stlaz, line 41: I see PSPs are involved in that operator. What happens to it when PSPs are removed in kube 1.22 and the way how SCCs and PSPs work diverges?

stlaz (commented):
A few questions in text.

Is this going to be supported and maintained by the compliance team?

enhancements/security/security-profiles-operator.md
kevinrizza: this is very out of date, but it does go into some of the high level concepts: https://github.com/operator-framework/enhancements/blob/master/enhancements/csvless-bundles.md

But more specifically, OLM already supports specific resource types other than CSV and CRD in operator bundles today. For example, we support certain Prometheus and ServiceMonitor objects as raw kube manifests applied by the bundle, along with some rbac related resources. I think that, while it might take us a little while before a set of arbitrary kube manifests replace the CSV entirely, we don't want to continue expanding the CSV api and instead OLM should just add these object types as support bundle objects to move this forward.

JAORMX:
> A few questions in text.
> 
> Is this going to be supported and maintained by the compliance team?

Yes. This is going to be supported by the Compliance Team; The plan is that also, alongside, the Node team (especially @saschagrunert ) is going to support this as well.

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+* Security Context Constraints to control what security knobs and handles are
+  service accounts able to tweak in their workloads.
+
+The default SELinux and Seccomp profiles work great for restricted workloads.
+However, once the workload needs a little more permissions, setting these
+security settings becomes non-trivial. In face of this, developers often
+respond by turning off these controls and running their workloads as privileged
+containers.
+
+The [Security Profiles Operator (SPO)](https://github.com/kubernetes-sigs/security-profiles-operator)
```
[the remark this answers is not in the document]
JAORMX: @stlaz in the initial pitch of this, the idea was to bring it as a Marketplace operator. But it was brought up that having it as a Marketplace Operator first would make it really hard for folks to adopt the security profiles, as it would not be sure that the operator would be there. So... if you try to install an operator that uses a security profile, the operator would be uninstallable unless the SPO would be there. Hence why we switched to proposing this as a CVO operator already.

enhancements/security/security-profiles-operator.md
JAORMX: only annotated containers get recorded. So, be default none would be unless you explicitly annotate your container.

enhancements/security/security-profiles-operator.md
JAORMX: PSPs are not involved in the operator.

enhancements/security/security-profiles-operator.md
JAORMX: Not at the moment, it's in my backlog to write one :smile: . The intent is to work on that in 4.9. And introduce this operator to OpenShift in 4.10.

enhancements/security/security-profiles-operator.md
dmesser: I think there was an EP at some point for something that we call CSV-less bundles. But independent of that we already support a bunch of extra manifests that are applied during install. So we can definitely talk about this at any point.

JAORMX: That sounds good to me! I'll remove the reference to the CSV from this EP. Can someone pass me some docs to read that offer examples of other manifests that the OLM supports? That way I could use those as examples for this proposal.

enhancements/security/security-profiles-operator.md
stlaz: I see, I based my observation solely on the readme, not on the actual code of the operator.

saschagrunert:
Is this blocked by any process right now? Or may we remove the WIP and ask for an additional round of review? :upside_down_face:

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+    matchLabels:
+      app: alpine
+```
+
+The operator automates the recording of profiles, and, when the
+workload is done, it's able to output a ready-to-use profile
+with the privileges that the application needs.
+
+The `SeccompProfile` objects rely on the [OCI Seccomp BPF Hook](https://github.com/containers/oci-seccomp-bpf-hook)
+to do the actual recording. So we'd need to ensure this hook is available in RHCOS
```
cgwalters, line 274: Having this be an extension seems OK, but on the other hand if it is basically a no-op unless the annotation is present, then perhaps it's simplest to ship this by default (e.g. as part of crio?).  I don't have a really strong opinon.

JAORMX: Having it enabled by default would be quite nice for usability IMO.

travier: How big is this package? If it's just a couple of KB, making an extension for that would probably not be worth it.

jhrozek: > Having this be an extension seems OK, but on the other hand if it is basically a no-op unless the annotation is present, then perhaps it's simplest to ship this by default (e.g. as part of crio?). I don't have a really strong opinon.

I really don't know enough about crio to decide if having `oci-seccomp-bpf-hook` part of crio or a crio dependency makes sense. 

Just a heads-up: I was testing `oci-seccomp-bpf-hook` on OCP earlier this week and at the moment, `bcc` (a dependency of  `oci-seccomp-bpf-hook`) requires bcc-tools for some reason. This dependency is being removed, but if you try installing the packages from 8.4, you'll get  bcc-tools which brings some python deps and kernel-devel. You'll also want the hook from 8.5 because there was a bogus crun dependency. With these two deps removed, we're down to:

- oci-seccomp-bpf-hook (1MB)
- bcc (676kB)
- clang-libs (23MB)
- libbpf (101kB)
- llvm-libs (21MB)

If RHCOS can ship these by default, even better, but an extension would be fine by me. As said, we need to have the dependencies trimmed first in Fedora and RHEL, though. All details are in https://issues.redhat.com/browse/CMP-927

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+
+Selinuxd is currently deployed as a container alongside the SPOD instance.
+In order to ensure that SELinux policies are installed early on when a
+node boots, it would be ideal to have selinuxd run directly on the node
+itself (either as a systemd unit or as a static container) as opposed to
+it running on a container managed by the Kubelet.
+
+On the other hand, the repository namespace selinuxd is hosted is currently
+a personal one. It would also be ideal to move that somewhere more appropriate.
+(The coreos namespace maybe?)
```
jhrozek, line 459: @cgwalters @cverna I would like to get your thoughts on where should selinuxd be deployed.

I think that having selinuxd as a daemon on the node would be best, because then any policies dropped in via MCO would be installed before workloads start. But it is another package on the OS and in addition a daemon that always runs..

A static pod would also be a possibility, selinuxd is capable of running in a container. What I see as a benefit of a deamon on the node is that we would be able to install policies also for static pods if selinuxd is configured to be started before kubelet.

If neither of these is acceptable, selinuxd could just remain part of the SPO and run as a deamonset. But then policies might be installed later than a workload is started..

Note that selinuxd is usable even w/o SPO as a way to take SELinux policies into use if an admin distributes them via an MCO or some other mechanism.

At the moment, selinuxd lives in a personal GH repo and we publish images to quay -- so the reason I'm trying to answer where should selinuxd run would also help me find a better home for the repo (should the project be part of the coreos org? openshift org? just be folded into SPO upstream?)

Thank you!

enhancements/security/security-profiles-operator.md
dmesser: @JAORMX see here https://docs.openshift.com/container-platform/4.7/operators/understanding/olm-packaging-format.html#olm-bundle-format_olm-packaging-format

enhancements/security/security-profiles-operator.md
travier: selinuxd would probably also be useful for Fedora CoreOS too. Could you create an issue to discuss that there? Would love to see it in action in a demo.

enhancements/security/security-profiles-operator.md
JAORMX: @dmesser @kevinrizza the doc makes sense! So, how trivial is to add new types to the supported types? I guess we'd then need to add `SelinuxProfiles` and `SeccompProfiles` as supported types, so folks could then add them to their bundles and the OLM could then install them.

Are there requirements for these types? e.g. required conditions that the objects should have or a status used to interact with the OLM?

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+    Ranges:                                     <none>
+  Supplemental Groups Strategy: RunAsAny        
+    Ranges:                                     <none>
+```
+
+This basicallly allows for setting all security settings for a pod, which
+might not be what a user intends to allow.
+
+To enable integration with the SPO and promote better security in workloads,
+we could add another SCC called `measured` that would look similar to this:
```
jhrozek, line 332: @sttts Hi Stefan, I wonder if you have time to help us flesh out the design of the SCC part? Feel free to tag someone else if you're not the right person to ask of course..

What I'm thinking now is that it would be nice to have a more generic way for an SCC to allow a selinux context or a seccomp profile without having to use the privileged SCC and at the same time only restrict the security profile used to the same namespace as the workload -- as opposed to adding one more default SCC like the `measured` example below suggests. The one extra SCC wouldn't be flexible enough, I can see some workloads e.g. wanting to use only certain capabilities or only certain volume types etc. Moreover, there would still need to be a piece that checks that only security profiles from the same namespace are used and I wasn't sure if OCP would like to have a webhook that checks all pods cluster-wide (a validating webhook was my first though to restrict the workloads to a subset of security profiles).

Do you think there would be a way to allow only using a selinux or a seccomp profile based on values a project is annotated with, more or less like UID ranges are allocated now? Then the SCCs in general could be extended to allow SELinux context strategy "RunAsRange" that would look at some project annotation and a similar mechanism could be used to allow all seccomp profiles from a project. Also, this way, the SCC enforcement would be decoupled from the actual security profiles management, so this mechanism might work even if there's no SPO or its APIs.

What I don't know is, if it's possible to let some trusted component that maintains the security profiles (SPO in this case) to annotate the projects with the security profiles deployed there. I'm not super familiar with the project annotations, but a quick look shows that only the display name and description annotations can be modified.

What do you think?

enhancements/security/security-profiles-operator.md
jhrozek: > selinuxd would probably also be useful for Fedora CoreOS too. Could you create an issue to discuss that there? Would love to see it in action in a demo.

I hope I found the right place - https://github.com/coreos/fedora-coreos-tracker/issues/854

enhancements/security/security-profiles-operator.md
jhrozek: @deads2k I see you did a lot of work in the SCC are, so I wonder if you have any opinion on the above?

sttts: @s-urbaniak @stlaz

stlaz: > The one extra SCC wouldn't be flexible enough, I can see some workloads e.g. wanting to use only certain capabilities or only certain volume types etc.

Isn't that why you can create your own? Or will this whole thing only ever work with the "measured" SCC?

> there would still need to be a piece that checks that only security profiles from the same namespace are used

Why limit security profiles to namespaces? Who gets to create them? Can I craft a malicious security profile in my namespace that would let me into other namespaces should any other service pick them up?

> would be a way to allow only using a selinux or a seccomp profile based on values a project is annotated with

This is already happening for SELinux (projects have the MCS part generated, although nothing else). You might be able to force seccomp on project's workloads in a similar fashion.

> I'm not super familiar with the project annotations, but a quick look shows that only the display name and description annotations can be modified.

I believe with enough privileges, you should be able to change any annotation on a namespace.

JAORMX: > Can I craft a malicious security profile in my namespace that would let me into other namespaces should any other service pick them up?

That's right. It would be ideal if then you could only use security profiles that are "approved" in your namespace. This will help in both auditing, but also security by keeping a clear multi-tenancy.

jhrozek: > > The one extra SCC wouldn't be flexible enough, I can see some workloads e.g. wanting to use only certain capabilities or only certain volume types etc.
> 
> Isn't that why you can create your own? Or will this whole thing only ever work with the "measured" SCC?

Yes, but the point of bringing the operator to OCP is that workloads would ship their own profiles. At the moment, the SCC allows you to list allowed profiles, or allow all or none. Creating an SCC per workload that ships a security profile doesn't seem very scalable. 

> 
> > there would still need to be a piece that checks that only security profiles from the same namespace are used
> 
> Why limit security profiles to namespaces? Who gets to create them? Can I craft a malicious security profile in my namespace that would let me into other namespaces should any other service pick them up?

Yes, that's the concern, especially with seccompProfiles. In the pod definition, `pods.spec.securityContext.seccompProfile` just points to a path on the node where the seccompProfile is stored. So the idea was that SPO would annotate projects with the seccomp or selinux profiles that are deployed there and SCC would enforce their use based on the annotations.

> 
> > would be a way to allow only using a selinux or a seccomp profile based on values a project is annotated with
> 
> This is already happening for SELinux (projects have the MCS part generated, although nothing else). You might be able to force seccomp on project's workloads in a similar fashion.
> 
> > I'm not super familiar with the project annotations, but a quick look shows that only the display name and description annotations can be modified.
> 
> I believe with enough privileges, you should be able to change any annotation on a namespace.

That's where I got confused, it [seemed](https://github.com/openshift/openshift-apiserver/blob/a6220d84588dd916fc13210d3484876832fd37bd/pkg/project/apis/project/validation/validation.go#L64) like right now only the description or displayname is allowed to be modified.

Of course, maybe this whole approach is over-engineered, because we'd have to add a new "strategy" for both selinux and seccomp that would allow all profiles from a list defined in a project, which is a big change. That's why I'm bringing this up even in a vague state of design.

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+
+The [Security Profiles Operator (SPO)](https://github.com/kubernetes-sigs/security-profiles-operator)
+is an upstream project aiming to ease installation and development of security
+profiles in Kubernetes. By using these profiles, developers would have a more
+granular way of setting permissions for their workloads, hopefully eliminating
+the need to run privileged containers
+
+### Goals
+
+* Integrate the Security Profiles Operator into OpenShift by having it
```
mrunalp, line 81: Can we list the motivation for including this by default in the core?
We are working hard to reduce our platform footprint, so the bar for including something by default is very high.

From a security posture, it makes sense that we provide good defaults out of the box that work for most of our customers. For the ones that do want to further lock down or loosen their security, they could install
this operator using OLM.

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+---
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
+kind: SelinuxProfile
+metadata:
+  name: errorlogger
+  namespace: my-namespace
+spec:
+  policy: |
+    (blockinherit container)
+    (allow process var_log_t ( dir ( open read getattr lock search ioctl add_name remove_name write )))
```
mrunalp, line 215: Is their guidance on how a customer can evaluate whether the SELinux policy they are installing going to make their cluster less secure?

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,636 @@
...
+* Seccomp to restrict the system calls that containers are allowed to do
+
+* Security Context Constraints to control what security knobs and handles are
+  service accounts able to tweak in their workloads.
+
+The default SELinux and Seccomp profiles work great for restricted workloads.
+However, once the workload needs a little more permissions, setting these
+security settings becomes non-trivial. In face of this, developers often
+respond by turning off these controls and running their workloads as privileged
+containers.
```
dhellmann, line 71: What sorts of developers? OCP developers? Users? Third-party operator authors?

enhancements/security/security-profiles-operator.md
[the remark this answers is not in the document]
dhellmann: Perhaps we should be talking about having a way for OLM operators to express dependencies on other operators.

enhancements/security/security-profiles-operator.md
jhrozek: Looking at the number of operators that run as privileged, locking down workloads with either seccomp or selinux is not very common. So, all of the above.

enhancements/security/security-profiles-operator.md
jhrozek: Is there such guidance for writing SELinux policy in general? We're working on a SELinux policy recording feature for SPO, would it help to suggest that, as opposed to hand-craft the policies so that the policy would only ever allow what the app is doing (presumably in CI) ?

enhancements/security/security-profiles-operator.md
jhrozek: > Can we list the motivation for including this by default in the core?

The main motivation was to not require that installing workloads that currently require to run as privileged to install an add-on, but rather provide a way to ship policies along with the operators.

> We are working hard to reduce our platform footprint, so the bar for including something by default is very high.
> 

Nod.

> From a security posture, it makes sense that we provide good defaults out of the box that work for most of our customers. For the ones that do want to further lock down or loosen their security, they could install
> this operator using OLM.

Perhaps the way to go might be a dependency on another operator, as said elsewhere. But ultimately, we should be locking down everything, including the core platform operators. And at that point, isn't it nicer to have something like SPO that lets you treat the security policies as first-class kubernetes objects?

enhancements/security/security-profiles-operator.md
JAORMX: @dhellmann our initial "customer" is internal OCP devs. We currently have more than 150 privileged containers running on the platform out of the box. This is quite worrying... Subsequently, we often do have customers that do want to lock down their workloads and are currently unable to since there's no feasible way of doing it. This work is progressing and unblocking the platform from these limitations, and eventually we'd like this operator to be an easy way for folks to lock down their workloads.

enhancements/security/security-profiles-operator.md
JAORMX: @mrunalp could you specify what you mean by guidance?

enhancements/security/security-profiles-operator.md
JAORMX: > From a security posture, it makes sense that we provide good defaults out of the box that work for most of our customers. For the ones that do want to further lock down or loosen their security, they could install
this operator using OLM.

If we want to provide good defaults, reducing the number of privileged containers we ship on the core OCP would be a good step (right now it's around 150+ privileged containers ). We believe that having this operator be installed by default would help us get there, as it would more easily enable our devs to start locking down their workload without needing to go through extra hoops.

dhellmann:
Based on our current desire slim down the release payload generally and to make it possible to have some components be optional in clusters (#922), the bar for adding new components to the release payload is going to be pretty high. So, I'm not sure this is something we really want to do. It would be good to bring the topic up on one of the regular architecture review calls for some discussion about the ultimate goals for this work, in case there are other approaches we could take to meet them.

jhrozek:
> Based on our current desire slim down the release payload generally and to make it possible to have some components be optional in clusters (#922), the bar for adding new components to the release payload is going to be pretty high. So, I'm not sure this is something we really want to do. It would be good to bring the topic up on one of the regular architecture review calls for some discussion about the ultimate goals for this work, in case there are other approaches we could take to meet them.

Thanks for the heads-up @dhellmann 

For the short term, I'm fine with making SPO optional, like we do for e.g. Compliance Operator or File Integrity Operator. The biggest short-term driver we see is from OCP customers to help secure their workloads with SELinux and seccomp policies which is at the moment a very manual process.

Looking at OLM docs, there seems to be a way for operators to [depend on other APIs](https://operator-framework.github.io/olm-book/docs/operator-dependencies-and-requirements.html), which would probably work for the case of another operator wishing to ship their security policies using the SPO provided APIs. Do you see any issues with that?

Let me know if rewording the enhancement proposal to make it clearer the operator would be optional at least in the short term helps.

jhrozek:
On Mon, Nov 08, 2021 at 08:06:37AM -0800, Jakub Hrozek wrote:
> > Based on our current desire slim down the release payload generally and to make it possible to have some components be optional in clusters (#922), the bar for adding new components to the release payload is going to be pretty high. So, I'm not sure this is something we really want to do. It would be good to bring the topic up on one of the regular architecture review calls for some discussion about the ultimate goals for this work, in case there are other approaches we could take to meet them.
> 
> Thanks for the heads-up @dhellmann 
> 
> For the short term, I'm fine with making SPO optional, like we do for e.g. Compliance Operator or File Integrity Operator. The biggest short-term driver we see is from OCP customers to help secure their workloads with SELinux and seccomp policies which is at the moment a very manual process.
> 
> Looking at OLM docs, there seems to be a way for operators to [depend on other APIs](https://operator-framework.github.io/olm-book/docs/operator-dependencies-and-requirements.html), which would probably work for the case of another operator wishing to ship their security policies using the SPO provided APIs. Do you see any issues with that?
> 
> Let me know if rewording the enhancement proposal to make it clearer the operator would be optional at least in the short term helps.

Done in the latest iteration.

enhancements/security/security-profiles-operator.md
```diff
@@ -0,0 +1,649 @@
...
+  - "@saschagrunert"
+reviewers:
+  - "@ashcrow"
+  - "@mrunalp"
+  - "@sttts"
+  - "@mrunalp"
+  - "@shawn-hurley"
+  - "@dmesser"
+approvers:
+  - TBD
```
dhellmann, line 15: If this is still an active proposal, you'll need to update to the new template to bring in the required metadata. We need an approver assigned, and it needs to be linked to a Jira or Bugzilla ticket for tracking the implementation.

jhrozek: Will do (probably on Monday).
Are there any requirements on the Jira ticket? Would https://issues.redhat.com/browse/CMP-1091 do? (This is our team's main epic tracking the work)

dhellmann: Since we drop the "status" field here, the idea is to provide a reader with some place to track implementation progress, and an epic like CMP-1091 would do that just fine.

jhrozek: I did the update locally, but can't push it at the moment - @JAORMX who opened the PR is on vacation and currently I don't have the push rights to his repo. I think he should be back next week, though.

PR #399, opened 2020-07-13
[machine-config-operator/baremetal] MCO declarative network configuration

bcrochet:
Enhancement proposal to extend MCO for declaritive network configuration.

PR #161, opened 2019-12-18
Deploy Kubernetes-nmstate with openshift

SchSeba:
Signed-off-by: Sebastian Sch <sebassch@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/openshift/enhancements/161)

PR #747, opened 2021-04-21
Enable Kubernetes NMstate by default for selected platforms 

yboaron:
Network interface configuration is a frequent requirement for some platforms, particularly baremetal.
In recent releases work was completed (see [1]) to enable Kubernetes NMstate (optionally via OLM) so that declarative configuration of secondary network interfaces is possible.

To improve the user-experience it is desirable to enable the NMState API by default on selected platforms where such configuration is common.
Having Kubernetes NMstate deployed by CVO will allow for example secondary NIC configuration via NodeNetworkConfigurationPolicy manifests at install time

[1] https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/mco-network-configuration.md

### DIFF
--- a/enhancements/network/kubernetes-nmstate-operator-to-cvo.md
+++ b/enhancements/network/kubernetes-nmstate-operator-to-cvo.md
@@ -1,0 +1,284 @@
+---
+title: add-kubernetes-nmstate-operator-to-cvo
+authors:
+  - "@bcrochet"
+  - "@yboaron"
+  - "@hardys"
+reviewers:
+  - "@cgwalters"
+  - "@derekwaynecarr"
+  - "@russellb"
+approvers:
+  - TBD
+creation-date: 2021-03-16
+last-updated: 2021-04-23
+status: provisional
+see-also:
+  - "/enhancements/machine-config/mco-network-configuration.md"
+---
+
+# Add Kubernetes NMState Operator to release payload
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Enable Kubernetes NMstate by default for selected platforms.
+
+Network interface configuration is a frequent requirement for some platforms, particularly baremetal.
+In recent releases [work was completed]("/enhancements/machine-config/mco-network-configuration.md")
+to enable Kubernetes NMstate (optionally via OLM) so that declarative configuration
+of secondary network interfaces is possible.
+
+To improve the user-experience it is desirable to enable the NMState API by default on selected platforms
+where such configuration is common.
+
+## Motivation
+
+In previous discussion it was noted that we may [not want NMState enabled on all platforms](https://github.com/openshift/enhancements/pull/161#discussion_r433303754) and also that [all tech-preview features must require opt-in](https://github.com/openshift/enhancements/pull/161#discussion_r433315534).
+
+Now, the support status of NMState is moving from tech-preview to fully supported, the main consideration is how to conditionally enable this only on platforms where it's likely to be needed.
+
+If we can enable Kubernetes NMstate as part of the release payload, we can improve the user experience on platforms where it is required - for example allow for NodeNetworkConfigurationPolicy to be provided via manifests at install-time via openshift-installer.
+
+Additionally, having the Kubernetes NMstate as part of the release payload will allow uniformity between the various platforms (CNV, IPI-Baremetal).
+
+Currently, CNV uses [Hyperconverged Cluster Operator - HCO](https://github.com/kubevirt/hyperconverged-cluster-operator) to deploy [custom operator - CNAO](https://github.com/kubevirt/cluster-network-addons-operator#nmstate), the CNAO installs Kubernetes NMstate, among other network related items.
+While IPI-Baremetal uses a [new operator](https://github.com/openshift/kubernetes-nmstate/blob/master/manifests/kubernetes-nmstate-operator.package.yaml) to install Kubernetes NMstate using OLM.
+
+## Q&A
+
+**Q:** why NMSTATE capability should run everywhere (on-prem, cloud, edge) in all delivery models (self-managed, hosted) ?
+
+**A:** If it was supported/possible, it would have been better to deploy nmstate capability only in required platforms.
+
+
+**Q:** Is the signal/noise ratio by having this API/controller everywhere worth the potential confusion for an SRE in a particular infra, location, or delivery model when debugging an issue?
+
+**A:** For the tech-preview phase we have a lightweight controller that only runs a deployment (replica=1) and NMSTATE CRD definition ,
+creating NMSTATE CR instance triggers the full API controller deployment.
+With this approach, we shouldn't have the full NMSTATE API everywhere.
+
+
+**Q:** What existing SLO could deliver the content and not confuse the user? e.g. could Cluster Network Operator deliver this content optionally based on its own local config like storage/csi?
+
+**A:** TBD
+
+
+**Q:** What is the resource budget for this API/controller?
+To enable it by default, we should understand its impact to the control plane:
+- write rate to the API server
+- number of clients that write to the resource (is it per node? is there a single controller co-located with api-server?)
+To enable it by default, we should understand its impact to each worker:
+- what does it watch?
+- per node daemon resource cost (cpu/memory)
+- per node daemon priority (is it node critical?)
+- per node authority to write to the corresponding resource (think node restriction admission plugin in k8s as example)
+- any unique backup/restore semantics introduced by the component
+
+**A:** The resources composed of operator and handler (operand) resources.
+Check [this document](https://docs.google.com/document/d/1qFumb1P01NOIl1gCq79Q0I8BkeaA5UQY9spqlOJjy-I/edit?usp=sharing) for detailed description of resources.
+
+
+**Q:** What is Nmstate  future impact to other topologies (single node or externalized control planes)?
+For single node
+understanding what namespace hosts the component, and ensuring it fits in the cpuset isolation for mgmt components
+For externalized control planes
+is this needed only for the 'first' mgmt cluster and its workers?
+is this needed for every worker in all agnostic clusters, and therefore should be hosted with an externalized control plane because its needed to support any 'join worker' flow.
+
+**A:** SNO deployments using none platform will be unaffected as this operator will not be enabled by default.
+
+TODO - do we need a way to enable this for none-platform, particularly for UPI metal as ideally we’d want all metal customers to have access to the same day-2 API (even if not by default)?
+
+For externalized controlplane deployments this would be needed for both the management cluster and the externalized controlplane, so that the configuration of NICs can be handled in a consistent way - it’s an open question how we’d handle this in the hybrid platform case though...
+
+
+**Q:** is it ok to deliver the content in a release payload, but enable it via some other configuration knob?
+
+**A:** The desire is to enable a common enabled-by-default API for network configuration (initially only for selected platforms where that is commonly needed).
+
+If that approach is not acceptable then delivering via the release payload with some way to opt-in would be reasonable.
+
+
+**Q:** What is the knob to turn on/off the capability? This is related to the SLO chosen to deliver this new meta-operator and what configuration it reads to know to turn it on. For example, if this is for selected platforms, the component would read the Infrastructure CR to determine if it should enable it all, but is there another knob that would allow opt-in/out?
+
+**A:** The current proposal is to couple this to the platform detected in the Infrastructure CR, so it can be enabled by default on selected platforms (particularly baremetal).
+
+There could also be some other interface at the SLO level to override that behavior and/or to enable opt-in for platforms where it’s not enabled by default.
+
+
+### Goals
+
+- Provide NMState APIs by default on desired platforms, [NetworkManager configuration should be updated](https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/mco-network-configuration.md#option-c-design) to support this capability on these platforms.
+- Allow using NMState APIs at install-time, where NIC config is a common requirement.
+
+### Non-Goals
+
+- Make persistent networking changes to nodes. That should be handled by Machine Config Operator ( check [this link](https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/mco-network-configuration.md#option-c) for more details).
+- Provide an API for controlplane network configuration via openshift-install (although in future a common API would be desirable for controlplane and secondary network interfaces)
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a user of Baremetal IPI I want to provide NodeNetworkConfigurationPolicy resources via manifests at install-time, to simplify
+my deployment workflow and avoid additional post-deploy steps.
+
+I'd like to avoid dependencies on additional registries in disconnected environments and enable core platform functions
+such as network configuration directly from the release payload.
+
+#### Story 2
+
+As a user of Openshift Virtualization I want to apply a NodeNetworkConfigurationPolicy to create network resources at nodes that will connet to VMs nics and use the already deployed kubernetes-nmstate version from openshift to reduce deployment complexity and maintenance.
+
+
+#### Story 3
+
+As a user of Baremetal IPI I want to configure (for example MTU, bonding) secondary network interfaces used for storage traffic.
+
+### Risks and Mitigations
+
+Exposes a method to modify host networking. If a configuration is applied that breaks connectivity to the API, then it will be rolled back automatically.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. Is it possible to install Kubernetes NMstate only on certain platforms (maybe using [cluster profiles](https://github.com/openshift/enhancements/blob/e598d529ba89e29bc0b48bfdb9818710a6392414/enhancements/update/cluster-profiles.md) )?
+
+2. Is it acceptable to adopt the [Cluster Baremetal Operator approach](https://github.com/openshift/enhancements/blob/4f5171c5a10f5fdeebc37cb6c6db85e3222e3ffe/enhancements/baremetal/an-slo-for-baremetal.md#not-in-use-slo-behaviors) ?
+
+### Test Plan
+
+Will be tested with an E2E test suite that also runs upstream. There is also a unit test suite.
+
+### Graduation Criteria
+
+**Note:** *Section not required until targeted at a release.*
+
+Define graduation milestones.
+
+These may be defined in terms of API maturity, or as something else. Initial proposal
+should keep this high-level with a focus on what signals will be looked at to
+determine graduation.
+
+Consider the following in developing the graduation criteria for this
+enhancement:
+
+- Maturity levels
+  - [`alpha`, `beta`, `stable` in upstream Kubernetes][maturity-levels]
+  - `Dev Preview`, `Tech Preview`, `GA` in OpenShift
+- [Deprecation policy][deprecation-policy]
+
+Clearly define what graduation means by either linking to the [API doc definition](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning),
+or by redefining what graduation means.
+
+In general, we try to use the same stages (alpha, beta, GA), regardless how the functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+**Examples**: These are generalized examples to consider, in addition
+to the aforementioned [maturity levels][maturity-levels].
+
+#### Dev Preview -> Tech Preview
+
+- Kubernetes NMState Operator is currently available in the Red Hat Catalog as Tech Preview
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Conduct load testing
+- E2E testing
+
+#### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
+
+### Upgrade / Downgrade Strategy
+
+If applicable, how will the component be upgraded and downgraded? Make sure this
+is in the test plan.
+
+Consider the following in developing an upgrade/downgrade strategy for this
+enhancement:
+
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade in order to keep previous behavior?
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade in order to make use of the enhancement?
+
+Upgrade expectations:
+
+- Each component should remain available for user requests and
+  workloads during upgrades. Ensure the components leverage best practices in handling [voluntary disruption](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/). Any exception to this should be
+  identified and discussed here.
+- Micro version upgrades - users should be able to skip forward versions within a
+  minor release stream without being required to pass through intermediate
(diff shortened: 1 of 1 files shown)


markmc:
I don't think we can justify this simply on the basis that installing an optional operator is "inconvenient and cumbersome"

In https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/mco-network-configuration.md I think we acknowledge that this configuration falls under the domain of the MCO, although we seem to be making a distinction about "only affecting post-kubelet". It also discusses only enabled the option C integration on bare-metal. I admit I'm confused.

If we do see this configuration in the domain of the MCO, and if we would like this capability to be enabled by default on all clusters, then I'd like to see the problem framed as an extension of the previous enhancement - e.g. considering the option of kubernetes-nmstate as an MCO operand.

hardys:
> I don't think we can justify this simply on the basis that installing an optional operator is "inconvenient and cumbersome"

Agreed we can perhaps be more specific about the user impact here, IIUC one of the reasons is so that secondary network configuration may be provided via NNCP manifests during initial deployment, vs requiring an optional add-on API and post-deploy configuration - @yboaron @celebdor do we have any more examples of the requirements driving this?

> In https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/mco-network-configuration.md I think we acknowledge that this configuration falls under the domain of the MCO, although we seem to be making a distinction about "only affecting post-kubelet". It also discusses only enabled the option C integration on bare-metal. I admit I'm confused.

I believe option C was implemented, as a compromises because MCO is supposed to own all persistent config, but it lacks the features required to enable NIC configuration (particularly [Machine specific config](https://github.com/openshift/machine-config-operator/issues/1720), the ability to apply some NIC config without rebooting the box, some kind of declarative API (e.g not just writing NM keyfiles), ability to rollback a bad config, etc.

The issue then, given that this kind of configuration is very common for CNV/baremetal is that the implemented API isn't available by default in those environments.

> If we do see this configuration in the domain of the MCO, and if we would like this capability to be enabled by default on all clusters, then I'd like to see the problem framed as an extension of the previous enhancement - e.g. considering the option of kubernetes-nmstate as an MCO operand.

Agreed, perhaps framing the problem as "deployed via CVO" is missing the bigger picture, which is that we've got a large API gap for on-premise users - we need a standard way to describe NIC config (for both secondary and controlplane networks, although currently the nmstate/NNCP integration only enables the former)

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,195 @@
...
+Currently, there are multiple methods of installing Kubernetes NMState into a cluster: 
+
+- CNV uses a [custom operator - CNAO](https://github.com/kubevirt/cluster-network-addons-operator#nmstate) that installs Kubernetes NMstate, among other network related items.
+The CNAO is deployed using [Hyperconverged Cluster Operator - HCO](https://github.com/kubevirt/hyperconverged-cluster-operator).
+To deploy Kubernetes NMstate it needs to deploy HCO --> CNAO --> Kubernetes NMstate
+- A [new operator](https://github.com/openshift/kubernetes-nmstate/blob/master/manifests/kubernetes-nmstate-operator.package.yaml) has been developed specifically to install Kubernetes NMstate using OLM.
+
+It is desirable for this to converge.
+
+Adding the Kubernetes NMstate as part of the CVO will allow for convenient and non-cumbersome use, which will also allow for uniformity between the various platforms (CNV, IPI-Baremetal).
```
hardys, line 41: I think we need to be more specific here, and perhaps include some links to historical discussion, e.g:

https://github.com/openshift/enhancements/pull/161#discussion_r441112022 

IIUC the conclusion there was this shouldn't be deployed unconditionally for all platforms, and I suspect we wanted to keep it an optional add-on due to the support status of the underlying nmstate components (which is/will-be no longer the case).

Since then however we have delivered operators which target only the `baremetal` platform e.g [cluster-baremetal-operator](https://github.com/openshift/enhancements/blob/master/enhancements/baremetal/an-slo-for-baremetal.md), so there is an established pattern of how to disable an operator on platforms where it's not relevant (although as mentioned by @russellb that's perhaps not ideal, but it's an option we can consider).

We know this is a core feature (not an optional add-on) for baremetal environments, and potentially also for vSphere and RHV?

So I think we should try to reframe this in terms of the user experience, and what is not currently possible (for example, specify secondary NIC config during the deployment phase), the CVO deploying the operator is really just an implementation detail related to achieving the desired user experience?

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Motivation
+
+### Goals
+
+* Provide a GA Kubernetes NMState Operator delivered in CVO for the use of OpenShift Virtualization at install time.
+* Provide a GA Kubernetes NMState Operator delivered in CVO for the use of OpenShift customers at install time.
```
hardys, line 48: As mentioned above, I'd reword this to something like "Provide NMState APIs by default on $platforms"

We should also reference https://github.com/openshift/enhancements/pull/399 since IIUC the goal here is to provide the APIs enabled via that work by default, on some subset of platforms where they are frequently needed?

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,195 @@
...
+
+* Make persistent networking changes to nodes. That should be handled by Machine Config Operator.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a user of CNV, I want to install Kubernetes-nmstate via the operator directly from the payload.
```
hardys, line 60: This should focus on the user experience I think, e.g describe the step-by-step flow (not in great detail, just an overview) that is needed today, and how it would be improved by having the nmstate integration available by default.

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
markmc: +1 we need to continue that earlier discussion of "this is a core capability on (at least) bare-metal, but not on cloud". This is the core issue.

romfreiman:
Just to verify - how many instances will it run with? Any affect on single node openshift?

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,217 @@
...
+
+As a user of Baremetal IPI I want to provide NodeNetworkConfigurationPolicy resources via manifests at install-time, to simplify
+my deployment workflow and avoid additional post-deploy steps.
+
+I'd like to avoid dependencies on additional registries in disconnected environments and enable core platform functions
+such as network configuration directly from the release payload.
+
+#### Story 2
+
+TODO - can we add a CNV story around simplification and maintenance overhead?
```
qinqon, line 80: As a user of Openshift Virtualization I want to apply a NodeNetworkConfigurationPolicy to create network resources at nodes that will connet to VMs nics and use the already deployed kubernetes-nmstate version from openshift to reduce deployment complexity and maintenance.

cgwalters:
Does nmstate have any support for e.g. wireguard?  That could be useful even in cloud platforms.

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Motivation
+
+In previous discussion it was noted that we may [not want NMState enabled on all platforms](https://github.com/openshift/enhancements/pull/161#discussion_r433303754) and also that [all tech-preview features must require opt-in](https://github.com/openshift/enhancements/pull/161#discussion_r433315534).
+
+Now, the support status of NMState is moving from tech-preview to fully supported, the main consideration is how to conditionally enable this only on platforms where it's likely to be needed.
+
+If we can enable Kubernetes NMstate as part of the CVO payload, we can improve the user experience on platforms where it is required - for example allow for NodeNetworkConfigurationPolicy to be provided via manifests at install-time via openshift-installer.
+
+Additionally, having the Kubernetes NMstate as part of the CVO will allow uniformity between the various platforms (CNV, IPI-Baremetal).
```
derekwaynecarr, line 51: lets not say 'cvo' and replace it with release payload.

phoracek:
> Does nmstate have any support for e.g. wireguard? That could be useful even in cloud platforms.

It does not support it now. We could open an RFE on nmstate to see if it is something they would consider adding.

derekwaynecarr:
Just recording some questions that I think will help frame the enhancement for use in future discussions.

>Justification for why this should run everywhere (on-prem, cloud, edge) in all delivery models (self-managed, hosted).

This would be important to emphasize if we expand this to more than "selected platfoms"

>Is the signal/noise ratio by having this API/controller everywhere worth the potential confusion for an SRE in a particular infra, location, or delivery model when debugging an issue?

I am assuming the API would only be installed on selected platforms as a result of this enhancement.

>What existing SLO could deliver the content and not confuse the user?  e.g. could Cluster Network Operator deliver this content optionally based on its own local config like storage/csi?

This enhancement is asking for the capability to be versioned with OCP (in the release payload), but not enabled on all platforms.  We need to evaluate which second level meta-domain operator makes the most sense to enable this, but I am against adding a net new second level operator.

>What is the resource budget for this API/controller?

To enable it by default, we should understand its impact to the control plane:

- write rate to the API server
- number of clients that write to the resource (is it per node?  is there a single controller co-located with api-server?)

To enable it by default, we should understand its impact to each worker:

- what does it watch?
- per node daemon resource cost (cpu/memory)
- per node daemon priority (is it node critical?)
- per node authority to write to the corresponding resource (think node restriction admission plugin in k8s as example)
- any unique backup/restore semantics introduced by the component

>Its future impact to other topologies (single node or externalized control planes)

For single node
- understanding what namespace hosts the component, and ensuring it fits in the cpuset isolation for mgmt components

For externalized control planes
- is this needed only for the 'first' mgmt cluster and its workers?
- is this needed for every worker in all agnostic clusters, and therefore should be hosted with an externalized control plane because its needed to support any 'join worker' flow.

>is it ok to deliver the content in a release payload, but enable it via some other configuration knob?

What is the knob to turn on/off the capability?  This is related to the SLO chosen to deliver this new meta-operator and what configuration it reads to know to turn it on.  For example, if this is for selected platforms, the component would read the `Infrastructure` CR to determine if it should enable it all, but is there another knob that would allow opt-in/out?

yboaron:
> Just recording some questions that I think will help frame the enhancement for use in future discussions.
> 
> > Justification for why this should run everywhere (on-prem, cloud, edge) in all delivery models (self-managed, hosted).
> 
> This would be important to emphasize if we expand this to more than "selected platfoms"
> 
> > Is the signal/noise ratio by having this API/controller everywhere worth the potential confusion for an SRE in a particular infra, location, or delivery model when debugging an issue?
> 
> I am assuming the API would only be installed on selected platforms as a result of this enhancement.
> 
> > What existing SLO could deliver the content and not confuse the user?  e.g. could Cluster Network Operator deliver this content optionally based on its own local config like storage/csi?
> 
> This enhancement is asking for the capability to be versioned with OCP (in the release payload), but not enabled on all platforms. We need to evaluate which second level meta-domain operator makes the most sense to enable this, but I am against adding a net new second level operator.
> 
> > What is the resource budget for this API/controller?
> 
> To enable it by default, we should understand its impact to the control plane:
> 
> * write rate to the API server
> * number of clients that write to the resource (is it per node?  is there a single controller co-located with api-server?)
> 
> To enable it by default, we should understand its impact to each worker:
> 
> * what does it watch?
> * per node daemon resource cost (cpu/memory)
> * per node daemon priority (is it node critical?)
> * per node authority to write to the corresponding resource (think node restriction admission plugin in k8s as example)
> * any unique backup/restore semantics introduced by the component
> 
> > Its future impact to other topologies (single node or externalized control planes)
> 
> For single node
> 
> * understanding what namespace hosts the component, and ensuring it fits in the cpuset isolation for mgmt components
> 
> For externalized control planes
> 
> * is this needed only for the 'first' mgmt cluster and its workers?
> * is this needed for every worker in all agnostic clusters, and therefore should be hosted with an externalized control plane because its needed to support any 'join worker' flow.
> 
> > is it ok to deliver the content in a release payload, but enable it via some other configuration knob?
> 
> What is the knob to turn on/off the capability? This is related to the SLO chosen to deliver this new meta-operator and what configuration it reads to know to turn it on. For example, if this is for selected platforms, the component would read the `Infrastructure` CR to determine if it should enable it all, but is there another knob that would allow opt-in/out?



> Just recording some questions that I think will help frame the enhancement for use in future discussions.
> 
> > Justification for why this should run everywhere (on-prem, cloud, edge) in all delivery models (self-managed, hosted).
> 
> This would be important to emphasize if we expand this to more than "selected platfoms"
> 
> > Is the signal/noise ratio by having this API/controller everywhere worth the potential confusion for an SRE in a particular infra, location, or delivery model when debugging an issue?
> 
> I am assuming the API would only be installed on selected platforms as a result of this enhancement.
> 
> > What existing SLO could deliver the content and not confuse the user?  e.g. could Cluster Network Operator deliver this content optionally based on its own local config like storage/csi?
> 
> This enhancement is asking for the capability to be versioned with OCP (in the release payload), but not enabled on all platforms. We need to evaluate which second level meta-domain operator makes the most sense to enable this, but I am against adding a net new second level operator.
> 
> > What is the resource budget for this API/controller?
> 
> To enable it by default, we should understand its impact to the control plane:
> 
> * write rate to the API server
> * number of clients that write to the resource (is it per node?  is there a single controller co-located with api-server?)
> 
> To enable it by default, we should understand its impact to each worker:
> 
> * what does it watch?
> * per node daemon resource cost (cpu/memory)
> * per node daemon priority (is it node critical?)
> * per node authority to write to the corresponding resource (think node restriction admission plugin in k8s as example)
> * any unique backup/restore semantics introduced by the component
> 
> > Its future impact to other topologies (single node or externalized control planes)
> 
> For single node
> 
> * understanding what namespace hosts the component, and ensuring it fits in the cpuset isolation for mgmt components
> 
> For externalized control planes
> 
> * is this needed only for the 'first' mgmt cluster and its workers?
> * is this needed for every worker in all agnostic clusters, and therefore should be hosted with an externalized control plane because its needed to support any 'join worker' flow.
> 
> > is it ok to deliver the content in a release payload, but enable it via some other configuration knob?
> 
> What is the knob to turn on/off the capability? This is related to the SLO chosen to deliver this new meta-operator and what configuration it reads to know to turn it on. For example, if this is for selected platforms, the component would read the `Infrastructure` CR to determine if it should enable it all, but is there another knob that would allow opt-in/out?

Thank you for raising these questions, a Q&A section was added to the enhancement.

bengland2:
just a comment - OCS could make use of nmstate for MTU, bonding, a bunch of things that would significantly improve its performance  if this was part of base system.  @obnoxxx

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,279 @@
...
+As a user of Baremetal IPI I want to provide NodeNetworkConfigurationPolicy resources via manifests at install-time, to simplify
+my deployment workflow and avoid additional post-deploy steps.
+
+I'd like to avoid dependencies on additional registries in disconnected environments and enable core platform functions
+such as network configuration directly from the release payload.
+
+#### Story 2
+
+As a user of Openshift Virtualization I want to apply a NodeNetworkConfigurationPolicy to create network resources at nodes that will connet to VMs nics and use the already deployed kubernetes-nmstate version from openshift to reduce deployment complexity and maintenance.
+
```
hardys, line 143: @bengland2 mentioned OCS as a potential user of this, for configuration of MTU/bonding when secondary networks are used for storage traffic, @yboaron can we capture that here as another user-story?

yboaron: Sure,
Thanks

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,284 @@
...
+To improve the user-experience it is desirable to enable the NMState API by default on selected platforms
+where such configuration is common.
+
+## Motivation
+
+In previous discussion it was noted that we may [not want NMState enabled on all platforms](https://github.com/openshift/enhancements/pull/161#discussion_r433303754) and also that [all tech-preview features must require opt-in](https://github.com/openshift/enhancements/pull/161#discussion_r433315534).
+
+Now, the support status of NMState is moving from tech-preview to fully supported, the main consideration is how to conditionally enable this only on platforms where it's likely to be needed.
+
+If we can enable Kubernetes NMstate as part of the release payload, we can improve the user experience on platforms where it is required - for example allow for NodeNetworkConfigurationPolicy to be provided via manifests at install-time via openshift-installer.
```
dhellmann, line 49: In #753 we are addressing a similar need with the performance-addon-operator. In that case, we're exploring the option of running the operator's `render` command outside of the installer between the `create manifests` and `create cluster` phases. Could a similar approach be used with kubernetes-nmstate?

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,284 @@
...
+**Q:** What is the knob to turn on/off the capability? This is related to the SLO chosen to deliver this new meta-operator and what configuration it reads to know to turn it on. For example, if this is for selected platforms, the component would read the Infrastructure CR to determine if it should enable it all, but is there another knob that would allow opt-in/out?
+
+**A:** The current proposal is to couple this to the platform detected in the Infrastructure CR, so it can be enabled by default on selected platforms (particularly baremetal).
+
+There could also be some other interface at the SLO level to override that behavior and/or to enable opt-in for platforms where it’s not enabled by default.
+
+
+### Goals
+
+- Provide NMState APIs by default on desired platforms, [NetworkManager configuration should be updated](https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/mco-network-configuration.md#option-c-design) to support this capability on these platforms.
```
dhellmann, line 120: After reading through this, it's not entirely clear what the "desired platforms" are. Is it just baremetal IPI?

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,284 @@
...
+Additionally, having the Kubernetes NMstate as part of the release payload will allow uniformity between the various platforms (CNV, IPI-Baremetal).
+
+Currently, CNV uses [Hyperconverged Cluster Operator - HCO](https://github.com/kubevirt/hyperconverged-cluster-operator) to deploy [custom operator - CNAO](https://github.com/kubevirt/cluster-network-addons-operator#nmstate), the CNAO installs Kubernetes NMstate, among other network related items.
+While IPI-Baremetal uses a [new operator](https://github.com/openshift/kubernetes-nmstate/blob/master/manifests/kubernetes-nmstate-operator.package.yaml) to install Kubernetes NMstate using OLM.
+
+## Q&A
+
+**Q:** why NMSTATE capability should run everywhere (on-prem, cloud, edge) in all delivery models (self-managed, hosted) ?
+
+**A:** If it was supported/possible, it would have been better to deploy nmstate capability only in required platforms.
```
sferich888, line 60: Can we elaborate more on 'why its not possible'? I find this to be an answer without context.

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Motivation
+
+In previous discussion it was noted that we may [not want NMState enabled on all platforms](https://github.com/openshift/enhancements/pull/161#discussion_r433303754) and also that [all tech-preview features must require opt-in](https://github.com/openshift/enhancements/pull/161#discussion_r433315534).
+
+Now, the support status of NMState is moving from tech-preview to fully supported, the main consideration is how to conditionally enable this only on platforms where it's likely to be needed.
+
+If we can enable Kubernetes NMstate as part of the release payload, we can improve the user experience on platforms where it is required - for example allow for NodeNetworkConfigurationPolicy to be provided via manifests at install-time via openshift-installer.
+
+Additionally, having the Kubernetes NMstate as part of the release payload will allow uniformity between the various platforms (CNV, IPI-Baremetal).
```
sferich888, line 51: Do these two operators run the risk of conflicting with one another because of this approach?

bengland2:
Eric, can  we prioritize baremetal for NMState?  That's the biggest OCS
need that I'm aware of, and involves some large customer accounts.
 cc'ing Eran Tamir, OCS PM.

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
hardys: Yes I think that could work, if the needed containers are in the release payload then even if we require an "advanced" interface at the installer phase it's an improvement on the current process where everything has to happen day-2 via OLM.  @yboaron would you agree?

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
hardys: The initial target is just the baremetal platform, to support both CNV and general baremetal cases, but ref discussion above with @bengland2 there are other potential use-cases e.g OCS which wouldn't necessarily be specific to baremetal?

So I'd like to see this enabled by default for baremetal, with a way to opt-in for other platforms.

The other option that you mention above is to add the components to the release payload, but don't enable by default for any platform, then require some "advanced" customization step e.g manifest to enable (on any platform).

phoracek: For OpenShift Virtualization it would be important too to have a way to opt-in on other platforms.

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
yboaron: Yep, that sounds like a good direction.

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,284 @@
...
+
+Currently, CNV uses [Hyperconverged Cluster Operator - HCO](https://github.com/kubevirt/hyperconverged-cluster-operator) to deploy [custom operator - CNAO](https://github.com/kubevirt/cluster-network-addons-operator#nmstate), the CNAO installs Kubernetes NMstate, among other network related items.
```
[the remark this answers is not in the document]
phoracek: It does. HCO and CNAO are both deployed through OLM. kubernetes-nmstate however is deployed directly by CNAO, without OLM. The reason is that we want a granular control over the lifecycle without a dependency on OLM as the U/S project targets vanilla Kubernetes environments too.

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
```diff
@@ -0,0 +1,284 @@
...
+
+**A:** The current proposal is to couple this to the platform detected in the Infrastructure CR, so it can be enabled by default on selected platforms (particularly baremetal).
+
+There could also be some other interface at the SLO level to override that behavior and/or to enable opt-in for platforms where it’s not enabled by default.
+
+
+### Goals
+
+- Provide NMState APIs by default on desired platforms, [NetworkManager configuration should be updated](https://github.com/openshift/enhancements/blob/master/enhancements/machine-config/mco-network-configuration.md#option-c-design) to support this capability on these platforms.
+- Allow using NMState APIs at install-time, where NIC config is a common requirement.
```
yuqi-zhang, line 121: Might be a dumb question, but will install time configuration be persistent via the MCO? Compared to the ephemeral option-C above that is. Otherwise wouldn't it be lost after the first reboot that happens on all RHCOS nodes?

cybertron: Install-time configuration would be persisted via the kubernetes-nmstate CR. For day 1 we would pull the config out of that and apply it directly to the image, then for day 2 the operator would take over and allow more dynamic changes.

yuqi-zhang: I think I misunderstood the exact boundaries of responsibilities. The MCO is just writing a config setup for the NMstate operator to use? Or is the MCO not involved at all in this nice flow

enhancements/network/kubernetes-nmstate-operator-to-cvo.md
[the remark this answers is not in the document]
sdodson: Yes, lets get a list into the enhancement. Here's my suggestion https://github.com/openshift/enhancements/pull/747#discussion_r661739839

PR #748, opened 2021-04-21
New proposal: Forwarder output parameter templates

alanconway:
Addresses the problem of separating logs from containers in the same pod,
for example application and sidecar containers.
Initial motivation is the JSON/Elasticsearch indexing problem but will be useful everywhere.

/assign @jcantril
/cc @sichvoge
/cc @ewolinetz
/cc @lukas-vlcek


enhancements/cluster-logging/output-parameter-templates.md
```diff
@@ -0,0 +1,153 @@
...
+
+### User Stories
+
+#### Send JSON logs from pods labeled app=myApp to separate indices based on container name
+
+``` yaml
+outputs:
+  - name: myApp
+    type: elasticsearch
+	indexPattern: '{.kubernetes.label.app.myApp}-{.kubernetes.container_name}'
```
jcantrill, line 94: What is the behavior when  the pod does not have label: `app=myapp`?

enhancements/cluster-logging/output-parameter-templates.md
```diff
@@ -0,0 +1,153 @@
...
+For example, to set an index of the form "label_value+container_name":
+
+``` yaml
+  elasticsearch:
+    indexPattern: '{.kubernetes.label.foo}+{.kubernetes.container_name}'
+```
+
+The template pattern is [JSONpath](https://goessner.net/articles/JsonPath/).
+JSONpath is used by [kubernetes](https://kubernetes.io/docs/reference/kubectl/jsonpath)
+and [fluetnd](https://docs.fluentd.org/plugin-helper-overview/api-plugin-helper-record_accessor)
```
jcantrill, line 60: spelling issue

enhancements/cluster-logging/output-parameter-templates.md
```diff
@@ -0,0 +1,153 @@
...
+    indexKey: kubernetes.label.foo
+```
+
+This proposal allows multiple record fields to be _combined_ into an output value.
+
+For example, to set an index of the form "label_value+container_name":
+
+``` yaml
+  elasticsearch:
+    indexPattern: '{.kubernetes.label.foo}+{.kubernetes.container_name}'
```
jcantrill, line 55: This is a dangerous offering for Elasticsearch as it creates a single index per container.  Imagine on a large OCP cluster with short lived build pods, crons, etc,  you will have significant shard explosion if you allow each container to have it's own index.  Additionally, modifying the shard replication leads to the need to potentially have significantly large ES clusters.  Customer's may want to do this not realizing the implications but I would N'ack this for our managed offering

alanconway: +1. It is a bad example. I'll remove and make sure we focus on realistic/useful examples.

enhancements/cluster-logging/output-parameter-templates.md
alanconway: Needs clarifying. Will update.

enhancements/cluster-logging/output-parameter-templates.md
alanconway: +1

enhancements/cluster-logging/output-parameter-templates.md
alanconway: It has the potential for abuse, but container names are not randomized strings like pod names, they tend to name the service in the container, so there will likely be only a small set of container names across the cluster, and a very small set in the application identified by a label. 
I see no other way to solve the problem of logging multi-container pods to ES, since sidecar containres are very likely to have different logging formats from the app container - they're usually created by completely unrelated teams. E.g. the Istio project likely has nothing to do with the local application project.  
Thoughts?

jcantrill: @igor-karpukhin @periklis @lukas-vlcek  Do we have enough metrics and alerts in place to provide enough visibility into the issues I foresee this change introducing?

alanconway: @jcantrill good point. Note this is also relevant to the changes we have already made - if the user picks a label with too many values in the cluster they can also overload ES. This whole user-chosen index thing is inherently prone to excessive ES indices if the user isn't careful.  That's been a concern since the start, but we don't have any better ideas (so far as I know) to prevent a single index from exploding due to incompatible JSON messages.

alanconway: @jcantrill can you lgtm this as an initial proposal? We can add details about metrics if need be later - that applies in any case to what we have already implemented and shipped.

jcantrill:
My general concerns are that in experienced customers will use this feature which will lead to index explosion and turn to us to resolve.  We need to ensure we have guidance and the proper metrics in place. 

/lgtm

PR #751, opened 2021-04-23
WIP - Disable Jenkins Pipeline Build Strategy

adambkaplan:
This is a draft proposal to disable the Jenkins Pipeline build strategy.
With this build strategy disabled, the Jenkins image can be removed from
the OpenShift payload and be delivered via CPaaS.
Jenkins fixes can then be released on their own cadence, independent of
the OpenShift release lifecycle.


adambkaplan:
/cc @akram @gabemontero @bparees @soltysh 

Brainstorming all of the touch points where Jenkins pipeline strategy could be invoked:

1. Webhooks
2. Web console
3. oc command line

Is there anything I am missing?

sbose78:
This is a step in the right direction, thanks for putting it together @adambkaplan .

Would you want to clarify that users still have the option of using Jenkinsfiles directly with their instance of Jenkins, instead of the BuildConfig pipeline strategy approach ?

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,249 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+OpenShift 3.0 included a way to invoke a Jenkins pipeline from an OpenShift build.
+The status of the build on the OpenShift cluster would reflect the state of the pipeline in Jenkins.
```
gabemontero, line 36: technically speaking, the jenkins pipeline status is stored as encoded json in annotations in the build object

see https://github.com/openshift/jenkins-sync-plugin/blob/master/src/main/java/io/fabric8/jenkins/openshiftsync/BuildSyncRunListener.java

the console then unpacks that json to provide the visualization it does for pipeline strategy

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,249 @@
...
+We now have that meaningful replacement with the general availability of OpenShift Pipelines.
+It is time to disable OpenShift's direct integration with Jenkins.
+
+## Motivation
+
+Distrbuting and supporting Jenkins directly has always been challenging for OpenShift.
+Jenkins is based in Java and is extended by a convoluted ecosystem of plugins, which are a constant source of bugs and CVEs.
+Historically, the sprawling nature of Jenkins and its dependencies has made it impossible for Red Hat to independently compile Jenkins and the libraries used by OpenShift's plugin extensions.
+We instead distribute Jenkins directly from community binaries, making it vulnerable to supply chain attacks from sophisticated attackers.
+Given the increasing prevalence of supply chain attacks globally (SolarWinds), relying on third parties for binary code is becoming less tolerable for product security.
```
gabemontero, line 51: Historical reference - Openshift has had this able to build from jenkins binary exception since Openshift 2

yes that is 2, not 3 nor 4

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,249 @@
...
+
+As an OpenShift release engineer
+I want to remove Jenkins from the OCP payload
+So that I can reduce the size of the payload
+And release Jenkins fixes on their own cadence
+
+### Implementation Details/Notes/Constraints [optional]
+
+Deprecation warnings for the Jenkins Pipeline Build Strategy will need to be updated in the next OpenShift release (OCP 4.N).
+Since Jenkins integrations have been officially been officially deprecated for well over one year, these warnings need to be strengthened so customers are fully aware that all direct Jenkins integrations will be removed.
```
gabemontero, line 96: you've got a double `been officially` in this line

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,249 @@
...
+Like the use of other deprecated APIs, OpenShift must also alert users if a Jenkins pipeline build strategy is used in a Build or BuildConfig.
+
+In the OCP 4.(N+1) release, builds which use the Jenkins pipeline build strategy should fail immediately.
+All documentation related to Jenkins pipeline builds will need to be removed, and replaced with a statement that the Jenkins pipeline strategy has been disabled.
+The updated documenation should include instructions on how to move a Jenkinsfile off of OpenShift.
```
gabemontero, line 101: the use of `curl` POST/Jenkins REST API and maybe the Jenkins CLI (though I think it may have fallen from favor as well) are both alternatives to comman line `oc start-build`

gabemontero: either a mention here and/or in the migration/alternatives section works for me  ... where ever you deem fit @adambkaplan 

but the fact command line invocation of jenkins pipelines is available outside of `oc start-build` seems important to me

gabemontero (changes requested):
some initial feedback @adambkaplan 

thanks for throwing this together

enhancements/builds/disable-jenkins-pipeline.md
adambkaplan: @spadgett was this console view an OCP 3-ism? Did we get rid of the pipeline view for Jenkins in OCP 4?

sbose78: I think, yes, in probably because I don't recollect seeing it anywhere.

sbose78:
> We do this today in our docs, worth pointing out in the EP.

Right, would be nice to point out in the EP that there's an alternative. Therefore, there's zero loss of functionality.

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,249 @@
...
+So that I can reduce the size of the payload
+And release Jenkins fixes on their own cadence
+
+### Implementation Details/Notes/Constraints [optional]
+
+Deprecation warnings for the Jenkins Pipeline Build Strategy will need to be updated in the next OpenShift release (OCP 4.N).
+Since Jenkins integrations have been officially been officially deprecated for well over one year, these warnings need to be strengthened so customers are fully aware that all direct Jenkins integrations will be removed.
+Like the use of other deprecated APIs, OpenShift must also alert users if a Jenkins pipeline build strategy is used in a Build or BuildConfig.
+
+In the OCP 4.(N+1) release, builds which use the Jenkins pipeline build strategy should fail immediately.
```
bparees, line 99: i don't think we have ever removed/disabled a v1 GA api.  i'm not sure who needs to sign off on this, but i think we'll need architecture team buy-in on whether this is allowed and if it is, what the process/mechanics need to be.

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,249 @@
...
+  or CNI may require updating that component before the kubelet.
+
+## Implementation History
+
+2020-04-23: Provisional enhancement draft
+
+## Drawbacks
+
+The idea is to find the best form of an argument why this enhancement should _not_ be implemented.
+
```
siamaksade, line 236: Customers who use Jenkins on OpenShift heavily rely on the pipeline strategy in Build
Configs to create their Jenkins jobs. The pipeline strategy has been deprecated for a while to signal customers that they have to move to an alternative approach for declaratively creating Jenkins jobs based on their Jenkinsfiles. Strong pushback and frequent escalations from customers should be expected if this feature is removed. 

Jenkins Operator initiative has been started to help customers with the transition when the pipeline strategy gets removed. However the Jenkins Operator on OpenShift is still in Tech Preview and customers cannot rely on it as an alternative.

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,249 @@
...
+Historically, the sprawling nature of Jenkins and its dependencies has made it impossible for Red Hat to independently compile Jenkins and the libraries used by OpenShift's plugin extensions.
+We instead distribute Jenkins directly from community binaries, making it vulnerable to supply chain attacks from sophisticated attackers.
+Given the increasing prevalence of supply chain attacks globally (SolarWinds), relying on third parties for binary code is becoming less tolerable for product security. 
+
+Jenkins itself is not native to Kubernetes and does not use a Java runtime that has been optimized for cloud-native environments (i.e. Quarkus).
+As a result, it has poor resilience when run on Kubernetes.
+The Jenkins templates do not run Jenkins in a highly available mode, leaving its primary instance vulnerable to downtime from slow restarts.
+The Jenkins container image is also very large - it is in fact the largest image in the OpenShift payload today.
+Removing Jenkins from the OpenShift payload has been a longstanding goal that can reduce the time needed to pull the OCP payload at install/updgrade time.
+
```
siamaksade, line 58: Why does removing Jenkins from payload requires removing the pipeline strategy from builds?

bparees: strictly speaking, it does not, however, if we remove it from the payload but keep the strategy:

1) customers will need to setup their own jenkins imagestream pointing to an appropriate image before their pipelines will work
2) as long as we keep the pipeline strategy alive, we need to keep the jenkins sync plugin up to date in an available jenkins image (sync plugin is necessary for the pipeline strategy to work)

(for the record i agree with your concerns about disabling the strategy, we need to do some due diligence here to figure out who is still using the strategy and what the impact would be on them if we disabled it)

enhancements/builds/disable-jenkins-pipeline.md
spadgett: @adambkaplan We added it in 4.x as well for parity. It looks like it's still there.

https://github.com/openshift/console/blob/master/frontend/public/components/build-pipeline.tsx

enhancements/builds/disable-jenkins-pipeline.md
gabemontero: part of moving it out of the payload to a new cpaas based image hosted on registry.redhat.io means you could still ship a samples jenkins imagestream and template, but remove the samples operator function that replaces the current image ref with the payload image.

in that sense then, jenkins would be just like all our other samples from SCL and middleware 

no need for users to set up the imagestream / template on their own

those templates and imagestreams still go through the process of being hosted in their "upstream" repo (in the jenkins case, https://github.com/openshift/jenkins) and then get pulled into https://github.com/openshift/library

bparees: > part of moving it out of the payload to a new cpaas based image hosted on registry.redhat.io means you could still ship a samples jenkins imagestream and template, but remove the samples operator function that replaces the current image ref with the payload image.

that's fair, though it doesn't work as cleanly for disconnected installs. and it loses the tight coupling between the jenkins image version and the ocp version (possibly not that critical any more if things have stabilized)

adambkaplan:
/hold

I don't think we can safely disable this strategy until we have good data on how many clusters are using the Jenkins pipeline strategy. Unfortunately to get data from telemetry, we need to add a recording rule to the cluster monitoring operator.

PR to add the recording rule is here: https://github.com/openshift/cluster-monitoring-operator/pull/1136. It will take some time for that to be backported to 4.6 (and have those 4.6 clusters upgrade) so we can get reliable numbers across the fleet.

enhancements/builds/disable-jenkins-pipeline.md
akram: True @siamaksade , the pipeline strategy is a convenient way to trigger the job creation in Jenkins and a way to provide customers a command line way to start the Jenkins build using `oc` . That's what the sync plugin does for this part.

It has probably not been explained explicitly what is the alternative with the deprecation. Here are the steps actually:
- User needs to create in Jenkins a new pipeline Job, with "Pipeline script definition from SCM" and set the correct git URL.
- User need to clic the "Start build" button in Jenkins console every time he wants to start a build.
- Kubernetes plugin takes care of the rest (starting agent, building, etc...)


It means that:
- `oc new-build` with pipeline strategy will say `Not supported`, it will not create a `BuildConfig` and `oc start-build` will not operate, as there is not `BuildConfig` of that type.
- The console will not reflect the pipeline state, as the `BuildConfig` does not exist.
- User who wants to follow the pipeline progress and status needs to reach Jenkins console.
- For clusters, with existing `BuildConfig` of pipeline strategy, `oc start-build` needs to inform customer that it is not supported anymore and that user has to start build manually from Jenkins console or use [jenkins-cli](https://www.jenkins.io/doc/book/managing/cli/ ) for instance.
- Builds started from Jenkins will still use agents as the agents lifecycle is managed by the Kubernetes plugin that we embed and pre-configure in our Jenkins image.

Talking about the Jenkins Operator to cover this area, the only added value would be for the Jenkins Operator to watch `BuildConfig` and `Build` objects for which strategy is `Pipeline` and then trigger the Jenkins build, which is actually, what we want to deprecate. And, having this feature has not been discussed until now for the Jenkins Operator as it was supposed to be a `kubernetes` compliant operator.

enhancements/builds/disable-jenkins-pipeline.md
soltysh: We never removed anything, at most we've made some fields no-ops. I'd start with @smarterclayton @derekwaynecarr to confirm. 
Most importantly do we have any numbers behind this, how many customers are actually using this? How painful

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,257 @@
...
+In OCP 4.N, an alert should be fired if a build is started with the Jenkins Pipeline build strategy.
+Likewise, an alert should be fired if there is any BuildConfig which uses the Jenkins Pipeline strategy.
+This data is already collected by openshift-state-metrics, therefore creating an alert is a matter of aggregating the right metrics.
+
+In OCP 4.(N+1), the build controller should immediately fail a build which utilizes the Jenkins pipeline strategy with an appropriate failure message.
+
+#### oc <new|start>-build
+
+In OCP 4.N, `oc new-build` should print a warning message if the generated/referenced BuildConfig uses the Jenkins pipeline strategy.
```
soltysh, line 113: I'm inclined to print the warning asap (ie. 4.8), since this functionality is already deprecated.

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,257 @@
...
+`oc start-build` should print a warning message if a build was started with the Jenkins pipeline strategy.
+
+In OCP 4.(N+1), `oc new-build` should not support the creation of new Builds with the Jenkins pipeline strategy.
+`oc start-build` may continue to create builds referencing the Jenkins pipeline strategy, but these builds will ultimately fail.
+
+#### Web Console
+
+User will continue to be able to deploy Jenkins on OpenShift by instantiating our provided Templates.
+
+In OCP 4.N, the Jenkins pipeline view of a build needs to display a warning that the strategy will be removed.
```
soltysh, line 123: Similarly to oc, this should happen asap, if this functionality is already deprecated.

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,257 @@
...
+## Alternatives
+
+### Only remove Jenkins from the OCP Payload
+
+OCP 3.x initially had a feature that auto-provisioned Jenkins if a JenkinsPipeline build was started.
+This capability was removed in the later OCP 3 releases, but its legacy was one reason that Jenkins was included in the OCP payload.
+
+Today users must deploy Jenkins via the provided templates to take advantage of OpenShift's integrations.
+The Samples Operator is responsible for installing the imagestream and templates for Jenkins, just like other Red Hat products delivered in this fashion.
+Jenkins is the only imagestream installed via the Samples Operator whose images exist in the payload.
```
soltysh, line 245: Maybe this is a good way forward initially. Get rid of the image, align the usage with the rest of samples. In parallel proceed with deprecation warnings everywhere with suggestions to use jenkins operator instead. 
A followup, would be to entirely drop the support once we have a viable replacement and users get sufficient time to migrate off the built-ins.

enhancements/builds/disable-jenkins-pipeline.md
siamaksade: In a developer survey we did last year with customers, >40% of respondents were using Jenkins on OpenShift and every customer we know that uses Jenkins on OpenShift uses the pipeline build strategy. The impact would be high.

enhancements/builds/disable-jenkins-pipeline.md
siamaksade: @akram the process you described is not declarative. It cannot be stored in a Git repository which is what customers are doing with the BuildConfig pipeline strategy and recreating all those jobs on a new cluster with a single `oc create -f jobs` command. 

This is the main point that impact customers heavily. They simply do not have a declarative way to create Jenkins pipelines anymore and have to create them one by one through the Jenkins UI every single time.

enhancements/builds/disable-jenkins-pipeline.md
soltysh: Yeah, that I'd probably start simple with what @adambkaplan writes towards the end of this document, to just remove the jenkins image from the payload, and align that image with the rest of samples operator. But leaving the strategy as is. This will have some downsides for users, which for disconnected environments would need additional step during mirroring, but I think that would send a clear singal for future chanegs.

enhancements/builds/disable-jenkins-pipeline.md
siamaksade: I'm much more comfortable with this approach despite the drawbacks that @bparees mentioned. I see it as an acceptable loss compared to removing the strategy altogether at this point.

enhancements/builds/disable-jenkins-pipeline.md
siamaksade: +1

enhancements/builds/disable-jenkins-pipeline.md
akram: What about using the `jenkins-cli` as suggested. It gets `xml` as an input containing the declaration of the job. That's not `oc` for sure, but, it seems to be a reasonnable tradeoff for the price of the deprecation of `Pipeline` build strategy.

sbose78:
Checking back on this after a week, I am OK with where this is heading : 

* Add deprecation message
* Move Jenkins image out of the payload
* Understand usage patterns https://github.com/openshift/cluster-monitoring-operator/pull/1136/files before dropping capability. 

Thanks @adambkaplan for this EP, helped have valuable conversations.

enhancements/builds/disable-jenkins-pipeline.md
adambkaplan: Proposal has been updated so that we aren't disabling the strategy. Instead we are introducing a feature gate/feature set.

adambkaplan:
/hold cancel

This proposal has been revamped such that the Jenkins pipeline strategy is enabled via a feature gate, rather than enabled by default. It is an attempt to strike a middle ground between discouraging Jenkins and maintaining current capabilities.

I think we are all in agreement that we can take Jenkins out of the payload once a CPaaS pipeline is established. If folks would like to see this as a separate EP I can split it out, or alternatively we can create a public JIRA epic and call it a day.

@sbose78 @gabemontero @akram @bparees @derekwaynecarr ptal

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,303 @@
...
+Clusters that enable the LegacyJenkinsPipeline feature set will be automatically reported to Telemeter.
+
+#### Documentation
+
+After Jenkins is removed from the payload, documentation related to Jenkins images will need to add mirroring instructions for disconnected clusters.
+Release notes will also need to inform admins that Jenkins images will need to be mirrored separately from the OpenShift payload.
+
+Documentation related to the Jenkins Pipeline build strategy will need to inform admins that the `LegacyJenkinsPipeline` feature set needs to be enabled for builds to succeed.
+
+### Risks and Mitigations
```
bparees, line 181: i'd like to see more detail around the imagestream transition from "imagestream that refs payload content" to "imagestream that refs external content"

For a connected cluster, this isn't a big deal and should be basically transparent when the cluster upgrades.

For a disconnected cluster it means:
1) on install, the imagestream won't import (if the image wasn't mirrored, which it won't be if they just mirrored the payload).  That's ok, we can document/talk through that
2) On upgrade what happens?  the imagestream gets updated and now points to some external registry and the imagestream can't be imported....  does it keep limping along on the last imported SHA (the one from the previous OCP release payload)?  And how bad a problem is that if that's what happens?  This is the scenario i'd be most worried about.

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,303 @@
...
+OpenShift 3.0 included a way to invoke a Jenkins pipeline from an OpenShift build.
+The status of the build on the OpenShift cluster would reflect the state of the pipeline in Jenkins.
+This was the first feature to support continuous integration/delivery processes on OpenShift.
+
+As OpenShift and Kubernetes have evolved, so has cloud-native CI/CD.
+Tekton and its downstream distribution - OpenShift Pipelines - provide a way of running CI/CD processes natively on Kubernetes.
+The Jenkins pipeline strategy was officially deprecated in OpenShift 4.1, but at that time there was no meaningful replacement.
+We now have that meaningful replacement with the general availability of OpenShift Pipelines.
+
+To discourage use of the Jenkins pipeline strategy, this proposal will remove Jenkins from the OCP payload and disable runs of Jenkins Pipeline builds by default.
```
bparees, line 42: i'm not sure these two things need to be coupled together.  Is there any value in doing one first and then the other later?
(e.g. putting pipelines behind a featuregate is easy and low risk/impact to customers.  Removing the image from the payload is somewhat higher risk, particularly for disconnected clusters)

in any case removing the payload content doesn't really discourage use of the strategy.  I'd say it's more the case that "as we discourage its use, we feel more empowered to remove the content from the payload"  (hence my suggestion that we gate pipelines first, and then remove the payload content later)

it doesn't have to be done separately, just throwing it out there as a risk mitigation option.

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,303 @@
...
+
+Similar to the `Drawbacks` section the `Alternatives` section is used to
+highlight and record other possible approaches to delivering the value proposed
+by an enhancement.
+
+### Completely disable the Jenkins Pipeline build strategy
+
+The first draft of this proposal sought to disable the Jenkins Pipeline build strategy in its entirety.
+Not only would we disable the Jenkins pipeline strategy, we would also remove Jenkins from the OCP payload and remove a lot of functionality from the Jenkins sync plugin.
+This proved untenable in the near term because we did not have sufficient telemetry to determine the full impact of this revocation.
```
bparees, line 285: since pipelines are already deprecated, i'm a little unclear on the value of feature-gating them but still letting people turn them on..... I guess the idea is that it puts one more barrier in front of people who are looking to use them, where we can make them aware they are opting into using a deprecated feature?

if so, we can ensure the feature gate makes it clear they are enabling a deprecated function?

I assume removal of the feature entirely is still the long term goal, that should probably be stated somewhere along w/ some indication of how we are going to decide when we can do that (e.g. telemetry)

enhancements/builds/disable-jenkins-pipeline.md
adambkaplan: Split this out into the new proposal - #751.

On upgrade the status will reflect the last imported SHA until the new image is mirrored. Eventually the Samples operator will fire an alert indicating that the Jenkins imagestream(s) have failed to import.

enhancements/builds/disable-jenkins-pipeline.md
adambkaplan: Split removal from the payload into its own EP - #841

enhancements/builds/disable-jenkins-pipeline.md
adambkaplan: > if so, we can ensure the feature gate makes it clear they are enabling a deprecated function?

The name I proposed for the feature set is `LegacyJenkinsPipeline` - perhaps using `DeprecatedJenkinsPipeline` would be a stronger name?

bparees: yes that name seems better/clearer

enhancements/builds/disable-jenkins-pipeline.md
adambkaplan: Removing Jenkins from the payload is being considered in a separate proposal - #841

adambkaplan:
@sbose78 @bparees did a dive into telemetry today, I feel confident we can move forward down this path. Sample set was ~800 4.7 clusters which run builds.

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,296 @@
...
+The bootstrap behavior will ensure that most clusters which rely on the JenkinsPipeline build strategy have the feature set enabled automatically.
+The exception to this rule will be clusters that have a separate feature set enabled prior to upgrade.
+The following query in Data Hub will reveal which clusters fall into this category:
+
+`count by (_id, name) (cluster_feature_set{name!=""} offset 180m) * on (_id) group_right() openshift:build_by_strategy:sum{strategy="jenkinspipeline"} offset 180m`
+
+As of this writing, no clusters fall into this category.
+This may change if clusters opt into tech preview features related to OpenShift builds.
+
+#### Documentation
```
gabemontero, line 158: Either with this documentation section or a separate section, I think we should be getting rid of https://github.com/openshift/origin/tree/master/examples/jenkins as part of the overall theme carried here on the de-prioritization and decoupling of Jenkins.

gabemontero (changes requested):
Just one minor ask @adambkaplan 

Also, curious how much consensus there is on this amongst you and the other reviewers.

enhancements/builds/disable-jenkins-pipeline.md
adambkaplan: We will need to keep Jenkins in our product documentation, so I don't want to confuse the matter here.

IMO the Jenkins example in origin can be moved to the openshift/jenkins repo, which is in line with our efforts to move Jenkins tests out of openshift/orign. I don't think we need to get sign off on this via an EP.

gabemontero: Moving from openshift/origin to openshift/jenkins is better than just removing.  I think we should go with your tweak on my idea.

That said, I'm not sure I agree with it not having to be called out in the EP.  Certain elements of the openshfit community still hold value in what is under the examples subdir of openshift/origin.

But perhaps I am in the minority there.

@bparees - you have as much historical participation in the examples subdir as I do.  WDYT?

bparees: > That said, I'm not sure I agree with it not having to be called out in the EP. Certain elements of the openshfit community still hold value in what is under the examples subdir of openshift/origin.

do they? we should be directing them to https://github.com/openshift/library/ instead.

the history of why they are there now is that we only had one repo, and also even once we started having multiple repos, tests were still expecting to get content from the origin repo.

i don't see a good reason to keep the content in origin at this point, as long as we put it in openshift/library (and openshift/library can import it from openshift/jenkins).  You get the templates as part of your OCP install anyway from the samples operator.

gabemontero: That "implementation" approach works for me as well @bparees 

Now, where are you with calling all of this out in @adambkaplan 's EP?  An implementation that should be included?  Or details he does not have to add ?

bparees: given that we've now had the conversation and defined a plan, it seems worth preserving it in the EP itself

adambkaplan: We already sync Jenkins imagestreams and templates from openshift/jenkins to openshift/library.

Updating the docs section to state that tests should use one of the templates delivered via the Samples operator.

gabemontero: FWIW my tech debt moving of e2e's from openshift origin to the jenkins related repos will achieve the result of no longer referencing https://github.com/openshift/origin/tree/master/examples/jenkins in our jenkins related e2e's

adambkaplan:
/cc @gabemontero @akram 

Rebased this, so it lost the lgtm label

akram:
it needs the approved label @adambkaplan , only you in our team seems to have permission to set it.

bparees:
> it needs the approved label @adambkaplan , only you in our team seems to have permission to set it.

i should be able to do it, i plan to take a pass through this and https://github.com/openshift/enhancements/pull/841 today.

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,299 @@
...
+Tekton and its downstream distribution - OpenShift Pipelines - provide a way of running CI/CD processes natively on Kubernetes.
+The Jenkins pipeline strategy was officially deprecated in OpenShift 4.1, but at that time there was no meaningful replacement.
+We now have that meaningful replacement with the general availability of OpenShift Pipelines.
+
+To discourage use of the Jenkins pipeline strategy, this proposal will disable runs of Jenkins Pipeline builds by default.
+Clusters which rely on the Jenkins pipeline strategy can enable it through cluster feature sets.
+
+## Motivation
+
+Distrbuting and supporting Jenkins directly has always been challenging for OpenShift.
```
bparees, line 47:
```suggestion
Distributing and supporting Jenkins directly has always been challenging for OpenShift.
```

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,299 @@
...
+
+#### Bootstrap Behavior
+
+When the `BuildJenkinsPipelineStrategy` feature gate is introduced, the openshift-controller-manager-operator will query all `BuildConfig` objects to see if any utilize the JenkinsPipeline strategy.
+If any BuildConfig uses the JenkinsPipline strategy, the DeprecatedJenkinsPipeline feature set will be turned on.
+
+The bootstrap check will only run if the default feature set is configured for the cluster.
+A status condition will indicate if this bootstrap check has completed.
+Once the bootstrap check is complete, it should not be run again.
+These protections will ensure that we do not overwrite feature sets that are enabled as a day 2 action, or overwrite pre-existing feature sets that allow updgrades.
```
bparees, line 130: when is this bootstrap check performed?  First startup after install, and first startup after upgrade to a version of OCP that has this functionality?

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,299 @@
...
+The operator for openshift-controller-manager will be enhanced to watch the cluster feature gates.
+If the cluster has the `BuildJenkinsPipelineStrategy` feature gate enabled, this gate will be passed to openshift-controller-manager's configuration.
+Openshift-controller-manager will then enable the strategy and allow JenkinsPipline builds to proceed as normal.
+
+#### Bootstrap Behavior
+
+When the `BuildJenkinsPipelineStrategy` feature gate is introduced, the openshift-controller-manager-operator will query all `BuildConfig` objects to see if any utilize the JenkinsPipeline strategy.
+If any BuildConfig uses the JenkinsPipline strategy, the DeprecatedJenkinsPipeline feature set will be turned on.
+
+The bootstrap check will only run if the default feature set is configured for the cluster.
```
bparees, line 127: why only if the default feature set is configured?

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,299 @@
...
+
+**Risk**: Bootstrap logic can overwrite other enabled features.
+
+*Mitigation*: Bootstrap logic is only invoked if the cluster uses the default feature set.
+
+**Risk:** Clusters need tech preview features AND the Jenkins pipeline build strategy enabled.
+
+*Mitigation*: Clusters with tech preview features are automatically marked as `NoUpgrade` clusters.
+If necessary, a cluster that needs tech preview features and the Jenkins pipeline build strategy can use the `CustomNoUpgrade` feature set,
+which gives admins the ability to enable individual feature gates on their own.
```
bparees, line 179: CustomNoUpgrade is really really hard to get right (I had a long discussion with @deads2k about this).

Can't they just enable both the TechPreviewNoUpgrade featureset *and* the DeprecatedJenkins featureset?

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,299 @@
...
+
+*Mitigation*: There is no real mitigation for this other than using the `CustomNoUpgrade` feature set.
+The downside of this approach is that the cluster cannot be upgraded to the next minor release.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. How many clusters use Jenkins pipeline builds and also enable a supported feature gate?
+   None today, but that may change if other build related features require a feature gate (example - tech preview items).
```
bparees, line 191: shouldn't featuresets/featuregates be additive?

my understanding is:
1) if you have no featureset enabled, you get the default features
2) if you enable techpreview you get the default features + tech preview ones
3) if you enable custom, you get exactly what you define in custom (you do not automatically get the default ones)
4) if you enable DeprecatedJenkins you get default+jenkins

I would expect if you enable multiple featuregates (other than Custom), you would get the union of all the features, no?  So it would only be problematic if some featureset(other than custom) needed to block some of the default features, and you wanted to use it w/ another featureset.


cc @deads2k

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,299 @@
...
+This proposal does not permanently disable the strategy - rather it discourages its use by requiring admins to enable a feature set.
+This will be documented in all places where Jenkins is referenced.
+
+Eventually we would like to make this feature set a `NoUpgrade` feature set.
+Doing so would keep the Jenkins pipeline strategy enabled, but prevent the cluster from being upgraded to the next OCP minor version.
+Hypothetically this is possible:
+
+* In OCP 4.N the feature set is renamed from `DeprecatedJenkinsPipeline` to `DeprecatedJenkinsPipelineNoUpgrade`
+* openshift-controller-manager-operator for OCP 4.N automatically renames the configured feature set from `DeprecatedJenkinsPipeline` to `DeprecatedJenkinsPipelineNoUpgrade`.
+* Once the upgrade to 4.N is complete, the `*NoUpgrade` feature set automatically marks the cluster as not upgradeable.
```
bparees, line 226: i don't think the idea of turning someone's prod cluster into a non-upgradeable cluster w/o them realizing that was going to happen.

I think there might be a better way to do this...... in 4.8/4.9 we are introducing the ability to define generic upgrade-gates that require admin acknowledgement.

If/when we have a 4.N that is going to remove jenkins, then in 4.N-1 we can add a required admin-ack gate that forces admins to acknowledge that the jenkins apis are being removed (and we can potentially only require that ack if there are buildconfigs that use jenkins).

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,299 @@
...
+
+Complete removal would depend on the number and nature of clusters which enable this feature set.
+This would require sign off from product management and relevant business units.
+
+### Upgrade / Downgrade Strategy
+
+On upgrade, previously enabled feature sets will remain enabled.
+The bootstrap logic will ensure that clusters which rely on Jenkins pipeline builds continue to function, unless that cluster has enabled a separate feature set.
+
+On downgrade, clusters with the DeprecatedJenkinsPipeline feature set will continue to function, but will not be upgradeable because the feature set is not known.
```
bparees, line 237: is that an existing upgrade gate?  that a cluster can't be upgraded if an unknown featureset is configured?

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,299 @@
...
+Admins will need to manually reset clusters to use the default feature set.
```
bparees, line 238: does that mean they'll also need to reset the bootstrap check before they upgrade, so that the bootstrap logic can run again and re-assert the featureset if they need it, on upgrade?

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,299 @@
...
+
+2021-04-23: Provisional enhancement draft
+2021-07-16: Update enhancement to disable via feature gates
+2021-07-20: Split out removal of Jenkins from the playload.
+
+## Drawbacks
+
+This proposal aims to disable behavior that previously functions by default.
+Jenkins remains a critical part of many customer development processes - at one point Jenkins was the most widely downloaded container image from the Red Hat container catalog.
+Requriring clusters to enable a feature set could spark confusion and concern amongst end customers.
```
bparees, line 255:
```suggestion
Requiring clusters to enable a feature set could spark confusion and concern amongst end customers.
```

enhancements/builds/disable-jenkins-pipeline.md
```diff
@@ -0,0 +1,299 @@
...
+
+Using a single supported feature set means that customers cannot enable other supported feature sets.
```
bparees, line 257: this sounds like a weird limitation of featuresets, @deads2k is it deliberate?

bparees (commented):
I've got some concerns around the fact that we can't enable more than one featureset, i'm interested to hear from @deads2k 

1) confirming that, and explaining why it's necessary behavior (it feels very limiting in our ability to use them)
2) whether that means featuresets are the right solution for this particular problem (maybe we need to use an alternate gating mechanism that isn't mutually exclusive with other configurations an admin may desire on their cluster)

bparees:
oh, one more comment:  i'm curious if you've reviewed https://docs.openshift.com/container-platform/4.8/rest_api/understanding-api-support-tiers.html#api-deprecation-policy_understanding-api-tiers and how this particular deprecation (And eventual removal) fits into what is described there?

it's somewhat complicated since our api commitments are generally written around resource types, not "specific fields within a resource type"

so i think it would be a worthwhile exercise for you to read our guidance there and see if it covers what you are trying to do, and if not, help us fill in the gaps.

PR #752, opened 2021-04-26
hack: fix the template lint job

dhellmann:
Placing the while loop for checking headers into a pipe meant the
variable it set with a return code was in a subshell, which did not
affect the return code of the parent. This version collects all of the
information about missing headers and if there are any it prints the
list and then sets the parent script return code correctly.

/assign @wking


wking (approved):
/lgtm

PR #753, opened 2021-04-26
workload-partitioning: add plan for configuration and enablement ownership

dhellmann:
The initial iteration of workload partitioning focused on a short path
 to a minimum viable implementation. This enhancement describes the
 loose ends for preparing the feature for GA at a high level, and
 explains the set of other design documents that need to be written
 separately during the next iteration.

### DIFF
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -134,7 +134,8 @@ have it generate manifests for the OpenShift installer to consume.
 2. Write an enhancement to describe the changes in
    performance-addon-operator to manage the kubelet and CRI-O
    configuration when workload partitioning is enabled, including the
-   `render` command.
+   `render` command. See https://issues.redhat.com/browse/CNF-2164 to
+   track that work.
 3. Write an enhancement to describe an API for enabling workload
    partitioning in an existing cluster.
 
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -145,7 +145,8 @@ deployment, and to simplify the configuration, we propose to extend
 the PAO with a `render` command, similar to the one in other
 operators, to have it generate manifests for the OpenShift installer
 to consume. This avoids any need to change the OpenShift installer to
-support this uncommon use case.
+support this uncommon use case, and is consistent with the way
+third-party networking solutions are installed.
 
 The `render` command will have an option to enable the management
 partition. When the option is given, the `render` command will create
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -82,6 +82,7 @@ the feature and to extend support to multi-node clusters.
    running cluster. We are going to continue to require the user to
    enable partitioning as part of deploying the cluster for another
    implementation phase.
+3. Support for external control plane topologies.
 
 ## Proposal
 
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -240,7 +240,8 @@ to partition the workloads into specific CPUs.
    in the cluster.
 8. Bootstrapping finishes and the cluster is launched.
 9. If the MCO does not apply KubeletConfig in step 7, it must do it
-   here.
+   here after the MCO is running in the cluster and can reconcile
+   normally.
 11. The cluster has partial partitioning enabled for `management`
     workloads. Pods for management workloads are mutated, but not
     actually partitioned.
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -159,10 +159,10 @@ input.
 
 For each PerformanceProfile, the PAO will render MachineConfig
 manifests with higher precedence than the default ones created by the
-installer during bootstrapping. These manifests will include all of
-the details for configuring `kubelet`, CRI-O, and `tuned` to know
-about and partition workloads, including the cpusets for each workload
-type.
+installer during bootstrapping, in the same way it does when running
+in the cluster. These manifests will include all of the details for
+configuring `kubelet`, CRI-O, and `tuned` to know about and partition
+workloads, including the cpusets for each workload type.
 
 The PAO `render` command also generates a manifest with the CRD for
 PerformanceProfile so bootstrapping does not block when trying to
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -1,0 +1,212 @@
+---
+title: rationalizing-configuration
+authors:
+  - "@dhellmann"
+  - "@mrunalp"
+  - "@browsell"
+  - "@MarSik"
+  - "@fromanirh"
+reviewers:
+  - "@deads2k"
+  - "@sttts"
+  - maintainers of PAO
+approvers:
+  - "@markmc"
+  - "@derekwaynecarr"
+creation-date: 2021-04-26
+status: implementable
+see-also:
+  - "/enhancements/management-workload-partitioning.md"
+---
+
+# Rationalizing Configuration of Workload Partitioning
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The initial iteration of workload partitioning focused on a short path
+to a minimum viable implementation. This enhancement describes the
+loose ends for preparing the feature for GA at a high level, and
+explains the set of other design documents that need to be written
+separately during the next iteration.
+
+## Motivation
+
+The management workload partitioning feature introduced in OCP 4.8
+enables pod workloads to be pinned to the kubelet reserved cpuset. The
+feature is enabled at install time and requires the reserved set of
+CPUs as input. That configuration is currently passed in via a
+manually created manifest containing a machine config resource. The
+set of CPUs used for management workload partitioning has to align
+with the reserved set configured by the performance-addon-operator
+(PAO) during day 2. For OCP 4.8, the onus is on the user to ensure
+these align, there is no interlock.
+
+The current implementation is prone to user error, because it has two
+entities configuring and managing the same data. This design document
+describes phase 2 of the implementation plan for workload partitioning
+to improve the user experience when configuring the feature.
+
+### Goals
+
+1. Describe the remaining work at a high level for architectural
+   review and planning.
+2. Document the division of responsibility between workload management
+   in components in the OpenShift release payload and PAO.
+3. List the other enhancements that need to be written, and their
+   scopes.
+
+### Non-Goals
+
+1. Resolve all of the details for each of the next steps. We're not
+   going to talk a lot about names or implementation details here.
+
+## Proposal
+
+### API
+
+The initial implementation is limited to single-node deployments. To
+support multi-node clusters, we need to solve a couple of concurrency
+problems. The first is providing a way for the admission hook
+responsible for mutating pod definitions to reliably know when the
+feature should be enabled. The admission hook cannot store state on
+its own, so we need a new API to indicate when the feature is
+enabled.
+
+We eventually want to extend the partitioning feature to support
+non-management workloads to include workload types defined by cluster
+admins. For now, however, we want to keep the interface as simple as
+possible.  Therefore, the new API will be read-only, and only used so
+that the admission hook (and other consumers) can tell when workload
+partitioning is active.
+
+The new API will need a controller to manage it. We propose to add a
+controller to the existing cluster-config-operator, to avoid adding
+the overhead of another image, pod, etc. to the cluster.
+
+We propose that the API be owned by core OpenShift, and defined in the
+`openshift/api` repository. A separate enhancement will be written to
+describe the API and controller in more detail.
+
+### Owner for Enablement
+
+Workload management needs two node configuration steps which overlap
+with work PAO is doing today:
+
+* Set reserved CPUs in kubelet config and enable the CPU manager
+  static policy
+* Set systemd affinity to match the reserved CPU set
+
+PAO already has logic for managing CPU pools and configuring kubelet
+and CRI-O. Since it already does a lot of what is needed to enable
+workload partitioning, we propose to extend it to handle the
+additional configuration currently being done manually by users,
+including the machine config with settings for kubelet and CRI-O.
+
+The effect of this decision is that the easiest way to use the
+workload partitioning feature will be through the
+performance-addon-operator. This means that the implementation of the
+feature is split across a few components, and not all of them are
+delivered in all clusters, but it avoids duplicating a lot of the work
+already done in the PAO.
+
+Workload partitioning is currently enabled as part of installing a
+cluster. This provides predictable behavior from the beginning of the
+life-cycle, and avoids extra reboots required to enable it in a
+running cluster (an important consideration in environments with short
+maintenance windows). To continue to support enabling the feature this
+way, and to simplify the configuration, we propose to extend the PAO
+with a `render` command, similar to the one in other operators, to
+have it generate manifests for the OpenShift installer to consume.
+
+### Future Work
+
+1. Write an enhancement to describe the API the admission hook will
+   use to determine when workload partitioning is enabled.
+2. Write an enhancement to describe the changes in
+   performance-addon-operator to manage the kubelet and CRI-O
+   configuration when workload partitioning is enabled, including the
+   `render` command.
+3. Write an enhancement to describe an API for enabling workload
+   partitioning in an existing cluster.
+
+### User Stories
+
+N/A
+
+### Risks and Mitigations
+
+There is some risk of shipping a feature with part of the
+implementation in the OpenShift release payload but the enabling tool
+delivered separately. PAO is considered to be a "core" OpenShift
+component, even though it is delivered separately. There is not a
+separate SKU for it, for example. The PAO team has been working
+closely with the node team on the implementation of this feature, so
+we do not anticipate any issues delivering the finished work in this
+way.
+
+## Design Details
+
+### Test Plan
+
+The other enhancements will provide details of the test plan(s)
+needed.
+
+### Graduation Criteria
+
+N/A
+
+#### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+- Enumerate service level indicators (SLIs), expose SLIs as metrics
+- Write symptoms-based alerts for the component(s)
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+N/A
+
+### Version Skew Strategy
+
+N/A
+
+## Implementation History
+
+- https://issues.redhat.com/browse/OCPPLAN-6065
+- https://issues.redhat.com/browse/CNF-2084
+
+## Drawbacks
+
+None
+
+## Alternatives
+
+We could move more of the PAO implementation into the release payload,
+so that users who want the workload partitioning feature do not need
+another component. This would either duplicate a lot of the existing
+PAO work or make future deliver of PAO updates more complicated by
+tying them to OpenShift releases.
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -139,12 +139,14 @@ operators, to have it generate manifests for the OpenShift installer
 to consume. This avoids any need to change the OpenShift installer to
 support this uncommon use case.
 
-The `render` command has an option to enable the management
-partition. When the option is given, the `render` command creates a
-manifest for the new enablement API with status data that includes the
-`management` workload type to enable partitioning. In the future, it
-might enable partitioning for all of the types mentioned in the input
-PerformanceProfiles, or based on some other input.
+The `render` command will have an option to enable the management
+partition. When the option is given, the `render` command will create
+a manifest for the new enablement API with status data that includes
+the `management` workload type to enable partitioning (the details of
+that API will be defined in a separate design document). In the
+future, the PAO might enable partitioning for all of the types
+mentioned in the input PerformanceProfiles, or based on some other
+input.
 
 For each PerformanceProfile, the PAO will render MachineConfig
 manifests with higher precedence than the default ones created by the
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -99,15 +99,24 @@ enabled.
 We eventually want to extend the partitioning feature to support
 non-management workloads to include workload types defined by cluster
 admins. For now, however, we want to keep the interface as simple as
-possible.  Therefore, the new API will only have status fields and
-will only be used so that the admission hook (and other consumers) can
-tell when workload partitioning is active.
+possible.  Therefore, even though the new API will have both spec and
+status fields, only the status fields will be used so that the
+admission hook (and other consumers) can tell when workload
+partitioning is active.
 
 To manage the risk associated with rolling out a complicated feature
 like this, we are going to continue to require it to be enabled as
 part of deploying the cluster. The API does not need to support
 configuring the workload partitioning in a running cluster, so no
 controller is needed, for now.
+
+Because of the extensive hardware-specific planning needed to use
+partitioning successfully, there is no clear business case for ever
+allowing the feature to be enabled in a running cluster. We will
+therefore also include an admission hook to prevent anyone from
+changing or deleting the API resource after it is created, following
+the pattern established in
+<https://github.com/openshift/kubernetes/blob/master/openshift-kube-apiserver/admission/customresourcevalidation/features/validate_features.go>.
 
 We propose that the API be owned by core OpenShift, and defined in the
 `openshift/api` repository. A separate enhancement will be written to
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -48,17 +48,24 @@ manually created manifest containing a machine config resource. The
 set of CPUs used for management workload partitioning has to align
 with the reserved set configured by the performance-addon-operator
 (PAO) during day 2. For OCP 4.8, the onus is on the user to ensure
-these align, there is no interlock.
+these align, there is no interlock.  The current implementation is
+prone to user error because it has two entities configuring and
+managing the same data.
 
-The current implementation is prone to user error because it has two
-entities configuring and managing the same data. This design document
-describes phase 2 of the implementation plan for workload partitioning
-to improve the user experience when configuring the feature.
+The 4.8 implementation is also limited to single-node deployments. We
+anticipate users of highly-available clusters also wanting to minimize
+the overhead of the management components to take full advantage of
+compute capacity for their own workloads.
+
+This design document describes phase 2 of the implementation plan for
+workload partitioning to improve the user experience when configuring
+the feature and to extend support to multi-node clusters.
 
 ### Goals
 
-1. Describe the remaining work at a high level for architectural
-   review and planning.
+1. Describe, at a high level, the remaining work to improve the user
+   experience and expand support to multi-node deployments for
+   architectural review and planning.
 2. Document the division of responsibility between workload management
    in components in the OpenShift release payload and PAO.
 3. List the other enhancements that need to be written, and their
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -144,7 +144,7 @@ windows). To continue to support enabling the feature during
 deployment, and to simplify the configuration, we propose to extend
 the PAO with a `render` command, similar to the one in other
 operators, to have it generate manifests for the OpenShift installer
-to consume. This avoids any need to change the OpenShift installer to
+as extra manifests. This avoids any need to change the OpenShift installer to
 support this uncommon use case, and is consistent with the way
 third-party networking solutions are installed.
 
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -10,6 +10,7 @@ reviewers:
   - "@deads2k"
   - "@sttts"
   - maintainers of PAO
+  - maintainers of MCO
 approvers:
   - "@markmc"
   - "@derekwaynecarr"
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -373,7 +373,10 @@ has a PerformanceProfile associated and the PAO is installed.
 If the MachineConfigPool does not match a PerformanceProfile, there
 will be no cpuset information and the PAO will generate MachineConfigs
 with partitioning enabled but not tied to a cpuset. This is somewhat
-safe, but may lead to unexpected behavior.
+safe, but may lead to unexpected behavior. The mutated pods would
+float across the full cpuset in the same way as if the feature
+enabled, however the cpushares would not be deducted from available
+CPUs, potentially leading to over commit scenarios.
 
 If the PAO is not installed, it will not generate the overriding
 MachineConfigs for the new MachineConfigPool. The nodes in the pool
--- a/enhancements/workload-partitioning/rationalizing-configuration.md
+++ b/enhancements/workload-partitioning/rationalizing-configuration.md
@@ -249,7 +249,9 @@ to partition the workloads into specific CPUs.
 13. Admission plugin uses the workload types on the workloads CR to
     decide when to mutate pods.
 14. CRI-O sees pods with workload annotations and uses the resource
-    request to set cpushares but not cpuset.
+    request to set cpushares but not cpuset. See the [Risks
+    section](#risks-and-mitigations) for details of the effects this
+    may have on the cluster.
 15. User installs the performance-addon-operator into the cluster.
 16. User creates PerformanceProfiles and adds them to the cluster.
 17. PAO generates new MachineConfigs, adding the cpuset information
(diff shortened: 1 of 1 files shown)


enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,213 @@
...
+
+We eventually want to extend the partitioning feature to support
+non-management workloads to include workload types defined by cluster
+admins. For now, however, we want to keep the interface as simple as
+possible.  Therefore, the new API will be read-only, and only used so
+that the admission hook (and other consumers) can tell when workload
+partitioning is active.
+
+The new API will need a controller to manage it. We propose to add a
+controller to the existing cluster-config-operator, to avoid adding
```
dhellmann, line 92: Is cluster-config-operator what we said?

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,213 @@
...
+problems. The first is providing a way for the admission hook
+responsible for mutating pod definitions to reliably know when the
+feature should be enabled. The admission hook cannot store state on
+its own, so we need a new API to indicate when the feature is
+enabled.
+
+We eventually want to extend the partitioning feature to support
+non-management workloads to include workload types defined by cluster
+admins. For now, however, we want to keep the interface as simple as
+possible.  Therefore, the new API will be read-only, and only used so
```
deads2k, line 87: A read-only API is very hard to build since it reflects status from somewhere.  I suggest a normal, CRD based API with a limited set of writers.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,213 @@
...
+that the admission hook (and other consumers) can tell when workload
+partitioning is active.
+
+The new API will need a controller to manage it. We propose to add a
```
deads2k, line 91: I'm not really clear on why this API isn't an API where the user expresses his desire and the node changes flow from this desire.

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: Maybe it's a terminology thing, but what we meant here is to have a CRD with only status fields that normal users wouldn't write to.

What other form of read-only APIs are there and how can we better clarify this?

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: We expect to extend the PerformanceProfile CRD to be used to turn this on. We don't want the admission hook in the core to depend on that CRD, since it is part of the performance-addon-operator and may not be installed in all clusters.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,213 @@
...
+including the machine config with settings for kubelet and CRI-O.
+
+The effect of this decision is that the easiest way to use the
+workload partitioning feature will be through the
+performance-addon-operator. This means that the implementation of the
+feature is split across a few components, and not all of them are
+delivered in all clusters, but it avoids duplicating a lot of the work
+already done in the PAO.
+
+Workload partitioning is currently enabled as part of installing a
```
deads2k, line 121: so is the flow here
1. user expresses desire for management CPU pool
2. PAO configures the kubeletconfig and systemd somehow
3. PAO watches the state of kubelet config and systemd and updates an API to indicate to the admission plugin: "this is now enabled"

dhellmann: Not quite.

We thought step 3 should not rely on PAO, because the admission plugin always needs to be able to answer the question about whether partitioning is enabled for a particular type. We also don't want to tie that enablement to the implementation details within kubelet or CRI-O. Instead, either the cluster-config-operator or machine-config-operator will have a new controller that looks at the nodes in the cluster to track when they advertise the partition-specific opaque resources (caused by the kubelet config being updated) and manage a status API (to be designed in another enhancement).

deads2k: > We thought step 3 should not rely on PAO, because the admission plugin always needs to be able to answer the question about whether partitioning is enabled for a particular type.

Above you indicated that the "PerformanceProfile CRD" is used to tell the PAO to configure this.  If the flow requires the PAO to enable, then requiring the PAO to set your status indicator CRD seems reasonable.  The admission plugin could still be active all the time, but at least the user desire and the state of the cluster would be managed by a single component.

dhellmann: So, the status API would be defined in the `openshift/api` repo and available to both the admission plugin and PAO. To keep the admission hook from having a dependency on something defined in PAO, that API needs to not be the PerformanceProfile itself. The CRD for the status API would always be installed, but there may not always be an instance. Something in PAO would create the instance and keep it up to date.

Does that make sense as an approach?

MarSik: Creating another object is easy enough, so yes, PAO can do this. The only question is whether it should be always enabled or whether we should create the instance based on something (master+worker case vs. SNO vs. RWN vs. three nodes total etc..).

deads2k: > The CRD for the status API would always be installed, but there may not always be an instance. Something in PAO would create the instance and keep it up to date.
> 
> Does that make sense as an approach?

That could work.  So could attaching a status field to an existing resource (is this related to Infrastructure?) that is empty until the PAO sets the value.

dhellmann: We have updated the text to always use the PAO to configure partitioning.

dhellmann:
@derekwaynecarr we have updated the text of this design based on the discussion we had about using the PAO during installation.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,460 @@
...
+3. Kubelet starts and sees the config file enabling partitioning. It
+   advertises the workload resource type on the Node and mutates
+   static pods with partitioning annotations.
+4. CRI-O sees pods with workload annotations and uses the resource
+   request to set cpushares and cpuset.
+
+If the MachineConfigPool does not match a PerformanceProfile, there
+will be no cpuset information and the PAO will generate MachineConfigs
+with partitioning enabled but not tied to a cpuset. This is somewhat
+safe, but may lead to unexpected behavior.
```
browsell, line 369: The mutated pods would float across the full cpuset as they would without the feature enabled, however the cpu shares would not get deducted from alloctable cpus potentially leading to over commit scenarios.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,460 @@
...
+
+## Summary
+
+The initial iteration of workload partitioning focused on a short path
+to a minimum viable implementation. This enhancement describes the
+loose ends for preparing the feature for GA at a high level, and
+explains the set of other design documents that need to be written
+separately during the next iteration.
+
+## Motivation
```
browsell, line 41: One other motivation is to allow this capability to be leveraged on non-single node configurations

dhellmann: Great point, I've updated the motivation and goal list to include that.

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: Thanks, I've included this info in this section.

dhellmann:
/cc @kikisdeliveryservice @sinnykumari @yuqi-zhang 

It would be good to have your feedback on this design. I don't think we anticipate any MCO changes, but it's involved in the processes we've outlined and I'd like to ensure we're not making any invalid assumptions.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,471 @@
...
+11. The cluster has partial partitioning enabled for `management`
+    workloads. Pods for management workloads are mutated, but not
+    actually partitioned.
+12. Kubelet starts and sees the config file enabling partitioning. It
+    advertises the workload resource type on the Node and mutates
+    static pods with partitioning annotations.
+13. Admission plugin uses the workload types on the workloads CR to
+    decide when to mutate pods.
+14. CRI-O sees pods with workload annotations and uses the resource
+    request to set cpushares but not cpuset.
```
MarSik, line 252: Would the same "pods floating and resources not accounted for" like in new MCP case happen here if steps 15..17 do not happen soon?

dhellmann: Yes, that's right. I can update the text to make that clear.

The reboot in step 18 fixes that by forcing all pods on the node to be rescheduled.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,479 @@
...
+input.
+
+For each PerformanceProfile, the PAO will render MachineConfig
+manifests with higher precedence than the default ones created by the
+installer during bootstrapping. These manifests will include all of
+the details for configuring `kubelet`, CRI-O, and `tuned` to know
+about and partition workloads, including the cpusets for each workload
+type.
+
+The PAO `render` command also generates a manifest with the CRD for
```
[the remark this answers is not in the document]
browsell: It will need to be extended to generate the CRD manifest as well.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,479 @@
...
+### User Stories
+
+#### Enabling and Configuring at the Same Time
+
+In this workflow, the user provides all of the information needed to
+fully partition the workloads from the very beginning.
+
+1. User runs the installer to `create manifests` to get standard
+   manifests without any partitioning.
+2. The user creates PerformanceProfile manifests for each known
```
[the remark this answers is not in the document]
browsell: Wouldn't we need those CR's to be applied though ?

enhancements/workload-partitioning/rationalizing-configuration.md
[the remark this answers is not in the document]
dhellmann: Right, this is a forward-looking statement about what the render command will do when the design is implemented.

enhancements/workload-partitioning/rationalizing-configuration.md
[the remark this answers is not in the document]
dhellmann: It does seem like we would normally want the PerformanceProfiles added to the cluster so that when the PAO is installed it has the same inputs available as the render command. What would be the benefit of not installing them?

I think we discussed it, but I don't remember the answer. Does the PAO need any of the other manifests the installer will create in step 1? If not, should we move step 3 before step 1 to make that clear?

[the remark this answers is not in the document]
MarSik: YEs, I think even the PerformanceProfile needs passed to the installer. For multiple reasons: all the generated objects will have an ownership reference pointing to it, and the user will not want to post it again post install (even though it would probably work too).

About the order, PAO does not need anything from the installer (yet), it is not reading anything. That might change as we might be checking the MCPs and Nodes to generate the right label selectors.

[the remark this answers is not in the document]
MarSik: My only concern is the PerformanceProfile CRD. That one is normally owned by OLM...

enhancements/workload-partitioning/rationalizing-configuration.md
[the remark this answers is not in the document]
MarSik: Yep, we are doing it backwards here. We have the initial implementation before the design :) But we need this part indeed, so the implementation will change.

enhancements/workload-partitioning/rationalizing-configuration.md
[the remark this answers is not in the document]
dhellmann: If `render` includes the CRD, then either the OLM version would overwrite it or OLM could simply update the ownership like some other operators do. We should verify that that works, though, in case we need to teach OLM that some things it installs might already exist.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,479 @@
...
+cluster, and that will not change as part of the implementation of
+this design. Enabling partitioning early in this way provides
+predictable behavior from the beginning of the life-cycle, and avoids
+extra reboots required to enable it in a running cluster (an important
+consideration in environments with short, fixed maintenance
+windows). To continue to support enabling the feature during
+deployment, and to simplify the configuration, we propose to extend
+the PAO with a `render` command, similar to the one in other
+operators, to have it generate manifests for the OpenShift installer
+to consume. This avoids any need to change the OpenShift installer to
```
sttts, line 147: I don't understand how the installer consumes these. Are you sure you write what you mean?

dhellmann: "consume" may be a bad choice of words, here. I think all we mean is that they are passed as extra manifests, and the installer does what it would normally do with any extra manifests.

```suggestion
as extra manifests. This avoids any need to change the OpenShift installer to
```

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+4. Installer bootstraps the cluster.
+5. Bootstrapping runs the machine-config-operator (MCO) `render`
+   command, which generates MachineConfig manifests with low
+   precedence.
+6. Bootstrapping uploads both sets of MachineConfig manifests, one
+   after the other.
+7. The machine-config-operator (MCO) applies MachineConfigs to nodes
+   in the cluster.
+8. Bootstrapping finishes and the cluster is launched.
+9. If the MCO does not apply KubeletConfig in step 7, it must do it
```
kikisdeliveryservice, line 240: a little confused on this step 9, the cluster is up and done, but the MCO will just roll some changes out? what is the trigger on this? does the user have to do something?

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+## Design Details
+
+### Open Questions
+
+1. What runs the PAO `render` command in the assisted installer
+   workflow?
+2. How hard do we need to work to ensure that the PAO version matches
+   the release payload version for the cluster being deployed?
+   What/who is responsible for that?
+3. We need to ensure MCO render handles the KubeletConfig CR generated
```
kikisdeliveryservice, line 420: I guess some of this wasnt' clear to me but mostly what we're talking about is PAO->kubelet/containerruntimes CRDs? or is PAO looking to create its own that it's expecting the MCO to render into a renderedconfig?

kikisdeliveryservice (commented):
A couple questions, but also think node should look at this since they own all things kubelet & container runtime in the MCO

cc: @mrunalp @rphillips

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+1. User runs the installer to `create manifests` to get standard
+   manifests without any partitioning.
+2. The user creates PerformanceProfile manifests for each known
+   MachineConfigPool and adds them to the installer inputs.
+3. The user runs the PAO `render` command to read the
+   PerformanceProfile manifests and create additional manifests, as
+   described above.
+4. User runs the installer to `create cluster`.
+5. Installer bootstraps the cluster.
+6. Bootstrapping runs the machine-config-operator (MCO) `render`
```
yuqi-zhang, line 195: I'm a bit confused as to what points 6-8 are pointing at. It sounds to me that you are trying to describe the mechanism for which the MCO bootstraps and ensures that the PAO-generated special kubelet/crio configs are ordered. But these steps don't really explicitly highlight that, and also is a bit vague as to the details.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+   command, which generates MachineConfig manifests with low
+   precedence.
+7. Bootstrapping uploads both sets of MachineConfig manifests, one
+   after the other.
+8. The machine-config-operator (MCO) applies MachineConfigs to nodes
+   in the cluster.
+9. Bootstrapping finishes and the cluster is launched.
+10. If the MCO does not apply KubeletConfig in step 8, it must do it
```
yuqi-zhang, line 203: I'm a bit confused. Up until now you do not explicitly mention kubeletconfigs, but instead use machineconfigs. Is this what the PAO is generating?

If so this is today not doable as a day 1 operation, and will effectively be day2 (extra reboots across all nodes) but will be automatic

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+1. The MCO applies the MachineConfigs to the node.
+2. Kubelet starts and sees the config file enabling partitioning. It
+   advertises the workload resource type on the Node and mutates
+   static pods with partitioning annotations.
+3. CRI-O sees pods with workload annotations and uses the resource
+   request to set cpushares and cpuset.
+
+#### Day 2: Add a New MachineConfigPool
+
+Adding a new MachineConfigPool is safe only if the MachineConfigPool
```
yuqi-zhang, line 362: all custom MCPs inherit from worker, so they must have worker-MCP based configs.

This is somewhat of an issue in the kubeletconfig use case today. The kubeletconfig doesn't really order based on pool priority but implied MCO naming-schema priority, which means if a worker pool and a custom pool both have custom kubeletconfigs, the one that actually gets laid down is alphabetically ordered, meaning that `custom, worker` will result in worker configs, and `worker,zz-custom` will result in zz-custom configs.

The suggestion today I have is not to have base worker kubeletconfigs if you have custom pools that need different ones. All nodes that require separate kubeletconfigs must be in a custom pool (empty worker pool). I think we should reconsider this section

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+If a MachineConfigPool exists without a matching PerformanceProfile,
+there will be no cpuset information and the PAO will generate
+MachineConfigs with partitioning enabled but not tied to a cpuset.
+This is somewhat safe, but may lead to unexpected behavior. The
+mutated pods would float across the full cpuset in the same way as if
+partitioning was not enabled and the cpushares would not be deducted
+from available CPUs, potentially leading to over commit scenarios.
+
+If the PAO is not installed into a cluster with workload partitioning
+enabled and configured, adding a new MachineConfigPool can be unsafe.
```
yuqi-zhang, line 404: Like above, I think we should carefully consider custom pool use cases and whether we want to support them at all.

yuqi-zhang (commented):
I have some general questions/concerns, but fundamentally speaking I think I am confused about the details of the proposal. Is my understanding correct in:

1. Today this is being shipped as 4.8 SNO only, with custom machineconfigs (and not kubelet/containerruntime config objects) being used to configure the node's kubelet/crio configuration to facilitate their workload partitioning
2. In the future, the PAO will have a separate pre-install command that allows users to generate custom kubeletconfigs/containerruntimeconfig manifests that get applied to the nodes, to facilitate workload partitioning.
3. this happens on firstboot during install (does not really work today via kubeletconfigs)
4. the support will extend to custom pools and day 2 (lots of caveats for this, is not directly supported today)
5. the actual CPU checking is not part of this, since this is all pre-install, so failures will just be the node failing to join the cluster (and openshift-install failing)

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: It wasn't clear to us whether the MCO render command consumed KubeletConfig resources, so that they would be included in the roll-out of the initial configuration to the host, or if that would have to happen once the cluster was up. From what @yuqi-zhang says above, it sounds like it's the latter, today. So the trigger is the MCO waking up and noticing that some of the extra manifests created by the PAO and added to the cluster during installation were KubeletConfig and that those haven't been applied.

The user can change the settings after the cluster is up using the PAO's PerformanceProfile API.

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: That's not any different from what we have today, right? I don't know what the best practices for using PerformanceProfiles is, but I don't expect that to change dramatically as a result of this work.

@browsell or @MarSik may have more insight into how they're being used today.

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: The PAO does define its own PerformanceProfile CRD, and it renders them down to KubeletConfig and MachineConfig resources already. Some of the work for this release will be to extend that to include the new settings, but we don't expect the MCO to understand any new APIs.

https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md#high-level-end-to-end-workflow has a bit more detail about what's being done manually today that we're trying to automate here.

dhellmann:
> I have some general questions/concerns, but fundamentally speaking I think I am confused about the details of the proposal. Is my understanding correct in:
> 
> 1. Today this is being shipped as 4.8 SNO only, with custom machineconfigs (and not kubelet/containerruntime config objects) being used to configure the node's kubelet/crio configuration to facilitate their workload partitioning

Yes, see https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md for some details about what go into those MachineConfigs.

> 2. In the future, the PAO will have a separate pre-install command that allows users to generate custom kubeletconfigs/containerruntimeconfig manifests that get applied to the nodes, to facilitate workload partitioning.

Right. The idea, at least for now, is that this PAO pre-install command will be the *only* way to turn on workload partitioning. The settings can be changed day 2, but it cannot be enabled day 2.

> 3. this happens on firstboot during install (does not really work today via kubeletconfigs)
> 4. the support will extend to custom pools and day 2 (lots of caveats for this, is not directly supported today)
> 5. the actual CPU checking is not part of this, since this is all pre-install, so failures will just be the node failing to join the cluster (and openshift-install failing)

Yeah, PAO has some logic for simple validation but workload partitioning is a non-trivial feature that requires the user to understand their hardware pretty deeply, so it's up to them to use it correctly. Over time we may be able to continue to improve the user experience.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+the `management` workload type to enable partitioning (the details of
+that API will be defined in a separate design document). In the
+future, the PAO might enable partitioning for all of the types
+mentioned in the input PerformanceProfiles, or based on some other
+input.
+
+For each PerformanceProfile, the PAO will render MachineConfig
+manifests with higher precedence than the default ones created by the
+installer during bootstrapping. These manifests will include all of
+the details for configuring `kubelet`, CRI-O, and `tuned` to know
```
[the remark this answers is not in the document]
MarSik: Correct, PAO uses KubeletConfiguration for setting up system cpu reserved, cpu and topology manager policies. Everything else is done by files and systemd services injected via MC Ignite.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+about and partition workloads, including the cpusets for each workload
+type.
+
+The PAO `render` command also generates a manifest with the CRD for
+PerformanceProfile so bootstrapping does not block when trying to
+install the PerformanceProfile manifests.
+
+### Future Work
```
derekwaynecarr, line 169: any special instructions on how to address or pull out the pao binary for eventually invoke the render command?

derekwaynecarr (changes requested):
I am in favor of adding the `render` command to the `pao` and using those manifests as inputs to the installer, its consistent with third-party CNIs or other supplementary content that is desired on initial install that is not delivered in core release image.  The assisted installer, openshift-install, and hive flows all support this pattern.

A few minor clarifications from my side, but will let @dhellmann respond to remaining comments from others before merging.

enhancements/workload-partitioning/rationalizing-configuration.md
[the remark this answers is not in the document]
yuqi-zhang: Ok, I just want to make sure that we are not overwriting ourselves. If multiple different machineconfigs refer to the same file, we may end up in an undesirable state

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+
+#### Enabling and Configuring at the Same Time
+
+In this workflow, the user provides all of the information needed to
+fully partition the workloads from the very beginning.
+
+1. User runs the installer to `create manifests` to get standard
+   manifests without any partitioning.
+2. The user creates PerformanceProfile manifests for each known
+   MachineConfigPool and adds them to the installer inputs.
```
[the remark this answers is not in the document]
yuqi-zhang: non-worker non-master MCPs cannot be booted into today (there are a lot of open discussions on this). So you'd effectively be booting first into worker then update into custom.

enhancements/workload-partitioning/rationalizing-configuration.md
yuqi-zhang: We did have some chatter about allowing day-1 kubeletconfig generation (adding the kubeletconfigcontroller into the bootstrap MCC), so that's a potential option if really needed

enhancements/workload-partitioning/rationalizing-configuration.md
yuqi-zhang: No, that is the behaviour today. I just want to make sure we are not overwriting our own configs. Custom pools need to be handled very carefully

enhancements/workload-partitioning/rationalizing-configuration.md
yuqi-zhang: Sure :) sorry, I can try to highlight the ordering:

The bootstrap node runs bootstrap versions of 1 MCController 1 MCServer. The MCC is responsible for generating master configs, and the MCS is responsible for serving that master config. 

At the time the MCC runs, it should see all manifests applied to the cluster (as templates directly), at least if it directly targets the MCO namespace. However there is no guarantee that other objects are ready and updated, which is where race conditions have occured in the past: if any of them get modified after bootstrap MCC has finished but the in-cluster MCC has not yet started. (also tangentially related to the kubeletconfig not being day-1 since there is nothing processing those objects into Machineconfigs yet).

The master nodes come up and form the real cluster, at which point it runs the "in-cluster" MCC and MCS, which generates the config to serve worker nodes.

So in that flow, any extra manifests, provided the are created before bootstrap are run, are no different than any template-shipped MCs that are applied day1. The only thing that matters then is the alphanumeric naming schema of the MC object, which determines their order of merging into the "rendered config". That is to say, if 2 MCs named `99-xx` and `11-xx` write to the same file, only the contents of `99-xx` will be there, and whatever is in `11-xx` is "silently dropped"

yuqi-zhang:
Thanks for the answers @dhellmann , I think I understand the proposal a bit more now. Reading https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md it seems that we shouldn't 

So my main remaining concerns boil down to:
1. day 2 support pertaining to `The settings can be changed day 2, but it cannot be enabled day 2.` (e.g. custom pools do not support day1)
2. custom pool support and ordering of MCs (world work today but want to make sure we consider ordering)

Which we can discuss further via another medium, if you'd like, doesn't necessarily have to be resolved as part of this enhancement

enhancements/workload-partitioning/rationalizing-configuration.md
MarSik: Yeah, this is not going to change.

Btw the fact that KubeletConfigs cannot be layered is already causing us some headaches. It is a different feature altogether, but we need to set cpu manager stuff and customers want to enable unsafe syscalls. And there just is no way to do that and keep ownership of just half of the config.

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: We do want to spend some time thinking about the UX there. Maybe that should happen in the PAO enhancement (item 2 below).

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+   experience and expand support to multi-node deployments for
+   architectural review and planning.
+2. Document the division of responsibility between workload management
+   in components in the OpenShift release payload and PAO.
+3. List the other enhancements that need to be written, and their
+   scopes.
+4. Minimize the complexity of installing a "normal" cluster while
+   still supporting this very special case.
+
+### Non-Goals
```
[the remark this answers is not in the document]
dhellmann: Done

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: I added a note to cover that.

enhancements/workload-partitioning/rationalizing-configuration.md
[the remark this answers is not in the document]
dhellmann: I've tried to clarify that most of the manifests are the same as what PAO already generates.

enhancements/workload-partitioning/rationalizing-configuration.md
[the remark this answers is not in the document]
dhellmann: If that's already a limitation, I think we can ignore it for now and address it independently of the workload-partitioning feature in the future.

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: @mrunalp @browsell @MarSik It sounds like we may have to deal with a race condition here.

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: > We did have some chatter about allowing day-1 kubeletconfig generation (adding the kubeletconfigcontroller into the bootstrap MCC), so that's a potential option if really needed

Yes, we're trying to simultaneously configure things properly from the very beginning of the cluster's life (to avoid some very ugly edge cases) and avoid a ton of reboots (to minimize deploy time, since metal reboots veeeery slowly).

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: I've added some wording to clarify the timing here.

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: I don't think we're going to be able to avoid supporting custom pools.

dhellmann:
I think I have responded to all of the open comments. I resolved threads that seemed ready. There are a couple of answered questions that I left open in case the answers are insufficient, and we have 2 potential issues to work through (status-only API types and the race described in https://github.com/openshift/enhancements/pull/753#discussion_r654072660). Since this is a high-level plan, if we can't resolve those quickly here perhaps we can defer discussion to #804

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+
+The initial implementation is limited to single-node deployments. To
+support multi-node clusters, we need to solve a couple of concurrency
+problems. The first is providing a way for the admission hook
+responsible for mutating pod definitions to reliably know when the
+feature should be enabled. The admission hook cannot store state on
+its own, so we need a new API to indicate when the feature is
+enabled.
+
+We eventually want to extend the partitioning feature to support
```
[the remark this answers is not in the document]
deads2k: > @deads2k had objections to status only fields in past when we handled aws resource tags.
> 
> [openshift/api#905](https://github.com/openshift/api/pull/905)

Yes.  Status-only fields need to be able to describe

1. how they will eventually allow human change if it is needed
2. how they react to a cluster-admin setting the field to a different value

I think this API can probably describe it, but it's important to consider it here.  For instance, I think the performance add-on operator could probably re-stomp status if it were changed (needs confirmation).  To allow a human to change it, the  performance add-on operator would need to learn that it could be externally controlled.

Clusters without the PAO won't be able to express "turn it on" without a spec field.  Would a spec field update be prevented under some conditions?

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,481 @@
...
+6. Bootstrapping uploads both sets of MachineConfig manifests, one
+   after the other.
+7. The machine-config-operator (MCO) applies MachineConfigs to nodes
+   in the cluster.
+8. Bootstrapping finishes and the cluster is launched.
+9. If the MCO does not apply KubeletConfig in step 7, it must do it
+   here after the MCO is running in the cluster and can reconcile
+   normally.
+11. The cluster has partial partitioning enabled for `management`
+    workloads. Pods for management workloads are mutated, but not
```
deads2k, line 246: is this the bit via the kube-apiserver admission or kubelet admission?

enhancements/workload-partitioning/rationalizing-configuration.md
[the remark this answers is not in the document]
MarSik: PAO cannot own this object or guarantee any reconcile loop. It is an optional operator installed by OLM.

PAO will merely know how to prepare it using render mode just like it does for MC, KubeletConfig and others to make it easy for admins. But any admin can do the same manually even without PAO.

enhancements/workload-partitioning/rationalizing-configuration.md
```diff
@@ -0,0 +1,478 @@
...
+enabled.
+
+We eventually want to extend the partitioning feature to support
+non-management workloads to include workload types defined by cluster
+admins. For now, however, we want to keep the interface as simple as
+possible.  Therefore, the new API will only have status fields and
+will only be used so that the admission hook (and other consumers) can
+tell when workload partitioning is active.
+
+To manage the risk associated with rolling out a complicated feature
```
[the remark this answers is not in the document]
MarSik: The current design does not expect any reconcile loop to exist, so what you describe is probably what will happen indeed.

enhancements/workload-partitioning/rationalizing-configuration.md
MarSik: Yep, this is something I touched upon in the PAO render mode enhancement. The model we want to use is to let the user execute it via `podman pullspec/pao:version <render-command-stuff>` from his laptop.

enhancements/workload-partitioning/rationalizing-configuration.md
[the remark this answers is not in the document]
dhellmann: We talked about this offline and agreed that we will have spec & status fields but that we will block changes using an admission plugin. See 160aa5d53d648f4d043c475a44a94a17db3d1ae8

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: Both kubelet and the admission hook will do the mutation, but CRI-O won't have the settings to place the containers on a specific cpuset.

deads2k:
update in https://github.com/openshift/enhancements/pull/753/commits/160aa5d53d648f4d043c475a44a94a17db3d1ae8 lgtm

kikisdeliveryservice:
Has node chimed in? MCO shouldn't be the only team signing off since node owns some of this and should take look.

enhancements/workload-partitioning/rationalizing-configuration.md
dhellmann: After speaking with @yuqi-zhang offline, I added a note in the risk section to cover this. See fe703ad7924eccf5eb27fcc50bc2ba44235a7727

dhellmann:
> Has node chimed in? MCO shouldn't be the only team signing off since node owns some of this and should take look.

Good point. @mrunalp is a co-author and @rphillips and @haircommander were involved in the initial implementation in the previous release. Here we're describing an API on top of that work, and I don't think the details change at the node level, but it wouldn't hurt for them to take a look.

PR #756, opened 2021-04-26
hack/template-lint: Actually, use PULL_BASE_SHA

wking:
Instead of 320ec318e3 (#755), to [avoid][1]:

    fatal: ambiguous argument 'origin/master..': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'

[`PULL_BASE_SHA` is part of Prow][2].

When folks call us outside of Prow, fall back to `HEAD`.  This isn't a great fallback for folks using the script in local development, but:

```console
$ PULL_BASE_SHA=origin/master hack/template-lint.sh
```

or whatever is possible, if a bit of a hack, if they don't like `HEAD`. And because folks can call the upstream remote whatever they want, `origin/master` isn't all that reliable either.

Also add the `--` suggested in Git's error message, to show that we intend this to be a revision ID and not a file path.

[1]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_enhancements/744/pull-ci-openshift-enhancements-master-markdownlint/1386818564115140608/artifacts/test/build-log.txt
[2]: https://github.com/kubernetes/test-infra/blob/73e6e4703b38b43fe262ba0554d642e9aebcc0e8/prow/jobs.md#job-environment-variables

### DIFF
--- a/hack/template-lint.sh
+++ b/hack/template-lint.sh
@@ -3,14 +3,17 @@
 # Check that the required section headers from the template are
 # present in the enhancement document(s).
 
+set -o errexit
+set -o nounset
+set -o pipefail
+
 TEMPLATE=$(dirname $0)/../guidelines/enhancement_template.md
 
 # We only look at the files that have changed in the current PR, to
 # avoid problems when the template is changed in a way that is
 # incompatible with existing documents.
 CHANGED_FILES=$(git log --name-only --pretty= "${PULL_BASE_SHA:-origin/master}.." -- \
-                    | grep '^enhancements' \
-                    | grep '\.md$' \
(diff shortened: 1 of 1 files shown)


wking:
I've pushed 74d7078 -> 88ba895 pivoting from `HEAD` to `origin/master` as the default for non-Prow use.  `origin/master` might not work if the user has used a different name for their remote, but `HEAD` wasn't a good match for folks looking to recover an already-committed change that was failing its Prow presubmits.  I'm not sure how often folks will actually run this locally, so I'm not strongly opinionated.

PR #757, opened 2021-04-27
tools: update enhancement newsletter to separate updates from new docs

dhellmann:
Separate out the PRs that change existing enhancements documents, and
only include a 1 line summary instead of repeating the original summary
from the document. This deemphasizes "typo fix" and other minor updates
to existing designs.

/assign @russellb

### DIFF
--- a/tools/cmd/report.go
+++ b/tools/cmd/report.go
@@ -8,7 +8,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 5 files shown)


dhellmann:
I tried to break the change into consumable commits, so it might be easier to review it that way instead of looking at the full diff.

dhellmann:
https://hackmd.io/4V_zWdl4QCSISfmLIs-gvw has some example output

PR #758, opened 2021-04-27
Security Context Constraints for CSI Volumes

adambkaplan:
Extend SCCs to set an allowlist for ephemeral CSI drivers.
CSI drivers can provide ephemeral CSI volume mounts, which are used for
things like Vault sealed secrets and the new Projected Resource CSI
driver.

### DIFF
--- a/enhancements/security/scc-csi-volumes.md
+++ b/enhancements/security/scc-csi-volumes.md
@@ -32,13 +32,13 @@ superseded-by: []
 Ephemeral CSI volumes are a newer type of volume in Kubernetes, which allows content to mounted into pods with underlying ephemeral file systems.
 OpenShift Security Context Constraints (SCCs) take an all-or-nothing approach when it comes to these volumes.
 A service account/user can either mount a `csi` volume, or it cannot.
-This proposal will enhnance SCCs to control which CSI drivers are allowed.
+This proposal will enhance SCCs to control which CSI drivers are allowed.
 
 ## Motivation
 
-Ephemeral CSI volumes are provided via CSI drivers that can be added to OpenShift.
-When a pod mounts a `csi` volume, it must specify the driver that will provide the underlying content.
-Cluster admins will need a means to approve which service accounts can use a particular CSI driver.
+Inline CSI volumes are provided via CSI drivers that can be added to OpenShift.
+When a pod mounts an inline `csi` volume, it must specify the driver that will provide the underlying content.
+Cluster admins will need a means to identify which drivers are considered safe for use, including in restricted namespaces.
 
 Ephemeral CSI drivers are expected to be used in the following use cases:
 
@@ -47,7 +47,7 @@ Ephemeral CSI drivers are expected to be used in the following use cases:
 
 ### Goals
 
-* Allow cluster admins to control which ephemeral CSI drivers can be mounted via SCCs.
+* Allow cluster admins to control which ephemeral CSI drivers can be used to mount `csi` volumes.
 
 ### Non-Goals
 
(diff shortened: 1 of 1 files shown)


enhancements/security/scc-csi-volumes.md
```diff
@@ -0,0 +1,194 @@
...
+To prevent the mounting of CSI volumes, admins should remove `csi` from the list of allowed volumes in the SCC.
+
+### Risks and Mitigations
+
+**Risk:** Default security context constraints will allow any CSI driver to be used
+
+*Mitigations:*
+
+- Only the `privileged` and `node-exporter` default SCCs allow CSI volumes to be mounted (they allow any volume mount)
+- Testing will ensure that new SCCs
```
bparees, line 107: ensure that new SCCs what?

enhancements/security/scc-csi-volumes.md
```diff
@@ -0,0 +1,234 @@
...
+
+#### Pod admission
+
+Pod that mount `csi` volumes cannot be admitted unless the chosen SCC allows `csi` volumes.
+In addition, the driver for the volume mount must have the `security.openshift.io/allow-csi-volumes-restricted` annotation.
+
+#### Allow CSI volumes in restricted namespaces
+
+Ephemeral CSI drivers will be promoted to the list of "safe" volume mounts, and be considered equivalent to a Secret or ConfigMap from a safety/restriction perspective.
+CSI volumes will be added to the list of allowed volume types in the `restricted` SCC.
```
adambkaplan, line 95: @deads2k I think we need to be careful here during cluster upgrades. Let's say in 4.9 we allow CSI volumes in the restricted SCC and update openshift-apiserver to check the driver annotation. During upgrade we can have a window where 4.9 SCC manifests are deployed and 4.8 apiservers are still alive. In that situation we would allow inline `csi` volumes on any pod.

What do you think of the following graduation criteria:

1. 4.9 `csi` volumes cannot be mounted unless the driver has the `security.openshift.io/allow-csi-volumes-restricted`  annotation
2. 4.10 we update the default SCCs to allow these volume mounts in restricted namespaces.

enhancements/security/scc-csi-volumes.md
```diff
@@ -0,0 +1,234 @@
...
+
+*Mitigation:*
+The upgrade test suites must ensure that pods cannot be created with `csi` volumes in a restricted namespace.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. Should we allow the privileged SCC to mount any CSI volume (via an `AllowAllCSIVolumeDrivers` flag)?
+   This has implications for OpenShift builds, which currently uses the `privleged` SCC.
```
adambkaplan, line 126: If such an admin "get of jail" flag is added, then we should consider a separate SCC for builds. We will likely want this anyway as we enable unprivileged builds.

enhancements/security/scc-csi-volumes.md
deads2k: If we go with an approach that annotates CSI drivers, then no SCC will change.  A 4.8 kube-apiserver will deny the pod.  A 4.9 kube-apiserver would run both admission plugins.

The CSI driver annotation approach means getting the storage team to implement this admission plugin.

adambkaplan: How does this help us for a pod that uses the restricted SCC? Won't they be denied admission?

enhancements/security/scc-csi-volumes.md
```diff
@@ -0,0 +1,234 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Ephemeral CSI volumes are a newer type of volume in Kubernetes, which allows content to mounted into pods with underlying ephemeral file systems.
```
bparees, line 32: Are "ephemeral" CSI volumes distinct from "CSI volumes" in general, particularly as regards the SCC changes being proposed here?   My impression is no, but this EP specifically states "ephemeral" a lot, implying to me some sort of difference.

enhancements/security/scc-csi-volumes.md
```diff
@@ -0,0 +1,234 @@
...
+#### Safe for Consumption Annotation
+
+CSI drivers that provide inline CSI volumes must mark themselves as safe for consumption.
+To mark a CSI driver safe, admins or an operator will need to add the `security.openshift.io/allow-csi-volumes-restricted` annotation to the driver's `CSIDriver` object:
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: CSIDriver
+annotations:
+  security.openshift.io/allow-csi-volumes-restricted: "true"
```
bparees, line 80: this is a pretty confusing name.  Is it allowing something? is it restricting it?  Is it allowing use of otherwise restricted things?  Restricting use of otherwise allowed things?

enhancements/security/scc-csi-volumes.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+#### Pod admission
+
+Pod that mount `csi` volumes cannot be admitted unless the chosen SCC allows `csi` volumes.
+In addition, the driver for the volume mount must have the `security.openshift.io/allow-csi-volumes-restricted` annotation.
+
+#### Allow CSI volumes in restricted namespaces
+
+Ephemeral CSI drivers will be promoted to the list of "safe" volume mounts, and be considered equivalent to a Secret or ConfigMap from a safety/restriction perspective.
```
bparees, line 94: promoted how?

enhancements/security/scc-csi-volumes.md
```diff
@@ -0,0 +1,234 @@
...
+All default SCCs installed by OpenShift with less restrictions will likewise allow `csi` volumes.
+When SCCs are ordered, the `csi` volume type will count as a "trivial" volume mount.
+
+### Risks and Mitigations
+
+**Risk:**
+Default security context constraints will allow any CSI driver to be used.
+
+*Mitigation:*
+Only the CSI drivers that are marked safe by an operator or admin will be allowed to mount `csi` volumes.
```
bparees, line 105: what does this mean for the existing SCCs that allow all CSI drivers today?  Are we going to potentially lock down users from being able to use CSI drivers they used to be able to use, because they aren't explicitly marked "safe"?

bparees (commented):
There is a lot of mixed use of "CSI volume", "CSI driver", "Inline CSI volume", "Ephemeral CSI volume", and "Ephemeral CSI driver".  It's difficult to tell if those are all the same thing, or are being treated distinctly/differently by this EP.  I think some term definitions and consistency of phrasing would make it easier to fully understand the implications of what is being proposed.

adambkaplan:
/close

I proposed this as an initial idea, patterning off of prior behavior. However, I'm not an expert in SCCs and the intricacies of extending them in this fashion.

@s-urbaniak this is a blocker for making the Shared Resource CSI driver available for GA. There is growing interest in ephemeral CSI drivers not just from the build team, but others who wish to add the Secret Store CSI driver to their clusters [1]. The Service Mesh team is also interested in this capability. Is this something the auth team can design for 4.10/4.11?

[1] https://secrets-store-csi-driver.sigs.k8s.io/

PR #759, opened 2021-04-27
workload-partitioning: reorganize

dhellmann:
We are going to be writing a few more enhancements related to the
workload partitioning feature, so create a subdirectory and move the
existing design doc there.

/cc @mrunalp @browsell


mrunalp (approved):
/lgtm

PR #770, opened 2021-05-05
update template to explain about keeping all sections

dhellmann:
Add a note to the process at the top of the template explaining that
all required sections must be kept in the published doc.

/assign @wking


wking (approved):
/lgtm

PR #771, opened 2021-05-05
add eus MCO enhancement

rphillips:


enhancements/update/eus-upgrades-mco.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Open Questions [optional]
+
+### Test Plan
+
+- CI tests are necessary which attempt to upgrade while violating kubelet to API
+compatibility, ie: 4.6 to 4.7 upgrade with MachineConfigPools paused, then check
+for Upgradeable=False condition to be set by the API Server Operator, assuming
+that our rules only allow for N-1 skew.
+- CI tests will be necessary to veferify the events emitted from the MCO
```
kikisdeliveryservice, line 90:
```suggestion
- CI tests will be necessary to verify the events emitted from the MCO
```

sdodson:
@rphillips Are we still pursuing this?

rphillips:
@sdodson yes.

enhancements/update/eus-upgrades-mco.md
```diff
@@ -0,0 +1,171 @@
...
+
+This also creates a scenario where admins wish to reduce both the duration and
+the disruption to workload associated with the upgrade.
+
+### Goals
+
+- The MCO will look at all the pools within the cluster, if there are paused
+pools the MCO will make sure the nodes have a n-2 skew.
+- If there is a paused pool with a node skew of greater than n-2, then an
+event will be emitted.
```
cgwalters, line 57: Will this block the control plane (master) pool from upgrading or not?

An event seems odd.  If the goal is to block, then I think we need to change the MCO to go `Upgradable=False` so the CVO will refuse upgrades, right?
If the goal is to *not* block...well, some sort of important/critical alert seems in order, not just an event right?

wking: > If the goal is to block...

The API-server is in charge of blocking; see #762.  The MCO is supposed to remind folks that they have old nodes, and provide suggestions about getting them updated.  I'd expected that to be via alerts, but I guess spitting out events periodically could accomplish the same thing.

sdodson: Looks like implementation is moving forward in https://github.com/openshift/machine-config-operator/pull/2658
I guess the assumption is that we'll follow up with alerts based on the events?

sinnykumari: I think current implementation PR is more generic, it will send event if any node have skew >= 2

PR #772, opened 2021-05-05
metallb: Update operator, CRD, and BGP stack plans

russellb:
This PR includes a few updates on the proposal for how to proceed with
integrating MetalLB with OpenShift

* Clarify that an OLM based operator, starting with an upstream
  implementation, is the preferred approach for the operator.

* Note that we're working on an upstream CRD based configuration
  interface, but detail a backup plan for layer2-only support using an
  existing OpenShift API.  This could potentially buy us another release
  worth of time to sort out our desired longer term configuration
  interface upstream.

* Add a pointer to our proposal upstream to adopt FRR as an option for
  providing BGP for MetalLB instead of MetalLB's existing custom BGP
  implementation.

### DIFF
--- a/enhancements/network/on-prem-service-load-balancers.md
+++ b/enhancements/network/on-prem-service-load-balancers.md
@@ -58,6 +58,7 @@ custom manner.  For more information on the existing `IngressIPs` feature:
 
 * https://github.com/openshift/api/blob/master/config/v1/types_network.go#L99
 * https://github.com/openshift/openshift-docs/pull/21388
+* https://docs.openshift.com/container-platform/4.7/networking/configuring_ingress_cluster_traffic/configuring-externalip.html
 
 ### Goals
 
@@ -229,35 +230,55 @@ doing all of your testing in a virtual network built on top of it.
 
 #### Operator
 
-This is the first of two alternatives for how we might integrate MetalLB in
-OpenShift.
-
 We must also create an operator for MetalLB.  We should develop an operator
-that is generally useful to the MetalLB community.  We should also have an
-OpenShift version of this operator for our use.
+that is generally useful to the MetalLB community.  The upstream operator has
+been proposed [here](https://github.com/metallb/metallb/pull/859). We should
+also have an OpenShift version of this operator for our use, which would
+contain any needed OpenShift specific customizations.
 
-It is assumed that the MetalLB operator would be managed by OLM as an optional
-additional component to be installed on on-premise clusters.  However, in the
-[ROADMAP.md
+The MetalLB operator will be managed by OLM as an optional additional component
+to be installed on on-premise clusters.  However, in the [ROADMAP.md
 document](https://github.com/openshift/enhancements/blob/master/ROADMAP.md),
 there is an item to "Front the API servers and other master services with
 service load balancers".  If this functionality is required at install time,
-the details on management of this operator may be revisited.
+the details on management of this operator may be revisited, since MetalLB may
+become a hard requirement in some environments.
 
 There is a start of a
 [metallb-operator](https://github.com/cybertron/metallb-operator) available and
 a [video demo](https://www.youtube.com/watch?v=WgOZno0D7nw).
 
-#### Alternative Integration: Cloud Controller Manager
+#### Configuration Interfaces
 
-An alternative integration approach would be via a cloud controller manager
-(CCM). An example of this is the [packet.net
-CCM](https://github.com/packethost/packet-ccm), which ensures MetalLB is
-deployed and also configures it properly to work in packet.net’s BGP
-environment.
+MetalLB's current configuration interface is a single `ConfigMap`.  We only
+want to support CRD based configuration through OpenShift.  The ultimate goal
+will be to enhance MetalLB with a native CRD based configuration interface.
+The design work has begun here:
 
-These integration options must be explored in more detail as part of a more
-detailed integration proposal.
+https://github.com/metallb/metallb/pull/833
+
+Since this is an upstream effort, the completion of this work may take longer
+than the timeline for when we want to ship the first piece of MetalLB support
+for OpenShift, which will be to support MetalLB in layer2 mode only.
+
+To mitigate this timing risk, we will consider adding a minimal configuration
+CRD to the MetalLB operator which only covers the configuration needed for
+layer2 mode. This will buy us some additional time to determine whether we will
+build a more complete interface within the core MetalLB project, or whether we
+will continue to extend the configuration interface provided by the operator.
+
+The operator will reconcile this minimal layer2-only configuration CRD by
+creating a MetalLB ConfigMap.
+
+#### FRR for BGP
+
+We have concerns about the long term viability of the built-in BGP stack in
+MetalLB.  We have multiple features we'd like to add to the BGP stack and feel
+the project is better served by relying on an external BGP implementation.  We
+are actively working upstream to propose the option of using FRR as the BGP
+stack behind MetalLB.
+
+https://github.com/metallb/metallb/pull/832
 
 ### Risks and Mitigations
 
@@ -374,6 +395,24 @@ to think it owns a `Service`.
 These risks to version skew must be mitigated through upstream community
 engagement and contributions, as well as informed management of upgrades in the
 MetalLB operator.
+
+### Graduation Criteria
+
+This feature is expected to be released as GA (layer2 only as a first release),
+and later BGP once the FRR integration and complete CRD based configuration
+interface are complete.
+
+#### Dev Preview -> Tech Preview
+
+N/A
+
+#### Tech Preview -> GA
+
+N/A
+
+#### Removing a deprecated feature
+
+N/A
 
 ## Implementation History
 


enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,101 @@ doing all of your testing in a virtual network built on top of it.
...
+spec:
+  ...
+  externalIP:
+    autoAssignCIDRs:
+    - 10.10.10.0/24
+```
+
+This makes OpenShift automatically set the `externalIP` field, but nothing will
+make sure those addresses are reachable automatically.  That's where MetalLB
+comes in.  If the OpenShift MetalLB operator is enabled, the operator can
```
msherif1234, line 283: what does metallb operator means ? I think we are making assumption here that metallb controler and speaker are deployed and metallb-system namespace is created so CNO can use this a trigger to form the Metallb's ConfigMap ?

enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,101 @@ doing all of your testing in a virtual network built on top of it.
...
+
+Note how the configuration copies over the CIDR for the externalIP allocator
+used in OpenShift.  This makes MetalLB expect to see addresses in this range
+used for load balancer Services.  Also note the use of `auto-assign: false`.
+In this scenario, we're letting OpenShift do the address allocation piece.
+
+If we support this as a configuration interface, customers will be able to opt
+for the more feature rich CRD interface offered by MetalLB natively
+once it is ready.  As long as the `AutoAssignCIDRs` feature is not enabled on
+the OpenShift side, there should not be a conflict.
```
msherif1234, line 312: how about abnormal cases where u have Metallb then someone config AutoAssignCIDRs , or the other way around do we need to detect and fail the config if the field is in use ?

enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,101 @@ doing all of your testing in a virtual network built on top of it.
...
+
+**TODO** -- There is one outstanding problem to make this fallback plan work.
+The OpenShift allocator sets spec.ExternalIPs, while MetalLB expects that an
+external IP allocator will set spec.LoadBalancerIP.  We must fix this mismatch
+on one side or the other.
```
msherif1234, line 317: ok this specific to auto-assign false, in which k8s resource is that , I didn't have to mess with that so I have no idea where that need to be set ?

msherif1234 (commented):
pls call out all metallb layers conifg limitation like customer will be limited to one addresspool, the only exposed config the pool address rnanage

enhancements/network/on-prem-service-load-balancers.md
markdgray: AutoAssignCIDRs is basically an IPAM method for service external IPs? So in this case, we would be giving the option to use this IPAM instead of the MetalLB one?

Whats the advantage of this over a custom CRD (that may not be the final one)? Users will still need to migrate from this configuration method to the new, final one?

markdgray: AutoAssignCIDRs is basically an IPAM method for service external IPs? So in this case, we would be giving the option to use this IPAM instead of the MetalLB one?

Whats the advantage of this over a custom CRD (that may not be the final one)? Users will still need to migrate from this configuration method to the new, final one?

enhancements/network/on-prem-service-load-balancers.md
russellb: "MetalLB opeartor is enabled" -- The MetalLB operator has been installed via OLM (Operator Lifecycle Manager) -- the method we use to manage the lifecycle of optional operators for a cluster.

I actually think the operator itself should generate the configmap, and not CNO.

I'll try to clarify this in the text

enhancements/network/on-prem-service-load-balancers.md
russellb: > AutoAssignCIDRs is basically an IPAM method for service external IPs? So in this case, we would be giving the option to use this IPAM instead of the MetalLB one?

Yes, that's right.

> Whats the advantage of this over a custom CRD (that may not be the final one)? Users will still need to migrate from this configuration method to the new, final one?

Advantages:

- no new CRD required.  It's a simple an existing configuration interface that should be able to support indefinitely (for a limited layer2 only use case anyway).  So I think this is a path that won't *require* any migration, unless a customer chooses to for more config capabilities.

Disadvantages:

- Doesn't support all of the config flexibliity for layer2 that metallb is capable of (would come later with our full CRD interface)

enhancements/network/on-prem-service-load-balancers.md
russellb: See this code in the metallb controller: https://github.com/metallb/metallb/blob/main/controller/service.go#L89-L96

It supports something else performing the IPAM function, but the field it looks at to see what IP has been requested is `LoadBalancerIP`, and not `ExternalIPs`, which is what the OpenShift IPAM knows how to set.  We need to reconcile this somehow.

Choices include ...

1) Change MetalLB to also look at spec.ExternalIPs (but this doesn't seem correct to me)

2) Change OpenShift to optionally set LoadBalancerIP instead.  It could be by duplicating the `AutoAssignCIDRs` functionality with a near identical option called `AutoAssignLBCIDRs` or something like that.

3) Write a controller that copies the IP from `ExternalIPs` to `LoadBalancerIP` on Services when we know we're running in this mode.  This feels pretty ugly, but would probably work ...

markdgray: 1. Agreed, I don't think this is correct either
2. With potentially different CIDRs?
3. A bit ugly but it was the first solution that I considered.

enhancements/network/on-prem-service-load-balancers.md
msherif1234: CNO is currently has the controller for most of the network CRD which is where AutoAssignCIDRs is part of

enhancements/network/on-prem-service-load-balancers.md
msherif1234: 2 this will require changes in the API repo  to add another field ?
AutoAssignLBCIDRs []string `json:"address-pools,omitempty"`

russellb: On 2) 
* Maybe we could raise an alert if both options were configured.  I don't think it makes sense for both to be used at the same time.
* Yes, this would be a new API addition as you describe.

3 seems like the "easiest" solution.  We could do this entirely in the OpenShift metallb operator and it would be part of how it works when it's operating in this "auto-configure based on AutoAssignCIDRs" mode.  It just feels like an ugly hack.

I think my preference is #2.  It seems the most explicit and clean to me.  It seems like we should have time to get this in.

squeed: I agree, Option 2, (with a suitable config flag) seems like the cleanest and easiest to implement.

russellb: Thanks, all.  The PR has been updated to reflect option 2 as the proposed approach.

It includes a proposed option name of `AutoAssignLoadBalancerCIDRs`, but I'm not tied to the name at all.

russellb:
@squeed @knobunc would the two of you mind being the approvers on this PR?

enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,123 @@ doing all of your testing in a virtual network built on top of it.
...
+
+https://github.com/metallb/metallb/pull/833
+
+Since this is an upstream effort, the completion of this work may take longer
+than the timeline for when we want to ship the first piece of MetalLB support
+for OpenShift, which will be to support MetalLB in layer2 mode only.
+
+To mitigate this timing risk, we have another option for a way to configure the
+layer2 mode of MetalLB by adding a small new feature to OpenShift's existing
+network config object.  OpenShift as an existing feature to automatically
```
squeed, line 265: typo: "OpenShift *has* an"

enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,123 @@ doing all of your testing in a virtual network built on top of it.
...
+spec:
+  ...
+  externalIP:
+    autoAssignLoadBalancerCIDRs:
+    - 10.10.10.0/24
+```
+
+`autoAssignCIDRs` and `audoAssignLoadBalancerCIDRs` must not be specified at
+the same time.  If both are specified, the cluster-network-operator should
+raise an alert about an invalid configuration.
```
squeed, line 302: I actually think we can catch this in CRD validation.

enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,123 @@ doing all of your testing in a virtual network built on top of it.
...
+
+The cluster network operator should have new code similar to its handling of
```
squeed, line 304: This is actually implemented in the openshift-controller-manager, FYI.

squeed:
This seems reasonable to me. If @sttts (or a delegate) wants to review the API implications, that would be nice. Otherwise I think this is basically mergeable.

enhancements/network/on-prem-service-load-balancers.md
russellb: thanks!  fixed this

enhancements/network/on-prem-service-load-balancers.md
russellb: cool, updated

enhancements/network/on-prem-service-load-balancers.md
russellb: thanks, fixed

russellb:
> This seems reasonable to me. If @sttts (or a delegate) wants to review the API implications, that would be nice. Otherwise I think this is basically mergeable.

Summary of the API impact for @sttts or a delegate:

We have an existing `AutoAssignCIDRs` option in the openshift network config for providing a range of IP addresses where OpenShift will set these addresses in the `ExternalIPs` field in the spec for Services of type: LoadBalancer.  We have a very similar use case here for having OpenShift manage a pool of addresses for the `LoadBalancerIP` field in the spec for Services of type: LoadBalancer.  They would be handled almost identically, except they set the allocated IP address in a different field.  There's more details in the PR text.

enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,123 @@ doing all of your testing in a virtual network built on top of it.
...
-It is assumed that the MetalLB operator would be managed by OLM as an optional
-additional component to be installed on on-premise clusters.  However, in the
-[ROADMAP.md
+The MetalLB operator will be managed by OLM as an optional additional component
+to be installed on on-premise clusters.  However, in the [ROADMAP.md
 document](https://github.com/openshift/enhancements/blob/master/ROADMAP.md),
 there is an item to "Front the API servers and other master services with
 service load balancers".  If this functionality is required at install time,
-the details on management of this operator may be revisited.
+the details on management of this operator may be revisited, since MetalLB may
```
dcbw, line 243: @russellb FYI @aojea has plans to make this not required, at least for Go-based things, by making client-go able to accept multiple API servers and balance/fail-over when required, instead of funneling into a single point of failure like the current keepalived/haproxy metal model or the slow-as-molasses-to-update cloud LB model, both of which we've had tons of problems with.

russellb: thanks!  that's good to know.

I think the impact of that would mean it's less likely that MetalLB would ever be a required component.

enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,123 @@ doing all of your testing in a virtual network built on top of it.
...
+metadata:
+  name: cluster
+spec:
+  ...
+  externalIP:
+    autoAssignLoadBalancerCIDRs:
+    - 10.10.10.0/24
+```
+
+`autoAssignCIDRs` and `audoAssignLoadBalancerCIDRs` must not be specified at
```
dcbw, line 300: @russellb how would this new field interact in cloud environments? Would it be disallowed there where the cloud provider integration handles LB address assignment? Will the CNO inspect the cloud platform and allow/deny this field automatically?

russellb: It could be useful for any platform where specifying `LoadBalancerIP` in the Service spec is allowed, but I don't know which of those that would be.

We could do something to flag this if the infrastructure platform is not none or baremetal and only allow it where we know it's valid and we have tested and documented it.

dcbw: @russellb I guess I was thinking about other cloud platforms that handle all the LBIP stuff themselves; if that's enabled, wouldn't this conflict with that or give the user the wrong impression of what will happen?

russellb: You're absolutely right.  The problem already exists to some degree.  Services have `LoadBalancerIP` in their spec, which I assume behaves differently depending on the platform.  We're raising the visibility of it by putting IPAM support for the field in our own API, though.  I think we should just document that we only test the use of this with metal (and point to the metallb operator).

enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,123 @@ doing all of your testing in a virtual network built on top of it.
...
+
+```yaml
+apiVersion: config.openshift.io/v1
+kind: Network
+metadata:
+  name: cluster
+spec:
+  ...
+  externalIP:
+    autoAssignLoadBalancerCIDRs:
```
abhat, line 296: should this field be part of the `ExternalIPConfig`?

russellb: Thanks for raising this.  It doesn't make sense there.  I moved it to a new section like:

```
    loadBalancer:
        autoAssignCIDRs:
            - ...
```

enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,119 @@ doing all of your testing in a virtual network built on top of it.
...
+```yaml
+apiVersion: config.openshift.io/v1
+kind: Network
+metadata:
+  name: cluster
+spec:
+  ...
+  loadBalancer:
+    autoAssignCIDRs:
+    - 10.10.10.0/24
```
danwinship, line 297: I think the naming should make it clearer that this is only expected to be useful in combination with MetalLB.

Additionally, given that fact, there's no reason this assignment necessarily needs to be configured in the cluster network config, or be implemented by openshift-controller-manager. It seems like this is something that would be useful to other MetalLB users too, and could perhaps be done upstream?

Also also, if it does end up in `Network.config.openshift.io`, you should not assume that the external IP config is necessarily a good model to follow in terms of how the config should look. That had the additional constraint that it had to be compatible with what we had done in 3.x.

enhancements/network/on-prem-service-load-balancers.md
```diff
@@ -229,35 +230,119 @@ doing all of your testing in a virtual network built on top of it.
...
+
+Note how the configuration copies over the CIDR for the LoadBalancerIP allocator
+used in OpenShift.  This makes MetalLB expect to see addresses in this range
+used for load balancer Services.  Also note the use of `auto-assign: false`.
+In this scenario, we're letting OpenShift do the address allocation piece.
+
+If we support this as a configuration interface, customers will be able to opt
+for the more feature rich CRD interface offered by MetalLB natively
+once it is ready.  As long as the `loadBalancer.autoAssignCIDRs` feature is not
+enabled on the OpenShift side, there should not be a conflict.
```
danwinship, line 335: so wait, does MetalLB already support `auto-assign: true`? If so, then why are we reimplementing it in OCP?

russellb: It does, yes, but via its ConfigMap config interface, and we want to only direct customers to use CRDs.  We're working on a full CRD config interface upstream, but it's unlikely to be ready in time for 4.9.  There's a link to that inline.

This is one approach to having an OpenShift specific CRD interface: a generally useful API for configuring load balancer IPAM (if it is indeed useful for any other platform, not actually sure).

enhancements/network/on-prem-service-load-balancers.md
russellb: > I think the naming should make it clearer that this is only expected to be useful in combination with MetalLB.

Would you say that the spec.LoadBalancerIP field for Services is also only useful for metallb?  I don't know the history of that field, or what other cloud providers can honor it, if any.

From an OpenShift perspective, indeed it's only metallb right now, but I expected it *could* be useful for some other cloud if any other supported that field, but I haven't gone looking for what else supports it.

> Additionally, given that fact, there's no reason this assignment necessarily needs to be configured in the cluster network config, or be implemented by openshift-controller-manager. It seems like this is something that would be useful to other MetalLB users too, and could perhaps be done upstream?

I answered this a bit to your other comment, but MetalLB does have its own IPAM, but no CRD config interface.

The goal here if we zoom out a bit is supporting a CRD based interface for configuring metallb in its layer2 mode.  This proposal actually started because we thought the existing externalIP allocator may be good enough, but it turned out metallb ignores that field, so then we proposed creating a new feature that mirrored the existing one.

An alternative would be to do this entirely within the downstream MetalLB operator.  We could have an OpenShift specific MetalLB config CRD, and start by only implementing as little as possible for layer2 mode, but still pursue a more complete CRD interface upstream that supports all features, including BGP configuration.

> Also also, if it does end up in `Network.config.openshift.io`, you should not assume that the external IP config is necessarily a good model to follow in terms of how the config should look. That had the additional constraint that it had to be compatible with what we had done in 3.x.

Fair point - open to suggestions.  It's a pretty simple thing, just need 1 or more CIDRs.

russellb:
Based on feedback from @dcbw and @danwinship in particular, I'm going to revise this to suggest that the new configuration option be done as MetalLB specific in the MetalLB operator instead of as a general OpenShift feature in the network config.  We're not convinced enough that it'd ever be useful outside of this context.

russellb:
@markdgray @msherif1234 @squeed @danwinship @dcbw This update reflects the change to creating a minimal CRD managed by the MetalLB operator which will cover layer2 mode only at first.

squeed:
Agreed, /lgtm

PR #774, opened 2021-05-06
network: Remove OpenShift-SDN requirement for Bare Metal Load Balancer

markdgray:
I think OpenShift-SDN is not a requirement for this but I want to clarify by raising this PR


russellb:
PR #772 includes changes that will fix the linting job.

I agree with this.  Architecturally I still want to strive for being CNI plugin agnostic, but when it comes to what we test and support, I'd agree that it makes sense to be ovn-kube only.

/approve

PR #775, opened 2021-05-06
Drop Node Hostname Resolution

cybertron:
In some Edge deployment scenarios, compute resources on remote
worker nodes can be in short supply. Deployers of these have
objected to the presence of unnecessary services on their nodes.
As a result, we would like to remove any services that are no
longer required for the cluster to function.

### DIFF
--- a/enhancements/network/drop-node-hostname-resolution.md
+++ b/enhancements/network/drop-node-hostname-resolution.md
@@ -1,0 +1,198 @@
+---
+title: drop-requirement-for-node-hostname-resolution
+authors:
+  - "@cybertron"
+reviewers:
+  - @yboaron
+  - TBD
+approvers:
+  - TBD
+creation-date: 2021-05-05
+last-updated: 2021-05-05
+status: implementable
+see-also:
+  - "/enhancements/network/baremetal-networking.md"
+  - https://github.com/openshift/enhancements/pull/654
+---
+
+# Drop Requirement for Node Hostname Resolution
+
+## Release Signoff Checklist
+
+- [X] Enhancement is `implementable`
+- [X] Design details are appropriately documented from clear requirements
+- [X] Test plan is defined
+- [X] Operational readiness criteria is defined
+- [X] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+When the initial baremetal IPI implementation was done, there was a requirement
+for node hostnames to be resolvable from other nodes. Since that time there
+have been changes in how bootstrapping happens (among other things) that seem
+to have removed this requirement. In order to provide node name resolution,
+baremetal IPI runs an independent coredns pod on each node that retrieves a
+list of nodes and provides DNS records for them. Some deployments, particularly
+resource-limited edge environments, do not want this additional service running
+on their nodes. We are proposing to remove the requirement for node hostnames
+to be resolvable so we can remove this service.
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


cybertron:
Note that this is a different coredns instance from the one deployed by the dns operator. This one is deployed as a static pod in on-prem clusters to provide the DNS resolution that other platforms get from their cloud. It doesn't have any resolution for cluster services. The only things we get from this coredns are:

- api-int.CLUSTER_DOMAIN
- master-0.CLUSTER_DOMAIN
- master-1.CLUSTER_DOMAIN
- etc...

We think that only the first one is actually required today.

enhancements/network/drop-node-hostname-resolution.md
```diff
@@ -0,0 +1,198 @@
...
+- [X] Enhancement is `implementable`
+- [X] Design details are appropriately documented from clear requirements
+- [X] Test plan is defined
+- [X] Operational readiness criteria is defined
+- [X] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+When the initial baremetal IPI implementation was done, there was a requirement
```
yboaron, line 31: Shouldn't it be 'on-prem IPI' instead of 'baremetal IPI' ?

enhancements/network/drop-node-hostname-resolution.md
```diff
@@ -0,0 +1,198 @@
...
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+On upgrade, the coredns static pod will be removed and replaced with an
+instance of dnsmasq running on the host that will only provide the api-int
+record. We will need to resolve port conflicts between coredns and dnsmasq
```
yboaron, line 157: I'm not sure if we should handle the corednד/dnsmasq conflict for upgrade/downgrade, 
the dns server (dnsmasq/coreDNS) is used only by the local node,
from what I recall, in MCO upgrade/downgrade path node being rebooted first and then start with the new MCO image (for upgrade: it will include  dnsmasq ,  for downgrade: it should include coredns)

enhancements/network/drop-node-hostname-resolution.md
```diff
@@ -0,0 +1,198 @@
...
+## Implementation History
+
+## Drawbacks
+
+As noted earlier, removing node hostname resolution from on-prem platforms
+will introduce a difference in behavior from the cloud platforms. Since
+efforts have been made to avoid reliance on DNS I don't see this as a major
+problem, but it is less than ideal.
+
+## Alternatives
```
yboaron, line 183: Q: is the main motivation is remote workers case? 

If the answer to the Q is yes, do you think that the following could be an alternative (though this will compromise DNS performance) ?
1. Master nodes will continue running CoreDNS as today
2. Remove core-DNS pod from workers 
3. Update resolv.conf prepender script  on workers to point to one of the masters  (api-vip )

cybertron:
As a followup to this, we had someone from the API team tell us these records are still required, although the specific examples mentioned were things we tested as working without the records. However, I've also recently been contacted by someone from the assisted installer team and learned they are using the DNS node names as part of their deployment process. I don't entirely understand how since the node has to join the cluster before its DNS record becomes available, but that's something we'll need to follow up on if we decide to revisit this.

romfreiman:
@cybertron who from the ai team? Not sure it's correct

cybertron:
@romfreiman I was chatting with Mat Kowalski. He was having trouble with a dual stack deployment and thought it was related to the fact that we only populated one IP version per node (which we've actually fixed recently).

PR #781, opened 2021-05-13
CORS-1650: RHEL 8 Server Worker/Infra Node Support

mtnbikenc:
This enhancement proposes to add RHEL 8 Server support for deploying
worker/infra nodes with [openshift-ansible](https://github.com/openshift/openshift-ansible).
RHEL8 nodes will be deployed and upgraded using the same Ansible
playbook process however, RPM content will be delivered using an
image in the release package.

### DIFF
--- a/enhancements/installer/rhel8_worker_support.md
+++ b/enhancements/installer/rhel8_worker_support.md
@@ -7,7 +7,7 @@ reviewers:
 approvers:
   - "@derekwaynecarr"
 creation-date: 2021-05-13
-last-updated: 2021-06-04
+last-updated: 2021-06-23
 status: provisional
 see-also:
   - "https://issues.redhat.com/browse/CORS-1650"
@@ -32,19 +32,15 @@ superseded-by:
 
 This enhancement proposes to add traditional RHEL 8 Server support for deploying worker/infra
 nodes with [openshift-ansible].  OpenShift will be deployed and upgraded on RHEL 8 nodes using
-the same [Ansible playbook process][adding-rhel-compute] however, OpenShift RPM content will be delivered using
-an image in the release payload.
+the same [Ansible playbook process][adding-rhel-compute].
 
 ## Motivation
 
-The process of scaling up traditional RHEL 8 worker nodes by using containerized OpenShift content will provide
-a better user experience than configuring OpenShift RPM repositories on hosts.  Using the
-[existing process][adding-rhel-compute] will allow near-term support of RHEL 8 nodes while developing enhanced
-longer-term RHEL 8 support.
+Using the [existing process][adding-rhel-compute] will allow near-term support
+of RHEL 8 nodes while developing enhanced longer-term RHEL 8 support.
 
 ### Goals
 
-1. Implement an os-content image for RHEL 8 OpenShift RPMs or files.
 1. Update [openshift-ansible] scaleup.yml and upgrade.yml playbooks to support traditional RHEL 8 workers
 
 ### Non-Goals
@@ -56,35 +52,21 @@ longer-term RHEL 8 support.
 
 ## Proposal
 
-Currently, for RHEL 7 worker node deployment, the user must [prepare the host][prepare-rhel]
-before running the scaleup.yml playbook.  This process can be cumbersome for
-automating deployment and also requires additional steps to update OpenShift repos before
-performing an OpenShift upgrade on the node.  For traditional RHEL 8 hosts, preparing OpenShift repos on the hosts will
-not be required because the necessary content will be delivered as an image in
-the release payload.  The RHEL 8 hosts will still need to be subscribed and have
-base channels enabled for package dependencies and operating system updates.
-
 The basic process for scaling up and upgrading RHEL worker nodes will remain the
-same.  The playbooks will be updated to download and mount the image with the OpenShift RPM
-content during the install/upgrade for RHEL 8 hosts.  The installation process will
+same.  The playbooks will be updated to install the appropriate RPMs based on
+the OS during the install/upgrade for RHEL 8 hosts.  The installation process will
 not change for RHEL 7 worker nodes.
 
 To migrate from RHEL 7 worker nodes, new RHEL 8 nodes should be provisioned,
 and the existing RHEl 7 worker nodes removed.
 
-[prepare-rhel]: https://docs.openshift.com/container-platform/4.7/machine_management/adding-rhel-compute.html#rhel-preparing-node_adding-rhel-compute
 
 ### User Stories
 
 1. As an OpenShift admin using [openshift-ansible], I want to deploy and upgrade
-   traditional RHEL 8 Server worker nodes without needing to configure OpenShift repos for installation.
-
-1. As an Openshift admin, I want to be able to deploy traditional RHEL 8 workers without
-   having to mirror OpenShift RPM content.
+   traditional RHEL 8 Server worker nodes.
 
 ### Open Questions [optional]
-
-1. Can we use existing os-content used for RHCOS to install required RPMs or files?
 
 ---
 
--- a/enhancements/installer/rhel8_worker_support.md
+++ b/enhancements/installer/rhel8_worker_support.md
@@ -49,11 +49,25 @@ longer-term RHEL8 support is developed.
 
 ### Non-Goals
 
-* Deploying/upgrading RHEL8 worker nodes through other mechanisms. i.e. operators
+* Deploying/upgrading RHEL8 worker nodes through other mechanisms, i.e. operators
+* Changing the deployment/upgrade process for RHEL 7 worker nodes 
 
 ## Proposal
 
-This is where we get down to the nitty gritty of what the proposal actually is.
+Currently, for RHEL 7 worker node deployment, the user must [prepare the host][prepare-rhel]
+before running the scaleup.yml playbook.  This process can be cumbersome for
+automating deployment and also requires additional steps to update repos prior
+to performing an upgrade.  For RHEL 8 hosts preparing repos on the hosts will
+not be required because the necessary packages will be delivered as an image in 
+the release package.  This removes the need of mirroring the release package as
+well as setting up repo mirrors for offline or disconnected deployments.
+
+The basic process for scaling up and upgrading RHEL worker nodes will remain the
+same.  The playbooks will be updated to download and mount the image with the RPM
+content during the install/upgrade for RHEL 8 hosts.  The install process will
+not change for RHEL 7 worker nodes.
+
+[prepare-rhel]: https://docs.openshift.com/container-platform/4.7/machine_management/adding-rhel-compute.html#rhel-preparing-node_adding-rhel-compute
 
 ### User Stories
 
--- a/enhancements/installer/rhel8_worker_support.md
+++ b/enhancements/installer/rhel8_worker_support.md
@@ -31,26 +31,27 @@ superseded-by:
 ## Summary
 
 This enhancement proposes to add RHEL 8 Server support for deploying worker/infra
-nodes with [openshift-ansible] RHEL8 nodes will be deployed and upgraded using
+nodes with [openshift-ansible] RHEL 8 nodes will be deployed and upgraded using
 the same Ansible playbook process however, RPM content will be delivered using
 an image in the release package.
 
 ## Motivation
 
-Customers desire to deploy RHEL8 nodes and using containerized content will provide
+Customers desire to deploy RHEL 8 nodes and using containerized content will provide
 a better user experience than configuring RPM repositories on nodes.  Using the
-existing process will allow near-term support of RHEL8 nodes while enhanced
-longer-term RHEL8 support is developed.
+existing process will allow near-term support of RHEL 8 nodes while enhanced
+longer-term RHEL 8 support is developed.
 
 ### Goals
 
-1. Implement an os-content image for RHEL8 RPMs
-1. Update [openshift-ansible] scaleup.yml and upgrade.yml playbooks to support RHEL8 workers
+1. Implement an os-content image for RHEL 8 RPMs
+1. Update [openshift-ansible] scaleup.yml and upgrade.yml playbooks to support RHEL 8 workers
 
 ### Non-Goals
 
-* Deploying/upgrading RHEL8 worker nodes through other mechanisms, i.e. operators
+* Deploying/upgrading RHEL 8 worker nodes through other mechanisms, i.e. operators
 * Changing the deployment/upgrade process for RHEL 7 worker nodes
+* Upgrading existing RHEL 7 worker nodes to RHEL 8
 
 ## Proposal
 
--- a/enhancements/installer/rhel8_worker_support.md
+++ b/enhancements/installer/rhel8_worker_support.md
@@ -5,7 +5,7 @@ authors:
 reviewers:
   - "@staebler"
 approvers:
-  - TBD
+  - "@derekwaynecarr"
 creation-date: 2021-05-13
 last-updated: 2021-05-13
 status: provisional
--- a/enhancements/installer/rhel8_worker_support.md
+++ b/enhancements/installer/rhel8_worker_support.md
@@ -1,0 +1,238 @@
+---
+title: rhel-8-server-worker-infra-node-support
+authors:
+  - "@mtnbikenc"
+reviewers:
+  - "@staebler"
+approvers:
+  - TBD
+creation-date: 2021-05-13
+last-updated: 2021-05-13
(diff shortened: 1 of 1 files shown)


mtnbikenc:
Epic link: https://issues.redhat.com/browse/CORS-1650

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,251 @@
...
+
+### Goals
+
+1. Implement an os-content image for RHEL8 RPMs
+1. Update [openshift-ansible] scaleup.yml and upgrade.yml playbooks to support RHEL8 workers
+
+### Non-Goals
+
+* Deploying/upgrading RHEL8 worker nodes through other mechanisms, i.e. operators
+* Changing the deployment/upgrade process for RHEL 7 worker nodes
```
staebler, line 53: What is the plan for existing RHEL 7 worker nodes to be upgraded to RHEL 8? Or would they just be replaced with new RHEL 8 nodes?

mtnbikenc: The current plan is that the customer would scale up new RHEL 8 workers and remove existing RHEL 7 workers.  We would not support a process to upgrade hosts in place.  I will add an item about not supporting OS upgrades.

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,252 @@
...
+## Motivation
+
+Customers desire to deploy RHEL 8 nodes and using containerized content will provide
+a better user experience than configuring RPM repositories on nodes.  Using the
+existing process will allow near-term support of RHEL 8 nodes while enhanced
+longer-term RHEL 8 support is developed.
+
+### Goals
+
+1. Implement an os-content image for RHEL 8 RPMs
```
sdodson, line 47: Do we know yet which teams will do this work? If it's coming from teams other than Installer can we make sure they have a representative added as a reviewer?

sdodson (approved):
looks fine to me, just want to make sure everyone involved in the work product gets a chance to review

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Proposal
+
+Currently, for RHEL 7 worker node deployment, the user must [prepare the host][prepare-rhel]
+before running the scaleup.yml playbook.  This process can be cumbersome for
+automating deployment and also requires additional steps to update repos prior
+to performing an upgrade.  For RHEL 8 hosts preparing repos on the hosts will
+not be required because the necessary packages will be delivered as an image in
+the release package.  This removes the need of mirroring the release package as
+well as setting up repo mirrors for offline or disconnected deployments.
```
cgwalters, line 64: It's rather surprising that the text in this proposal only mentions "RHCOS" once in a later question.   Because obviously this model is a whole big advantage of what we've implemented with RHCOS today.

Also I'd really like to reinforce the story that RHCOS *is* RHEL, it's just managed by the MCO.  So we could phrase this as something more like "Support traditional RHEL8 workers, except align more with the way RHEL CoreOS works by shipping OS updates in a container" or so?

cgwalters: Also there's an important flip side to this - the reason one might want to use traditional RHEL8 is *precisely* because one has existing management tools in place for it.  For example, one might be using Satellite "content views" which have existed for RHEL for a long time.  How would this model of shipping RPMs-in-containers interact with these types of external tooling?   Wouldn't these users want tooling that supported e.g. extracting the bits from this RPM and importing it to Satellite?

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,252 @@
...
+
+How will security be reviewed and by whom? How will UX be reviewed and by whom?
+
+Consider including folks that also work outside your immediate sub-project.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. How will this be implemented for OKD?  Is this a concern?
```
ashcrow, line 101: I don't believe that this has any affect on OKD as the base is FCOS.

enhancements/installer/rhel8_worker_support.md
sferich888: Where (on what hosts) / how will a customer use this image? Will they be expected to extract it from the image and host a repo on a host (if so which one); or would this image run on the cluster and offer the RPM content to the cluster (and this is all automated)?

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,252 @@
...
+
+The basic process for scaling up and upgrading RHEL worker nodes will remain the
+same.  The playbooks will be updated to download and mount the image with the RPM
+content during the install/upgrade for RHEL 8 hosts.  The install process will
+not change for RHEL 7 worker nodes.
+
+[prepare-rhel]: https://docs.openshift.com/container-platform/4.7/machine_management/adding-rhel-compute.html#rhel-preparing-node_adding-rhel-compute
+
+### User Stories
+
```
sferich888, line 74: Are we missing a user story or two? 

As an OpenShift admin I want to setup my 'isolated' infrastructure for RHEL 8 hosts; so that they can update (and/or be managed). 

As an OpenShift admin I want to update the repo-content in my 'isolated' infrastructure (network).

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,252 @@
...
+1. As an OpenShift admin using [openshift-ansible], I want to deploy and upgrade
+   RHEL 8 Server worker nodes without needing to prepare hosts for installation.
+
+1. As an Openshift admin, I want to be able to deploy RHEL 8 workers without
+   having to mirror RPM content.
```
sferich888, line 79: Should we just integrate better with Satellite of this?

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,252 @@
...
+## Summary
+
+This enhancement proposes to add RHEL 8 Server support for deploying worker/infra
+nodes with [openshift-ansible] RHEL 8 nodes will be deployed and upgraded using
+the same Ansible playbook process however, RPM content will be delivered using
+an image in the release package.
+
+## Motivation
+
+Customers desire to deploy RHEL 8 nodes and using containerized content will provide
```
miabbott, line 40: Is the use case for deploying RHEL 8 nodes from scratch?  Or would customers get the "base RHEL 8" content onto a host and then install the OCP specific RPMS(i.e.`cri-o`, `openshift-hyperkube`, etc)?

mtnbikenc: The intent is that we are only installing the packages necessary for OCP from the container.  This alleviates the need of switching repo configuration for OCP upgrades between .y releases.  The rest of the OS content would come through the customer's existing OS management processes.

enhancements/installer/rhel8_worker_support.md
mtnbikenc: We need to better understand the customer's pain points with repo management.

enhancements/installer/rhel8_worker_support.md
mtnbikenc: The content would come from the release payload and would not be something the user would need to update or manage, other than mirroring the release payload for offline installs.

enhancements/installer/rhel8_worker_support.md
mtnbikenc: I will update this question based on this response.

enhancements/installer/rhel8_worker_support.md
mtnbikenc: I will include more verbiage about 'traditional' RHEL 8 support.

mtnbikenc: > Where (on what hosts) / how will a customer use this image? Will they be expected to extract it from the image and host a repo on a host (if so which one); or would this image run on the cluster and offer the RPM content to the cluster (and this is all automated)?

The customer would not interact directly with the image.  The openshift-ansible playbooks would orchestrate the process of pulling the image, extracting or mounting the content, and then installing the content.

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,252 @@
...
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement proposes to add RHEL 8 Server support for deploying worker/infra
+nodes with [openshift-ansible] RHEL 8 nodes will be deployed and upgraded using
+the same Ansible playbook process however, RPM content will be delivered using
```
dhellmann, line 35: The same process as what? RHEL 7? Is there a reference for that for those of us coming in between its implementation and this enhancement?

enhancements/installer/rhel8_worker_support.md
dhellmann, line 64: There was some discussion of the image size in the architecture call today. Another reason we might not want to add this new image to the release payload is that it wouldn't be used for every customer or cluster (potentially not even "a lot" of them), and as a general principle we try to only put things into the payload that are very widely used.

If the hurdle is really setting up or mirroring the RPM repo, can we do something to make that easier in a way that doesn't affect every cluster? Could we script it? Or distribute the image described here outside of the release payload in a way that makes it easy to consume in clusters where it's needed?

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,252 @@
...
+### Goals
+
+1. Implement an os-content image for RHEL 8 RPMs
+1. Update [openshift-ansible] scaleup.yml and upgrade.yml playbooks to support RHEL 8 workers
+
+### Non-Goals
+
+* Deploying/upgrading RHEL 8 worker nodes through other mechanisms, i.e. operators
+* Changing the deployment/upgrade process for RHEL 7 worker nodes
+* Upgrading existing RHEL 7 worker nodes to RHEL 8
```
dhellmann, line 54: Does that mean if a user wanted to reuse a host they would have to deploy RHEL 8 from scratch, or just that we are not automating the upgrade from 7 to 8?

enhancements/installer/rhel8_worker_support.md
yuqi-zhang, line 35: Could you also add to this explicit MCO requirements, if that's relevant?

enhancements/installer/rhel8_worker_support.md
sdodson: `Customers desire to add existing subscribed RHEL 8 hosts to an OpenShift 4.9+ cluster as workers.` or something along those lines to address Micah's question, subscribed being important part because we need to be able to assume base channels are enabled to fulfill any deps.

enhancements/installer/rhel8_worker_support.md
sdodson: I'd go ahead and mention that they should be re-provisioned, we're relatively light touch but I don't think we want to attempt any sort of upgrade. If they want a fully managed solution, for now, they should be using RHCOS.

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Proposal
+
+Currently, for RHEL 7 worker node deployment, the user must [prepare the host][prepare-rhel]
+before running the scaleup.yml playbook.  This process can be cumbersome for
+automating deployment and also requires additional steps to update repos prior
+to performing an upgrade.  For RHEL 8 hosts preparing repos on the hosts will
```
sdodson, line 61:
```suggestion
to performing an upgrade.  For RHEL 8 hosts preparing OpenShift repos on the hosts will
```

To clarify that we're speaking only of OpenShift specific repos, the hosts should be subscribed and have base channels enabled which is outside the scope of what we're managing.

enhancements/installer/rhel8_worker_support.md
sdodson: I don't think we want to talk about Satellite, avoiding the OCP repo management is a specific goal of this enhancement and given they've already got to mirror the release image it seems like we shouldn't concern ourselves with that in this enhancement.

enhancements/installer/rhel8_worker_support.md
sdodson: Should we mention that traditional Fedora workers are specifically out of scope?

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,252 @@
...
+How will security be reviewed and by whom? How will UX be reviewed and by whom?
+
+Consider including folks that also work outside your immediate sub-project.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. How will this be implemented for OKD?  Is this a concern?
+1. Can we use existing os-content used for RHCOS to install required RPMs?
```
sdodson, line 102:
```suggestion
1. Can we use existing os-content used for RHCOS to install required RPMs or files?
```
Given the discussion during the arch call where it was mentioned perhaps we just extract raw files out of the os-content image.

enhancements/installer/rhel8_worker_support.md
mtnbikenc: [openshift-ansible](https://github.com/openshift/openshift-ansible) which is linked in the previous sentence is the 'Ansible playbook process' being referred to.  I will add a link to the openshift docs for adding RHEL hosts.
https://docs.openshift.com/container-platform/4.7/machine_management/adding-rhel-compute.html

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,262 @@
...
+
+1. As an OpenShift admin using [openshift-ansible], I want to deploy and upgrade
+   traditional RHEL 8 Server worker nodes without needing to configure OpenShift repos for installation.
+
+1. As an Openshift admin, I want to be able to deploy traditional RHEL 8 workers without
+   having to mirror OpenShift RPM content.
+
+### Open Questions [optional]
+
+1. Can we use existing os-content used for RHCOS to install required RPMs or files?
```
ashcrow, line 87: I think the key here is more so matching the versions of foundational OCP components (`cri-o`, `kubelet`, etc..). Since traditional RHEL workers are, by definition managed by the system administrators directly there could easily be additional important packages required by the user. I've not thought through this deeply but I could see this as a bit of a wrinkle.

yuqi-zhang: RHEL7 nodes today don't quite match the same version of RHCOS packages, but for RHEL8 it may make sense to ensure the e.g. kubelet packages match between RHCOS and RHEL8 nodes?

mtnbikenc:
Updated.

Remove references to using an os-content image

After investigation of using an image to deliver RPM content for RHEL8
workers, it was found to be a sub-optimal approach for limited user
experience improvement.

dhellmann:
> Updated.
> 
> Remove references to using an os-content image
> 
> After investigation of using an image to deliver RPM content for RHEL8
> workers, it was found to be a sub-optimal approach for limited user
> experience improvement.

It would be good to capture this detail in the Alternatives section.

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,244 @@
...
+not change for RHEL 7 worker nodes.
+
+To migrate from RHEL 7 worker nodes, new RHEL 8 nodes should be provisioned,
+and the existing RHEl 7 worker nodes removed.
+
+
+### User Stories
+
+1. As an OpenShift admin using [openshift-ansible], I want to deploy and upgrade
+   traditional RHEL 8 Server worker nodes.
```
dhellmann, line 67: What platforms do customers tend to want this for? All? Metal? Specific clouds? It probably doesn't matter in terms of this doc, but it helps add context.

enhancements/installer/rhel8_worker_support.md
```diff
@@ -0,0 +1,244 @@
...
+
+* Deploying/upgrading traditional RHEL 8 worker nodes through other mechanisms, i.e. operators
+* Changing the deployment/upgrade process for RHEL 7 worker nodes
+* Upgrading existing RHEL 7 worker nodes to RHEL 8
+* OKD and traditional Fedora worker support
+
+## Proposal
+
+The basic process for scaling up and upgrading RHEL worker nodes will remain the
+same.  The playbooks will be updated to install the appropriate RPMs based on
```
dhellmann, line 56: Is there a link to the docs for the process we're matching?

PR #784, opened 2021-05-17
Allow installer to include/exclude components based on user select install solution

bparees:


bparees:
/assign @derekwaynecarr 
cc @ecordell @staebler

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+## Summary
+
+This enhancement proposes a mechanism for cluster installers to exclude one or more optional components for
+their installation which will determine which payload components are/are not installed in their cluster.
+Core components are defined as the set of Second Level Operators managed by the Cluster Version Operator
+which today cannot be disabled until after completing the install and editing a CVO override, or editing
+the CVO overrides as part of rendering+editing manifests.
+
+The proposed UX is to make this a first class part of the install config api with the implementation
+being arguments supplied to the CVO to filter the user-selected manifests.
```
staebler, line 35: I think there is more involved in the installer implementation of this than simply passing along arguments to the CVO.
1. Presumably the installer would need to know which addons can be excluded so that the installer can do validation on the user input.
2. Some addons may require the installer to omit manifests that it would normally create. For example, if the baremetal-operator addons were omitted, then the installer would need to omit the [manifest that configures the metal provisioning](https://github.com/openshift/installer/blob/master/data/data/manifests/openshift/baremetal-provisioning-config.yaml.template).
3. There would probably be an expectation by the user that the installer would make some opinionated decisions on behalf of the user about which addons to exclude. Again with the baremetal-operator addon, my expectation as a user would be that the installer would configure the exclusions automatically so that the baremetal-operator addon was excluded when using a platform other than "baremetal".

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Motivation
+
+There is an increasing desire to move away from "one size fits all" cluster installations, and
+towards flexibility about what should/should not exist in a new cluster out of the box.  This can
+be seen in efforts such as hypershift, single node, and cloud-ready-containers.  Each of these
```
romfreiman, line 41: nit: code-ready-containers

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+
+The CVO already has the ability to respect annotations on resources, as can be seen
+[here](https://github.com/openshift/cluster-kube-apiserver-operator/blob/c03c9edf5fddf4e3fb1bc6d7afcd2a2284ca03d8/manifests/0000_20_kube-apiserver-operator_06_deployment.yaml#L10) and leveraged [here](https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/assets/cluster-version-operator/cluster-version-operator-deployment.yaml#L47-L48).
+This proposal consists of two parts:
+
+1) Formalizing a concept of an "addon" annotation which allows a given resource to be excluded based
+on installer input. For example the console deployment resource could be annotated as
+
+```yaml
+annotations:
```
romfreiman, line 83: just to verify, you're expecting ALL console related manifests to have this annotation, right?
If so, I would state it more clearly here - that it's not a single manifest.

bparees: yeah. will clarify, thanks.

enhancements/installer/component-selection.md
bparees: ack

enhancements/installer/component-selection.md
bparees: 1) maybe?  specifying things to exclude that don't actually exclude anything isn't a great UX but i dunno that it's a requirement that we tell you your exclusion didn't match anything.

2) not sure i follow this one.  I'd expect all related manifests to be annotated such that they get excluded as a group.  Could the manifest that configures the metal provisioning not also be annotated?   Is it instantiated by the CVO or by the installer itself?

3) i'd classify that as a potential future goal/possibility, but not a requirement here

romfreiman: on the same topic - will the user be able to say exclude bm-operator but install bm platform? If not, how the installer will know to block it?

bparees: I tried to cover these sort of scenarios here:
https://github.com/openshift/enhancements/pull/784/files#diff-4b412132f6f354bd03403cc85e104b445d5ff581abebae3d0b6fb69c58f094e7R120-R122

so we can discuss if that's "sufficient/acceptable" or not, but the position of this EP was to punt on such things.

romfreiman: `maybe? specifying things to exclude that don't actually exclude anything isn't a great UX but i dunno that it's a requirement that we tell you your exclusion didn't match anything.
`
I think it will be a bad experience if the user will have a typo. Maybe, we can find a way for the CVO to verify it during bootstrap and fail early. kinda cvo-render --verify-components

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+an addon that was excluded in the original install.  The new resources should be
+created.  The new resources should *not* be created.
+
+4) After installing a cluster, change the set of addons that are included/excluded.
+Newly included addon resources should be created, newly excluded ones should be
+deleted.  (Do we want to support this?)
+
+
+
+### Graduation Criteria
```
romfreiman, line 171: maybe it worth adding a section how do we mark those clusters in telemetry and in general (some configmap? Infrastructue CR? ) . So it will be clear, both for trends and support how they were installed.

enhancements/installer/component-selection.md
staebler: >     2. not sure i follow this one.  I'd expect all related manifests to be annotated such that they get excluded as a group.  Could the manifest that configures the metal provisioning not also be annotated?   Is it instantiated by the CVO or by the installer itself?

It is ultimately added to the cluster via CVO. So the manifest could have the annotation. The installer would need to add that annotation since the installer creates the manifest.

enhancements/installer/component-selection.md
cfergeau: 'CodeReady Containers' would be even better :)

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+1. Do we want to constrain this functionality to turning off individual components?  We could
+also use it to
+  a) turn on/off groups of components as defined by "solutions" (e.g. a "headless" solution
+  which might turn off the console but also some other components).  This is what CLUSTER_PROFILES
+  sort of enable, but there seems to be reluctance to expand the cluster profile use case to include
+  these sorts of things.
+  b) enable/disable specific configurations such as "HA", where components could contribute multiple
+  deployment definitions for different configurations and then the installer/CVO would select the correct
+  one based on the chosen install configuration (HA vs single node) instead of having components read/reconcile
+  the infrastructure resource.
```
hardys, line 139: IMO it would be useful to expand the scope to include opt-in to optional components, for example see the discussion on #747 where we'd like to make the kubernetes-nmstate operator optionally available on any platform (and probably enabled by default on some where it's commonly used, e.g baremetal)

enhancements/installer/component-selection.md
hardys: > on the same topic - will the user be able to say exclude bm-operator but install bm platform? If not, how the installer will know to block it?

Seems like that is the scope of this proposal, but as @staebler mentions it implies some additional validation in the installer, e.g taking the baremetal example - it's reasonable to do `create manifests` with the baremetal SLO disabled (perhaps to deploy via UPI or assisted where those components may not be required), but a user should not be allowed to do `create cluster` with that configuration, because deployment of any worker hosts will fail.

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+
+2. How does the admin enable a component post-install if they change their mind about what components
+they want enabled?  Do we need/want to allow this?
```
dhellmann, line 142: Could we produce OLM versions of the add-ons, and let them use OLM to install the components later if they aren't present from the start?

If we do that, could we move the add-ons out of the release payload and *always* manage them via OLM, even if it means the installer defaults to generating the OLM manifests to trigger installing the add-on?

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+
+Which resources ultimately get installed for a given cluster would be the set of resources encompassed
+by the CLUSTER_PROFILE(if any), minus any resources explicitly excluded by the excluded addons configuration.
+
+Examples of candidate components to be treated as addons:
+
+* console
+* imageregistry
+* samples
+* baremetal operator
```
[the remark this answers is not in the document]
bparees: quite honestly this suggestion came from @derekwaynecarr so i'm going to defer all the various questions people are raising about it, to him :)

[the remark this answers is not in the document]
bparees: (I mean the suggestion of baremetal operator being optional)

enhancements/installer/component-selection.md
bparees: it's plausible, but olm versions of things that are deemed "essential" brings its own set of challenges:

1) disconnected installs (you need to mirror the olm catalog also, or we need to include an olm catalog in the payload)
2) independent versioning/release cadence and ensuring version-matrix compatibility
3) install ordering (no olm component can possibly come up until OLM itself comes up).  - likely not a concern for things we have deemed "Addons" but just noting it.

these are definitely things we're going to be starting to explore (looks like cert-manager may be the first candidate) but at least right now it raises more questions than it answers, for me.

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+1) Install clusters w/ the various add-on components included/excluded and confirm
+that the cluster is functional but only running the expected add-ons.
+
+2) Upgrade a cluster to a new version that includes new resources that belong to
+an addon that was included in the original install.  The new resources should be
+created.
+
+3) Upgrade a cluster to a new version that includes new resources that belong to
+an addon that was excluded in the original install.  The new resources should be
+created.  The new resources should *not* be created.
```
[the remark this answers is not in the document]
bparees: yup, thanks

enhancements/installer/component-selection.md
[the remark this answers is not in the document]
bparees: (yes, apparently that's what i meant. updating)

enhancements/installer/component-selection.md
bparees: for now i want to constrain this to "how do we make the set of things the CVO currently installs be dynamic instead of static"

mechanisms for having OLM-based components be a first class part of an install experience is going to be a follow-on item.

enhancements/installer/component-selection.md
bparees: good point, will add wording about this

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Open Questions
+
+
+1. Do we want to constrain this functionality to turning off individual components?  We could
+also use it to
+  a) turn on/off groups of components as defined by "solutions" (e.g. a "headless" solution
+  which might turn off the console but also some other components).  This is what CLUSTER_PROFILES
+  sort of enable, but there seems to be reluctance to expand the cluster profile use case to include
+  these sorts of things.
```
dhellmann, line 135: Part of the issue with profiles is managing the proliferation of definitions of groupings of individual components as we try to predict what users will or won't prioritize including in their clusters. We may have the same challenge with defining solutions or configurations. Starting with the low level components individually will give users complete control, and we can layer some sort of grouping concept on top later if we find it necessary.

enhancements/installer/component-selection.md
dhellmann: Perhaps a better post-install solution is to just let the user modify the CVO resource that records the list of skipped components? Then when the CVO reconciles, it would (un)install components based on the change.

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+
+2. How does the admin enable a component post-install if they change their mind about what components
+they want enabled?  Do we need/want to allow this?
+
+3. What are the implications for upgrades if a future upgrade would add a component which would have
+been filtered out during install time?  The install time choices need to be stored somewhere in
+the cluster and used to filter applied resources during upgrades also.  My understanding is today
+this is handled with CLUSTER_PROFILES and EXCLUDE_ANNOTATIONS by setting the env vars on the CVO
+pod, but if we want to allow the set to be changed (see (2), we need a more first class config
+resource that is admin editable)
```
dhellmann, line 149: How would a new component have been filtered out if the user didn't know to do so because it didn't exist?

enhancements/installer/component-selection.md
```diff
@@ -0,0 +1,246 @@
...
+an addon that was included in the original install.  The new resources should be
+created.
+
+3) Upgrade a cluster to a new version that includes new resources that belong to
+an addon that was excluded in the original install.  The new resources should *not* be
+created.
+
+4) After installing a cluster, change the set of addons that are included/excluded.
+Newly included addon resources should be created, newly excluded ones should be
+deleted.  (Do we want to support this?)
```
dhellmann, line 167: Does it cost us anything extra to support it, other than the testing effort?

enhancements/installer/component-selection.md
dhellmann: We have more recent requirements to be able to install the metal tools on platforms other than `baremetal`, so that's probably no longer a good example of something we would expect the installer to exclude automatically.

That said, I think we can ignore the automated guidance angle, at least in a first iteration. We can make it fancier/smarter later, but the first version should have simple rules: The default guidance is to install everything. If you want less than everything, you tell the CVO and it leaves out the things you don't want.

enhancements/installer/component-selection.md
bparees: at this point the EP is directed towards this approach, basically grouping resources at the second-level-operator level (technically people can annotate the resources however they want, but the expected usage will be to annotate component-related resources with a common name that the admin can then opt in/out of).

slightly more complicated is the case where 2 or more components have interdependencies and what we do if the admin only disables one of them.  I'm not sure we've identified any examples of this that need to be disable-able in the first pass, though.

enhancements/installer/component-selection.md
bparees: because the new component (or new resource for an existing component) uses the same "component name" that was part of the initial filter list during install.

enhancements/installer/component-selection.md
bparees: allowing previously excluded components to be added to the cluster is relatively easy.

allowing previously included components to be removed is hard, per the other comment thread (CVO doesn't know how to clean stuff up today, and can't be sure it knows everything an operator may have done)

And allowing it at all means we need a user editable config resource that the CVO watches to update its filter list.  If you can't change it after install time, you don't need any of that.

bparees:
new PR: 
https://github.com/openshift/enhancements/pull/922

Issue #783, opened 2021-05-17
template-lint.sh can't handle committed file renames


sgreene570:
I ran into some issues with `make lint` when trying to relocate an enhancement to a different top level folder.

Here is a trivial reproducer that is entirely a hypothetical example:

```
mv enhancements/builds/volume-mounted-resources.md enhancements/builds/test-name.md
git add .
git commit -m "test"
make lint
```

And you are presented with the following error:
```
+ /workdir/hack/template-lint.sh
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
grep: enhancements/builds/volume-mounted-resources.md: No such file or directory
enhancements/builds/volume-mounted-resources.md missing "## Release Signoff Checklist"
enhancements/builds/volume-mounted-resources.md missing "## Summary"
enhancements/builds/volume-mounted-resources.md missing "## Motivation"
enhancements/builds/volume-mounted-resources.md missing "### Goals"
enhancements/builds/volume-mounted-resources.md missing "### Non-Goals"
enhancements/builds/volume-mounted-resources.md missing "## Proposal"
enhancements/builds/volume-mounted-resources.md missing "### User Stories"
[...]

PR #785, opened 2021-05-17
tools: update template-lint.sh to ignore missing files

dhellmann:
Under some conditions git log reports a deleted or renamed file as
changed. If the file isn't present, we shouldn't check it.

Addresses #783


wking (approved):
/lgtm

PR #787, opened 2021-05-20
WIP: Add no drain upgrade filter

michaelgugino:
See https://github.com/openshift/machine-config-operator/pull/2582 for a POC flow in the MCD

### DIFF
--- a/enhancements/update/skip-drain-on-upgrade.md
+++ b/enhancements/update/skip-drain-on-upgrade.md
@@ -1,0 +1,150 @@
+---
+title: skip-drain-on-upgrade
+authors:
+  - "@michaelgugino"
+reviewers:
+  -
+approvers:
+  -
+creation-date: 2021-05-20
+last-updated: 2021-05-20
+status: WIP
+---
+
+# Skip drain on upgrade for certain pods
+
+Define an API that allows users/OpenShift developers/administrators to opt-out
+pods from being drained on upgrade.
+
+This is particularly relevant for stateful workloads.  Today, an upgrade
+triggers a drain of each node, thereby evicting pods running on that node,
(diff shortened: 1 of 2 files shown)


enhancements/update/skip-drain-on-upgrade.md
```diff
@@ -0,0 +1,150 @@
...
+period (say, 15 minutes).
+
+# Avoid application outages when not draining pods
+
+The point of draining is two fold.  First, it's to cleanly shutdown workloads
+so they can be rescheduled elsewhere.  Second, it's to validate PDBs for each
+pod and ensure we don't cause an application to lose too many replicas at once.
+
+If we're not evicting a pod, we need to come up with some other alternative
+to prevent a total application outage.
```
dhellmann, line 37: If we had an annotation on Pods that means "during drains, do not delete me" but the upgrade logic honored the PDB by not rebooting the node (or maybe not even starting to drain it), would that address both goals?

michaelgugino: We need to reboot the nodes to utilize the new ostree.  Also, there's no mechanism for restarting services instead of rebooting.

So, I think we ultimately do want to reboot the node and drain other pods, but this is about not deleting a particular pod for 'expensive to recreate a particular pod' purposes.  EG, you have some stateful app with local storage (possibly using an attached NVMe or similar), you don't want to shuffle that data, you just want to restart a node and have the pod restarted.

dhellmann: Right. For single-node, my understanding is that MCO skips the drain for all pods and just reboots the node (drain wouldn't work in that case, because there would be no other node to receive the pod). I wonder if we could reuse that behavior here, but more selectively by marking the Pods that should not be drained before a reboot.

PR #788, opened 2021-05-20
CONVENTIONS.md: add PromQL query to calculate adequate memory request

fpetkovski:
This commit adds a Prometheus query which can be used to calculate the difference between the requested memory of each container, and the actual memory which should be requested according to memory request convention.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -286,6 +286,22 @@ end-to-end job run and the component being tuned uses 350m on the same
 run, the request for the component being tuned should be set to
 100m/600m * 350m ~= 58m.
 
+The following PromQL query illustrates how to calculate the adequate CPU request
+for all containers in the `openshift-monitoring` namespace:
+
+```PromQL
+# CPU request / usage of the SDN container
+scalar(
+  max(kube_pod_container_resource_requests{namespace="openshift-sdn", container="sdn", resource="cpu"}) /
+  max(max_over_time(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{namespace="openshift-etcd", container="etcd"}[120m]))) * 
+# CPU usage of each container in the openshift-monitoring namespace
+max by (pod, container) (node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{namespace="openshift-monitoring"})
+```
+
+> Please note that pods which run on control-plane nodes must use the etcd container as their baseline.
+The example above uses the SDN container for all pods in the `openshift-monitoring` namespace in order to simplify the
+Prometheus query. Since the `cluster-monitoring-operator` runs on control plane nodes, its CPU request should be evaluated against the `etcd` container.
+
 Kubernetes resource requests for memory are [measured in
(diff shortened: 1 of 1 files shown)


CONVENTIONS.md
```diff
@@ -306,6 +306,18 @@ the Prometheus instance in the cluster to query the
 `kube_pod_resource_request` and `kube_pod_resource_limit` metrics and
```
simonpasquier, line 306: should we remove `kube_pod_resource_limit` since it is stated before that cluster components shouldn't have resource limits?

simonpasquier (commented):
It might be handy to include the PromQL query to compare CPU requests with actual usage.

simonpasquier:
/assign @dhellmann 

Assigning Doug since he updated this section recently.

CONVENTIONS.md
dhellmann: We could do that in another PR, if we're sure nothing actually does set limits.

dhellmann (commented):
This looks good to me. I'm not a PromQL expert, so I'll leave it to @simonpasquier to sign off on the details. It would be cool if we could show some sample output, but if that's a graph it probably wouldn't be easy to do here so it's not a big deal.

Thanks!

/approve

CONVENTIONS.md
```diff
@@ -286,6 +286,18 @@ end-to-end job run and the component being tuned uses 350m on the same
 run, the request for the component being tuned should be set to
 100m/600m * 350m ~= 58m.
 
+The following PromQL query illustrates how to calculate the adequate CPU request
+for all containers in the `openshift-monitoring` namespace:
```
simonpasquier, line 290: the query mixes the openshift-etcd and openshift-monitoring namespaces.

fpetkovski: The formula for calculating the adequate request depends on the etcd usage and request. This is why we have both the target namespace (openshift-monitoring in this case) and the reference namespace (openshift-etcd). 

Does this resolve the question, or did you have something else in mind>

simonpasquier: Ah right I should have read the paragraph above indeed :)

Having said that all monitoring components (except CMO which runs on control plane nodes) should use SDN as the baseline.

fpetkovski: Ack. The query should be updated to use SDN as the baseline. I also added a note that pods which run on control-plane nods should be evaluated against etcd.

CONVENTIONS.md
```diff
@@ -286,6 +286,22 @@ end-to-end job run and the component being tuned uses 350m on the same
...
+
+```PromQL
+# CPU request / usage of the SDN container
+scalar(
+  max(kube_pod_container_resource_requests{namespace="openshift-sdn", container="sdn", resource="cpu"}) /
+  max(max_over_time(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{namespace="openshift-etcd", container="etcd"}[120m]))) *
```
simonpasquier, line 296:
```suggestion
  max(max_over_time(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{namespace="openshift-sdn", container="sdn"}[120m]))) * 
```

PR #791, opened 2021-05-24
Support RAID and BIOS configuration for baremetal IPI deployments

hs0210:
Add the support of RAID and BIOS configuration for baremetal IPI deployments.

Signed-off-by: Hu Shuai <hus.fnst@cn.fujitsu.com>


enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Motivation
+
+With the new feature of [RAID configuration](https://github.com/metal3-io/baremetal-operator/pull/292) and the upcoming feature of [BIOS configuration](https://github.com/metal3-io/baremetal-operator/pull/302), openshift already has or will have the ability to configure RAID and BIOS for baremetal.
+
+However, these features only work with a running cluster, users cannot configure desired RAID and BIOS during IPI deployments.
+
+### Goals
+
+- Allow users to configure RAID and BIOS in ***install-config.yaml***.
```
dhellmann, line 46: We try not to list the implementation being proposed as a goal, since we might receive feedback that leads to a different implementation. Here I think it would be more accurate to say that we want to allow users to manage the RAID and BIOS settings for control plane hosts during installation, since worker nodes are already managed by the BMO (even when creating the cluster).

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,133 @@
...
+          address: IPAddress
+          username: UserName
+          password: PassWord
+        bootMACAddress: MACAddress
+        raid: RAIDConfig
+        firmware: BIOSConfig
+```
+
+RAID feature has been implemented in **BMO**, so the *raid* field here is the same as the [BMH](https://github.com/metal3-io/baremetal-operator/blob/399f5ef7ee3831014c1425250bc4fa49641a8709/config/crd/bases/metal3.io_baremetalhosts.yaml).
+The *firmware* field is the same as the *spec.firmware* field in **BMH** which is been advancing by [#302](https://github.com/metal3-io/baremetal-operator/pull/302).
```
dhellmann, line 89: We typically try to keep the API of install-config small, and limited to things that almost every deployment needs to manage. I don't think that applies to RAID or BIOS settings, even for bare metal deployments.

The installer runs in multiple phases. Typically a user will run `create cluster` and get all of the phases, but they can also run `create manifests` and `create cluster` separately. Between those steps, they can add and modify the manifests used to create the cluster.

Is it possible to pass the RAID and BIOS settings this way, instead of through install-config? The user would run `create manifests` and then add or replace the `BareMetalHost` manifests created by the installer. The `create manifests` step would read the new files and the interface to terraform in the installer code would need to understand the BIOS and RAID fields.

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Process the fields in installer
+
+For master nodes, call the `BuildTargetRAIDCfg` method in **BMO** to process the *raid* field into *target_raid_config*, and finally write *target_raid_config* into the ***terraform.baremetal.auto.tfvars.json*** file.
```
dhellmann, line 93: If we're going to import that code from the BMO into the installer, we will want to ensure that it is in a package with minimal dependencies (rather than in the controller implementation, for example). How is the code organized in the BMO today? Will we need to rearrange anything?

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,133 @@
...
+### Risks and Mitigations
+
+TBD
+
+## Design Details
+
+### Test Plan
+
+- Unit tests for determining the configuration of RAID and BIOS passed to Ironic meeting expectations.
+- e2e tests for determining the configuration of RAID and BIOS configured during the IPI deployments.
```
dhellmann, line 111: CI for OpenShift uses VMs to simulate bare metal. Do we have an emulator for testing the BIOS API? Can we test RAID configuration using VMs?

dhellmann (commented):
/ok-to-test

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Thanks for your review. Done.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: > Is it possible to pass the RAID and BIOS settings this way, instead of through install-config?

Yes, it is possible. We used this way as an [alternative](https://github.com/openshift/enhancements/pull/791/files#diff-b06bf476cab4c460016a94b59469061638772af0097367a394ad9a23fee21c83R123-R125) before. 

> The `create manifests` step would read the new files and the interface to terraform in the installer code would need to understand the BIOS and RAID fields.

I think you may mean the `create cluster` step here. We would modify the installer code to make the interface to terraform understand the BIOS and RAID fields.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: > How is the code organized in the BMO today?

Now the raid related code is in the ironic package.

> Will we need to rearrange anything?

It may be better to create a new package for raid related code for reducing dependency.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: > Do we have an emulator for testing the BIOS API?

No, currently we don't have an emulator for testing the BIOS API.

> Can we test RAID configuration using VMs?

Yes, we can use VMs to test software RAID.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hardys: Modifying the manifests seems like a reasonable compromise, provided we can consume that data via terraform in a reasonably clean way in the installer code.

One thing to mention is that the BMH resources will still get persisted in the cluster, so the modified master BMH resources will have the BIOS/RAID config, but also be `externallyProvisioned: true`

Provided the BMO can handle that, it's probably OK, but I wanted to note it's a corner case since BMH resources probably won't normally be defined that way.

hs0210: @hardys Thanks for your review.

> the modified master BMH resources will have the BIOS/RAID config, but also be externallyProvisioned: true

This is indeed what you said. We plan to add this note to the proposal.

enhancements/baremetal/baremetal-config-raid-and-bios.md
dtantsur: https://opendev.org/openstack/sushy-tools has rudimentary BIOS support, we can use it for some basic testing probably.

enhancements/baremetal/baremetal-config-raid-and-bios.md
dtantsur: +1 to using BMH API for BIOS, especially since we're going to have 2 slightly different approaches to BIOS settings.

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,149 @@
...
+RAID feature has been implemented in **BMO**, so the *raid* field here is the same as the [BMH](https://github.com/metal3-io/baremetal-operator/blob/399f5ef7ee3831014c1425250bc4fa49641a8709/config/crd/bases/metal3.io_baremetalhosts.yaml).
+The *firmware* field is the same as the *spec.firmware* field in **BMH** which is been advancing by [#302](https://github.com/metal3-io/baremetal-operator/pull/302).
+
+#### Process the fields in installer
+
+For control plane hosts, call the `BuildTargetRAIDCfg` method in **BMO** to process the *raid* field into *target_raid_config*, and finally write *target_raid_config* into the ***terraform.baremetal.auto.tfvars.json*** file.
+
+#### Process the fields in terraform-provider-ironic
+
+Add *target_raid_config* and *bios_settings* fields to terraform-provider-ironic API,
```
dtantsur, line 93: Let's omit the `target_` prefix maybe? Terraform is declarative, `target_` fields are part of the ironic state-driven approach.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: > https://opendev.org/openstack/sushy-tools has rudimentary BIOS support, we can use it for some basic testing probably.

@dtantsur Thanks for your advice.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Done.

hs0210:
@hardys @dtantsur I [updated the implementation details](https://github.com/openshift/enhancements/pull/791/commits/7d2baa83f1575d8fbdb808109db1e63a0ea15be3). PTAL. Because the clean steps of software RAID and hardware RAID are different, if *raid* field is processed into logicalDisks on the installer side, it will be difficult for terraform-provider-ironic to know whether it is a software RAID or a hardware RAID, so the processing of RAID and BIOS is unified to terraform -provider-ironic side.

hs0210:
@hardys @dtantsur We have completed [the PoC](https://github.com/openshift/enhancements/pull/791/commits/c049224c44fcc073b84aed2e139b0de45226da8b) and confirmed this feature is indeed feasible through e2e test. PTAL.

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,151 @@
...
+  -
+superseded-by:
+  -
+---
+
+# Config RAID and BIOS for Baremetal IPI deployments
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
```
dtantsur, line 28: I suspect you need to check it

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,151 @@
...
+configuration of RAID and BIOS for control plane hosts and then passes it to Ironic([Baremetal Operator(BMO)](https://github.com/metal3-io/baremetal-operator)
+has supported the process of RAID and BIOS for worker nodes).
+
+### User Stories
+
+With the addition of this feature, the users can configure RAID and BIOS in IPI deployments.
+
+### Implementation Details/Notes/Constraints
+
+#### Add two fields
```
dtantsur, line 69: I assume in the new version you don't add any new fields other than what has already been added to BMO?

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,151 @@
...
+
+RAID feature has been implemented in **BMO**, so the *raid* field here is the same as the [BMH](https://github.com/metal3-io/baremetal-operator/blob/399f5ef7ee3831014c1425250bc4fa49641a8709/config/crd/bases/metal3.io_baremetalhosts.yaml).
+The *firmware* field is the same as the *spec.firmware* field in **BMH** which is been advancing by [#302](https://github.com/metal3-io/baremetal-operator/pull/302).
+
+#### Process the fields in installer
+
+For control plane hosts, transform the *raid* and *firmware* fields into json respectively and then write them into the
+***terraform.baremetal.auto.tfvars.json*** file as the values of *raid_config* and *bios_settings* respectively.
+
+#### Process the fields in terraform-provider-ironic
```
dtantsur, line 92: I'm curious how it will be done. Terraform is known to be bad with nested JSON fields, you may have to pass everything as a string..

(just food for thought, not really a comment on this spec)

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,151 @@
...
+## Design Details
+
+### Test Plan
+
+- Unit tests for determining the configuration of RAID and BIOS passed to Ironic meeting expectations.
+- e2e tests for determining the configuration of RAID and BIOS configured during the IPI deployments.
+
+### Graduation Criteria
+
+TBD
```
dtantsur, line 116: You need to fill all `TBD` with something, at least with `N/A` if the item does not apply.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Sorry for the misunderstanding, yes, we did pass the RAID and BIOS to terraform-provider-ironic in [string form](https://github.com/hs0210/installer/commit/9072a656176309daca0eee595f9587df1eb3c860#diff-3b0c0219b1206253af5d430d464ea73c976c339a055510a0d9456d4dfb0419afR107-R108).

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Done.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Yes. I have updated it.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Done.

dtantsur:
/lgtm

Looks good from my perspective, thank you!

hs0210:
@hardys Could you help review this proposal?

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,149 @@
...
+
+#### Process the fields in installer
+
+For control plane hosts, transform the *raid* and *firmware* fields into string respectively and then write them into the
+***terraform.baremetal.auto.tfvars.json*** file as the values of *raid_config* and *bios_settings* respectively.
+
+#### Process the fields in terraform-provider-ironic
+
+Add *raid_config* and *bios_settings* fields to terraform-provider-ironic API, transform them back to struct format, and then call the
+methods in **BMO** to process the two fields to perform [manual cleaning](https://docs.openstack.org/ironic/latest/admin/cleaning.html#manual-cleaning).
```
hardys, line 93: This sounds like we'll make terraform-provider-ironic run BMO code, but IIUC that's not the case - we'll implement equivalent functionality in terraform-provider-ironic via gophercloud directly, right?

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,149 @@
...
+- [Implementation on openshift/installer](https://github.com/hs0210/installer/tree/ipi-support-raid-bios)
+- [Implementation on openshift-metal3/terraform-provider-ironic](https://github.com/hs0210/terraform-provider-ironic/tree/ipi-support-raid-bios)
+
+## Drawbacks
+
+This will increase the number of steps in IPI deployments and take longer.
+
+## Alternatives
+
+Add RAID and BIOS configuration by modifying the manifest files **~/clusterconfigs/openshift/99_openshift-cluster-api_hosts- \*.yaml** generated by executing `openshift-baremetal-install --dir ~/clusterconfigs create manifest`.
```
hardys, line 147: This alternative is the same as the proposal above - we could perhaps make it clearer if we say something like:

```
It is already possible to modify the manifest files generated by the installer, but currently any changes to bios or RAID config will only have any affect for workers hosts.  It is desirable to enable configuration of both master and worker hosts via the same method.
```

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,149 @@
...
+
+This approach takes advantage of already supported features, so it does not need to modify any source code, but it only works for worker nodes.
```
hardys, line 149: Another alternative is to add this data to the install-config, but we discussed it and decided modifying manifests would be a reasonable alternative, given that this is an "advanced" API, and that it's not yet supported for all vendors/drivers.

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,149 @@
...
+N/A
+
+## Implementation History
+
+- [Implementation on openshift/installer](https://github.com/hs0210/installer/tree/ipi-support-raid-bios)
+- [Implementation on openshift-metal3/terraform-provider-ironic](https://github.com/hs0210/terraform-provider-ironic/tree/ipi-support-raid-bios)
+
+## Drawbacks
+
+This will increase the number of steps in IPI deployments and take longer.
```
hardys, line 143: This will only be the case when manifests specify the bios/RAID config though, correct?

hardys:
> @hardys Could you help review this proposal?

A couple of small comments/questions but overall I think this looks good - thanks!

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Sorry, currently we did use some methods in BMO. If needed, we can implement equivalent functionality in terraform-provider-ironic via gophercloud directly.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Yes.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Thanks, done.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Yes, I have updated the Alternatives section.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: Hi! Ping! PTAL @hardys

hs0210:
@hardys @dtantsur Hi, I sent PRs to terraform-provider-ironic and installer as WIP based on this proposal, PTAL. 
- https://github.com/openshift-metal3/terraform-provider-ironic/pull/53
- https://github.com/openshift/installer/pull/5196

> At present, BMO already has related raid and bios processing methods, so our PoC runs BMO code in terraform-provider-ironic for building raid clean steps and bios settings.
pros: this can avoid reinventing the wheel.
cons: this will introduce a lot of vendor packages to terraform-provider-ironic, but we can rearrange some code organization in BMO to minimal dependencies(e.g. create a new package for raid related code).
Do you think running BMO code is acceptable? Or we have to implement this via gophercloud directly in terraform-provider-ironic?

@hardys Currently [PR53](https://github.com/openshift-metal3/terraform-provider-ironic/pull/53) using several BMO methods such as [BuildRAIDCleanSteps](https://github.com/openshift-metal3/terraform-provider-ironic/pull/53/commits/d3b42e54c51b9867467728e89cee4fde81df4c46#diff-373c79495b73be3d62528c3a124f7bdf8966d21f21e2f8ab49dfc1754bf4eec9R693) and [BuildBIOSSettings](https://github.com/openshift-metal3/terraform-provider-ironic/pull/53/commits/d3b42e54c51b9867467728e89cee4fde81df4c46#diff-373c79495b73be3d62528c3a124f7bdf8966d21f21e2f8ab49dfc1754bf4eec9R667-R675). Could you give some comments on whether we can use the BMO code in terraform-provider-ironic in the [above discussion](https://github.com/openshift/enhancements/pull/791#discussion_r696274430)?

hs0210:
Hi! Ping! Thanks.

hs0210:
Hi! Ping! Thanks.

enhancements/baremetal/baremetal-config-raid-and-bios.md
kirankt: I think implementing the logic in the terraform-provider-ironic is the way to go. The code there doesn't vendor BMO and we'd like to keep it that way.

hs0210: @kirankt Thanks for your review, but the community has [discussed this matter](https://github.com/openshift-metal3/terraform-provider-ironic/pull/53#issuecomment-917993321), and the final decision is to use BMO's existing code, but the BMO code structure will be adjusted, and [the adjustment is still under discussion](https://github.com/metal3-io/baremetal-operator/pull/1017).

kirankt: From what I read, I don't think there is consensus there yet. @stbenjam what are you thoughts on vendoring BMO into the ironic TF provider.

hs0210: @kirankt I updated the section `Process the fields in terraform-provider-ironic`, PTAL.

hs0210:
Hi, [#1017](https://github.com/metal3-io/baremetal-operator/pull/1017) has been merged, and based on this I updated the section [Process the fields in terraform-provider-ironic](https://github.com/openshift/enhancements/pull/791/commits/879a2dfe7565a23c2989d7a23b1b5e129df9f5b5). PTAL.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: @kirankt, I got a comment from @zaneb under PR53, and he suggested [import all of the bmo modules](https://github.com/openshift-metal3/terraform-provider-ironic/pull/53#discussion_r762031284) on terraform-provider-ironic side. I also think that it is fine to vendor bmo modules because it can avoid reinventing the wheel. WDYT?

zaneb: We [are already](https://github.com/openshift/installer/blob/master/go.mod#L480) in the installer so it seems to me it can't make anything worse.

enhancements/baremetal/baremetal-config-raid-and-bios.md
```diff
@@ -0,0 +1,174 @@
...
+
+This enhancement proposes to add the support of RAID and BIOS configuration for baremetal IPI deployments.
+
+## Motivation
+
+With the new features of [RAID configuration](https://github.com/metal3-io/baremetal-operator/pull/292) and [BIOS configuration](https://github.com/metal3-io/baremetal-operator/pull/302), openshift already has the ability to configure RAID and BIOS for baremetal.
+
+However, these features only work with a running cluster, users cannot configure desired RAID and BIOS during IPI deployments.
+
+### Goals
```
romfreiman, line 46: Will it be applicable for adding nodes as well?
What about ZTP support?

hs0210: > Will it be applicable for adding nodes as well?

The latest OCP version already supports configuring RAID and BIOS when adding worker nodes. And this proposal is dedicated to supporting the RAID and BIOS configuration of the master node during the IPI process.

> What about ZTP support?

This proposal have nothing to do with ZTP.

enhancements/baremetal/baremetal-config-raid-and-bios.md
hs0210: I reverted the section `Process the fields in terraform-provider-ironic` accordingly, PTAL.

hs0210:
@dtantsur @zaneb @hardys @kirankt Hi, regarding this proposal, all the existing issues have been addressed. The biggest controversy has always been about whether vendor BMO on the terraform-provider-ironic side. Following the merge of [PR1017](https://github.com/metal3-io/baremetal-operator/pull/1017) and the results of community discussions, it was decided to import all of the BMO modules. For now, this proposal has been updated to the latest accordingly. Please review. If there is anything else I need to do to advance the merge of this proposal, please let me know. I hope this proposal can be merged asap. Thanks a lot.

elfosardo (approved):

hardys:
/approve

I think this is reasonable - we're not adding any new APIs here, only wiring in data from the existing manifests so we can achieve the same configuration for controlplane and worker machines

PR #666, opened 2021-02-23
[WIP]adding new userequest audit policy 

EmilyM1:
This is the addition to our existing Audit Policies which will add a userrequest policy.
UserRequests logs the requests of system:authenticated:oauth users on the resources they request.
This addition will also include a new test suite that will ensure that any previous Audit Policy(Policies) that are added will not expose tokens, secrets, routes and some configmaps. The tests will fail and block the merging on misconfigured policies, in particular, those logging at the incorrect -level. It also deals with the changes needed to `oc must-gather` and scalability for upcoming policies.

PR #792, opened 2021-05-26
kube-apiserver/audit: Add group customRules

sttts:
This adds a `customRules` list for by-group profiles to the apiserver audit configuration:

```
kind: APIServer
apiVersion: config.openshift.io/v1
spec:
  ...
  audit:
    profile: None | Default | WriteRequestBodies | AllRequestBodies
    customRules:
    - group: system:authenticated:oauth
      profile: None |Default | WriteRequestBodies | AllRequestBodies
```

### DIFF
--- a/enhancements/kube-apiserver/audit-policy.md
+++ b/enhancements/kube-apiserver/audit-policy.md
@@ -41,8 +41,13 @@ The API is meant to enable customers with stronger audit requirements than the a
 (from _metadata_-only level, over _request payloads_ to _request and response payload_ level) of audit
 logs, accepting the increased resource consumption of the API servers.
 
-This API is intentionally **very limited in filtering events**. More advanced filtering is to be done via an external mechanism (post-filtering).
-It was proven through performance tests (compare alternatives section) that even a uniform profile for all events is acceptable with small two-digit percent overhead.
+This API is intentionally **very limited in filtering events**. More
+advanced filtering is to be done via an external mechanism
+(post-filtering), e.g. via
+[jq](https://docs.openshift.com/container-platform/4.7/security/audit-log-view.html#security-audit-log-basic-filtering_audit-log-view).
+It was proven through performance tests (compare alternatives section)
+that even a uniform profile for all events is acceptable with small
+two-digit percent overhead.
 
 In addition to the uniform policy, we allow custom profiles by groups the request user is member of.
 
@@ -529,10 +534,18 @@ apiservercontrollerset.NewAPIServerControllerSet(
 )
 ```
 
-As audit policy file is not static anymore with the addition of custom rules, this approach will not work anymore. Instead, we need
-a dynamic audit policy controller that computes the policy depending on the apiserver configuration resource and copies that as a
-`ConfigMap` into the operand namespace. Then the revision controller will notice it changing and assigns a new operand revision that
-causes a rollout. We will add the audit policy controller to the APIServer controller set in library-go:
+As the audit policy file is not static anymore with the addition of
+custom rules, this approach will not work anymore. Instead, we need a
+dynamic audit policy controller in the apiserver operators that
+
+1. computes the policy depending on the apiserver configuration
+   resource
+2. and copies that as a `ConfigMap` into the operand namespace.
+
+Then the revision controller will notice it changing and assigns a new
+operand revision that causes a rollout. We will add the audit policy
+controller to the APIServer controller set in library-go, shared by
+our three apiserver operators:
 
 ```go
 apiservercontrollerset.NewAPIServerControllerSet(
--- a/enhancements/kube-apiserver/audit-policy.md
+++ b/enhancements/kube-apiserver/audit-policy.md
@@ -529,7 +529,10 @@ apiservercontrollerset.NewAPIServerControllerSet(
 )
 ```
 
-As audit policy file is not static anymore with the addition of custom rules, this approach will not work anymore. Instead, we need a dynamic audit policy controller that computes the policy depending on the apiserver configuration resource and copies that as a `ConfigMap` into the operand namespace. Then the revision controller will notice it changing and assigns a new operand revision that causes a rollout. We will add the audit policy controller to the APIServer controller set in library-go:
+As audit policy file is not static anymore with the addition of custom rules, this approach will not work anymore. Instead, we need
+a dynamic audit policy controller that computes the policy depending on the apiserver configuration resource and copies that as a
+`ConfigMap` into the operand namespace. Then the revision controller will notice it changing and assigns a new operand revision that
+causes a rollout. We will add the audit policy controller to the APIServer controller set in library-go:
 
 ```go
 apiservercontrollerset.NewAPIServerControllerSet(
--- a/enhancements/kube-apiserver/audit-policy.md
+++ b/enhancements/kube-apiserver/audit-policy.md
@@ -40,8 +40,10 @@ The API is meant to enable customers with stronger audit requirements than the a
 (from _metadata_-only level, over _request payloads_ to _request and response payload_ level) of audit
 logs, accepting the increased resource consumption of the API servers.
 
-This API is intentionally **not about filtering events**. Filtering is to be done via an external mechanism (post-filtering).
-It was proven through performance tests (compare alternatives section) that this trade-off is acceptable with small two-digit percent overhead.
+This API is intentionally **very limited in filtering events**. More advanced filtering is to be done via an external mechanism (post-filtering).
+It was proven through performance tests (compare alternatives section) that even a uniform profile for all events is acceptable with small two-digit percent overhead.
+
+In addition to the uniform policy, we allow custom profiles by groups the request user is member of.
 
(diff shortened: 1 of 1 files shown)


EmilyM1:
The linter has been changed recently. It requires some new headers in each enhancement. 
https://github.com/openshift/enhancements/pull/666/files#diff-731ba465970a87a66f9744cfdd69a780f81235315e7ff6ea1232f7ef41ac7d7dR355 that location to add them seemed to work for me.

enhancements/kube-apiserver/audit-policy.md
```diff
@@ -40,8 +41,10 @@ The API is meant to enable customers with stronger audit requirements than the a
 (from _metadata_-only level, over _request payloads_ to _request and response payload_ level) of audit
 logs, accepting the increased resource consumption of the API servers.
 
-This API is intentionally **not about filtering events**. Filtering is to be done via an external mechanism (post-filtering).
-It was proven through performance tests (compare alternatives section) that this trade-off is acceptable with small two-digit percent overhead.
+This API is intentionally **very limited in filtering events**. More advanced filtering is to be done via an external mechanism (post-filtering).
```
EmilyM1, line 44: maybe we could explain the post filtering more, like mentioning jq options? Your call

enhancements/kube-apiserver/audit-policy.md
```diff
@@ -230,7 +252,10 @@ apiVersion: config.openshift.io/v1
 spec:
   ...
   audit:
-    profile: Default | WriteRequestBodies | AllRequestBodies
+    profile: None | Default | WriteRequestBodies | AllRequestBodies
```
EmilyM1, line 255: Do we want audit.profile listed? Is customRules.group.profile on line 258 enough? Or are lines 255 and 258 both needed?
Is None a choice we want for profiles? I don't know if that's an option right now.

enhancements/kube-apiserver/audit-policy.md
```diff
@@ -261,61 +279,294 @@ the respective file-based audit policy of
...
+			Name: "audit",
+		},
+	},
+	...,
+)
+```
+
+As audit policy file is not static anymore with the addition of custom rules, this approach will not work anymore. Instead, we need
+a dynamic audit policy controller that computes the policy depending on the apiserver configuration resource and copies that as a
+`ConfigMap` into the operand namespace. Then the revision controller will notice it changing and assigns a new operand revision that
```
EmilyM1, line 534: I could be wrong, but I thought the file was for the operand/apiserver and the ConfigMap was for the operators.

enhancements/kube-apiserver/audit-policy.md
sttts: added a link to https://docs.openshift.com/container-platform/4.7/security/audit-log-view.html#security-audit-log-basic-filtering_audit-log-view. @bergerhoffer did a great job describing that in our docs.

enhancements/kube-apiserver/audit-policy.md
sttts: `None` is new. We need it to reduce logging, e.g. for the case where logs are restricted to authenticated oauth users.

enhancements/kube-apiserver/audit-policy.md
sttts: the configmap is created by the operator in the operand namespace. The revision controller makes a copy with the `-<revision>` suffix and redeploys the apiserver with the new configmap mounted.

enhancements/kube-apiserver/audit-policy.md
k-keiichi-rh: `None` is really helpful for users who want to reduce the amount of audit logs.
But the current limitation `With None set for (2), the cluster is only under limited support, i.e. for support cases, the customer will be asked to set this to one of the other three values. We will document this in the API documentation.` prevents to use this profile. There are no users who don't want Red Hat support. So I think no users use this profile even if it's implemented. Are there any expected use cases for `None` profile? 
I understand using `None` only is under limited support. But if customers use `None` + some customRules, is it not the limited support?

enhancements/kube-apiserver/audit-policy.md
EmilyM1: That clears it up. So it's operator initiated and consumed by operand 👍

enhancements/kube-apiserver/audit-policy.md
```diff
@@ -261,61 +284,302 @@ the respective file-based audit policy of
...
+
+1. computes the policy depending on the apiserver configuration
+   resource
+2. and copies that as a `ConfigMap` into the operand namespace.
+
+Then the revision controller will notice it changing and assigns a new
+operand revision that causes a rollout. We will add the audit policy
+controller to the APIServer controller set in library-go, shared by
+our three apiserver operators:
+
```
EmilyM1, line 549: Will there be an increase in time to init the apiserver pod because we are changing from a static resource and our now dependent on the creation of one?

enhancements/kube-apiserver/audit-policy.md
```diff
@@ -261,61 +284,302 @@ the respective file-based audit policy of
...
+dynamic audit policy controller in the apiserver operators that
+
+1. computes the policy depending on the apiserver configuration
+   resource
+2. and copies that as a `ConfigMap` into the operand namespace.
+
+Then the revision controller will notice it changing and assigns a new
+operand revision that causes a rollout. We will add the audit policy
+controller to the APIServer controller set in library-go, shared by
+our three apiserver operators:
```
EmilyM1, line 548: When you say audit policy controller here, is this a reference to a new function that should be created?

enhancements/kube-apiserver/audit-policy.md
sttts: You can upgrade the cluster and we don't say no to support requests. But if we require audit logs, the user is expected to switch the profile to Default or higher. Am happy to hear a better wording for this.

Basically, the user story this should prevent: "No, we cannot increase audit logging for this production cluster because of X. So we cannot provide audit logs to support".

cc @sferich888 @Anandnatraj

enhancements/kube-apiserver/audit-policy.md
sttts: yep.

enhancements/kube-apiserver/audit-policy.md
sttts: For the apiserver point of view, nothing changes. It is still a mounted file on disk. The work is done by the operator to put the right ConfigMap in place. But this is instant, i.e. changing a kube-apiserver + redeploy (what we did before) and changing the ConfigMap + redeploy (what we want to do now) take the same amount of time.

enhancements/kube-apiserver/audit-policy.md
sttts: yes, a new, small controller in the apiserver operator set in library-go. It will continuously compute the necessary audit policy and writes it to the ConfigMap when it changes.

enhancements/kube-apiserver/audit-policy.md
EmilyM1: Oh, I see! I was worried about needing to increase timeouts, but it looks like it's a non-issue. Thanks.

enhancements/kube-apiserver/audit-policy.md
```diff
@@ -104,74 +114,86 @@ depth using the new API.
...
-We propose to add an `audit.profile` field to `apiservers.config.openshift.io/v1` which selects the
-audit policy to be deployed to all OpenShift-provided API servers in the cluster. From the beginning
-we provide the following profiles:
+We propose to add an `audit` struct to `apiservers.config.openshift.io/v1` with
+1. a `customRules` list for custom profiles matching requests by a group the requesting user is part of.
+2. a `profile` field for requests that don't match any rule in (1).
+This struct specifies the audit policy to be deployed to all OpenShift-provided API servers in the cluster.
 
-- `Default`: this is the existing default policy. `Default` is also the default value in the `APIServer` custom resource:
+The generated `audit.k8s.io/v1beta1` has a constant preamble:
```
EmilyM1, line 123: Also will need to specify that it will go to `all users` if a group isn't inputted by the user

sttts: yes. If you don't specify the group, it applies to all. By putting the all-groups rules at the end, they act as fallback when no per-group rule applies.

enhancements/kube-apiserver/audit-policy.md
k-keiichi-rh: Thank you for explanation. I misread the above sentence. I agree to this policy. For support requests, it's natural to request further information to troubleshoot customer's issues if required.

enhancements/kube-apiserver/audit-policy.md
```diff
@@ -261,61 +284,302 @@ the respective file-based audit policy of
...
+apiVersion: config.openshift.io/v1
+spec:
+  ...
+  audit:
+    profile: None | Default | WriteRequestBodies | AllRequestBodies
+    customRules:
+    - group: system:authenticated:oauth
+      profile: None |Default | WriteRequestBodies | AllRequestBodies
+```
+
```
EmilyM1, line 424: Why do we need the seperate `Profile` then also the list
- group
- profile
Kinda a repeat of the question here:
https://github.com/openshift/api/pull/930/files#r645170272

Anandnatraj:
LGTM

PR #793, opened 2021-05-26
authentication: describe group syncing for OIDC identity providers

stlaz:
This enhancement describes how group synchronization upon user login would work for OIDC-based identity providers.

cc @s-urbaniak @sttts @deads2k

### DIFF
--- a/enhancements/authentication/synchronizing-user-groups-on-login.md
+++ b/enhancements/authentication/synchronizing-user-groups-on-login.md
@@ -1,0 +1,287 @@
+---
+title: synchronizing-user-groups-on-login
+authors:
+  - "@stlaz"
+reviewers:
+  - "@sur"
+  - "@slaskawi"
+approvers:
+  - "@deads2k"
+  - "@sttts"
+creation-date: 2021-05-22
+last-updated: yyyy-mm-dd
+status: implementable
+see-also:
+replaces:
+superseded-by:
+---
+
+# Synchronizing User Groups on Login
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+OpenShift should be capable of synchronizing user groups from a 3rd party
+identity provider upon a user logging in to their OpenShift cluster.
+
+## Motivation
+
+In the real world, OpenShift is not the owner of user identities but these
+are stored in a 3rd party identity provider system. When OpenShift retrieves
+such an identity, it usually only gathers information about the user's unique
+identification in that IdP and optionally a username that's preferred for the
+user if it's different from its identifier.
+
+It is a common practice to group users in logical groups on the identity provider
+side and sometimes it would make sense to bring the groups over to the OpenShift
+cluster as well. The identity providers oftentimes allow a protocol-specific path
+to retrieve group information, which is something OpenShift should leverage in those
+cases where it is desired.
+
+### Goals
+
+- allow retrieving user's groups from a 3rd party OIDC identity provider into
+  OpenShift when such a user logs in
+
+### Non-Goals
+
+- implement group synchronization for an identity provider type other than
+  OIDC
+- keep the groups retrieved by OpenShift from a 3rd party identity provider
+  synchronized between user logins
+- for a given session, only have the user appear in the groups synchronized
+  from the identity provider associated with this session
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+I am an administrator of an OpenShift cluster that uses identities from
+my company's identity provider. The users of the identity provider are
+organized in a complex and well-though-out group structure. I would like
+OpenShift to respect the group structure so that I can assign privileges
+to the users on OpenShift side according to their group membership in the
+identity provider.
+
+### Implementation Details
+
+#### OIDC Group Retrieval
+
+Neither the OIDC core specification nor JWT RFC mention a standard claim that would serve for user group
+storage and so none of the standard scopes define accessing it. While this could work without
+a need for any further configuration, we may need a way to specify extra scopes, which can
+already be done in the oauth-server configuration by specifying `extraScopes` in the identity
+provider configuration (`oauth.config.openshift.io/cluster`).
+
+A possible follow-up implementation might include being able to
+[request specific claims](https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter)
+from the UserInfo endpoint or from an `id_token`, which would allow for a more fine-grained
+configuration to reach the group claim.
+
+#### API
+
+The `OpenIDClaims` config API structure shall be expanded by a new `Groups`
+field:
+
+```golang
+// OpenIDClaims contains a list of OpenID claims to use when authenticating with an OpenID identity provider
+type OpenIDClaims struct {
+    ...
+
+    // groups is the list of claims whose values should be converted into OpenShift groups
+    // that will include the user
+    // +optional
+    Groups []string `json:"groups,omitempty"`
+}
+```
+
+The group object (`groups.user.openshift.io.com/v1`) should be annotated with
+`oauth.openshift.io/idp.<idpname>: "synced"`
+where `<idpname>` marks the identity provider name from the configuration. It should
+also get an `oauth.openshift.io/generated: true` if it hasn't previously existed.
+
+#### oauth-server
+
+When the OIDC provider has group synchronization configured, the oauth-server
+must attempt to retrieve the group claim (or claims) from the `id_token` it
+obtains from the provider after a successful authentication, and should do
+the same for the JSON data retrieved from the OIDC's UserInfo endpoint
+should the OIDC be exposing this endpoint in its OIDC discovery data. The
+UserInfo response must be authoritative in that case.
+
+Once the server has the user's groups information, it will look up the
+OpenShift groups the user appears in, and it will either add to/remove from/create
+the necessary groups for the user. The user should only be removed from the groups
+that were created for that specific identity provider (as per the annotation
+from the [API section](#api)). If such a group ends up being empty, it should
+be removed if it contains the `oauth.openshift.io/generated: true` annotation.
+
+When a group that would otherwise be created on sync already exists, the
+oauth-server should only add the user among the group members and add an
+`oauth.openshift.io/idp.<idpname>` annotation to mark users of that IdP appear
+in that group.
+
+This implementation is not solving an issue when a user can be sourced from
+multiple identity providers when `mappingMethod: add` is configured for these
+providers. Such a user would get group memberships in all the groups from
+all matching identity providers. An alternative implementation will be
+discussed in the [Alternatives section](#alternatives).
+
+### Risks and Mitigations
+
+Given that the group synchronization only happens on logins, there is a time
+window during which the synchronization of groups between the identity provider
+and OpenShift is not guaranteed. This would typically be an issue when someone's
+group membership gets revoked.
+
+In order to force OpenShift group membership synchronization on group membership
+revocation on the identity provider side, one might:
+
+1. revoke all the OAuthAccessTokens for a given user by removing them, leveraging
+   a `userName` field selector, thus forcing the user to relogin:
+
+```bash
+       oc delete oauthaccesstoken --field-selector=userName="<username>"
+```
+
+2. remove the user from the OpenShift group manually
+
+Users can also be added to groups in the IdP, in which case they are more
+likely to willingly cause a resync by `oc login` themselves.
+
+## Design Details
+
+### Open Questions
+
+### Test Plan
+
+The authentication operator already has a way to configure Keycloak and
+communicate with its API. Use that to add test users and groups to test:
+
+1. user can still log in when groups are configured but no group
+   claims are received
+2. on login, user gets added to a previously non-existent group when
+   the group appears in the configured claims
+3. on login, user gets added to an existing group that appears in the
+   retrieved group claims
+4. on login, user gets removed from any existing group that was created
+   for the given identity provider (per annotation) when the group
+   no longer appears among those retrieved from the OIDC
+
+
+### Follow-up Work
+
+To synchronize groups in a time-based manner, an additional controller in the
+oauth-server would be necessary. OIDC does not provide a standardized group
+membership and reverse group membership lookups. That means that such a controller
+would have to cache the access tokens (and possibly refresh tokens) retrieved
+from the OIDC after a successful user login, and use them to query the UserInfo
+endpoint. This could be troublesome, however, because:
+
+1. the access token we get from the OIDC might be shortlived, if there are no
+   refresh tokens we can't synchronize the groups for the given user for too long
+2. we might end up generating a lot of traffic to the OIDC provider
+3. the OIDC may not expose/implement the UserInfo endpoint
+
+### Graduation Criteria
+
+The feature should be considered mature once it lands.
+
+#### Dev Preview -> Tech Preview
+
+#### Tech Preview -> GA
+
+#### Removing a deprecated feature
+
+### Upgrade / Downgrade Strategy
+
+### Version Skew Strategy
+
+## Implementation History
+
+Major milestones in the life cycle of a proposal should be tracked in `Implementation
+History`.
+
+## Drawbacks
+
+There are two drawbacks to the current solution:
+
+1. the groups go out of sync between logins
+- This is quite a common problem with all systems that receive delegated identities.
+  Third party vendors don't usually provide a standardized watch channel that could be
+  effectively and reliably observed, and so most systems either only synchronize on user
+  login or attempt to perform group retrievals in regular intervals, but real real-time
+  synchronization is usually impossible.
+
+2. the groups are assigned to a user, not an `<identity provider:user>` touple
+- Some configurations allow a user to be sourced by multiple identity providers, where
+  each of the providers might have different groups configured. The implemenentation
+  described in this enhancement does not distinguish the groups when the user attempts
+  to use their token and just uses all the OpenShift groups the user appears in.
+
+## Alternatives
+
+### oauth-server alternative
+
+The OAuthAccessToken, User and Groups are the three things that the current OpenShift
+authenticator retrieves during token validation.
+
+To solve the issue of the user only appearing in the groups of
+`union(current identity provider groups, OpenShift groups)`, we could take an
+alternative approach and only synchronize the groups into the OAuthAccessToken for
+the given user and session, since, as noted above, the access token is one of the
+three objects introspected during authentication, and is the only one of the three
+that is sure to describe the current session.
+
+Having the groups stored within the access token would likely make group
+synchronization in between logins (a possible future expansion) harder to perform
+because to assure the groups in the token actually come from the given identity provider,
+the field should be immutable (as all the other OAuthAccessToken fields are).
+
+### Using vanilla Kubernetes OIDC config
+
+Another alternative approach involves using the [Kubernetes OIDC authenticator](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server)
+directly to retrieve identities. This way the OpenShift authentication stack does
+not get involved at all. This means that:
+
+- No `OAuthAuthorizeToken`/`OAuthAccessToken` get created.
+- No `Identity`/`User`/`Groups` object get created.
+- Neither `ServiceAccount` or `OAuthClient` and the credentials derived from them
+  will be respected when applications attempt to retrieve user's credentials from
+  the identity provider. That is, unless the 3rd party identity provider does not
+  implement their own integration with OpenShift, which is unlikely.
+- A third-party identity provider is not likely to know about the OAuthAccessToken
+  scopes that OpenShift uses that some components may rely upon (e.g. "user:info"
+  scope for components that only need to be able to get the user data). This means
+  that the 3rd party OIDC needs to be flexible in order to allow requesting unknown
+  scopes and scopes with a given prefix (for the [role scopes](https://docs.openshift.com/container-platform/4.7/authentication/tokens-scoping.html#scoping-tokens-role-scope_configuring-internal-oauth)).
+- The upstream configuration for OIDC does not currently allow setting the
+  "userInfo.Extra" fields of the authentication response so even if the 3rd party
+  OIDC identity provider allows custom scopes, there is currently no way for the
+  OpenShift-specific `scopeAuthorizer` to learn of these.
+- Any component that relied on an OpenShift-specific retrieval of user information
+  (==> `get` on `User` with a given accesstoken) would need to be rewritten to
+  either use the OIDC token's claims or retrieve that information from the UserInfo
+  endpoint of the OIDC identity provider.
+- In order for `oc`, `oauth-proxy` and OpenShift web console to work with the given
+  3rd party OIDC identity provider, the user must configure the `oauthMetadata` of
+  the `authentication.openshift.io/cluster` resource so that these clients are
+  redirected to the correct URLs.
+
+  - Alternatively, this configuration may stay the same, but in that case a user
+    would have to configure an OIDC identity provider in the `oauth.openshift.io/cluster`
+    resource so that they are able to use it. This would lead to an inconsistent
+    behavior where logins with an ID token retrieved from the OIDC identity provider
+    would have the OIDC provider groups, but logins with username/password in
+    `oc`, `oauth-proxy`, OpenShift web console would not + the behavior would
+    likely differ in all the other points outlined in this section.


enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,235 @@
...
+#### OIDC Group Retrieval
+
+OIDC core specification does not mention a standard `id_token` claim that
+would serve for user group storage and so none of the standard scopes define
+accessing it. That means retrieving the groups will very likely always require
+the user to configure `extraScopes` in their identity provider configuration
+(oauth.config.openshift.io/cluster).
+
+A possible follow-up implementation might include being able to
+[request specific claims](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims)
```
s-urbaniak, line 84: Did you rather mean requesting individual claims https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter ?

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,235 @@
...
+    // groups is the list of claims whose values should be converted into OpenShift groups
+    // that will include the user
+    // +optional
+    Groups []string `json:"groups,omitempty"`
+}
+```
+
+The group object should be annotated with `oauth.openshift.io/idpname: <idpname>`
+where `<idpname>` marks the identity provider name from the configuration. It should
+also get an `oauth.openshift.io/generated: true` if it hasn't previously existed.
```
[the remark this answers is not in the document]
s-urbaniak: oh, sorry, i think i am dumb ;) you mean the synchronized group object i guess. Do we have a concrete type for that already that we can reference here similar to OpenIDCLaims?

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,235 @@
...
+
+#### oauth-server
+
+When the OIDC provider has group synchronization configured, the oauth-server
```
s-urbaniak, line 111: "configured" means implictely:
1. an `extraScopes` entry is configured (optional)
2. a non-empty set of `Groups` entries is configured
?

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,235 @@
...
+
+- allow retrieving user's groups from a 3rd party OIDC identity provider into
+  OpenShift when such a user logs in
+
+### Non-Goals
+
+- implement group synchronization for an identity provider type other than
+  OIDC
+- keep the groups retrieved from a 3rd party identity provider synchronized
+  between user logins
```
s-urbaniak, line 59: maybe be too obvious, but maybe let's add support for non-oidc providers as an explicit non-goal.

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,235 @@
...
+Once the server has the user's groups information, it will look up the
+OpenShift groups the user appears in, and it will either add to/remove from/create
+the necessary groups for the user. The user should only be removed from the groups
+that were created for that specific identity provider (as per the annotation
+from the [API section](#api)). If such a group ends up being empty, it should
+be removed if it contains the `oauth.openshift.io/generated: true` annotation.
+
+When a group that would otherwise be created on sync already exists, the
+oauth-server should only add the user among the group members and add an
+`oauth.openshift.io/idpname` annotation to mark it was created for the
```
s-urbaniak, line 126: could this be flapping if we have multiple identity providers configured?

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: oh yes, wrong link, thanks :+1:

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Implementation Details/Notes/Constraints [optional]
+
+#### OIDC Group Retrieval
+
+OIDC core specification does not mention a standard `id_token` claim that
+would serve for user group storage and so none of the standard scopes define
+accessing it. That means retrieving the groups will very likely always require
+the user to configure `extraScopes` in their identity provider configuration
+(oauth.config.openshift.io/cluster).
```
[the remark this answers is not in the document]
stlaz: since it does not appear in the Core specification, I wouldn't use a default. Ignoring unknown scopes is recommended but I can imagine some obsure IdPs erroring.

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: 2. should be enough

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,235 @@
...
+obtains from the provider after a successful authentication, and should do
+the same for the `id_token` retrieved from the OIDC's UserInfo endpoint
+should the OIDC be exposing this endpoint in its OIDC discovery data.
+
+Once the server has the user's groups information, it will look up the
+OpenShift groups the user appears in, and it will either add to/remove from/create
+the necessary groups for the user. The user should only be removed from the groups
+that were created for that specific identity provider (as per the annotation
+from the [API section](#api)). If such a group ends up being empty, it should
+be removed if it contains the `oauth.openshift.io/generated: true` annotation.
```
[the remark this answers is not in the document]
stlaz: I would like to avoid unnecessary knobs, so I would either leave them there or remove them, but wouldn't add an option to choose, at least not yet.

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: hmmm, good point, I think it I'll have to actually include the IdP name in the annotation itself

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Motivation
+
+In the real world, OpenShift is not the owner of user identities but these
+are stored in a 3rd party identity provider system. When OpenShift retrieves
+such an identity, it usually only gathers information about the user's unique
+identification in that IdP and optionally a username that's preferred for the
+user if it's different from its identifier.
+
+It is a common practice to group users in logical groups on the identity provider
```
slaskawi, line 43: The most generic spec we could leverage on is OpenID Connect. Typically, groups information might be put in the Access Token (from oAuth), ID Token (OpenID Connect) or we may grab it from User Info Endpoint (also OpenID Connect). Here are some examples from my local Keycloak instance:

* [Access Token with Groups Mapper turned on](https://jwt.io/#debugger-io?token=eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICJmZkVnekFYUnR3bXpvU0xJWXJCbXJ4VDBfVkQ2bmMzV1AtaDRJM05HTXRnIn0.eyJleHAiOjE2MjI0NDQ3NzksImlhdCI6MTYyMjQ0NDcxOSwianRpIjoiOWQyZjA4ZTAtNzY2Ni00YTZjLTk2OWYtZTdhOTM2NDIzNWE4IiwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdDo4MDgwL2F1dGgvcmVhbG1zL21hc3RlciIsImF1ZCI6WyJtYXN0ZXItcmVhbG0iLCJhY2NvdW50Il0sInN1YiI6IjYwMWY5N2NjLTQ2NzgtNGY2Ny04ZTNhLWIwZmZmMmRhZDllNiIsInR5cCI6IkJlYXJlciIsImF6cCI6ImFkbWluLWNsaSIsInNlc3Npb25fc3RhdGUiOiIyYmE2MzZhNy02ZTA5LTQ5OTgtYWFlNS0yNmU5NjU2ZWM4MDIiLCJhY3IiOiIxIiwicmVhbG1fYWNjZXNzIjp7InJvbGVzIjpbImNyZWF0ZS1yZWFsbSIsImRlZmF1bHQtcm9sZXMtbWFzdGVyIiwib2ZmbGluZV9hY2Nlc3MiLCJhZG1pbiIsInVtYV9hdXRob3JpemF0aW9uIl19LCJyZXNvdXJjZV9hY2Nlc3MiOnsibWFzdGVyLXJlYWxtIjp7InJvbGVzIjpbInZpZXctaWRlbnRpdHktcHJvdmlkZXJzIiwidmlldy1yZWFsbSIsIm1hbmFnZS1pZGVudGl0eS1wcm92aWRlcnMiLCJpbXBlcnNvbmF0aW9uIiwiY3JlYXRlLWNsaWVudCIsIm1hbmFnZS11c2VycyIsInF1ZXJ5LXJlYWxtcyIsInZpZXctYXV0aG9yaXphdGlvbiIsInF1ZXJ5LWNsaWVudHMiLCJxdWVyeS11c2VycyIsIm1hbmFnZS1ldmVudHMiLCJtYW5hZ2UtcmVhbG0iLCJ2aWV3LWV2ZW50cyIsInZpZXctdXNlcnMiLCJ2aWV3LWNsaWVudHMiLCJtYW5hZ2UtYXV0aG9yaXphdGlvbiIsIm1hbmFnZS1jbGllbnRzIiwicXVlcnktZ3JvdXBzIl19LCJhY2NvdW50Ijp7InJvbGVzIjpbIm1hbmFnZS1hY2NvdW50IiwibWFuYWdlLWFjY291bnQtbGlua3MiLCJ2aWV3LXByb2ZpbGUiXX19LCJzY29wZSI6Im9wZW5pZCBlbWFpbCBwcm9maWxlIiwiZW1haWxfdmVyaWZpZWQiOmZhbHNlLCJncm91cHMiOlsiY3JlYXRlLXJlYWxtIiwiZGVmYXVsdC1yb2xlcy1tYXN0ZXIiLCJvZmZsaW5lX2FjY2VzcyIsImFkbWluIiwidW1hX2F1dGhvcml6YXRpb24iXSwicHJlZmVycmVkX3VzZXJuYW1lIjoiYWRtaW4ifQ.XjJWgzw17mZOiLBXclWTCjfw1XE9xv9D6s0N0TGNzqihsXzDxZ6Y1vDovIuMYHqtk3-Vs6Ro53_gLF3BDpx8sc66cYOriCjlQGB-eYQuXMFmwa3vzM2I3I_TGXZIJAkEhV9-J1Zr5_DbluHYbb2CaNnkvSZcNBTdR-1DHBIbG4An5LanGkIuPqA_WDSb8H264TgJfM96SFDHkOz-kSuFkKVV_Pm6-oI6OzVl_Xq2Cp2UEhcva-8bdADdOWNkLa8O9KBL_fr9wL2zdv3iULWOMx2aQb6rcS_j5u7v9EzL9S9U2HDK6b-fzwkwJVgidJG0tmfpNe0oRbRQ4bpB5k8pAA&publicKey=-----BEGIN%20PUBLIC%20KEY-----%0AMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAshFo9sR3JdjbxtCzhuXD%0ANbZKZH%2FjaDT6BBJpiSK%2BvVMotGdooDTu1HYPiEwTQ6F1cj6AULiyRB2%2FXjoUEa7S%0AKqoCjG1S9dGaPPp2EccbI7qjtuQDn9J10JErPVPI6vMiH0jX%2BwzYZEibfoNGeK05%0AJ5BZ1sonhmtXQ8lPIpsI8anAYo7yzuKOgKsW2ABjAONtoUUJO7YElRSXtrtqtRAi%0A4tlwAdZOSFDZW4nMY5YIcnzuu29WYsbkEsBNWU8VTAZG3r8UoanyaUIXzEC9iRJr%0AQnWy65eceIzofZP0sDN1L8MldmmytvugbP6hPiMqYc2yGGiPI2yj%2BIQ6mUu%2FoVz%2F%0A%2FQIDAQAB%0A-----END%20PUBLIC%20KEY-----%0A)
* [ID Token obtained with "scope=openid"](https://jwt.io/#debugger-io?token=eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICJmZkVnekFYUnR3bXpvU0xJWXJCbXJ4VDBfVkQ2bmMzV1AtaDRJM05HTXRnIn0.eyJleHAiOjE2MjI0NDQ3NzksImlhdCI6MTYyMjQ0NDcxOSwiYXV0aF90aW1lIjowLCJqdGkiOiIyYzA1ZjdhNi0wY2VkLTRiN2UtYTE5Ny05OTlmYTlmZmM4N2EiLCJpc3MiOiJodHRwOi8vbG9jYWxob3N0OjgwODAvYXV0aC9yZWFsbXMvbWFzdGVyIiwiYXVkIjoiYWRtaW4tY2xpIiwic3ViIjoiNjAxZjk3Y2MtNDY3OC00ZjY3LThlM2EtYjBmZmYyZGFkOWU2IiwidHlwIjoiSUQiLCJhenAiOiJhZG1pbi1jbGkiLCJzZXNzaW9uX3N0YXRlIjoiMmJhNjM2YTctNmUwOS00OTk4LWFhZTUtMjZlOTY1NmVjODAyIiwiYXRfaGFzaCI6Ink4ekY3U2VDNXhzaTVKN3lhdmFRcXciLCJhY3IiOiIxIiwiZW1haWxfdmVyaWZpZWQiOmZhbHNlLCJncm91cHMiOlsiY3JlYXRlLXJlYWxtIiwiZGVmYXVsdC1yb2xlcy1tYXN0ZXIiLCJvZmZsaW5lX2FjY2VzcyIsImFkbWluIiwidW1hX2F1dGhvcml6YXRpb24iXSwicHJlZmVycmVkX3VzZXJuYW1lIjoiYWRtaW4ifQ.mWQMJ2YlyEcM0muhvKm71EXBwN5yjL15uYh-3tJetjsWElMV6MKHGKquMx426hkq-ojpWG5fhKhT4YTkiHNEHzWKg4K0pvb6vLabddrQiMpHAzZgIMogKDQn7PVsNxH03Fsepy7mk-WT8VJm9af0enGbDGiQRrV1SGUmGYCWqF5xJDIpgtD3mwTwpQflCZkoZywBslrQQ7JbI9s6Db_MrljAgSUqTQBz_j4mKSFWe2RVf1Pp8YHCXw40F69fGftCnNCZeooZuYvCNs7x_Yg1nEWHpRoi1oQvrjd5zusV14MINOQuF3WYSSojhP0WEAiAioOtHZ96p2fhF9gcuiFsWg&publicKey=-----BEGIN%20PUBLIC%20KEY-----%0AMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAshFo9sR3JdjbxtCzhuXD%0ANbZKZH%2FjaDT6BBJpiSK%2BvVMotGdooDTu1HYPiEwTQ6F1cj6AULiyRB2%2FXjoUEa7S%0AKqoCjG1S9dGaPPp2EccbI7qjtuQDn9J10JErPVPI6vMiH0jX%2BwzYZEibfoNGeK05%0AJ5BZ1sonhmtXQ8lPIpsI8anAYo7yzuKOgKsW2ABjAONtoUUJO7YElRSXtrtqtRAi%0A4tlwAdZOSFDZW4nMY5YIcnzuu29WYsbkEsBNWU8VTAZG3r8UoanyaUIXzEC9iRJr%0AQnWy65eceIzofZP0sDN1L8MldmmytvugbP6hPiMqYc2yGGiPI2yj%2BIQ6mUu%2FoVz%2F%0A%2FQIDAQAB%0A-----END%20PUBLIC%20KEY-----%0A)
* [User Info Endpoint](https://gist.github.com/slaskawi/647d903186f7af4c2a390ce9cd620e3d)

The biggest problem I can see is that there's no standard "groups" claim (see the [spec](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-json-web-token)), which means there's is no guarantee any of those will expose groups in a standard way. We'll rely on vendor-specific behavior here.

So how about implementing a mechanism that will scan the Access and the ID Tokens and try to sync groups automatically - without the additional round-trip to the User Info endpoint?

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,238 @@
...
+cases where it is desired.
+
+### Goals
+
+- allow retrieving user's groups from a 3rd party OIDC identity provider into
+  OpenShift when such a user logs in
+
+### Non-Goals
+
+- implement group synchronization for an identity provider type other than
```
slaskawi, line 56: In this case, let's mention that this enhancement is for the OpenID Connect IdPs explicitly. Once we state this, we could assume that we have an ID Token at hand.

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,238 @@
...
+  OIDC
+- keep the groups retrieved by OpenShift from a 3rd party identity provider
```
slaskawi, line 58: Actually, if we decided to extract groups information from tokens, we'd get this one almost for free.

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,238 @@
...
+organized in a complex and well-though-out group structure. I would like
+OpenShift to respect the group structure so that I can assign privileges
+to the users on OpenShift side according to their group membership in the
+identity provider.
+
+### Implementation Details
+
+#### OIDC Group Retrieval
+
+OIDC core specification does not mention a standard `id_token` claim that
```
slaskawi, line 78: Could you please rephrase it a bit? I think you mean there is no a standard "groups" claim in JWT nor OpenID Connect specs?

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,238 @@
...
+would serve for user group storage and so none of the standard scopes define
+accessing it. That means retrieving the groups will very likely always require
+the user to configure `extraScopes` in their identity provider configuration
```
slaskawi, line 81: OpenID Connect requires adding an additional scope (`openid`) to the token requests. Once we add it, we'll receive an ID Token, which should contain information about the user. As I mentioned earlier, there is no guarantee it will contain groups, but emitting them is up to the IdP's configuration. 

With that in mind, perhaps we could treat the Extra Scopes part as something optional? Most of the administrators should be able to make their IdPs to include groups in the token responses.

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,238 @@
...
+
+    // groups is the list of claims whose values should be converted into OpenShift groups
+    // that will include the user
+    // +optional
+    Groups []string `json:"groups,omitempty"`
+}
+```
+
+The group object (`groups.user.openshift.io.com/v1`) should be annotated with
+`oauth.openshift.io/idp.<idpname>: "synced"`
```
slaskawi, line 107: Perhaps we should also have an annotation with a timestamp, e.g. `oauth.openshift.io/idp.<idpname>.synctime: "<timestamp>"`? This could be useful when investigating synchronization issues between IdPs.

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,238 @@
...
+Once the server has the user's groups information, it will look up the
+OpenShift groups the user appears in, and it will either add to/remove from/create
+the necessary groups for the user. The user should only be removed from the groups
+that were created for that specific identity provider (as per the annotation
+from the [API section](#api)). If such a group ends up being empty, it should
+be removed if it contains the `oauth.openshift.io/generated: true` annotation.
+
+When a group that would otherwise be created on sync already exists, the
+oauth-server should only add the user among the group members and add an
+`oauth.openshift.io/idp.<idpname>` annotation to mark users of that IdP appear
```
slaskawi, line 129: If we decided to add a timestamp, we should add it here too.

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Open Questions
+
+1. Do we care about mapping the groups to their source identity provider per
+   session (i.e. a user is only a member of the groups sourced from the IdP
+   that they logged in to + any OpenShift, non-synced groups)?
+
+### Test Plan
+
+The authentication operator already has a way to configure Keycloak and
```
slaskawi, line 170: What are the 3rd party IdPs we'll use for testing? Keycloak? Google? Maybe something else?

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,238 @@
...
+3. on login, user gets added to an existing group that appears in the
+   retrieved group claims
+4. on login, user gets removed from any existing group that was created
+   for the given identity provider (per annotation) when the group
+   no longer appears among those retrieved from the OIDC
+
+
+### Follow-up Work
+
+To synchronize groups in a time-based manner, an additional controller in the
```
slaskawi, line 186: Syncing groups from an ID Token could help here as well. We could always check groups upon user login.

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,238 @@
...
+authenticator retrieves during token validation.
+
+To solve the issue of the user only appearing in the groups of
+`union(current identity provider groups, OpenShift groups)`, we could take an
+alternative approach and only synchronize the groups into the OAuthAccessToken for
+the given user and session, since, as noted above, the access token is one of the
+three objects introspected during authentication, and is the only one of the three
+that is sure to describe the current session.
+
+Having the groups stored within the access token would likely make group
```
slaskawi, line 235: Also, the Access Token could explode in its size if a user belongs to gazillion groups at the same time (which is very likely).

slaskawi:
Great proposal @stlaz! 

Most of my comments I could shortened to two things:
- How about extracting groups information from an ID Token instead of User Info Endpoint?
- How about synchronizing user upon his login (assuming we'll have an ID Token at hand)?

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: > typically, groups information might be put in the Access Token (from oAuth)
I don't think that's a very typical use, access tokens are usually (as described in the RFC) just opaque strings, although some are trying to put some semantics to it: https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/

> So how about implementing a mechanism that will scan the Access and the ID Tokens and try to sync groups automatically - without the additional round-trip to the User Info endpoint?
yes, the the introspection of the ID token is what we'll do in the implementation, but we won't be touching the access tokens at all

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: this is a non-goal, so what you say is already implied, no?

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: No, that would only be synced on logins.

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: I think I can just rephrase that there are no specific "group" claims, mentioning `id_token` here is confusing, you're right.

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: > Once we add it, we'll receive an ID Token, which should contain information about the user. 

not all the information is going to be included, but perhaps we can get rid of most of this paragraph

`ExtraScopes` are already optional

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: Perhaps we could merge that into this annotation. Although I am a little hesitant because that might be too many updates to just the annotation itself and not really doing much to the object.

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: We can use anything that provides a container and has a REST API to work with users and groups. I wouldn't use more than a single IdP though, writing vendor-specific integration tests starts being time-consuming.

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: Syncing groups on user login is the intent of this enhancement :slightly_smiling_face: The interesting part is what happens when the user is already logged in since currently there is no communication with the 3rd party IdP during that time today (that is the <user login, OpenShift access token expires> time interval).

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: True

stlaz:
> How about extracting groups information from an ID Token instead of User Info Endpoint?

we'll use whichever is available

> How about synchronizing user upon his login (assuming we'll have an ID Token at hand)?

that's the point of this enhancement :slightly_smiling_face: Perhaps I failed horribly when the intent is not visible from the doc.

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: But this is fine I guess. This is why you should use short-lived tokens in general. Upon a refresh (exchanging a Refresh Token for an Access Token) you should get a new ID Token as well. So the general remediation for this problem is to shorted the Access Token live time.

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: Ok, this works for me. But at some point in time, we'll probably need to mention at least one "certified" IdP in the documentation (or show some example, how to use this feature). It would be perfect if that example, and what we test automatically were the same things.

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: The update may happen asynchronously in the background. I have no issues with that. 

The important question is do we want to update the timestamp on every ID Token rotation? In other words, what will happen when we exchange a Refresh Token for a new Access Token and get a new ID Token "by the way". This rotation will probably happen pretty often (Access Token lifetime set in Keycloak is 5 min by default).

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: Thanks!

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: Ok, you're right! Thanks!

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: What will happen when we refresh an Access Token? Would we trigger the sync in this scenario as well?

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: Ok, makes sense!

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: We're not currently refreshing access tokens because we don't need to. User login is a one-shot operation where we retrieve what we can from the OIDC provider and then discard whichever tokens we got.

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: We may need to clarify what we expect of the OIDC provider to be capable of doing but this should generally work, without any IdP being put as "this product really works the best".

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: We don't do any refreshes though. Also we'll have to consider the very likely different lifetimes of a 3rd-party refresh token and OpenShift OAuthAccessToken.

slaskawi: Ok, understood. Thanks for clarification. 

So I'm just wondering - could we assume that a user needs to log out and in in order to refresh his roles? I'm trying to avoid any background synchronization mechanism in order to simplify the implementation.

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: Ok, so if we do not plan to certify any of the 3rd party IdP, we might need to implement a flexible mechanism for obtaining groups from the User Info Endpoint. In case of Keycloak, we know the structure (here's [an example](https://gist.github.com/slaskawi/647d903186f7af4c2a390ce9cd620e3d)) but I'm not sure we can assume the same layout for other IdPs. What do you think?

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: oooh ok - I forgot that we're considering the federation use case here. Thanks for clearing this one out!

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: Clarified in other comments. We don't rotate those tokens.

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: Clarified off-list - this is where `OpenIDClaims` comes into play.

slaskawi (approved):
Clarified all the doubts - LGTM!

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: They only need to log in again, which would update groups for all their other sessions, too.

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: On a further thought, I would refrain from synctime in the annotations, purely to avoid updates that only update the annotation time.

s-urbaniak:
/lgtm
/approve
final nits discussed OOB :tada: :shipit:

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: Could we create an event or an audit log for that instead? I believe this functionality might be crucial for sorting permission settings order problems in the future.

enhancements/authentication/synchronizing-user-groups-on-login.md
slaskawi: Ok, thanks!

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,247 @@
...
+It is a common practice to group users in logical groups on the identity provider
+side and sometimes it would make sense to bring the groups over to the OpenShift
+cluster as well. The identity providers oftentimes allow a protocol-specific path
+to retrieve group information, which is something OpenShift should leverage in those
+cases where it is desired.
+
+### Goals
+
+- allow retrieving user's groups from a 3rd party OIDC identity provider into
+  OpenShift when such a user logs in
```
deads2k, line 52: doesn't this assume that all groups presented are 
1. valid for openshift
2. desired in the cluster
3. are the same for all tokens issued

?

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,247 @@
...
+  login or attempt to perform group retrievals in regular intervals, but real real-time
+  synchronization is usually impossible.
+
+2. the groups are assigned to a user, not an `<identity provider:user>` touple
+- Some configurations allow a user to be sourced by multiple identity providers, where
+  each of the providers might have different groups configured. The implemenentation
+  described in this enhancement does not distinguish the groups when the user attempts
+  to use their token and just uses all the OpenShift groups the user appears in.
+
+## Alternatives
```
deads2k, line 232: Can you describe the shortcomings of using OIDC directly from the kube-apiserver?  I suspect that write up may become even more important over time.

I think the primary issue is the usage of serviceaccounts and their tokens as automatic oauthclients which can be used for in-cluster addons automatically on openshift.  This shortcoming may be enough to justify work like this, but it may also justify why certain integration paths for external token issuers fall short and give a concrete path for another project to become a token issuer.

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: > doesn't this assume that all groups presented are
>
>    1. valid for openshift
>    2. desired in the cluster
 
It does. I suppose we could make lists to allow/block certain groups appear in the generic provider API.

> 3. are the same for all tokens issued
yes

enhancements/authentication/synchronizing-user-groups-on-login.md
stlaz: That's a good point, I'll try to think about this and will add a section here.

stlaz: Outlined a few issues that came to my mind

enhancements/authentication/synchronizing-user-groups-on-login.md
```diff
@@ -0,0 +1,287 @@
...
+the given user and session, since, as noted above, the access token is one of the
+three objects introspected during authentication, and is the only one of the three
+that is sure to describe the current session.
+
+Having the groups stored within the access token would likely make group
+synchronization in between logins (a possible future expansion) harder to perform
+because to assure the groups in the token actually come from the given identity provider,
+the field should be immutable (as all the other OAuthAccessToken fields are).
+
+### Using vanilla Kubernetes OIDC config
```
deads2k, line 251: This section is great.

deads2k:
> /lgtm

Open a PR adding @sur and @stlaz as approvers in the authentication dir please.

/approve

for @sur.

PR #343, opened 2020-05-26
cluster-wide oauth-proxy settings

deads2k:
The [oauth-proxy](https://github.com/openshift/oauth-proxy) is a commonly used proxy which allows a naive application to make use of in-cluster authentication and authorization to protect itself on an openshift platform without changing the application itself. This effectively makes the application appear as an extension to our platform, examples include: logging kibana, monitoring grafana, etc. There are many configuration values that are considered as related to platform configuration or security that the oauth-proxy could auto-detect.

@s-urbaniak @lilic @stlaz @sttts This is what I'm thinking about for some of the settings.  Would it help?

PR #797, opened 2021-05-31
custom routes: set up an explicit trust to the oauth-server

stlaz:
Extends the "custom route configuration" enhancement document with a way to trust the oauth-server for components that need to request an OpenShift access token.

/assign @deads2k 
/cc @s-urbaniak

### DIFF
--- a/enhancements/ingress/custom-route-configuration.md
+++ b/enhancements/ingress/custom-route-configuration.md
@@ -59,10 +59,11 @@ slightly different documentation.
 
 ### Goals
 
-1. provide a way for cluster-admins to see all the routes that have flexible names and serving certificates in the cluster.
+1. provide a way for cluster-admins to see all the routes that have flexible names and serving certificates in the cluster
 2. provide a single API that is consistent for every route they need to configure
 3. provides a way to have an operator with scope limited permissions read the serving cert/key pairs
-4. allow a cluster-admin to specify a different DNS name.
+4. allow a cluster-admin to specify a different DNS name
+5. provide a way for other components to trust the oauth-server with a custom serving certificate
 
 ### Non-Goals
 
@@ -193,6 +194,23 @@ pairs.
 Library-go is also an ideal spot to provide functionality to inject route hosts.  Something like our StaticResource
 controller which automatically handles the API status and the required route changes.
 
+#### Trusting the oauth-server
+
+Some OpenShift components need to communicate with the OpenShift oauth-server in order to
+retrieve access tokens on behalf of the user that wants to use them. Typical examples include
+the web console and anything that uses the oauth-proxy, like the monitoring tools. A custom oauth-server
+serving certificate might break trust from the other components in the cluster.
+
+The authentication operator needs to publish the oauth-server's  trust root in a well-known location so
+that components that need to trust the oauth-server can use it. The authentication operator will therefore
+always create a configMap `openshift-config-managed/oauth-server-cert` which is going to include the
+oauth-server's serving certificate.
+
+The certificate should be found in the `ca-bundle.crt` key in the `data` of such a configMap.
+
+The operators managing the components that need to trust the oauth-server have to consume this configMap
+and configure the components' so that they trust the certificate.
+
 ### Risks and Mitigations
 
 Some components have attempted to resolve similar issues in one off ways.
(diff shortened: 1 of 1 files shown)


enhancements/ingress/custom-route-configuration.md
```diff
@@ -193,6 +194,25 @@ pairs.
...
+that components that need to trust the oauth-server can use it. The authentication operator will therefore
+always create a configMap `openshift-config-managed/oauth-server-cert` which is going to include:
+
+1. the closest-to-leaf CA certificate to the oauth-server serving certificate if that is possible
+2. the oauth-server's serving certificate if no CA certificate that signed it is known
+
+The certificate should be found in the `ca-bundle.crt` key in the `data` of such a configMap.
+
+The operators managing the components that need to trust the oauth-server have to consume this configMap
+and configure the components' so that they trust the certificate.
```
s-urbaniak, line 214: isn't this something oauth-proxy could do autonomously to prevent redundant implementations?

stlaz: It would be a possibility. Perhaps something to consider in https://github.com/openshift/enhancements/pull/343?

enhancements/ingress/custom-route-configuration.md
```diff
@@ -193,6 +194,25 @@ pairs.
...
+retrieve access tokens on behalf of the user that wants to use them. Typical examples include
+the web console and anything that uses the oauth-proxy, like the monitoring tools. A custom oauth-server
+serving certificate might break trust from the other components in the cluster.
+
+The authentication operator needs to publish the oauth-server's  trust root in a well-known location so
+that components that need to trust the oauth-server can use it. The authentication operator will therefore
+always create a configMap `openshift-config-managed/oauth-server-cert` which is going to include:
+
+1. the closest-to-leaf CA certificate to the oauth-server serving certificate if that is possible
+2. the oauth-server's serving certificate if no CA certificate that signed it is known
```
s-urbaniak, line 209: @slaskawi question: @stlaz and me brainstormed if this is actually too complex. Instead, we could simplify this and just return the configured serving cert as-is. Do you see pros/cons?

slaskawi: I think the main reason to use CAs as far away from root as possible is helpful for blacklisting. This way, when you invalidate CAs close to the leaf, there's no reason to regenerate certs between the leaf and the root. 

But I guess there's no CA blacklisting mechanism implemented in OpenShift whatsoever. If my assumption is correct - I guess it's safe to use the serving cert and simplify this part.

stlaz: modified so that only the serving-cert is present

enhancements/ingress/custom-route-configuration.md
```diff
@@ -193,6 +194,23 @@ pairs.
...
 controller which automatically handles the API status and the required route changes.
 
+#### Trusting the oauth-server
+
+Some OpenShift components need to communicate with the OpenShift oauth-server in order to
+retrieve access tokens on behalf of the user that wants to use them. Typical examples include
+the web console and anything that uses the oauth-proxy, like the monitoring tools. A custom oauth-server
+serving certificate might break trust from the other components in the cluster.
+
+The authentication operator needs to publish the oauth-server's  trust root in a well-known location so
```
deads2k, line 204: you don't always have the trust.  You sometimes only have the serving cert.  The golang TLS negotiation doesn't require a chain back to the root.

enhancements/ingress/custom-route-configuration.md
```diff
@@ -193,6 +194,23 @@ pairs.
...
+that components that need to trust the oauth-server can use it. The authentication operator will therefore
+always create a configMap `openshift-config-managed/oauth-server-cert` which is going to include the
```
deads2k, line 206: `oauth-serving-cert` since that says what its for as opposed to the noun "server" which could be a client-cert.

deads2k:
This seems like a reasonable idea.  I agree that the oauth-proxy binary should automatically look for this and we probably want that particular configmap to be readable by system:authenticated.

enhancements/ingress/custom-route-configuration.md
stlaz: ack, I think this is a leftover from a previous version.

slaskawi (approved):
/lgtm

PR #800, opened 2021-06-02
updates: capture offline discussion on eus-upgrades-mvp

dhellmann:
Incorporate the outcome of the thread on #762 terminating at
https://github.com/openshift/enhancements/pull/762/files#r626609451
into the doc for better discoverability.

/priority important-soon
/cc @crawford
/cc @markmc
/cc @sdodson
/cc @sttts
/cc @deads2k
/assign @derekwaynecarr

### DIFF
--- a/enhancements/update/eus-upgrades-mvp.md
+++ b/enhancements/update/eus-upgrades-mvp.md
@@ -275,6 +275,38 @@ customers.
 
 ## Design Details
 
+### Offline Discussion
+
+To recap an offline discussion about this for those that were not able
+to attend based on [review comments on the original
+PR](https://github.com/openshift/enhancements/pull/762/files#r626609451):
+
+* The API server ships with a set of built-in types (pod and node).
+* The upstream project via API review/human or project operator
+  oversight ensures that skew between API server and kubelet is n-2.
+* The API server operator is being asked to report Upgradeable=False
+  to capture that API review/human/operational oversight because it is
+  the component that must not upgrade.
+* The Machine Config Operator is the component that must upgrade, so
+  reporting Upgradeable=False is confusing.
+* The Machine Config Operator should report a message in its reporting
+  that admins should perform an action.
+
+I (@derekwaynecarr) agree that the API server (and its operators)
+should not have to understand all API types it serves and all its
+clients. I think its reasonable for API types that are built-in, whose
+upstream set of human operators/api-reviewers/etc. are trying to
+ensure that skew in either direction (api-server or kubelet) is
+preserved that both operators in our context should report a desired
+action (apiserver operator should say I cannot upgrade, machine config
+operator should promote an upgrade).
+
+For types that are not built-in (CRDs, etc.), it is extremely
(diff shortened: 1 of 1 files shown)


enhancements/update/eus-upgrades-mvp.md
```diff
@@ -275,6 +275,38 @@ customers.
 
 ## Design Details
 
+### Offline Discussion
+
+To recap an offline discussion about this for those that were not able
+to attend based on [review comments on the original
+PR](https://github.com/openshift/enhancements/pull/762/files#r626609451):
+
+* The API server ships with a set of built-in types (pod and node).
```
sttts, line 284: unclear why this matters?

Do you want to say that the pod subresources are implemented through APIs served by the kubelet and called by the kube-apiserver? This puts a constraint on maximal skew.

enhancements/update/eus-upgrades-mvp.md
```diff
@@ -275,6 +275,38 @@ customers.
...
+  that admins should perform an action.
+
+I (@derekwaynecarr) agree that the API server (and its operators)
+should not have to understand all API types it serves and all its
+clients. I think its reasonable for API types that are built-in, whose
+upstream set of human operators/api-reviewers/etc. are trying to
+ensure that skew in either direction (api-server or kubelet) is
+preserved that both operators in our context should report a desired
+action (apiserver operator should say I cannot upgrade, machine config
+operator should promote an upgrade).
```
sttts, line 302: I still disagree with this assessment. This is not about REST API skew (of built-in resources) at all, and hence kubelet as an API client does not matter. This is about kube-apiserver->kubelet API usage to implement certain subresources. Hence, the apiserver is the client and the kubelet dictates which APIs it serves. And therefore it makes sense that the apiserver has to block upgrades.

sttts: To be clear: the outcome of my logic is the same: the ckao will report Upgradable=False.

enhancements/update/eus-upgrades-mvp.md
dhellmann: I'm not sure what the background is, I wasn't in the meeting. I'm capturing @derekwaynecarr's comment from the original PR.

enhancements/update/eus-upgrades-mvp.md
dhellmann: @sttts Does that mean this enhancement doc needs a different change? I want to capture the argument in the body, so maybe this text needs to be proposed as an update to the alternatives section, instead of here? I'm not aware of what the final outcome was, so help me work out what updates are needed, please.

sdodson: I believe the intent is really to capture the leading component must not upgrade, in this scenario that's the apiserver. The details around API types and compatibility are just in support of that.

sdodson:
/lgtm
This seems like a decent clarification and summary of offline discussions.

dhellmann:
Thanks, @sdodson !

PR #802, opened 2021-06-04
WIP Workload partitioning API enhancement

MarSik:
This is part of a bigger Workload partitioning enhancement that deals with the trigger API specifics.

Still heavy WIP.

Signed-off-by: Martin Sivák <msivak@redhat.com>

### DIFF
--- a/enhancements/management-workload-partitioning-api.md
+++ b/enhancements/management-workload-partitioning-api.md
@@ -79,7 +79,7 @@ The short term goal is to get enough functionality to create the `management` pa
 
 - Long term goal of allowing cluster admins to create their own partitions is not part of this proposal for now
 - Day 2 creation of partitions is not part of the proposal
-- Namespace or MCP specific partitions are not part of the proposal 
+- Namespace or MCP specific partitions are not part of the proposal
 
 ## Proposal
 
--- a/enhancements/management-workload-partitioning-api.md
+++ b/enhancements/management-workload-partitioning-api.md
@@ -1,0 +1,239 @@
+---
+title: management-workload-partitioning-api
+authors:
+  - "@marsik"
+  - "@dhellmann"
+  - "@mrunalp"
+  - "@browsell"
+reviewers:
+  - "@deads2k"
+  - "@staebler"
+  - TBD
+approvers:
+  - "@smarterclayton"
+  - "@derekwaynecarr"
+  - "@markmc"
+creation-date: 2021-06-04
+last-updated: 2021-06-04
+status: provisional
+see-also:
+  - "/enhancements/management-workload-partitioning.md"
+---
+
+# Management workload partitioning API
+
+
+To get started with this template:
+1. **Pick a domain.** Find the appropriate domain to discuss your enhancement.
+1. **Make a copy of this template.** Copy this template into the directory for
+   the domain.
+1. **Fill out the "overview" sections.** This includes the Summary and
+   Motivation sections. These should be easy and explain why the community
+   should desire this enhancement.
+1. **Create a PR.** Assign it to folks with expertise in that domain to help
+   sponsor the process.
+1. **Merge at each milestone.** Merge when the design is able to transition to a
+   new status (provisional, implementable, implemented, etc.). View anything
+   marked as `provisional` as an idea worth exploring in the future, but not
+   accepted as ready to execute. Anything marked as `implementable` should
+   clearly communicate how an enhancement is coded up and delivered. If an
+   enhancement describes a new deployment topology or platform, include a
+   logical description for the deployment, and how it handles the unique aspects
+   of the platform. Aim for single topic PRs to keep discussions focused. If you
+   disagree with what is already in a document, open a new PR with suggested
+   changes.
+1. **Keep all required headers.** If a section does not apply to an
+   enhancement, explain why but do not remove the section. This part
+   of the process is enforced by the linter CI job.
+
+The `Metadata` section above is intended to support the creation of tooling
+around the enhancement process.
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The overall feature summary is present in the [Management workload partitioning](management-workload-partitioning.md) feature. This document focuses on how the feature enablement API and its interactions will look like.
+
+## Motivation
+
+This section is for explicitly listing the motivation, goals and non-goals of
+this proposal. Describe why the change is important and the benefits to users.
+
+### Goals
+
+This feature is about designing a read-only API that will describe the enabled workload partitions (types, classes, etc.). This information is needed for kubelet to start exposing the right resources as well for the admission webhook to know when the pod manipulation is needed.
+
+It is expected this API will be created at the installation process. Either manually or using the Performance addon operator render mode.
+
+The short term goal is to get enough functionality to create the `management` partition where all OpenShift infrastructure could be placed. This partition is cluster wide and only affects pods with certain pre-agreed-upon labels and placed in specific pre-annotated namespaces.
+
+### Non-Goals
+
+- Long term goal of allowing cluster admins to create their own partitions is not part of this proposal for now
+- Day 2 creation of partitions is not part of the proposal
+- Namespace or MCP specific partitions are not part of the proposal 
+
+## Proposal
+
+The proposal is to define a new cluster-wide Custom Resource Definition that would describe the allowed partition names in the status section. That way it hints at being a read only object where no user/admin input or modifications are expected.
+
+```yaml
+apiVersion: workloadpartitioning.openshift.io/v1
+kind: WorkloadPartitions
+metadata:
+  # arbitrary name, all objects of this Kind should be processed and merged
+  name: management-partition
+status:
+  # list of strings, defines partition names to be exposed by kubelet
+  globalPartitionNames:
+    - management
+```
+
+To allow for future extensibility and possible multiple sources of workload partition names (coming from customers, the installer, or other operators, etc.), we propose that there might be multiple `WorkloadPartitions` objects injected into the cluster. The expected behavior is that all components would just merge all the defined names together.
+
+Only the cluster administrator will have the ability to create or manipulate the WorkloadPartitions objects. Anyone will be allowed to read them.
+
+### User Stories
+
+Please check the user stories in the parent [Workload partitioning management](management-workload-partitioning.md) feature.
+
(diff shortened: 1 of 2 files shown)


enhancements/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,239 @@
...
+
+It is expected this API will be created at the installation process. Either manually or using the Performance addon operator render mode.
+
+The short term goal is to get enough functionality to create the `management` partition where all OpenShift infrastructure could be placed. This partition is cluster wide and only affects pods with certain pre-agreed-upon labels and placed in specific pre-annotated namespaces.
+
+### Non-Goals
+
+- Long term goal of allowing cluster admins to create their own partitions is not part of this proposal for now
+- Day 2 creation of partitions is not part of the proposal
+- Namespace or MCP specific partitions are not part of the proposal
```
dhellmann, line 82: I'm not sure how widely we've discussed the different types of scope we might want to have, so phrasing this a little differently may make it clearer.

```suggestion
- Partitions with scope less than the entire cluster (such as being limited to Namespaces or MCPs) are not part of the proposal
```

enhancements/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,239 @@
...
+status:
+  # list of strings, defines partition names to be exposed by kubelet
+  globalPartitionNames:
+    - management
+```
+
+To allow for future extensibility and possible multiple sources of workload partition names (coming from customers, the installer, or other operators, etc.), we propose that there might be multiple `WorkloadPartitions` objects injected into the cluster. The expected behavior is that all components would just merge all the defined names together.
+
+Only the cluster administrator will have the ability to create or manipulate the WorkloadPartitions objects. Anyone will be allowed to read them.
+
```
dhellmann, line 103: We should probably say explicitly that we don't have a controller to reconcile the resource because it is not expected to change.

enhancements/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,239 @@
...
+- Long term goal of allowing cluster admins to create their own partitions is not part of this proposal for now
+- Day 2 creation of partitions is not part of the proposal
+- Namespace or MCP specific partitions are not part of the proposal
+
+## Proposal
+
+The proposal is to define a new cluster-wide Custom Resource Definition that would describe the allowed partition names in the status section. That way it hints at being a read only object where no user/admin input or modifications are expected.
+
+```yaml
+apiVersion: workloadpartitioning.openshift.io/v1
```
dhellmann, line 89: In the original enhancement we use `workload.openshift.io` in some annotations. Should we do that here, too, and make this something like `partitioning.workload.openshift.io` What is standard/preferred for OpenShift APIs?

Alternatively, do we anticipate other API inputs related to workloads that might live on this CR later, so we should not include "partition" in the name?

enhancements/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,239 @@
...
+kind: WorkloadPartitions
+metadata:
+  # arbitrary name, all objects of this Kind should be processed and merged
+  name: management-partition
+status:
+  # list of strings, defines partition names to be exposed by kubelet
+  globalPartitionNames:
```
dhellmann, line 96: Bike shed: We should be consistent and either call it "global" or "cluster" but not use those two terms interchangeably.

enhancements/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Drawbacks
+
+This feature and its implementation might lead cluster administrators to start using it for manual overrides of the core components like cpu and/or topology manager. That would be an anti-pattern and the presentation of this feature should clearly state that.
+
+## Alternatives
+
+Similar to the `Drawbacks` section the `Alternatives` section is used to
+highlight and record other possible approaches to delivering the value proposed
+by an enhancement.
```
dhellmann, line 235: Elsewhere we've discussed the possible need to add per-workload parameters, like scope. The proposal above implies those would end up as separate lists, which I think is fine. We should include the other form here in the Alternatives section, for completeness. Our notes doc has an example as

```yaml
workloadTypes:
    - name: user-defined-workload-type
      scope: Pool
    - name: management
      Scope: ClusterWide
```

enhancements/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,239 @@
...
+metadata:
+  # arbitrary name, all objects of this Kind should be processed and merged
+  name: management-partition
+status:
+  # list of strings, defines partition names to be exposed by kubelet
+  globalPartitionNames:
+    - management
+```
+
+To allow for future extensibility and possible multiple sources of workload partition names (coming from customers, the installer, or other operators, etc.), we propose that there might be multiple `WorkloadPartitions` objects injected into the cluster. The expected behavior is that all components would just merge all the defined names together.
```
dhellmann, line 100: In other places where we have cluster-scoped configuration resources like this we look for 1 well-defined name, often `cluster`. I'm not sure how we decide between 1 and many, though. Maybe in this case many makes sense if we assume there may be multiple creators but that the CRD only has status fields. Do we really anticipate multiple creators?

enhancements/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,239 @@
...
+What are the caveats to the implementation? What are some important details that
+didn't come across above. Go in to as much detail as necessary here. This might
+be a good place to talk about core concepts and how they relate.
+
+### Risks and Mitigations
+
+1) An admin creates a WorkloadPartitions object manually after the cluster has been running for some time on a cluster that already has partitioning enabled
+1) An admin creates the WorkloadPartitions object manually after the cluster has been running for some time on a cluster with no workload partitioning enabled
+1) An admin deletes the WorkloadPartitions object that was created during the install process
+1) A random user manages to create a WorkloadPartitions object due to a bug in the defined RBAC rules
```
dhellmann, line 121: This list is a good start. We should add some detail about what ill effects might result from each of these cases.

enhancements/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,239 @@
...
+The proposal is to define a new cluster-wide Custom Resource Definition that would describe the allowed partition names in the status section. That way it hints at being a read only object where no user/admin input or modifications are expected.
+
+```yaml
+apiVersion: workloadpartitioning.openshift.io/v1
+kind: WorkloadPartitions
+metadata:
+  # arbitrary name, all objects of this Kind should be processed and merged
+  name: management-partition
+status:
+  # list of strings, defines partition names to be exposed by kubelet
```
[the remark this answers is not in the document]
MarSik: Oh yeah, you are right.

enhancements/management-workload-partitioning-api.md
MarSik: If we allow custom partitions in the future then I would say yes, there might be multiple owners in such case (customer creates one and PAO renders another one). I just did not want to close the door prematurely.

enhancements/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,239 @@
...
+end to end tests.**
+
+#### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
+
+### Upgrade / Downgrade Strategy
+
+There is currently no active component managing the WorkloadPartitions custom resource. Since we do not want to allow enabling workload partitioning for existing clusters there does not have to be any upgrade logic.
```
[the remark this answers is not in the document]
MarSik: Isn't this the same I have in risks? New CR created in a cluster with no workload partitioning enabled.

enhancements/management-workload-partitioning-api.md
MarSik: Good point!

enhancements/management-workload-partitioning-api.md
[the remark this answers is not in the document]
dhellmann: True, yes, that's the same scenario. Maybe referencing the risk here is enough, so we can avoid duplication.

enhancements/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Non-Goals
+
+- Long term goal of allowing cluster admins to create their own partitions is not part of this proposal for now
+- Day 2 creation of partitions is not part of the proposal
+- Namespace or MCP specific partitions are not part of the proposal 
+
+## Proposal
+
+The proposal is to define a new cluster-wide Custom Resource Definition that would describe the allowed partition names in the status section. That way it hints at being a read only object where no user/admin input or modifications are expected.
```
Tal-or, line 86: It is not clear from the proposal who should own this CRD?

dhellmann: That's explained a little better in #753 but we should probably summarize it here. The CRD will be defined in `openshift/api`.

enhancements/workload-partitioning/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,249 @@
...
+  # that should be configured on the kubelet and CRI-O level.
+  clusterPartitionNames:
+    - management
+```
+
+It is expected this API will be created at the installation process. Either manually or using the Performance Addon Operator render mode.
+
+To allow for future extensibility and possible multiple sources of workload partition names (coming from customers, the installer, or other operators, etc.), we propose that there might be multiple `WorkloadPartitions` objects injected into the cluster. The expected behavior is that all components would just merge all the defined names together.
+
+There is no controller or reconcile loop as part of this proposal. Only the cluster administrator will have the ability to create or manipulate the WorkloadPartitions objects. Anyone will be allowed to read them.
```
cynepco3hahue, line 101: We still will run reconciliation under the PAO once the `WorkloadPartitions` resource updated.

enhancements/workload-partitioning/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Implementation Details/Notes/Constraints [optional]
+
+What are the caveats to the implementation? What are some important details that
+didn't come across above. Go in to as much detail as necessary here. This might
+be a good place to talk about core concepts and how they relate.
+
+### Risks and Mitigations
+
+1) An admin creates a WorkloadPartitions object manually after the cluster has been running for some time on a cluster that already has partitioning enabled
```
cynepco3hahue, line 117: looks like something happened with ordering

enhancements/workload-partitioning/management-workload-partitioning-api.md
```diff
@@ -0,0 +1,249 @@
...
+apiVersion: workload.openshift.io/v1
+kind: WorkloadPartitions
+metadata:
+  # arbitrary name, all objects of this Kind should be processed and merged
+  name: management-partition
+status:
+  # List of strings, defines partition names that will be recognized by the
+  # workload partitioning webhook. This list will also inform PAO about partitions
+  # that should be configured on the kubelet and CRI-O level.
+  clusterPartitionNames:
```
cynepco3hahue, line 93: hm what about CPUs that should be used for the CPU pinning or will it use reservedCPUs by default?

MarSik: The API server does not care. Kubelet / CRIO will get the right ids from PAO.

PR #812, opened 2021-06-22
Add proposal about Node Operator

saschagrunert:
Adding an enhancement for a dedicated CRI-O operator.

Please take a first look @rphillips @nee1esh @mrunalp

### DIFF
--- a/enhancements/node/node-operator.md
+++ b/enhancements/node/node-operator.md
@@ -1,0 +1,316 @@
+---
+title: node-operator
+authors:
+  - "@saschagrunert"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2021-06-22
+last-updated: 2021-06-22
+status: provisional
+see-also: {}
+replaces: {}
+superseded-by: {}
+---
+
+# Node Operator (NOP)
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The overall goal of creating this new operator in OpenShift is the encapsulation
+of CRI-O and kubelet related configurations, metrics and alerts within a single
+domain and namespace.This provides on one hand a higher flexibility for the Node
+team, while on the other hand the dependency on external teams gets reduced. In
+the long-term the Node team would only rely on OpenShift internal APIs to manage
+the container runtime and kubelet related configurations. Additional deployments
+for proxying runtime metrics or providing telemetry access are part of this
+operator, too.
+
+## Motivation
+
+Most of the node related configurations are today part of the [Machine Config
+Operator (MCO)][mco]. Because the MCO got bigger over time, their long-term goal
+is to provide an API for machine related configuration purposes. The new
+operator would use this API in a later migration step.
+
+The CRI-O metrics are today served by using a [`kubelet`
+ServiceMonitor][kubelet-monitor] and a corresponding Prometheus relabeling. This
+happens as part of the [Cluster Monitoring Operator (CMO)][cmo] project. The CMO
+team also aims to provide an API for external teams to integrate their
+metrics retrieval into the OpenShift cluster monitoring. In the same way as for
+the MCO, the new node operator would use this API to integrate the runtime
+metrics retrieval natively into OpenShift by using a dedicated ServiceMonitor
+secured by TLS and RBAC.
+
+[mco]: https://github.com/openshift/machine-config-operator
+[cmo]: https://github.com/openshift/cluster-monitoring-operator
+[kubelet-monitor]: https://github.com/openshift/cluster-monitoring-operator/blob/8ec92e1/assets/control-plane/service-monitor-kubelet.yaml#L95-L116
+
+### Goals
+
+- Near-term:
+  - Create a new Node owned project to be prepared for future changes of the MCO
+    and CMO.
+  - Build in CRI-O metrics retrieval by using TLS and RBAC with the help of a
+    kube-rbac-proxy DaemonSet.
+  - Include TLS certificate rotation (on disk as well as in a secret) for the
+    CRI-O metrics.
+- Long-term:
+  - Migrate the kubelet and CRI-O configurations into the operator.
+  - Include [OpenTelemetry][opentelemetry] support for a node.
+
+[opentelemetry]: https://opentelemetry.io
+
+### Non-Goals
+
+Re-implementing logic from the MCO or CMO, we will rely on their APIs if they're
+available.
+
+## Proposal
+
+The main proposal is integrating a new operator into OpenShift, which is owned
+by the Node team. This includes the whole supply chain around CI/CD and release
+management.
+
+The first implementation of the operator will consist of:
+
+- A dedicated namespace where the operator lives and acts on.
+- A controller managing the reconciliation of the statically provided assets,
+  which are:
+  - RBAC rules as well as a ServiceAccount for the kube-rbac-proxy based metrics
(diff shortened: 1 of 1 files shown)


enhancements/node/cri-o-operator.md
```diff
@@ -0,0 +1,307 @@
...
+- Create a new Node owned project to be prepared for future changes of the MCO
+  and CMO.
+- Build in CRI-O metrics retrieval by using TLS and RBAC with the help of a
+  kube-rbac-proxy DaemonSet.
+- Include TLS certificate rotation (on disk as well as in a secret) for the
+  CRI-O metrics.
+
+### Non-Goals
+
+Anything which is already part of the MCO or CMO. For example, the CRI-O
```
haircommander, line 69: I would more consider this a goal, as ctrcfg handling is in the MCO now. 

also, is kubelet configuration also a goal, or will we have another operator to manage kubelet configuration?

enhancements/node/cri-o-operator.md
```diff
@@ -0,0 +1,307 @@
...
+### Risks and Mitigations
+
+#### A new project to integrate
+
+The highest risk for OpenShift as well as the Node team is that we have to
+maintain and integrate a new operator into the product. This will add initially
+higher efforts on the team, while gaining more flexibility in a mid-to-long
+term. A mitigation around this to keep the focus on metrics in the first place,
+while having in mind to change the operator later on to fulfill the APIs from
+the MCO and CMO.
```
haircommander, line 155: maybe also mention the risk of having a new operator managed by CVO (if we want that--it seems like we may)

enhancements/node/cri-o-operator.md
```diff
@@ -0,0 +1,307 @@
...
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The overall goal of creating this new operator in OpenShift is the encapsulation
+of CRI-O related configurations, metrics and alerts within a single domain and
+namespace. This provides on one hand a higher flexibility for the Node team,
```
haircommander, line 32: the summary here implies something the rest of this doc does not go into: what are we going to do with ContainerRuntimeConfig handling? is that being moved here, or is this only cri-o metrics related?

saschagrunert:
Okay it looks like we have to find a new name since we have to handle kubelet configuration in the future too. I'll also integrate the open telemetry efforts in the use cases in the next iteration.

saschagrunert:
@QiWang19 please take a look too if you find some time.

enhancements/node/cri-o-operator.md
saschagrunert: I think so, which also means we have to find a better name for the project.

haircommander: the Node Operator (NO) :)

saschagrunert: That's a great name! Is it a rule to have 3 characters for an operator abbreviation? This would mean we have to call it NOP.

haircommander:
xref https://hackmd.io/UzgrWuCLQ-u9bMh_JhRKOA as this implementation will likely depend on how the MCO refactor goes

enhancements/node/cri-o-operator.md
haircommander: or if 4 are allowed, it could also technically be the NOpe

saschagrunert: We should do a survey.

enhancements/node/cri-o-operator.md
```diff
@@ -0,0 +1,307 @@
...
+The operator will consist of:
+
+- A dedicated namespace where the operator lives and acts on.
+- A controller managing the reconciliation of the statically provided assets,
+  which are:
+  - RBAC rules as well as a ServiceAccount for the kube-rbac-proxy based metrics
+    server.
+  - SecurityContextConstraints to allow HostPorts and HostNetwork for the
+    DaemonSet.
+  - A DaemonSet to retrieve the metrics for every node.
```
derekwaynecarr, line 88: is this referring to a world where cri-o is the single source of truth for metrics (and not cadvisor embedded in kubelet)?

derekwaynecarr: or another way to frame this, what prerequisite work is needed upstream before inclusion of this operator makes sense?

haircommander: AFAIU this is cri-o specific metrics (image pull time is a good example), not the metrics coming from cAdvisor as part of https://github.com/kubernetes/enhancements/pull/2364

haircommander: i.e. there should be no requirements on upstream kube, just cri-o

saschagrunert: Yes, in the first place we only refer to the CRI-O metrics.

enhancements/node/cri-o-operator.md
saschagrunert: Changed it to "Node Operator" for now.

enhancements/node/cri-o-operator.md
saschagrunert: Yeah, made it more clear to handle kubelet and runtime related configurations too if an API is available.

enhancements/node/cri-o-operator.md
saschagrunert: Added the risk :+1:

enhancements/node/node-operator.md
```diff
@@ -0,0 +1,316 @@
...
+
+- Near-term:
+  - Create a new Node owned project to be prepared for future changes of the MCO
+    and CMO.
+  - Build in CRI-O metrics retrieval by using TLS and RBAC with the help of a
+    kube-rbac-proxy DaemonSet.
+  - Include TLS certificate rotation (on disk as well as in a secret) for the
+    CRI-O metrics.
+- Long-term:
+  - Migrate the kubelet and CRI-O configurations into the operator.
```
sdodson, line 69: Aside from metrics configuration and export below I don't see any other examples of kubelet and cri-o configuration mentioned. Is that because it's just not in scope for the initial proposal?

Once we get there I'm curious how we'd address coupling of configuration with the binaries presently managed by MCO updating CoreOS. Not sure if that's in scope or not.

haircommander: ^ that is a question I have, as well as how are we going to pull the NO images? right now we use podman to pull the mcd and run a one-off to configure before cri-o starts. would we use cri-o to pull the NO images (thus requiring a cri-o restart), or would we do a similar thing with podman, then configure cri-o

saschagrunert: > Aside from metrics configuration and export below I don't see any other examples of kubelet and cri-o configuration mentioned. Is that because it's just not in scope for the initial proposal?

Yes, the idea was that the Node Operator only holds the configuration, while we use an API of the MCO later to apply the config. 

> ^ that is a question I have, as well as how are we going to pull the NO images?

I would say that we use the standard way and let CRI-O pull the image. How we apply the configuration later on would be still managed by the MCO (API), right?

saschagrunert:
@rphillips @mrunalp PTAL

saschagrunert:
Outcome from a sync within the team: Find out if there is an MCO enhancement already in the works and if it makes sense to depend on it.

saschagrunert:
We will put this on hold and refactor on top of the outcome of https://issues.redhat.com/browse/GRPA-3104
/hold

dhellmann:
@saschagrunert we don't usually freeze the lifecycle of reviews in this repo because we want design discussions to move through quickly enough that implementation is not blocked.

If you need help finding reviewers, please post a message in #forum-arch on slack and we'll see if we can find the right people.

If this design isn't related to work that's actually ongoing, it's better to let the PR close and reopen it when the work is prioritized. That way the design details are fresh in everyone's mind when the code is being reviewed.

saschagrunert:
@dhellmann sound good, let's close this one for now.

PR #815, opened 2021-06-28
CONSOLE-2425: Support localization of dynamic plugins

jhadvig:
Updating dynamic plugins enhancement per discussion with @vojtechszocs and @christianvogt.

Linking implementation PR - https://github.com/openshift/console/pull/9196

PTAL

/assign @spadgett

### DIFF
--- a/enhancements/console/dynamic-plugins.md
+++ b/enhancements/console/dynamic-plugins.md
@@ -10,7 +10,7 @@ reviewers:
 approvers:
   - "@bparees"
 creation-date: 2020-08-18
-last-updated: 2021-06-30
+last-updated: 2021-07-22
 status: implementable
 ---
 
@@ -349,6 +349,33 @@ shows a message in a toast notification indicating that there is an update
 available and the user must refresh their browser to see the changes. (We won't
 refresh the page automatically to avoid possibly losing data if the user is
 entering something into a form or the YAML editor.)
+
+### Localization
+
+Info on how Console is handling i18n is in this [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/console/internationalization.md).
+
+Console uses [react-i18next](https://github.com/i18next/react-i18next) for i18n,
+and dynamic plugins must use react-i18next as well.
+
+All dynamic plugins must use a single react-i18next [namespace](https://www.i18next.com/principles/namespaces),
+named after the plugin, e.g. for `kubevirt` the filename would be
+`plugin__kubevirt.json`. Localization resources need to be served
+by the plugin service under the `locales/{language}/{namespace}.json`
+path relative to the `basePath` defined in the `ConsolePlugin` resource.
+All dynamic plugins must use the `plugin__` namespace prefix, e.g.
+`plugin__knative` or `plugin__kubevirt`. The request for the dynamic
+plugin localization resources will be proxied by console backend.
+For example, the `kubevirt` plugin localization resource
+in the `en` language will be requested at path
+`/locales/en/plugin__kubevirt.json`
+
+Here's a code example of how the `kubevirt` plugin would translate a message:
+```js
+const VMHeading = () => {
+  const { t } = useTranslation();
+  return <h1>{t('plugin__kubevirt~Virtual Machine')}</h1>;
+};
+```
 
 ### Error Handling
 


jhadvig:
The `dynamic-pugins.md` file is missine following section and subsections:

```
Checking enhancements/console/dynamic-plugins.md
enhancements/console/dynamic-plugins.md missing "### User Stories"
enhancements/console/dynamic-plugins.md missing "#### Dev Preview -> Tech Preview"
enhancements/console/dynamic-plugins.md missing "#### Tech Preview -> GA"
enhancements/console/dynamic-plugins.md missing "#### Removing a deprecated feature"
```

Not sure if thats acceptable.

dhellmann:
> The `dynamic-pugins.md` file is missine following section and subsections:
> 
> ```
> Checking enhancements/console/dynamic-plugins.md
> enhancements/console/dynamic-plugins.md missing "### User Stories"
> enhancements/console/dynamic-plugins.md missing "#### Dev Preview -> Tech Preview"
> enhancements/console/dynamic-plugins.md missing "#### Tech Preview -> GA"
> enhancements/console/dynamic-plugins.md missing "#### Removing a deprecated feature"
> ```
> 
> Not sure if thats acceptable.

Instead of removing sections, we want to explicitly say which ones are not applicable to a given enhancement. That makes it easier for reviewers to consider whether maybe the section should have information in it, without having to remember what all of the sections are.

Please restore the missing sections in the appropriate place based on the template.

enhancements/console/dynamic-plugins.md
```diff
@@ -272,6 +272,35 @@ available and the user must refresh their browser to see the changes. (We won't
 refresh the page automatically to avoid possibly losing data if the user is
 entering something into a form or the YAML editor.)
 
+### Localization
+
+Console's frontend will use `/locales/resource.json?lng={{lng}}&ns={{ns}}`
+as i18next loadPath.
```
spadgett, line 278: This seems like an implementation detail that we can leave out of this enhancement since it has no impact on the dynamic plugins themselves. It prevents us from having to update the enhancement proposal is the path ever changes.

enhancements/console/dynamic-plugins.md
```diff
@@ -272,6 +272,35 @@ available and the user must refresh their browser to see the changes. (We won't
...
+
+#### Dynamic Plugins
+
+Dynamic plugins will need to contain their own localization files and serve them
+per requested. All dynamic plugins must use a single localization namespace
+per language, called `default.json`.
```
spadgett, line 284: I think we'd want the namespace to match whatever the code is using (so if it's `plugin:dev-console` that filename should match). We might be careful with special characters like `:`... Maybe something like `plugin__` is a better prefix if we want to avoid conflicts?

enhancements/console/dynamic-plugins.md
```diff
@@ -272,6 +272,35 @@ available and the user must refresh their browser to see the changes. (We won't
...
+All dynamic plugins are expected to use `plugin:` namespace prefix, e.g.
+`plugin:dev-console` or `plugin:kubevirt-plugin`.
+The request for the dynamic plugin locales will be proxied by console to plugin's `Pod`,
+using it's `Service`.
+Example request path for `console-demo-plugin` locales will look like:
+`/locales/resource.json?lng=en&ns=plugin:console-demo-plugin'`.
+This request will be proxied to:
+`<service-name>.<service-namespace>.svc.cluster.local:<service-port>/locales/{language}/default.json`
+
+#### Static Plugins
```
spadgett, line 294: I'd leave the section on static plugins out of the enhancement proposal. We're working to migrate away from those, and they're loaded the same way as core console translations are.

spadgett (commented):
/cc @christianvogt @vojtechszocs

enhancements/console/dynamic-plugins.md
spadgett: Right, but I think using a different namespace in the code compare to the filename is confusing and not how i18next typically works. Simplest would be to make them the same IMO.

christianvogt: I'm indifferent to use default or the plugin name itself.

As for the `:`, what's wrong with it as a separator? Certainly read better imo.

jhadvig: > I'm indifferent to use default or the plugin name itself.

+1

> As for the :, what's wrong with it as a separator? Certainly read better imo.

yeah I think its more readable as well. not sure what is the issue

spadgett: It's the same reason I wouldn't recommend putting a space in the filename if it can be easily avoided. It needs to be escaped in URLs, and we need to be sure i18next and any dev tools we use handle special characters correctly. Have we tested that i18next-http-backend properly URL encodes `:` in the `loadPath`? Using underscores instead sidesteps the problem entirely.

I actually find `__` more readable as a separator because it gives more space between the prefix and actual name. Maybe I'm just used to BEM classnames at this point.

jhadvig:
@spadgett @christianvogt updated the enhancement per comments. PTAL

jhadvig:
Also added these four sections to make the linter happy 
```
"### User Stories"
"#### Dev Preview -> Tech Preview"
"#### Tech Preview -> GA"
"#### Removing a deprecated feature"
```

vojtechszocs:
> Also added these four sections to make the linter happy
> 
> ```
> "### User Stories"
> "#### Dev Preview -> Tech Preview"
> "#### Tech Preview -> GA"
> "#### Removing a deprecated feature"
> ```

@jhadvig FYI, I've also added these sections in my PR #820 (3 user stories, other sections left blank)

enhancements/console/dynamic-plugins.md
```diff
@@ -272,6 +274,21 @@ available and the user must refresh their browser to see the changes. (We won't
 refresh the page automatically to avoid possibly losing data if the user is
 entering something into a form or the YAML editor.)
 
+### Localization
```
spadgett, line 277: We should link to the console i18n enhancement here for background on how we do i18n.

enhancements/console/dynamic-plugins.md
```diff
@@ -272,6 +274,21 @@ available and the user must refresh their browser to see the changes. (We won't
...
+
+Dynamic plugins will need to contain their own localization files and serve them
```
spadgett, line 279: I'd preference this by saying that console uses react-i18next for i18n, and plugins must use react-i18next as well.

enhancements/console/dynamic-plugins.md
```diff
@@ -272,6 +274,21 @@ available and the user must refresh their browser to see the changes. (We won't
...
+per requested. All dynamic plugins must use a single localization namespace
+per language, named after the namespace, e.g. for `kubevirt-plugin` the filename
```
spadgett, line 281: Namespaces aren't really per language, they span langues. I'd say plugins must use a single react-i18next namespace with a link to https://www.i18next.com/principles/namespaces.

enhancements/console/dynamic-plugins.md
```diff
@@ -272,6 +274,21 @@ available and the user must refresh their browser to see the changes. (We won't
 refresh the page automatically to avoid possibly losing data if the user is
 entering something into a form or the YAML editor.)
 
+### Localization
+
+Dynamic plugins will need to contain their own localization files and serve them
+per requested. All dynamic plugins must use a single localization namespace
```
vojtechszocs, line 280: I think we should stress the fact that each dynamic plugin is expected to provide all of its localized strings in a single file (using a single i18n namespace).

```
Plugins should localize strings rendered in console UI and provide a single localization
file per language, to be served as part of the plugin's assets.

For example, a plugin named `kubevirt-plugin` would have its localization files hosted
at `/locales/{language}/kubevirt-plugin.json`.

Plugins must use `plugin__` namespace prefix for all of their localized strings, e.g.
`plugin__kubevirt-plugin~Virtual Machines`. This allows console to determine whether
the namespace is related to a dynamic plugin and fetch the appropriate resource.
```

The doc already mentions that
> The console backend will proxy the plugin assets from the `Service` using the service CA bundle.

so I'd leave out the details regarding proxy URL (`/locales/resource.json?lng={{lng}}&ns={{ns}}`).

vojtechszocs (changes requested):

enhancements/console/dynamic-plugins.md
```diff
@@ -272,6 +274,21 @@ available and the user must refresh their browser to see the changes. (We won't
...
+per language, named after the namespace, e.g. for `kubevirt-plugin` the filename
+would be `kubevirt-plugin.json`. Localization files need to be served under
+`/locales/{language}/{namespace}.json` path.
+All dynamic plugins are expected to use `plugin__` namespace prefix, e.g.
+`plugin__knative-plugin` or `plugin__kubevirt-plugin`. The request for the dynamic
```
spadgett, line 285: Maybe use `plugin__knative` and `plugin__kubevirt` as examples? I think in most cases you wouldn't have `-plugin` in the plugin name.

A code snippet showing a real example would be helpful.

```ts
const VMHeading = () => {
  const { t } = useTranslation();
  return <h1>{t('plugin__kubevrit~Virtual Machine')}</h1>;
};
```

enhancements/console/dynamic-plugins.md
spadgett, line 280: nit: I'd leave out "and serve them when requested." I think that should be apparent, and we have a specific example below.

spadgett (approved):
/approve

The approach looks good to me, just a few comments on the text.

enhancements/console/dynamic-plugins.md
```diff
@@ -272,6 +274,21 @@ available and the user must refresh their browser to see the changes. (We won't
...
+plugin locales will be proxied by console to plugin's `Pod`, using it's `Service`.
+Example request path for `console-demo-plugin` locales will look like:
+`/locales/resource.json?lng=en&ns=plugin__console-demo-plugin'`.
+This request will be proxied to:
+`<service-name>.<service-namespace>.svc.cluster.local:<service-port>/locales/{language}/{namespace}.json`
```
spadgett, line 290:
```suggestion
`<service-name>.<service-namespace>.svc.cluster.local:<service-port>/locales/{language}/plugin__console-demo-demo.json`
```

enhancements/console/dynamic-plugins.md
jhadvig: @spadgett hmmm, but wont the locale file contain the `-plugin` string? I though it would. Or are we counting that it will contain it and in that case we done need to `ns` to contain it ?

enhancements/console/dynamic-plugins.md
jhadvig: Im a bit confused here based on https://github.com/openshift/enhancements/pull/815#discussion_r662469872. shall the `-plugin` be part of the namespace or not ?

enhancements/console/dynamic-plugins.md
spadgett: I'm saying usually the plugin name will be something like `kubevrit` instead of `kubevirt-plugin`. Adding `-plugin` on the end is a bit redundant. We don't typically name deployments `kubevirt-deployment` and pods `kubevirt-pod` for example. If there is no `-plugin` suffix in the `ConsolePlugin` resource name, then it wouldn't be in the JSON filename or i18n namespace.

enhancements/console/dynamic-plugins.md
spadgett: Only if the `ConsolePlugin` resource name has `-plugin` in it. In this case, the plugin name is `console-demo-plugin` (as stated above in line 287), but for most plugins I'd expect `-plugin` *not* to be in the name.

I think it would be better to use a real example here like `kubevirt` instead of `console-demo-plugin`.

jhadvig:
@spadgett comments addressed. PTAL

enhancements/console/dynamic-plugins.md
```diff
@@ -350,6 +350,33 @@ available and the user must refresh their browser to see the changes. (We won't
...
 
+### Localization
+
+Info on how Console is handling i18n is in this [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/console/internationalization.md).
+
+Console uses [react-i18next](https://github.com/i18next/react-i18next) for i18n,
+and dynamic plugins must use react-i18next as well.
+
+All dynamic plugins must use a single react-i18next [namespace](https://www.i18next.com/principles/namespaces),
+named after the namespace, e.g. for `kubevirt` the filename would be
```
spadgett, line 361:
```suggestion
named after the plugin, e.g. for `kubevirt` the filename would be
```

enhancements/console/dynamic-plugins.md
```diff
@@ -350,6 +350,33 @@ available and the user must refresh their browser to see the changes. (We won't
...
+`/locales/{language}/{namespace}.json` path.
+All dynamic plugins are expected to use `plugin__` namespace prefix, e.g.
+`plugin__knative` or `plugin__kubevirt`. The request for the dynamic
+plugin locales will be proxied by console to plugin's `Pod`, using it's `Service`.
+Example request path for `knative` plugin locales will look like:
+`/locales/resource.json?lng=en&ns=plugin__knative'`.
+This request will be proxied to:
+`<service-name>.<service-namespace>.svc.cluster.local:<service-port>/locales/{language}/{namespace}.json`
+
+The following is an example of kubevirt plugin locales usage in code:
```
spadgett, line 372:
```suggestion
Here's a code example of how the `kubevirt` plugin would translate a message: 
```

enhancements/console/dynamic-plugins.md
```diff
@@ -350,6 +350,33 @@ available and the user must refresh their browser to see the changes. (We won't
...
+
+Info on how Console is handling i18n is in this [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/console/internationalization.md).
+
+Console uses [react-i18next](https://github.com/i18next/react-i18next) for i18n,
+and dynamic plugins must use react-i18next as well.
+
+All dynamic plugins must use a single react-i18next [namespace](https://www.i18next.com/principles/namespaces),
+named after the namespace, e.g. for `kubevirt` the filename would be
+`plugin__kubevirt.json`. Localization files need to be served under
+`/locales/{language}/{namespace}.json` path.
```
spadgett, line 363:
```suggestion
`plugin__kubevirt.json`. Localization resources need to be served
by the plugin service under the
`locales/{language}/{namespace}.json` path relative to the `basePath`
defined in the `ConsolePlugin` resource.
```

enhancements/console/dynamic-plugins.md
```diff
@@ -350,6 +350,33 @@ available and the user must refresh their browser to see the changes. (We won't
...
+All dynamic plugins are expected to use `plugin__` namespace prefix, e.g.
+`plugin__knative` or `plugin__kubevirt`. The request for the dynamic
+plugin locales will be proxied by console to plugin's `Pod`, using it's `Service`.
+Example request path for `knative` plugin locales will look like:
+`/locales/resource.json?lng=en&ns=plugin__knative'`.
+This request will be proxied to:
+`<service-name>.<service-namespace>.svc.cluster.local:<service-port>/locales/{language}/{namespace}.json`
```
spadgett, line 370: I'd keep the plugin example consistent with the code example below, so let's use kubevirt.

```suggestion
plugin localization resources will be proxied by console backend.
For example, the `kubevirt` plugin localization resource
in the `en` language will be requested at path
`/locales/en/plugin__kubevirt.json`
```

spadgett (commented):
@jhadvig Thanks. Don't forget to change the `last-updated` date at the top of the enhancement.

enhancements/console/dynamic-plugins.md
```diff
@@ -350,6 +350,33 @@ available and the user must refresh their browser to see the changes. (We won't
...
+
+Console uses [react-i18next](https://github.com/i18next/react-i18next) for i18n,
+and dynamic plugins must use react-i18next as well.
+
+All dynamic plugins must use a single react-i18next [namespace](https://www.i18next.com/principles/namespaces),
+named after the plugin, e.g. for `kubevirt` the filename would be
+`plugin__kubevirt.json`. Localization resources need to be served
+by the plugin service under the `locales/{language}/{namespace}.json`
+path relative to the `basePath` defined in the `ConsolePlugin` resource.
+All dynamic plugins are expected to use `plugin__` namespace prefix, e.g.
```
spadgett, line 365: Nit: This isn't only an expectation, it's a requirement

```suggestion
All dynamic plugins must use the `plugin__` namespace prefix, e.g.
```

spadgett (commented):
Sorry, just one point of clarification I missed in the last review. It's important to note that the plugin prefix is required, not just a convention.

jhadvig:
@spadgett comment addressed :)

spadgett (approved):
/lgtm

PR #816, opened 2021-06-28
Ingress: Add API for configuring ingress connection timeouts

rfredette:
This enhancement describes several new API fields to configure the length of the timeout period for connections through the router

### DIFF
--- a/enhancements/ingress/request-timeout-configuration.md
+++ b/enhancements/ingress/request-timeout-configuration.md
@@ -19,6 +19,7 @@ creation-date: 2021-06-17
 last-updated: 2021-07-14
 status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced|informational
 see-also:
+  - https://github.com/openshift/enhancements/pull/461
 replaces:
 superseded-by:
 ---
@@ -27,9 +28,9 @@ superseded-by:
 
 ## Release Signoff Checklist
 
-- [ ] Enhancement is `implementable`
-- [ ] Design details are appropriately documented from clear requirements
-- [ ] Test plan is defined
+- [X] Enhancement is `implementable`
+- [X] Design details are appropriately documented from clear requirements
+- [X] Test plan is defined
 - [ ] Operational readiness criteria is defined
 - [ ] Graduation criteria for dev preview, tech preview, GA
 - [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
@@ -38,7 +39,7 @@ superseded-by:
 
 This document proposes the addition of several API fields into the ingress
 controller API, allowing the cluster admin to adjust how long routers will hold
-open connections that are awaiting a response.
+open connections under various circumstances.
 
 ## Motivation
 
@@ -53,11 +54,13 @@ Allow admins to configure the following HAProxy connection timeout parameters:
 - timeout client-fin
 - timeout server
 - timeout server-fin
+- timeout tunnel
 - tcp-request inspect-delay
 
 ### Non-Goals
 
 - Allow admins to configure every HAProxy timeout parameter
+- Allow users to configure timeouts per route
 
 ## Proposal
 
@@ -69,6 +72,7 @@ spec:
     clientFinTimeout: "1s"
     serverTimeout: "30s"
     serverFinTimeout: "1s"
+	tunnelTimeout: "1h"
     tlsInspectDelay: "10s"
 ```
 
@@ -100,11 +104,11 @@ responding to them.
 #### Router Image Environment Variables
 
 `timeout client`, `timeout client-fin`, `timeout server`, and `timeout
-server-fin` were all configurable in 3.X, and the router image still supports
-configuring them from the environment variables
+server-fin` `timeout tunnel` were all configurable in 3.X, and the router image
+still supports configuring them from the environment variables
 `ROUTER_DEFAULT_CLIENT_TIMEOUT`, `ROUTER_CLIENT_FIN_TIMEOUT`,
-`ROUTER_DEFAULT_SERVER_TIMEOUT`, and `ROUTER_DEFAULT_SERVER_FIN_TIMEOUT`,
-respectively.
+`ROUTER_DEFAULT_SERVER_TIMEOUT`, `ROUTER_DEFAULT_SERVER_FIN_TIMEOUT`, and
+`ROUTER_DEFAULT_TUNNEL_TIMEOUT` respectively.
 
 A new environment variable will be defined, `ROUTER_INSPECT_DELAY`, which will
 control the `tcp-request inspect-delay` variable.
@@ -146,7 +150,7 @@ this proposal are set sufficiently high.
 
 ### Open Questions
 
-- How do these timeout variables affect websocket connections (if at all)?
+None
 
 ### Test Plan
 
@@ -159,6 +163,7 @@ spec:
     clientFinTimeout: "3s"
     serverTimeout: "60s"
     serverFinTimeout: "4s"
+	tunnelTimeout: "30m"
     tlsInspectDelay: "5s"
 ```
 - verify that the router deployment contains the following values:
@@ -177,6 +182,8 @@ spec:
           value: "60s"
         - name: "ROUTER_DEFAULT_SERVER_FIN_TIMEOUT"
           value: "4s"
+		- name: "ROUTER_DEFAULT_TUNNEL_TIMEOUT"
+		  value: "30m"
         - name: "ROUTER_TLS_INSPECT_DELAY"
           value: "5s"
         ...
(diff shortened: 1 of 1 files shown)


enhancements/ingress/request-timeout-configuration.md
```diff
@@ -0,0 +1,183 @@
...
+
+TODO
+
+## Proposal
+
+Add the following configurable variables to the ingresscontroller API:
+```yaml
+spec:
+  tuningOptions:
+    clientTimeout: "30s"
```
sgreene570, line 63: Is there a better way to express these timeouts than strings? If not, we should make sure to define the format and what limits HAProxy might place on these values come API design time.

sgreene570: (my main concern is someone specifying a value that breaks HAProxy reloads such as "99999d" on accident)

enhancements/ingress/request-timeout-configuration.md
```diff
@@ -0,0 +1,183 @@
...
+
+A new environment variable will be defined, `ROUTER_INSPECT_DELAY`, which will
+control the `tcp-request inspect-delay` variable. If no unit is specified, the
+value represents the delay in milliseconds, but time units can be specified,
+following the time format outlined in [section 2.4. of the HAProxy 2.2
+documentation](https://github.com/haproxy/haproxy/blob/v2.2.0/doc/configuration.txt).
+
+### Risks and Mitigations
+
+TODO
```
sgreene570, line 94: Do cloud LB timeouts come into play here at all?

enhancements/ingress/request-timeout-configuration.md
rfredette: I think it's tough to validate the field at the API level because I don't know of a good way to allow users to specify units while also validating that the timeout is not longer than the max allowed value.

If validation at the API level is preferred, input could be taken as an integer, assuming a particular time unit (my inclination would be `ms`), which would make it easy to have min/max values, but without being able to specify the units, there's more chance the user making a conversion error when inputting the timeouts.

Another option is to follow the same policy as the hard-stop-after timeout, where the ingress operator accepts any valid time as a string in the HAProxy time format, but if the value is higher than the max allowed, it gets truncated to the max value (i.e. `99999d` would become `2147483647ms`) and an error is logged.

Between the two, I'd prefer the latter behavior since it improves readability and reduces the chance of user error, but I'm open to other perspectives

sgreene570: Ah, right, I forgot about our timeout value clipping logic. Maybe we could take this oppurtunity to move that to a more formal location (I think in the past we talked about moving the HAProxy timeout clip function to library-go, or something along those lines).

sgreene570: aha! https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/deployment.go#L1262-L1263

enhancements/ingress/request-timeout-configuration.md
```diff
@@ -0,0 +1,231 @@
...
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This document proposes the addition of several API fields into the ingress
+controller API, allowing the cluster admin to adjust how long routers will hold
+open connections that are awaiting a response.
```
Miciah, line 41:
```suggestion
open connections under various circumstances.
```
These settings include request deadline and client and server inactivity timeouts, so it's a broader set of options than just response deadline.

enhancements/ingress/request-timeout-configuration.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Motivation
+
+Customers need the ability to adjust how long HAProxy holds connections open,
+either by extending the timeout to accomodate slower backends or clients, or by
```
Miciah:
```suggestion
either by extending the timeout to accommodate slower backends or clients, or by
```

enhancements/ingress/request-timeout-configuration.md
```diff
@@ -0,0 +1,231 @@
...
+Allow admins to configure the following HAProxy connection timeout parameters:
+- timeout client
+- timeout client-fin
+- timeout server
+- timeout server-fin
+- tcp-request inspect-delay
+
+### Non-Goals
+
+- Allow admins to configure every HAProxy timeout parameter
```
Miciah, line 60: Maybe also:
````
- Allow users to configure per-route timeouts.
```

enhancements/ingress/request-timeout-configuration.md
```diff
@@ -0,0 +1,231 @@
...
+  tuningOptions:
+    clientTimeout: "30s"
+    clientFinTimeout: "1s"
+    serverTimeout: "30s"
+    serverFinTimeout: "1s"
+    tlsInspectDelay: "10s"
+```
+
+All variables expect time values. If no units are specified, the router assumes
+milliseconds.
```
Miciah, line 76: The Go definition should use the `metav1.Duration` type, which is parsed using `time.ParseDuration`:
```go
// Duration is a wrapper around time.Duration which supports correct
// marshaling to YAML and JSON. In particular, it marshals into strings, which
// can be used as map keys in json.
type Duration struct {
	time.Duration `protobuf:"varint,1,opt,name=duration,casttype=time.Duration"`
}
```
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/duration.go#L24-L29

If you use `metav1.Duration`, the unit suffix is required.

enhancements/ingress/request-timeout-configuration.md
```diff
@@ -0,0 +1,231 @@
...
+days, 20 hours). If a time greater than the maximum is specified for any of the
+API fields added in this proposal, the timeout will be set to the maximum
+allowed value, and an error message will be logged by the ingress operator.
+
+### Risks and Mitigations
+
+Setting timeout values in the ingress controller will only affect timeouts
+caused by ingress. Other hops in the connection, such as the backend or the
+cloud LB, may still cause connections to time out even if the fields added in
+this proposal are set sufficiently high.
```
Miciah, line 143: Related: https://github.com/openshift/enhancements/pull/461 (it might be worth adding a `see-also:` link in the metadata at the top of this file).

enhancements/ingress/request-timeout-configuration.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Design Details
+
+### Open Questions
+
+- How do these timeout variables affect websocket connections (if at all)?
```
Miciah, line 149: The timeout for WebSocket connections can be configured using `timeout tunnel`, for which we have the `ROUTER_DEFAULT_TUNNEL_TIMEOUT` environment variable.  It would make sense to add an API field to configure that as well.

Miciah, line 149: How about `hard-stop-after`?  We added an annotation to configure `hard-stop-after` in https://github.com/openshift/cluster-ingress-operator/pull/522, but a formal API field would be useful.

enhancements/ingress/request-timeout-configuration.md
Miciah: By the way, https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions is relevant here.  It looks like the names you chose are consistent with the guidance there.  (There is some interesting nuance between "timeouts" and "deadlines",  but I think these are all timeouts.)

rfredette: Good to know; I was not aware of the deadline vs timeout guidance, but these fields do seem to align with the timeout definition.

enhancements/ingress/request-timeout-configuration.md
rfredette: It would probably be good to implement `hard-stop-after` in the API, but since it's already configurable with the annotation, let's leave that addition for a later date.

enhancements/ingress/request-timeout-configuration.md
```diff
@@ -69,6 +72,7 @@ spec:
     clientFinTimeout: "1s"
     serverTimeout: "30s"
     serverFinTimeout: "1s"
+	tunnelTimeout: "1h"
```
Miciah, line 75: The indentation is inconsistent on this line (tab instead of spaces).

enhancements/ingress/request-timeout-configuration.md
```diff
@@ -0,0 +1,254 @@
...
+`spec.tuningOptions.serverFinTimeout` in the ingresscontroller API to a lower
+value, forcing those connections to close sooner if my application stops
+responding to them.
+
+### Implementation Details/Notes/Constraints
+
+#### Router Image Environment Variables
+
+`timeout client`, `timeout client-fin`, `timeout server`, and `timeout
+server-fin` `timeout tunnel` were all configurable in 3.X, and the router image
```
Miciah, line 107:
```suggestion
`timeout client`, `timeout client-fin`, `timeout server`, `timeout
server-fin`, and `timeout tunnel` were all configurable in 3.X, and the router image
```

enhancements/ingress/request-timeout-configuration.md
rfredette: Good catch. It looks like my editor didn't do tabs-to-spaces conversion for some reason...

enhancements/ingress/request-timeout-configuration.md
```diff
@@ -177,6 +182,8 @@ spec:
           value: "60s"
         - name: "ROUTER_DEFAULT_SERVER_FIN_TIMEOUT"
           value: "4s"
+		- name: "ROUTER_DEFAULT_TUNNEL_TIMEOUT"
+		  value: "30m"
```
Miciah, line 186: Lines 166, 185, and 186 also have tabs instead of spaces.

enhancements/ingress/request-timeout-configuration.md
Miciah: Might be worth adding as a non-goal.

Miciah:
Thanks!
/lgtm
/accept

PR #817, opened 2021-06-28
Baremetal IPI Network Configuration for Day-1

cybertron:
Describe user-facing API for day-1 network customizations in the IPI workflow,
with particular focus on baremetal where such configuration is a common
requirement.

Co-authored-by: Steve Hardy <shardy@redhat.com>


enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Goals
+
+* Define API for day-1 network customizations
+* Enable common on-premise network configurations (bond+vlan, static ips) via IPI
+
+### Non-Goals
+
+* Platforms other than `baremetal`, although the aim is a solution which could be applied to other platforms in future if needed.
+* Provide a consistent (ideally common) user API for deployment and post-deployment configuration. Getting agreement on a common API for day-1 and day-2 has proven contentious and will be addressed later.
```
hardys, line 66: We should perhaps qualify the statement re common API, maybe switch the order of this bullet with the "Enabling kubernetes-nmstate by default", then here we can say something like:

"Getting agreement on a common API for day-1 and day-2 has stalled due to lack of consensus around enabling kubernetes-nmstate APIs (which are the only API for day-2 currently) by default"

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+* Enabling kubernetes-nmstate by default for day-2 networking is discussed via
+[another proposal](https://github.com/openshift/enhancements/pull/747)
+* Handling of advanced networking options such as bridges and bonds. The implementation may support them, but they are not the priority for this work.
```
hardys, line 69: I'm not sure this is correct - the #1 request for baremetal is bond+vlan, since it's the cheapest way to achieve both a dedicated provisioning network and a highly available MachineNetwork.

Maybe this should say something like "Advanced network configuration which is not achievable via simple network manager keyfiles?"

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+for reasons of network ownership or concerns over reliability I can't use DHCP
+and therefore need to provide a static configuration for my primary network.
+
+There is no way to provide [MachineSpecific Configuration in OpenShift](https://github.com/openshift/machine-config-operator/issues/1720) so I am
+forced to use the UPI flow which is less automated and more prone to errors.
+
+### Risks and Mitigations
+
+In some existing workflows, kubernetes-nmstate is used to do network configuration on day-2. Using a different interface for day-1 introduces the potential for mismatches and configuration errors when making day-2 changes. It should be possible to mitigate this by generating the NetworkManager keyfiles using the nmstatecli, which will allow the same nmstate config to be used directly in the day-2 configuration.
+
```
hardys, line 102: We could also mention here that such a documented approach could also provide an easier path for users if/when a common nmstate based API is available in future?

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+    id=eth0
+    uuid=18e0cec7-041c-4fb3-957c-a60c80dd9b85
+    type=ethernet
+    etc...
+master-1:
+  eth0:
+    [connection]
+    id=eth0
+    etc...
+```
```
hardys, line 145: I think the current plan is to enable this interface only for the `baremetal` platform, so there may be concerns about adding a new top-level install-config section.  In that case, we could instead add the config to the platform-specific `hosts` list (which already contains other host-specific data needed to automate deployment)

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+
+#### API Open Questions
+
+* How do we map an individual machine's keyfiles to the machine? Hostname was previously suggested because it was already an available label in NNCP, but if we're not using NNCP then there might be a better way.
+
```
hardys, line 150: I think the current plan is to initially enable this interface only for the `baremetal` platform, so there may be concerns about adding a new top-level install-config section.  

In that case, we could instead add the config to the platform-specific `hosts` list (which already contains other host-specific data needed to automate deployment), and this also solves the mapping to a host (since we're already consuming this per-host data in the `baremetal` flow, e.g BMC credentials)

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+* If we're using RHCOS functionality, do we still need the custom image build discussed below? Or can we just pass the keyfiles as part of the deployment process?
+
+* What is the best way to encode the data above into secret(s)?
```
hardys, line 153: @dtantsur @zaneb - can you help with this question?  I'm guessing we'll need the metal3 image building tool to reference a secret containing the config (either per-host or per-cluster) - but here we're talking about a secret containing some encoded NM keyfiles.

In the [upstream design](https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/image-builder-integration.md#design-details) we talk about a networkData secret, which could imply the openstack-specific networkData schema, I guess we can have an image builder backend which just consumes keyfiles instead?

Then in future if/when we can make progress on the question of a common day1/day2 API we could swap that out for something that can build from some nmstate derived data/resource to generate the secret.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Processing user configuration
+
+#### Deployment of the controlplane hosts via terraform
+
+TODO - we'll have to match the keyfiles to the master Machines, then generate image/images for deployment via terraform/ironic
```
hardys, line 159: I think we can do this by consuming the install-config data, then generating some files to be consumed on the bootstrap VM by the [ironic startup script](https://github.com/openshift/installer/blob/master/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template) - this can then generate the images for each controlplane host prior to starting any ironic services (terraform waits for ironic-api to start before attempting to deploy)

This raises the question of if we can detect when the configuration is common for all hosts - it'll be simpler to just generate 3 images (with predictable filenames so we can pass the necessary URL to terraform/Ironic), but obviously that wastes some disk space in the bond+vlan case (where the same config will be applied to all 3 master hosts).

@kirankt I think you've been looking at this area, any thoughts?

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+
+#### Deployment of compute/worker hosts
+
+TODO - this will rely on metal3 image-building proposal https://github.com/metal3-io/metal3-docs/pull/183
```
hardys, line 163: @zaneb can you help add more detail to this section?

enhancements/network/baremetal-ipi-network-configuration.md
kirankt: I agree with Steve. This should belong in the hosts section of the baremetal platform. e.g.
```
platform:
  baremetal:
    hosts:
      - name: openshift-master-0
        networkData:
          eth0:
            [connection]
            id=eth0
            uuid=18e0cec7-041c-4fb3-957c-a60c80dd9b85
            type=ethernet
etc...     
```

enhancements/network/baremetal-ipi-network-configuration.md
kirankt: I am assuming that the installer will create this secret asset. We need to figure out which namespace this will end up in.

enhancements/network/baremetal-ipi-network-configuration.md
kirankt: Right now we consume one rhcos image for all control plane nodes from the rhcos-stream data. I don't think it will be necessary to send separate copies of the images for each control plane node. All we have to do is send separate network data to the ironic agent and the agent should perform the copy network action independently.

https://github.com/dtantsur/ironic-agent-image/blob/08c3803db844e5bd50d21017c7f0c81158dc6a1a/hardware_manager/ironic_coreos_install.py#L80

Although, I'm not sure why we're using metadata here rather than just passing networkdata directly. Perhaps @dtantsur can clarify.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+to enable such configuration in the IPI flow.
+
+## Motivation
+
+Since the introduction of baremetal IPI, a very common user request is how
+to configure day-1 networking, and in particular the following cases which are not currently possible:
+
+* Deploy with OpenShift Machine network on a tagged (non-default) VLAN
+* Deploy with OpenShift Machine network using static IPs (no DHCP)
+
```
racedo, line 50: Deploy with OpenShift Machine network using bonded NICs (and often together with tagged VLANs) is a common user request at the same level of importance. I know it's mentioned below and if having to choose, having static IPs would be a higher prio but static/bonds/VLANs are the basic most commonly required configurations.

We also need options for setting MTUs and for bonding policies.

enhancements/network/baremetal-ipi-network-configuration.md
racedo: Fully agree with Steve on the need for bond and VLAN as the basic most required config (obviously, along with static IPs).

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a baremetal IPI user, I want to deploy via PXE and achieve a highly
+available Machine Network configuration in the most cost/space effective
+way possible.
+
```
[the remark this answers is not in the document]
cybertron: Allowing users to statically assign addresses in the provisioning DHCP server is probably possible, but I'm not sure it's worth tackling at this time. Users who are uncomfortable with us assigning the provisioning addresses can use virtualmedia and skip DHCP entirely. The implementation of that would also be completely separate from what we're currently proposing. It wouldn't be configuration applied to the nodes, it would be configuration applied to the provisioning DHCP server.

So for the initial implementation I would be inclined to only focus on your first and third cases. If there's significant requests for the second option we can always implement that in the future.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Sure, I can do that.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Okay, this sounds reasonable. I'd be more concerned about using a baremetal-specific structure if we had the common nmstate API, but since this implementation isn't going to be useful to other groups like OVNK anyway I'm good with it.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Sounds like a reasonable optimization, but if we don't get it for the initial implementation it will still work so I'd consider this a stretch goal if we manage to finish this feature early. :-)

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Okay, these two points were the primary motivations for this work, but I'll try to capture some of the other things we'd also like to configure as well.

enhancements/network/baremetal-ipi-network-configuration.md
zaneb: https://gist.github.com/zaneb/9d809ab69753f170c91b2726ab565697

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+However, we have been unable to get agreement on a method to deliver the operator on day-1, which makes this problematic. It's possible we could implement some way to pull just the NNCP CRD from the operator payload and isntall it on day-1, but at this point we likely don't have time to do that and it's not clear we even want to if the operator can't be installed on day-1 too. This would essentially orphan the CRs after the initial deployment completes.
+
+### Provide raw nmstate data
+
+We could expose an interface to the user which accepts raw nmstate data (not a CR wrapper).  This would mean a common DSL to the NodeNetworkConfigurationPolicy resource, but not directly reusing that API, so there is risk of drift between the two APIs.
+
+The user would have to pass the nmstate data via the install-config, and also as a Secret for use on scale-up, then also define a NNCP resource if day-2 management of the configuration is required (which could be risky if the configuration isn't exactly the same...)
```
zaneb, line 177: Couldn't the installer set up all of these?

Or the CAPBM could create the Secret to use for day-1 automatically from a NNCP resource that is to be used for day-2.

enhancements/network/baremetal-ipi-network-configuration.md
racedo: @cybertron oh ok, so the workflow you have in mind could be basic host network config on day 1 and then have it fully configured right after finishing the install with knmstate? If this made sense, we could pass knmstate configurations with the installer to be applied once the knmstate operator is deployed (and automate its deployment as part of the install workflow). On the flip side, users would have to deal with a combination of NetworkManager keyfiles and nmstate, unless some conversion of sorts could be implemented.

hardys: @racedo ideally we'd pass in the same data consumed by kubernetes-nmstate to the installer (indeed that was our [original plan](https://github.com/hardys/enhancements/blob/day1_networking/enhancements/installer/network_configuration.md))

However discussion on #747 has stalled, and I think based on conversations with @dhellmann and @derekwaynecarr we're not getting knmstate (and associated NNCP CRD API) into the release payload.

So, we need an interim solution which does mean users have to deal with a combination of keyfiles and nmstate.

There is a conversion possible using the nmstate CLI (which @cybertron mentions in this PR), so I think we'll just have to document that until a more fully-integrated solution is possible (either adding knmstate to the release payload, or more likely adding some way to automatically enable OLM operators at install-time)

enhancements/network/baremetal-ipi-network-configuration.md
hardys: I guess that raises the question (already mentioned in the alternatives) of whether we want this to be nmstate format or just inlined keyfiles?  I guess we need the interface to enable adding nmstate support even if we start with keyfiles.

Also for those familiar with OpenStack `networkData` implies a different format - maybe we should call this something else e.g `networkConfig` or `networkFiles`  (then adding e.g `nmstateConfig` later would be simple)?

enhancements/network/baremetal-ipi-network-configuration.md
hardys: Ack thanks - I think we have to decide between keyfiles and the nmstate DSL I guess - if the builder could consume the nmstate data, transform to keyfiles then embed that gets us closer to the day-2 kubernetes-nmstate API.

But then we would still have two types of nmstate data persisted - that may result in problems keeping nmstate versions/format in sync (between the builder container and kubernetes-nmstate)

enhancements/network/baremetal-ipi-network-configuration.md
hardys: Ok, sounds like we may have a few possible optimizations here - maybe we could add them to the PR as possible future work so we don't lose track?

enhancements/network/baremetal-ipi-network-configuration.md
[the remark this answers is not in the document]
hardys: IMO we should treat the provisioning network as out-of-scope for this PR - there are already ways for customers to control that network (either by disabling it and using virtual-media, or by making that network [Unmanaged](https://docs.openshift.com/container-platform/4.7/installing/installing_bare_metal_ipi/ipi-install-installation-workflow.html#additional-install-config-parameters_ipi-install-configuration-files) and configuring their own DHCP server)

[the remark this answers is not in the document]
hardys: @cybertron perhaps add configuration of the provisioning network as an explicit non-goal?

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+means of providing the configuration to the OS is required.
+
+In the UPI flow this is achieved by consuming user-provided NetworkManager
+keyfiles, as an input to `coreos-install --copy-network`, but there is
+no corresponding user interface at the openshift-install level.
+
+### Goals
+
+* Define API for day-1 network customizations
+* Enable common on-premise network configurations (bond+vlan, static ips) via IPI
```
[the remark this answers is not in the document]
hardys: Yeah I guess bond configuration could differ between hosts or groups of hosts (different nic names etc) - I think the API described here is basically always config per-host, but we can then perhaps optimize by detecting common config (to avoid duplicate images).

I also wonder if we have per-host config in the install-config.yaml if we can make yaml anchors/aliases work so that e.g in the case where you have a single bond/vlan config for every host you don't have to copy/paste it for each host.

enhancements/network/baremetal-ipi-network-configuration.md
hardys: The problem is ref #747 that the NNCP CRD provided via kubernetes-nmstate doesn't exist at the point of install - it's currently only applied if/when you install the operator day-2 from OLM.

We could look at making CAPBM look for such a resource in future though, then fall back to using secrets (which might resolve some of the concerns around day1/day2 interface overlap is we can make it clear to users which API they should interact with)

enhancements/network/baremetal-ipi-network-configuration.md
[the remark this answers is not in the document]
racedo: Ack, that makes sense.

enhancements/network/baremetal-ipi-network-configuration.md
zaneb: Ah, I was missing that context.
I wonder if the objections to #747 could be overcome by saying we'll enable the operator at install time whenever the user specifically provides network configuration in the installer, rather than always on particular platforms.

hardys: @zaneb perhaps - the impression I got from discussion with @dhellmann and @derekwaynecarr is that we don't want knmstate in the release payload at all, because it's not relevant to all platforms (although arguably it could be e.g to support secondary interfaces for OCS) (there are also some unresolved questions around support status and ownership to deal with).

But if we're going to block core-platform functions like network configuration from the release payload long term, it probably does mean we need some install-time way to enable optional dependencies from OLM or other sources.

dankenigsberg, line 177: I think that this problem, of having 3 different copies of the same thing at
* install-config
* secret
* day-2 NNCP

exists also in the proposed solution; it's not a special downside of this alternative.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Use Kubernetes-NMState NodeNetworkConfigurationPolicy custom resources
+
+If we were able to install the NNCP CRD at day-1 then we could use that as the configuration interface. This has the advantage of matching the configuration syntax and objects used for day-2 network configuration via the operator.
+
+However, we have been unable to get agreement on a method to deliver the operator on day-1, which makes this problematic. It's possible we could implement some way to pull just the NNCP CRD from the operator payload and isntall it on day-1, but at this point we likely don't have time to do that and it's not clear we even want to if the operator can't be installed on day-1 too. This would essentially orphan the CRs after the initial deployment completes.
+
+### Provide raw nmstate data
+
+We could expose an interface to the user which accepts raw nmstate data (not a CR wrapper).  This would mean a common DSL to the NodeNetworkConfigurationPolicy resource, but not directly reusing that API, so there is risk of drift between the two APIs.
```
dankenigsberg, line 175: @tyll & @cathay4t may correct me, but I think that NetworkManager keyfiles are similarly susceptible to drift. For example, I'm not sure that the NM team would be happy to learn that secret created for RHEL-8 NM are being used with RHEL-9 workers. NMState exists in order to provide a stable, human-readable representation of node network state; and in that it is more committed for backward and forward compatibility.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+
+#### Deployment of compute/worker hosts
+
+TODO - this will rely on metal3 image-building proposal https://github.com/metal3-io/metal3-docs/pull/183
+
+## Alternatives
+
+### Use Kubernetes-NMState NodeNetworkConfigurationPolicy custom resources
+
+If we were able to install the NNCP CRD at day-1 then we could use that as the configuration interface. This has the advantage of matching the configuration syntax and objects used for day-2 network configuration via the operator.
```
dankenigsberg, line 169:
```suggestion
If we were able to install the (NNCP CRD)[https://nmstate.io/kubernetes-nmstate/user-guide/102-configuration] at day-1 then we could use that as the configuration interface. This has the advantage of matching the configuration syntax and objects used for day-2 network configuration via the operator.
```

enhancements/network/baremetal-ipi-network-configuration.md
dankenigsberg, line 177:
```suggestion
The user would have to pass the nmstate data via the install-config, and also as a Secret for use on scale-out, then also define a NNCP resource if day-2 management of the configuration is required (which could be risky if the configuration isn't exactly the same...)
```

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Provide raw nmstate data
+
+We could expose an interface to the user which accepts raw nmstate data (not a CR wrapper).  This would mean a common DSL to the NodeNetworkConfigurationPolicy resource, but not directly reusing that API, so there is risk of drift between the two APIs.
+
+The user would have to pass the nmstate data via the install-config, and also as a Secret for use on scale-up, then also define a NNCP resource if day-2 management of the configuration is required (which could be risky if the configuration isn't exactly the same...)
+
+#### Create a net-new NMState Wrapper CR
+
+The [assisted-service](https://github.com/openshift/assisted-service/blob/master/docs/crds/nmstate.yaml) has created a new NMState wrapper CR.
```
dankenigsberg, line 181: I think that they just moved those docs

```suggestion
The [assisted-service](https://github.com/openshift/assisted-service/blob/0b0e3677ae83799151d11f1267cbfa39bb0c6f2e/docs/hive-integration/crds/nmstate.yaml) has created a new NMState wrapper CR.
```

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: If we use nmstate yaml for install-config then we need nmstate available on the provisioning host so we can convert it to keyfiles. I'm fine with doing that (it's how we'd have to process NNCP records too), but it does add a dependency. I'm not sure how controversial that will be. It's also a bit more complex implementation-wise than straight keyfiles.

One option would be to replace the interface name with the filename, i.e. eth0 -> eth0.nmconnection. That's going to have to happen at some point anyway, and then we could potentially add support for eth0.yaml, which would indicate nmstate data. But maybe that's a little too magic?

+1 to not overloading networkData.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Thanks!

cybertron (commented):
I'll push my updated local copy of the doc later when we've had a chance to wrap up some of the existing conversations here.

enhancements/network/baremetal-ipi-network-configuration.md
hardys: The install-config data doesn't exist after installation, but yes I agree - we'll potentially have two copies of network configuration which isn't ideal.

Until we get a fully supported kubernetes-nmstate and an API derived from that included by default in OCP, I don't see how this can be avoided though.  We had a [previous draft proposal](https://github.com/hardys/enhancements/blob/b6a348dc7951b792fc338b1d0a237b023f18ec88/enhancements/installer/network_configuration.md) which aimed to use NNCP for both day-1 and day-2 - but unless that API is available on day-1 (which seems unlikely given #747 is stalled) that doesn't seem to be a viable option.

enhancements/network/baremetal-ipi-network-configuration.md
hardys: That's a good point - given that it's python based I think it's highly unlikely we'll get the nmstate CLI into RHCOS - we'd need to run it inside a container.

As @dankenigsberg mentions below one advantage of using nmstate would be it may be easier to handle changes in NM configuration syntax after upgrades, but until #201 is resolved we don't actually lifecycle bootimages on upgrade (presumably this will be needed before any el8->el9 based RHCOS upgrade though)

On balance I'm leaning towards starting with keyfiles, but then adding nmstate support in a future release - we can bikeshed the interface later but yes the connection suffix does sound a little magical :)  I'd maybe prefer a new section to support the alternative syntax

enhancements/network/baremetal-ipi-network-configuration.md
dankenigsberg: We agree about this. But it is wrong to attach this to this alternative; this text should move up.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: This is very interesting. One of the objections to enabling NMState was that keyfiles provide the same functionality and the same backward compatibility. If that's not true then it eliminates one of the major arguments against NMState.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Starting with keyfiles makes a lot of sense. On some level we'll need keyfiles for day 1 since that's what we'll pass in to the images. Layering another interface over that should be much simpler once we have keyfile support.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Yeah, this entry was largely referring to a design we've abandoned for now. I'll update it to compare/contrast with the current proposal. I think we've already discussed the drawbacks in terms of duplicate configuration in the Risks and Mitigations section.

enhancements/network/baremetal-ipi-network-configuration.md
hardys: Perhaps @cgwalters may have some comments here - we did discuss this recently and he mentioned that the keyfile interface supported for UPI has proven stable, but I suspect if/when any major-rhel-version rebase happens for RHCOS we may be faced with some larger incompatibilities.

I'm not clear what our strategy would be if/when such a major-version bump happens, would we anticipate a path for in-place upgrades or is it more likely to be a reinstall?

cathay4t: bias: I am developer of nmstate.

 * The keyfile is not standardized as yaml/json. The only specification like document I found is http://www.gtkbook.com/gtkbook/keyfile.html . Using keyfile directly will need your project sharing mutual understanding of keyfile spec as NetworkManager.
 * Besides creating NM connection files, nmstate has done a lot workarounds on complex network layouts like activation order, link depedency, preserving link in up state with best efforts, rollback on verification failure. Eventually, you will find yourself facing the old problems nmstate had.
 * IMHO. Keyfile of NetworkManager has never been treated as API.

cathay4t: @thom311 As NetworkManager developer, any suggestions?

EdDev: I want to address this from a different angle.
nmstate was originally created to provide a declarative API to a node network configuration and state.

We found that the work to translate specific applications network configurations is duplicated between teams & groups, each investing a lot of manpower and maintenance effort. As we are in favor of sharing logic, nmstate came to be and allowed applications to describe the desired end results, letting it do its best to accomplish it and validate at the end success.
With the help of NetworkManager in the backend, it provides most of the needs an application is looking for without the need to invest in scripts and low level logic.
And as I see it, it has another strong advantage over anything else: It is an API used by high level application logic across openship. Using it will just be a cross (OCP) applications API and will allow issues found in one place to be shared with the others.

I would consider this higher level API over anything lower, especially if we stay in the openshift domain, where nmstate is already very active.
If there are problems which nmstate is bringing, I would like to see them raised so we could provide solutions there.

thom311: > IMHO. Keyfile of NetworkManager has never been treated as API.

keyfiles (like also ifcfg-rh files) are totally considered an stable API.

Otherwise, when updating NetworkManager to a newer version, you'd have to migrate the configuration or your system's behavior would change. That would be very undesirable.

We also explain them in manual pages `man nm-settings-keyfile` and `man nm-settings` (where there arguably is room for improvement). The lower layers indeed use glib's [`GKeyFile` API](https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html), which is documented and basically parses the files to/from a dict-of-dict of strings. libnm also provides an API for converting `GKeyFile` to/from `NMConnection` (which is an API that should simply handling of what are plain string dictionaries).

It's fully supported to edit/pre-deploy a profile in keyfile format.

> I'm not clear what our strategy would be if/when such a major-version bump happens

NetworkManager's major releases don't break API (as such, they are not really "major" according to semantic versioning).
We call 1.32.0 a major release and like all our major releases are supposed to be fully backward compatible (including keyfile format). Well, technically, the [NEWS file](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/d9c0d43879e8420dda6482b05341dcfeedf7be43/NEWS#L11) for 1.32.0 lists some changes in behaviors, but these changes are really supposed not to affect users in a bad way. So maybe I should say: we really try hard not to break behavior in a way that affects real users in a bad way.

RHEL9 and RHEL8 use pretty much the same NetworkManager anyway... like every Linux distro.

---

I am pretty sure that also nmstate considers its API stable, so that isn't really a unique feature to either of them.

cybertron: Thanks for all the discussion. For the purposes of this enhancement I don't think nmstate is an option, but I hope we can revisit this topic in a future release.

cybertron:
Okay, I've pushed an updated version of this based on the comments so far. The two remaining points I'm aware of that need to be addressed are:
* How do we encode the user-provided config into a secret? Base64 encode in the format discussed in the user API section? Maybe as JSON instead of YAML?
* Any more details about the master deployment process? I gave a (very) high level overview, but someone with more experience in that process can probably do better.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Processing user configuration
+
+#### Deployment of the controlplane hosts via terraform
+
+We will map the keyfiles to their appropriate Machine using the host field
+of the baremetal install-config. The keyfiles will then be added to custom
+images for each Machine built by Terraform and Ironic.
+
+TODO - Does anyone have more detail to add here?
```
hardys, line 205: @kirankt and/or @dtantsur perhaps you can help with more details here?

I think it will be something like:
* During startup of the provisioning services on the bootstrap VM, we will build a customized image (for each master?) containing the network configuration specified in the install-config
* This image will be passed to Terraform/Ironic to ensure each host is deployed with the appropriate configuration
* The configuration will be persisted to the cluster via a secret, so that day-2 rebuild of the controlplane hosts is possible

Re the secret, I'm not entirely clear if we'll expect to create a secret per-host, or a single secret with the list of hosts?  That probably depends on the image-building implementation so I guess either could work.

enhancements/network/baremetal-ipi-network-configuration.md
hardys: Ok I think the consensus is we'll go with keyfiles as a first step, then when in future the integration of nmstate ref #747 is progressed, we could revisit adding an nmstate based day-1 API.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,241 @@
...
+## Alternatives
+
+### Use Kubernetes-NMState NodeNetworkConfigurationPolicy custom resources
+
+If we were able to install the [NNCP CRD](https://nmstate.io/kubernetes-nmstate/user-guide/102-configuration)
+at day-1 then we could use that as the configuration interface. This has the advantage of matching the configuration syntax and objects used for day-2 network configuration via the operator.
+
+However, we have been unable to get agreement on a method to deliver the operator on day-1, which makes this problematic.
+It's possible we could implement some way to pull just the NNCP CRD from the operator payload and isntall it on day-1, but at this point we likely don't have time to do that and it's not clear we even want to if the operator can't be installed on day-1 too.
+This would essentially orphan the CRs after the initial deployment completes.
```
dhellmann, line 225: What we're doing in the PAO is having its render mode create the CR needed to have OLM install the operator when the cluster comes up. Does that not work in this case?

cybertron: Because a lot of baremetal deployments are done disconnected, we can't rely on the operator being available to OLM. In many environments the deployer will need to take additional steps to mirror the content into their environment before it can be deployed. In addition, because of the sheer amount of content that needs to be mirrored for OLM there may be deployers who don't want to do that at all. I know there are discussions happening that may improve the mirroring situation, but nothing that will happen soon enough to help us here.

Also, part of the lack of agreement on NMState for day-1 was whether we should use it at all. I hope we can eventually get past that since it's already in use for day-2, but we just ran out of time to have that discussion if we wanted to deliver this in 4.9.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,241 @@
...
+
+* What is the best way to encode the data above into secret(s)?
+
+### Processing user configuration
+
+#### Deployment of the controlplane hosts via terraform
+
+We will map the keyfiles to their appropriate Machine using the host field
+of the baremetal install-config. The keyfiles will then be added to custom
+images for each Machine built by Terraform and Ironic.
```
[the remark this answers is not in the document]
cybertron: Sure. This is part of why I left the TODO because I'm not super-familiar with the process here. :-)

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: > Re the secret, I'm not entirely clear if we'll expect to create a secret per-host, or a single secret with the list of hosts? That probably depends on the image-building implementation so I guess either could work.

Yeah, that's why I left that as an open question. I don't have enough familiarity to say which works better. I'm hoping we can get input from the people closer to the implementation.

cybertron:
Added the missing sections and tried to address the comments so far.

enhancements/network/baremetal-ipi-network-configuration.md
dhellmann: OK, it would be good to capture that level of detail here in this document.

I'm not sure I buy in to the "mirroring is hard" argument. We're certainly expecting them to mirror the PAO to get those features if they want it. There's also work going on to make [the process for pruning the catalog](https://docs.openshift.com/container-platform/4.6/operators/admin/olm-restricted-networks.html#olm-pruning-index-image_olm-restricted-networks) easier.

hardys: I think the main point is that OpenShift today lacks any API for day-1 host network configuration (other than the keyfiles passthrough provided by `coreos-installer`

IMO that is a core function of the platform, and it's not ideal to ask users to enable optional add-on components to achieve it, but since there isn't consensus on that (ref #747) we've decided to go with the interface that does exist in the meantime e.g keyfiles - that doesn't preclude adding some higher-level API in future, but it also means we don't have to block on it.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,279 @@
...
+    etc...
+```
+
+In install-config this would look like:
+```yaml
+platform:
+  baremetal:
+    hosts:
+      - name: openshift-master-0
+        networkConfig:
```
kirankt, line 180: I've been thinking about the implementation. Whilst this looks logically neat, it would be hard to type into golang. My suggestion would be to use the following format for the install-config. Let me know what you think:

```
platform:
  baremetal:
    hosts:
      - name: openshift-master-0
        networkConfig:
          - interface: eth0
            data: |
              [connection]
              id=eth0
               ...
          - interface: eth1
            data: |
              [connection]
              id=eth1
               ...


```

enhancements/network/baremetal-ipi-network-configuration.md
kirankt: In the current implementation, we only deal with one image for all control plane hosts. We could make this separate, which would mean three different copies of, essentially the same, OS image. I like the idea of separation TBH, because it would open up the possibility of deploying non-homogeneous control plane hardware in the future. If we go this route, then the installer will create 3 secrets per control plane host.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,279 @@
...
+
+#### API Open Questions
+
+* What is the best way to encode the data above into secret(s)?
+
+### Processing user configuration
+
+#### Deployment of the controlplane hosts via terraform
+
+We will map the keyfiles to their appropriate Machine using the host field
```
kirankt, line 201: I think this should say host instead of machine.

enhancements/network/baremetal-ipi-network-configuration.md
dankenigsberg: @cybertron would you spell out why nmstate is "not an option"? What should its authors do for it to become one? As I see it "raw nmstate" has one drawback: it requires to define an nmstate container to be pulled in by the installer and during scale-out to convert nmstate yaml to nm key files.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Based on some separate discussion of this feature, it sounds like the configs will be associated with a BareMetalHost object. Is that correct? If so I'll use that to be precise about what we mean.

cybertron: Okay, never mind. I see that's discussed in literally the next section of the document (a part I didn't write, in my defense).

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Okay, that works for me. One secret per host is conceptually easier too. If we decide we want hosts to be able to share secrets at some point that can probably be added.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: @dankenigsberg Mostly because of the gridlock in the discussion on #747 that Steve linked. We've had trouble convincing OpenShift to accept NMState as part of the day 1 content, and that makes pulling a container as part of the installation problematic.

Originally we wanted to use NMState, but because of the timeframe for delivery of the feature we didn't have time to continue the prerequisite discussions and decided to pursue the lower level option, which hopefully won't preclude adding an NMState layer on top in the future.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: In the interest of not reiterating the entire discussion in #747 here, can we just link to that PR, explain that we did not have time to resolve it, and note that we want to revisit it in the future once we have the base keyfile functionality implemented?

enhancements/network/baremetal-ipi-network-configuration.md
dtantsur: From the ironic's perspective we operate on the level of ignition configuration. What you put there doesn't concern us.

enhancements/network/baremetal-ipi-network-configuration.md
dtantsur: Yeah, a secret per host is easier to reason about.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,279 @@
...
+images for each host built by Terraform and Ironic.
+
+TODO - Does anyone have more detail to add here?
+
+#### Deployment of compute/worker hosts
+
+BareMetalHost resources for workers will be created with the Secret containing the network data referenced in the `preprovisioningNetworkData` field defined in the Metal³ [image builder integration design](https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/image-builder-integration.md#custom-agent-image-controller).
+This will cause the baremetal-operator to create a PreprovisioningImage CRD and wait for it to become available before booting the IPA image.
+
+An OpenShift-specific PreprovisioningImage controller will use the provided network data to build a CoreOS IPA image with the correct network data in place, and the baremetal-operator will then use this to boot the Host.
```
dtantsur, line 212: nit: s/network data/ignition configuration/ (for clarity)

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,279 @@
...
+
+If not overridden, the contents of the same network data secret will be passed to the Ironic custom deploy step for CoreOS that installs the node.
```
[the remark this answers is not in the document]
zaneb: Yes, it will get passed to the deploy as NetworkData (not as UserData), so CoreOS installer will certainly ignore it (although code installed by the actual userdata could theoretically make use of it after the fact).

enhancements/network/baremetal-ipi-network-configuration.md
trozet: @hardys I would argue that we need network configuration on all platforms. OVN requires OVS network configuration on every single platform, which today is done by the MCO ovs-configuration bash script (creating network manager config) and would be much better served by nmstate or something more robust.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Okay, I've done what I said in my previous comment. Let me know what you think.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Ack, done.

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: I think the next paragraph covers this now.

enhancements/network/baremetal-ipi-network-configuration.md
hardys: @trozet I completely agree, but since we've not managed to achieve consensus on how to do that (ref #747) the scope of this PR is limited to using the interfaces which are currently supported (e.g keyfiles).

enhancements/network/baremetal-ipi-network-configuration.md
dankenigsberg: @cybertron I believe that the "raw nmstate" alternative is unrelated to #747 and its gridlock. It basically says "instead of inventing a new yaml format for holding NM keyfiles, let us use nmstate schema. This requires the user to install nmstatectl on the RHEL where they run openshift-install and requires us to run nmstatectl on scale out". These are downsides, and you can opt against this alternative, but it *is* an option. Frankly, I don't think that it would take a larger timeframe to implement.

phoracek: @cybertron if one of the issues was nmstate using Python, there is now a PoC of nmstate rewritten in Rust. Would that help sway the decision?

zaneb: It has come to my attention that this proposal/discussion is predicated on the assumption that the image builder's Secret must contain keyfiles.

An alternative would be if the Secret contained the nmstate data and the image builder container runs nmstatectl to convert it to keyfiles, before building them into the ignition. Provided that there is a supportable RPM available for nmstate then I don't see any obstacle to this. It's sad we won't be able to use the Python API from golang and will have to rely on calling out to it as a separate process but I assume that shouldn't be a problem.

The upstream metal³ API for the image builder doesn't require any particular format for the network data, but it does require that it be delivered in a Secret. There is no scenario in which the image builder ends up being able to read the NNCP directly. So if we want to avoid drift we will inevitably need a controller to copy data from the NNCP to the Secret. This can be trivial if they're in the same format. (It's not like k8s-nmstate provides the keyfiles in the NNCP Status, so if the image builder were built to expect keyfiles then it would just mean that this controller would have to include the nmstatectl conversion instead of doing it in the image builder. That's fine but it wouldn't actually buy us anything.)

Using nmstate from the outset would avoid having to change the format of the data we expect the user to provide to the installer in future once we have a method of installing k8s-nmstate at day 1; we could simply provide the same nmstate data to the Secret and the NodeNetworkConfigurationPolicy, or add the trivial controller somewhere (perhaps in CAPBM?) to sync them.

The only downside I can think of is that if we eventually decide against (ever) installing the NNCP CRD and instead implement some sort of Machine-specific configuration for the MCO, the installer won't have keyfiles available to pass to it.

cybertron: @phoracek I don't think Python was ever the objection, but I'll keep that in mind as we continue the discussions.

@dankenigsberg I'm not sure whether adding a dependency to the installer is necessarily trivial, particularly since we've gotten feedback that keyfiles are the supported network configuration interface in coreos and therefore we should use them instead. That's an argument we just didn't have time to chase if we were going to deliver this as originally planned. Now that it has slipped out of the 4.9 release anyway we may be able to revisit some of these topics.

Zane's proposal might have the least friction as it involves installing nmstate into a container image we control so we don't need anyone's permission to do that. The question of whether to use nmstate as the input format at all may still come up, but we can cross that bridge when we come to it. It does also have the drawback of making the solution further baremetal-specific, but given the difficulty we've had in reaching broad consensus maybe limiting the scope is a feature not a bug. :-)

dankenigsberg: @zaneb ++

zaneb: > @phoracek I don't think Python was ever the objection, but I'll keep that in mind as we continue the discussions.
> 
> @dankenigsberg I'm not sure whether adding a dependency to the installer is necessarily trivial

Just to add some context on the constraints here, the installer is written in golang and it is delivered as a single binary with **_no_** dependencies. So anything that requires a Python, or even Rust, program at install time is a non-starter.

dankenigsberg: > the installer is written in golang and it is delivered as a single binary with no dependencies.
Thanks @zaneb, I did not know. This constraint defies the ancient Unix principle of multiple applications interacting with each other over well-defined interfaces. It is sad that we need to re-learn it.

zaneb: @dankenigsberg well, note that the installer isn't delivered as an RPM, nor is it restricted to platforms we control, nor even to Unix-like platforms. If potential customers wanting to try out OpenShift in their cloud account are spending their time wrestling with getting the right version of Python on their Windows laptop then we are losing a good proportion of them as customers.
That said, I didn't make the rules :)

enhancements/network/baremetal-ipi-network-configuration.md
kirankt: +1 for filenames to replace the interface name in the API.

enhancements/network/baremetal-ipi-network-configuration.md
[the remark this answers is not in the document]
dtantsur: So, what's the action item on this line? Should we somehow start handling `networkData`? Should this line be reworded to say "`userData` containing an ignition configuration writing the given keyfiles"?

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,281 @@
...
+
+N/A
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+There should be little impact on upgrades and downgrades. Nodes are deployed with network configuration baked into the image, which means it will remain over upgrades or downgrades. NetworkManager keyfiles are considered a stable interface so any version of NetworkManager should be able to parse them equally.
```
dtantsur, line 239: Should we talk about the whole CoreOS IPA business here? It does have upgrade impact.

enhancements/network/baremetal-ipi-network-configuration.md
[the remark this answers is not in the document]
zaneb: I don't think there's anything to do, this is just documenting the (future) existence of that behaviour.

enhancements/network/baremetal-ipi-network-configuration.md
zaneb: I think that needs another enhancement proposal.

enhancements/network/baremetal-ipi-network-configuration.md
kirankt: We can probably simplify the config a bit and also use the suggestion to include file names like so:

```yaml
platform:
  baremetal:
    hosts:
      - name: openshift-master-0
        networkConfig:
        - eth0.nmconnection: |
            [connection]
            id=eth0
            uuid=18e0cec7-041c-4fb3-957c-a60c80dd9b85
            type=ethernet
            etc...
```

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,281 @@
...
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Currently in the IPI flow, there is no way to provide day-1 network configuration
+which is a common requirement, particularly for baremetal users.  We can build
+on the [UPI static networking enhancements](rhcos/static-networking-enhancements.md)
```
kirankt, line 42: NIT. This relative URL doesn't work correctly. Suggestion: We can use: https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/static-networking-enhancements.md

enhancements/network/baremetal-ipi-network-configuration.md
kirankt: I also wanted to clarify which format we're all leaning towards: NM Keyfiles or nmstate config. Either way, I hope I hope we're just consuming a blob of network config data. I was just looking at the nmstate schema and its pretty extensive. I don't think we should include it in the install-config. At least in this proposal. 
https://github.com/nmstate/nmstate/blob/base/libnmstate/schema.py

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,281 @@
...
+        - interface: eth0
+          data: |
+            [connection]
+            id=eth0
+            uuid=18e0cec7-041c-4fb3-957c-a60c80dd9b85
+            type=ethernet
+            etc...
+```
+
+Because the initial implementation will be baremetal-specific, we can put the
```
sferich888, line 190: baremetal-specific or host specific?

zaneb: specific to the baremetal platform

zaneb:
I believe that this version resolves the main concern that folks had (around the input format, NetworkManager keyfiles vs. nmstate). Is there anything else preventing this moving forward?
/assign @trozet

kirankt:
Can we come to a consensus and approve this enhancement, please? I believe we have enough to move forward with the implementation. Thanks.

sadasu:
@trozet and @staebler could you please provide your comments? This enhancement proposes how we would be passing day-1 networking config via the installer so would like to get an OK from both of you. Thanks!

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,281 @@
...
+
+* A new section in install-config.
+* A secret that contains base64-encoded content for the keyfiles.
+
+These are not mutually exclusive. If we implement the install-config option, we will still need to persist the configuration in a secret so it can be used for day-2. In fact, the initial implementation should likely be based on secrets, and then install-config integration can be added on top of that to improve the user experience. This way the install-config work won't block implementation of the feature.
+
+The data provided by the user will need to have the following structure:
+```yaml
+[hostname]:
+  [interface name]:
```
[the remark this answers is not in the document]
hardys: IMO it's not a chicken/egg problem, but the user will need knowledge of the hardware to provide the correct interface name.

It's outside the scope of this proposal, but in future we could consider adding validations (compare user config with nic names detected during inspection), and/or nic aliases which are less directly coupled to the specfic hardware.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,275 @@
...
+# Baremetal IPI Network Configuration
+
+Describe user-facing API for day-1 network customizations in the IPI workflow,
+with particular focus on baremetal where such configuration is a common
+requirement.
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
```
hardys, line 32: We could probably mark the status as implementable and check both these I think, given the pretty thorough discussion so far?

hardys:
/lgtm

Not approving since I contributed to an earlier draft of this, but overall this lgtm and I think outlines an actionable plan which won't impact any platforms outside IPI baremetal (long term OpenShift should gain a platform-agnostic API for network config IMO, but that's outside the scope of this iteration)

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,275 @@
...
+openshift-master-1:
+  interfaces:
+    - name: eth0
+      etc...
+```
+
+In install-config this would look like:
+```yaml
+platform:
+  baremetal:
```
trozet, line 171: should we put it under baremetal if there might be plans in the future to use it on other platforms? I'm thinking here about replacing ovs-configuration with this.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,275 @@
...
+
+As this feature targets day-1 configuration there should be no version skew. Day-2 operation will be handled by other components which are outside the scope of this document.
+
+## Implementation History
+
+4.9: Initial implementation
+
+## Drawbacks
+
+Using NetworkManager keyfiles as the user interface for network configuration makes day-2 configuration more problematic. The only existing mechanism for modifying keyfiles after initial deployment is machine-configs, which will trigger a reboot of all affected nodes when changes are made. It also provides no validation or rollback functionality.
```
trozet, line 245: If the keyfiles are laid down as machine-configs, and we modify them in ovs-configuration, wont that be an issue?

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,275 @@
...
+      etc...
+```
+
+In install-config this would look like:
+```yaml
+platform:
+  baremetal:
+    hosts:
+      - name: openshift-master-0
+        networkConfig:
```
trozet, line 174: this will support all types of nmstate configuration right?

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,275 @@
...
+<hostname>: <nmstate configuration>
+<hostname 2>: <nmstate configuration>
+etc...
+```
+
+For example:
+```yaml
+openshift-master-0:
+  interfaces:
+    - name: eth0
```
trozet, line 159: so this works if you know the NIC name. What about doing something like TripleO did with introspection... identifying NICs dynamically and mapping them to networks. I get this might not be applicable to the scope of this enhancement, just thinking longer term how that would work with this API, or if that would need a new API?

An example of this is I might have a scenario where I want my kapi traffic to go over the ctlplane network, but I want to define another network for all my default egress OVN traffic.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,275 @@
...
+RHCOS already provides a mechanism to specify NetworkManager keyfiles during deployment of a new node. We need to expose that functionality during the IPI install process, but preferably using [nmstate](https://nmstate.io) files as the interface for a more user-friendly experience. There are a couple of options on how to do that:
+
+* A new section in install-config.
+* A secret that contains base64-encoded content for the keyfiles.
+
+These are not mutually exclusive. If we implement the install-config option, we will still need to persist the configuration in a secret so it can be used for day-2. In fact, the initial implementation should likely be based on secrets, and then install-config integration can be added on top of that to improve the user experience. This way the install-config work won't block implementation of the feature.
+
+The data provided by the user will need to have the following structure:
+```yaml
+<hostname>: <nmstate configuration>
```
trozet, line 150: I know you mentioned at first it will be per host, but I think that will be really annoying for some users. Assume in my scenario all my hosts are homogeneous and I have 120 of them. That's a giant config. Could we just support applying the same config to all nodes at least initially as well?

trozet (commented):
overall looks good. Some comments inline. @fedepaol can you PTAL?

enhancements/network/baremetal-ipi-network-configuration.md
cybertron: Yeah, that's a good point. Although I suppose every time MCO rewrites them the script would run again to redo the changes? I know we have deployers doing this today and it works, but I don't know for sure that they're using OVNK.

enhancements/network/baremetal-ipi-network-configuration.md
zaneb: For the implementation we won't care either way. It's just a question of whether we can give it an interface that makes sense.

enhancements/network/baremetal-ipi-network-configuration.md
fedepaol: Just to mention, I know that the nmstate folks (@qinqon in particular) are working on a dynamic way to generate nmstate configurations depending on the current network layout (i.e. defining the bridge on top of the default gateway, instead of sticking the interface name).

Not sure of the state, because we discussed it some time ago, but I think that would solve this issue.

enhancements/network/baremetal-ipi-network-configuration.md
fedepaol: +1 for default config / templated configs and some sort of node selector mechanism

enhancements/network/baremetal-ipi-network-configuration.md
qinqon: Hi, yes we are working on a tool named `nmpolicy` that will solve something at least similar to that we have a design document with examples, here it's the example you are interested on I think.

```yaml
capture:
  default-gw: routes.running.destination=="0.0.0.0/0"  
  base-iface-routes: routes.running.next-hop-interface==capture.default-gw.routes.running[0].next-hop-interface
  base-iface:  interfaces.name==capture.default-gw.routes.running[0].next-hop-interface
  bridge-routes: capture.base-iface-routes | routes.running.next-hop-interface:="br1"
  delete-base-iface-routes: capture.base-iface-route | routes.running.state:="absent"
  bridge-routes-takeover: capture.delete-base-iface-routes.routes.running + capture.bridge-routes.routes.running
desiredState:
  interfaces:
    - name: br1
      description: Linux bridge with base interface as a port
      type: linux-bridge
      state: up
      ipv4: {{ capture.base-iface.interfaces[0].ipv4 }}
      bridge:
        options:
          stp:
            enabled: false
        port:
        - name: {{ capture.base-iface.interfaces[0].name }}
  routes:
    config: {{ capture.bridge-routes-takeover.running }}
 }}
```

fedepaol:
> overall looks good. Some comments inline. @fedepaol can you PTAL?

left couple of comments, looks good!

enhancements/network/baremetal-ipi-network-configuration.md
zaneb: I don't think that will help; nmstatectl cannot run on the host, it must run on the cluster and produce keyfiles that are incorporated into the ISO before it even boots on the host. We don't currently have a way of building e.g. a container image into the ISO, so everything but the Ignition (which contains the keyfiles) must come over the network and therefore cannot be required to set up the network.

qinqon: The nmstate team is reimplementing nmstatectl so it's just a binary without python dependencies, will that help with that ? They are also going to implement a flag to configure directly on the kernel with netlink bypassing NetworkManager.

romfreiman:
Did you take into account existing implementation of the assisted installer and ztp? Are those collide? Or can work together?

hardys:
> @hardys I mostly asking on the nmstate. What will happen if networkConfig will be set in the install config as proposed here, but ZTP flow will be used? Can it work together? Or the networkConfig will be ignored?

@romfreiman until/unless we converge the flows the networkConfig (and new install-config interface) won't be used by the ZTP flow - the ZTP flow builds an ISO with the network configuration embedded already - that process should continue to work regardless of this new flow, but hopefully we can look at ways to converge in future.

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,275 @@
...
+
+#### Create a net-new NMState Wrapper CR
+
+The [assisted-service](https://github.com/openshift/assisted-service/blob/0b0e3677ae83799151d11f1267cbfa39bb0c6f2e/docs/hive-integration/crds/nmstate.yaml) has created a new NMState wrapper CR.
+
+We probably want to avoid a proliferation of different CR wrappers for nmstate
+data, but one option would be to convert that (or something similar) into a common
+OpenShift API, could such an API be a superset of NNCP e.g also used for day-2?
+
+This would mean we could at least use a common configuration format based on nmstate with minimal changes (or none if we make it _the_ way OpenShift users interact with nmstate), but unless the new API replaces NNCP there is still the risk of configuration drift between the day-1 and day-2 APIs. And we still need a Secret to generate the PreprovisioningImage.
```
flaper87, line 267: > but unless the new API replaces NNCP there is still the risk of configuration drift between the day-1 and day-2

Agreed, maybe this is a good opportunity to bump this conversation? Try to find some alignment here?

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,275 @@
...
+In the sections below we will describe the user-facing API that contains network configuration, and the proposed integration for each of these cases.
+
+### User-facing API
+
+RHCOS already provides a mechanism to specify NetworkManager keyfiles during deployment of a new node. We need to expose that functionality during the IPI install process, but preferably using [nmstate](https://nmstate.io) files as the interface for a more user-friendly experience. There are a couple of options on how to do that:
+
+* A new section in install-config.
+* A secret that contains base64-encoded content for the keyfiles.
+
+These are not mutually exclusive. If we implement the install-config option, we will still need to persist the configuration in a secret so it can be used for day-2. In fact, the initial implementation should likely be based on secrets, and then install-config integration can be added on top of that to improve the user experience. This way the install-config work won't block implementation of the feature.
```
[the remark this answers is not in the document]
hardys: Providing the data via manifests is probably a viable alternative, but can you expand on the concerns just so we all have relevant context if/when any subsequent proposal gets created?

/cc @cybertron @zaneb @sadasu what are your thoughts here, is it reasonable to fall back to manifests containing the underlying secret data given this latest feedback?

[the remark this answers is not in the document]
hardys: One problem with only exposing the Secret API to users, is they will also need to modify all of the installer-generated BareMetalHost manifests (so they reference the desired secret(s)).

That's likely to be error prone (particularly for environments with a lot of worker hosts and per-host configuration e.g static IPs), obviously it's scriptable but this is why we wanted to enable input of this data via the install-config.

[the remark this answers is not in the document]
zaneb: > I would prefer to keep the install-config integration as a separate enhancement.

For all intents and purposes, this _is_ a separate enhancement for the install-config integration.

> I have serious concerns over the appropriateness of embedding nmstate files into the install config. I would like to start with secrets in manifests that the user can modify.

It's just the same data with the added downside that you'd have to specify the Secret wrapper (TBH users should never need to know that it's stored in a Secret) and get the name right so that the installer can guess which BaremetalHost to link it to. I assume you're suggesting this as a temporary workaround until we can agree on the right interface for the installer?

enhancements/network/baremetal-ipi-network-configuration.md
```diff
@@ -0,0 +1,275 @@
...
+authors:
+- "@cybertron"
+- "@hardys"
+- "@zaneb"
+reviewers:
+- "@kirankt"
+- "@dtantsur"
+- "@zaneb"
+approvers:
+- "@trozet"
```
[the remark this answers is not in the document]
cybertron: Yep, will do.

enhancements/network/baremetal-ipi-network-configuration.md
cgwalters: Do we actually forsee the installer trying to parse/validate this data?  That seems hard.

Ultimately this data is just passed through into an Ignition config provided to the node, right?

I really see this as leading towards a general ability to inject ignition (rendered indirectly perhaps via [butane](https://github.com/coreos/butane/)) or other tools to specific hosts.

xref https://github.com/openshift/machine-config-operator/issues/1720

Regarding nmstate specifically; I think this actually has the same issue as e.g. offering `butane` configs.
It may be simplest to have `openshift-install` only accept final rendered ignition configs to start.  Then an administrator who wants to generate NM configs via nmstate (or use `butane` or another custom tool) needs to run that outside of `openshift-install`.

zaneb: > Do we actually forsee the installer trying to parse/validate this data? That seems hard.

No, I don't think the installer would validate anything beyond the fact that it's valid yaml.

> Ultimately this data is just passed through into an Ignition config provided to the node, right?

Yes.

> I really see this as leading towards a general ability to inject ignition (rendered indirectly perhaps via [butane](https://github.com/coreos/butane/)) or other tools to specific hosts.

Not sure I agree here. The network data is only one part of the ignition we give the host. The rest involves pulling the ironic agent container and running it as a service. There's no way we want users to even have to know that exists.

enhancements/network/baremetal-ipi-network-configuration.md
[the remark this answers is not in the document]
zaneb: Good point Steve.
I agree that might work (and perhaps this was @staebler's intention) as a temporary measure to enable us to test the rest of the flow while we agree on what the interface to the installer should be (although it will require a bunch of extra rework), but if editing the manifests is the only interface to it then we haven't really shipped anything usable to customers.

PR #820, opened 2021-06-30
CONSOLE-2904: Update Console dynamic plugins feature page

vojtechszocs:
- minor updates reflecting the current implementation details
- update `disable-plugins` query parameter description to match changes in openshift/console#9363

cc @spadgett

### DIFF
--- a/enhancements/console/dynamic-plugins.md
+++ b/enhancements/console/dynamic-plugins.md
@@ -79,6 +79,23 @@ contributing to a single monorepo.
 
 ## Proposal
 
+### User Stories
+
+#### Story 1
+
+As a user of OpenShift, I should be able to utilize the currently installed operator
+APIs and features in console frontend via plugins delivered through these operators.
+
+#### Story 2
+
+As a developer of OpenShift, I should be able to develop, build and deploy plugins
+on the cluster to expose APIs and features of the given operator in console frontend.
+
+#### Story 3
+
+As an admin of an OpenShift cluster, I should be able to list plugins available on
+the cluster and enable/disable plugins upon operator install or at any point later.
+
 ### Module Federation
 
 Console will use [Webpack module federation](https://webpack.js.org/concepts/module-federation/)
@@ -393,6 +410,18 @@ matures.
 
 Static plugins are already a supported feature. Any existing static plugin that
 is migrated to a dynamic plugin will need to have the same support level.
+
+#### Dev Preview -> Tech Preview
+
+None
+
+#### Tech Preview -> GA
+
+None
+
+#### Removing a deprecated feature
+
+None
 
 ### Upgrade / Downgrade Strategy
 
--- a/enhancements/console/dynamic-plugins.md
+++ b/enhancements/console/dynamic-plugins.md
@@ -208,7 +208,7 @@ extension points will be available.
   "version": "1.2.3",
   "displayName": "My Plugin",
   "dependencies": {
-    "@console/pluginAPI": "*"
+    "@console/pluginAPI": "4.8.x"
   },
   "extensions": [
     {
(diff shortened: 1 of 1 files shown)


enhancements/console/dynamic-plugins.md
```diff
@@ -87,10 +87,12 @@ to dynamically load code from another application while sharing dependencies.
 It also allows console to share its components with plugins built and bundled
 separately.
 
-Plugins will need to be built with Webpack 5 to use module federation.
+Plugins will need to be built with Webpack 5+ which includes native support for
+module federation.
 
-We have a pull request with a [working prototype](https://github.com/openshift/console/pull/6101)
```
vojtechszocs, line 92: I've removed this section since openshift/console#6101 is already merged and listed below in "Implementation History" section. If you think we should keep it, let me know and I'll restore it.

enhancements/console/dynamic-plugins.md
```diff
@@ -185,63 +187,122 @@ extension points will be available.
...
-| Property | Description |
-| ----- | -------- |
-| `name: string` | Used as a unique identifier. Each plugin must have a unique name. |
-| `version: string` | The version of the plugin. Version must be parsable by node-semver, eg. “1.2.3” |
-| `pluginAPI: string` | Semver range of compatible OpenShift console versions. |
-| `displayName?: string` | User friendly display name. |
-| `description?: string` | The description of the plugin. |
-| `extensions: Extension[]` | List of extensions contributed by the plugin. |
-| `dependencies?: { [pluginName: string]: string }` | A plugin which depends on other plugins must explicitly define their dependencies. The key of the map is the plugin name and the value is a semver version range. The plugin is not loaded if a dependency is not met. |
+<table>
```
vojtechszocs, line 222: Using HTML table since Markdown-specific table syntax does not support multi-line code blocks.

If there's a better way to do this, please let me know.

enhancements/console/dynamic-plugins.md
```diff
@@ -363,6 +424,8 @@ console will only load the correct plugin.
 ## Implementation History
 
 * 2020-07-24 - Initial [dynamic plugin PR](https://github.com/openshift/console/pull/6101) opened.
+* 2020-11-26 - Allow static plugins to use new extension types with code references
```
vojtechszocs, line 427: Listed openshift/console#7163 here since it marks an important milestone in the transition process from static to dynamic plugins.

enhancements/console/dynamic-plugins.md
```diff
@@ -185,63 +204,122 @@ extension points will be available.
...
 ```json
 {
-  "$schema": "/path/to/schema/manifest.json",
-  "id": "my-plugin",
+  "name": "my-plugin",
   "version": "1.2.3",
-  "pluginAPI": "4.7.x",
   "displayName": "My Plugin",
+  "dependencies": {
+    "@console/pluginAPI": "*"
```
vojtechszocs, line 211: Hm, this should probably be `4.7.x` instead of `*` since the latter is not very explanatory.

enhancements/console/dynamic-plugins.md
```diff
@@ -280,10 +358,10 @@ This prevents an uncaught error from causing the application to white-screen
...
 and show a message in the notification drawer to let users know.
 
-Additionally, the console can contain a `?disable-plugins` query parameter with
-a comma separated list of plugin names. When this parameter is present, the
-corresponding dynamic plugins are not loaded. If `all` is passed, all dynamic
-plugins will be disabled.
+Additionally, console users can disable specific or all dynamic plugins that
+would normally get loaded by console via `?disable-plugins` query parameter.
+The value of this parameter is either a comma separated list of plugin names
+(disable specific plugins) or an empty string (disable all plugins).
```
spadgett, line 364: When we get to the point where everything including the admin perspective is a dynamic plugin, it might not make sense to disable all plugins. But it's useful for now, and I think we can cross that bridge when we come to it.

spadgett (approved):
/lgtm

PR #826, opened 2021-07-07
Creating initial skeleton of eo v2 api proposal

ewolinetz:
In an effort to further simplify using EO and continue to improve upon the UX, we want to introduce a V2 for our CRD so that we can switch to a t-shirt size model instead and let the operator make informed decisions on the cluster structure.

cc @portante @jcantrill @alanconway

### DIFF
--- a/enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
+++ b/enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
@@ -1,0 +1,319 @@
+---
+title: elasticsearch-v2-api-and-infrastructure
+authors:
+  - "@ewolinetz"
+reviewers:
+  - "@jcantrill"
+  - "@bparees"
+  - "@alanconway"
+approvers:
+  - "@jcantrill"
+  - "@bparees"
+  - "@alanconway"
+creation-date: 2021-07-02
+last-updated: 2021-06-02
+status: implementable
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# elasticsearch-v2-api-and-infrastructure
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [x] Migration plan is defined
+- [x] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Currently when interacting with the Elasticsearch Operator via the Elasticsearch CRD, one needs to have a lot of knowledge on how to best configure
+the Elasticsearch cluster node's roles, memory/cpu, and replication count. We improve the user experience of the operator, and provide a more performant
+cluster that can operate at scale, by both simplifying the API and moving the knowledge of the desired infrastructure into the operator itself.
+
+## Motivation
+
+Currently the majority of our Elasticsearch bugs are performance related and are usually clusters that have been scaled vertically rather than horizontally.
+This can be summed up to the fact that it can be difficult to correctly decide the memory that each node should be allocated, and the number of nodes
+that should be in the cluster.
+
+Given each Elasticsearch node's uses its memory differently based on the role is is performing, having nodes that act as "All-In-One" causes resources
+to not be used in optimal ways and usually results an over allocation of resources.
+This over allocation can make it hard to justify horizontal scaling, which leads to more performance related issues.
+
+Reducing this decision to t-shirt sizes helps abstract this complicated understanding from users and simplifies the experience to picking the
+most appropriate t-shirt given their use case.
+
+For EO developers, this has the added benefit of allowing us to move away from the use of UUIDs which would simplify our code.
+
+### Goals
+The specific goals of this proposal are:
+
+* Outline the responsibility for the Elasticsearch Operator regarding desired infrastructure
+
+* Provide a v2 elasticsearch CRD API
+
+* Provide upgrade strategies from v1 to v2
+
+We will be successful when:
+
+* Users of Elasticsearch Operator no longer need to specify their node roles and count
+* Users can specify a t-shirt size instead that can better fit their needs
+* We can continue to improve upon the t-shirt size recommendations to make our customers successful
+
+### Non-Goals
+
+* N/A
+
+### Current workflow vs Proposed
+
+Currently, a consumer like Jaeger or Cluster Logging is responsible for the following for their ES cluster:
+
+1. Creating and updating their elasticsearch CR(s) so that EO can manage their cluster
+    * This includes breaking up nodes by roles
+    * Specifying (usually pass through) of node requirements
+
+With this proposal the workflow would be:
+
+1. Create their elasticsearch CR specifying a particular sized cluster (based on t-shirt sizes)
+
+Current specification:
+
+```yaml
+apiVersion: "logging.openshift.io/v1"
+kind: "Elasticsearch"
+metadata:
+  name: "elasticsearch"
+spec:
+  managementState: "Managed"
+  nodeSpec:
+    resources:
+      limits:
+        memory: 1Gi
+      requests:
+        cpu: 100m
+        memory: 1Gi
+  nodes:
+  - nodeCount: 1
+    roles:
+    - client
+    - data
+    - master
+    storage:
+      storageClassName: gp2
+      size: 10Gi
+  redundancyPolicy: ZeroRedundancy
+  indexManagement:
+    policies:
+    - name: infra-policy
+      pollInterval: 1m
+      phases:
+        hot:
+          actions:
+            rollover:
+              maxAge:   2m
+        delete:
+          minAge: 5m
+    mappings:
+    - name:  infra
+      policyRef: infra-policy
+      aliases:
+      - infra
+      - logs.infra
+```
+
+Proposed:
+
+```yaml
+apiVersion: "logging.openshift.io/v1"
+kind: "Elasticsearch"
+metadata:
+  name: "elasticsearch"
+spec:
+  storage:
+    storageClassName: gp2
+    size: 10Gi
+  size: 0x.medium
+  indexManagement:
+    infraPolicies:
+    - phases:
+      hot:
+        actions:
+          rollover:
+            maxAge:   2m
+      delete:
+        minAge: 5m
+    appPolicies:
+    - phases:
+      hot:
+        actions:
+          rollover:
+            maxAge: 2m
+      delete:
+        minAge: 5m
+```
+
+## Proposal
+
+Currently the Elasticsearch CR is overly complicated and allows for too much granular control over the data nodes, which places the expectation
+of expertise on the user rather than containing it to the operator (which is where it should reside). This proposal seeks to outline a v2 to provide
+a more simple means of interacting with the Elasticsearch Operator to ask for a managed Elasticsearch cluster that can be better
+configured for performance.
+
+### User Stories
+
+#### As an user of EO, I want to have to provide minimal information for my cluster to be created for me
+
+As an user I want to just be able to provide information such as a t-shirt size (which is outlined further for me) and my storage information
+and have an elasticsearch cluster spun up for me so that I do not need to be an expert and configure my node roles and node resources.
+
+### Implementation Details
+
+#### Assumptions
+
+* Currently the EO CRD version is V1 and this would be a V2. We would assume that EO is able to react for both a V1 and V2 and if the user creates a V2
+  elasticsearch CR with the same name as an existing V1, the operator would migrate the V1 backing storage naming to be used for V2 (read: it would
+  rename the PVCs and rebind the backing PV so that it matches the updated naming convention).
+
+### Risks and Mitigations
+
+## Not Breaking Other Operators who Depend on EO
+
+As part of adding another version of the CRD that EO will respond to, we need to ensure that we can continue to support those that remain on V1 so that
+they are able to migrate their usage to V2 in a timely manner that suits them (typically we give 1-2 releases since announcement before deprecating).
+
+## Design Details
+
+### PVC current and proposed
+
+Current storage PVCs + backing:
+```bash
+$ oc get pvc
+NAME                                        STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
+elasticsearch-elasticsearch-cd-w5afmzhp-1   Bound    pvc-b7c88b5e-0513-4912-8d13-0d4b81e20fe4   10Gi       RWO            gp2            13m
+elasticsearch-elasticsearch-cd-w5afmzhp-2   Bound    pvc-f3193fc1-da91-4267-b819-3b7b0af76cd6   10Gi       RWO            gp2            13m
+elasticsearch-elasticsearch-cd-w5afmzhp-3   Bound    pvc-2936caa5-9311-4548-b3ce-7a2ec5c439ea   10Gi       RWO            gp2            13m
+elasticsearch-elasticsearch-m-xpk5ybty      Bound    pvc-c90afaab-b4d4-47ff-8dd6-15f411ebe999   10Gi       RWO            gp2            13m
+```
+
+Proposed storage PVCs + backing:
+```bash
+$ oc get pvc
+NAME                      STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
+elasticsearch-0           Bound    pvc-b7c88b5e-0513-4912-8d13-0d4b81e20fe4   10Gi       RWO            gp2            13m
+elasticsearch-1           Bound    pvc-f3193fc1-da91-4267-b819-3b7b0af76cd6   10Gi       RWO            gp2            13m
+elasticsearch-2           Bound    pvc-2936caa5-9311-4548-b3ce-7a2ec5c439ea   10Gi       RWO            gp2            13m
+elasticsearch-master      Bound    pvc-c90afaab-b4d4-47ff-8dd6-15f411ebe999   10Gi       RWO            gp2            13m
+```
+
+### T-Shirt sizing
+
+To take inspiration from EC2 node sizing and the Loki Operator the Elasticsearch Operator would provide a more simple interface to customers
+in the form of t-shirt sizes that would encapsulate replication policies, range of node count, resources provided per node.
+
+This would need to be further fleshed out in a spike to determine what a Medium (the 90% use case) would be sized at, however some example sizes
+could be:
+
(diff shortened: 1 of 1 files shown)


alanconway (approved):

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,300 @@
...
+```yaml
+apiVersion: "logging.openshift.io/v1"
+kind: "Elasticsearch"
+metadata:
+  name: "elasticsearch"
+spec:
+  storage:
+    storageClassName: gp2
+    size: 10Gi
+  size: 0x.medium
```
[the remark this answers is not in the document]
ewolinetz: i talk about it a bit more under the `t-shirt sizes` header... it implies 0 redundancy

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,300 @@
...
+in the form of t-shirt sizes that would encapsulate replication policies, range of node count, resources provided per node.
+
+This would need to be further fleshed out in a spike to determine what a Medium (the 90% use case) would be sized at, however some example sizes
+could be:
+
+* 0x.Small - A small (e.g. Dev test cluster) with zero redundancy. This would likely only be able to support retention up to 72 hours.
+  * This could be comprised of just 1 master node and 1-3 data+ingest nodes with 1:1 separate coordinating nodes (matching data nodes)
+* 1x.Medium - A cluster that should be able to cover 90% of our customer use cases and provide single redundancy
+  * 3 masters, 3-5 data+ingest nodes and 1:1 coordinating nodes
+* 1x.Large - For very large customer clusters with single redundancy
```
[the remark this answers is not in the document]
ewolinetz: redundancy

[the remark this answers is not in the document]
ewolinetz: also these are just example names, these aren't set in stone

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,300 @@
...
+
+* Currently the EO CRD version is V1 and this would be a V2. We would assume that EO is able to react for both a V1 and V2 and if the user creates a V2
+  elasticsearch CR with the same name as an existing V1, the operator would migrate the V1 backing storage naming to be used for V2 (read: it would
+  rename the PVCs and rebind the backing PV so that it matches the updated naming convention).
+
+### Risks and Mitigations
+
+## Not Breaking Other Operators who Depend on EO
+
+As part of adding another version of the CRD that EO will respond to, we need to ensure that we can continue to support those that remain on V1 so that
```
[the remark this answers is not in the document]
ewolinetz: those usually come in the form of a transformation hook though right? I'll spell it out a bit more in the doc but my thought was this:
If there exists a V1 CR named "elasticsearch" in the namespace "openshift-logging"
and then a V2 CR named "elasticsearch" in the namespace "openshift-logging" is created
The operator would ignore the V1 one and migrate the objects from V1 to V2 (objects primarily being currently used storage)

My main concern with doing a straight migration is that we would need to try to extrapolate the desired t-shirt size based on the current configurations which may be very difficult to get correct. It may also be that the customer's current configuration does not correctly meet their demands so doing a migration could leave them with similar issues (mainly if their storage capacity isn't high enough or if their cluster is under resourced)

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,300 @@
...
+
+### T-Shirt sizing
+
+To take inspiration from EC2 node sizing and the Loki Operator the Elasticsearch Operator would provide a more simple interface to customers
+in the form of t-shirt sizes that would encapsulate replication policies, range of node count, resources provided per node.
+
+This would need to be further fleshed out in a spike to determine what a Medium (the 90% use case) would be sized at, however some example sizes
+could be:
+
+* 0x.Small - A small (e.g. Dev test cluster) with zero redundancy. This would likely only be able to support retention up to 72 hours.
```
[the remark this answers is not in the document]
ewolinetz: This was only meant to be an example. there would likely need to be a spike card created to determine what each t-shirt size should be. We had done something like this already with the Loki Operator.

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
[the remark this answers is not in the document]
alanconway: Yep, it's a trade off: the migration hook is a "standard" way of doing this and allows you to provide a v2-only operator, but it requires that an automatic v1->v2 translation will not break existing v1 deployments.

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,310 @@
...
+- [x] Test plan is defined
+- [x] Migration plan is defined
+- [x] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Currently when interacting with the Elasticsearch Operator via the Elasticsearch CRD, one needs to have a lot of knowledge on how to best configure
+the Elasticsearch cluster node's roles, memory/cpu, and replication count. In order to continue to improve the user experience of the operator
+and provide a performant cluster that can operate at scale we should simplify the API and move the knowledge of desired infrastructure into the
```
portante, line 36: Perhaps consider writing this last sentence as:

> We improve the user experience of the operator, and provide a more performant cluster that can operate at scale, by both simplifying the API and moving the knowledge of the desired infrastructure into the operator itself.

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,310 @@
...
+operator itself.
+
+## Motivation
+
+Currently the majority of our Elasticsearch bugs are performance related and are usually clusters that have been scaled vertically rather than horizontally.
+Given each node responsibility of Elasticsearch uses its memory differently, having nodes that act as "All-In-One" causes resources to not be used
+in optimal ways and usually requires an over allocation of resources (which makes it hard to justify horizontal scaling, which leads to performance
+related issues). It would also allow us to move away from the idea of node UUIDs and make naming conventions more generic and simple.
```
portante, line 44: Does this last sentence describe a motivation? Or does it describe an additional benefit provided by this work to improve "Day 2" operations?

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,310 @@
...
+Currently when interacting with the Elasticsearch Operator via the Elasticsearch CRD, one needs to have a lot of knowledge on how to best configure
+the Elasticsearch cluster node's roles, memory/cpu, and replication count. In order to continue to improve the user experience of the operator
+and provide a performant cluster that can operate at scale we should simplify the API and move the knowledge of desired infrastructure into the
+operator itself.
+
+## Motivation
+
+Currently the majority of our Elasticsearch bugs are performance related and are usually clusters that have been scaled vertically rather than horizontally.
+Given each node responsibility of Elasticsearch uses its memory differently, having nodes that act as "All-In-One" causes resources to not be used
+in optimal ways and usually requires an over allocation of resources (which makes it hard to justify horizontal scaling, which leads to performance
```
portante, line 43: It is not clear to me how this motivation maps to the changes proposed.

It "looks" like the ability specify individual roles is being removed in-favor of a small set of generic categories in which all the smarts for how to use Elasticsearch are placed.

> Currently the majority of our Elasticsearch bugs are performance related and are usually clusters that have been scaled vertically rather than horizontally.

How will this interface change address this?  Is it by providing a set of "t-shirt sizes" that map more closely to what the user wants so that they can scale correctly?  If so, then we might want to say that following this sentence.

What is nice about that is the t-shirt sizes in the operator can be adjusted over time as necessary.

What is difficult about that is when the t-shirt doesn't fit for the use case.

> Given each node responsibility of Elasticsearch uses its memory differently, having nodes that act as "All-In-One" causes resources to not be used in optimal ways and usually requires an over allocation of resources (which makes it hard to justify horizontal scaling, which leads to performance related issues).

There is a lot in this sentence.  It seems that what we are saying here is that is too difficult to express what the user needs for their environment with the existing API, and/or the user can easily make mistakes which cause performance problems.  So the motivation is that the new API helps prevent the user from shooting themselves in the foot by reducing the problem to picking the appropriate t-shirt size.

This is a laudable goal.  So then it is all about providing the proper t-shirt sizes.

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,310 @@
...
+
+* Outline the responsibility for the Elasticsearch Operator regarding desired infrastructure
+
+* Determine a v2 elasticsearch CRD API
+
+* Discuss upgrade strategies from v1 to v2
+
+We will be successful when:
+
+* Users of Elasticsearch Operator no longer need to specify their node roles and count
```
portante, line 57: Isn't this describing success in terms of an implementation detail instead of user success?  Perhaps something like:

>  * Users are able to properly pick a t-shirt size that fits their need
>  * This new API gives us added flexibility to modify the available t-shirt sizes as necessary

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,310 @@
...
+Given each node responsibility of Elasticsearch uses its memory differently, having nodes that act as "All-In-One" causes resources to not be used
+in optimal ways and usually requires an over allocation of resources (which makes it hard to justify horizontal scaling, which leads to performance
+related issues). It would also allow us to move away from the idea of node UUIDs and make naming conventions more generic and simple.
+
+### Goals
+The specific goals of this proposal are:
+
+* Outline the responsibility for the Elasticsearch Operator regarding desired infrastructure
+
+* Determine a v2 elasticsearch CRD API
```
portante, line 51: s/Determine/Provide/

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,310 @@
...
+
+* Discuss upgrade strategies from v1 to v2
```
portante, line 53: Is it really a goal to "discuss" upgrade strategies?  Or is it a goal to "Provide upgrade strategies from v1 to v2".

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,310 @@
...
+            maxAge: 2m
+      delete:
+        minAge: 5m
+```
+
+## Proposal
+
+Currently the Elasticsearch CR is overly complicated and allows for too much granular control over the data nodes, which places the expectation
+of expertise on the user rather than containing it to the operator (which is where it should reside). This proposal seeks to outline a v2 to provide
+a more simple means of interracting with the Elasticsearch Operator to ask for a managed Elasticsearch cluster that can be better
```
portante, line 155: `s/interracting/interacting`

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,310 @@
...
+configured for performance.
```
portante, line 156: > Currently the Elasticsearch CR is overly complicated and allows for too much granular control over the data nodes, which places the expectation of expertise on the user rather than containing it to the operator (which is where it should reside).

This feels like a motivation statement.  Not sure this is really proposing anything.

Perhaps a proposal statement could be: "Provide a v2 API which simplifies the configuration YAML to provide storage size, data retention times, and a "t-shirt" size summary configuration.

> This proposal seeks to outline a v2 to provide a more simple means of interacting with the Elasticsearch Operator to ask for a managed Elasticsearch cluster that can be better configured for performance.

It is not clear to me how t-shirt sizes can work when you allow the user to specify storage size and data retention times.  Seems like that still allows them to shoot themselves in the foot.

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,310 @@
...
+
+### User Stories
+
+#### As an user of EO, I want to have to provide minimal information for my cluster to be created for me
+
+As an user I want to just be able to provide information such as a t-shirt size (which is outlined further for me) and my storage information
+and have an elasticsearch cluster spun up for me so that I do not need to be an expert and configure my node roles and node resources.
```
portante, line 163: This is a great user story ... up until we give them t-shirt size, storage size, and retention.

I think we might want to consider ask the user for:

- max clients to support
- max # of separate indices
- max data retention times

And then use that to compute a t-shirt size that matches.  We can refine the exact inputs, but there is a formula of sorts from all the recommendations that are out there.

ewolinetz: the idea was actually that we would provide a "calculator" that would sit out there for 5 minutes to determine how much logs are generated and based on that extrapolate how many logs are created a day.

I think the most appropriate of those would be "max data retention times", however we still need to have an understanding of what size of each node will the customer allow ES to use -- hence the t-shirt size.

The shirt size doesn't need to strictly be cpu/memory but it is a templated way to let users tell us what size nodes to use and then let us scale up from there to keep the cluster in a sane state.

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
ewolinetz: its simplifying the code base -- more of an additional benefit for supporting use cases

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
ewolinetz: this section isn't user success -- its a success for the team/product

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
ewolinetz: If we know our data retention goals the operator can ensure it is scaled up enough based on the resources it is allowed to allocate per node. that, or it tells the users that there is no way it can meet that goal... which would be in the form of an alert

ewolinetz:
@alanconway @periklis @jcantrill any way we can get more eyes on this and merged in/have further suggestions made?

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,319 @@
...
+
+### Implementation Details
+
+#### Assumptions
+
+* Currently the EO CRD version is V1 and this would be a V2. We would assume that EO is able to react for both a V1 and V2 and if the user creates a V2
+  elasticsearch CR with the same name as an existing V1, the operator would migrate the V1 backing storage naming to be used for V2 (read: it would
+  rename the PVCs and rebind the backing PV so that it matches the updated naming convention).
```
periklis, line 180: Is renaming a PVC possible? IIRC the only changes allowed on PVCs after creation are changes on metadata labels and annotations.

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,319 @@
...
+in the form of t-shirt sizes that would encapsulate replication policies, range of node count, resources provided per node.
+
+This would need to be further fleshed out in a spike to determine what a Medium (the 90% use case) would be sized at, however some example sizes
+could be:
+
+* 0x.Small - A small (e.g. Dev test cluster) with zero redundancy. This would likely only be able to support a small amount of retention.
+  * This could be comprised of just 1 master node and 1-3 data+ingest nodes with 1:1 separate coordinating nodes (matching data nodes)
+* 1x.Medium - A cluster that should be able to cover 90% of our customer use cases and provide single redundancy
+  * 3 masters, 3-5 data+ingest nodes and 1:1 coordinating nodes
+* 1x.Large - For very large customer clusters with single redundancy
```
periklis, line 225: How does the naming convention apply to `MultipleRedudancy` and `FullRedudancy` as supported today?

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
```diff
@@ -0,0 +1,319 @@
...
+
+* Intended plan for upgrading would be:
+  * If there exists a V1 CR named "elasticsearch" in the namespace "openshift-logging"
+    and then a V2 CR named "elasticsearch" in the namespace "openshift-logging" is created then
+    the operator would ignore the V1 one and migrate the objects from V1 to V2 (objects primarily being currently used storage)
+
+  * The main concern with doing a straight migration (via transformation hook) is the need to try to extrapolate a desired t-shirt
+    size based on the current configurations which may be very difficult to get correct. It may also be that the customer's current
+    configuration does not correctly meet their demands so doing a migration could leave them with similar issues (mainly if their
+    storage capacity isn't high enough or if their cluster is under resourced)
```
periklis, line 292: What about upgrading the tracing instances?

periklis, line 292: If PVC renaming is not possible, we would need a mechanism to migrate PVCs, too

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
ewolinetz: you can't change the name of the pvc.. maybe i should reword this. the idea is we would change the name we use for PVC creation so we would create a PVC that follows that naming convention and then rebind the old PVC's backing PV to this new PVC and then remove the old one... effectively renaming.

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
ewolinetz: these are just example names, not ones that are set in stone. its up to the implementation to decide that

enhancements/cluster-logging/elasticsearch-v2-api-and-infrastructure.md
ewolinetz: there should be no difference. i simply gave "openshift-logging" as an example

PR #828, opened 2021-07-08
Add admin ack upgrade gate

jottofar:


enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,130 @@
...
+CVO currently blocks minor level upgrades when overrides are set, a resource deletion is pending, or an operator has set its upgradeable condition to false.
+This enhancement adds a generic upgrade blocking mechanism for blocking minor level upgrades.
+This generic upgrade blocker can consist of one or more items each of which will block a minor level upgrade.
+Each generic upgrade blocker item will be associated with some well-defined task to be completed by an administrator.
+Once the task has been completed the administrator will acknowledge completion and its associated item will no longer block minor level upgrades.
+
+## Motivation
+
+OCP 4.9 uses kubernetes v1.22 which is removing a large number of [APIs][k8s-1.22-removed-apis].
+We suspect that many customer clusters are still using these APIs through helm charts, operators, generic workloads, administrative tools, etc.
```
sdodson, line 40: I'd be sure to call out that we're specifically concerned about latent external use of APIs removed in 1.22. The cluster will generate alerts for active use.

Mostly just want to make it clear that we're not just being lazy, we believe there is no reasonable way for a Cluster Operator to provide a strong enough signal here.

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,138 @@
...
+### Non-Goals
+
+## Proposal
+
+CVO's existing framework for blocking minor level upgrades will be leveraged.
+The existing framework periodically checks upgradeable conditions.
+A new [upgradeable check method][upgradeableCheck-interface] will be added to the existing framework to implement this upgrade blocking mechanism.
+Use of this existing framework also allows this new upgrade blocking mechanism to be bypassed by using the `force` upgrade option.
+
+The new method will check a newly defined configmap for one or more predefined keys.
```
sdodson, line 62: Should these be predefined in code or could this be payload driven, ie: we add configmaps like this and each one is an independent ack.
```
apiVersion: v1
data:
  acknowledged: "false"
  message: |
    Kubernetes 1.22 and thus OpenShift 4.9 remove several APIs previously deprecated, please review the referenced documentation
  referenceURI: https://docs.openshift.com/yatta-yatta
kind: ConfigMap
metadata:
  annotations:
    clusterversion.openshift.io/admin-ack-request: "true"
    release.openshift.io/create-only: "true"
  name: admin-ack-request-kube-122-api-removals
  namespace: openshift-cluster-version
```

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,130 @@
...
+* This upgrade blocking mechanism must be compatible with SD, ACM, IKS/ROKS, Hive, and Hypershift.
+
+### Non-Goals
+
+## Proposal
+
+CVO's existing framework for blocking minor level upgrades will be leveraged.
+The existing framework periodically checks upgradeable conditions.
+A new [upgradeable check method][upgradeableCheck-interface] will be added to the existing framework to implement this upgrade blocking mechanism.
+Use of this existing framework also allows this new upgrade blocking mechanism to be bypassed by using the `force` upgrade option.
```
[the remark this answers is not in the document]
sdodson: If not I guess we may need version ranges as part of the schema so once they've upgraded to 4.9 we stop asking them about this specific ack?

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Implementation Details/Notes/Constraints
+
+### Risks and Mitigations
+
+* A common risk is human error resulting in an incorrect key/value pair being placed in the configmap.
+This type of error will not change `Upgradeable=false` and therefore the administrator will know a mistake has been made since the `Upgradeable` condition message will continue to be displayed.
+CVO will log an error whenever an unrecognized key is found in the configmap.
+* Another risk is that the configmap gets removed.
+This will result in `Upgradeable=false` and the standard `Upgradeable` condition message will be displayed with the additional message that the configmap cannot be found and should be created. (OPTION: Should CVO create the configmap?)
```
sdodson, line 93: This seems fine if the individual acks are enumerated in code, if they're a shipped manifest then we'd need them to be create-only and one per ACK?

enhancements/update/upgrades-blocking-on-ack.md
sdodson: This may be getting too close to needing a CRD / API.

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,138 @@
...
+Clusters upgraded to 4.9 will no longer have access to these [APIs][k8s-1.22-removed-apis].
+This new upgrade blocking mechanism will be used in an attempt to require administrators to read detailed documentation regarding the removed APIs and the possible impact.
+This OCP documentation will list which APIs have been removed and how an administrator checks their cluster for use of those APIs.
+
+4.8.z versions that are upgradeable to 4.9 will contain this enhancement and have the configmap `openshift-cluster-version/upgrade-ack`.
+CVO will set `Upgradeable=false` which in turn blocks cluster upgrades to 4.9 until the administrator has added the key/value pair `ack-4.9-api-removal-upgrade-approved: "true"` to the configmap.
+Until this is done the `Conditions` console page will display an `Upgradeable` condition message directing the administrator to the relevant OCP documentation.
+
+If the administrator wants to upgrade to 4.9 without performing the acknowledgment `force` can be used.
+
```
bparees, line 84: user stories are generally more in the form of:

as a cluster admin, i want to be warned that apis will be removed if i upgrade to 4.9 before i accidentally upgraded and be instructed how to check my cluster for potential impact

as a cluster admin i want to be able to override the ack requirement and upgrade anyway

as an OCP developer i want a way to introduce additional ACKs that admins will be required to set before upgrading to future minor OCP versions

as an ocp product team we do not want customers to be surprised by workloads breaking due to apis being removed after they upgrade


etc

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,138 @@
...
+### Implementation Details/Notes/Constraints
+
+### Risks and Mitigations
+
+* A common risk is human error resulting in an incorrect key/value pair being placed in the configmap.
+This type of error will not change `Upgradeable=false` and therefore the administrator will know a mistake has been made since the `Upgradeable` condition message will continue to be displayed.
+CVO will log an error whenever an unrecognized key is found in the configmap.
+* Another risk is that the configmap gets removed.
+This will result in `Upgradeable=false` and the standard `Upgradeable` condition message will be displayed with the additional message that the configmap cannot be found and should be created. (OPTION: Should CVO create the configmap?)
+CVO will log an error whenever there is an issue accessing the configmap.
```
bparees, line 94: presumably upgradeable will also be false in this case

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### Tech Preview -> GA
+
+N/A. This is not expected to be released as Tech Preview.
+
+#### Removing a deprecated feature
+
+### Upgrade / Downgrade Strategy
+
+These keys must be examined for validity for the next minor release and removed if no longer valid. For example, if the given 4.9 API breakage is no longer considered an issue when 4.10 is available the CVO `ack-4.9-api-removal-upgrade-approved` based logic and related documentation must be removed from 4.9.
```
bparees, line 119: i generally agree that the logic can/should be removed (specifically because we don't want a newly installed 4.9 cluster to be blocked from upgrading to 4.10 because the configmap is missing this key), but if we do remove the logic that expects this key, and the admin does not remove the key from the configmap (no reason for them to do so), then isn't the CVO going to start complaining about unknown keys in the configmap?

bparees: in general i think we want to encourage the admin to leave the previous acks in place, so they serve as a bit of history/context.  (this is also why i think it's important that the key names include the version they apply to)

enhancements/update/upgrades-blocking-on-ack.md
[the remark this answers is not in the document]
jottofar: No it's not. Yeah, sounds like being version aware will also help in other ways as well per https://github.com/openshift/enhancements/pull/828#discussion_r667161973

enhancements/update/upgrades-blocking-on-ack.md
jottofar: I like defining the key and message as you have above rather than hard-coding in CVO. I did interpret `we define a particular key name that the admin must create in the configmap with a value of “true” ` literally and write up as such thinking perhaps we didn't want to make it too easy for the admin to do the ack.  But we can still have a configmap to define the keys - its message and even it's valid version info - and  the separate configmap I defined that would be used by the admin to accomplish the ack'ing and checked by CVO. Plus we may want to keep the admin out of the "key config" configmap and limit to the less busy configmap used only for acking.

/cc @bparees

enhancements/update/upgrades-blocking-on-ack.md
jottofar: Perhaps another reason to have both a "keys config" configmap and an ack configmap with the above only applying to the latter.

enhancements/update/upgrades-blocking-on-ack.md
jottofar: Yes. That last statement is actually part of the above risk bullet `Another risk...`

enhancements/update/upgrades-blocking-on-ack.md
bparees: > we didn't want to make it too easy for the admin to do the ack. But we can still have a configmap to define the keys - its message and even it's valid version info - and the separate configmap I defined that would be used by the admin to accomplish the ack'ing and checked by CVO. 

this direction matches a conversation scott and i were having on slack.  I definitely don't want it to be "too easy" but having the gates be defined by a separate config resource seems reasonable to me.

enhancements/update/upgrades-blocking-on-ack.md
bparees: yeah, that risk is defined as "the configmap gets removed", i want to explicitly cover the risk "the configmap can't be accessed" (apiserver issue, rbac issue, etc).

enhancements/update/upgrades-blocking-on-ack.md
jottofar: Agree, not user stories. I put it under a new heading. I would have left the `User Stores` section empty since it seems requirements are pretty well flushed out but since you already had some good ones I put them in and will think about it some more.

enhancements/update/upgrades-blocking-on-ack.md
jottofar: made some additions

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,157 @@
...
+Each key represents a specific upgrade blocking condition that must be acknowledged before minor level upgrades are allowed.
+An administrator acknowledges a condition by adding the corresponding key with a value of `true` to the configmap.
+Values other than `true` will be ignored and therefore not clear the upgrade block.
+The keys will be named to make it explicitly clear what the administrator is acknowledging and thereby agreeing to.
+CVO will set `Upgradeable=false` which in turn blocks cluster upgrades to the next minor release until it finds each of the predefined keys in the configmap with a value of `true`.
+For each key blocking an upgrade, CVO will generate a detailed message pointing to relevant OCP documentation explaining what the administrator is expected to do before acknowledging task completion, i.e. entering the corresponding key into the configmap with a value of `true`.
+
+The new configmap `upgrade-ack` will be installed as part of the release in the namespace `openshift-cluster-version`.
+It will initially have no keys since the administrator is expected to add the relevant key as part of acknowledgement.
```
LalatenduMohanty, line 75: How the admin will be communicated or get to know that the config needs to me modified. I guess from the clusterversion status condition? or an alert? Also how we will we communicate the admin the exact command to edit the config map e.g. the patch command or the `oc edit command` .

enhancements/update/upgrades-blocking-on-ack.md
LalatenduMohanty: I think we need to clearly mention that option. I am in favor of having two config maps. One gets created by CVO for a situation where we need approval from admin to move forward with update, 2nd Admin creates to give approval. This way admin can delete and recreate if required without messing things up.

enhancements/update/upgrades-blocking-on-ack.md
LalatenduMohanty: I agree. I do not think we need to delete the configmap . But if admins want to delete it, they should be fine doing that right?

LalatenduMohanty (changes requested):

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,157 @@
...
+- [X] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+CVO currently blocks minor level upgrades when overrides are set, a resource deletion is pending, or an operator has set its upgradeable condition to false.
+This enhancement adds a generic upgrade blocking mechanism for blocking minor level upgrades.
```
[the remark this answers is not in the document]
jottofar: It's the blocking mechanism itself that we're adding which, yes, can be "resolved" (vs, bypass which sounds like `--force`) by the admin doing an ack. I explain that in the last two sentences.

enhancements/update/upgrades-blocking-on-ack.md
jottofar: The way it's written, since I saw no requirement about an alert, is the admin will be notified exactly as an admin is now when CVO blocks a minor release upgrade for other reasons. As far as what the admin needs to do before to get pass the block, the Condition message itself is to be very detailed and point admin to updated doc's with details:

> For each key blocking an upgrade, CVO will generate a detailed message pointing to relevant OCP documentation explaining what the administrator is expected to do before acknowledging task completion, i.e. entering the corresponding key into the configmap with a value of `true`.`

We could add an alert if that's what the consensus is but then that makes this minor level upgrade blocker differ from the others. Does it need to be?

enhancements/update/upgrades-blocking-on-ack.md
jottofar: Trying to nail down configmap(s), take a look...

I'm thinking there is one configmap per "gate" that defines the gate - name, message, applicable versions. CVO reads these configmaps to know what gates it needs to check, applicable version(s), message to generate if blocking. Then there's another configmap that the admin will be directed to to add the appropriate annotation, i.e. to do the ACK, for all defined gates. CVO is checking this configmap to ensure all defined gates have been entered or CVO will block minor upgrade.

For example below are two gates each defined by a configmap. Based on this config CVO will be checking `upgrade-ack` configmap (defined at bottom) for annotations `ack-4.9-api-removal-upgrade-approved: "true"` and `ack-another-ack: "true"` 
otherwise minor upgrades will be blocked.

```
apiVersion: v1
data:
  ackAnnoName: "ack-4.9-api-removal-upgrade-approved"
  message: |
    Kubernetes 1.22 and thus OpenShift 4.9 remove several APIs previously deprecated, please review the referenced documentation
  referenceURI: https://docs.openshift.com/yatta-yatta
  versions: "4.9"
kind: ConfigMap
metadata:
  name: admin-ack-request-kube-122-api-removals
  namespace: openshift-cluster-version
---
apiVersion: v1
data:
  ackAnnoName: "ack-another-ack"
  message: |
    Here's another ack, please review the referenced documentation...
  versions: "4.9,4.10"
kind: ConfigMap
metadata:
  name: admin-ack-another
  namespace: openshift-cluster-version
```
I'm thinking the admin does not have to create this configmap but merely update it with the appropriate anno which will be explained in detail in the doc's the admin is directed to:
```
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
  name: upgrade-ack
  namespace: openshift-cluster-version
```

bparees: the version numbers should probably indicate "you need this ack if you are on this version" rather than "you need this ack before you can go to this version".  in case we ever allow skipping versions somehow.

jottofar: Actually, wouldn't my one-per-gate example configmaps conflict because they both contain the same keys, e.g. ackAnnoName, etc?

For that reason, and because I think it just makes more sense, we should have one configuration config map that defines everything about each "gate" as in:
```
apiVersion: v1
metadata:
  name: admin-ack-config
  namespace: openshift-cluster-version
data:
   upgrade.ack.configuration: |
      upgrade.ack.gates {
         gate: [
            {
               name: "ack-4.9-api-removal-upgrade-approved"
               valid.in.versions: "4.9"
               message: |
                  Kubernetes 1.22 and thus OpenShift 4.9 remove several APIs previously deprecated, please review the referenced documentation
               referenceURI: https://docs.openshift.com/yatta-yatta
            }
            {
               name: "ack-foo"
               valid.in.versions: "4.9,4.10"
               message: |
                  Foo bar, please review the referenced documentation
               referenceURI: https://docs.openshift.com/foo-bar
            }
         ]
      }

CVO reads 0 to n gates and for each read, if it applies to the current OCP version, uses the `name` field to look it up in the `upgrade-ack` configmap and if not found to have been "ACK'ed" blocks minor upgrades.
If we decided a gate no longer applied, for example `ack-foo` becomes OBE in 4.10.9, we could just remove it from the above configmap.

File: enhancements/update/upgrades-blocking-on-ack.md (hunk 3 above)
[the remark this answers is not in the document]
Comment (jottofar) [reply enhancements/update/upgrades-blocking-on-ack.md:64]: Added version requirement to blocker name.

File: enhancements/update/upgrades-blocking-on-ack.md (hunk 2 above)
Comment (jottofar) [reply enhancements/update/upgrades-blocking-on-ack.md]: After discussing with @sdodson and @bparees we decided against additional configmap(s) due to added complexity of maintaining and consuming and instead go with simplicity of hard-coded key names.

File: enhancements/update/upgrades-blocking-on-ack.md (hunk 7 above)
Comment (jottofar) [reply enhancements/update/upgrades-blocking-on-ack.md]: Removal of keys no longer required since their name will version scope them.

File: enhancements/update/upgrades-blocking-on-ack.md [hunk 10]
```diff
@@ -0,0 +1,162 @@
...
+## Proposal
+
+### General Implementation
+
+CVO's existing framework for blocking minor level upgrades will be leveraged.
+The existing framework periodically checks upgradeable conditions.
+A new [upgradeable check method][upgradeableCheck-interface] will be added to the existing framework to implement this upgrade blocking mechanism.
+Use of this existing framework also allows this new upgrade blocking mechanism to be bypassed by using the `force` upgrade option.
+
+The new method's logic will be driven by predefined configmap key names each of which represents a potential minor level upgrade blocker.
```
bparees, line 66: this isn't accurate anymore, right? it's just hardcoded logic.

enhancements/update/upgrades-blocking-on-ack.md
bparees: I think as long as we are satisfied w/ the existing UX when an operator is in an "upgradeable=false" status, that is fine.

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,162 @@
...
+## Drawbacks
+
+* CVO will contain a hard-coded key name for each specific acknowledge based upgrade block.
+Configuring these keys through one or more configuration configmaps was considered but decided against due to the added complexity of maintaining and consuming the additional configmaps.
+* Keys will not be portable across products that use a different versioning scheme than OCP.
+* `Upgradeable=false` is the default state and results if there is any issue accessing the configmap or a key is invalid.
+
+## Alternatives
+
+[k8s-1.22-removed-apis]: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22
```
bparees, line 161: in what sense is this an alternative?

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,162 @@
...
+[upgradeableCheck-interface]: https://github.com/jottofar/cluster-version-operator/blob/e1adca47977bbea635b7fb2dc2473525e512a539/pkg/cvo/upgradeable.go#L132
```
bparees, line 162: not clear why this is under "alternatives"

enhancements/update/upgrades-blocking-on-ack.md
jottofar: Yes, hardcoded logic. I just mean that the hardcoded logic is based on `predefined configmap key names`. I will reword to make that more clear.

enhancements/update/upgrades-blocking-on-ack.md
jottofar: It's not. It's just the link definitions. They only show up when file is viewed in raw form. I'll pust something under the `Alternatives` heading so it is not confusing in the raw.

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,162 @@
...
+### General Implementation
+
+CVO's existing framework for blocking minor level upgrades will be leveraged.
+The existing framework periodically checks upgradeable conditions.
+A new [upgradeable check method][upgradeableCheck-interface] will be added to the existing framework to implement this upgrade blocking mechanism.
+Use of this existing framework also allows this new upgrade blocking mechanism to be bypassed by using the `force` upgrade option.
+
+The new method's logic will be driven by predefined configmap key names each of which represents a potential minor level upgrade blocker.
+The key name will include an OCP major/minor verion number that indicates to which OCP upgrade target version the key applies.
+For example, a key name containing `4.9` would only apply to an OCP cluster currently at version 4.8 and can therefore only block 4.8 to 4.9 upgrades.
```
sdodson, line 68: This may be an implementation detail that we need not call out here, but is the controller strictly parsing the defined key names for the version or is it just expected to be defined in the hardcoded gating code? If we're relying on the key name perhaps we need to define the format for the names?

jottofar:
/retitle Bug 1978376: Add ack based upgrade blocker

enhancements/update/upgrades-blocking-on-ack.md
jottofar: Defined key name format.

sdodson:
/cc @csrwng @relyt0925 @jeremyeder 
Can we please get your review on this as a relatively urgent matter. We need to get this enhancement to a state where we can get the implementation into 4.9 and backported to 4.8.z before we allow upgrades to 4.9.

bparees:
@csrwng @relyt0925 @jeremyeder specifically what we are looking for from you is confirmation that SD/IKS/Hypershift will be able to give admins a way to set this additional ACK prior to upgrade, or at least that you understand the implication for custom logic/actions needed before being able to upgrade a cluster.  (tldr: upgrading to 4.9 will require a new key/value entry be created in a specific configmap which serves as the admin "acking" they understand the implication of 4.9 api removals, before you can start the cluster upgrade).

bparees:
@dgoodwin @jnpacker adding you guys for ACM/Hive oversight.  Same ask as https://github.com/openshift/enhancements/pull/828#issuecomment-880725437

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,181 @@
...
+This OCP documentation will list which APIs have been removed and how an administrator checks their cluster for use of those APIs.
+
+4.8.z versions that are upgradeable to 4.9 will contain this enhancement and have the configmap `openshift-cluster-version/upgrade-ack`.
+CVO will set `Upgradeable=false` which in turn blocks cluster upgrades to 4.9 until the administrator has added the key/value pair `ack-4.8-apis-being-removed-in-4.9: "true"` to the configmap.
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    ack-4.8-apis-being-removed-in-4.9: "true"
```
dgoodwin, line 102: Any reason we're using an annotation here instead of a data key in the config map itself?

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,181 @@
...
+  name: upgrade-ack
+  namespace: openshift-cluster-version
+```
+Until this is done the `Conditions` console page will display an `Upgradeable` condition message directing the administrator to the relevant OCP documentation.
+
+If the administrator wants to upgrade to 4.9 without performing the acknowledgment `force` can be used.
```
dgoodwin, line 108: Cloud credential operator has done something similar to this where if you're using manual mode, where an administrator is responsible for managing all credentails/permissions, we require the admin to annotate every CredentailsRequest to indicate they've done the work to prepare for an upgrade.

@akhil-rane @joelddiaz would we be interested in using this approach instead? It implies an "ack" key be automatically present for every minor version for clusters in manual mode.

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Goals
+
+* Acknowledgment requires the explicit update of a cluster resource by an administrator.
+* For the initial release in 4.8.z, administrator acknowledgment of API removal will be required before a minor level upgrade to 4.9 is allowed.
+* Upgrade blocking mechanism must be generic such that it can be repurposed in the future.
+* It must be clear what an administrator is acknowledging.
+* Administrators can bypass the upgrade blocking mechanism by using the existing CVO "force" implementation.
+* The UI will clearly indicate when this upgrade blocking mechanism is blocking upgrades, why, and what needs to be done to clear the block.
+* This upgrade blocking mechanism must be compatible with SD, ACM, IKS/ROKS, Hive, and Hypershift.
```
dgoodwin, line 53: Hive doesn't really deal with upgrades today but it's something we've always talked about.

@jnpacker what do you think, does this fit best in ACM or Hive?

Will there be a programatic way for a tool initiating the upgrade to query the ack's required for an upcoming release image the cluster may upgrade to? Something in the release image ideally.

enhancements/update/upgrades-blocking-on-ack.md
bparees: i missed this, i would also have expected it to be data in the configmap.

enhancements/update/upgrades-blocking-on-ack.md
bparees: > Will there be a programatic way for a tool initiating the upgrade to query the ack's required for an upcoming release image the cluster may upgrade to? Something in the release image ideally.

there isn't one currently defined, no.  When you say something in the release image, what do you have in mind?

enhancements/update/upgrades-blocking-on-ack.md
sdodson: Agreed, should just be a key.

enhancements/update/upgrades-blocking-on-ack.md
dgoodwin: some of kind of manifest that something like ACM or Hive could parse to put infront of the user, here are the things you need to opt-in for before this cluster will upgrade. 

Actually that leads to some questions, how does the CVO know what keys it needs? Do we backport what is needed for 4.9 into 4.8.z? Or is the CVO actively examining possible upgrade endpoints and again, parsing out the keys, so it knows it's not upgradable. (apologies if I missed this)

enhancements/update/upgrades-blocking-on-ack.md
sdodson: I think CCO current implementation is nice because the "ack" is in context of the CredentialsRequest, while they could certainly patch them without looking at the context of the actual request it sort of puts it in the admin's face.

enhancements/update/upgrades-blocking-on-ack.md
bparees: The gating logic will go into 4.8.z, not 4.9 (for this particular gate anyway).  Right now the plan is to just code the logic in the CVO itself.  We had an earlier iteration where the gates were defined via resources, but it started to get complicated.

dgoodwin: How could ACM/Hive know what flags need to be set?

bparees: honestly my thought was the ACM would be specifically aware of the acks a particular OCP update requires and include code accordingly.

but we are discussing moving the list of required gates into a configmap again.

enhancements/update/upgrades-blocking-on-ack.md
joelddiaz: IIRC, you only annotate the CCO's config CR, not the individual CredentialsRequest CRs

enhancements/update/upgrades-blocking-on-ack.md
jottofar: Changed to use `data:`

enhancements/update/upgrades-blocking-on-ack.md
jottofar: Based on team discussion added additional configmap (for a total of two new configmaps) to contain keys.

dgoodwin: This looks good, thanks!

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,210 @@
...
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: admin-gates
+  namespace: openshift-cluster-version
+data:
+  ack-4.8-kube-122-api-removals-in-4.9: |
+     Detailed description of what admin needs to do...Read documentation here https://
+  ack-4.8-gate-2: |
```
wking, line 77: CVO's current ConfigMap handler [uses `mergeMap` for `data`][1], and [`mergeMap` ensures that manifest-side keys have the requested values][2], but does not remove keys which are not mentioned in the manifest.  Are we comfortable with this ConfigMap being append-forever?  Or do we want to define a pruning strategy?  Like "CVO will special-case logic to remove keys from both ConfigMaps unless they match 4.y for the current or a future y"?  Or something?

[1]: https://github.com/openshift/cluster-version-operator/blob/3a68652568e9075c23f491bc8c037942bd67ec82/lib/resourcemerge/core.go#L18
[2]: https://github.com/openshift/cluster-version-operator/blob/3a68652568e9075c23f491bc8c037942bd67ec82/lib/resourcemerge/meta.go#L28-L41

bparees: personally i don't see an issue w/ append forever, but i'm interested to hear concerns

1) this should be rare
2) we are planning a UX around this(CVO status reporting) so it's not like an admin will need to look through this configmap to figure out what applies to their cluster currently

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,210 @@
...
+Major milestones in the life cycle of a proposal should be tracked in `Implementation History`.
+
+## Drawbacks
+
+* Keys will not be portable across products that use a different versioning scheme than OCP.
+* `Upgradeable=false` is the default state and results if there is any issue accessing either configmap or if an invalid key is found in the `admin-acks` configmap.
+
+## Alternatives
+
+No alternatives have been discussed.
```
wking, line 207: Can we mention [the cloud-cred approach][1] ([`cloudcredential.openshift.io/upgradeable-to`][2])?  Because with this enhancement we're moving from "each component handles this directly and feeds the CVO via their ClusterOperator's `Upgradeable` condition" to "that's still allowed, but we're also providing this CVO-hosted UI", and I think it's worth discussing our motivation.  Probably:

> Admins may not want to go hunting through a bunch of component-specific UXes.  And mumble, mumble, about why browsing `ClusterOperator` `Upgradeable=False` condition messages wouldn't be enough to point admins at the steps they should be taking...

or something?

[1]: https://github.com/openshift/enhancements/pull/828#discussion_r670630101
[2]: https://github.com/openshift/cloud-credential-operator/pull/286/files#diff-f6d93feef6d59ca8ced14b54fe0058d41f14de44450baf6b723cc1c2ec280c54R132

joelddiaz: CCO's needs for admin signoff are ongoing (not one-offs). So we came up with a signoff where the admin says "i'm okay to upgrade to 4.X" (this should in theory allow upgrades from one EUS version to the next EUS version).
But it is odd for CCO to have its own mechanism that exists outside of this ConfigMap approach. Perhaps it is justified as CCO's needs are not release-specific, but ongoing pre-upgrade maintenance expected from the admin (and CCO's requirements about what is required from the admin is dependent on how CCO is configured to run eg. Manual/Mint/Passthrough/etc).

bparees: > Admins may not want to go hunting through a bunch of component-specific UXes. And mumble, mumble, about why browsing ClusterOperator Upgradeable=False condition messages wouldn't be enough to point admins at the steps they should be taking...

The motivation in my mind is more that this isn't a specific component limitation/issue.  (You can argue that the apiserver should set upgradeable=false since the apiserver is removing the apis, but the apiserver isn't going to break if you upgrade to 4.9, its your workloads that are going to break)

but yes, agree that this shouldn't be setting a new precedent that the CVO manages upgradeable=false for everything.

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,210 @@
...
+For example, a key name beginning with `ack-4.8` would only apply to an OCP cluster currently at version 4.8.
+The key's value will be a detailed message pointing to relevant OCP documentation explaining what the administrator is expected to do before acknowledging task completion.
+If running within the applicable version, CVO will check a newly defined configmap, `admin-acks`, for the key.
+The key represents a specific upgrade blocking condition that must be acknowledged before a minor level upgrade is allowed.
+An administrator acknowledges a condition by adding the corresponding key with a value of `true` to the configmap `admin-acks`.
+Values other than `true` will be ignored and therefore not clear the upgrade block.
+The keys will be named to make it explicitly clear what the administrator is acknowledging and thereby agreeing to.
+CVO will set `Upgradeable=false` which in turn blocks cluster upgrades to the next minor release until it finds each of the applicable predefined keys in the configmap with a value of `true`.
```
csrwng, line 89: Just to clarify, is this `Upgradeable=false` condition going to be set on the ClusterVersion status? Currently the Upgradable condition seems to only be set on ClusterOperator status(es)

bparees: i'll be honest, i thought the CVO already had its own clusteroperator resource but I see that it does not seem to.  We could certainly introduce one, but since @sdodson and @jottofar got this far w/o raising the question i'm going to hope that means they had something else in mind.

jottofar: CVO currently sets ClusterOperatorStatusCondition `Upgradeable=false` when cluster version overrides exist or there is a pending CVO initiated resource deletion (via the recently added manifest deletion support). This would add an additional one for ack(s).

bparees: sets it where?  on ClusterVersion?

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,210 @@
...
+This enhancement adds a generic upgrade blocking mechanism for blocking minor level upgrades.
+This generic upgrade blocker can consist of one or more items each of which will block a minor level upgrade.
+Each generic upgrade blocker item will be associated with some well-defined task to be completed by an administrator.
+Once the task has been completed the administrator will acknowledge completion and its associated item will no longer block minor level upgrades.
+
+## Motivation
+
+OCP 4.9 uses kubernetes v1.22 which is removing a large number of [APIs][k8s-1.22-removed-apis].
+We suspect that many customer clusters are still using these APIs through helm charts, operators, generic workloads, administrative tools, etc.
+We believe there is no effective way for a Cluster Operator to detect the use of these APIs by external applications.
```
dofinn, line 41: 4.8 is introduced `apirequestcount` that highlights usage of depreciating APIs in next release. This covers this use-case to a degree but not completely. 

```
[~ {production} (cluster:default)]$ oc get apirequestcounts --as backplane-cluster-admin | grep 1.22
NAME REMOVEDINRELEASE REQUESTINCURRENTHOUR REQUESTSINLAST24H
certificatesigningrequests.v1beta1.certificates.k8s.io                     1.22               0                       373
clusterrolebindings.v1beta1.rbac.authorization.k8s.io                      1.22               32                      2721
customresourcedefinitions.v1beta1.apiextensions.k8s.io                     1.22               0                       488
ingresses.v1beta1.extensions                                               1.22               6                       428
ingresses.v1beta1.networking.k8s.io                                        1.22               0                       0
mutatingwebhookconfigurations.v1beta1.admissionregistration.k8s.io         1.22               0                       253
openshiftapiservers.v1.operator.openshift.io                                                  189                     13223
operators.v1.operators.coreos.com                                                             29                      1822
rolebindings.v1beta1.rbac.authorization.k8s.io                             1.22               0                       0
roles.v1beta1.rbac.authorization.k8s.io                                    1.22               0                       0
validatingwebhookconfigurations.v1beta1.admissionregistration.k8s.io       1.22               0                       170
```

dofinn:
Unpacking an SD perspective...

**RBAC**
Managed Openshift (OSD/ROSA/ARO) would require access be granted to the [dedicated-admin role](https://github.com/openshift/managed-cluster-config/tree/master/deploy/ccs-dedicated-adminsl) (not applicable in ARO?) and of course could be reviewed with cluster-admin if the customer has access to it. 

**OCM**
Would CVO write a metric for admin-ack cases that can we exposed via telemetry so OCM/ClusterServices can consume it?  This would enable surfacing of it via the OCM console so customers can review each "blocker" for each cluster they own. 

**4.8->4.9 use case** 

https://github.com/openshift/enhancements/pull/828/files#r672722878

For this use-case in particlar for 4.8-4.9, we could make use of the new `apirequestcount` API (coming 4.8) that actually shows depreciation and usage of APIs in the next Y stream release. There is also alerts raised to highlight use of the depreciating APIs

```
[~ {production} (clusterID:default)]$ oc get apirequestcounts --as backplane-cluster-admin | grep 1.22
NAME REMOVEDINRELEASE REQUESTSINCURRENTHOUR REQUSTSINLAST24H
certificatesigningrequests.v1beta1.certificates.k8s.io                     1.22               0                       373
clusterrolebindings.v1beta1.rbac.authorization.k8s.io                      1.22               32                      2721
customresourcedefinitions.v1beta1.apiextensions.k8s.io                     1.22               0                       488
ingresses.v1beta1.extensions                                               1.22               6                       428
ingresses.v1beta1.networking.k8s.io                                        1.22               0                       0
mutatingwebhookconfigurations.v1beta1.admissionregistration.k8s.io         1.22               0                       253
rolebindings.v1beta1.rbac.authorization.k8s.io                             1.22               0                       0
roles.v1beta1.rbac.authorization.k8s.io                                    1.22               0                       0
validatingwebhookconfigurations.v1beta1.admissionregistration.k8s.io       1.22               0                       170
```

**managed-upgrade-operator** 

Given that we (SD) utilizes [MUO](https://github.com/openshift/managed-upgrade-operator), upgrades are blocked based on the presence of alerts. so the 4.8->4.9 use case higlighed in this enhancement could potentially already be covered. 

cc @mrbarge  @jharrington22  @cblecker 

---

Final thoughts. 

+1 from me on the basis that we can push the admin-ack status around with metrics via telemetry so we can include the state in the OCM ecosystem and make it consumable by other operators (MUO)

enhancements/update/upgrades-blocking-on-ack.md
sdodson: I don't see an issue with it either.

enhancements/update/upgrades-blocking-on-ack.md
sdodson: On the cluster-version operator's status, correct @jottofar ?

enhancements/update/upgrades-blocking-on-ack.md
sdodson: I think you got the point based on your mention that it's not complete, but just to be sure I'll clarify. While we can use the new `apirequestcount` in 4.8 to measure and alert on active use of these APIs it doesn't capture atent use, think a CI build pipeline outside of the cluster that hasn't run for months but uses removed APIs to push a critical workload update after it's been upgraded to 4.9. This is a prompt for the admin to go round up all those use cases and work with their userbase and ISVs.

enhancements/update/upgrades-blocking-on-ack.md
wking: yup, we have `Upgradeable` on the ClusterVersion, and it's a union of what we collect from ClusterOperator and the CVO's own opinions on things like this (currently the presence of meaningful `overrides` in ClusterVersion's `spec`.  And any in-progress deletions.  [Source][1]).

[1]: https://github.com/openshift/cluster-version-operator/blob/d58378108819cf8455600521bc5186acb7d201da/pkg/cvo/upgradeable.go#L239-L241

wking:
> Would CVO write a metric for admin-ack cases that can we exposed via telemetry so OCM/ClusterServices can consume it? This would enable surfacing of it via the OCM console so customers can review each "blocker" for each cluster they own.

Hmm.  I can see the appeal of something like a mass-ack for many clusters at once.  But I dunno if that's worth the bump to Telemetry cardinality.  Do we really expect to have a many-cluster-mass-ack for this?  Or is whatever we do today to surface `Upgradeable=False` sufficient from the OCM perspective?

enhancements/update/upgrades-blocking-on-ack.md
jottofar: I don't think there's an action here?

I don't see this as setting a new precedent either. We're setting Upgradeable=false based on a cluster wide condition - the general case being we've asked the admin to do something and then ack that it's done.

enhancements/update/upgrades-blocking-on-ack.md
csrwng: @wking if the Upgradeable condition is not present, can we assume that it is true?

dofinn:
> > Would CVO write a metric for admin-ack cases that can we exposed via telemetry so OCM/ClusterServices can consume it? This would enable surfacing of it via the OCM console so customers can review each "blocker" for each cluster they own.
> 
> Hmm. I can see the appeal of something like a mass-ack for many clusters at once. But I dunno if that's worth the bump to Telemetry cardinality. Do we really expect to have a many-cluster-mass-ack for this? Or is whatever we do today to surface `Upgradeable=False` sufficient from the OCM perspective?

The other option would be to retro fit SRE/SD needs via https://github.com/openshift/osd-metrics-exporter and then remoteWrite to SD MST (Observatorium). This would require OCM/CS to then ingest metrics from Telemetry + MST. This implimentation would also a "shift right" cc @jeremyeder

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,210 @@
...
+Relying solely on automated API detection to block upgrades could result in blocking for APIs that are not actively used any longer or not blocking because an application was not active during the check.
+Having a human in the loop to drive the analysis and make the final decision is the surest way to prevent customers from upgrading from 4.8.z to 4.9.z and encountering missing API related breakage.
```
jewzaam, line 43: This doesn't prevent upgrades when API related breakage will happen, it moves the responsibility to the administrator.

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Goals
+
+* Acknowledgment requires the explicit update of a cluster resource by an administrator.
+* For the initial release in 4.8.z, administrator acknowledgment of API removal will be required before a minor level upgrade to 4.9 is allowed.
+* Upgrade blocking mechanism must be generic such that it can be repurposed in the future.
+* It must be clear what an administrator is acknowledging.
+* Administrators can bypass the upgrade blocking mechanism by using the existing CVO "force" implementation.
```
jewzaam, line 51: So any customer forcing upgrades as their normal means of upgrade will never see this warning.  Are there metrics on how many upgrades done to date that force and would skip this check?  Note ARO upgrades executed by SRE (not by customer) always "force" though ARO does not execute minor upgrades.. just including as a note of a large scale use of forced upgrades.

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,210 @@
...
+* The UI will clearly indicate when this upgrade blocking mechanism is blocking upgrades, why, and what needs to be done to clear the block.
+* This upgrade blocking mechanism must be compatible with SD, ACM, IKS/ROKS, Hive, and Hypershift.
```
jewzaam, line 53: When I think of SD compatibility I have a key concern: how do we keep customers from directly manipulating this configuration?  If this is part of a new CRD then RBAC can be managed better and webhooks created to block operations from cluster-admins if needed on OSD, ROSA, and ARO.  If this configuration is included in something already existing like ClusterVersion it becomes difficult to cordon permissions from customer.  If it is something generic like a configmap it may be OK if it's a specific name and we never permit customers to create that resource.  This isn't just in protecting the blocking of edges but also from keeping customers from setting the block manually.  Not clear how this will be used in the future so I have to assume behavior can change, therefore I'd like clear ways to manage permissions for users. 

So a part of this should be SD can easily prevent customers from directly manipulating this configuration.

enhancements/update/upgrades-blocking-on-ack.md
bparees: correct....are you questioning that as an axiom, or looking for better phrasing, or something else?

fundamentally this EP represents us saying "we cannot solve this problem for admins, therefore we need admins to be accountable for addressing it, with our best guidance"

enhancements/update/upgrades-blocking-on-ack.md
bparees: the two configmaps in this EP will have well defined names and live in a well defined namespace.  Does that satisfy your concerns?

(one of them defines the gates and should never be edited by an admin at all..if they try, the CVO is going to stomp the change back anyway, the other is where the admin puts their ack of the gate)

enhancements/update/upgrades-blocking-on-ack.md
jottofar: The `Conditions` console page displays the message regardless. `force` just allows this and other blockers to be ignored for purposes of an upgrade.

enhancements/update/upgrades-blocking-on-ack.md
jewzaam: Yes, especially if it's not editable.  If that changed, as this does talk of using this pattern for other things, having a static name and creation managed by the platform will address rbac needs.

enhancements/update/upgrades-blocking-on-ack.md
jewzaam: Better phrasing, shifting responsibility sounds like the best approach in this situation.

```suggestion
Having a human in the loop to drive the analysis and make the final decision is the surest way to prevent customers from upgrading from 4.8.z to 4.9.z and reduce instances of missing API related breakage by allowing administrators to assess the impact per cluster and impacted application.
```

enhancements/update/upgrades-blocking-on-ack.md
bparees: when we talk of using this pattern for other things, we mean there might be other cases where we add additional key/values to the gate configmap, and require the admin ack them in the ack configmap.

the actual configmap name/namespace would be the same one that's being defined here, not new/additional ones.

enhancements/update/upgrades-blocking-on-ack.md
jottofar: If I understand your question correctly @csrwng, the condition we're proposing to add in this enhancement, which would be added [here](https://github.com/openshift/cluster-version-operator/blob/d58378108819cf8455600521bc5186acb7d201da/pkg/cvo/upgradeable.go#L238), can be nil which is checked [here](https://github.com/openshift/cluster-version-operator/blob/d58378108819cf8455600521bc5186acb7d201da/pkg/cvo/upgradeable.go#L37).

enhancements/update/upgrades-blocking-on-ack.md
jottofar: Reworded with slight change. Our goal here is to eliminate API breakage rather than to just reduce.

enhancements/update/upgrades-blocking-on-ack.md
csrwng: Thanks, from a hypershift perspective I just want to make sure that we have a clear signal as to whether the cluster is upgradeable by either looking at ClusterVersion or ClusterOperator(s). We don't rely on the CVO to perform an upgrade, but we want to report in our status whether the cluster can be upgraded and let the service provider make the decision whether to force it or not.

sdodson:
@dofinn You'll get Upgradeable=False conditions backhauled through telemetry, one of which would be this new condition as well as alerts generated by the `apirequestcount` alerts that exist already. Is that sufficient for your needs or is there something else that's a hard requirement to enable OSD/ROSA/etc?

dofinn:
> @dofinn You'll get Upgradeable=False conditions backhauled through telemetry, one of which would be this new condition as well as alerts generated by the `apirequestcount` alerts that exist already. Is that sufficient for your needs or is there something else that's a hard requirement to enable OSD/ROSA/etc?

@sdodson Could you please provide an example of what this metric would look like? If we can surface that upgradeable is either true/false and a related condition as to why that true/false exists, then i think that is sufficient.

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,211 @@
...
+  - "@jottofar"
+reviewers:
+  - "@LalatenduMohanty"
+  - "@bparees"
+  - "@sdodson"
+  - "@csrwng"
+  - "@relyt0925"
+  - "@jeremyeder"
+approvers:
+  - TBD
```
dhellmann, line 13: We should get someone to sign up to be approvers for when we have consensus. Maybe @sdodson is a good person?

enhancements/update/upgrades-blocking-on-ack.md
dhellmann: The detail in this thread would make a good addition to the body of the doc.

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,211 @@
...
+Having a human in the loop to drive the analysis and make the final decision is the surest way to prevent customers from upgrading from 4.8.z to 4.9.z and eliminate instances of missing API related breakage by allowing administrators to assess the impact per cluster and impacted application.
+
+### Goals
+
+* Acknowledgment requires the explicit update of a cluster resource by an administrator.
+* For the initial release in 4.8.z, administrator acknowledgment of API removal will be required before a minor level upgrade to 4.9 is allowed.
+* Upgrade blocking mechanism must be generic such that it can be repurposed in the future.
+* It must be clear what an administrator is acknowledging.
+* Administrators can bypass the upgrade blocking mechanism by using the existing CVO "force" implementation.
+* The UI will clearly indicate when this upgrade blocking mechanism is blocking upgrades, why, and what needs to be done to clear the block.
```
dhellmann, line 53: I don't see anyone from the console team on the reviewer list at the top of the doc. Who should be pulled in to coordinate for this goal?

enhancements/update/upgrades-blocking-on-ack.md
```diff
@@ -0,0 +1,211 @@
...
+This will result in `Upgradeable=false` and the standard `Upgradeable` condition message will be displayed with the additional message that the configmap cannot be found and should be created.
+CVO will recreate the `admin-gates` configmap if removed since it will be managed by CVO.
+CVO will log an error whenever the `admin-acks` configmap has been removed.
+* Unable to access the `admin-acks` configmap.
+This will result in `Upgradeable=false` and the standard `Upgradeable` condition message will be displayed with the additional message that the configmap cannot be accessed.
+CVO will log an error whenever there is an issue accessing the `admin-acks` configmap.
+* Unable to access the `admin-gates` configmap.
+Since CVO will not be able to determine what if any minor level upgrade blocks are defined CVO will set `Upgradeable=false` and set the displayed `Upgradeable` condition message to indicate that the `admin-gates` configmap cannot be accessed.
+CVO will log an error whenever there is an issue accessing the `admin-gates` configmap.
+
```
dhellmann, line 167: Is there a risk for EUS updates if only some of the acks are provided, allowing a partial upgrade stopping at 4.x+1 but not progressing to 4.x+2? Or does the logic check for all required acks between 4.x and the target version?

enhancements/update/upgrades-blocking-on-ack.md
dhellmann: If there was an alternative discussed and rejected, it would be good to capture that info in the doc for completeness.

enhancements/update/upgrades-blocking-on-ack.md
bparees: we don't necessarily need anyone, the current assumption is that the console already exposes info when upgrades are blocked, and that the mechanism we are using here is one that the console already watches.  That assumption needs to be confirmed/validated of course, but we don't need the console team explicitly to do that.

if the assumption is incorrect, then we'll need more details in this EP about what additional console function is needed and then we'd need the console team to review. We'd likely start that conversation w/ sam padgett.

enhancements/update/upgrades-blocking-on-ack.md
bparees: there is no automated update from one EUS to the next, it's already a manual process where the admin does each sequential update explicitly, so while they might update from 4.x to 4.x+1 and then find they can't progress to 4.x+2 w/o doing some additional ack(which i agree isn't great if they suddenly find they need a week to resolve the ack before they can continue their upgrade process), it's not like it's blocking an automated "get me from 4.x to 4.x+3" process the admin kicked off(because no such process exists today, afaik).

i'm not really sure how we'd solve this.  
1) when you're at 4.6, 4.6 does not know what requirements may exist to get from 4.8 to 4.9, unless we start backporting stuff all the way to 4.6 once we discover it in 4.8/4.9.

2) when you're on 4.6 we don't actually know your intention as far as what level you are intending to get to next (as i said, there's no "upgrade me to 4.10" button, it's just "upgrade me to 4.7.  and then upgrade me to 4.8.  and then.....", so it seems difficult to communicate "hey, fyi, you're on 4.6 and that's fine, and you can get to 4.7 and 4.8, that's fine too, but if you intend to get to 4.9 you're going to get stuck at some point".  We might have a hint based on you being on the "EUS" channel instead of the "stable" channel, though i'm not clear what our longterm vision is for that channel(are we going to put 4.7+4.8+4.9 into that channel when 4.10EUS comes out?  or are we going to tell people to switch to the stable channel to get to 4.10 and then switch back to the EUS channel?).

it feels like something that belongs in the EUS upgrade guidance(hopefully we have such a thing), i'm sure this isn't the only "gotcha" we want admins to think about when they are getting ready to go from 4.6 to 4.10, as opposed to just going from 4.6 to 4.7.

enhancements/update/upgrades-blocking-on-ack.md
bparees: the main alternative i can think of is we debated whether the apiserver should be the component setting upgradeable=false and looking for the ack, since it is the component that is "removing" the apis.

we rejected it because:
1) the apiserver is removing the apis, but the apiserver is not blocked from upgrading because of that..it would be more than happy to upgrade, regardless of what apis are in use.
2) this mechanism feels like a generically useful concept so implementing it in a way that we can reuse it for other similar gates that require manual admin action felt right, vs tying it to a specific component.


I guess another rejected alternative was to just use the apiserver's data about what apis are in use, and block upgrades if removed apis are in use.  It was rejected because that information isn't sufficient to make an informed choice (just because you used the api last week doesn't mean you care about it, and just because you didn't use it in the last week doesn't mean you don't have a tool or workload that is going to try to use it tomorrow)

enhancements/update/upgrades-blocking-on-ack.md
sdodson: I think it is, though perhaps not quite as clearly in line 43.

enhancements/update/upgrades-blocking-on-ack.md
sdodson: Given the CVO would be updated at each step along that path I think there's no real concern here.

We've sort of sputtered out on ensuring that 4.N can signal all of the requirements to get to 4.N+4 and I don't see us achieving that goal so I'm not too worried about it. If we stick with 4 minor stride between EUS versions maybe we try to address this in 4.10 to 4.14, which means we'd need to revisit this as soon as we start 4.10 dev.

enhancements/update/upgrades-blocking-on-ack.md
sdodson: @jottofar Do you mind copying and pasting the two alternatives that Ben mentioned here and then I think we're good to approve and merge. I don't see anything else outstanding that would block us from merging this.

jottofar: Done

enhancements/update/upgrades-blocking-on-ack.md
dhellmann: OK, it sounds like we can address my concerns via documentation. It would be good to include it as a risk in the design, though.

sdodson:
/lgtm
/approve
I don't think there's anything outstanding that prevents us from starting on implementation at this point.

PR #830, opened 2021-07-12
WIP Add enhancement for automated cluster backups

marun:
/cc @hexfusion @deads2k @smarterclayton

### DIFF
--- a/enhancements/disaster-recovery/automated-cluster-backups.md
+++ b/enhancements/disaster-recovery/automated-cluster-backups.md
@@ -4,10 +4,12 @@ authors:
   - "@marun"
 reviewers:
   - "@hexfusion"
+  - "@lilic"
   - "@deads2k"
   - "@smarterclayton"
 approvers:
   - "@hexfusion"
+  - "@lilic"
   - "@deads2k"
   - "@smarterclayton"
 creation-date: 2021-07-08
@@ -49,10 +51,13 @@ in a control plane node being taken offline in the process of attempting to
 follow the documented backup procedure.
 
 Finally, OpenShift 4.9 is intended to ship with etcd 3.5, and this will mean
-an upgrade for existing clusters from 3.4. Ensuring support for automated
-backups in the 4.8 edge required to upgrade to 4.9 will increase the
-likelyhood that customers will have cluster backups available in case of any
-trouble with the etcd upgrade.
+an upgrade for existing clusters from etcd 3.4. Since etcd will not support
+downgrade from 3.5 to 3.4, downgrade of a OpenShift cluster from 4.9 to 4.8
+will only be possible by restoring from backup. Ensuring support for
+automated backups in the 4.8 edge required to upgrade to 4.9 will increase
+the likelyhood that customers will have cluster backups available in case a
+downgrade is required.
+
 
 ### Goals
 
--- a/enhancements/disaster-recovery/automated-cluster-backups.md
+++ b/enhancements/disaster-recovery/automated-cluster-backups.md
@@ -1,0 +1,319 @@
+---
+title: automated-cluster-backups
+authors:
+  - "@marun"
+reviewers:
+  - "@hexfusion"
+  - "@deads2k"
+  - "@smarterclayton"
+approvers:
+  - "@hexfusion"
+  - "@deads2k"
+  - "@smarterclayton"
+creation-date: 2021-07-08
+last-updated: 2021-07-08
+status: implementable
+see-also:
+  - "https://docs.google.com/document/d/1LyZQMgLvgd81iQndcDcjAVN25ZEhCxVGMttoiRkSU6g/edit#"
+  - "https://docs.openshift.com/container-platform/4.7/backup_and_restore/backing-up-etcd.html"
+  - "https://rancher.com/docs/rancher/v2.x/en/backups/v2.5/configuration/backup-config/"
+  - "https://www.openshift.com/blog/ocp-disaster-recovery-part-1-how-to-create-automated-etcd-backup-in-openshift-4.x"
+---
+
+# Automated Cluster Backups
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Enable the configuration of automated cluster backups.
+
+## Motivation
+
+The current documented procedure for performing a backup of an OpenShift
+cluster is manually initiated. This reduces the likelyhood of a timely backup
+being available in the event of catastrophic cluster failure.
+
+The procedure also requires gaining a root shell on a control plane
+node. Shell access to OpenShift control plane nodes access is generally
+discouraged due to the potential for affecting the reliability of the
+node. Indeed, at least one customer reports having had operator error result
+in a control plane node being taken offline in the process of attempting to
+follow the documented backup procedure.
+
+Finally, OpenShift 4.9 is intended to ship with etcd 3.5, and this will mean
+an upgrade for existing clusters from 3.4. Ensuring support for automated
+backups in the 4.8 edge required to upgrade to 4.9 will increase the
+likelyhood that customers will have cluster backups available in case of any
+trouble with the etcd upgrade.
+
+### Goals
+
+1. One-time cluster backup can be initiated without requiring a root shell.
+1. Automated backups can be configured as a day 2 operation.
+1. Cluster backups are written to the host filesystem of control plane nodes.
+
+### Non-Goals
+
+1. Backup to cluster (i.e. PVC) or cloud storage (e.g. S3).
+1. Support day 1 configuration of automated backups.
+1. Automate cluster restoration.
+1. Rewrite the existing cluster backup script.
+
+## Proposal
+
+- Add a new API singleton supporting backup configuration
+  - Suggested resource name `backups.config.openshift.io`
+    - Separate backup configuration from the etcd operator in case
+      implementation migrates to a dedicated backup operator.
+  - Fields:
+    - spec.reason `string`
+      - Set to a value different from the most recent reason, will prompt a
+        one-time backup.
+      - Default: `""`
+    - spec.schedule `string`
+      - [Standard cron expression](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax)
+      - Results in backups scheduled repeatedly to successive control plane
+        nodes.
+      - Default: `""`
+    - spec.maxDurationWithoutBackup `duration`
+      - If a backup has not successfully completed in the configured
+        interval, a critical alert will be generated for this and every
+        subsequent interval. The alert will either indicate that backups are
+        not configured, or supply the most recent reason a backup failed.
+      - Setting to `0` will disable alerting.
+      - Default: `1h`
+    - spec.retentionCount `int`
+      - The maximum number of backups to retain.
+      - If the number of succesful backups (as determined by recorded
+        status) matches `retentionCount`, the oldest backup will removed
+        before a new backup is initiated.
+      - Default: 15
+
+- TODO Need a way to track backups once initiated
+  - New API type?
+  - Metadata to store
+    - openshift version
+    - node the backup was performed on
+      - Would also be where the backup was stored
+    - path the backup was written to
+    - time
+    - size
+    - success/failure
+      - failure reason
+  - Stored as spec or status?
+  - Remove backup api resources as part of removing backups
+
+
+- Implement a new controller that responds to the backup configuration
+  - The new controller will initially be shipped with the etcd operator for simplicity
+  - If `maxDurationWithoutBackup` is non-zero and is exceeded without a
+    successfull backup, create a critical alert. This is intended to ensure
+    that cluster admins are encouraged to configure automated backups if only
+    to disable the alerting.
+  - The cost of taking automated backups should be spread over all the
+    control plane nodes to minimize the IO impact on any one node.
+    - Spreading backups across available nodes has the added benefit of
+      maximizing the chances of retaining a viable backup should one or more
+      control plane nodes be lost.
+
+- If a backup is configured to occur due to `backupReason` or `schedule`:
+  - Given the set of control plane nodes
+  - Given the set of the successful backups recorded in status
+  - Only a single backup pod should be running at a given time. If a backup
+    pod is still running:
+    - If the pod has been running longer than a maximum period (TODO How
+      long?), terminate it and record the backup as having failed.
+    - If the pod has been running for less than the maximum period, defer the
+      scheduled backup for (TODO How long)
+  - If the number of successful backups matches `retentionCount`
+    - Ensure removal of the oldest backup from the node that it was written to
+    - Remove record of removed backup from the api?
+    - Log removal of the backup
+  - Find the set of candidate nodes that have fewer backups than other nodes
+    - Could be 1 node (e.g. 2 nodes have 1 backup and 1 node has zero)
+    - Could be a subset of nodes (e.g. 1 node has 1 backup and 2 nodes have
+      zero)
+    - Could be all nodes (i.e. no successful backups have been taken)
+  - Schedule a backup pod to a random member of the set of candidate nodes
+    - A backup pod should invoke the `cluster-backup.sh` script on the host
+      via chroot and write the backup data to the host.
+    - A backup pod should first check that the available disk space on the
+      node it is running on is a multiple of the size of the node's
+      /var/lib/etcd path. A node needs a minimum amount of available storage
+      to operate reliably and automated backup must make every effort not to
+      exhaust a node's available storage.
+  - When a backup pod terminates, record the succcess or failure to a backup
+    resource
+
+### User Stories
+
+- I want to be alerted after installation or upgrade to the availability of
+  an automated backup capability and be given an indication of the importance
+  of configuring it in a timely manner.
+
+- I want to initiate a cluster backup without requiring a root shell on a
+  control plane node so as to minimize the risk involved.
+
+- I want to schedule recurring cluster backups so that I have recent cluster
+  state to recover with in the event of disaster (i.e. losing 2 control plane
+  nodes).
+
+- I want to have failure to take cluster backups for more than a configurable
+  period to be reported to me via critical alerts.
+
+- I want to minimize the operational impact of taking regular backups.
+
+### Risks and Mitigations
+
+Spreading the spread the cost of automated backups over all control plane
+nodes may not be sufficient to avoid negatively impacting cluster performance
+on IO-constrained hosts. Ideally it would be possible to limit the IO
+overhead of taking a backup via cgroup limits, but it's not clear that this
+would be achievable without the granularity provided by cgroups v2 (not
+slated for GA until RHEL9).
+
+Automated backups relying on host storage could fill up the disks of control
+plane nodes, rendering them inoperable. This is intended to be mitigated by a
+check for available disk space that ensures backups are only initiated on
+nodes that have a multiple of the space required to backup the node's
+/var/lib/etcd path.
+
+Relying on node storage may hamper recovery if all control plane nodes are
+lost. It might make sense to document the requirement to periodically ship
+backup data off the cluster. Given that the current restore procedure
+requires at least one control plane node, though, a cluster is only
+recoverable from backup if at least one of the backup-containing control
+plane nodes is available.
+
+## Design Details
+
+### Open Questions
+
+- Under what circumstances should backup pods that do not terminate
+  successfully or do not terminate within a reasonable period of time be
+  recreated?
+
+- If the node targeted for backup does not have sufficient space, should
+  there be an attempt to take a backup on a different node? This question is
+  complicated by spreading backups across available nodes since backing up to
+  a different node may require removing a backup that is not the oldest.
+
+### Test Plan
+
+- Comprehensive unit testing of
+  - Changes in backup configuration
+  - Scheduling logic
+  - Backup retention
+  - Pre-backup check for disk space available
+
+- Update DR e2e testing to initiate one-time backups via the controller
+
+- Add new e2e golden-path testing of scheduled backups
+
+### Graduation Criteria
+
+**Note:** *Section not required until targeted at a release.*
+
+Define graduation milestones.
+
+These may be defined in terms of API maturity, or as something else. Initial proposal
+should keep this high-level with a focus on what signals will be looked at to
+determine graduation.
+
+Consider the following in developing the graduation criteria for this
+enhancement:
+
+- Maturity levels
+  - [`alpha`, `beta`, `stable` in upstream Kubernetes][maturity-levels]
+  - `Dev Preview`, `Tech Preview`, `GA` in OpenShift
+- [Deprecation policy][deprecation-policy]
+
+Clearly define what graduation means by either linking to the [API doc definition](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning),
+or by redefining what graduation means.
+
+In general, we try to use the same stages (alpha, beta, GA), regardless how the functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+**Examples**: These are generalized examples to consider, in addition
+to the aforementioned [maturity levels][maturity-levels].
+
+#### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+- Enumerate service level indicators (SLIs), expose SLIs as metrics
+- Write symptoms-based alerts for the component(s)
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+
+**For non-optional features moving to GA, the graduation criteria must include
+end to end tests.**
+
+#### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
+
+### Upgrade / Downgrade Strategy
+
+Given that this feature is strictly additive and is not dependent or depended
+upon by other components, no special consideration need be given to handling
+upgrades and downgrades.
+
+### Version Skew Strategy
+
+Given that this feature is strictly additive and is not dependent or depended
+upon by other components, no special consideration need be given to handling
+version skew.
+
+## Implementation History
+
+Major milestones in the life cycle of a proposal should be tracked in `Implementation
+History`.
+
+## Drawbacks
+
+The idea is to find the best form of an argument why this enhancement should _not_ be implemented.
+
+## Alternatives
+
+Q: The built-in
+  [CronJob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/)
+  feature would appear to be duplicated by this feature. Why not just
+  implement on top of `CronJob`?
+
+A: It is desirable to spread backups across control plane nodes to minimize
+   the io impact on any one node and to maximize the potential for a given
+   control plane node to have a recent backup to restore from. It does not
+   appear to be possible to layer the scheduling logic required to achieve
+   these capabilities on top of `CronJob` scheduling.
+
+Q: Why not schedule backups to nodes with the built-in
+   [Job](https://kubernetes.io/docs/concepts/workloads/controllers/job/)
+   feature instead of creating pods directly?
+
+A: `Job` has the ability to reschedule a pod to another node if that node
+   becomes unavailable (e.g. due to failure or reboot). This is not desirable
+   in the context of an attempt to backup a cluster since nodes are not
+   fungeable. A node targeted for backup is chosen to spread both IO impact
+   and backup availability across available nodes, and `Job` has no way of
+   taking these considerations into account.
--- a/enhancements/disaster-recovery/automated-cluster-backups.md
+++ b/enhancements/disaster-recovery/automated-cluster-backups.md
@@ -35,7 +35,8 @@ see-also:
 
 ## Summary
 
-Enable the configuration of automated cluster backups.
+Enable the configuration of automated backup of etcd and the static pod
+content that underpin an OpenShift 4.x self-hosted cluster.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


dhellmann:
If this is a priority for the current quarter or release you can use `/priority important-soon` to indicate that.

hexfusion (commented):
thanks @marun added few notes for first pass

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,319 @@
...
+      - Results in backups scheduled repeatedly to successive control plane
+        nodes.
+      - Default: `""`
+    - spec.maxDurationWithoutBackup `duration`
+      - If a backup has not successfully completed in the configured
+        interval, a critical alert will be generated for this and every
+        subsequent interval. The alert will either indicate that backups are
+        not configured, or supply the most recent reason a backup failed.
+      - Setting to `0` will disable alerting.
+      - Default: `1h`
```
romfreiman, line 92: 1h? why do u need such high frequency?

romfreiman: The major drawback is in case something is corrupted, or a user caused some issue and will notice only the next day , you'll continuously  override all old copies.
Doesn't 1d make more sense?

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,319 @@
...
+1. Automate cluster restoration.
+1. Rewrite the existing cluster backup script.
+
+## Proposal
+
+- Add a new API singleton supporting backup configuration
+  - Suggested resource name `backups.config.openshift.io`
+    - Separate backup configuration from the etcd operator in case
+      implementation migrates to a dedicated backup operator.
+  - Fields:
```
romfreiman, line 76: I would add a way to specify dedicated partition, and I think it should be the recommended approach. Otherwise, there is always a risk that some back will kill the node (or at least fill the fs).

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,319 @@
...
+    - spec.reason `string`
+      - Set to a value different from the most recent reason, will prompt a
+        one-time backup.
+      - Default: `""`
+    - spec.schedule `string`
+      - [Standard cron expression](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax)
+      - Results in backups scheduled repeatedly to successive control plane
+        nodes.
+      - Default: `""`
+    - spec.maxDurationWithoutBackup `duration`
```
romfreiman, line 86: will this alert also address that backup didnt happen due to low disk capacity?

marun: It would happen due to a backup not occurring for _any_ reason.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: Adding another partition for this use case is not a near-term option, hence the emphasis on not performing a backup if there is even a minute chance of filling the disk.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: - We're not intending to enable backups by default. The user will be prompted (by the 'backup not completed within x amount of time alert) to configure backups to a configuration appropriate to their needs when the cluster upgrades to a version that supports it.
- The number we set as a default is pretty arbitrary at this point - it's a placeholder. I would expect to solicit input from actual users rather than any of us assuming we know best. 
- A backup that is more than a short interval old would be problematic if the cluster used cluster storage. Most csi plugins use the api as their source of truth. In the event of _any_ storage changes occurring since the backup was taken, it is likely that the api state wouldn't match the storage state and reconciliation could be difficult or even impossible.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,319 @@
...
+
+
+- Implement a new controller that responds to the backup configuration
+  - The new controller will initially be shipped with the etcd operator for simplicity
+  - If `maxDurationWithoutBackup` is non-zero and is exceeded without a
+    successfull backup, create a critical alert. This is intended to ensure
+    that cluster admins are encouraged to configure automated backups if only
+    to disable the alerting.
+  - The cost of taking automated backups should be spread over all the
+    control plane nodes to minimize the IO impact on any one node.
```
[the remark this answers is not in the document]
marun: Does it follow that IO cost should be removed as a consideration?

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,319 @@
...
+        subsequent interval. The alert will either indicate that backups are
+        not configured, or supply the most recent reason a backup failed.
+      - Setting to `0` will disable alerting.
+      - Default: `1h`
+    - spec.retentionCount `int`
+      - The maximum number of backups to retain.
+      - If the number of succesful backups (as determined by recorded
+        status) matches `retentionCount`, the oldest backup will removed
+        before a new backup is initiated.
+      - Default: 15
```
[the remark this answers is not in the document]
marun: It's a completely arbitrary number. In the pantheon of things to worry about in this enhancement the default values are bottom of the list and it may make more sense to worry about them in the context of implementation PRs.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,319 @@
...
+        nodes.
+      - Default: `""`
+    - spec.maxDurationWithoutBackup `duration`
+      - If a backup has not successfully completed in the configured
+        interval, a critical alert will be generated for this and every
+        subsequent interval. The alert will either indicate that backups are
+        not configured, or supply the most recent reason a backup failed.
+      - Setting to `0` will disable alerting.
+      - Default: `1h`
+    - spec.retentionCount `int`
```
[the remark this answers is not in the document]
marun: No, total. As per the proposal to spread backups across nodes, the total number of backups figures into retention (remove the oldest backup regardless of the node) and scheduling (schedule a backup on one of the nodes with the lease number of retained backups).

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,319 @@
...
+    - success/failure
+      - failure reason
+  - Stored as spec or status?
+  - Remove backup api resources as part of removing backups
+
+
+- Implement a new controller that responds to the backup configuration
+  - The new controller will initially be shipped with the etcd operator for simplicity
+  - If `maxDurationWithoutBackup` is non-zero and is exceeded without a
+    successfull backup, create a critical alert. This is intended to ensure
```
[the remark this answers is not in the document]
marun: Are you asking for the backup controller to set the controller degraded as a way of preventing an upgrade unless either of the following conditions are met?

 - a backup exists that is within the window defined by `maxDurationWithoutBackup`
 - backups have been disabled

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,319 @@
+---
+title: automated-cluster-backups
+authors:
+  - "@marun"
+reviewers:
+  - "@hexfusion"
+  - "@deads2k"
+  - "@smarterclayton"
```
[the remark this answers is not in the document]
marun: Done

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,319 @@
...
+discouraged due to the potential for affecting the reliability of the
+node. Indeed, at least one customer reports having had operator error result
+in a control plane node being taken offline in the process of attempting to
+follow the documented backup procedure.
+
+Finally, OpenShift 4.9 is intended to ship with etcd 3.5, and this will mean
+an upgrade for existing clusters from 3.4. Ensuring support for automated
+backups in the 4.8 edge required to upgrade to 4.9 will increase the
+likelyhood that customers will have cluster backups available in case of any
+trouble with the etcd upgrade.
```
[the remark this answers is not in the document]
marun: Done

enhancements/disaster-recovery/automated-cluster-backups.md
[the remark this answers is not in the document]
hexfusion: yes that makes sense I believe.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Enable the configuration of automated cluster backups.
```
deads2k, line 38: needs specificity.  velero promises the same.  This is specifically about backing up etcd state and NOT about backing PVs or node state, correct?

enhancements/disaster-recovery/automated-cluster-backups.md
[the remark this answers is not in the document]
hexfusion: well we want to validate 
- is pulling snapshot across the wire less disruptive to leader than having to serve the snapshot locally with a lcoal I/O hit to disk.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+### Goals
+
+1. One-time cluster backup can be initiated without requiring a root shell.
+1. Automated backups can be configured as a day 2 operation.
+1. Cluster backups are written to the host filesystem of control plane nodes.
+
+### Non-Goals
+
+1. Backup to cluster (i.e. PVC) or cloud storage (e.g. S3).
+1. Support day 1 configuration of automated backups.
```
deads2k, line 71: seems like an unusual exclusion

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+1. Automate cluster restoration.
+1. Rewrite the existing cluster backup script.
+
+## Proposal
+
+- Add a new API singleton supporting backup configuration
+  - Suggested resource name `backups.config.openshift.io`
+    - Separate backup configuration from the etcd operator in case
+      implementation migrates to a dedicated backup operator.
```
deads2k, line 80: I read top to bottom, but this reasoning seems a little weak to me.  If a later backup operator is created, I would expect it to find a way to honor or migrate from existing, individual  backup configurations that were focused on particular operands.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+  - Fields:
+    - spec.reason `string`
+      - Set to a value different from the most recent reason, will prompt a
+        one-time backup.
+      - Default: `""`
+    - spec.schedule `string`
+      - [Standard cron expression](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax)
+      - Results in backups scheduled repeatedly to successive control plane
```
deads2k, line 88: This is not clear to me.  Can you provide an example of "every two hours" for me?

Also, why a full cron syntax instead of a simple duration?

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+        nodes.
+      - Default: `""`
```
deads2k, line 90: tell me what a default means please.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+    - spec.maxDurationWithoutBackup `duration`
+      - If a backup has not successfully completed in the configured
```
deads2k, line 92: does this mean that if a backup starts at 1am and doesn't finish by 2am, the alert will sound.  If I have a 3h interval for starting these, does it mean that if I have a backup at 5am that takes 10minutes, will I get an alert at 6:10am?

The fact that I have to ask means the doc needs update.

enhancements/disaster-recovery/automated-cluster-backups.md
deads2k: Depending on what you mean, it seems that this default may have to auto-adjust

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+        interval, a critical alert will be generated for this and every
+        subsequent interval. The alert will either indicate that backups are
```
deads2k, line 94: If backups are not configured, I would expect a different alert.

enhancements/disaster-recovery/automated-cluster-backups.md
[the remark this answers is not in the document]
deads2k: This doesn't seem like enough granularity for retention.  I suggest a retention stanza with this as an element. Retention is a thing people often slice on 
1. absolute number
2. size
3. time
4. weighted decimation: keep every hour for 24 hours.  keep every day for a week.  keep every week for a month.  keep every month for a year.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+      - Setting to `0` will disable alerting.
+      - Default: `1h`
+    - spec.retentionCount `int`
+      - The maximum number of backups to retain.
+      - If the number of succesful backups (as determined by recorded
+        status) matches `retentionCount`, the oldest backup will removed
+        before a new backup is initiated.
+      - Default: 15
+
+- TODO Need a way to track backups once initiated
```
deads2k, line 105: Yes you do.  I suggest a separate CR type to represent each backup you expect to have.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+  - New API type?
+  - Metadata to store
```
deads2k, line 107: some of this metadata should exist on-disk as well.

enhancements/disaster-recovery/automated-cluster-backups.md
deads2k: > Yes you do. I suggest a separate CR type to represent each backup you expect to have.

Current me (no guarantee of future me) sees some benefit to writing this into a configmap and live copying that onto disk so that you have this data when etcd is non-functional and you have to make decisions.

Current me is also debating whether we would just serialize this separate CR down instead.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+    - time
+    - size
+    - success/failure
+      - failure reason
+  - Stored as spec or status?
+  - Remove backup api resources as part of removing backups
+
+
+- Implement a new controller that responds to the backup configuration
+  - The new controller will initially be shipped with the etcd operator for simplicity
```
deads2k, line 121: this seems reasonable to me.

enhancements/disaster-recovery/automated-cluster-backups.md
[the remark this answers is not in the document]
deads2k: I suggest easing in with a warning alert first.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+1. Cluster backups are written to the host filesystem of control plane nodes.
+
+### Non-Goals
+
+1. Backup to cluster (i.e. PVC) or cloud storage (e.g. S3).
+1. Support day 1 configuration of automated backups.
+1. Automate cluster restoration.
+1. Rewrite the existing cluster backup script.
+
+## Proposal
```
deads2k, line 75: Others may know this, but I do not.  How do I take an etcd backup?  I'd like either a link to where we already describe how to do this in *our* docs or enhancements or two 10-step lists of how to mechanically do this by hand on a master and another list for a follower.

enhancements/disaster-recovery/automated-cluster-backups.md
deads2k: > Also, why a full cron syntax instead of a simple duration?

ok, I see why.  I don't want every 24 hours.  I want everyday at 2am.  Please describe the reasoning in the doc for future me?

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+  - Only a single backup pod should be running at a given time. If a backup
+    pod is still running:
+    - If the pod has been running longer than a maximum period (TODO How
+      long?), terminate it and record the backup as having failed.
+    - If the pod has been running for less than the maximum period, defer the
+      scheduled backup for (TODO How long)
+  - If the number of successful backups matches `retentionCount`
+    - Ensure removal of the oldest backup from the node that it was written to
+    - Remove record of removed backup from the api?
+    - Log removal of the backup
```
deads2k, line 144: log before you remove the record to make an indempotent command.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+  - Find the set of candidate nodes that have fewer backups than other nodes
+    - Could be 1 node (e.g. 2 nodes have 1 backup and 1 node has zero)
+    - Could be a subset of nodes (e.g. 1 node has 1 backup and 2 nodes have
+      zero)
+    - Could be all nodes (i.e. no successful backups have been taken)
+  - Schedule a backup pod to a random member of the set of candidate nodes
+    - A backup pod should invoke the `cluster-backup.sh` script on the host
+      via chroot and write the backup data to the host.
+    - A backup pod should first check that the available disk space on the
```
deads2k, line 153: backup pod or backup script?  This sounds like protection a script we provide should have.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+## Design Details
+
+### Open Questions
+
+- Under what circumstances should backup pods that do not terminate
+  successfully or do not terminate within a reasonable period of time be
+  recreated?
+
+- If the node targeted for backup does not have sufficient space, should
+  there be an attempt to take a backup on a different node? This question is
```
deads2k, line 210: Yes, this should be attempted.

deads2k: Why is this failure "special"?  Why wouldn't all failures have failover behavior like this?

enhancements/disaster-recovery/automated-cluster-backups.md
deads2k: I'd like to see a section outlining the directory structure on the node and the content contained in it for backups.

deads2k: I'd like to see enough metadata available on a master to build an automated, "restore the right backup" command on a master.  This should contain enough detail that we're sure we write enough data from our own controller and be detailed enough that a customer could provide congruent metadata for his own backup process if he wished to restore using our script.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+- I want to minimize the operational impact of taking regular backups.
+
+### Risks and Mitigations
+
+Spreading the spread the cost of automated backups over all control plane
+nodes may not be sufficient to avoid negatively impacting cluster performance
+on IO-constrained hosts. Ideally it would be possible to limit the IO
+overhead of taking a backup via cgroup limits, but it's not clear that this
+would be achievable without the granularity provided by cgroups v2 (not
+slated for GA until RHEL9).
```
jeremyeder, line 186: This is probably my biggest concern, having backup activity adversely affect the performance of the cluster.  One way to indirectly rate limit backups without cgroups v2 might be to constrain the backup's CPU.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+
+Automated backups relying on host storage could fill up the disks of control
+plane nodes, rendering them inoperable. This is intended to be mitigated by a
+check for available disk space that ensures backups are only initiated on
+nodes that have a multiple of the space required to backup the node's
+/var/lib/etcd path.
```
jeremyeder, line 192: What other pre-flight checks might we make?  Should it look for certain alerts firing?  What if etcdmemberdown is firing?

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+
+Relying on node storage may hamper recovery if all control plane nodes are
+lost. It might make sense to document the requirement to periodically ship
+backup data off the cluster. Given that the current restore procedure
```
jeremyeder, line 196: Should we make an off-cluster schedule shipper (e.g. S3 or similar...) part of this enhancement?

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+- Write symptoms-based alerts for the component(s)
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
```
jeremyeder, line 271: In this particular instance, should load testing move to the dev preview section?

jeremyeder: Also, the metrics and SLIs...should be considered during dev preview phase.

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,324 @@
...
+The procedure also requires gaining a root shell on a control plane
+node. Shell access to OpenShift control plane nodes access is generally
+discouraged due to the potential for affecting the reliability of the
+node. Indeed, at least one customer reports having had operator error result
+in a control plane node being taken offline in the process of attempting to
+follow the documented backup procedure.
+
+Finally, OpenShift 4.9 is intended to ship with etcd 3.5, and this will mean
+an upgrade for existing clusters from etcd 3.4. Since etcd will not support
+downgrade from 3.5 to 3.4, downgrade of a OpenShift cluster from 4.9 to 4.8
```
deads2k, line 55: If your desire in 4.9 is to irreversibly switch to etcd 3.5, then this feature needs to save a backup *before* upgrading to 4.9, which suggests you would need this feature in 4.8.  

1. Is the timing listed here stale (and you mean 4.10)?
2. Is this supposed to be backportable?
3. Is 4.9 going to have two stages of upgrade that involve "updated to 3.4+backup, then updated to 3.5"?
4. Is 4.8 going to have a manual, one-off backup and then status reporting?

marun: @hexfusion can confirm, but https://github.com/openshift/enhancements/pull/830#discussion_r671425446 suggests a desire to block upgrades to 4.9 unless a recent backup exists.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: I don't think this is a 'dev preview' type of feature. @Anandnatraj can confirm, but I believe the urgency of this feature suggests providing it GA in whatever release it is first included in.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: The initial scoping is explicitly targeting an MVP that is generally useful for all clusters, not just those that have the ability to utilize off-cluster storage.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: @hexfusion Are there any conditions other than quorum loss and lack of storage that should preclude taking a backup?

enhancements/disaster-recovery/automated-cluster-backups.md
marun: @hexfusion How effective would it be limit a backup pod's CPU as suggested? I was under the impression that the constraint was almost entirely IO.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: This question is highlighting the potential for conflict between how many backups a customer wants to retain and how to ensure those backups are spread across nodes so that the loss of any one node or any one backup does not compromise the ability to recover a cluster. 

@hexfusion Do you see a tension between what we think might be reasonable (e.g. 3 backups per node) vs what customers uninformed about the consequences of time-travel might want?

enhancements/disaster-recovery/automated-cluster-backups.md
marun: Done

enhancements/disaster-recovery/automated-cluster-backups.md
marun: Updated to reference the script.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: @deads2k Would you please elaborate as to why a performance-impacting action should change without human input?

enhancements/disaster-recovery/automated-cluster-backups.md
marun: Thanks to 4.x being self-hosted, it's not just about etcd backups but also content from /etc/kubernetes, but to your point I'll clarify that.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: The intention here is MVP, and as per @Anandnatraj adding supporting for configuring backups via the installer was considered out-of-scope.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: I'm also attempting to impose a degree of logical separation between the etcd operator and disaster recovery, since the latter touches a lot more things than etcd. Especially when we start talking about automating recovery, etcd is only one piece of the puzzle. Does this additional context suggest to you preemptively putting the configuration in a new CRD, or would you prefer that the configuration be incorporated in to the CEO configuration?

enhancements/disaster-recovery/automated-cluster-backups.md
marun: Done

enhancements/disaster-recovery/automated-cluster-backups.md
marun: "" -> disabled 

Done.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: What's the value of serializing into a configmap? I'm not sure why a CR wouldn't be a sufficient source-of-truth.

enhancements/disaster-recovery/automated-cluster-backups.md
hexfusion: I believe the operator needs to have an understanding of the expected/current cluster id for etcd. So that we can ensure we are taking backups of that cluster explicitly. Also on restore, we would want to validate backup clusterID is. This might be something we want to stamp on every static pod resource as metadata in addition to revision? 

There are other backup level metadata we would like to store for validation purposes
- etcd db checksum
- checksum on static pod resources
- ocp version of backup

tl;dr we need to ensure validity of backups to protect support from dirty/invalid state as the root cause for restore failure. This also might be a security consideration.

enhancements/disaster-recovery/automated-cluster-backups.md
hexfusion: the perf footprint is:
- CPU/Memory/Disk for leader to serve a snapshot
- wire cost for state travel 
- destination storage medium I/O hit

enhancements/disaster-recovery/automated-cluster-backups.md
```diff
@@ -0,0 +1,329 @@
...
+
+### Non-Goals
+
+1. Backup to cluster (i.e. PVC) or cloud storage (e.g. S3).
+1. Support day 1 configuration of automated backups.
+1. Automate cluster restoration.
+1. Rewrite the existing cluster backup script.
+
+## Proposal
+
```
hexfusion, line 77: abstract thought experiment:

would it be possible to have a controller schedule a static pod that mounts a ramdisk. Then take the snapshot to that mount? This way we could use the current static-pod controller process to handle pruning as well. RAM seems easier to come by than storage in most cases. So CU could have the option if they had the RSS resources to mitigate disk cost by snapshotting to ramdisk followed by a slow burn to hostDisk.

Once snapshot hits ramdisk we could use something like rsync --bwlimit  to limit socket I/O bandwidth from ramdisk to hostDisk. Once complete mark success. Its more complex but reusing static-pod controller might be interesting on its own?

enhancements/disaster-recovery/automated-cluster-backups.md
hexfusion: as a snapshot is a quorate request we don't need to think about quorum, no quorum means the client request fails.

enhancements/disaster-recovery/automated-cluster-backups.md
marun: @hexfusion Is it necessary to consider every resource consumed by backup, or is there a subset of them that can be considered 'limiting factors' whose impact it is more important to minimize?

enhancements/disaster-recovery/automated-cluster-backups.md
marun: aka the only pre-flight we need is for available storage?

enhancements/disaster-recovery/automated-cluster-backups.md
marun: How likely is it that the control plane nodes on any given cluster will have sufficient RAM to spare to buffer backups in a ramdisk?

hexfusion: The RAM would need to be dedicated during cluster provisioning. So it's probably not practical.

enhancements/disaster-recovery/automated-cluster-backups.md
hexfusion: I would expect no one to want time travel if they understood the implications. in which case the goal should be validation of the etcd state file on disk as a valid restore medium. backup and time travel should be break glass

enhancements/disaster-recovery/automated-cluster-backups.md
hexfusion: well, we would want to validate the cost of not taking backup on the leader node. the etcd backend is likely to grow soon so these are longer-term considerations. For example, I can see the etcd db being 2x what it is today (16GB) in the not-so-distant future. In which case another check would be isLeader().

enhancements/disaster-recovery/automated-cluster-backups.md
lilic: Yes, I agree we might have to backport something to 4.8 to ensure a backup is taken. From my small experiments, it wasn't possible to easily downgrade due to the schema changes.

marun:
/close

This has been pushed to 4.10. Closing for now, will leave the branch up for future.

PR #831, opened 2021-07-12
aos-sno-pair enhancement

mshitrit:
Signed-off-by: Michael Shitrit <mshitrit@redhat.com>


dhellmann:
If this is a priority for the current quarter or release, you can use `/priority important-now` to indicate that.

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,142 @@
...
+Where this is not possible, there are existing capabilities that we can leverage by enabling the _Corosync two_node, last_man_standing, and wait_for_all_ options (See: man votequorum) and configuring a fencing delay in Pacemaker.
+
+In combination, these options provide the following behaviour:
+- Both sides retain quorum in the event of a network split.
+- Service recovery is normally predicated on successful fencing.
+- When the fenced node reboots, it will not be able to obtain quorum until it has seen its peer (preventing an infinite fencing cycle).
+- To avoid both sides fencing each other at the same time, fencing is typically configured to occur after a delay on one of the nodes. That node will perform recovery in the event of a true peer failure, but will always lose in the response to a pure network failure.
+
+### Risks and Mitigations
+- Etcd, and by inference k8s and OpenShift clusters, cannot function without quorum.
```
sttts, line 99: how does this matter if the design explicitly says that etcd and cluster state in general is not shared, but totally independent?

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+This enhancement describes setting up a two nodes cluster with a reduced footprint which will serve edge cases.
+
+## Motivation
+
+Some companies have a need for a highly available container management solution that fits within a reduced footprint.
+- The hardware savings are significant for customers deploying many remote sites (kiosks, branch offices, restaurant chains, etc), most notably for edge computing and RAN specifically.
+- The physical constraints of some deployments prevents more than two nodes (planes, submarines, satellites, and also RAN).
```
cgwalters, line 32: The foundation of Kubernetes is etcd, which is built on extensive proven research into the Raft consensus protocol, and has lots of domain knowledge, tooling etc.  The fact that all cluster types today use etcd means we have symmetry across many cloud and bare metal cases.

I can certainly see that going from 2 to 3 nodes may increase cost for some of these users.  But it's also an enormous cost to OpenShift to add a new type of "HA" cluster and I don't think we can gloss over that at all.   I understand pacemaker etc. are also well understood and deployed in many scenarios, but those are (AFAIK) today mostly bare metal.  So we will lose out on the synergy we get today from making cloud/metal strongly aligned.

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+- Some locations will not have reliable network connections or limited bandwidth (once again submarines, satellites, and disaster areas such as after hurricanes or floods)
+  
+### Goals
+
+Provide the ability for all-in-one OpenShift clusters to operate as a **predefined** pair in an **active-active** configuration, detect when its peer has died and **automatically** take over its workloads after ensuring it is safe to do so.
```
dhellmann, line 37: Is this the goal? Based on what has been said elsewhere, the clusters aren't actually aware of each other, it's the applications running on them that are paired. Did I misunderstand those other comments?

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+- **Replication of application data** between appliances.
+- DR for services in clusters with **more than a single node.**
+- DR for services running on **two or more all-in-one clusters.**
+- Enable services uptime during a **remote cluster upgrade process.**
+
+## Proposal
+
+Assuming workloads that require some form of coordination between the two nodes (either at-most-one semantics or simply to reduce machine load during normal operation):
+
+- Pacemaker and Corosync from the RHEL-HA cluster stack are deployed [in Pods](https://github.com/beekhof/installer/blob/aio/pod-cluster.yaml.in) on both OCP clusters, and they are configured to form a two node cluster.
```
dhellmann, line 51: That's a 2 node pacemaker cluster, not a 2 node OpenShift cluster, right?

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+- The Pacemaker cluster is configured so that on peer failure the cluster can reliably ensure that the workload on an unhealthy peer is stopped before starting the workload on the healthy peer. Configuration [flexibility](https://drive.google.com/file/d/1aw_78490rTQGsxEFcZytMe8lasZErLMm/view?usp=sharing) is provided here to allow for the anticipated variety of customer requirements.
```
dhellmann, line 52: Google Drive isn't public. Please export all images into a format that can be added to this repository and linked or embedded in the doc.

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+
+Care must be taken to ensure that paths are truly independent and avoid single points of failure (SPoF). Crossover cables are often the simplest way of meeting this requirement and are therefore highly encouraged.
+
+#### Fencing
+In the event of a peer failure, any **workloads requiring singleton semantics** (such as StatefulSets or kubevirt VMs) require the existing node to be put into a safe state (stopped or otherwise known to be not running the workloads) before they can be started on the surviving peer.  The traditional term for this is “**fencing**” and it is an **optional** part of RHEL-HA clusters.
+
+Customer environments without any singleton workloads may not require fencing, but need to be comfortable with the possibility of workloads being active on both peers at the same time.
+
+For customers that do have singleton workloads, the RHEL-HA stack supports a wide variety of BMCs including those conforming to the IPMI and RedFish standards, as well as SSH and libvirt for testing.
```
dhellmann, line 80: What API will be provided for configuring access to those BMCs?

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+  - Reasons mentioned in favor:
+    - It ensures the secondary node works when you need it. Not only it is working, but the whole list of dependencies it carries with it (DNS...)
+    - Performance.
+    - Cost. The secondary node is not just sitting there, wasting resources.
+    - LESS maintenance procedures. You don't need to failover/failback on upgrades, for example.
+- Do we really need automated failover, or is manual confirmation sufficient?
+- How much overlap is there between the goals as laid out here and Telco RAN requirements?
+
+
+### Test Plan
```
dhellmann, line 116: It would be good to have some detail about testing.

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Graduation Criteria
```
dhellmann, line 118: Are we going to release this in a preview state first or go straight to GA?

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### Removing a deprecated feature
+
+### Upgrade / Downgrade Strategy
+
+### Version Skew Strategy
+
+## Implementation History
+
+## Drawbacks
```
dhellmann, line 134: This approach introduces a non-kubernetes HA approach to an OpenShift environment, requiring users to understand both kubernetes and RHEL-HA.

enhancements/single-node/aos-sno-pair.md
dhellmann: Everywhere else that I've seen this discussed we talked about it being a thing that sits on top of OpenShift and provides HA for an app. It's aware of the OpenShift API, and manages a Deployment, but it doesn't provide HA for OpenShift itself and doesn't rely on etcd quorum.

However, some parts of this document make me think that either I'm wrong and this is about 2-node etcd HA, or we did originally explore 2-node etcd HA using this approach but abandoned it. @mshitrit could you scan through for references to etcd and make sure they're relevant and not misleading about what is being made highly available?

enhancements/single-node/aos-sno-pair.md
mshitrit: That is two Single Node OpenShift (SNO) clusters, each SNO is independent and unaware of the other one.
Pacemaker is involved in  maintaining the High Availability stack that is on top.

enhancements/single-node/aos-sno-pair.md
mshitrit: I think that you understood correctly the part that indeed each SNO is independent.
However both SNO + HA Stack compromise the all-in-one OpenShift cluster (AOS) which is another entity.
The above applies to AOS entity and not to the SNOs.

enhancements/single-node/aos-sno-pair.md
mshitrit: >but it doesn't provide HA for OpenShift itself 

I'm not completely sure what you mean by `OpenShift itself`.
If you mean that each Single Node OpenShift (SNO) that compromise the  all-in-one OpenShift cluster (AOS) isn't HA by itself you are correct.
However the AOS as an entity is HA.

dhellmann: > > but it doesn't provide HA for OpenShift itself
> 
> I'm not completely sure what you mean by `OpenShift itself`.
> If you mean that each Single Node OpenShift (SNO) that compromise the all-in-one OpenShift cluster (AOS) isn't HA by itself you are correct.
> However the AOS as an entity is HA.

Maybe the confusion is coming from the way we're using terminology around HA and the name we're giving this enhancement. This isn't about making the platform itself highly available, it's about making the application highly available.

Normally when we want an OpenShift cluster to run in a HA configuration, we say it needs at least 3 nodes. That provides for multiple replicas of etcd, the API server, etc. and makes the _platform_ highly available.

This proposal does not do that. It is about ensuring that an _application_ running on top of 2 completely distinct OpenShift instances ends up in a highly available configuration. However, those OpenShift instances are not highly available, either individually or as a pair. Someone trying to configure the instances via their APIs has to go directly to one and then the other to change both. If the etcd database or API server fails on one, access won't be redirected to the other.

I think we should call this enhancement something that more accurately describes the layer where the implementation is operating, and not "single-node pair". Using the word application in the name would help.

This design doc also talks about etcd in a way that makes the distinction between platform and application HA confusing.

For example, line 99 mentions losing etcd quorum as a risk. But etcd won't be HA, we will have 2 separate etcd instances and they don't need to have a combined quorum. 

The alternative on line 139 also mentions etcd. I don't know what "etcd learners" are or how that would relate to running HA between 2 deployments. Maybe in that case we just need a little more detail?

enhancements/single-node/aos-sno-pair.md
dhellmann: I'm not sure what you mean by "compromise".

enhancements/single-node/aos-sno-pair.md
dhellmann: Saying that two single node instances of OpenShift form a two node OpenShift cluster is confusing. The nodes are not part of the same OpenShift cluster.

enhancements/single-node/aos-sno-pair.md
mshitrit: Sorry typo - I meant compose

dhellmann: Ah, ok, that makes more sense. I think my comments in the other thread about platform vs. application apply here, too. Let's carry on that conversation in the other thread so it's all in one place.

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+#### Quorum handling
+_The "two node cluster" is a use case that requires special consideration.  With a standard two node cluster, each node with a single vote, there are 2 votes in the cluster. Using the simple majority calculation (50% of the votes + 1)  to calculate quorum, the quorum would be 2.  This means that the both nodes would always have to be alive for the cluster to be quorate and operate._
+<br>Source: man votequorum
+
+For environments in which customers will allow it, Corosync also ships a lightweight quorum arbitrator that can be run locally or in the cloud on any non-cluster node.  In such cases, no special quorum handling is required and careful placement of the arbitrator also ensures that the surviving peer is reachable by its intended clients.
+
+Where this is not possible, there are existing capabilities that we can leverage by enabling the _Corosync two_node, last_man_standing, and wait_for_all_ options (See: man votequorum) and configuring a fencing delay in Pacemaker.
+
+In combination, these options provide the following behaviour:
+- Both sides retain quorum in the event of a network split.
```
hexfusion, line 93: This is an interesting behavior that should have more details. Again are we talking about etcd quorum?

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+- Service recovery is normally predicated on successful fencing.
+- When the fenced node reboots, it will not be able to obtain quorum until it has seen its peer (preventing an infinite fencing cycle).
+- To avoid both sides fencing each other at the same time, fencing is typically configured to occur after a delay on one of the nodes. That node will perform recovery in the event of a true peer failure, but will always lose in the response to a pure network failure.
+
+### Risks and Mitigations
+- A two node etcd cluster does not have quorum whenever the communication between the peers is interrupted (by network or peer failure).
```
hexfusion, line 99: This has been touched upon in other commends but it should be crystal clear if this is 2 single node etcd clusters in the design or a 2 node etcd quorum. 2 node etcd has a higher percentage failure case from hardware perspective than single node without gaining any fault tolerance. With future ARM deployment capabilities I am questioning the true cost savings over engineering/support complexity as well.

enhancements/single-node/aos-sno-pair.md
hexfusion: typically this is called split-brain so I would wonder by this would be a goal.

enhancements/single-node/aos-sno-pair.md
hexfusion: learner design doc for ref: https://etcd.io/docs/v3.5/learning/design-learner/

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+This enhancement describes setting up a two nodes cluster with a reduced footprint which will serve edge cases.
```
beekhof, line 26:
```suggestion
This enhancement does **not** propose replacing or otherwise modifying OpenShift's use of etcd.

This enhancement proposes a layer on top of exactly two standard Single Node OpenShift clusters that monitors for node-level failures and uses existing OpenShift APIs to scale up Deployments once the peer has been put into a safe state.
```

enhancements/single-node/aos-sno-pair.md
beekhof: @cgwalters I've updated the summary to better indicate that we're not proposing messing with OCP's use of etcd.

beekhof: @hexfusion Yep, I'm still banging on about this :-)  

Learner is cool, but won't solve a 2-node network split scenario on its own.  Strictly speaking both sides should loose quorum and the cluster would be hosed.  If a winner is pre-ordained (eg. the leader), there is a decent chance that's the one that will fail and the cluster is still hosed.  If both sides are forced to retain quorum, you still need a mechanism to prevent volumes being mounted, and StatefulSets from being started, on both sides.

enhancements/single-node/aos-sno-pair.md
beekhof:
```suggestion
Provide the ability for exactly two all-in-one OpenShift clusters to operate as a **predefined** pair in an **active-active** configuration, detect when its peer has died and **automatically** take over its workloads after ensuring it is safe to do so.
```

enhancements/single-node/aos-sno-pair.md
beekhof: Agreed.  Lets standardize on either `RHEL-HA cluster` or `Pacemaker cluster` to refer to the higher level 2-node entity

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+#### Fencing
+In the event of a peer failure, any **workloads requiring singleton semantics** (such as StatefulSets or kubevirt VMs) require the existing node to be put into a safe state (stopped or otherwise known to be not running the workloads) before they can be started on the surviving peer.  The traditional term for this is “**fencing**” and it is an **optional** part of RHEL-HA clusters.
+
+Customer environments without any singleton workloads may not require fencing, but need to be comfortable with the possibility of workloads being active on both peers at the same time.
+
+For customers that do have singleton workloads, the RHEL-HA stack supports a wide variety of BMCs including those conforming to the IPMI and RedFish standards, as well as SSH and libvirt for testing.
+
+It also supports indirect fencing methods like poison-pill which can optionally make use of shared storage for out-of-band signalling and heartbeating where available.
+
+#### Quorum handling
```
beekhof, line 84:
```suggestion
#### Quorum handling within the RHEL-HA cluster layer
```

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+_The "two node cluster" is a use case that requires special consideration.  With a standard two node cluster, each node with a single vote, there are 2 votes in the cluster. Using the simple majority calculation (50% of the votes + 1)  to calculate quorum, the quorum would be 2.  This means that the both nodes would always have to be alive for the cluster to be quorate and operate._
```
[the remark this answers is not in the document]
beekhof: Yep. Not referring to etcd quorum here

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+<br>Source: man votequorum
+
+For environments in which customers will allow it, Corosync also ships a lightweight quorum arbitrator that can be run locally or in the cloud on any non-cluster node.  In such cases, no special quorum handling is required and careful placement of the arbitrator also ensures that the surviving peer is reachable by its intended clients.
```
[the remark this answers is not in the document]
beekhof: RHEL-HA

enhancements/single-node/aos-sno-pair.md
beekhof: Correct. This is not etcd.
I believe this is all covered in our product documentation.
@mshitrit maybe dig up a link to something relevant

enhancements/single-node/aos-sno-pair.md
beekhof: @mshitrit can you update this section to better indicate that this is why we're **not** trying to create a a 2-node etcd cluster

enhancements/single-node/aos-sno-pair.md
beekhof: I don't think graduating to GA is assumed at this point.
At the moment, we're just looking to socialize the concept prior to putting a demo in the hands of customers for feedback.  There are still issues around config and data synchronization that we need to have a story for before this is something we can GA.

enhancements/single-node/aos-sno-pair.md
beekhof: Only when things go wrong 😇  🤣

enhancements/single-node/aos-sno-pair.md
beekhof: RHEL-HA includes a python based `fence-agents` package that the RHEL cluster uses to interface with BMC hardware.  For configuring the RHEL cluster, there is a `pcs` command line tool that we will need to drive or wrap. possibly based on the contents of a new CRD that describes the rest of the RHEL cluster (including peer ips).

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+- DR for services in clusters with **more than a single node.**
+- DR for services running on **two or more all-in-one clusters.**
+- Enable services uptime during a **remote cluster upgrade process.**
+
+## Proposal
+
+Assuming workloads that require some form of coordination between the two nodes (either at-most-one semantics or simply to reduce machine load during normal operation):
+
+- Pacemaker and Corosync from the RHEL-HA cluster stack are deployed [in Pods](https://github.com/beekhof/installer/blob/aio/pod-cluster.yaml.in) on both OCP clusters, and they are configured to form a RHEL-HA cluster which is composed of two independent SNO clusters.
+- The Pacemaker cluster is configured so that on peer failure the cluster can reliably ensure that the workload on an unhealthy peer is stopped before starting the workload on the healthy peer. Configuration [flexibility](https://github.com/openshift/enhancements/blob/fb3faa767a04866ee6d6ef46127636bd086df658/enhancements/single-node/aos-sno-resources/2%20Node%20HA%20Menu.svg) is provided here to allow for the anticipated variety of customer requirements.
```
dhellmann, line 53: I think the URL can just point to the local file and rely on relative URL resolution. Otherwise every time we update the image we have to update  the link, too.

enhancements/single-node/aos-sno-pair.md
dhellmann: Is there going to be a separate enhancement to describe that API, or should it be included here?

enhancements/single-node/aos-sno-pair.md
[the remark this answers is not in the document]
dhellmann: We've recently had some conversations about the difference in support lifecycles for RHEL and OCP. It would be good to call that out as a risk in this doc.

enhancements/single-node/aos-sno-pair.md
beekhof: The CRD will be included here once we finalize the details

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,143 @@
...
+## Implementation History
+
+## Drawbacks
+This approach introduces a non-kubernetes HA approach to an OpenShift environment, requiring users to understand both kubernetes and RHEL-HA.
+
+## Alternatives
+
+1. Support an active/passive configuration with 1 master + 1 worker.  Workloads stay running on the worker if the master dies (unless they require rescheduling before the master is recovered), but the master takes over if the worker dies.
+2. Run with a single active etcd copy, but make use of [etcd learners](https://github.com/etcd-io/etcd/blob/master/Documentation/learning/design-learner.md) to provide a non-live replica that can be promoted in the event the active copy fails.
+   Learners will gain self-promotion abilities for automated failover in etcd 3.5.  Would require additional engineering (either in etcd or some component driving it) for the learner to distinguish between a network and leader failures, and/or ensure the leader is no longer accepting writes before promoting itself.
```
beekhof, line 140:
```suggestion
   Learners will gain self-promotion abilities for automated failover in etcd 3.5.  Would require additional engineering (either in etcd or some component driving it) for the learner to distinguish between a network and leader failures, and/or ensure the leader is no longer accepting writes before promoting itself.
   Strictly speaking both sides should loose quorum and the cluster would be hosed. If a winner is pre-ordained (eg. the leader), there is a significant chance that's the one that will fail and the cluster is still hosed. If both sides are forced to retain quorum, you still need a mechanism to prevent volumes being mounted, and StatefulSets from being started, on both sides.
```

enhancements/single-node/aos-sno-pair.md
beekhof: done

enhancements/single-node/aos-sno-pair.md
mshitrit: [here]( https://docs.google.com/presentation/d/1tAMRBDOK5PUzS69t-BY8-Dxjw5HRYRHAx4b2nNrhjMY/edit#slide=id.gb14b7c36b7_0_0) is the presentation, I think it covers it quite nicely.

beekhof: @mshitrit I meant the RHEL-HA product documentation

enhancements/single-node/aos-sno-pair.md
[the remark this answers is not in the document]
beekhof: Aren't RHEL based products generally supported for much longer than OCP?  What specific concerns have been raised?

The RHEL-HA team has a long history of being extremely responsive when bugs are found and proactive when it comes to backports.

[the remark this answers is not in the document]
dhellmann: IIRC, the RHEL lifetime was less than what we've been asked to support for OCP. Either way, the timelines don't always align, which makes extended support challenging sometimes.

enhancements/single-node/aos-sno-pair.md
beekhof: @dhellmann Might be a little premature. We're not actively talking about GA yet.

mshitrit: Added some basic details.
Per Andrew's comment I assume it'll expand more if necessary when it'll be more relevant.

enhancements/single-node/aos-sno-pair.md
dhellmann: Listing those items as requirements for GA is exactly the sort of detail we should capture early, so it isn't lost.

enhancements/single-node/aos-sno-pair.md
```diff
@@ -0,0 +1,158 @@
...
```
hexfusion, line 154: Just a note automated failover did not land in etcd 3.5.

enhancements/single-node/aos-sno-pair.md
mshitrit: I've added a template on how we plan to fill GA Criteria.
We will fill in more concrete details based on milestones progress.

mshitrit:
The conversation arrived at the conclusion that while the design is not objectionable, the result should not be part of ocp.
Will follow up with the ACM crew to see if it aligns with their goals.

PR #833, opened 2021-07-14
kube-apiserver: Fallback on Failing Revisions of Static Pods

sttts:
Static pods operators use a revisioned pod manifest and revisioned configmap and secret manifests to roll out and start a new configuration of the operand in an atomic way.

Configuration can be bad in different ways. In HA OpenShift, the static pod operator will wait for operands on one node to start up and to become healthy and ready. The operator will stop rolling out bad revisions further.

In SNO, there is no other pod to ensure availability when a roll-out of a new revision happens. Hence, a bad revision can be fatal for the cluster, especially for the kube-apiserver and etcd

This enhancement is describing a fallback mechanism that will make the operand to revert to the last healthy revision when the new revision fails, and how the operator will notice this event and how it will react.

The fallback mechanism is opt-in by the operator, and kube-apiserver and etcd operators (and potentially kcm and ks) have to consult the deployment topology in the infrastructure resource to decide.

### DIFF
--- a/enhancements/kube-apiserver/startup-monitor.md
+++ b/enhancements/kube-apiserver/startup-monitor.md
@@ -41,7 +41,7 @@ Static pods operators use a revisioned pod manifest and revisioned configmap and
 
 Configuration can be bad in different ways. In HA OpenShift, the static pod operator will wait for operands on one node to start up and to become healthy and ready. The operator will stop rolling out bad revisions further.
 
-In SNO, there is no other pod to ensure availability when a roll-out of a new revision happens. Hence, a bad revision can be fatal for the cluster, especially for the kube-apiserver and etcd
+In SNO, there is no other pod to ensure availability when a roll-out of a new revision happens. Hence, a bad revision can be fatal for the cluster, especially for the kube-apiserver and etcd.
 
 This enhancement is describing a fallback mechanism that will make the operand to revert to the last healthy revision when the new revision fails, and how the operator will notice this event and how it will react.
 
@@ -84,6 +84,10 @@ If the operand pod does not start up in N minutes, e.g. due to:
 - not answering on the expected port (connection refused)
 - healthy but never readyz
 
+but not in case of
+
+- etcd not healthy
+
 the task for startup-monitor is to fall back:
 
 1. when detecting problems with the new revision, the startup-monitor will copy the pod-manifest of the `/etc/kubernetes/static-pods/last-known-good` link (or the previous revision if the link does not exist, or don't do anything if there is no previous revision as in bootstrapping) into `/etc/kubernetes`.
@@ -104,13 +108,13 @@ As long as the startup-monitor notices an operand pod-manifest of a different re
 
 The startup-monitor watches the mirror pod for readiness, i.e. by
 
-1. watch `/var/log/kube-apiserver/termination.log` for the first start-up attempt (beware of the race of startup-monitor startup and kube-apiserver startup). Set Reason=*NeverStartedUp* when this times out.
-2. watch `/var/log/kube-apiserver/termination.log` for more than one start-up attempt. Set Reason=*CrashLooping* if more than one is found and the monitor times out.
-3. check https://localhost:6443/healthz. Set Reason=*Unhealthy* if this is red and the monitor times out.
-4. check https://localhost:6443/readyz. Set Reason=*NotReady* if this is red and the monitor times out. Reason=EtcdUnhealthy if the etcd post-start-hook never finished. In all case: message should contain the unfinished post-start-hooks.
+1. watch `/var/log/kube-apiserver/termination.log` for the first start-up attempt (beware of the race of startup-monitor startup and kube-apiserver startup). Set Reason=**NeverStartedUp** when this times out.
+2. watch `/var/log/kube-apiserver/termination.log` for more than one start-up attempt. Set Reason=**CrashLooping** if more than one is found and the monitor times out.
+3. check https://localhost:6443/healthz. Set Reason=**Unhealthy** if this is red and the monitor times out.
+4. check https://localhost:6443/readyz. Set Reason=**NotReady** if this is red and the monitor times out. Reason=EtcdUnhealthy if the etcd post-start-hook never finished. In all case: message should contain the unfinished post-start-hooks.
 4. get the mirror pod from the localhost kube-apiserver
-5. check the revision annotation is the expected one. Set Reason=*NotReady* if this is red and the monitor times out.
-6. checking status.ready. Set Reason=*NotReady* if this is red and the monitor times out.
+5. check the revision annotation is the expected one. Set Reason=**NotReady** if this is red and the monitor times out.
+6. checking status.ready. Set Reason=**NotReady** if this is red and the monitor times out.
 
 ### Retries
 
--- a/enhancements/kube-apiserver/startup-monitor.md
+++ b/enhancements/kube-apiserver/startup-monitor.md
@@ -53,8 +53,8 @@ In SNO, we only have one chance to start the new static pod for the kube-apiserv
 
 Bad configurations can have many reasons:
 - **the pod manifest can be wrong**: e.g. by YAML syntax, not validating as a pod or some deeper semantical error:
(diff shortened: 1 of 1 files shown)


enhancements/kube-apiserver/startup-monitor.md
```diff
@@ -0,0 +1,219 @@
...
+- free port wait loop timeout
+- crash-looping
+- not answering on the expected port (connection refused)
+- healthy but never readyz
+
+the task for startup-monitor is to fall back:
+
+1. when detecting problems with the new revision, the startup-monitor will copy the pod-manifest of the `/etc/kubernetes/static-pods/last-known-good` link (or the previous revision if the link does not exist, or don't do anything if there is no previous revision as in bootstrapping) into `/etc/kubernetes`.
+
+It will add annotations to the old revision pod manifest:
```
mfojtik, line 95: is having annotation enough? how would that be discovered by a human?

if admin make config change that that change is being reverted by non-ready API server, how do they know they did something wrong?

sttts: The operator will publish this in degraded annotation.

sttts: clarified.

enhancements/kube-apiserver/startup-monitor.md
```diff
@@ -0,0 +1,234 @@
...
+
+- recover from a badly formatted or invalid pod manifest
+- recover from bad config observer output
+- recover from bad revisioned ConfigMaps and Secrets.
+
+### Non-Goals
+
+- recover from invalid non-revisioned certificates
+- recover from missing or invalid files on the masters given by static paths outside of the `/etc/kubernetes` directory (e.g. some certs or kubeconfig the operand is consuming)
+- optimize downtime of the API beyond in fallback case to O(60s). This mechanism is for the disaster case and only survival counts. If downtime is `O(5min)` or even more, this is completely ok.
```
soltysh, line 74: Pod manifest validation seems like a reasonable addition to the operator, although I wonder if this should be covered here as well.

sttts: dry-run can catch a reasonable amount of these I would say. Though I bet it's not complete.

enhancements/kube-apiserver/startup-monitor.md
```diff
@@ -0,0 +1,234 @@
...
+
+2. if the operand becomes ready, the startup-monitor will link the revision to `/etc/kubernetes/static-pods/last-known-good`, and then remove its own pod-manifest from `/etc/kubernetes`, and hence commit suicide.
+
+As long as the startup-monitor notices an operand pod-manifest of a different revision (by checking the operator manifest in `/etc/kubernetes` and the revision annotation), it will do nothing and just wait, continuing to watch the operand pod-manifest. This is important to avoid races on startup, and avoid problems on downgrade before this mechanism was introduced.
+
+### Readiness
+
+The startup-monitor watches the mirror pod for readiness, i.e. by
+
+1. watch `/var/log/kube-apiserver/termination.log` for the first start-up attempt (beware of the race of startup-monitor startup and kube-apiserver startup). Set Reason=**NeverStartedUp** when this times out.
```
p0lyn0mial, line 108: what does it mean in practice? reading a line and comparing the content and timestamp?

enhancements/kube-apiserver/startup-monitor.md
```diff
@@ -0,0 +1,234 @@
...
+### Coordination of Installer and Startup-Monitor
+
+Startup-monitor and installer to non-atomic work, and hence we need to coordinate work between them to protect the following case:
+an installer is in progress and wants to install a new revision, the current revision is not ready and we are about to fall back to the previous version. The installer writes the new file and we immediately overwrite it.
+
+Hence, we will write a lock file to /var/lock/kube-apiserver-installer.lock, and make the installer and the startup-monitor readiness procedure to wait for it.
+
+### Pruner
+
+The pruner today does not know about the `last-known-good` symlink to the last successful revision. It must be extended to ignore it and the linked revision.
```
p0lyn0mial, line 150: would it be okay to extend the pruner controller to read the annotations from the kas.pod and put found revision onto a protected list before starting the pruner pod?

enhancements/kube-apiserver/startup-monitor.md
```diff
@@ -0,0 +1,234 @@
...
+
+#### Tech Preview -> GA
+
+#### Removing a deprecated feature
+
+### Upgrade / Downgrade Strategy
+
+On upgrade, if that even matters for SNO pre-GA, the `last-known-good` symlink won't exist on first rollout. We will fall back to the previous revision then.
+
+On downgrade, the startup-monitor might still exist in `/etc/kubernetes`. We will make it wait until it sees an operand pod-manifest of the same revision before doing anything (see proposal).
```
p0lyn0mial, line 217: could the startup monitor exit when the pod revision is > than the current revision for N minutes?

enhancements/kube-apiserver/startup-monitor.md
p0lyn0mial: would it make sense to keep track of `pid` instead/additionally?

enhancements/kube-apiserver/startup-monitor.md
```diff
@@ -0,0 +1,234 @@
...
+
+The startup-monitor watches the mirror pod for readiness, i.e. by
+
+1. watch `/var/log/kube-apiserver/termination.log` for the first start-up attempt (beware of the race of startup-monitor startup and kube-apiserver startup). Set Reason=**NeverStartedUp** when this times out.
+2. watch `/var/log/kube-apiserver/termination.log` for more than one start-up attempt. Set Reason=**CrashLooping** if more than one is found and the monitor times out.
+3. check https://localhost:6443/healthz. Set Reason=**Unhealthy** if this is red and the monitor times out.
+4. check https://localhost:6443/readyz. Set Reason=**NotReady** if this is red and the monitor times out. Reason=EtcdUnhealthy if the etcd post-start-hook never finished. In all case: message should contain the unfinished post-start-hooks.
+4. get the mirror pod from the localhost kube-apiserver
+5. check the revision annotation is the expected one. Set Reason=**NotReady** if this is red and the monitor times out.
+6. checking status.ready. Set Reason=**NotReady** if this is red and the monitor times out.
```
p0lyn0mial, line 114: would it make sense to add a step(s) that would `get, list, watch` some resources?

enhancements/kube-apiserver/startup-monitor.md
```diff
@@ -0,0 +1,234 @@
...
+- 5 min is already a very generous duration for kube-apiserver startup. In normal situation we are in the area of O(30s).
+
+The timeout is to be specified in the operator. Hence, another operator might want different timeouts.
+
+### Retries
+
+The readiness procedure above has the risk of considering external reasons as fatal reasons for kube-apiserver readiness. E.g. etcd unstable or kubelet not updating the mirror pod. Both would led us to believe the new revision is broken.
+This is different in HA OpenShift because there is no time limit exists for a new static pod to start up. It can keep crash-looping for hours and eventually start-up to finish the roll-out.
+
+In order to mitigate this difference in error behaviour, a failing revision in SNO will be retried to roll-out after >=N minutes, N=O(10), a reasonable time for the cluster to recover from a previous attempt. In the time in-between, the fallback mechanism will move the cluster back to the old revision, until the retry is started.
```
p0lyn0mial, line 128: I hope there is a way to force a new rollout without bumping the revision counter.

enhancements/kube-apiserver/startup-monitor.md
```diff
@@ -0,0 +1,234 @@
...
+## Implementation History
+
+Major milestones in the life cycle of a proposal should be tracked in `Implementation
+History`.
+
+## Drawbacks
+
+## Alternatives
+
+- we have talked about starting the new apiserver in parallel and to have some proxy or iptables mechanism to switch over only when the new revision is ready. This was rejected as we don't have the memory resources on nodes to run two apiservers in parallel.
```
aojea, line 232: the state of the art on zero disruption is "socket takeover", keeping the new and old process in parallel and transfer the sockets, you can read more here:
https://copyconstruct.medium.com/file-descriptor-transfer-over-unix-domain-sockets-dcbbf5b3b6ec
https://blog.apnic.net/2021/03/29/how-facebook-achieves-disruption-free-updates-with-zero-downtime/

enhancements/kube-apiserver/startup-monitor.md
sttts: there is.

enhancements/kube-apiserver/startup-monitor.md
sttts: follow-up

enhancements/kube-apiserver/startup-monitor.md
sttts: it would be restarted by kubelet.

PR #836, opened 2021-07-15
ETCD-58: Add enhancement for automated defragmentation

hexfusion:
This enhancement is the documentation for the defragmentation controller which shipped in 4.9.

### DIFF
--- a/enhancements/etcd/automated-defrag.md
+++ b/enhancements/etcd/automated-defrag.md
@@ -10,7 +10,7 @@ approvers:
   - "@smarterclayton"
   - "@deads"
 creation-date: 2021-07-14
-last-updated: 2021-07-14
+last-updated: 2021-08-06
 status: implementable
 
 see-also:
@@ -33,8 +33,8 @@ Enable automated defragmentation of the etcd cluster state.
 
 ## Motivation
 
-As defragmentation is an important part of cluster health, managing this operation with a controller aligns with the 
-principals and promises of OpenShift 4. 
+As defragmentation is an important part of cluster health, managing this operation with a controller aligns with the
+principals and promises of OpenShift 4.
 
 ### Goals
 
@@ -46,8 +46,8 @@ principals and promises of OpenShift 4.
 
 ## Proposal
 
-We propose to add a defrag controller to the `cluster-etcd-operator` that will check the endpoints of the etcd 
-cluster every 10 minutes for observable signs of fragmentation. 
+We propose to add a defrag controller to the `cluster-etcd-operator` that will check the endpoints of the etcd
+cluster every 10 minutes for observable signs of fragmentation.
 
 [openshift/cluster-etcd-operator PR #625](https://github.com/openshift/cluster-etcd-operator/pull/625)
 
@@ -57,8 +57,8 @@ cluster every 10 minutes for observable signs of fragmentation.
 - I want to ensure that heavy churn workloads do not negatively impact cluster performance.
 - I want to ensure that cluster operands utilize only the necessary resources to provide service.
 
-Large production clusters such as OpenShift CI feel the pain of fragmentation daily. The constant churn of resources 
-from creating and deleting many jobs can quickly lead to resource bloat as the result of fragmentation. This growth can 
+Large production clusters such as OpenShift CI feel the pain of fragmentation daily. The constant churn of resources
+from creating and deleting many jobs can quickly lead to resource bloat as the result of fragmentation. This growth can
 lead to OOM events and downtime. This controller was prototyped as a result of a CI outage.
 
 ### Implementation Details/Notes/Constraints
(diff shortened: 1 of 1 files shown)


enhancements/etcd/automated-defrag.md
```diff
@@ -0,0 +1,235 @@
...
+### Implementation Details/Notes/Constraints
+
+## etcd Storage Backend
+etcd's storage backend is powered by `bbolt` a B+Tree memory mapped key/value store with MVCC support. While MVCC is
+necessary to support fully concurrent transactions and the heart of the `Watch` functionality.
+MVCC poses a performance challenge for `bbolt` because the db can store many versions of the keys value. These
+versions can build up indefinitely and result in memory and disk performance issues. The solution is to regularly
+compact the key space.
+
+compaction is essentially a delete action in `bbolt`, where the prevision versions of a keys value are pruned.
```
lilic, line 73: Nit:

```suggestion
Compaction is essentially a delete action in `bbolt`, where the prevision versions of a keys value are pruned.
```

enhancements/etcd/automated-defrag.md
```diff
@@ -0,0 +1,235 @@
...
+    // dbSizeInUse is the size of the backend database logically in use, in bytes, of the responding member.
+    DbSizeInUse int64 `protobuf:"varint,9,opt,name=dbSizeInUse,proto3" json:"dbSizeInUse,omitempty"`
+}
+```
+
+From these two fields we can conclude the total fragmented space on that etcd members backend store. If the
+percentage of bytes of the total keyspace are over a defined threshold percentage. The controller will 
+utilize maintenance API again and issue a `DefragmentRequest` RPC.
+
+Defragmentation is I/O blocking thus important to only defrag a single etcd at a time. Additionally, to reduce
```
lilic, line 100: How is this going to be achieved, to restrain defrag to a single etcd at a time? Could we detail that a bit, sorry if I missed it somewhere?

enhancements/etcd/automated-defrag.md
```diff
@@ -0,0 +1,235 @@
...
+defragment the store is directly related to the amount of free pages that exists. For this reason we feel it
+makes sense to continuously observe the clusters state and defrag frequently along the cadence of 
+compaction resulting in shorter intervals of disruption by keeping the number of pages limited and the size of the 
+database smaller.
+
+The criteria for defragmentation is defined as:
+
+- Cluster must be observed as `HighlyAvailableTopologyMode`.
+- The etcd cluster must report all members as healthy.
+- `minDefragBytes` this is a minimum database size in bytes before defragmentation should be considered.
```
lilic, line 111: What is this value?

enhancements/etcd/automated-defrag.md
```diff
@@ -0,0 +1,235 @@
...
+TODO show performance graphs.
+
+### Risks and Mitigations
+
+- SNO: As defragmentation is I/O blocking it is unsafe for non HA clusters. Mitigation to this risk includes a check in 
+  the controller for control-plane topology and will exit 0 if non HA.
+  
+- datafile corruption: Recently a bug was fixed in upstream etcd[1] which could lead to corruption of the key space 
+  as a result of defragmentation if it happened while etcd was being terminated. Mitigation to this risk includes 
+  heavy exercise in CI and telemetry will alert `etcdMemberDown` if one of the etcd instances goes down as the
```
lilic, line 124: etcdMemberDown fires after 1 minute, is this short or long enough for this?

lilic (commented):
Great work and really detailed proposal! Few questions from my side only.

enhancements/etcd/automated-defrag.md
hexfusion: I am thinking 100MB to reduce early defrag during install. Its possible this is not needed but I think 100MB would be reasonable.

lilic: Sounds good, now I see the need to send this data back to telemetry, thanks!

enhancements/etcd/automated-defrag.md
hexfusion: the timeout for the call is 45s technically we could see a wait that long but we would miss 1m.

enhancements/etcd/automated-defrag.md
hexfusion: will add in more detail

lilic (approved):
/lgtm

lilic (approved):
/lgtm

Thanks!

hexfusion:
/remove-lifecycle rotten

This needs to get merged

dhellmann:
@hexfusion is it in a state where it's ready to go now? I see @deads2k is listed as an approver, does anyone else need to look at it?

hexfusion:
@dhellmann This is more or less documentation as the feature will be shipped in 4.9.

cc @mfojtik

sdodson:
/approved
Assuming since this shipped in 4.9 we should land this, if there are material differences between final impelmentation and what's defined in this enhancement please find some time to update the enhancement contents.

sdodson:
The linter appears to only check modified files so I'm going tomerge this, whomever touches this file next will have to clean up the linting problem.

PR #821, opened 2021-07-02
enhancements/update/targeted-update-edge-blocking: Propose a new enhancement

wking:
This enhancement proposes a mechanism for blocking edges for the subset of clusters considered vulnerable to known issues with a particular update or target release.

This is the third iteration of this proposal, replacing #426 (update-service-side decisions based on Telemetry queries) and #737 (update-service-side decisions based on client-supplied query parameters).  I'm now recommending cluster-side decisions, with the update service declaring conditional edges and telling the cluster-side tools how to decide if the cluster is vulnerable to the known bugs impacting its update decisions.

PR #465, opened 2020-09-02
Insights operator up to date gathering

martinkunc:
This is suggesting how Insights Operator could become untied from release cycle and offer stability and dynamic changes when adding new data.

PR #837, opened 2021-07-16
Conditional Data Gathering for Insights Operator

Sergey1011010:


enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,199 @@
...
+
+Collecting data on some condition such as an alert is quite common pattern 
+during the root-cause analysis and the ability to define the rules 
+what to collect on which condition externally will help to do that faster.
+
+### Goals
+
+- Implement validation of the rules for conditional gatherer
+- Create a service providing these rules
+- Fetch them from there and apply to the conditional gatherer
```
tremes, line 50: I would probably mention the Git repository here as the source of the gathering rules defintions. I aslo think we should be explicit so instead of saying:
```
Fetch them from there ....
```
we should say something like `Fetch the rules definitions from the service API` or something like that.

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,199 @@
...
+against JSON schema and possibly some review process. 
+We have created a PoC for the repo: 
+
+https://github.com/tremes/insights-operator-gathering-rules
+
+or the version with an example of JSON schema validation:
+
+https://github.com/tremes/insights-operator-gathering-rules/tree/schema 
+
+2. There's a service living in cloud.redhat.com fetching the rules from the repo and providing them through its API.
```
tremes, line 106: I think we should already say `console.redhat.com`

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,199 @@
...
+- For each rule, there's at least one gathering function
+- Only implemented conditions can be used
+- Alert name from `alert_is_firing` condition should be a string of length between 1 and 128 
+and consist of only alphanumeric characters
+- Only implemented gathering functions can be used
+- Namespace from `gather_logs_of_namespace` function should be a string of length between 1 and 128, 
+match the next regular expression `^[a-zA-Z]([a-zA-Z0-9\-]+[\.]?)*[a-zA-Z0-9]$` and start with `openshift-`
+- Tail lines from `gather_logs_of_namespace` function should be an integer between 1 and 4096
+- Namespace from `gather_image_streams_of_namespace` function should be a string of length between 1 and 128, 
+match the next regular expression `^[a-zA-Z]([a-zA-Z0-9\-]+[\.]?)*[a-zA-Z0-9]$` and start with `openshift-`
```
tremes, line 122: Not sure but these are maybe impl details from the perspective of this document.

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,199 @@
...
+## Motivation
+
+Collecting data on some condition such as an alert is quite common pattern 
+during the root-cause analysis and the ability to define the rules 
+what to collect on which condition externally will help to do that faster.
+
+### Goals
+
+- Implement validation of the rules for conditional gatherer
+- Create a service providing these rules
```
tremes, line 49: I think this service is described here very briefly and would deserve more. What about e.g authentication, reading the data from the repository etc.?

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,199 @@
...
+
+Empty
+
+### Risks and Mitigations
+
+The potential risks could come from an attacker spoofing the config somehow (if they got access to the repo),
+all they could do is let the insights operator collect more data and send it to the c.r.c, but because of validation
+on the operator side, the potential of collecting data is limited. For example, 
+we check that namespaces start with `openshift-`, we're also limiting the amount of potentially collected data by,
+for example, introducing the limit for amount of collected logs.
```
dhellmann, line 145: Is the service going to pull directly from the GitHub repo? Or do we mirror that somewhere internally? How does the system behave if GitHub is unavailable for some reason?

Is there any way for an attacker to introduce a data collection rule that would expose sensitive information about the cluster?

Does the validation protect against DDoS attacks from collecting excessive information?

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Graduation Criteria
+
+Empty
+
+### Upgrade / Downgrade Strategy
+
+The conditional gatherer itself (e.g. adding new/removing old conditions or gathering functions) is 
+modified through a standard process like all other operator's code. Only the rules to collect X with Y params on Z
+are updated through the new process described above.
```
dhellmann, line 176: Does this enhancement need to cover schema changes for the config files, or is that addressed in another document already?

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,199 @@
...
+
+A gathering function consists of its name and parameters, implemented functions are:
+
+- `logs_of_namespace` which collects logs from each container in the namespace from parameter `namespace`
+limiting it to only last N lines from parameter `tail_lines`
+- `image_streams_of_namespace` which collects image streams of the namespace from parameter `namespace`
+
+The proposal is to implement the next process of adding new rules:
+
+1. We have a repo with json configs defining the rules. The repo will have some simple CI with validation
```
dhellmann, line 96: We have this repo already, or we will create it? Where?

enhancements/insights/conditional-data-gathering.md
Sergey1011010: > Is the service going to pull directly from the GitHub repo? 

Yes

> How does the system behave if GitHub is unavailable for some reason?

The service sitting in the middle of the repo and Insights Operator is not implemented yet, but it can cache the latest version of the rules in case GitHub fails. If the service living in console.redhat.com is not available, then Insights Operator would not collect conditional data, but still collect all other data like before. 

> Is there any way for an attacker to introduce a data collection rule that would expose sensitive information about the cluster?

No. In case they somehow get access to the repo, they would be able to introduce a new rule, for example, collecting logs or imagestreams from other openshift related namespace and make Insights Operator send an archive containing this data, but all the anonymization would still be applied (we remove ssh keys, certs and so on) and the data would still be sent to our pipeline, the attacker cannot redirect where the archive goes to. 

> Does the validation protect against DDoS attacks from collecting excessive information?

Right now it limits the amount of data collected by each gathering rule, but I think we also need to introduce the limit for the number of rules the operator would fetch. But even now, we have a timeout for each gatherer and it would give up after some time (but we would still get other non conditional data as before).

enhancements/insights/conditional-data-gathering.md
Sergey1011010: It's not, I will add it to the doc.

enhancements/insights/conditional-data-gathering.md
Sergey1011010: Not yet, it's supposed to live in https://github.com/RedHatInsights

enhancements/insights/conditional-data-gathering.md
dhellmann: OK, could you cover these risks in the doc? For example, if Github has an outage we wouldn't be able to add new rules. That's a small risk, but it's there and we should acknowledge it. If we think the other items are not risks, we should explain why -- they're still questions about risk that could be raised about the design.

enhancements/insights/conditional-data-gathering.md
dhellmann: OK, that detail would be good to include in the document.

enhancements/insights/conditional-data-gathering.md
Sergey1011010: I'll elaborate more on the risks, thanks

enhancements/insights/conditional-data-gathering.md
Sergey1011010: fixed

enhancements/insights/conditional-data-gathering.md
Sergey1011010: I think it's better to keep it, maybe move to another sections, but it should be there cuz it's important to understand how safe the solution is

enhancements/insights/conditional-data-gathering.md
Sergey1011010: Described more. Didn't mention anything about authentication because now we aren't gonna have any.

enhancements/insights/conditional-data-gathering.md
Sergey1011010: The idea of how the service would work has changed lil bit, now we're gonna include the rules in the image so the service doesn't need to talk to GitHub. Added a few more things to the risks.

enhancements/insights/conditional-data-gathering.md
Sergey1011010: Added as well

enhancements/insights/conditional-data-gathering.md
mklika: are we ok to resolve this - i.e. keep it in the doc ... @tremes ?

tremes: I am fine with resolving this conversation.

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,231 @@
...
+History`.
+
+## Drawbacks
+
+None?
+
+## Alternatives
+
+Thin IO idea (https://github.com/openshift/enhancements/pull/465) could also solve this, but thin IO was about adding
+new code bypassing the standard release process which could potentially be very dangerous.
```
wking, line 226: This proposal is keeping the rule development in the existing insights-operator repository.  But then it's delivering them via a new, central console.redhat.com API, so you can end up with new-rule-vs-older-insights-operator-code combinations that have not been tested by CI/QE.  And I don't see anything addressing [the signing and restricted-network concerns][1] raised in the previous PR.  The OLM approach floated in previous discussion seems like it would cover both of those without needing a new API.

[1]: https://github.com/openshift/enhancements/pull/465#issuecomment-770977399

Sergey1011010: Well, there will be basic tests, but yes, we can't test every single combination. The Thin IO was very different, it was about running new code, here we have only freedom to run existing functions with different validated parameters.

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,231 @@
...
+
+https://github.com/tremes/insights-operator-gathering-rules
+
+The JSON schema can be found here:
+
+https://raw.githubusercontent.com/openshift/insights-operator/f9b762149cd10ec98079e48b8a96fc02a2aca3c6/pkg/gatherers/conditional/gathering_rule.schema.json
+
+2. There will be a service living in console.redhat.com which will have the rules baked into the container and will
+provide all the rules through its API. The very first version of the service is going to be simple, but later we may
+add some filtering on the API level (for example by cluster version).
```
bparees, line 108: having the config/gathering-rules be based on the OCP version seems like a core requirement as the particular logs you want to gather may change between versions, as may the alerts that are relevant.

i'd prefer to see a design that takes this into account from the start rather than trying to add version-awareness in later.

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,231 @@
...
+
+3. Insights Operator fetches a config with the rules from the service and unmarshalls JSON to Go structs.
```
bparees, line 110: i'd like to see some discussion in the "alternatives" section as to why just baking the logic into the insights operator itself was rejected  (concern that you'll need to update the config more dynamically than waiting for customers to upgrade their clusters?)

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,231 @@
...
+against JSON schema and possibly some review process. The repo should live in https://github.com/RedHatInsights
+We have created a PoC for the repo:
+
+https://github.com/tremes/insights-operator-gathering-rules
+
+The JSON schema can be found here:
+
+https://raw.githubusercontent.com/openshift/insights-operator/f9b762149cd10ec98079e48b8a96fc02a2aca3c6/pkg/gatherers/conditional/gathering_rule.schema.json
+
+2. There will be a service living in console.redhat.com which will have the rules baked into the container and will
```
bparees, line 106: is there precedence for this in the insights operator already (routinely calling back to console.redhat.com to get some config data?)

enhancements/insights/conditional-data-gathering.md
bparees: how often/when will the insights operator poll for updates to the config?

Sergey1011010: The current design makes it get the config at every run of conditional gatherer, which is 2 hours for the most of clusters.

enhancements/insights/conditional-data-gathering.md
Sergey1011010: Config - not, but we're getting the aggregator's data (https://cloud.redhat.com/api/insights-results-aggregator/v1/clusters/%s/report)

tremes: In other words we're getting an Insights report (from console.redhat.com) for each uploaded IO archive to provide basic Insights overview for the Insights Advisor in the OCP console.

enhancements/insights/conditional-data-gathering.md
tremes: Well I think a change in the gathered data is not problematic, is it? I mean there's a need to have the information about OCP version, but note it's already in our data (we gather the [version](https://github.com/openshift/insights-operator/blob/master/docs/insights-archive-sample/config/version.json). Checking the OCP version could therefore be on the actual rule evaluation IMO. 
Considering an alert name change, I think it will require a new condition regardless of the OCP version.

bparees: you're saying the rule itself would be something like "when cluster version=X.Y.Z && alertFooFiring then gather log X"

?

I think that covers my concern if so.  What I want to avoid is the case where someone defines a rule like "when alertFooFiring gather logX" that only applies to v4.n, then then on v4.n+1 what you need is "when alertFooFiring gather logY"

tremes: Yes that was my reasoning.

jholecek-rh: I sense misunderstanding (I raised the concern in the past that we are overloading the term "rule"):

1. There are rules of the conditional gathering engine that control what/when the Insights Operator gathers data. Ben seems to be talking about these.
2. There are Insights rules that consume the gathered data and produce recommendations to customers. Tomáš seems to be talking about these.

Ben's request for conditional gathering rules to support version-based conditions makes sense to me. While the Insights Operator does this by default in a way - a specific OCP version runs a specific Insights Operator version, we can still realize in the future that it would be useful to gather more information for a specific OCP version, and this would enable it (limited to harcoded gatherers, of course).

Sorry if I'm totally wrong and just adding noise.

bparees: > Insights Operator does this by default in a way - a specific OCP version runs a specific Insights Operator version

my understanding is that the conditional gathering rules are not defined in the operator itself, so it doesn't matter what operator version is running, the gather rules are defined/served by the external service, so the operator on the cluster needs a way to decide whether or not the gather rule should be used, based on the cluster version the operator is running on (there may be other conditions that determine whether or not the gather rule applies, as well).

So what i'd expect is an api that allows the consuming operator to decide "should i use this gather rule, or ignore it"

How the api provides that version info (encoded into each rule, or by serving up blocks of rules where each block of rules has a field that says which cluster versions should consume that block of rules) i don't particularly care.

tremes: @jholecek-rh correct IMO. Yes sorry I was probably mixing things little bit (we would need some better naming here). Yeah so I was talking about the Insights rule (not the condition rule). 

Well, I think I've changed my mind a bit in the sense that the OCP version should be considered as a type of condition then (triggering X) - basically what Ben suggested.

jholecek-rh: I just did add noise. Sorry. :slightly_frowning_face: Tomáš should think twice before pointing me to a discussion next time. :slightly_smiling_face: 

* Every Insights Operator  version has a hardcoded set of data that it gathers, no conditional gathering rules involved. This set differs between versions. This is what I meant by the sentence.
* You are correct that the conditional gathering rules are an external resource independent of the Insights Operator version.
* The conditional gathering rules can only invoke existing gatherers. Available gatherers differ between Insights Operator versions so one conditional gathering rule can produce different results on different OCP versions.

None of this really matters because you where discussing two options for conditional gathering based on cluster version:

1. Conditional gathering rules are tagged with versions they apply to. When the Insights Operator downloads the gathering rules, it specifies the version it is running and gets only gathering rules for that version.
2. Version conditions are standard part of the gathering rule conditions. The Insights Operator always downloads all rules, but executes only those whose conditions are satisfied on the cluster version where it is running.

If Tomáš meant to ask me which one I prefer, I think that both options are equally powerful. Both options need to support version ranges to avoid "duplicate" rules like the example given by Ben.

The differences are
* Server-side complexity and an extra mechanism for specifying applicable versions (option 1)
* Payload size and client-side load (option 2, the Insights Operator evaluating all rules)

I tend to prefer option 2:
* Simpler implementation
* More straightforward conditions of gathering rules
* I believe we can keep the amount of gathering rules at a reasonable size by retiring rules for old OCP versions

I hope I am making sense now. :flushed:

Sergey1011010: Current vision for how the versioning of the gathering rules would be implement is that we would have another condition, something like this:
```json
[{
  "conditions": [
    {
      "type": "alert_is_firing",
      "params": {
        "name": "SamplesImagestreamImportFailing"
      }
    },
    {
      "type": "cluster_version_matches",
      "params": {
        "version": "4.8.*"
      }
    }
  ],
  "gathering_functions": {
    "logs_of_namespace": {
      "namespace": "openshift-cluster-samples-operator",
      "tail_lines": 100
    },
    "image_streams_of_namespace": {
      "namespace": "openshift-cluster-samples-operator"
    }
  }
}]
```

And the operator would decide which rules to apply, but even later we can also add filtering on the service so that it returns only rules for the specific version. But nothing of it is implemented at the moment.

jholecek-rh: What is the exact semantics of `cluster_server_matches`, please? A quick search in the Insights Operator repo didn't give any hits.

Sergey1011010: @jholecek-rh it's not implemented yet. We can probably use semantic versioning or something. I'm gonna describe how it would work in the enchancement.

jholecek-rh: I see. In that case, please make sure to include a convenient option to specify ranges (i.e. greater than and smaller than operations).

Sergey1011010: Mentioned it here https://github.com/openshift/enhancements/pull/837/files#diff-5db15f2380b3edb22792b4057f2e242db90d08b5c98a67fd210351b1c1a08c41R94

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,236 @@
+---
+title: conditional-data-gathering
+authors:
+  - "@Sergey1011010"
+reviewers:
+  - "@inecas"
+  - "@tremes"
+  - "@smarterclayton"
+approvers:
+  - "@smarterclayton"
```
dhellmann, line 10: Unless @smarterclayton has explicitly agreed to be the approver for this design, we probably want to find someone with more time. Maybe someone from @eparis' group?

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,236 @@
...
+- Create a service taking gathering rules from the git repo and providing them through the API
+- Fetch the rules definitions from the service's API and apply them to the conditional gatherer
+
+### Non-Goals
+
+Empty
+
+## Proposal
+
+Right now we have conditional gatherer which can collect data based on some conditions
```
dhellmann, line 58: Both this design and #821 include a form of rule that looks for characteristics of a cluster before making a decision about what to do. The rules are for different purposes, but it would be worth spending some time considering whether we could share a data structure (nested within the API) and ideally evaluation code, so that we have just 1 implementation. Perhaps if not as a first version, we could have a goal of converging? And if not, let's document our reasoning.

/cc @wking @sdodson

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,236 @@
...
+which are defined in the code, for example the next config would collect 100 lines of logs
+from each container in namespace `openshift-cluster-samples-operator`
+and image streams of the same namespace when alert `SamplesImagestreamImportFailing` is firing
+and add it to the next archive.
+
+```json
+[{
+  "conditions": [{
```
dhellmann, line 66: "conditions" has a rather specific connotation within a kubernetes API. Maybe we can find a different term here? "rules", for example?

Sergey1011010: Well, currently we call conditions + gathering_functions a gathering rule, maybe we need to find better terms cuz it already caused confusing with insights data analysis rules.

enhancements/insights/conditional-data-gathering.md
Sergey1011010: Sure. How do I find a right person?

enhancements/insights/conditional-data-gathering.md
dhellmann: Yeah, it sounds like "rules" is too overloaded as well.

enhancements/insights/conditional-data-gathering.md
dhellmann: Good question. Maybe start a thread in #forum-arch on slack?

enhancements/insights/conditional-data-gathering.md
Sergey1011010: Well ClusterCondition could potentially be shared, it's very similar, but not sure if we would benefit from it. About evaluation code it depends if we would have any common conditions, so far we have AlertIsFiring condition and ClusterVersionMatches will be implemented and you have PromQLClusterCondition. Not sure if it worth the efforts since it may complicate adding new conditions for both parties.

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,237 @@
...
+### Risks and Mitigations
+
+The potential risks could come from an attacker spoofing the config somehow (if they got access to the repo),
+all they could do is let the insights operator collect more data and send it to the c.r.c, but because of validation
+on the operator side, the potential of collecting data is limited and there still would be
+the same anonymization of potentially sensitive data as before.
+For example, we check that namespaces start with `openshift-`, we're also limiting
+the amount of potentially collected data by, for example, introducing the limit for amount of collected logs
+(per container and the number of containers) and, in the worst case, if the conditional gathering takes too much time
+it would be stopped by the timeout and we would just get less data.
```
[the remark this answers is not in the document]
bparees: is there a backstop on the overall size of the logs that can be gathered/shipped?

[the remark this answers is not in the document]
Sergey1011010: There's an implicit one coming from the other limits (number of gathering rules and number of gathering functions for each rule and each function can't collect more than N lines), we don't have anything like "limit the total number of log lines conditional gatherer would get during one run", but we do have a limit for the archive size (8 Mb) and when it's reached no more records are added to the archive. We still get this partially filled archive with the error that some data was lost tho.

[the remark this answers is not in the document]
bparees: ok

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,237 @@
...
+we would just have less data in the archives (everything except conditional gatherer's data),
+an error would be written to the logs and to the metadata which would allow us to know
+that something is broken and we need to bring the service back.
+
+In case GitHub is not available, we won't be able to add new rules, but the old ones would still be returned by
+the service.
+
+## Design Details
+
+Empty
```
smarterclayton, line 183: The total cluster impact (cpu consumed by operator, api, nodes) as a result of this gather should be approximately constant as a function of the number and type of input rules, to prevent scenarios where a mis configuration creates an expensive gather set.  I recommend rate limiting the gathering process for this component so that no more than X impact can be done in Y time regardless of the number or complexity of rules.  I would also like to see design points on the estimated cost of the gather per rule / rule type relative to other known costs within the operator.

Sergey1011010: Besides the limit for the number of rules we also have timeout for each gatherer, so if it takes more than `GatheringInterval / 2`, the gatherer stops collecting the data. For the most of clusters which have gathering interval of 2 hours the timeout is gonna be 1 hour.

Sergey1011010:
@sttts, @deads2k, could you please review the api?

enhancements/insights/conditional-data-gathering.md
Sergey1011010: The task to unify APIs was created https://issues.redhat.com/browse/CCXDEV-5878

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,240 @@
...
+
+### Goals
+
+- Implement validation of the rules for conditional gatherer
+- Create a service taking gathering rules from the git repo and providing them through the API
+- Fetch the rules definitions from the service's API and apply them to the conditional gatherer
+
+### Non-Goals
+
+Empty
```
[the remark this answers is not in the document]
Sergey1011010: addeed

enhancements/insights/conditional-data-gathering.md
tdosek: Guys what about "Tentative gathering" the term should hopefully lead people towards the actual fact that this is not happening every time. We won't be able to bypass all the tech lingo though.

Sergey1011010: In the other enhancement (https://github.com/openshift/enhancements/pull/821/files) with which we unified the APIs they call it "cluster condition", maybe we can also use this term since our conditions do pretty much the same thing.

iNecas: Looking at the synonyms for `condition` in dictionaries, `situational`, `contextual` gathering might fit, but seem like they don't offer a good work for `conditions` word here. 

How about `s/conditions/constrains/g`?

bparees: these are "rules for times/situations in which we should gather these additional pieces of data", right?

conditions seems fine.  rules also seems fine.  constraints does not seem to match.

bparees: another option if you want to get fancy is `heuristic`

iNecas: Yes, I still feel like `conditions` are still the best fit for what this thinks mean.

tremes:
There's one last discussion opened and it seems leaning toward current wording with "conditions" word. 
/approve

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,240 @@
+---
+title: conditional-data-gathering
+authors:
+  - "@Sergey1011010"
+reviewers:
+  - "@inecas"
+  - "@tremes"
+  - "@sttts"
+  - "@deads2k"
```
bparees, line 9: console team should be represented as a reviewer

bparees: sorry, i mixed up my PRs. ignore that comment.

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,240 @@
...
+Right now we have conditional gatherer which can collect data based on some conditions
+which are defined in the code, for example the next config would collect 100 lines of logs
+from each container in namespace `openshift-cluster-samples-operator`
+and image streams of the same namespace when alert `SamplesImagestreamImportFailing` is firing
+and add it to the next archive.
+
+```json
+[{
+  "conditions": [{
+    "type": "alert_is_firing",
```
wking, line 68: Do we want to register this in [the cluster-condition type registry][1]?  You could cover these same conditions with [the already-registered `PromQL` type][2], but I see from looking at the implemention behind the already-landed openshift/insights-operator#519 that a type specific to alerts allows for more convenient result caching than the generic PromQL type allows for.  If the Insights and cluster-version operators end up supporting divergent sets of types, we may want to explicitly document that in the type registry?  Or maybe we should unify our cluster-condition implementations in library-go to avoid diverging?  Or...?

Also, see David rejecting lowercase type names, as documented [here][3].  So `AlertIsFiring` or some such?

[1]: https://github.com/openshift/enhancements/blob/2cc2d9b331532c852878a7c793f3a754914c824e/enhancements/update/targeted-update-edge-blocking.md#cluster-condition-type-registry
[2]: https://github.com/openshift/enhancements/blob/2cc2d9b331532c852878a7c793f3a754914c824e/enhancements/update/targeted-update-edge-blocking.md#promql
[3]: https://github.com/openshift/enhancements/blob/2cc2d9b331532c852878a7c793f3a754914c824e/enhancements/update/targeted-update-edge-blocking.md#lowercase-type-names

Sergey1011010: Well, so far we unified the APIs, not the code. Reusing code is planned in some future, but it's not urgent at all, so I'd rather do it after the enhancement. While PromQL would work, so far there is no need to make it that flexible. 

I'll make it CamelCase, thanks.

enhancements/insights/conditional-data-gathering.md
```diff
@@ -0,0 +1,256 @@
...
+
+Not really a risk, but in case the service is not available or provides invalid config for the operator,
+we would just have less data in the archives (everything except conditional gatherer's data),
+an error would be written to the logs and to the metadata which would allow us to know
+that something is broken and we need to bring the service back.
+
+In case GitHub is not available, we won't be able to add new rules, but the old ones would still be returned by
+the service.
+
+## Design Details
```
tremes, line 188: I am missing here few words about the authentication. I guess the Insights operator will have to authenticate (using the "cloud.openshift.com" token from the `pull-secret`) to the new service. 

I think it would be also good to describe the communication between the operator and the service more. How often will we query the service? What will the service API return - will be there some metadata e.g version or some time of update of the conditions? I remember there were some discussions about caching so this would be good to mention as well IMO.

Sergey1011010: Added some info about implementation. Do you think we need the authentication? The rules aren't secret after all

tremes: I am not sure, I guess everything running in crc requires authentication. We can check with @tisnik.

enhancements/insights/conditional-data-gathering.md
```diff
@@ -187,7 +187,44 @@ the service.
 
 ## Design Details
 
-Empty
+On the initial start the conditional gatherer will try to fetch the config from the service
+by doing http(s) request to the appropriate endpoint `https://cloud.redhat.com/.../gathering_rules`.
+There is a timeout of one minute for the request. The response is then cached and sent again only when 12 hours pass
```
tremes, line 192: Does it mean the Insights operator will be querying the service every 12h? Would it make sense to make this value configurable?

Sergey1011010: Yes, why not

enhancements/insights/conditional-data-gathering.md
tisnik: @tremes Officially it would be needed. This is two-way thing as it allows the cluster (operator) to trust with whom it communicates (so not only for us to trust the cluster)

PR #838, opened 2021-07-16
KNIP-1770: Add proposal for exporting local storage device health metrics

rohantmp:
This adds a proposal for exporting health metrics for each local storage
device collected via tools like smart.

[view preview](https://github.com/rohantmp/enhancements/blob/diskMonitoring/enhancements/local-storage/health-monitoring.md)

@gnufied @jsafrane @sabose @sp98 @priyanka19-98 

Signed-off-by: Rohan CJ <rohantmp@gmail.com>

### DIFF
--- a/enhancements/local-storage/health-monitoring.md
+++ b/enhancements/local-storage/health-monitoring.md
@@ -1,0 +1,155 @@
+---
+title: local-device-health-monitoring
+authors:
+  - "@rohantmp"
+reviewers:
+  - "@jan--f"
+  - "@simonpasquier”
+  - "@sp98" 
+  - "@dhellman"
+
+approvers:
+  - "@simonpasquier”
+  - "@dhellman"
+creation-date: 2021-07-16
+last-updated: 2021-09-01
+status: implementable
+see-also:
+replaces:
+superseded-by:
+---
+
+# Local Device Health Monitoring
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement proposes that we export health metrics for each local device on each node.
+
+## Motivation
+
+Local device vendors expose health information about storage devices via APIs like [SMART](https://en.wikipedia.org/wiki/S.M.A.R.T.). It very useful to know when a storage is about to fail, to differentiate between storage hardware related errors vs purely software ones, etc.
+
+### Goals
+
+- Expose health information about local storage devices in a way that can generate alerts
+
+### Non-Goals
+
+- Align with sig-storage [health monitoring for persistent volumes](https://kubernetes.io/blog/2021/04/16/volume-health-monitoring-alpha-update/)
+  - justification: it is also useful to monitor system drives and CSI only deals with persitent volumes, and migrating to CSI is an extremely long term effort.
+
+## Proposal
+
+[node_exporter](https://github.com/openshift/node_exporter) currently exports per-node metrics, we will use it to collect and export metrics for each device.
+
+### User Stories
+
+- As an admin, I would like to be alerted when a storage device is unhealthy and be linked to remediation steps.
+- As an admin, I would like if components (such as OpenShift Data Foundation) could access storage device health and failover faster.
+
+### Implementation Details/Notes/Constraints
+
+Implement a sidecar in the [node_exporter](https://github.com/openshift/node_exporter) component under [Cluster Monitoring Operator](https://github.com/openshift/cluster-monitoring-operator) that will scrape the metrics from the node and export them via the [textfile exporter](https://github.com/openshift/node_exporter/tree/master/text_collectors) (an exporter that scans `.prom` files in the
+`/var/node_exporter/textfile` directory and exports them.
+
+The Cluster Monitoring Operator will deploy the sidecar when the platform is Baremetal or None by modifying the `node_exporter` [daemonset](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/node-exporter/daemonset.yaml). The platform will be detected by querying the `infrastructure` resource.
+
+After that, the maintainance effort will mostly centered on the code the image runs to get health and update its `.prom` file. The logic for this collector can be in a subdirectory of CMO and have it's own OWNERS file allowing the ODF team to own it.
+
+We will be collecting metrics via [libstoragemgmt](https://github.com/libstorage/libstoragemgmt) and calling it via [libstoragemgmt-golang](https://github.com/libstorage/libstoragemgmt-golang/).
+If a block device is a partion or logical volume, we will only expose health for the parent device.
+
+
+### Ownership and Maintenance
+
+- The `node_exporter` daemonset creation is owned by the Openshift-Monitoring team. The PR for modifying the daemonset based on infrastructure to include the sidecar can be contributed by ODF, but Monitoring team will have to review.
+- The collector sidecar image and logic will live in a separate repository owned and maintained by the ODF team.
+- A new bugzilla subcomponent will be created under openshift-monitoring, but the default assignee will be ODF and ODF will be responsible for triaging the issues.
+
+### Risks and Mitigations
+
+- The golang bindings for libstoragemgmgt need some review and polish.
+  - Mitigation: The ODF(OpenShift Data Foundation) team will contribute to and help the RHEL team maintain libstoragemgmt golang
+
+## Design Details
+
+Each device will have a metric `storage_device_health` metric associated with it that has a value of 0,1, or 2. Where 0 is healthy state, 1 is a warning state, and 2 is a failed state. Each metric will be labeled by node, device-name, and device-id (where available).
+
+| storage_device_health value | Meaning |
+| --------------------------- | ------- |
+| 0                           | Healthy |
+| 1                           | Warning |
+| 2                           | Failed  |
+
+### Open Questions
+
+### Test Plan
+
+**Note:** *Section not required until targeted at a release.*
+
+- E2E tests will deploy the component and verify that metrics are being emitted per device
+
+### Graduation Criteria
+
+**Note:** *Section not required until targeted at a release.*
+
+**For non-optional features moving to GA, the graduation criteria must include
+end to end tests.**
+
+#### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+- Write symptoms-based alerts for the component(s)
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+
+#### Removing a deprecated feature
+
+Does not apply
+
+### Upgrade / Downgrade Strategy
+
+
+
+### Version Skew Strategy
+
+
+
+## Implementation History
+
+
+## Drawbacks
+
+Because writing files to be read by the textfile collector cannot be done on-demand in response to a request from prometheus, we will have to poll at our own frequency.
+However, the frequency can be made tunable.
+
+## Alternatives
+
+- Implement in Local Storage Operator:
+  - `LocalVolumeDisocveryResults` already enumerate and refresh information about the disk in a per-node `LocalVolumeDiscoveryResult`, and a [PR](https://github.com/openshift/local-storage-operator/pull/249) is in progress to add metrics to each of the daemonsets that update the CR.
+  - `LocalVolumeDiscovery` can be configured via `nodeSelector` and `toleration` to match any set of nodes or all nodes. We could have the metrics be for all nodes or just for the ones that LocalVolumeDiscovery is configured to watch. Whether this should be controlled by the same (singleton) CR is an [open question](#Open-Questions    )
+
+- We *could* create a new component for this.
+
+Disqualified alternatives:
+- We considered node-problem-detector, but it doesn't export metrics only sets node conditions and creates events.
+
+## Infrastructure Needed
+
+- New repository for collector image.
+- New bugzilla subcomponent with default assignee managed by ODF team.


enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,132 @@
...
+
+- As an admin, I would like to be alerted when a storage device is unhealthy and be linked to remediation steps.
+- As an admin, I would like if components (such as OpenShift Data Foundation) could access storage device health and failover faster.
+
+
+### Implementation Details/Notes/Constraints
+
+`LocalVolumeDisocveryResults` already enumerate and refresh information about the disk in a per-node `LocalVolumeDiscoveryResult`, and a [PR](https://github.com/openshift/local-storage-operator/pull/249) is in progress to add metrics to each of the daemonsets that update the CR.
+
+`LocalVolumeDiscovery` can be configured via `nodeSelector` and `toleration` to match any set of nodes or all nodes. We could have the metrics be for all nodes or just for the ones that LocalVolumeDiscovery is configured to watch. Whether this should be controlled by the same (singleton) CR is an [open question](#Open-Questions    )
```
sp98, line 68: are we planning to provide metrics for all the disks? Even those that are not used by LSO?

sp98 (changes requested):

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,132 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement proposes that we export health metrics for each local device on each node. 
+
+## Motivation
+
+Local device vendors expose health information about storage devices via APIs like [SMART](https://en.wikipedia.org/wiki/S.M.A.R.T.). It very useful to know when a storage is about to fail, to differentiate between storage relate errors vs purely software ones, etc.
```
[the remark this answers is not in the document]
mykaul:
```suggestion
Local device vendors expose health information about storage devices via APIs like [SMART](https://en.wikipedia.org/wiki/S.M.A.R.T.). It very useful to know when a storage is about to fail, to differentiate between storage hardware related errors vs purely software ones, etc.
```

enhancements/local-storage/health-monitoring.md
rohantmp: yes, I'll add that

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+## Drawbacks
+
+
+
+## Alternatives
+
+- We explored node_exporter, but it is a rootless container, so SMART data can't be checked.
+- We considered node-problem-detector, but it doesn't export metrics only sets node conditions and creates events.
```
fabiand, line 128: IMHO node-problem.detector is using cluster level data to detect problems, but not by itself export node level data

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,132 @@
...
+- We *could* create a new component for this.
```
fabiand, line 129: I'm missing some reasoning of why LSO is _not_ the right place.

While I agree that other components can own disks of a host, LSO is one of them, thus it looks like to be like a good starting point at least. The good is alos that LSO can correlate it to PVs

gnufied: In general argument against including these metrics in LSO is - LSO is not a general purpose disk management solution. It exists to allow creation of PVs from user selected disks. In other words - it only concerns itself with PV creation from selected disks and if task was to monitor health of PVs that LSO creates, it would be a suitable place to add health checks. 

In this case though - task is to monitor health of all disks, including root disk, boot disk etc and in which case - LSO does not seem like a suitable place to report such data.  There are some advantages of keeping LSO lightweight and easier to maintain and using it as a kitchen sink makes maintenance difficult.

gnufied:
@rohantmp can you also add some description about how metrics would look like?

enhancements/local-storage/health-monitoring.md
rohantmp: @gnufied I'm not sure I understand the consequences, the solution has to be maintained somewhere. Can you clarify the possible negative situations from putting it in LSO so I can put it in the enhancement?

enhancements/local-storage/health-monitoring.md
rohantmp: Agreed, just included it because it was brought up frequently and I've included why we chose not to use it. Should I put it in another section?

enhancements/local-storage/health-monitoring.md
sabose: If LSO creates PVs from local disks, I think it seems to be the place to monitor health of those disks as well. Like Rohan, I haven't quite understood the argument against adding this feature to LSO.

enhancements/local-storage/health-monitoring.md
gnufied: > IMHO node-problem.detector is using cluster level data to detect problems, but not by itself export node level data.

I do not understand this. node-problem-detector runs a daemon on each node and reports data/metrics from the nodes. I do not think it uses cluster level data.

gnufied:
We discussed this in Slack with Openshift architects and current consensus is to use node-exporter to emit smart data metrics. There is already an collector for this - https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/blob/master/smartmon.py and it is possible to make privileged collectors that read smart data and dump into a textfile, so as node exporter can export it. 

We generally want smart metrics to be available whenever possible and we do not want this to be an optional feature and hence node-exporter is considered best fit for this. It does however mean that - we can't blink LEDs of disks based on disk health, since such a feature is too far out of scope for node exporter.

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,137 @@
...
+## Drawbacks
+
+
+
+## Alternatives
+
+- We *could* create a new component for this.
+
+Disqualified alternatives:
+- We explored node_exporter, but it is a rootless container, so SMART data can't be checked.
```
dhellmann, line 133: Why can't node exporter be made to run as root?

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,137 @@
...
+
+This enhancement proposes that we export health metrics for each local device on each node.
+
+## Motivation
+
+Local device vendors expose health information about storage devices via APIs like [SMART](https://en.wikipedia.org/wiki/S.M.A.R.T.). It very useful to know when a storage is about to fail, to differentiate between storage hardware related errors vs purely software ones, etc.
+
+### Goals
+
+- Expose health information about local storage devices in a way that can generate alerts
```
dhellmann, line 45: Is the plan to expose everything or is there a subset of the data we know would be most useful?

enhancements/local-storage/health-monitoring.md
rohantmp: My understanding is that upstream won't go for it based on [this issue](https://github.com/prometheus/node_exporter/issues/782#issuecomment-356160910) and asking around on #forum-monitoring.

enhancements/local-storage/health-monitoring.md
rohantmp: Currently, the discussion has been around a simple boolean. Normal/Abnormal (lifting those terms from the CSI implementation). I'll add a more fleshed out design for the metrics ASAP

enhancements/local-storage/health-monitoring.md
dhellmann: Why can't OpenShift run it as root?

gnufied: Running this as privileged is not a problem in OCP. We are already running some collectors in privileged containers as part of same node-exporter pod.

dhellmann: If that's the case, why was this approach disqualified?

rohantmp: Hi @gnufied, would be useful if you could involve our team in this discussion. Can you link to the slack conversation?

rohantmp:
> we can't blink LEDs of disks based on disk health, since such a feature is too far out of scope for node exporter.

This does seem like a feature that is too important to preclude.

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Non-Goals
+
+- Align with sig-storage [health monitoring for persistent volumes](https://kubernetes.io/blog/2021/04/16/volume-health-monitoring-alpha-update/)
+  - justification: it is also useful to monitor system drives and CSI only deals with persitent volumes, and migrating to CSI is an extremely long term effort.
+
+## Proposal
```
jsafrane, line 52: I'd appreciate definition of "device heath" here (or below). It can be very different thing to different people.

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,137 @@
...
+
+
+### Implementation Details/Notes/Constraints
+
+`LocalVolumeDisocveryResults` already enumerate and refresh information about the disk in a per-node `LocalVolumeDiscoveryResult`, and a [PR](https://github.com/openshift/local-storage-operator/pull/249) is in progress to add metrics to each of the daemonsets that update the CR.
+
+`LocalVolumeDiscovery` can be configured via `nodeSelector` and `toleration` to match any set of nodes or all nodes. We could have the metrics be for all nodes or just for the ones that LocalVolumeDiscovery is configured to watch. Whether this should be controlled by the same (singleton) CR is an [open question](#Open-Questions    )
+
+We will be collecting metrics via [libstoragemgmt](https://github.com/libstorage/libstoragemgmt) and calling it via [libstoragemgmt-golang](https://github.com/libstorage/libstoragemgmt-golang/).
+If a block device is a partion or logical volume, we will only expose health for the parent device.
```
jsafrane, line 70: In case of LVM it can be array of devices. Whose health will be reported?

dhellmann:
> > we can't blink LEDs of disks based on disk health, since such a feature is too far out of scope for node exporter.
> 
> This does seem like a feature that is too important to preclude. Does this mean that we should create a new component for this?

Let's tackle one part of the problem at a time. If we focus this enhancement on collecting the data to let someone know when there is an issue, that data can be used by a tool that knows how to turn on the identification light.

jan--f:
Iiuc the current proposal is to add device health data as labels to existing block device metrics?
Exposing device health metrics on their own as opposed to adding them as labels to existing labels should be preferred. To correlate two metrics, say one about block devices and the one about device health, the Prometheus query language can be used to match on a common label, in this case the parent block device.

Either deployment model (using the `node_exporter`'s `textfile_collector` or using `libstoragemgmt` in a standalone exporter) should work. Afaik there were discussions about this before. The `node_exporter` route would require running the collector script in a sidecar container and someone to own and maintain that deployment, as well as the actual collector script.

cc @simonpasquier for visibility

dhellmann:
Since we're talking about using node-exporter, should this proposal be moved out of the local-storage folder and into the monitoring folder?

Who on the monitoring team will sign up to be the approver for the enhancement?

jan--f:
Monitoring is not the right place for this. We are take care of the Monitoring stack itself and provide the framework, not individual exporters or the respective alerts that use this framework. Using node_exporter in this context really only means using node_exporter to expose the metrics, The metrics themselves need to be created by a separate component. See also https://github.com/prometheus/node_exporter#textfile-collector
The interface to node_exporter is only the text file containing the metrics, that are to be exported. I think local-storage is the best location for this.

 I think @simonpasquier would be the first choice as an approver, but in his absence I'll happily cover that.

dhellmann:
OK, it's not clear to me which team manages node-exporter itself and its configuration within OpenShift. I assume that team would want to be involved in the discussion here. Is that the monitoring team? Someone else?

jan--f:
CMO manages and deploys `node_exporter`, so that is the monitoring teams responsibility indeed. without this the Monitoring stack and an OpenShift deployment won't function properly.
However a SMART exporter based on the textfile collector interface is only loosely coupled to the `node_exporter`. Any file with the suffix `.prom` that is present in the configured directory ([/var/node_exporter/textfile](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/node-exporter/daemonset.yaml#L37)) and contains the correct syntax will be exported by `node_exporter`. How those files get created and updated is out of scope for `node_exporter`. We do run some exporter like that but only once in an init container.

A SMART exporter that run run regularly needs to be run in its own container. Development and maintenance of such optional addition is something the Monitoring team can't provide.

dhellmann:
@jan--f so the idea is that anything can drop files in that path on the node and node-exporter will pick them up? If I understand correctly, I think it's reasonable to say that decoupling means the storage team can do the work.

jan--f:
> @jan--f so the idea is that anything can drop files in that path on the node and node-exporter will pick them up? If I understand correctly, I think it's reasonable to say that decoupling means the storage team can do the work.

Yes that's correct :+1:

rohantmp:
@dhellmann is this something we want on every OCP cluster? If we created another operator for this it would be optional right?

Or do we add it to the OCP manifests somehow?

mykaul:
> @dhellmann is this something we want on every OCP cluster? If we created another operator for this it would be optional right?

On every bare-metal cluster, it makes sense.

dhellmann:
> > @dhellmann is this something we want on every OCP cluster? If we created another operator for this it would be optional right?
> 
> On every bare-metal cluster, it makes sense.

Right, we've had a request to add this to bare metal clusters for a while now. It would also be useful if it could be enabled when the platform is "none" because a lot of those clusters are also really bare metal. It probably doesn't make sense to enable it on cloud platforms.

Does the LSO already look at the infrastructure config resource to make any other decisions?

rohantmp:
LSO doesn't currently look at the infra config. We do this in ODF/OCS though so I'm familiar. 
The issue I currently see is that LSO is something the user has to manually install.

What would be a good approach for including it automatically based on install config?

rohantmp:
[The docs](https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/operators.md) say that to add something to CVO it has to be a Cluster Operator. Seems like the right place for this might be in the monitoring operator? We (ODF/OCS team) would be glad to make the contribution. 

@jan--f Would it be a good candidate for a `node_exporter` sidecar or another daemonset deployed by CMO?

rohantmp:
@jan--f ^^

jan--f:
@rohantmp As noted in https://github.com/openshift/enhancements/pull/838#issuecomment-893426723 CMO is responsible for deploying and maintaining the monitoring framework, not for _optional_ exporters and their relevant artifacts.

A smart exporter (even if it only creates the `.prom` file that `node_exporter` exposes) needs to know when to run, which hardware to query and how to query it. For CMO that strikes me as out of scope, we don't really have any knowledge about the infrastructure and hardware.
Would ODF/OCS not be the right place for this? Iiuc it already deploys optional exporters that integrate with Monitoring. Would this be an option?

rohantmp:
@jan--f LSO might be a better place, but it's still an optional operator. I'm looking for a way to make this automatically deploy if the platform type is `None` or `Baremetal`  as @dhellmann and @mykaul indicated. The only way I have found to achieve that so far is via a ClusterOperator in the release manifest and I don't think this qualifies for a new one. Is there another way?

dhellmann:
> > not for optional exporters and their relevant artifacts.
> 
> @jan--f I don't think this would be optional exactly. I'm looking for a way to make this automatically deploy if the platform type is `None` or `Baremetal` as @dhellmann and @mykaul indicated.
> 
> LSO(Local Storage Operator) might be a better place (ODF is a layer above LSO and a paid product), but it's still an optional operator. The only way I have found to achieve that so far is via a ClusterOperator in the release manifest and I don't think this qualifies for a new one. Is there another way?

We have a couple of operators (both in the release payload and delivered via OLM) that look at the platform type in the infrastructure configuration resource to make decisions about what to deploy.

simonpasquier:
I'd agree that it'd be better if the full solution (exporter + deployment) is owned by a single team. But I think that using node_exporter would be still possible if no other option is practical.

To reduce the implications for the monitoring team, the sidecar program reading disk statistics could detect the platform type and decide whether or not it should dump metrics to the text file by itself. It would mean that the sidecar is always running (but consuming very little resources if the platform isn't `Baremetal` or `None`). @dhellmann suggested that since CMO already [reads](https://github.com/openshift/cluster-monitoring-operator/blob/e7e6b8c0a4dbb0f53a99d9cd3d5f2a54422fd90b/pkg/operator/operator.go#L587-L608) the infrastructure config for other purposes, it could pass the platform type to the sidecar via an environment variable (a program's argument would also work) to avoid the program hitting the API itself. The other alternative would be for CMO to adjust the daemonset definition (adding or removing the sidecar accordingly) but it would require more work for the monitoring team.

Existing text collectors live https://github.com/openshift/node_exporter/tree/master/text_collectors and we can have a dedicated directory + OWNERS list for this additional collector redirecting to whoever is responsible for it. My assumption is that once it gets plumbed into CMO (e.g. added to the daemonset definition), the monitoring team wouldn't have much maintenance work.

rohantmp:
@simonpasquier , I think that this solution meets all requirements, thanks. I'll update the proposal to reflect this as the way to go forward and add a comment when I'm done.

simonpasquier:
@rohantmp from a monitoring team's perspective, I'd like the updated proposal to clarify the ownership and areas of responsibility. For example:
* which component ships the alerts derived from the collected metrics? If it ends up being the cluster monitoring operator, we need to clarify who's going to own that code.
* which team (e.g. BZ component) will be responsible for triaging and fixing bugs?

cc @openshift/openshift-team-monitoring for other inputs.

Can you also summarize the discussions that happened over Slack regarding local-storage-operator vs. node_exporter?

rohantmp:
>  -   which component ships the alerts derived from the collected metrics? If it ends up being the cluster monitoring operator, we need to clarify who's going to own that code.
>  -   which team (e.g. BZ component) will be responsible for triaging and fixing bugs?


@simonpasquier updated with a new "Ownership and Maintenance" section :)

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,142 @@
...
+### Version Skew Strategy
+
+
+
+## Implementation History
+
+
+## Drawbacks
+
+
```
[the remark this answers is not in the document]
rohantmp: How do we know when prometheus scrapes if we're going for textfile?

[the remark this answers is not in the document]
dhellmann: If I remember correctly, node_exporter reads the file when Prometheus scrapes, but it is up to the external data collection tool to create the file and keep it up to date -- there is no trigger to run the app when scraping occurs.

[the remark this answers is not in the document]
rohantmp: So we would be polling right? I'll try to gather some data about CPU utilization. We could tune the frequency of polling.

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,142 @@
...
+- A new bugzilla subcomponent will be created under openshift-monitoring, but the default assignee will be ODF and ODF will be responsible for triaging the issues.
+
+### Risks and Mitigations
+
+- The golang bindings for libstoragemgmgt need some review and polish.
+  - Mitigation: The ODF(OpenShift Data Foundation) team will contribute to and help the RHEL team maintain libstoragemgmt golang
+
+## Design Details
+
+Each device will have a metric `storage_device_health` metric associated with it that has a value of 0,1, or 2. Where 0 is healthy state, 1 is a warning state, and 2 is a failed state. Each metric will be labeled by node, device-name, and device-id (where available).
```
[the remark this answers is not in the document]
dhellmann: Thanks!

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,142 @@
...
+
```
dhellmann, line 85: When we've discussed adding other hardware-related metrics in the past there was some concern about the cardinality of the data and how much new data would need to be processed and stored by Prometheus. Do we have those sorts of concerns for this data?

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,142 @@
...
+  - `LocalVolumeDisocveryResults` already enumerate and refresh information about the disk in a per-node `LocalVolumeDiscoveryResult`, and a [PR](https://github.com/openshift/local-storage-operator/pull/249) is in progress to add metrics to each of the daemonsets that update the CR.
+  - `LocalVolumeDiscovery` can be configured via `nodeSelector` and `toleration` to match any set of nodes or all nodes. We could have the metrics be for all nodes or just for the ones that LocalVolumeDiscovery is configured to watch. Whether this should be controlled by the same (singleton) CR is an [open question](#Open-Questions    )
+
+- We *could* create a new component for this.
+
+Disqualified alternatives:
+- We considered node-problem-detector, but it doesn't export metrics only sets node conditions and creates events.
+
+## Infrastructure Needed
+
```
dhellmann, line 142: We said in the meeting earlier today that we wanted a new repository to be created to hold the reporter image configuration. If you include that detail here, we can work out a good name for it and then the request to DPP for the new repo will be easier to approve.

enhancements/local-storage/health-monitoring.md
rohantmp: I guess this was for when we were considering per-PV metrics. LVM volumes count as a logical volume. We wouldn't expose metrics for them, just their parents.

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,142 @@
...
+After that, the maintainence effort will mostly centered on the code the image runs to get health and update its `.prom` file. The logic for this collector can be in a subdirectory of CMO and have it's own OWNERS file allowing the ODF team to own it.
+
+We will be collecting metrics via [libstoragemgmt](https://github.com/libstorage/libstoragemgmt) and calling it via [libstoragemgmt-golang](https://github.com/libstorage/libstoragemgmt-golang/).
+If a block device is a partion or logical volume, we will only expose health for the parent device.
+
+
+### Ownership and Maintenance
+
+- The `node_exporter` daemonset creation is owned by the Openshift-Monitoring team. The PR for modifying the daemonset based on infrastructure to include the sidecar can be contributed by ODF, but Monitoring team will have to review.
+- The sidecar image and logic will live in the Cluster Monitoring Operator repo, but maintenance and ownership will be with the ODF team via a OWNERS file.
```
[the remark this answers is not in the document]
rohantmp: updated

enhancements/local-storage/health-monitoring.md
rohantmp: There are currently plans to add one new metric per storage device.

enhancements/local-storage/health-monitoring.md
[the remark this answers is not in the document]
rohantmp: Commented this under drawbacks.

enhancements/local-storage/health-monitoring.md
simonpasquier: It would be good to provide an estimation of how many timeseries would be exposed overall (e.g. `for a node with N physical storage devices, the exporter would expose (number of metrics) * (number of storage device) = 1*N = N timeseries`).

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,155 @@
...
+| 1                           | Warning |
+| 2                           | Failed  |
+
+### Open Questions
+
+### Test Plan
+
+**Note:** *Section not required until targeted at a release.*
+
+- E2E tests will deploy the component and verify that metrics are being emitted per device
```
simonpasquier, line 99: So I would expect that we have a job running on baremetal and this job should run in the yet-to-be-defined exporter repository as well as github.com/openshift/cluster-monitoring-operator?

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,155 @@
...
+### Version Skew Strategy
+
+
+
+## Implementation History
+
+
+## Drawbacks
+
+Because writing files to be read by the textfile collector cannot be done on-demand in response to a request from prometheus, we will have to poll at our own frequency.
```
simonpasquier, line 138: We could also have the future exporter behave as a regular exporter with a /metrics endpoint.

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,155 @@
...
+
+The Cluster Monitoring Operator will deploy the sidecar when the platform is Baremetal or None by modifying the `node_exporter` [daemonset](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/node-exporter/daemonset.yaml). The platform will be detected by querying the `infrastructure` resource.
+
+After that, the maintainance effort will mostly centered on the code the image runs to get health and update its `.prom` file. The logic for this collector can be in a subdirectory of CMO and have it's own OWNERS file allowing the ODF team to own it.
+
+We will be collecting metrics via [libstoragemgmt](https://github.com/libstorage/libstoragemgmt) and calling it via [libstoragemgmt-golang](https://github.com/libstorage/libstoragemgmt-golang/).
+If a block device is a partion or logical volume, we will only expose health for the parent device.
+
+
+### Ownership and Maintenance
```
simonpasquier, line 72: We need also to mention the service monitor (if we don't go the textfile collector's route) and the alerting rules. My idea is that they live in the cluster-monitoring-operator (because they need an operator to be deployed) but owned by the ODF team (again it should be feasible to put them in a dedicated directory with separate OWNERS file).

simonpasquier (commented):
Something that isn't yet clear to me is whether the collector is a textfile collector (with node_exporter serving the text file) or a regular Prometheus exporter (e.g. with a `/metrics` HTTP endpoint). The first approach is probably easier to implement but it introduces a delay in metrics collection. It might be also harder to detect when something goes wrong in the collector (stale metrics).

enhancements/local-storage/health-monitoring.md
jan--f: Not sure if this is the right place, but should this proposal also mention how this health status is derived? Will the code look at SMART attributes or the result of SMART test runs or something else entirely?

If this would rely on SMART's testing feature for example, these tests need to be scheduled as well.

I think the implementation choice could inform whether using the textfile collector interface or a stand-alone exporter is preferable. Say, if the health status is determined based on SMART tests, a standalone exporter, that queries the device synchronously might be overly resource intensive, as running those tests is typically done fairly rarely.

dhellmann:
> Something that isn't yet clear to me is whether the collector is a textfile collector (with node_exporter serving the text file) or a regular Prometheus exporter (e.g. with a `/metrics` HTTP endpoint). The first approach is probably easier to implement but it introduces a delay in metrics collection. It might be also harder to detect when something goes wrong in the collector (stale metrics).

Line 61 talks about using a text file collector. I think as a first version, that approach makes a lot of sense because it should be quick to deliver. Your points about noticing when the collector dies make sense, but I don't know how big of a risk that is.

Is it reasonable to start with a sidecar and see how that goes, and build a more robust solution if we need to?

pcuzner:
Thanks @rohantmp for looping me in.

There's a couple of things that confuse me with this proposal that primarily relate to the user interaction.

- a disk health alert is divorced from the PV right? If so how does an admin determine the impact of a warn event to applications on the platform? 
- how are you identifying the device in the metric...presumably not by /dev/sdb?
- how will this approach work with the UI? For example, the alert isn't the only interaction - the UI should be able to show the devices right..and if so, the health and characteristics of the device should be presented - so instead of examining the attrributes of a CR that represents a device the UI needs to run prometheus queries every n secs?
- I've seen in here that the LED interactions are out of scope - which given the strategy to offload device health to a prometheus exporter is understandable. But I'm confused as to the manual workflow that follows an alert - it would be great if the proposal included the workflow that it's going to deliver to the customer
- the example metric is only covering health for a device - so do we already cover the other attributes that libstoragemgmt provides somewhere else, for example hdd or flash; transport type:sas, sata, pcie; device rpm. Users already expect this kind of information, the vmware UI is a great example that customers likely measure other solutions by (I know I would!)

jan--f:
> Is it reasonable to start with a sidecar and see how that goes, and build a more robust solution if we need to?

I think so yes. One thing we could think about is to add a run counter as a metric (I suppose the unix time would work). If such a metric doesn't change in too long we'd have a signal that the actual job isn't updating the metrics file. The signal would have to be somewhat course, say "hasn't changed in the last 24h" or so depending on the frequency of the scrape job.

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Implementation Details/Notes/Constraints
+
+Implement a sidecar in the [node_exporter](https://github.com/openshift/node_exporter) component under [Cluster Monitoring Operator](https://github.com/openshift/cluster-monitoring-operator) that will scrape the metrics from the node and export them via the [textfile exporter](https://github.com/openshift/node_exporter/tree/master/text_collectors) (an exporter that scans `.prom` files in the
+`/var/node_exporter/textfile` directory and exports them.
+
+The Cluster Monitoring Operator will deploy the sidecar when the platform is Baremetal or None by modifying the `node_exporter` [daemonset](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/node-exporter/daemonset.yaml). The platform will be detected by querying the `infrastructure` resource.
+
+After that, the maintainance effort will mostly centered on the code the image runs to get health and update its `.prom` file. The logic for this collector can be in a subdirectory of CMO and have it's own OWNERS file allowing the ODF team to own it.
```
jan--f, line 66: One implementation detail we might want to call out specifically here that with text collectors its good practice to update the `.prom` file atomically in order to avoid races between the collector and the node_exporter accessing the file.

This can be as simple as this:
```
./exporter_script > metrics.prom.new
mv metrics.prom.new metrics.prom
```

jan--f:
> Thanks @rohantmp for looping me in.
> 
> There's a couple of things that confuse me with this proposal that primarily relate to the user interaction.
> 
> * a disk health alert is divorced from the PV right? If so how does an admin determine the impact of a warn event to applications on the platform?

There was a discussion about that, but it seems it was lost in an update. Iirc there is already a metric about PVs already that has a device label. Prometheus can match this PV metric with the device health metric.

> * how are you identifying the device in the metric...presumably not by /dev/sdb?

I believe that is the plan and corresponds to the aforementioned PV metric.

> * how will this approach work with the UI? For example, the alert isn't the only interaction - the UI should be able to show the devices right..and if so, the health and characteristics of the device should be presented - so instead of examining the attrributes of a CR that represents a device the UI needs to run prometheus queries every n secs?

This would certainly work, though it might be worth adding a recording rule. However this proposal wouldn't require a change in what works today. I do think UI interaction could be considered out of scope for this enhancement proposal.

> * I've seen in here that the LED interactions are out of scope - which given the strategy to offload device health to a prometheus exporter is understandable. But I'm confused as to the manual workflow that follows an alert - it would be great if the proposal included the workflow that it's going to deliver to the customer
> * the example metric is only covering health for a device - so do we already cover the other attributes that libstoragemgmt provides somewhere else, for example hdd or flash; transport type:sas, sata, pcie; device rpm. Users already expect this kind of information, the vmware UI is a great example that customers likely measure other solutions by (I know I would!)

I would propose both the above points are out of scope for this enhancement proposal, purely going by $title.

pcuzner:
> > * a disk health alert is divorced from the PV right? If so how does an admin determine the impact of a warn event to applications on the platform?
> 
> There was a discussion about that, but it seems it was lost in an update. Iirc there is already a metric about PVs already that has a device label. Prometheus can match this PV metric with the device health metric.
> 
> > * how are you identifying the device in the metric...presumably not by /dev/sdb?
> 
> I believe that is the plan and corresponds to the aforementioned PV metric.
> 
Thanks @jan--f
@rohantmp can this be added back into the design. Obviously /dev/sdb would be bad so it would be great for the design to describe how the metric can be utilised to help the end user understand the impact of the disk health issues. And since the scope of this effort centers around a metric, presumably the kubeadmin user can't actually see device health from the cli?

rohantmp:
Just back today after moving to a new city, catching up on everything :)

rohantmp:
Fiddling with a manual PoC to confirm that the correlation will go smoothly. Will add back how the UI could consume it after that.

rohantmp:
We've run into a problem. The HP vendor tool that allows us to read the health data from the RAID adapter (even if the devices are being used in passthrough mode) has a license that forbids distribution. The libstoragemgmt support for these vendors are dependent on the vendor tools, and from what I understand in RHEL the documented way to do this was for the user to provide the binary shipped by the vendor.

I'm currently trying to think of a workflow where something similar can be done in Openshift where the user can link to the package in a configmap and the plugin can download it or build an image with that package? Still sounds very hacky, I would love suggestions.

This will also impact the design as it would require user input. So the default monitoring solution can work for normal generic HBAs, but we would need user intervention for things like HPE smart arrays.

dhellmann:
@rohantmp would it be useful to deliver a version that doesn't require the extra tools, and then add support for that in a later release?

dhellmann:
We had some discussion about the HP question in slack, so I want to capture the outcome here so it can be added to the doc.

It is correct that we cannot deliver third-party tools with licenses that don't allow us to rebuild or repackage them.

Because node-exporter will run the container with elevated privileges, introducing a way for users to specify an arbitrary image to run would bypass the normal RBAC that protects the cluster from pods running with those privileges. Any user would could configure the image would be able to run their workload with elevated privileges on every node.

The primary alternative is to have someone create and support an operator that can include images containing the third-party tools. That could be done by the hardware vendor themselves, an integrator, or someone else. We could offer a "base" operator implementation to which someone could add their own image, but it's not clear there is a lot of value in investing in that versus showing them the operator SDK.

Cluster metric collection is restricted to openshift-* namespaces. We don't normally want other operators deployed into those namespaces, because it introduces an opportunity for something to overwhelm the cluster monitoring stack with too much metric data. For the same reason, we wouldn't want the third-party monitoring to inject data through node-exporter's text file input mode using something like a shared volume.

Users could enable workload monitoring and use it to collect data from the third-party operator, but that may introduce unwanted overhead to the cluster.

That leaves us with:

1. This enhancement should not worry about hardware that requires third-party tools that we cannot distribute.
2. We may want to provide some documentation or a blog post describing how to create an operator that does collect data using those third-party tools.
3. There is an open question about how we would want to support reporting from a third-party operator that would need to be solved before we try to do 2.

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,155 @@
...
+- A new bugzilla subcomponent will be created under openshift-monitoring, but the default assignee will be ODF and ODF will be responsible for triaging the issues.
+
+### Risks and Mitigations
+
+- The golang bindings for libstoragemgmgt need some review and polish.
+  - Mitigation: The ODF(OpenShift Data Foundation) team will contribute to and help the RHEL team maintain libstoragemgmt golang
+
+## Design Details
+
+Each device will have a metric `storage_device_health` metric associated with it that has a value of 0,1, or 2. Where 0 is healthy state, 1 is a warning state, and 2 is a failed state. Each metric will be labeled by node, device-name, and device-id (where available).
```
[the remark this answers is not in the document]
jan--f: Seconded, it would be great to have some details regarding this. This could simply be smarts device health property or be based on various smart counters. How the later is calculated is an interesting detail.

enhancements/local-storage/health-monitoring.md
```diff
@@ -0,0 +1,155 @@
...
+`/var/node_exporter/textfile` directory and exports them.
+
+The Cluster Monitoring Operator will deploy the sidecar when the platform is Baremetal or None by modifying the `node_exporter` [daemonset](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/node-exporter/daemonset.yaml). The platform will be detected by querying the `infrastructure` resource.
+
+After that, the maintainance effort will mostly centered on the code the image runs to get health and update its `.prom` file. The logic for this collector can be in a subdirectory of CMO and have it's own OWNERS file allowing the ODF team to own it.
+
+We will be collecting metrics via [libstoragemgmt](https://github.com/libstorage/libstoragemgmt) and calling it via [libstoragemgmt-golang](https://github.com/libstorage/libstoragemgmt-golang/).
```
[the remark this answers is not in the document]
jan--f: Afair Ceph's smart metric collection is part of the OSD code, so it seems unlikely this can be made into a stand-alone tool.

jan--f:
> > We've run into a problem. The HP vendor tool that allows us to read the health data from the RAID adapter (even if the devices are being used in passthrough mode) has a license that forbids distribution. The libstoragemgmt support for these vendors are dependent on the vendor tools, and from what I understand in RHEL the documented way to do this was for the user to provide the binary shipped by the vendor.
> 
> Would that also be the case if we use regular smartctl via the existing scripts instead of libstoragemgmt?

I believe it would. The issue is that the storage controller itself doesn't implement the SMART interface or allows to access the disks attached to it through it.

PR #841, opened 2021-07-20
Remove Jenkins from the OCP Payload

adambkaplan:
Provisional proposal to remove Jenkins images from the OpenShift
payload. This would allow Jenkins primary and agent images to be
delivered on their own cadence, just like any other on-top application.

### DIFF
--- a/enhancements/builds/remove-jenkins-payload.md
+++ b/enhancements/builds/remove-jenkins-payload.md
@@ -1,0 +1,239 @@
+---
+title: remove-jenkins-payload
+authors:
+  - "@adambkaplan"
+reviewers:
+  - "@gabemontero"
+  - "@akram"
+approvers:
+  - "@bparees"
+  - "@sbose78"
+  - "@derekwaynecarr"
+creation-date: 2021-07-20
+last-updated: 2021-08-24
+status: implementable
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Remove Jenkins from the OCP Payload
+
+## Release Signoff Checklist
+
+- [ x ] Enhancement is `implementable`
+- [ x ] Design details are appropriately documented from clear requirements
+- [ x ] Test plan is defined
+- [ x ] Operational readiness criteria is defined
+- [ x ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This proposal aims to remove Jenkins images from the OCP payload and replace them with images delivered via CPaaS.
+
+## Motivation
+
+Distributing and supporting Jenkins directly has always been challenging for OpenShift.
+Due to OpenShift's tight coupling to Jenkins via the Jenkins Pipeline build strategy,
+OpenShift 4 has included a Jenkins distribution as well as a default set of agent images used to run Jenkins pipelines.
+This meant that any bug fixes or upgrades to Jenkins needed to go through the often arduous backport process to get delivered to customers.
+
+Jenkins itself is not native to Kubernetes and does not use a [Java runtime](https://www.jenkins.io/doc/administration/requirements/java/) that has been optimized for cloud-native environments (i.e. Quarkus).
+As a result, it has poor resilience when run on Kubernetes.
+The Jenkins templates do not run Jenkins in a highly available mode, leaving its primary instance vulnerable to downtime from slow restarts.
+The Jenkins container image is also very large - it is in fact the largest image in the OpenShift payload today.
+Removing Jenkins from the OpenShift payload has been a longstanding goal that can reduce the time needed to pull the OCP payload at install/updgrade time.
+
+With the GA release of OpenShift Pipelines, we would like to decouple Jenkins from OpenShift, distributing Jenkins just like any other application.
+This would allow bug fixes in our Jenkins distribution to be released on its own cadence.
+Customers can continue to deploy Jenkins on OpenShift via its ImageStream and associated templates.
+In this fashion, Jenkins becomes just another application on OpenShift and does not retain any preferential status.
+The maintenance of Jenkins is also simplified - only one image version will be released and tested against supported OCP versions.
+
+### Goals
+
+* Move Jenkins out of the OCP Payload.
+
+### Non-Goals
+
+* Document how JenkinsPipeline strategy builds can be migrated off of OpenShift.
+* Migrate Jenkins users to OpenShift Pipelines/Tekton.
+* Stop distributing the Jenkins image and our associated OpenShift plugins.
+* Disable the JenkinsPipeline build strategy.
+
+## Proposal
+
+### User Stories
+
+As a developer using Jenkins on OpenShift
+I want to continue using my Jenkinsfiles
+So that I can continue to use Jenkins for my CI/CD processes
+
+As an OpenShift release engineer
+I want to remove Jenkins from the OCP payload
+So that I can reduce the size of the payload
+And release Jenkins fixes on their own cadence
+
+### Implementation Details/Notes/Constraints [optional]
+
+#### Remove From Payload
+
+Jenkins and its related agent images will be removed from the OCP payload.
+The delivery of these container images will be replaced with a CPaaS-based release pipeline.
+Jenkins images will be published to the Red Hat Container Catalog, just like any other layered product.
+The current [openshif4/ose-jenkins](https://catalog.redhat.com/software/containers/openshift4/ose-jenkins/) image will need to be deprecated in favor of the new non-payload image.
+
+Once the CPaaS pipeline is ready, the Jenkins templates and imagestreams delivered in OpenShift via Samples operator will need to reference the container catalog images (just like other layered products).
+The Samples Operator can then remove its special logic for Jenkins imagestreams and templates.
+
+#### Jenkins / OCP Compatibility
+
+The label `io.openshift.versions` will be added to the Jenkins image via a Dockerfile LABEL instruction.
+Its value will be an inclusive range of OCP major+minor versions that the image is officially supported on.
+For example, the current Jenkins image would set the value `v4.6-v4.8`.
+
+In the event a breaking change is introduced, the `ocp-*` backport branch will maintain the Dockerfile that sets the appropriate OCP supported version label.
+The main branch will then update the `io.openshift.versions` label to indicate that prior OCP versions are no longer supported.
+For example, if a breaking change is introduced in OCP 4.9, the branch `ocp-4.8` will be added to all Jenkins repositories.
+The main branch will then be update the `io.openshift.versions` label on the image to `v4.9`, while the image produced from the `ocp-4.8` branch will continue to use `v4.6-v4.8`.
+
+#### Jenkins Plugin Version Scheme
+
+Jenkins plugins for OpenShift currently use the 1.0.<z> versioning scheme.
+This versioning scheme will be updated so that the minor Y version corresponds to the lowest supported version of OCP 4 for the plugin.
+For example, the next release of the openshift-client plugin should be 1.6.0, since OCP 4.6 is the lowest supported version of OCP for the plugin (that we can support).
+
+The main/master branch for Jenkins and its plugins will be assumed to support the version of OpenShift under development.
+If a breaking change is introduced, a separate branch named `ocp-4.n` should be created to handle bug backports, with `n` representing the highest supported version of OpenShift.
+The requisite delivery pipelines will also need to be updated so that bugfix builds can be issued from `ocp-*` branches.
+
+#### Jenkins Image Tagging Scheme
+
+The Jenkins image will use the supported OCP version as the basis of its tagging scheme:
+
+1. `latest` - latest image build
+2. `4` - latest image that supports OCP 4
+3. `4.<Y>`- latest image that supports up to OCP 4.Y
+4. `4.<Y>-N` - latest image that supports up to OCP 4.Y, with incrementing ordinal `N`.
+
+CPaaS and Errata will be configured so that tags for `4.Y` are set appropriately for the given branch, based on the `io.openshift.versions` Dockerfile label above.
+For example, if the Dockerfile has the `v4.6-v4.9` version label, the image produced by CPaaS will have the following tags:
+
+- `latest`
+- `4`
+- `4.6`, `4.7`, `4.8`, and `4.9`
+- `4.6-N`, `4.7-N`, `4.8-N`, `4.9-N` (`N` being an increasing ordinal from brew)
+
+The Jenkins imagestreams delivered by the Samples operator will be updated to use this new tagging scheme, with tags added for `latest`, `4`, and the current `4.<Y>` version.
+The existing imagestream tag `2` will also be maintained as long as Jenkins is based on Jenkins v2.
+
+#### Web Console
+
+User will continue to be able to deploy Jenkins on OpenShift by instantiating our provided Templates.
+As long as the Jenkins Pipeline strategy can be run on a fully supported cluster (i.e upgrades are allowed), existing console integrations with Jenkins should remain in place.
+
+#### Telemetry / Operational Readiness
+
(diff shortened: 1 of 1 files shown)


enhancements/builds/remove-jenkins-playload.md
```diff
@@ -0,0 +1,195 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This proposal aims to remove Jenkins images from the OCP payload and replace them with images delivered via CPaaS.
+
+## Motivation
+
+Distrbuting and supporting Jenkins directly has always been challenging for OpenShift.
```
bparees, line 37:
```suggestion
Distributing and supporting Jenkins directly has always been challenging for OpenShift.
```

enhancements/builds/remove-jenkins-playload.md
```diff
@@ -0,0 +1,195 @@
...
+As a result, it has poor resilience when run on Kubernetes.
+The Jenkins templates do not run Jenkins in a highly available mode, leaving its primary instance vulnerable to downtime from slow restarts.
+The Jenkins container image is also very large - it is in fact the largest image in the OpenShift payload today.
+Removing Jenkins from the OpenShift payload has been a longstanding goal that can reduce the time needed to pull the OCP payload at install/updgrade time.
+
+With the GA release of OpenShift Pipelines, we would like to decouple Jenkins from OpenShift, distributing Jenkins just like any other application.
+This would allow bug fixes in our Jenkins distribution to be released on its own cadence.
+Customers can continue to deploy Jenkins on OpenShift via its ImageStream and associated templates.
+In this fashion, Jenkins becomes just another application on OpenShift and does not retain any preferential status.
+The maintentance of Jenkins is also simplified - only one image version will be released and tested against supported OCP versions.
```
bparees, line 52:
```suggestion
The maintenance of Jenkins is also simplified - only one image version will be released and tested against supported OCP versions.
```

bparees: we tried to ship only a single jenkins image for all ocp versions back in the 3.x days.  It eventually bit us, hard, due to api changes.  I would strongly advise against trying to go down that road again.  I think you're going to need to ship/maintain a jenkins version tag per ocp version.

specifically, when every customer is using the "latest" tag, and you need to make an incompatible change (one that is necessary to support 4.12 for example, but breaks 4.8) you can't do it, unless you can find a way to make the change be ocp version aware and do the right thing in both cases, which may be incredibly difficult particularly if the issue is in some jenkins plugin code or something.

enhancements/builds/remove-jenkins-playload.md
```diff
@@ -0,0 +1,195 @@
...
+Imagestream imports should not be impacted by version skew - imagestreams delivered by the Samples
+operator are not updated until the new version of the Samples operator is deployed and wins leader election.
+
+## Implementation History
+
+2021-07-20: Provisional enhancement draft
+
+## Drawbacks
+
+Removing Jenkins from the payload will help with the mainenance burden, because Jenkins fixes will not need to be backported to each OCP release.
```
bparees, line 174: per my comments above, you may still need to maintain an image per OCP version, but at least you'll have more flexibility in doing the backports and shipping it, outside the OCP cadence.

enhancements/builds/remove-jenkins-playload.md
```diff
@@ -0,0 +1,195 @@
...
+However, as long as the Jenkins pipelie strategy exists, the sync plugin will need to be maintained and updated.
```
bparees, line 175:
```suggestion
However, as long as the Jenkins pipeline strategy exists, the sync plugin will need to be maintained and updated.
```

enhancements/builds/remove-jenkins-playload.md
```diff
@@ -0,0 +1,195 @@
...
+
+#### Removing a deprecated feature
+
+The Jenkins pipeline strategy has been officially deprecated since OCP 4.1.
+This proposal does not permanently disable the strategy - rather it changes how we deliver an on-top application.
+Documentation will be updated to reflect areas where customers are impacted (mainly for network restricted environments).
+
+### Upgrade / Downgrade Strategy
+
+On upgrade, the Jenkins imagestream will be updated to reflect the new non-payload image by the Samples operator.
```
bparees, line 159: how does a cluster get a new jenkins image when they are not upgrading OCP itself?  Will the imagestream be configured for scheduled import so that it picks up changes automatically?    (that may be undesirable for some customers, so really it probably needs to be a choice)

(even when upgrading OCP, nothing will cause a re-import unless the imagestreamtag name that's being imported changes)

enhancements/builds/remove-jenkins-playload.md
bparees: At a minimum you should probably publish multiple tags even if they use the same content for now.  And ensure that OCP clusters import the tag that's aligned to their OCP version.  That way if you *do* need to break compatibility in the future, you can just branch and start publishing 2 different images at that point.

enhancements/builds/remove-jenkins-playload.md
adambkaplan: @sbose78 does our Helm toolchain provide a means to upgrade a chart and update existing chart deployments, similar to what we have for OperatorHub? If so, this might be the best path forward (and would require us to deliver as a Helm chart instead of a template)

bparees: i'm not sure how that would solve the problem of "new jenkins image pushed to the 4.9 tag, 4.9 clusters pick up the new image sha"?

sbose78: > does our Helm toolchain provide a means to upgrade a chart and update existing chart deployments, similar to what we have for OperatorHub?

Yes, it does. 

The chart authors would typically make a submission for the new version of the software, example, https://github.com/openshift-helm-charts/charts/tree/main/charts/partners/hashicorp/vault . Our UI/CLI experiences would pick this up and prompt users to upgrade if a newer version is available.

enhancements/builds/remove-jenkins-playload.md
```diff
@@ -0,0 +1,195 @@
...
+This proposal aims to remove Jenkins images from the OCP payload and replace them with images delivered via CPaaS.
+
+## Motivation
+
+Distributing and supporting Jenkins directly has always been challenging for OpenShift.
+Due to OpenShift's tight coupling to Jenkins via the Jenkins Pipeline build strategy,
+OpenShift 4 has included a Jenkins distribution as well as a default set of agent images used to run Jenkins pipelines.
+This meant that any bug fixes or upgrades to Jenkins needed to go through the often arduous backport process to get delivered to customers.
+
+Jenkins itself is not native to Kubernetes and does not use a Java runtime that has been optimized for cloud-native environments (i.e. Quarkus).
```
gabemontero, line 42: if you'd like to include some sort of reference to Jenkins and Quarkus @adambkaplan, there minimally is https://www.jenkins.io/sigs/platform/

gabemontero (commented):
just a minor offering of an additional reference @adambkaplan

enhancements/builds/remove-jenkins-playload.md
adambkaplan: @sbose78 will our Helm chart repo, like our OperatorHub repo, live in an index image that gets updated regularly?

enhancements/builds/remove-jenkins-playload.md
adambkaplan: Spoke with @akram on this topic - we are already comfortable adding backwards compatibility to the Jenkins plugins. For instance, some of oauth plugin bits have code that supports 4.2 and earlier versions. Our current practice is to keep the same plugin versions across all supported releases.

I think this proposal will need to identify a versioning scheme so that if we do need to break backwards compatibility, we can do so safely.

adambkaplan: @bparees doesn't every operator have this problem? Have we established a pattern for OLM operator teams to maintain branches/versions for backwards compatiblity? Or are we mapping new territory here?

enhancements/builds/remove-jenkins-playload.md
```diff
@@ -0,0 +1,225 @@
...
+
+#### Jenkins Image Tagging Scheme
+
+The Jenkins image will use the maximum supported OCP version as the basis of its tagging scheme:
+
+1. `latest` - latest image build
+2. `4` - latest image that supports OCP 4
+3. `4.<Y>`- latest image that supports up to OCP 4.Y
+4. `4.<Y>.<Z>` - latest image that supports up to OCP 4.Y, with bugfix advisory Z
+
```
adambkaplan, line 97: @akram I'm a bit naive when it comes to what CPaaS can do. Can we configure it to produce multiple tags from the same branch?

For example - if the image in main/master supports OCP 4.6, can we configure CPaaS to produce the following tags:

1. `latest`, `4`
2. `4.8`, `4.8.0`
3. `4.7`, `4.7.<z>`
4. `4.6`, `4.6.<z>`

Similarly, when we branch for backwards compatibility, can we tell CPaaS to produce a different set of tags on the backport branch?

enhancements/builds/remove-jenkins-playload.md
bparees: first, we need to distinguish between talking about the operand image and the operator image.

With respect to operator images:  teams version their operators, always.  And they can decide which OCP catalog a given operator version goes into ("v1.1 of my operator only goes into the v4.7 OCP catalog"), though in theory an OCP 4.8 cluster can consume the v4.7 OCP catalog, but normally it does not.

With respect to operands:  The operator contains image references to the images it needs for operands. Those are normally SHA pinned when the operator bundle is constructed, so the operator is tied to a specific operand image.  However the operator can also be written to consume its operands in a non-SHA pinned way, or even in a way where you choose which operand image/version you want when you instantiate whatever thing the operator is providing.

but tldr: yes, operators and their associated operand are normally versioned.

As for how operator teams handle this..... yes they would normally maintain release branches.  That is what the logging operator team does, for example, but of course they are coming from the OCP world.  Some of the less OCP-coupled operator teams maintain branches named based on their operand version, though.

enhancements/builds/remove-jenkins-playload.md
```diff
@@ -0,0 +1,225 @@
...
+#### Jenkins Plugin Version Scheme
+
+Jenkins plugins for OpenShift currently use the 1.0.<z> versioning scheme.
+This versioning scheme will be updated so that the minor Y version corresponds to the lowest supported version of OCP 4 for the plugin.
+For example, the next release of the openshift-client plugin should be 1.6.0, since OCP 4.6 is the lowest supported version of OCP for the plugin (that we can support).
```
adambkaplan, line 102: This will help with backports for plugins. If we make a breaking change because OCP changed underneath us, we increase the minor version number.

enhancements/builds/remove-jenkins-playload.md
```diff
@@ -0,0 +1,225 @@
...
+
+The main/master branch for Jenkins and its plugins will be assumed to support the version of OpenShift under development.
+If a breaking change is introduced, a separate branch named `ocp-4.n` should be created to handle bug backports, with `n` representing the highest supported version of OpenShift.
+The requisite delivery pipelines will also need to be updated so that bugfix builds can be issued from `ocp-*` branches.
+
+#### Jenkins / OCP Compatibility
+
+The label `io.openshift.versions` will be added to the Jenkins image via a Dockerfile LABEL instruction.
+Its value will be an inclusive range of OCP major+minor versions that the image is officially supported on.
+For example, the current Jenkins image would set the value `v4.6-v4.8`.
```
adambkaplan, line 112: This is useful metadata to have for users who pull by SHAs. An image tagged `4.6.z` could run on OCP 4.8 just fine, and vice versa.

enhancements/builds/remove-jenkins-playload.md
adambkaplan: Verified - today's templates use imagechange triggers. On upgrade the imagestream tags should reimport, and clusters will pick up the new tags.

enhancements/builds/remove-jenkins-playload.md
adambkaplan: Proposal has been updated to include branching and versioning scheme. We will carry backport branches when breaking changes are introduced.

enhancements/builds/remove-jenkins-playload.md
adambkaplan: Reviewed with the CPaaS folks - the answer is yes, sort of. CPaaS + Errata lets us pick which tags get pushed to the container catalog.

adambkaplan:
bump @bparees 

Gone ahead and marked this EP "implementable" based on our discussion.

@gabemontero @akram it's worth you taking a look for lgtm

bparees:
/approve
(will let the build/jenkins team give the final lgtm signoff)

enhancements/builds/remove-jenkins-payload.md
```diff
@@ -0,0 +1,239 @@
...
+3. `4.<Y>`- latest image that supports up to OCP 4.Y
+4. `4.<Y>.0-YYYYMMDDHHMM-<git-sha>` - latest image that supports up to OCP 4.Y, produced at the provided datetime YYYYMMDDHHMM format and source commit sha.
+
+CPaaS and Errata will be configured so that tags for `4.Y` are set appropriately for the given branch, based on the `io.openshift.versions` Dockerfile label above.
+For example, if the Dockerfile has the `v4.6-v4.9` version label, the image produced by CPaaS will have the following tags:
+
+- `latest`
+- `4`
+- `4.6`, `4.7`, `4.8`, and `4.9`
+- `4.6.0-YYYYMMDDHHMM-<git-sha>`, `4.7.0-YYYYMMDDHHMM-<git-sha>`, `4.8.0-YYYYMMDDHHMM-<git-sha>`, `4.9.0-YYYYMMDDHHMM-<git-sha>`
```
adambkaplan, line 126: @akram please verify that this tagging scheme if feasible. If yes, then you can go ahead and add lgtm to this proposal.

akram: @adambkaplan I don't see CPaaS being able to specify a custom tagging scheme. It relies on OSBS/brew at the end of the day, and we just got tags like:
`jenkins-4.8-ordinal` where ordinal is an incremented sequence from and by brew.

sbose78:
Thanks for putting this up, on my list to review today.

enhancements/builds/remove-jenkins-payload.md
adambkaplan: Updated so the tag scheme uses the ordinal from brew

akram:
That looks good to me @adambkaplan , before putting the stamp, I want to ask if there is nothing specific that we may have forgotten related to air gapped environment? I don't think about anything, but, I am questioning yall.

adambkaplan:
> I want to ask if there is nothing specific that we may have forgotten related to air gapped environment? I don't think about anything, but, I am questioning yall.

I think the answer here is that Jenkins will need to be mirrored separately, just like any other imagestream. I've called this out in the EP as something that needs to be documented, as current customers will not be used to doing this.

akram:
Makes sense. Great EP ; thanks @adambkaplan 
/LGTM

gabemontero:
> > I want to ask if there is nothing specific that we may have forgotten related to air gapped environment? I don't think about anything, but, I am questioning yall.
> 
> I think the answer here is that Jenkins will need to be mirrored separately, just like any other imagestream. I've called this out in the EP as something that needs to be documented, as current customers will not be used to doing this.

In fact, the docs somewhat get into this, noting that jenkins amongst the samples is a special case, given the image comes from the payload.  See https://docs.openshift.com/container-platform/4.8/openshift_images/samples-operator-alt-registry.html#installation-restricted-network-samples_samples-operator-alt-registry

Feels like at one point we had specifics in the "disconnected support" section, but I am not seeing it in https://docs.openshift.com/container-platform/4.8/installing/installing-mirroring-installation-images.html#installing-preparing-samples-operator

PR #842, opened 2021-07-21
Restore tenant configuration for Loki forwarder.

alanconway:
This configuration will not be required for our default Loki instance, but may be required for custom Loki instances.

/cc @periklis
/assign @jcantrill

### DIFF
--- a/enhancements/cluster-logging/forward-to-loki.md
+++ b/enhancements/cluster-logging/forward-to-loki.md
@@ -21,8 +21,8 @@ superseded-by:
 - [X] Enhancement is `implementable`
 - [X] Design details are appropriately documented from clear requirements
 - [X] Test plan is defined
-- [ ] Operational readiness criteria is defined
-- [ ] Graduation criteria for dev preview, tech preview, GA
+- [X] Operational readiness criteria is defined
+- [X] Graduation criteria for dev preview, tech preview, GA
 - [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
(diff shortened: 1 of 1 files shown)


enhancements/cluster-logging/forward-to-loki.md
```diff
@@ -148,6 +148,12 @@ The following optional output fields are Loki-specific:
   Example: `kubernetes.labels.foo` => `kubernetes_labels_foo`.\
   **Note**: `kubernetes.host` is *always* be included, even if not requested.
   It is required to ensure ordered label streams.
+- `tenantKey`: (string, default=`"kubernetes.namespaceName"`) \
```
periklis, line 151: We don't provide a default for in-cluster.

periklis: In addition do we really want to define as tenant key a label key or simply leave it as arbitrary data?

alanconway: A fixed tenant ID limits us to static tenants pre-defined by the log admin based on input filters. This has turned out to be inadequate for other cases (JSON, cloudwatch groups etc.), where the logging system needs to respond to new label values or new namespaces appearing in the cluster without a manual re-config each time. I think the same applies to tenants - you can't implement "namespace as tenant" without it unless you know all your namespaces in advance.

alanconway:
@periklis @jcantrill  can I get an LGTM here? This brings the enhancement proposal up to date with the code that merged to the CLO.

PR #843, opened 2021-07-21
Node Healthcheck Operator

rgolangh:
Signed-off-by: Roy Golan <rgolan@redhat.com>

### DIFF
--- a/enhancements/node-healthcheck-operator.md
+++ b/enhancements/node-healthcheck-operator.md
@@ -11,7 +11,7 @@ reviewers:
   - "@slintes"
   - "@dhellmann"
 approvers:
-  - "@abeekof"
+  - "@abeekhof"
   - "@n1r1"
 ---
 
--- a/enhancements/node-healthcheck-operator.md
+++ b/enhancements/node-healthcheck-operator.md
@@ -85,7 +85,7 @@ expected to track the [node maintenance lease proposal]
 - Detection of Machines that never become healthy Nodes is not part of the initial implementation.  Machine API based cluster can/should continue to use MHC at this time.
 
 ## Proposal
-The main use-case we target is UPI, user provisioned infrastructure, which do not have
+The main use-case being targeted is clusters which, for technical and/or political reasons, do not have
 the Machine-API controlling them. This enables any kind of recovery mechanism to be built and handle remediation.
 
 #### Stories
@@ -447,4 +447,3 @@ started right away.
 [poison pill]: https://github.com/poison-pill/poison-pill
 [node maintenance lease proposal]: https://github.com/kubernetes/enhancements/pull/1411/
 [remediation API]:
-
--- a/enhancements/node-healthcheck-operator.md
+++ b/enhancements/node-healthcheck-operator.md
@@ -37,7 +37,22 @@ that comply with the existing  [api].
 
 Enable clusters that do not have a functioning Machine API to detect and respond to node-level failures in order to recover workloads , and/or restore cluster capacity.
 
-The existing Machine Health Check controller requires a functioning machine-api. This prevents its use in the majority of Kubernetes and OpenShift clusters.
+The majority of OpenShift clusters are UPI based and do not have Machine objects
+and/or a functioning Machine API - which is a prerequisite for using the Machine 
+HealthCheck controller.
+
+There are a variety of reasons, even on cloud, that customers might want or need
+to pursue a UPI strategy *for deployment*, but that doesn't need to prevent them
+from being able to use a standard solution for protecting their workloads, or imply 
+that they would prefer to create a bespoke solution.
+
+There are a number of barriers to retrofitting a Machine API to an existing UPI 
+cluster, but the most common raised are: complexity, technical limitations, and 
+customer policies which prevent BMCs or platform credentials from being stored 
+within the cluster).
+
+While there remains a need for UPI, there is also a need to allow UPI workloads to 
+be made highly available, and customers look to Red Hat to provide a solution.
 The proposed Node Healthcheck Controller (NHC) provides a generic solution that fits all installation types, together with an external remediation API that allows a pluggable remediation system to be installed and remediate unhealthy nodes.
 
 ### Goals
@@ -430,6 +445,5 @@ started right away.
 [poison pill]: https://github.com/poison-pill/poison-pill
 [node maintenance lease proposal]: https://github.com/kubernetes/enhancements/pull/1411/
 [remediation API]:
-
 
 
--- a/enhancements/node-healthcheck-operator.md
+++ b/enhancements/node-healthcheck-operator.md
@@ -1,0 +1,435 @@
+Node Health Check Enhancement 
+
+---
+title: node-healthcheck-controller
+authors:
+  - "@rgolangh"
+reviewers:
+  - "@abeekof"
+  - "@n1r1"
+  - "@mshitrit"
+  - "@slintes"
+  - "@dhellmann"
+approvers:
+  - "@abeekof"
+  - "@n1r1"
+---
+
+# Node health checking
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift/docs
+
+
+## Summary
+
+
+Enable opt in for automated node health checking and create objects
+for unhealthy nodes, to be handled later by external controllers
+that comply with the existing  [api].
+
+## Motivation
+
+Enable clusters that do not have a functioning Machine API to detect and respond to node-level failures in order to recover workloads , and/or restore cluster capacity.
+
+The existing Machine Health Check controller requires a functioning machine-api. This prevents its use in the majority of Kubernetes and OpenShift clusters.
+The proposed Node Healthcheck Controller (NHC) provides a generic solution that fits all installation types, together with an external remediation API that allows a pluggable remediation system to be installed and remediate unhealthy nodes.
+
+### Goals
+
+- Create an installation-type agnostic node health check solution
+
+- Feature parity with Machine Health Check controller
+
+- Providing a default strategy combined with [poison-pill].
+
+- Endorse usage of external [remediation api]
+
+- Allow defining different unhealthy criteria by node selectors, and have a sane strategy for handling nodes that match multiple NHC CRs.
+
+- Provide defaults health checks and default remedy custom resource creation.
+
+- Long term - Make Machine Health Check controller only implement the external remediation api and drop the default Machine API based remediation mechanism
+
+
+### Non-Goals
+
+- Coordination of remedy operation - operators/controllers implementers
+expected to track the [node maintenance lease proposal]
+
+- Recording long-term stable history of all health-check failures or remediation
+  actions.
+
+- Detection of Machines that never become healthy Nodes is not part of the initial implementation.  Machine API based cluster can/should continue to use MHC at this time.
+
+## Proposal
+The main use-case we target is UPI, user provisioned infrastructure, which do not have
+the Machine-API controlling them. This enables any kind of recovery mechanism to be built and handle remediation.
+
+#### Stories
+- As an admin of a cluster I want the cluster to detect Node failures and
+ initiate recovery, so that workloads and volumes with at-most-one semantics
+  can be run/mounted elsewhere.
+
+- As an admin of a cluster I want the cluster to detect Node failures and
+ initiate recovery, so that we can attempt to restore cluster capacity.
+
+- As an OCP engineer, I want the NHC to consume the same logic for detecting
+ failures, to avoid code duplication and provide a consistent experience
+  compared to MHC.
+
+- As an OCP engineer, I want the NHC to use the same interface for initiating
+ recovery, so that we can reuse existing mechanisms (like [poison pill] and metal3) and
+  avoid code duplication.
+
+To fulfil the above we create a [Node Health Check custom resource (NCR)](anchor todo) that specifies:
+1. health criteria: list of criterias and selector
+2. template: a template of an [external custom resource (remediation CR)](link todo)
+ to create in response to a match.
+
+### Unhealthy criteria
+A node target is unhealthy when the node meets the unhealthy node conditions criteria defined.
+If any of those criterias are met for longer than given timeouts, the controller
+creates a remediation CR according to the template in the NHC CRD.
+Timeouts will be defined as part of the NHC CRD .
+
+### Implementation Details
+
+#### NodeHealthCheck CRD
+- Enable watching a node based on a label selector. 
+- Enable defining an unhealthy node criteria (based on a list of node conditions).
+- Enable setting a threshold of unhealthy nodes. If the current number is at or above this threshold no further action will take place. This can be expressed as an int or as a percentage of the total targets in the pool.
+
+E.g:
+- Create a remedy CR when node is `ready=false` or `ready=Unknown` condition for more than 5m.
+- I want to avoid triggering remediation if 40% or more of the targets of this pool are unhealthy at the same time.
+
+
+```yaml
+apiVersion: remediation.medik8s.io/v1alpha1
+kind: NodeHealthCheck
+metadata:
+  name: example
+  # this CRD is cluster-scope, namespace is not needed  
+spec:
+  maxUnhealthy: "49%"
+  selector:
+    matchExpressions:
+      key: node-role.kubernetes.io/worker
+      operator: exists
+  unhealthyConditions:
+    - type: "Ready"
+      status:  "Unknown"
+      duration: "300s"
+   - type: "Ready"
+     status:  "False"
+     duration: "300s"
+  remediationTemplate:
+    kind: PoisonPillRemediationTemplate
+    apiVersion: poison-pill.medik8s.io/v1alphaX
+    name: M3_REMEDIATION_GROUP
+status:
+  healthyNodes: 5
+  obeservedNodes: 5
+```
+
+#### NodeHealthCheck controller
+Watch:
+- Watch NodeHealthCheck resources
+- Watch nodes with an event handler e.g controller runtime `EnqueueRequestsFromMapFunc` which returns NodeHealthCheck resources.
+
+Reconcile:
+- Periodically enqueue a reconcile request
+- Fetch all nodes targets. E.g:
+
+
+- Calculate the number of unhealthy targets.
+- Compare current number against `maxUnhealthy` threshold and temporary short
+ circuits if the threshold is met.
+- Create remedy objects for unhealthy targets according to the template in NHC object
+
+### Notes/Constraints
+
+The functionality of this controller have a lot in common with the Machine Health
+Check controller except that its creating objects and not
+performing any actions per se. When deploying this controller make
+sure 1. you have a controller implementing the API, e.g [poison pill]
+2. the machine health check controller is not running, or in case of a mixed
+ the node health checks selector avoid targeting nodes with machines.
+
+
+### Risks and Mitigations
+
+#### Remediation Interaction with ClusterAutoScalar
+
+Misconfiguration of nodes provisioning can cause nodes to fail to get operational.
+Remediation is blind to such problems and may interfere with diagnosing the problem.
+The backoff mechanism mentioned in MHC may be suited for Machines, because they
+have the MachineSet name that can be recorded and tracked. Without the machine API
+our knowledge of the nodes could not be handled in such a generic way. There is
+simply no generic way to record a node configuration origin in reliable way without
+putting provider specific annotations.
+Potentially the NHC resource could contain label selectors to track
+repeating remediation and mark them as [incrementally back-off](https://issues.redhat.com/browse/OCPCLOUD-800). e.g
+
+Suggested backoff specification under the NHC resource may look like this:
+```yaml
+kind: NodeHealthCheck
+...
+spec:
+  selector:
+     ...
+  # optional
+  backoff:
+    type: exponential
+    limit: 10m
+```
+
+Caveat of this approach is that we need to take care of conflicting or overlapping backoff definitions.
+Setting global catchall backoff strategy is impossible or just coarse without specifying a selector (how do we know what to backoff from?)
+
+Currently backoff is not implemented.
+
+#### Multiple remediation provider deployed
+
+Multiple providers could act on remediation objects. On non heterogeneous cluster
+this may even make sense. It is not the goal of this controller to coordinate
+those, but providers will need disparate set of nodes to handle based on labels,
+or operator on a node if they can hold a [maintenance lease][node maintenance lease proposal].
+
+
+## Design Details
+
+### Logical Conditions Matching
+
+The resource will be given a list of conditions to track, per a selection of
+nodes, and will evaluate each condition in the list. It's a logical evaluation i.e
+it is enough to evaluate a single condition to true per node to mark as unhealthy.
+
+```
+unhealthyConditions:
+   - type:    "Ready"
+     status:  "Unknown"
+     duration: "300s"
+   - type:    "Ready"
+     status:  "False"
+     duration: "300s"
+```
+In the above example only 1 condition needs to match to mark a node non-healthy.
+
+### Max Unhealthy count
+
+To create remediation object (i.e act for remediation) the overall nodes in the
+selection groups must not exceed `maxUnhealthy` threshold. The value can be
+an int or a percentage, similar to v1/apps `maxUnavailable`:
+```
+spec:
+ maxUnhealthy: 30%
+```
+
+### Default spec values:
+selector: {matchingExpression: { Key: node-role.kubernetes.io/worker, Operator: Exists } }
+maxUnhealthy: 49%
+unhealthyConditions: [ {type: NotReady, status: Unknown, duration: 300s},{ type: Ready, status: False, 300s} ]
+remediationTemplate: n/a
+
+### Mandatory spec fields:
+remediationTemplate
+
+### Optional spec fields:
+  - selector
+  - maxUnhealthy
+  - unhealthyConditions
+
+
+### Status sub-resource fields
+observedNodes - specifies how much nodes are observed by this object selector
+heathyNodes     - specify the number of healthy nodes
+inFlightRemediations - a mapping of a remediaiton object name (== the node name) and the time it was created. Removed
+when the remediation object deleted.
+
+
+### Remediation Object creation
+Add the target Node to ownerReferences. We want that object to be garbage
+collected with the referenced Node
+
+
+### Controller details
+leader election: true
+Sync period: 60s - determines how frequent reconciles happen regardless of resource changes.
+watches NHC objects
+watches Node objects - this will make the reconcile more efficient. It will allow a sync period with longer intervals
+
+# [rgolan] the rest is WIP #
+
+### Open Questions [optional]
+
+This is where to call out areas of the design that require closure before deciding
+to implement the design.  For instance,
+ > 1. This requires exposing previously private resources which contain sensitive
+  information.  Can we do this?
+
+### Test Plan
+
+**Note:** *Section not required until targeted at a release.*
+
+Consider the following in developing a test plan for this enhancement:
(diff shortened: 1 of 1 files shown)


enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+Reconcile:
+- Periodically enqueue a reconcile request
+- Fetch all nodes targets. E.g:
+
+
+- Calculate the number of unhealthy targets.
+- Compare current number against `maxUnhealthy` threshold and temporary short
+ circuits if the threshold is met.
+- Create remedy objects for unhealthy targets according to the template in NHC object
```
dhellmann, line 154: What are remedy objects? Is that an API for another controller?

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+### Notes/Constraints
+
+The functionality of this controller have a lot in common with the Machine Health
+Check controller except that its creating objects and not
+performing any actions per se. When deploying this controller make
+sure 1. you have a controller implementing the API, e.g [poison pill]
```
dhellmann, line 161: This dependency should be called out earlier in the proposal section to make it clear that the new controller notices unhealthy nodes, but relies on some other actor to resolve any problems.

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+To create remediation object (i.e act for remediation) the overall nodes in the
+selection groups must not exceed `maxUnhealthy` threshold. The value can be
+an int or a percentage, similar to v1/apps `maxUnavailable`:
+```
+spec:
+ maxUnhealthy: 30%
+```
+
+### Default spec values:
```
dhellmann, line 235: These are the defaults for the NodeHealthCheck spec? Why were these defaults selected?

nit: This section would be easier to read if it was formatted as YAML, or at least wrapped in '```' so the markdown parser does not try to interpret it.

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+
+### Remediation Object creation
+Add the target Node to ownerReferences. We want that object to be garbage
+collected with the referenced Node
+
+
+### Controller details
+leader election: true
+Sync period: 60s - determines how frequent reconciles happen regardless of resource changes.
```
dhellmann, line 264: Can the user change this value?

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+2. template: a template of an [external custom resource (remediation CR)](link todo)
+ to create in response to a match.
+
+### Unhealthy criteria
+A node target is unhealthy when the node meets the unhealthy node conditions criteria defined.
+If any of those criterias are met for longer than given timeouts, the controller
+creates a remediation CR according to the template in the NHC CRD.
+Timeouts will be defined as part of the NHC CRD .
+
+### Implementation Details
```
dhellmann, line 101: What is the plan to package and deliver this new component? Is it going into the OCP release payload, or are we going to deliver it through OLM?

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+- Coordination of remedy operation - operators/controllers implementers
+expected to track the [node maintenance lease proposal]
+
+- Recording long-term stable history of all health-check failures or remediation
+  actions.
+
+- Detection of Machines that never become healthy Nodes is not part of the initial implementation.  Machine API based cluster can/should continue to use MHC at this time.
+
+## Proposal
```
dhellmann, line 70: It would be useful to have a subsection describing how the feature provided by the new controller is used, end to end. Something like

1. User installs the NodeHealthCheck operator
2. User creates a NodeHealthCheck resource describing rules for monitoring nodes. (How do they decide what rules apply? Is there more configuration to do?)
3. The controller sees that a node fails the check rule described.
4. The controller does ... (I think it's creating instructions for a remediation controller?)
5. The remediation controller does ...
6. The node is deleted/rebooted/reprovisioned/other

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+Enable opt in for automated node health checking and create objects
+for unhealthy nodes, to be handled later by external controllers
+that comply with the existing  [api].
+
+## Motivation
+
+Enable clusters that do not have a functioning Machine API to detect and respond to node-level failures in order to recover workloads , and/or restore cluster capacity.
+
+The existing Machine Health Check controller requires a functioning machine-api. This prevents its use in the majority of Kubernetes and OpenShift clusters.
```
dhellmann, line 40: It would be good to have more justification here for why we need that feature for user-managed deployments and why we don't simply recommend creating those deployments with all of the automation that the IPI workflow brings with it.

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+  - "@rgolangh"
+reviewers:
+  - "@abeekof"
+  - "@n1r1"
+  - "@mshitrit"
+  - "@slintes"
+  - "@dhellmann"
+approvers:
+  - "@abeekof"
+  - "@n1r1"
```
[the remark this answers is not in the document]
JoelSpeed: I'll review this next week. I've already been briefed on the general concept so should have sufficient context

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift/docs
+
+
+## Summary
+
+
+Enable opt in for automated node health checking and create objects
+for unhealthy nodes, to be handled later by external controllers
+that comply with the existing  [api].
```
beekhof, line 34:
```suggestion
This enhancement proposes the creation of a new OLM-delivered operator
that supports NodeCondition-based health checking of UPI worker nodes.

The new operator utilises the same External Remediation API as Machine 
Healthcheck to trigger recovery of affected nodes and workloads.
```

enhancements/node-healthcheck-operator.md
beekhof: Wouldn't that basically be a justification of why UPI exists?
Given that a majority of deployments are UPI, I'd have thought that to be assumed knowledge.

beekhof:
```suggestion
The majority of OpenShift clusters are UPI based and do not have Machine objects
and/or a functioning Machine API - which is a prerequisite for using the Machine 
HealthCheck controller.

There are many possible barriers to retrofitting a Machine API to an existing UPI 
cluster, but the most common raised are: complexity, technical limitations, and 
customer policies which prevent BMCs or platform credentials from being stored 
within the cluster).

While there remains a need for UPI, there is also a need to allow UPI workloads to 
be made highly available.
```

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+- Provide defaults health checks and default remedy custom resource creation.
+
+- Long term - Make Machine Health Check controller only implement the external remediation api and drop the default Machine API based remediation mechanism
+
+
+### Non-Goals
+
+- Coordination of remedy operation - operators/controllers implementers
+expected to track the [node maintenance lease proposal]
```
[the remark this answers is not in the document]
beekhof: +1 for inline links :)

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+- Recording long-term stable history of all health-check failures or remediation
+  actions.
```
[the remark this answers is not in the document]
beekhof: yes

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+- Detection of Machines that never become healthy Nodes is not part of the initial implementation.  Machine API based cluster can/should continue to use MHC at this time.
```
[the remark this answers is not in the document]
beekhof: I'm not sure exactly where to capture the above information, but it does represent the one usecase we can point to for **not** trying to replace MHC at this time.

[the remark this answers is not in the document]
beekhof:
```suggestion
- Replace Machine Healthcheck for IPI clusters
```

enhancements/node-healthcheck-operator.md
beekhof: OLM. See suggest revision to the summary

enhancements/node-healthcheck-operator.md
beekhof: The External Remediation API that's also used by MHC

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+### Unhealthy criteria
+A node target is unhealthy when the node meets the unhealthy node conditions criteria defined.
+If any of those criterias are met for longer than given timeouts, the controller
+creates a remediation CR according to the template in the NHC CRD.
+Timeouts will be defined as part of the NHC CRD .
+
+### Implementation Details
+
+#### NodeHealthCheck CRD
```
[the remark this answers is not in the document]
beekhof: Derived from the MHC CRD

enhancements/node-healthcheck-operator.md
dhellmann: I would expect a UPI deployment to use external tools for monitoring, for example, just as it does for provisioning. Similarly, I would expect a user who wants automation for managing their cluster to be fully integrated to use the IPI tools, even if that means redeploying. So, having some sort of detail about cases where the user wants or needs a UPI deployment but also wants the automation described here would help explain why we want to support a new component duplicating the functionality of one we already have.

beekhof: TBH, that seems punitive.  You didn't want X so you can't have Y either.

IIUC there are a variety of reasons, even on cloud, that customers might need to pursue a UPI strategy, but there are zero technical reasons that needs to prevent them from being able to use a standard solution for protecting their workloads.

We both heard customer0 very clearly say they would much rather use a solution from Red Hat than write their own and I see no reason they would be an exception.

I thought we were trying to get away from this idea that how you installed a cluster locks you into a certain feature set.

dhellmann: Those are good justifications for solving the problem, and would be good additions to the doc.

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+sure 1. you have a controller implementing the API, e.g [poison pill]
+2. the machine health check controller is not running, or in case of a mixed
+ the node health checks selector avoid targeting nodes with machines.
+
+
+### Risks and Mitigations
+
+#### Remediation Interaction with ClusterAutoScalar
+
+Misconfiguration of nodes provisioning can cause nodes to fail to get operational.
```
JoelSpeed, line 170: If the node doesn't become operational, how will the NHC remediate this? I don't really understand how the NHC would have this kind of issue at all, it doesn't know which nodes are meant to be coming up, which is where this issue occurs in MHC

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+```
+
+Caveat of this approach is that we need to take care of conflicting or overlapping backoff definitions.
+Setting global catchall backoff strategy is impossible or just coarse without specifying a selector (how do we know what to backoff from?)
+
+Currently backoff is not implemented.
+
+#### Multiple remediation provider deployed
+
+Multiple providers could act on remediation objects. On non heterogeneous cluster
```
JoelSpeed, line 200: What happens when NHC and MHC are in play at the same time?

dhellmann:
For some reason Github won't let me reply in a thread to Joel's comment https://github.com/openshift/enhancements/pull/843#discussion_r678208612

> There's nothing that prevents a customer from using Machine API on a UPI cluster, so I don't think we should be implying MHC doesn't work on most clusters. Granted most UPI clusters don't use MAPI, but that's not what is stated here.
>
> Are we explicitly trying to support platforms that don't currently have Machine API support at all?

I think the most important use for this is likely on-prem metal clusters deployed via UPI. I don't know if it's possible to enable the machine API there at all, or if it is possible to do it without giving the cluster credentials to the BMCs. Adding some of that background to the Motivation section would help.

enhancements/node-healthcheck-operator.md
beekhof: Answering the easy question first...

> Are we explicitly trying to support platforms that don't currently have Machine API support at all?

@JoelSpeed absolutely. that is the primary goal

> There's nothing that prevents a customer from using Machine API on a UPI cluster, so I don't think we should be implying MHC doesn't work on most clusters. Granted most UPI clusters don't use MAPI, but that's not what is stated here.

I'm not sure I see the distinction...
There may not be an internal technical reason[1], but presumably the majority of customers that deploy with UPI aren't just being lazy and there is something about the customer environment (technical or political) that prevents them turning on MAPI after installation too.

[1] Though based on experience with Assisted Installer and metal clusters, I would argue it is far from a trivial exercise.

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+authors:
+  - "@rgolangh"
+reviewers:
+  - "@abeekof"
+  - "@n1r1"
+  - "@mshitrit"
+  - "@slintes"
+  - "@dhellmann"
+approvers:
+  - "@abeekof"
```
beekhof, line 14:
```suggestion
  - "@abeekhof"
```

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
+Node Health Check Enhancement 
+
+---
+title: node-healthcheck-controller
+authors:
+  - "@rgolangh"
+reviewers:
+  - "@abeekof"
```
beekhof, line 8:
```suggestion
  - "@abeekhof"
```

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+- Providing a default strategy combined with [poison-pill].
+
+- Endorse usage of external [remediation api]
+
+- Allow defining different unhealthy criteria by node selectors, and have a sane strategy for handling nodes that match multiple NHC CRs.
+
+- Provide defaults health checks and default remedy custom resource creation.
+
+- Long term - Make Machine Health Check controller only implement the external remediation api and drop the default Machine API based remediation mechanism
```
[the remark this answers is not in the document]
beekhof: I think this step is optional and ultimately orthogonal to the proposal.

```suggestion
```

If we did head down this track, I believe it could be achieved without changing the MHC CRD.

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+expected to track the [node maintenance lease proposal]
+
+- Recording long-term stable history of all health-check failures or remediation
+  actions.
+
+- Detection of Machines that never become healthy Nodes is not part of the initial implementation.  Machine API based cluster can/should continue to use MHC at this time.
+
+## Proposal
+The main use-case we target is UPI, user provisioned infrastructure, which do not have
+the Machine-API controlling them. This enables any kind of recovery mechanism to be built and handle remediation.
```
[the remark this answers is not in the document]
beekhof: I/we should probably stop using UPI as shorthand for "clusters without MAPI" to avoid creating this kind confusion.

To me, why a given cluster doesn't have MAPI or how it got to that state, is less relevant than the number of customers that can't benefit from the existing solution.

```suggestion
The main use-case being targeted is clusters which, for technical and/or political reasons, do not have
the Machine-API controlling them. This enables any kind of recovery mechanism to be built and handle remediation.
```

enhancements/node-healthcheck-operator.md
beekhof: NHC will do nothing if it detects a MHC CR (except for the one covering spot instances)

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+- Calculate the number of unhealthy targets.
+- Compare current number against `maxUnhealthy` threshold and temporary short
+ circuits if the threshold is met.
+- Create remedy objects for unhealthy targets according to the template in NHC object
+
+### Notes/Constraints
+
+The functionality of this controller have a lot in common with the Machine Health
+Check controller except that its creating objects and not
```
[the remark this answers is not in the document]
beekhof: We did approach CAPI with that idea, however there was no enthusiasm at the time.
This is not hugely surprising since it effectively deemphasises the role of Machines and MAPI.

We're still open to the idea of adapting the existing controller (that's basically how NHC was created), however I would have concerns about the speed at which progress can be made.  MHC was a fine MVP 2.5 years ago but it really hasn't been allowed to progress much since.

enhancements/node-healthcheck-operator.md
beekhof: NHC's problem is being able to detect such a condition, remediation is delegated to external operators.
As discussed below, today NHC doesn't detect this scenario and is one reason why it doesn't replace MHC.

Even if NHC did detect the condition, the primary goal is to ensure that protected workloads are not running there (so that they can be started elsewhere). Which, even if the NHC does nothing, is true by default for a node that has never contacted the scheduler.

While restoring compute capacity in the presence of non-transient errors is not possible without MAPI[1], this doesn't detract from being able to automatically and safely move the workloads elsewhere.

[1] And sometimes even *with* MAPI.  Reprovisioning a physical node with a bad disk or DIMM is going to give you back the same broken machine.

enhancements/node-healthcheck-operator.md
beekhof: They were inherited from the MHC implementation

enhancements/node-healthcheck-operator.md
beekhof: I'm not sure we need to call this out anymore. The operator dependancies automatically pulls in poison pill since this was originally drafted.

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+performing any actions per se. When deploying this controller make
+sure 1. you have a controller implementing the API, e.g [poison pill]
+2. the machine health check controller is not running, or in case of a mixed
+ the node health checks selector avoid targeting nodes with machines.
```
beekhof, line 163:
```suggestion
sure the machine health check controller is not configured.
```

enhancements/node-healthcheck-operator.md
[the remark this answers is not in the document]
dhellmann: Let's capture this info in the doc.

enhancements/node-healthcheck-operator.md
JoelSpeed: Should probably add a link out to the section for MHC that describes this behaviour

enhancements/node-healthcheck-operator.md
JoelSpeed: I understand that NHC doesn't detect this condition, but the topic of this paragraph seems to imply it is borrowing the back off mechanism implemented in MHC for this very reason.

Given NHC only knows about Nodes and doesn't know about their owners, i don't see how it can track or relate nodes in a way that a backoff could be implemented, unless this was a global backoff for the entire NHC?

How often do we predict that this kind of backoff would be required?

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+  # optional
+  backoff:
+    type: exponential
+    limit: 10m
+```
+
+Caveat of this approach is that we need to take care of conflicting or overlapping backoff definitions.
+Setting global catchall backoff strategy is impossible or just coarse without specifying a selector (how do we know what to backoff from?)
+
+Currently backoff is not implemented.
```
JoelSpeed, line 196: Probably don't need this comment, enhancements are written before anything is implemented normally 😉

enhancements/node-healthcheck-operator.md
JoelSpeed: This needs to be called out in the enhancement

enhancements/node-healthcheck-operator.md
beekhof: I think it's more trying to say why we *can't* borrow the same exact mechanism from MHC.

Since NHC is primarily useful on clusters without MAPI, the Node would have the same name when it comes back though - tracking how many times the same node was remediated is possible in those circumstances, for anywhere else it could be handled globally (rg. I remediated N nodes in M minutes, I should stop)

enhancements/node-healthcheck-operator.md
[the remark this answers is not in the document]
beekhof: Seems like it belongs in the alternatives section. I don't think it needs to be a prerequisite for "provide a solution for non-MAPI clusters".

enhancements/node-healthcheck-operator.md
JoelSpeed: That's good context, please make sure that's reflected into the enhancement in this section

enhancements/node-healthcheck-operator.md
rgolangh: Ack

enhancements/node-healthcheck-operator.md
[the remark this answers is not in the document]
rgolangh: ack

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+### Unhealthy criteria
+A node target is unhealthy when the node meets the unhealthy node conditions criteria defined.
+If any of those criterias are met for longer than given timeouts, the controller
+creates a remediation CR according to the template in the NHC CRD.
+Timeouts will be defined as part of the NHC CRD .
+
+### Implementation Details
+
+#### NodeHealthCheck CRD
+- Enable watching a node based on a label selector.
```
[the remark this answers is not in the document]
beekhof: "Parity with MHC" basically.  Not sure we need to rehash the MHC design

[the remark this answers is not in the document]
dhellmann: OK, feature parity is listed in the goals but I guess in my last reading the link between the goal and these details wasn't obvious. I don't think any changes are needed here.

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,449 @@
...
+- Create a remedy CR when node is `ready=false` or `ready=Unknown` condition for more than 5m.
+- I want to avoid triggering remediation if 40% or more of the targets of this pool are unhealthy at the same time.
+
+
+```yaml
+apiVersion: remediation.medik8s.io/v1alpha1
+kind: NodeHealthCheck
+metadata:
+  name: example
+  # this CRD is cluster-scope, namespace is not needed
```
JoelSpeed, line 135: Would be good to document somewhere why this should be cluster scoped

Can you think of a possible future where this is used in a centrally managed infrastructure kind of scenario, eg HyperShift where it would run on the control plane? If this is a use case you wish to support at any point, making this cluster scoped is not desirable

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,449 @@
...
+    name: M3_REMEDIATION_GROUP
+status:
+  healthyNodes: 5
+  obeservedNodes: 5
+```
+
+#### NodeHealthCheck controller
+Watch:
+- Watch NodeHealthCheck resources
+- Watch nodes with an event handler e.g controller runtime `EnqueueRequestsFromMapFunc` which returns NodeHealthCheck resources.
```
JoelSpeed, line 161: Nodes create a LOT of events when they are updating their health checks, watching them will probably mean you are constantly reconciling. You may want to filter the events somehow here

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,435 @@
...
+
+Reconcile:
+- Periodically enqueue a reconcile request
```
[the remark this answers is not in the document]
JoelSpeed: If you are watching nodes which send events regularly, periodically reconciling is probably not required

enhancements/node-healthcheck-operator.md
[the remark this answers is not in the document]
JoelSpeed: I don't think this has been captured yet

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,449 @@
...
+
+
+### Risks and Mitigations
+
+#### Remediation Interaction with ClusterAutoScalar
+
+Misconfiguration of nodes provisioning can cause nodes to fail to get operational.
+Remediation is blind to such problems and may interfere with diagnosing the problem.
+The backoff mechanism mentioned in MHC may be suited for Machines, because they
+have the MachineSet name that can be recorded and tracked. Without the machine API
```
JoelSpeed, line 190: I don't think MHC knows anything about MachineSets 🤔  I also don't think we have a backoff mechanism, what is this referring to?

enhancements/node-healthcheck-operator.md
JoelSpeed: Still needs to be added

enhancements/node-healthcheck-operator.md
JoelSpeed: These don't match the MHC defaults. The only one of these fields we actually default is `maxUnhealthy` which defaults to 100% (ie turn off this feature)

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,449 @@
...
+### Default spec values:
+selector: {matchingExpression: { Key: node-role.kubernetes.io/worker, Operator: Exists } }
+maxUnhealthy: 49%
+unhealthyConditions: [ {type: NotReady, status: Unknown, duration: 300s},{ type: Ready, status: False, 300s} ]
+remediationTemplate: n/a
+
+### Mandatory spec fields:
+remediationTemplate
+
+### Optional spec fields:
```
JoelSpeed, line 261: Conditions and selector should be required as they are in MHC

enhancements/node-healthcheck-operator.md
JoelSpeed: Not normally, and 60s is far too frequent IMO. This means the controller will be doing a full list against the API for all objects it cares about every 60s. Typically this would be more on the order of minutes than seconds, eg IIRC, in MAPI we use 10m for the sync period. Others even use hours.

I think having a sync period of 60s puts unnecessary load on the API server without much benefit

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,449 @@
...
+The majority of OpenShift clusters are UPI based and do not have Machine objects
+and/or a functioning Machine API - which is a prerequisite for using the Machine 
+HealthCheck controller.
+
+There are a variety of reasons, even on cloud, that customers might want or need
+to pursue a UPI strategy *for deployment*, but that doesn't need to prevent them
+from being able to use a standard solution for protecting their workloads, or imply 
+that they would prefer to create a bespoke solution.
+
+There are a number of barriers to retrofitting a Machine API to an existing UPI
```
enxebre, line 51: Are these barriers referring to baremetal or also cloud envs?

enhancements/node-healthcheck-operator.md
```diff
@@ -0,0 +1,449 @@
...
+## Motivation
+
+Enable clusters that do not have a functioning Machine API to detect and respond to node-level failures in order to recover workloads , and/or restore cluster capacity.
+
+The majority of OpenShift clusters are UPI based and do not have Machine objects
+and/or a functioning Machine API - which is a prerequisite for using the Machine 
+HealthCheck controller.
+
+There are a variety of reasons, even on cloud, that customers might want or need
+to pursue a UPI strategy *for deployment*, but that doesn't need to prevent them
```
enxebre, line 47: It'd be nice to mention a few to make it clear why moving to mapi automated machine management is not an option.

enhancements/node-healthcheck-operator.md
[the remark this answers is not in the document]
enxebre: If this is a long term desire then I think it belongs to the non-goal section for this proposal.

PR #844, opened 2021-07-23
host-port-registry: add 9446 port

andfasano:
This is used by the baremetal-operator health endpoint.

See also https://github.com/openshift/cluster-baremetal-operator/pull/180

### DIFF
--- a/dev-guide/host-port-registry.md
+++ b/dev-guide/host-port-registry.md
@@ -85,6 +85,7 @@ Ports are assumed to be used on all nodes in all clusters unless otherwise speci
 | 9102  | kube-proxy | sdn | 4.7 | metrics, third-party network plugins only |
 | 9103  | ovn-kubernetes node kube-rbac-proxy | sdn || metrics |
 | 9258  | cluster-cloud-controller-manager-operator | cluster infra | 4.9 | metrics, control plane only |
+| 9446  | baremetal-operator | kni | 4.9 | healthz; baremetal provisioning, control plane only |
 | 9537  | crio      | node || metrics |
 | 9641  | ovn-kubernetes northd | sdn | 4.3 | control plane only, ovn-kubernetes only |
 | 9642  | ovn-kubernetes southd | sdn | 4.3 | control plane only, ovn-kubernetes only |


kirankt (approved):
/lgtm

andfasano:
Note: the markdown job is failing as it looks like the template is applied also to files that are not technically an enhancement request, as in this case

dhellmann:
#846 moves this file to the dev-guide directory to avoid the linter check.

dhellmann:
/cc @knobunc 

Is there a policy for choosing these ports beyond ensuring they're in the range? What we have now seems scattered.

dev-guide/host-port-registry.md
```diff
@@ -84,6 +84,7 @@ Ports are assumed to be used on all nodes in all clusters unless otherwise speci
 | 9102  | ovn-kubernetes master kube-rbac-proxy | sdn || metrics, control plane only, ovn-kubernetes only |
 | 9102  | kube-proxy | sdn | 4.7 | metrics, third-party network plugins only |
 | 9103  | ovn-kubernetes node kube-rbac-proxy | sdn || metrics |
+| 9446  | baremetal-operator | kni | 4.9 | healthz; baremetal provisioning, control plane only |
```
dhellmann, line 87: There are other healthz ports down in the 10000 range (see line 97 and below). Should this component be using that port range, too?

andfasano: @dhellmann do we have any recommendation for allocating health ports in that range? In that case maybe we could add also a note in the current document.

andfasano: I see that also the ccmo used a port in the 9k range: 

```
 9259  | cluster-cloud-controller-manager-operator | cluster infra | 4.9 | healthz, control plane only
```

(as a side note, it was defined in the `localhost-only` paragraph, which seems to be a better choice also for us, given that both operator are using the host networking).
I don't think we're strictly requiring a specific range, so in case we could move it into the 10k one

dhellmann: #618 includes some guidance for host port allocation

andfasano: Thanks @dhellmann . So IIUC, if the #618 propsal will get approved we should then move that port under the 20k (due the metal3 host networking)?
In the meantime would be possible to merge the current PR to reflect in the doc the current status?

dhellmann:
Sure, let's merge this since it reflects what we have today. You can get someone else on the team to verify the numbers and leave the lgtm.

/approve

dev-guide/host-port-registry.md
zaneb: #618 is only referring to workers though, and these ports are only on the control plane.

PR #848, opened 2021-07-26
lint: loosen some linter rules

dhellmann:
Some of the linter rules applied here make it inconvenient to import
documents written in other repositories.

/cc @wking
/cc @deads2k

### DIFF
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -43,3 +43,6 @@ config:
   blanks-around-lists: false
   single-trailing-newline: false
   no-multiple-blanks: false
+
+  # Do not require language declarations for code blocks.
+  fenced-code-language: false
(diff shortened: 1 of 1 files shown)


deads2k:
I don't object to the rule.  Is there really no way to override on a per-PR basis?  @Miciah is extremely creative with bash if we just need to find a way to select the right files.

dhellmann:
> I don't object to the rule for net-new docs. Is there really no way to override on a per-PR basis? @Miciah is extremely creative with bash if we just need to find a way to select the right files.

`hack/template-lint.sh` has some logic for finding the files changed in a PR and that could be added to `hack/markdownlint.sh`.

I'm not sure if I like that or not, though. Non-enhancement files seem much more likely to see changes over time, so we penalize future maintainers of those files if we don't make them conform to the linter rules up front.

dhellmann:
I think we agreed we don't need this.

PR #849, opened 2021-07-29
NE-310 Enhancement proposal for HSTS route admission plugin

candita:
Updates to the enhancement proposal after reaching feature complete.

### DIFF
--- a/enhancements/ingress/global-admission-hsts.md
+++ b/enhancements/ingress/global-admission-hsts.md
@@ -4,7 +4,7 @@ authors:
   - "@cholman"
 
 reviewers:
-  - "@danehans" 
+  - "@danehans"
   - "@frobware"
   - "@knobunc"
   - "@Miciah"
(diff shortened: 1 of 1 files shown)


enhancements/ingress/global-admission-hsts.md
```diff
@@ -542,12 +683,13 @@ $ ./batch_annotate.sh filename.txt
 #### Testing
 Accompanying unit and e2e test changes will be added to exercise the `RequiredHSTSPolicies` type.
 The admission plugin guarantees the following:
-* Routes with HSTS without a TLS termination policy of edge or re-encrypt will be rejected
+* Routes with HSTS without a TLS termination policy of edge or re-encrypt may retain the annotation, but will not apply the header
```
Miciah, line 686: It's HAProxy that would apply the header, so the language you have now in this item describes HAProxy behavior, not the admission plugin.  It should describe the plugin's behavior: 
```suggestion
* Routes with HSTS without a TLS termination policy of edge or re-encrypt will be accepted without checking the annotation.
```

candita: Thanks!

PR #850, opened 2021-07-29
Enable flows collection

mariomac:
This RFE defines the steps required to facilitate the enablement and collection
of network Flows' tracing from the user side, as well as the status retrieval.
It describes the user stories to switch on/off the flows both from the command-line
interface and the web console, and depicts the implementation details required in
the existing OVN-Kubernetes component, as well as the new components that should
be created.

Related JIRA issue: https://issues.redhat.com/browse/NETOBSERV-22


enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,307 @@
...
+
+However, we need to set this address dynamically to ensure that each pod will forward
+the flows to the collector in its same node.
+
+We need to extend the previous configuration to allow replacing the `collectors` property
+by one of the following properties:
+
+* `nodePort: (int)`: if set, the OVN will discover the node IP and submit the flows to the
+  port indicated by such property. E.g. `nodePort: 30023`.
+* `selector: (map)`: if set, the OVN will fetch for any `NodePort` matching
```
jotak, line 197: Can you clarify, this selector is used to match a `Service` object, right? Then to extract its `NodePort`. And further will use this port, but associated with its own host IP, not the IP of the service... is that correct?

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,307 @@
...
+
+#### Task 3: create a Flow operator to manage the flows status
+
+After this task is finished:
+* The operator can communicate with OVN-Kubernetes pods to patch their configuration.
+* There is an exposed API to perform the various operations: enable, disable, get status...
+* You can operate flows through the `oc` CLI command: enable, disable, get status...
+
+This task could involve setting new RBAC permissions to allow interaction between the
+Flow operator and the OVF-Kubernetes pods.
```
jotak, line 227: typo s/OVF/OVN/ I guess

enhancements/network/enable-flows-collection.md
mariomac: Yes. It's correct. I'll clarify it in the text

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,307 @@
...
+## Motivation
+
+Flows (NetFlow, IPFIX) emission, collection and processing may have an overall
+impact in the system performance. By this reason, we need to allow users to
+switch on and off the flows' tracing feature.
+
+It is also important to allow the users to know the status of the current flows
+tracing (on or off).
+
+### Goals
```
jotak, line 47: I think we should also add more controls in configuring flows emission, especially for fine-tuning the ipfix cache settings which can play an important role to mitigate potential issues with a too high ingestion rate.
Basically, ovs allows to control cache timeout and cache max flows that will modify the size of aggregates, hence the number of records sent to collector.

mariomac: Do you mean, making this options visible to the customer via UI or `oc`? In any case, I'm afraid that adding it here would increase too much the scope of the RFE. Maybe it's better to create another RFE with the additions to the UI as well as the changes required in the CNO to enable such configuration. What do you think?

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,352 @@
...
+In order to ensure an optimal usage of the network, we should restrict flow's
+emission-collection traffic to the local node.
+
+According to the [architectural description of the flows' emitter/collectors pipeline](
+https://docs.google.com/document/d/1kzNfTLXhMolu8VTcH0m9hmvFluN9yKMy7TSNIXJVTrA/edit#heading=h.kulf936ct32p),
+the collector will be deployed as a DaemonSet, to ensure that there will be an
+instance of the collector for each node.
+
+Exposing the collector endpoint as a Kubernetes ClusterIP Service would cause the traffic
+to be distributed across all the nodes (unless you enable the `ServiceInternalTrafficPolicy`
```
astoycos, line 195: `ServiceInternalTrafficPolicy` Is not yet implemented in  OVN-Kubernetes

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,352 @@
...
+cluster-wide feature gate, which at this moment is in alpha status, only available
+for the latest Kubernetes 1.21 version).
+
+To limit the flows' traffic to internal nodes, each collector should expose a node
+port, so each OVN instance only has to know the node IP and Port (which might be
+fixed or discovered).
+
+From the user-side configuration of the IPFix cache and sampling, we should use a `ConfigMap` and
```
astoycos, line 203: The config map would need to be exposed to the CNO, not ovn-k directly

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,352 @@
...
+kind: Network
+metadata:
+  name: cluster
+spec:
+  exportNetworkFlows:
+    collectors:
+      - 192.168.1.99:2056
+```
+
+However, we need to set this address dynamically to ensure that each pod will forward
```
astoycos, line 226: If we wan't to set this dynamically then it can no longer be a configurable field in the ovnkube-node daemonset, since currenlty whenever it's updated, all the ovnkub-node pods are deleted and rolled back out.

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,352 @@
...
+the flows to the collector in its same node.
+
+We need to extend the previous configuration to allow replacing the `collectors` property
```
astoycos, line 229: What's in charge of creating the collector service + associated collector daemonset? Is it the CNO, Cluster Logging Operator, or something else?

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,352 @@
...
+    role: flows-collector
+  ```
+
+#### Task 2: implement new discovery options in OVN-Kubernetes
+
+Before [OVN-Kubernetes invokes OpenVSwitch to enable flows](https://github.com/ovn-org/ovn-kubernetes/commit/ecc4047f5d0e732c267fccb2eabd64b5894b9f9a#diff-9c3b7372332fd16d60dba539a94dd0f819d262d1fbabd1e5d9c09e47da5af0ffR119-R128),
+if the network provider is configured for discovery, it needs to fetch the host/port and pass it
+to OpenVSwitch, in the `targets=` argument.
+
+If, by any reason, the Collector host:port can't be discovered (e.g. because the collector is not
```
astoycos, line 248: I don't think this is something that should be implemented in core OVN-K meaning  the health of the collector service is not something OVN-K should fetch and have to report

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,352 @@
...
+https://docs.google.com/document/d/1kzNfTLXhMolu8VTcH0m9hmvFluN9yKMy7TSNIXJVTrA/edit#heading=h.kulf936ct32p),
+the collector will be deployed as a DaemonSet, to ensure that there will be an
+instance of the collector for each node.
+
+Exposing the collector endpoint as a Kubernetes ClusterIP Service would cause the traffic
+to be distributed across all the nodes (unless you enable the `ServiceInternalTrafficPolicy`
+cluster-wide feature gate, which at this moment is in alpha status, only available
+for the latest Kubernetes 1.21 version).
+
+To limit the flows' traffic to internal nodes, each collector should expose a node
```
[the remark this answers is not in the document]
jotak: My understanding here is that we could use directly the host IP, not the service IP, to avoid LB. The Service here is not really mandatory, it's only used to declare the port to be used. Probably we shouldn't use a Service at all and declare the port in the exporter config, rather than having a diverted use of a Service.

enhancements/network/enable-flows-collection.md
mariomac: Clarifying it into the text.

enhancements/network/enable-flows-collection.md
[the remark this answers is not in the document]
mariomac: Joel is right. I'm rewriting it to avoid any misunderstanding.

enhancements/network/enable-flows-collection.md
mariomac: We, the Network Observability Team, will implement the new collectors.

enhancements/network/enable-flows-collection.md
mariomac: In principle, the new configuration options should be static (they just provide information to discover the host IP). I guess the discovery process should be triggered periodically to see if there is a new or different collector within the node, then change the openvswitch configuration. But the new `nodePort` and `selector` configuration options should never change.

enhancements/network/enable-flows-collection.md
mariomac: I see your point. Probably we could create another component to work as a health checker.

enhancements/network/enable-flows-collection.md
mariomac: After re-reading, I see that we should move this configuration to the same configmap that we will use to pass other configuration.

enhancements/network/enable-flows-collection.md
mariomac: Would this mean that, after an update in the config map, the CNO would need to Restart OVN-K to pass the new configuration? In principle is not a big issue as the configuration should be pretty stable.

jotak: Sorry I didn't read this section carefully. Operators usually don't rely on config maps, it's replaced with their CRDs, so I think it's the same here with the ipfix cache config, it should be configured via the `operator.openshift.io/v1` `Network` resource I guess (or doesn't it make sense for some reason?)

enhancements/network/enable-flows-collection.md
jotak: or the other way around :)

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,357 @@
...
+of network flows' tracing from the user side, as well as the status retrieval.
+It describes the user stories to switch on/off the flows both from the command-line
+interface and the web console, and depicts the implementation details required in
+the existing OVN-Kubernetes component, as well as the new components that should
+be created. We consider IPFix as the unique protocol to be supported by now, but
+we might also consider using other protocols like Netflow.
+
+## Motivation
+
+Flows (IPFIX) emission, collection and processing may have an overall
```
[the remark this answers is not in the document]
mariomac: For an initial version, our plan was to enable it for the entire cluster. Of course our product roadmap is always under constant refinement but as far as I know it is not planned yet @jotak @stleerh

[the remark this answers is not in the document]
jotak: Yes, this is how it currently works actually, regardless this enhancement proposal: flow exports can already be turned on but it's cluster-wide (as per [this enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/network/netflow.md)). We will surely need something more fine-grained at some point but the idea was to start without changing this part.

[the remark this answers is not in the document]
amorenoz: Understood, thanks @mariomac @jotak

jotak:
One question I have, maybe for @astoycos : is it possible to trigger on/off flow collection without restarting the ovn-k pods? If I'm correct, the current process is:

1. Change detected in Network CR by CNO, triggering change in the daemonsets' pod spec
2. Pod spec change triggers pods restart with new flow export config
3. On ovn-k pod startup, reading the config results in executing ovs-vsctl command to configure flow export in OvS

So would it be possible to run this ovs-vsctl command not based on pod spec (env), but through another mechanism that spares us from restarting the pods? wdyt?

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,362 @@
...
+
+* `sampling`: optional integer, in range 1 to 4,294,967,295
+  The  rate  at  which  packets should be sampled and sent to each
+  target collector. If not specified, defaults to 400, which means
+  one  out of 400 packets, on average, will be sent to each target
+  collector.
+
+Other arguments might be considered, such as `enable-input-sampling`, `enable-output-sampling`,
+or `enable-tunnel-sampling`.
+
```
stleerh, line 182: Let's add sampling as one of the parameters that can be configured in the CLI and Web Console.

enhancements/network/enable-flows-collection.md
```diff
@@ -0,0 +1,362 @@
...
+    collectors:
+      - 192.168.1.99:2056
+```
+
+However, we need to set this address dynamically to ensure that each pod will forward
+the flows to the collector in its same node.
+
+While the above `Network` configuration is still valid for the current use cases, the Network
+Observability components will provide a new CRD whose configuration is exposed to the CNO, which
+will use it to configure the OVN-Kubernetes operator. For example:
```
jotak, line 234: I've drawn a diagram that would summarize the interaction between the CNO and our new NOO (network observability operator) : https://docs.google.com/drawings/d/1zTbGOHUZlm3mKUcnb3s7SdXQbJcyoOIGieDzYOjxlxg/edit

Basically, here's how it would work:

NOO creates a ConfigMap for OvS, and CNO is able to understand it and will be watching it for changes.

For the CNO, this config map content sums up with any OvS configuration that potentially exists in the Network CR. For example, if a user has configured an IP for netflow exports in the Network CR AND has configured a Netobs CR for the NOO, OvS will export flows to both destinations.

Currently, the CNO interacts with the OVN pods via the DaemonSet definition and environment variables set on some of the OVN containers. New environment variables would be introduced to account for the new configuration needed by the NOO.

The OVN pods need to use this new config while setting up the OvS flow exports.

enhancements/network/ovs-flow-export-configuration.md
```diff
@@ -0,0 +1,261 @@
...
+
+However, we need to set this address dynamically to ensure that each pod will forward
+the flows to the collector in its same node.
+
+While the above `Network` configuration is still valid for the current use cases, the NOO will generate a new `ConfigMap` that is exposed to the CNO, which
+will use it to configure the OVN pods. The CNO has to watch for any change in this configuration.
+
+See [diagrams](https://docs.google.com/document/d/1XxaFl4Gl1VbClfcPngkFCETv0anuFO6Klowh8MFZtJc/edit#) showing this interaction.
+
+The `ConfigMap` content sums up with any configuration that is already present in the Network CR. For example, if a user has configured an IP for netflow exports in the Network CR AND has configured a Netobs CR for the NOO, OvS will export to both destinations.
```
martinkennelly, line 174: What is meant by "sums up"? Do you mean you will take information present in Network CR and put it also in the config map or do you mean the CNO will implement what's specified in the network CR and the configmap?

enhancements/network/ovs-flow-export-configuration.md
```diff
@@ -0,0 +1,261 @@
...
+  sampling: <int>
+  cacheActiveTimeout: <int (seconds)>
+  cacheMaxFlows: <int>
+```
+
+Assuming that each OVS instance will forward flows to the same node (to minimize network traffic
+between nodes), the `nodePort` property will tell the CNO which port of its node should send data
+to. The node IP is discoverable by OVN pods or the CNO.
+
+`nodePort` and `sharedTarget` are mutually exclusive: the former will be set by the NOO when the collector is deployed as a DaemonSet, the latter will be set when deployed as Deployment + Service. So, when `sharedTarget` is used, the CNO will simply need to copy it to the ovn-k pods environment variable for IPFIX export that already exists.
```
martinkennelly, line 196: Forgive my ignorance here, but I thought OvS is running directly on a node and not within a pod. How can it export flows to a cluster service IP that is defined for sharedTarget?

enhancements/network/ovs-flow-export-configuration.md
```diff
@@ -0,0 +1,261 @@
...
+
+The `ConfigMap` content sums up with any configuration that is already present in the Network CR. For example, if a user has configured an IP for netflow exports in the Network CR AND has configured a Netobs CR for the NOO, OvS will export to both destinations.
+
+Here is the new `ConfigMap` definition:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ovs-ipfix-config
```
martinkennelly, line 182: This config map is exclusively for ipfix config? If its not, I would group and indent the ipfix specific options. I believe cache* options are ipfix exclusive right?

enhancements/network/ovs-flow-export-configuration.md
jotak: I'm not sure *how* it works internally but I can tell that it works since it's how I'm using it for a bunch of months now: deploying goflow with a service, grabbing its cluster IP, and setting it to `OVN_IPFIX_TARGETS` on `ovnkube-node` daemonset

enhancements/network/ovs-flow-export-configuration.md
jotak: The idea here is to have it exclusively for ipfix, yes. This is what we need for this enhancement. But maybe we can change as you suggest, in case we want also to support netflow v5 later. I'll do that (and rename the CM accordingly)

enhancements/network/ovs-flow-export-configuration.md
jotak: (by the way.. forgive my English, with "sums up" I actually meant "adds up")

enhancements/network/ovs-flow-export-configuration.md
```diff
@@ -0,0 +1,261 @@
...
+      - 192.168.1.99:2056
+```
+
+However, we need to set this address dynamically to ensure that each pod will forward
+the flows to the collector in its same node.
+
+While the above `Network` configuration is still valid for the current use cases, the NOO will generate a new `ConfigMap` that is exposed to the CNO, which
+will use it to configure the OVN pods. The CNO has to watch for any change in this configuration.
+
+See [diagrams](https://docs.google.com/document/d/1XxaFl4Gl1VbClfcPngkFCETv0anuFO6Klowh8MFZtJc/edit#) showing this interaction.
```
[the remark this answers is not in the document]
dhellmann: Google Apps documents are not visible to anyone who is not a member of the organization, so we try to avoid using links to them or embedded images from them in design docs here. Please copy all images to this repo. You can create a sub-folder to hold this enhancement and its images together to help keep things organized.

enhancements/network/ovs-flow-export-configuration.md
knobunc: A process running on the node can still access the service IP address.  The host sends the service ip range in to ovn.

enhancements/network/ovs-flow-export-configuration.md
```diff
@@ -0,0 +1,267 @@
...
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ovs-flows-config
+  namespace: # CNO's namespace
+spec:
+  ipfix:
+    sharedTarget: <service ip+port>
```
astoycos, line 184: This `sharedTarget` meaning here is a little confusing, maybe adding a diagram showing that it will connect to the goflow collector daemonset pod for each node would help (something like what you made [here](https://docs.google.com/document/d/1kzNfTLXhMolu8VTcH0m9hmvFluN9yKMy7TSNIXJVTrA/edit#heading=h.7yb4abpdeo6w))

I don't think this needs to block the kep from merging though and can be changed in a follow up commit if needed

PR #854, opened 2021-08-03
CONVENTIONS: Daemonsets should use maxAvailability 10%

dankenigsberg:
@smarterclayton's statement in https://bugzilla.redhat.com/show_bug.cgi?id=1920209#c14 is important enough to be documented as a convention in this document. Daemonsets should use `maxAvailability: 10%`.

According to @sdodson 10% generally reduces a 2.5hr daemonset rollout to something like 20 minutes in 250 node clusters.
Further optimization may change the somewhat-arbitrary value of 10%, but as of OpenShift 4.8 it is working fine.

Thanks @tiraboschi for pointing me to it.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -199,6 +199,17 @@ To avoid disruption and mass pod-death, it is important to
   * Components that support workloads directly must not disrupt end-user workloads during upgrade or reconfiguration
     * E.g. the upgrade of a network plugin must serve pod traffic without disruption (although tiny increases in latency are allowed)
     * All components that currently disrupt end-user workloads must prioritize addressing those issues, and new components may not be introduced that add disruption
+* All dameonsets of OpenShift components should use the maxUnavailable rollout strategy, allowing up to 10% of instances to be taken down at once.
+Any bugs that block that should be fixed.
+`10%` is an arbitrary ratio that ensures that most of the cluster is up and
+serving clients, while a big-enough chunk of the cluster is upgradable.
+There wasn't a huge amount of analysis beyond the fact that it significantly
+improved upgrade times and hit our targets. Additionally, 10% is roughly in
+line with the minimum disruption experienced by the default sized cluster
+whenever we reboot nodes and that's mandatory in order to complete an upgrade.
+Keeping the default of `1` makes an operator upgrade unacceptably slowly on
+big clusters because only one host running the daemonset can be drained (and upgraded) at a time.
+
 
 #### Priority Classes
 


tiraboschi (approved):

dhellmann (changes requested):
As we add rules like this, we need to include enough information for the reader to understand where each rule came from and why it is in place. Do we have that information for this rule?

sdodson:
It's done so that widely deployed Daemonsets don't take ages to upgrade in large clusters. 10% generally reduces a 2.5hr daemonset rollout to something like 20 minutes in 250 node clusters. BTW, we mention 25% on line 140. I think that's the new guidance, I've not measured the impact but I wouldn't assume it makes a huge difference.

dankenigsberg:
> It's done so that widely deployed Daemonsets don't take ages to upgrade in large clusters. 10% generally reduces a 2.5hr daemonset rollout to something like 20 minutes in 250 node clusters. BTW, we mention 25% on line 140. I think that's the new guidance, I've not measured the impact but I wouldn't assume it makes a huge difference.

I think that the 25% refers to deployments, not daemonsets, hence the need for an explicit statement. I chose 10% as an [ipse dixit](https://en.wikipedia.org/wiki/Ipse_dixit) not because it is better or worse than 25%.

CONVENTIONS.md
```diff
@@ -199,6 +199,12 @@ To avoid disruption and mass pod-death, it is important to
   * Components that support workloads directly must not disrupt end-user workloads during upgrade or reconfiguration
     * E.g. the upgrade of a network plugin must serve pod traffic without disruption (although tiny increases in latency are allowed)
     * All components that currently disrupt end-user workloads must prioritize addressing those issues, and new components may not be introduced that add disruption
+* All dameonsets in OpenShift should be at `maxUnavailable: 10%`.
+Any bugs that block that should be fixed.
+`10%` is an arbitrary ratio that ensures that most of the cluster is up and
+serving clients, while a big-enough chunk of the cluster is upgradable.
+Keeping the default of `1` makes your operator upgrade unacceptably slowly on
```
dhellmann, line 206:
```suggestion
Keeping the default of `1` makes an operator upgrade unacceptably slowly on
```

CONVENTIONS.md
```diff
@@ -199,6 +199,12 @@ To avoid disruption and mass pod-death, it is important to
   * Components that support workloads directly must not disrupt end-user workloads during upgrade or reconfiguration
     * E.g. the upgrade of a network plugin must serve pod traffic without disruption (although tiny increases in latency are allowed)
     * All components that currently disrupt end-user workloads must prioritize addressing those issues, and new components may not be introduced that add disruption
+* All dameonsets in OpenShift should be at `maxUnavailable: 10%`.
```
dhellmann, line 202:
```suggestion
* All dameonsets of OpenShift components should use the maxUnavailable rollout strategy, allowing up to 10% of instances to be taken down at once.
```

CONVENTIONS.md
```diff
@@ -199,6 +199,12 @@ To avoid disruption and mass pod-death, it is important to
...
+Any bugs that block that should be fixed.
+`10%` is an arbitrary ratio that ensures that most of the cluster is up and
+serving clients, while a big-enough chunk of the cluster is upgradable.
+Keeping the default of `1` makes your operator upgrade unacceptably slowly on
+big clusters.
```
dhellmann, line 207:
```suggestion
big clusters because only one host running the daemonset can be drained (and upgraded) at a time.
```

dhellmann (commented):
The additional detail looks good, thank you.

I proposed some wording changes for tone and detail.

Do we need to have a discussion of the 10% value?

sdodson:
> Do we need to have a discussion of the 10% value?

10% is fine with me and as Dan mentioned it's specifically what Clayton suggested.

CONVENTIONS.md
sdodson:
```suggestion
* All rollingUpdate dameonsets in OpenShift should utilize `maxUnavailable: 10%` to ensure a timely rollout on large clusters.
```
The phrase "rollout strategy" seems a bit too close to the spec field names in my opinion. here's what it actually looks like
```
spec:
  updateStrategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 10%
```

dhellmann: Would it make sense to put that YAML snippet right into the text?

dhellmann:
> > Do we need to have a discussion of the 10% value?
> 
> 10% is fine with me and as Dan mentioned it's specifically what Clayton suggested.

"Because Clayton said" isn't good enough justification unless there's an explanation backing it up.

sdodson:
> "Because Clayton said" isn't good enough justification unless there's an explanation backing it up.

I thought you were just asking if we needed to debate 10% vs 25% in this PR, which if you are then I'd just say that 10% is what we have in 4.8 and that's gotten us to our target of completing the upgrade of cluster scoped resources within two hours. It's not clear that going to 25% would improve things in anything other than clusters that are 4-10 nodes in size and those already hit well under the two hour mark.

I agree we should explain why we're recommending changes from the default of `maxUnavailable: 1`.

dhellmann:
> > "Because Clayton said" isn't good enough justification unless there's an explanation backing it up.
> 
> I thought you were just asking if we needed to debate 10% vs 25% in this PR, which if you are then I'd just say that 10% is what we have in 4.8 and that's gotten us to our target of completing the upgrade of cluster scoped resources within two hours. It's not clear that going to 25% would improve things in anything other than clusters that are 4-10 nodes in size and those already hit well under the two hour mark.
> 
> I agree we should explain why we're recommending changes from the default of `maxUnavailable: 1`.

Yeah, sorry, my reply came across a bit brusque.

If we just want to say something like "we picked 10% at 4.8 and will review it based on experience", that's probably OK. It would be useful to know how someone picked 10%, even if it was just plucked out of thin air as a starting point, so that whoever does that future review has some basis for understanding how hard we tried to get the right value at the beginning and what factors were considered.

dankenigsberg:
> I agree we should explain why we're recommending changes from the default of maxUnavailable: 1.

@sdodson I believe that the current text does so. Let me know if something is still missing.

sdodson:
There wasn't a huge amount of analysis beyond the fact that it significantly improved upgrade times and hit our targets. Additionally, 10% is roughly in line with the minimum disruption experienced by the default sized cluster whenever we reboot nodes and that's mandatory in order to complete an upgrade.

dhellmann:
> There wasn't a huge amount of analysis beyond the fact that it significantly improved upgrade times and hit our targets. Additionally, 10% is roughly in line with the minimum disruption experienced by the default sized cluster whenever we reboot nodes and that's mandatory in order to complete an upgrade.

OK, let's put that info into the body of the doc.

dankenigsberg:
> OK, let's put that info into the body of the doc.

It somewhat repeats what I already written there, but I can tolerate this. If you prefer a more succinct text, please suggest it.

dhellmann:
I'm happy with this version. @sdodson may want one more review.

/approve

sdodson:
/lgtm
I think it's clear enough for anyone that would be reading this content.

PR #855, opened 2021-08-03
 ingress: Add bind-options enhancement

m-yosefpor:
This enhancement extends the IngressController API to allow the user to
specify bindOptions in HostNetwork strategy and enables them to run multiple
instances of ingresses on the same node in the HostNetwork strategy, with
mitigating port binding conflicts.

**Implementation:**
openshift/api: https://github.com/openshift/api/pull/965
openshift/cluster-ingress-operator: https://github.com/openshift/cluster-ingress-operator/pull/634

### DIFF
--- a/enhancements/ingress/bind-options.md
+++ b/enhancements/ingress/bind-options.md
@@ -1,0 +1,327 @@
+---
+title: bind-options
+authors:
+  - "@m-yosefpor"
+reviewers:
+  - "@alebedev87"
+  - "@Miciah"
+  - "@frobware"
+approvers:
+  - @knobunc
+  - @tjungblu
+creation-date: 2021-08-04
+last-updated: 2021-08-04
+status: implementable
+see-also:
+  - https://github.com/openshift/cluster-ingress-operator/issues/633
+  - https://github.com/openshift/api/issues/964
+replaces:
+superseded-by:
+---
+
+# Ingress Bind Options
+
+## Release Signoff Checklist
+
+- [X] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement extends the IngressController API to allow the users to
+specify bind options in the HostNetwork strategy and enables them to run multiple
+instances of the ingress controller on the same node using the HostNetwork strategy, therefore
+mitigating port binding conflicts.
+
+## Motivation
+
+When using the HostNetwork strategy for ingress controllers, the default 80, 443, 10443, 10444, 1936 ports on the host are bound by HAProxy for http, https, no_sni, sni and stats correspondingly.
+However, those ports might be occupied by other processes (such as another set of ingress controllers), which makes it impossible to run multiple sets of ingress controllers on the same nodes with HostNetwork strategy.
+In OpenShift 3.11, it was possible to listen on custom host ports via setting these environment variables in the router's DeploymentConfig ([documentation link](https://docs.openshift.com/container-platform/3.11/architecture/networking/routes.html#env-variables)):
+
+    - `ROUTER_SERVICE_HTTP_PORT=80`
+    - `ROUTER_SERVICE_HTTPS_PORT=443`
+    - `ROUTER_SERVICE_SNI_PORT=10444`
+    - `ROUTER_SERVICE_NO_SNI_PORT=10443`
+    - `STATS_PORT=1936`
+
+However there is no option to specify custom ports in `IngressController.operator.openshift.io/v1` object right now.
+
+Having multiple sets of ingress controllers is useful for the router sharding, routers with different policies (public, private), routers with different configuration options, etc.
+Running in HostNetwork is a strict requirement in some scenarios (e.g. environments with custom PBR rules). Also routers with the HostNetwork strategy have shown to have better performance (see [OKD4 benchmarks](https://docs.okd.io/latest/scalability_and_performance/routing-optimization.html))
+
+Using custom node selectors to ensure different ingress controllers run on different nodes is not always feasible, when there are not many nodes in the cluster.
+
+### Goals
+
+
+### Goal
+
+Enable cluster administrators to run multiple set of ingress controllers on the same node with `HostNetwork` endpoint publishing strategy, by configuring the bindings for the following ports:
+- HTTP
+- HTTPS
+- Stats
+
+### Non-Goals
+
+1. Enabling cluster administrators to configure custom ports on the ingress controllers that use any endpoint publishing strategy other than `HostNetwork`.
+2. Enabling cluster administrators to configure custom ports on the ingress controllers for `SNI` and `NO_SNI` ports.
+
+The reasons for number 2 to be a non-goal are:
+
+- The sysadmin’s goal is really only to configure alternative ports for 80/443/1936 and to be able to point an external LB to these ports.
+- Requiring the sysadmin to configure these additional ports would mean we would need additional documentation to explain why the sysadmin needed to care about something that is really an implementation detail of the HAProxy configuration (and this internal implementation detail could conceivably change, in which case the API would become cruft).
+- The larger API surface comes with an increased risk of misconfiguration and port conflicts that an automated solution could prevent.
+- Requiring 5 ports instead of just 3 reduces scalability for users who want to run a large number of host-network router pods on the same node.
+- Requiring HAProxy-specific configuration makes migration to another proxy (such as Contour) less straightforward.
+
+
+## Proposal
+
+To enable cluster administrators to configure bind options on the ingress controllers that use the `HostNetwork` endpoint publishing strategy:
+the IngressController API is extended by adding an optional `BindOptions` field with type `*IngressControllerBindOptions` to the `HostNetworkStrategy` struct:
+
+    ```go
+    // HostNetworkStrategy holds parameters for the HostNetwork endpoint publishing
+    // strategy.
+    type HostNetworkStrategy struct {
+        // ...
+
+        // bindOptions defines parameters for binding haproxy in ingress controller pods.
+        // All fields are optional and will use their respective defaults if not set.
+        // See specific bindOptions fields for more details.
+        //
+        //
+        // Setting fields within bindOptions is generally not recommended. The
+        // default values are suitable for most configurations.
+        //
+        // +optional
+        BindOptions *IngressControllerBindOptions `json:"bindOptions,omitempty"`
+    }
+    ```
+
+`IngressControllerBindOptions` is the following struct:
+
+    ```go
+    // IngressControllerBindOptions specifies options for binding haproxy in ingress controller pods
+    type IngressControllerBindOptions struct {
+        // Ports are used to set the ports on which the ingress controller will be bound.
+        //
+        // If unset, the default values are applied as follows:
+        //
+        //   HTTP: 80
+        //   HTTPS: 443
+        //   STATS: 1936
+        //
+        // +kubebuilder:validation:Optional
+        // +optional
+        Ports *IngressControllerPorts `json:"ports,omitempty"`
+    }
+
+    // IngressControllerPorts specifies ports on which ingress controller pods will be bound
+    type IngressControllerPorts struct {
+        // http defines the port number which HAProxy process binds for
+        // http connections. Setting this field is generally not recommended. However in
+        // HostNetwork strategy, default http 80 port might be occupied by other processess
+        //
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:validation:Minimum=1
+        // +kubebuilder:validation:Maximum=30000
+        // +kubebuilder:default:=80
+        // +optional
+        HTTP int32 `json:"http,omitempty"`
+
+        // https defines the port number which HAProxy process binds for
+        // https connections. Setting this field is generally not recommended. However in
+        // HostNetwork strategy, default https 443 port might be occupied by other processess
+        //
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:validation:Minimum=1
+        // +kubebuilder:validation:Maximum=30000
+        // +kubebuilder:default:=443
+        // +optional
+        HTTPS int32 `json:"https,omitempty"`
+
+        // stats is the port number which HAProxy process binds
+        // to expose statistics on it. Setting this field is generally not recommended.
+        // However in HostNetwork strategy, default stats port 1936 might
+        // be occupied by other processess
+        //
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:validation:Minimum=1
+        // +kubebuilder:validation:Maximum=30000
+        // +kubebuilder:default:=1936
+        // +optional
+        Stats int32 `json:"stats,omitempty"`
+    }
+    ```
+
+The following example configures two ingress controllers with the HostNetwork strategy that can run on the same nodes without port conflicts:
+
+    ```yaml
+    ---
+    apiVersion: operator.openshift.io/v1
+    kind: IngressController
+    metadata:
+    name: default-1
+    namespace: openshift-ingress-operator
+    spec:
+    endpointPublishingStrategy:
+        type: HostNetwork
+    ---
+    apiVersion: operator.openshift.io/v1
+    kind: IngressController
+    metadata:
+    name: default-2
+    namespace: openshift-ingress-operator
+    spec:
+    endpointPublishingStrategy:
+        type: HostNetwork
+        hostNetwork:
+            bindOptions:
+                ports:
+                    http: 11080
+                    https: 11443
+                    stats: 1937
+    ```
+
+### Validation
+
+Specifying `spec.endpointPublishingStrategy.type: HostNetwork` and omitting
+`spec.endpointPublishingStrategy.hostNetwork` or
+`spec.endpointPublishingStrategy.hostNetwork.bindOptions` is valid and implies
+the default behavior, which is to use the default ports to bind to. The API validates
+that any value provided for the ports defined in
+`spec.endpointPublishingStrategy.hostNetwork.bindOptions` is greater than `1`, and lower than `30000` (to avoid conflicts with 30000-32767 nodePort services range).
+
+
+### User Stories
+
+#### As a cluster administrator, I need to configure bind ports for my IngressController
+
+To satisfy this use-case, the cluster administrator can set the ingress controller's `spec.endpointPublishingStrategy.hostNetwork.bindOptions`.
+
+
+### API Extensions
+
+This enhancement proposal modifies `IngressController.operator.openshift.io/v1`.
+
+### Implementation Details
+
+Implementing this enhancement requires changes in the following repositories:
+
+* openshift/api
+* openshift/cluster-ingress-operator
+
+OpenShift Cluster Ingress Operator, creates a deployment for Router with environment variables for port bindings which OpenShift Router already respects:
+`ROUTER_SERVICE_HTTP_PORT`, `ROUTER_SERVICE_HTTPS_PORT`, `STATS_PORT`.
+
+Also the deployments should not have ip:port conflicts for `SNI_PORT` and `NO_SNI_PORT`. There are multiple ways to achieve this:
+
+- The cluster ingress operator configures the router to use a unique loopback address for the sni/nosni frontends. This loopback address would be determined deterministically, for example by hashing the ingresscontroller’s name.
+As a result, the sni/nosni port numbers wouldn’t need to change to avoid the conflict as the routers would be using different IP addresses. This can be done with the introduction of two new environment variables for the router:`ROUTER_SERVICE_SNI_IP`, `ROUTER_SERVICE_SNI_IP`, and the configuration of the frontend to bind on this IPs.
+The cluster ingress operator sets these environment variables for the router deployment from `127.0.0.1/8` CIDR based on the ingress controller name hash.
+- The router gets rid of the loopback hop using a Unix domain socket or other solution, which would have other advantages as well ([router's PR](https://github.com/openshift/router/pull/326)).
+
+### Risks and Mitigations
+
+#### Conflicting ports
+
+Users might define conflicting ports which would cause HAProxy process to fail at startup.
+
+A mitigation to this risk is to implement a validation feature in the reconciliation loop of the Cluster Ingress Operator
+to ensure all the ports defined in the `bindOptions` section are unique, and return an error with a meaningful message if there are conflicting ports.
+
+#### Usage of X-Forwarded headers
+
+The default forwarded header policy of the ingress controller is to append X-Forwarded headers. This results into the insertion of `X-Forwarded-Port` header on OpenShift router side (HAProxy acts as a reverse proxy).
+If this was the first insertion (L4 load balancer in front of the router) the router's port will be the only one seen by the server (application inside OpenShift).
+And if the server relies on `X-Forwarded-*` headers (like OpenShift image registry for instance) this may give a wrong impression that the client used the port set in `X-Forwarded-Port` which in case of the custom port binding may be different.
+
+A mitigation to this risk is to change the forwarded header policy on the ingress controller or on the route level: [documentation](https://docs.openshift.com/container-platform/4.8/networking/ingress-operator.html#nw-using-ingress-forwarded_configuring-ingress).
+
+## Design Details
+
+### Test Plan
+
+The controller that manages the ingress controller Deployment and related
+resources has unit-test coverage; for this enhancement, the unit tests are
+expanded to cover the additional functionality:
+
+1. Create an ingress controller that enables the `HostNetwork` endpoint publishing strategy type without specifying `spec.endpointPublishingStrategy.hostNetwork.bindOptions`.
+2. Verify that the IngressController configures:
+    - `ROUTER_SERVICE_HTTP_PORT=80`
+    - `ROUTER_SERVICE_HTTPS_PORT=443`
+    - `STATS_PORT=1936`
+
+3. Update the ingress controller to specify `spec.endpointPublishingStrategy.hostNetwork.bindOptions`.
+4. Verify that the ingress controller updates the router deployment to specify the corresponding values for
+`ROUTER_SERVICE_HTTP_PORT`, `ROUTER_SERVICE_HTTPS_PORT`, `STATS_PORT`.
+
+
+The operator has end-to-end tests; for this enhancement, the following test can
+be added:
+
+1. deploy host network ingress controller with the default port bindings
+2. deploy another host network controller with the same node placements but with `bindOptions` field set with the ports not conflicting with the first ingress controller
+3. check that the second ingress controller becomes ready
+4. check that the first ingress controller is still ready
+
+### Graduation Criteria
+
+N/A.
+
+#### Dev Preview -> Tech Preview
+
+N/A.  This feature will go directly to GA.
+
+#### Tech Preview -> GA
+
+N/A.  This feature will go directly to GA.
+
+#### Removing a deprecated feature
+
+N/A.  We do not plan to deprecate this feature.
+
+### Upgrade / Downgrade Strategy
+
+On upgrade, the default configuration remains in effect.
+
+If a cluster administrator upgraded to 4.9, then use different ports for bindOptions, and then downgraded to 4.8, the downgrade would turn all the ingress controller ports with HostNetwork strategy back to default ports.
+The administrator would be responsible for making sure there are not any port conflicts when downgrading to OpenShift 4.8.
+
+
+### Version Skew Strategy
+
+N/A.
+
+### Operational Aspects of API Extensions
+
+N/A.
+
+#### Failure Modes
+
+N/A.
+
+#### Support Procedures
+
+N/A.
+
+## Implementation History
+
+This enhancement is being implemented in OpenShift 4.9.
+
+In OpenShift 3.3 to 3.11 it was possible to use `ROUTER_SERVICE_HTTP_PORT`, `ROUTER_SERVICE_HTTPS_PORT`, `ROUTER_SERVICE_SNI_PORT`, `ROUTER_SERVICE_NO_SNI_PORT`, `STATS_PORT` environment variables.
+
+
+## Alternatives
+
+N/A
+
+
+## Drawbacks
+
+N/A


enhancements/ingress/bind-options.md
```diff
@@ -0,0 +1,268 @@
...
+## Summary
+
+This enhancement extends the IngressController API to allow the user to
+specify bindOptions in HostNetwork strategy and enables them to run multiple
+instances of ingresses on the same node in the HostNetwork strategy, with
+mitigating port binding conflicts.
+
+## Motivation
+
+When using HostNetwork strategy for ingressControllers, the default 80, 443, 10443, 10444, 1936 ports on the host are binded on HAProxy for http, https, no_sni, sni and stats correspondingly. However, those ports might be occupied by other processes (such as another set of ingressControllers), which makes it impossible to run multiple set of ingressControllers on the same nodes with HostNetwork strategy. In OKD3.11, it was possible to listen on custom host ports via setting these env vars in Routers DeploymentConfig, however there is not any options to specify custom ports in `IngressController.operator.openshift.io/v1` object right now.
```
alebedev87, line 38: Can you mention which environment variables were used in OpenShift 3.11? Links to the documentation would also we helpful.

Also, some minor remarks: `s/binded on HAProxy/bound by HAProxy/`, `s/to run on multiple set/to run on multiple sets/`, `s/OKD3.11/OpenShift 3.11/`.

enhancements/ingress/bind-options.md
```diff
@@ -0,0 +1,268 @@
...
+Having multiple set of ingressControllers is useful for router sharding, routers with different policies (public, private), routers with different configuration options, etc. Running in HostNetwork is a strict requirement in some scenarios (e.g. environemtns with custom PBR rules). Also there might be some performance considerations to run routers with HostNetwork strategy.
+
+Using custom nodeSelectors to ensure different ingressControllers runs on different nodes is not always feasible, when there are not many nodes in the cluster.
+
+### Goals
+
+
+### Goal
+
+Enable cluster administrators to configure the following ports bindings on IngressControllers that use the "HostNetwork" endpoint publishing strategies:
```
alebedev87, line 49:
```suggestion
Enable cluster administrators to configure IngressControllers which use `HostNetwork` endpoint publishing strategy with the bindings for the following ports:
```

enhancements/ingress/bind-options.md
```diff
@@ -0,0 +1,268 @@
...
+- NoSNI
+- Stats
+
+### Non-Goals
+
+1. Enabling cluster administrators to configure custom ports on IngressControllers that use any endpoint publishing strategy other than "HostNetwork".
+
+## Proposal
+
+To enable cluster administrators to configure IngressControllers to configure bindOptions on IngressControllers that use the "HostNetwork" endpoint publishing strategies:, the IngressController API is extended by adding an optional `BindOptions` field with type `*IngressControllerBindOptions` to the `HostNetworkStrategy` struct:
```
alebedev87, line 62:
```suggestion
To enable cluster administrators to configure IngressControllers to configure bindOptions on IngressControllers that use the "HostNetwork" endpoint publishing strategies: the IngressController API is extended by adding an optional `BindOptions` field with type `*IngressControllerBindOptions` to the `HostNetworkStrategy` struct:
```

enhancements/ingress/bind-options.md
```diff
@@ -0,0 +1,268 @@
...
+        sniPort: 11444
+        noSniPort: 11445
+        statsPort: 1937
+```
+
+### Validation
+
+Specifying `spec.endpointPublishingStrategy.type: HostNetwork` and omitting
+`spec.endpointPublishingStrategy.hostNetwork` or
+`spec.endpointPublishingStrategy.hostNetwork.bindOptions` is valid and specifies
```
alebedev87, line 179:
```suggestion
`spec.endpointPublishingStrategy.hostNetwork.bindOptions` is valid and implies
```

enhancements/ingress/bind-options.md
```diff
@@ -0,0 +1,268 @@
...
+
+
+### Implementation Details
+
+Implementing this enhancement requires changes in the following repositories:
+
+* openshift/api
+* openshift/cluster-ingress-operator
+
+OpenShift Cluster Ingress Operator, creates a deployment for Router with environment variables for port bindings which Openshift Router already respects. `ROUTER_SERVICE_HTTP_PORT`, `ROUTER_SERVICE_HTTPS_PORT`, `ROUTER_SERVICE_SNI_PORT`, `ROUTER_SERVICE_NO_SNI_PORT`, `STATS_PORT`.
```
alebedev87, line 200:
```suggestion
OpenShift Cluster Ingress Operator, creates a deployment for Router with environment variables for port bindings which OpenShift Router already respects: `ROUTER_SERVICE_HTTP_PORT`, `ROUTER_SERVICE_HTTPS_PORT`, `ROUTER_SERVICE_SNI_PORT`, `ROUTER_SERVICE_NO_SNI_PORT`, `STATS_PORT`.
```

enhancements/ingress/bind-options.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Risks and Mitigations
+
+
+Users might define conflicting ports which causes haproxy process fail to start. A mitigation to this risk is to implement a validation feature in cluster-ingress-operator in reconsiler loop to ensure all the ports defined in the `bindOptions` section are unique, and return err if there are conflicting ports with a meaningful error message.
```
alebedev87, line 205:
```suggestion
Users might define conflicting ports which would cause HAProxy process to fail at startup. A mitigation to this risk is to implement a validation feature in the reconciliation loop of the Cluster Ingress Operator to ensure all the ports defined in the `bindOptions` section are unique, and return an error with a meaningful message if there are conflicting ports.
```

enhancements/ingress/bind-options.md
```diff
@@ -0,0 +1,268 @@
...
+
+Also if the underlying IngressController implementation were to change away from HAProxy to a different implementation, some values such as `sniPort` and `noSniPort` might not be needed and loose their meaning. Also the `statsPort` teminology might be different for other tools such as `Envoy`, which `metricPort` might be more meaningful in that case.
```
alebedev87, line 207: I agree about `sni` and `nosni` names which are pure HAProxy things but `stats` was already chosen to be a generic name detached from any concrete implementation.
BTW, it looks like Envoy is also favoring the name statistics: [statistics architecture](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/observability/statistics#arch-overview-statistics).

enhancements/ingress/bind-options.md
```diff
@@ -0,0 +1,268 @@
...
+## Design Details
+
+### Test Plan
+
+The controller that manages the IngressController Deployment and related
+resources has unit-test coverage; for this enhancement, the unit tests are
+expanded to cover the additional functionality.
+
+The operator has end-to-end tests; for this enhancement, the following test can
+be added:
```
alebedev87, line 219: What you described below is actually the unit test that you did. The e2e tests typically involve the deployment of the ingress controllers on the real Kubernetes/OpenShift cluster.

`cluster-ingress-operator` repository has a dedicated directory for e2e tests and it already has a scenario for the host network type of ingress controller: [here](https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/operator_test.go#L727). I was thinking that this e2e test case can be updated or a new one can be created to try a scenario similar to this:
- deploy host network ingress controller with the default port bindings
- deploy another host network controller with the same node placements but with `bindOptions` field set with the ports not conflicting with the first ingress controller
- check that the second ingress controller becomes ready
- check that the first ingress controller is still ready

enhancements/ingress/bind-options.md
```diff
@@ -0,0 +1,268 @@
...
+instances of ingresses on the same node in the HostNetwork strategy, with
+mitigating port binding conflicts.
+
+## Motivation
+
+When using HostNetwork strategy for ingressControllers, the default 80, 443, 10443, 10444, 1936 ports on the host are binded on HAProxy for http, https, no_sni, sni and stats correspondingly. However, those ports might be occupied by other processes (such as another set of ingressControllers), which makes it impossible to run multiple set of ingressControllers on the same nodes with HostNetwork strategy. In OKD3.11, it was possible to listen on custom host ports via setting these env vars in Routers DeploymentConfig, however there is not any options to specify custom ports in `IngressController.operator.openshift.io/v1` object right now.
+
+Having multiple set of ingressControllers is useful for router sharding, routers with different policies (public, private), routers with different configuration options, etc. Running in HostNetwork is a strict requirement in some scenarios (e.g. environemtns with custom PBR rules). Also there might be some performance considerations to run routers with HostNetwork strategy.
```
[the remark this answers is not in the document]
m-yosefpor: Policy Based Routing rules, e.g. when we have source-ip based routing rules such as
```
ip rule add from 192.168.1.0/24 table mytable prio 1
```

In our case, we have multiple default GWs, and each packet should move back from the interface which has arrived to the node. In such cases, the pod routing rules and tables should match the ones on the host, otherwise the traffic is not routed correctly. Although If there are only static route records in main routing table, network namespaces works fine, but with PBR rules, the pods need to run on the hostNetwork, unless someone wants to do hacky workaround like adding the rules with initContainers (which is also not possible in-case of ingress pods as the deployment is created by CIO not cluster-admins)

For performance considerations, I meant some users want to run the ingress pods in hostNetwork since it has a better performance (probably there should be better wording for that instead of `performance considerations`)

better perfomance of hostNetwork routers has been shown in the OKD docs benchmarks (https://docs.okd.io/latest/scalability_and_performance/routing-optimization.html).


Shall I mention the above details in proposal too?

enhancements/ingress/bind-options.md
m-yosefpor: - env vars mentioned
- link to openshift 3.11 documentation added
- typos fixed

enhancements/ingress/bind-options.md
m-yosefpor: applied

enhancements/ingress/bind-options.md
m-yosefpor: applied

enhancements/ingress/bind-options.md
m-yosefpor: applied

enhancements/ingress/bind-options.md
m-yosefpor: applied

enhancements/ingress/bind-options.md
m-yosefpor: applied

enhancements/ingress/bind-options.md
m-yosefpor: the last sentence removed

enhancements/ingress/bind-options.md
m-yosefpor: Moved them under the unit tests and added the mentioned scenario for e2e test.

m-yosefpor:
@alebedev87 Thanks for your detailed and comprehensive review.

I've applied all your suggestions and also added your ID to reviewers list. Also Fixed the pipeline error.

Only the PBR rules and hostNetwork performance discussion is unresolved yet.

enhancements/ingress/bind-options.md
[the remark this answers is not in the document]
alebedev87: No, I think it's fine like this.

About the performance it wasn't clear to me whether you meant performance benefits or penalties. Yes, may be a different wording or even the link to OKD doc would make the phrase absolutely clear.

enhancements/ingress/bind-options.md
```diff
@@ -0,0 +1,295 @@
...
+    spec:
+    endpointPublishingStrategy:
+        type: HostNetwork
+        hostNetwork:
+        bindOptions:
+            httpPort: 11080
+            httpsPort: 11443
+            sniPort: 11444
+            noSniPort: 11445
+            statsPort: 1937
```
alebedev87, line 184: My only uncertainty about this API is whether we should make it more generic or not. I don't quite appreciate the repetition of `Port` in all the fields. I was thinking about 2 alternatives:
1. generic
```yaml
apiVersion: operator.openshift.io/v1
kind: IngressController
metadata:
name: default-2
namespace: openshift-ingress-operator
spec:
endpointPublishingStrategy:
    type: HostNetwork
    hostNetwork:
      bindOptions:
          ports:
            http: 11080
            https: 11443
            sni: 11444
            nosni: 11445
            stats: 1937
```
2. even more generic
```yaml
apiVersion: operator.openshift.io/v1
kind: IngressController
metadata:
name: default-2
namespace: openshift-ingress-operator
spec:
endpointPublishingStrategy:
    type: HostNetwork
    hostNetwork:
      bindOptions:
          ports:
            - name: "http"
              port: 11080
            - name: "https"
              port: 11443
            - name: "sni"
              port: 11444
            - name: "nosni"
              port: 11445
            - name: "stats"
              port: 1937
```

What do you think @m-yosefpor ? You can push back if you disagree, it's perfectly fine. As a matter of fact I don't have a strong opinion on this, just wanted to discuss the alternatives.

cc @Miciah, @frobware I would like to know your opinion on the whole API and on this comment in particular.

m-yosefpor: @alebedev87 I completely agree 👌 the "port" word repetition in the fields is undesirable. Option 1 definitely seems better than what currently is. For option 2,  it depends on the controller : for HAProxy, I think it is overkill as there are not any other ports other than those currently defined in router config, and I think it's very unlikely that they change in the near future. However, if the underlying controller changes to sth else, the more generic method in option 2 would be better. It also mitigates the part mentioned in the "risks" section about this.

Also a number 3 option would be sth like this:

```
apiVersion: operator.openshift.io/v1
kind: IngressController
metadata:
name: default-2
namespace: openshift-ingress-operator
spec:
endpointPublishingStrategy:
    type: HostNetwork
    hostNetwork:
      portBindings:
        http: 11080
        https: 11443
        sni: 11444
        nosni: 11445
        stats: 1937
```
 
Is it likely to put anything else under `bindOptions` other than port numbers anywhere in the future? If you think there would not be any other parameter related to bindings other than port numbers in HostNetwork strategy, another option would be renaming the `bindOptions` instead. 

Or an option 4, the corresponding generic definition like option 2, with `bindOptions` renamed:

```
endpointPublishingStrategy:
    type: HostNetwork
    hostNetwork:
      portBindings:
        - name: "http"
          port: 11080
 ```

P.S. on controller implementation: I don't think the openshift community is going to change the ingressController implementation from HAProxy to contour (or anything else) ever, although they might use contour for supporting Gateway-api (formerly Service-api) and probably introducing new CRDs/COs for lifecycle of the controller. Am I right?

Comment (alebedev87) [reply enhancements/ingress/bind-options.md:184]: Option 3 looks good to me too. I don't think we plan to do anything other than ports for `bindOptions`. However keeping the name more open for further enhancements sounds tempting. When I was looking at the API  I had `bind` Linux syscall in mind, from that point of view `bindOptions` looks quite self-contained and leave enough room for other fields (i.e. address).
About other implementations of the ingress controller, API was designed to be as generic as possible, even Contour can take the place of HAProxy, no concrete plans for the moment though.

Well, at the end I have to choose something, so I vote for the option 1. Also I would like to know what you guys (@Miciah, @frobware) think about the options.

Comment (frobware) [reply enhancements/ingress/bind-options.md:184]: At the moment I would lean more to option 2. A list without hard-coded names which only make sense (ATM) for haproxy.

Comment (alebedev87) [reply enhancements/ingress/bind-options.md:184]: @m-yosefpor, please find the outcome of the discussion @frobware @Miciah and I had on this API enhancement:

We have some concerns about putting the sni/nosni port configuration in the API:

- The sysadmin’s goal is really only to configure alternative ports for 80/443/1936 and to be able to point an external LB to these ports.
- Requiring the sysadmin to configure these additional ports would mean we would need additional documentation to explain why the sysadmin needed to care about something that is really an implementation detail of the HAProxy configuration.  (And this internal implementation detail could conceivably change, in which case the API would become cruft.)
- The larger API surface comes with an increased risk of misconfiguration and port conflicts that an automated solution could prevent.
- Requiring 5 ports instead of just 3 reduces scalability for users who want to run a large number of host-network router pods on the same node.
- Requiring HAProxy-specific configuration makes migration to another proxy (such as Contour) less straightforward.  

As an alternative to requiring the sysadmin to configure the sni/nosni ports, how about the following?

- The sysadmin configures alternatives for ports 80/443/1936.
- The cluster ingress operator configures the router to use these alternative ports, and configures HAProxy to use a unique loopback address for the sni/nosni frontends.  
     - This loopback address would be determined deterministically, for example by hashing the ingresscontroller’s name.  For example, the “default” ingresscontroller could use 127.1.2.3, and the “custom” ingresscontroller could use 127.4.5.6.
     - The sni/nosni port numbers wouldn’t need to change to avoid the conflict as the routers would be using different IP addresses.  

On further thought, the need to configure the binding address:port for the sni/nosni frontends raises the question of whether we need this internal hop over a TCP connection.  We’ll look into why we have that hop, and whether we can get rid of the loopback hop using a Unix domain socket or other solution, which would have other advantages as well.

Comment (m-yosefpor) [reply enhancements/ingress/bind-options.md:184]: @alebedev87 Thank you very much for your detailed response. Great conclusions! Actually I thought about CIO to decide the ports for sni/nosni, but then I realized it would be very situation dependent and raises much more issues. The idea of listening on different lo addresses (and better, a unix domain) is awesome.

So I'll modify the proposal to include only the 3 mentioned ports. And for the format, @frobware was leaning towards option2, but now that we are going to remove sni/nosni, should we go with option 1 instead?

File: enhancements/ingress/bind-options.md (hunk 9 above)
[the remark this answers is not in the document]
Comment (m-yosefpor) [reply enhancements/ingress/bind-options.md]: fixed.

File: enhancements/ingress/bind-options.md (hunk 10 above)
Comment (alebedev87) [reply enhancements/ingress/bind-options.md:184]: Sorry @m-yosefpor, I was off for a week.
Yes, I think the option 1 with only 3 ports (http,https,stats) would be good.

Comment (m-yosefpor) [reply enhancements/ingress/bind-options.md:184]: @alebedev87 sorry I was also off for a week (dealing with COVID). I'm going to start working on this again. Another challenge that we faced `X-Forward-Port` is set to the port the router is listening on. So if you change the port to `10443`, then `X-Forwarded-Port: 10443`. Most users have an external LB on standard port 80 and 443, and want to loadbalance to other ports (e.g. 10080, 10443). However with this header set, `image-registry` will stop working as it redirects users to `X-Forwarded-Port`, and the user tries to connect to this port of external LB which fails. There might be some other apps which will face this issue too. Should I also mention this in the enhancement? any changes needed in image-registry? or introducing a new option in routers to override X-Forwarded-Port to any custom value? (currently it is possible to set policy for forward headers in routes, but  it does not help with this issue, as the only option is to set `haproxy.router.openshift.io/set-forwarded-headers: never` which then the X-Forwarded-For will be lost as well.

Comment (alebedev87) [reply enhancements/ingress/bind-options.md:184]: @m-yosefpor I'm not sure I understood the problem you faced and whether it's worth being mentioned in the enhancement. Note that you have 4 different policies of how to handle `X-Forwarded-*` headers on OpenShift ingress controller's side: [Using X-Forwarded headers](https://docs.openshift.com/container-platform/4.8/networking/ingress-operator.html#nw-using-ingress-forwarded_configuring-ingress).

Comment (m-yosefpor) [reply enhancements/ingress/bind-options.md:184]: The problem is when somebody uses another port (other than standard 80 and 443), for example 10080 and 10443 for their ingressControllers, the most probable use-case is they have an external loadbalancer in front of routers listening on standard 80 and 443 port, which loadbalances the traffic to 10080 and 10443 ports. For example


11.1.1.1 (external loadbalancer listening on port 443) -> 10.2.1.1:10443 (OCP node IP, with 10443 port specified in `spec.endpointPublishingStrategy.hostNetwork.bindOptions.ports.https` )


A record for ingress: `*.apps.<cluster_name>.<base_domain>.` IN A 11.1.1.1

Then many apps will not work properly, including openshift-image-registry which probably causes e2e tests to fail.

And yes the policy of `X-Forwarded headers` will fix the issue, however we should override the default header policy from `append` to `never` in ingressController, when `spec.endpointPublishingStrategy.hostNetwork.bindOptions` are set to anything other than `http: 80, https: 443`.

```
apiVersion: operator.openshift.io/v1
kind: IngressController
metadata:
  name: default
  namespace: openshift-ingress-operator
spec:
  httpHeaders:
    forwardedHeaderPolicy: Never # <<<<<< this is Append by default which should be overrided by Never, even if user specifies that.
```


The error in case of `append` policy for openshift-image-registry is:

```
Error response from daemon: Get https://image-registry.apps.<cluster_name>.<base_domain>./v2/: Get https://registry.apps.<cluster_name>.<base_domain>.:10443/openshift/token?account=my-sa&client_id=docker&offline_token=true: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers) (Client.Timeout exceeded while awaiting headers)
```

The reason why registry will stop working in this case is that `X-Forwarded-Port` header is set to ``spec.endpointPublishingStrategy.hostNetwork.bindOptions.https` (10443 in this case), will causes a redirect from 443 to 10443 will cause time out (as this port is not even exposed on external LB).

Comment (m-yosefpor) [reply enhancements/ingress/bind-options.md:184]: Also it is a limitation of listening on non-standard ports, to not to be able to use `append` or `replace` policies, and causes the source IP to be missing, unless we provide other properties in `IngressController`:

e.g. having another separate policy for `X-Forwarded-Port`

```
apiVersion: operator.openshift.io/v1
kind: IngressController
metadata:
  name: default
  namespace: openshift-ingress-operator
spec:
  httpHeaders:
    forwardedHeaderPolicy: Append
    forwardedPortHeaderPolicy: Never # <<< new property
```

Comment (alebedev87) [reply enhancements/ingress/bind-options.md]: Hi @m-yosefpor!

Thank you for the investigation of this problem. Sorry for the late reply I was trying to understand why you saw this problem.

1. Image registry indeed uses the `X-Forwarded-*` headers to retrieve the host and port of the original request. This is needed to send the token retrieval link to the original client (`<original-host>:<original-port>/openshift/token?account=my-sa&client_id=docker&offline_token=true`). The code of the image registry [favors the explicit port](https://github.com/openshift/image-registry/blob/master/pkg/origin-common/util/httprequest/httprequest.go#L82) from `X-Forwarded-Port` header over the implicit from `X-Forwarded-Host` and [always gets](https://github.com/openshift/image-registry/blob/master/pkg/origin-common/util/httprequest/httprequest.go#L63) the first value of them.
2. As you know IngressController's default forwardedHeaderPolicy is `Append` which means that `X-Forwarded-*` headers are appended with HAProxy values so that the original values stay the first in the list. In the current HAProxy configuration we append the following x-forwarded headers:
```
X-Forwarded-For
X-Forwarded-Host
X-Forwarded-Port
X-Forwarded-Proto
```

I tried to exercise the behavior of HAProxy using a simple httpserver which responds with all the headers and uses `SchemeHost` implementation from the image registry on OpenShift 4.8.5. Here is the result with simple curl:
```bash
$ curl http://httpserver-default.apps.<public-domain>

Accept=*/*
Forwarded=for=XXX.XXX.XXX.XXX;host=httpserver-default.apps.<public-domain>;proto=http
User-Agent=curl/7.61.1
X-Forwarded-For=XXX.XXX.XXX.XXX
X-Forwarded-Host=httpserver-default.<public-domain>
X-Forwarded-Port=80
X-Forwarded-Proto=http

Now: http httpserver-default.<public-domain>
Before: http httpserver-default.<public-domain>:80
```
Note that `Now` and `Before` outputs are 2 versions of `SchemeHost` function which is [the one](https://github.com/openshift/image-registry/blob/master/pkg/origin-common/util/httprequest/httprequest.go#L59) used by the image registry to get the original host, port and scheme. The current (`Now`) version skips the port if it's 80 for http and 443 for https. 

Here is the output which tried to emulate a LB in between (LB is supposed to insert x-forwarded headers):
```bash
$ curl -H 'X-Forwarded-Host: originaldomain.com' -H 'X-Forwarded-Port: 123' http://httpserver-default.apps.<public-domain>

Accept=*/*
Forwarded=for=XXX.XXX.XXX.XXX;host=httpserver-default.apps.<public-domain>;proto=http
User-Agent=curl/7.61.1
X-Forwarded-For=XXX.XXX.XXX.XXX
X-Forwarded-Host=originaldomain.com,httpserver-default.apps.<public-domain>
X-Forwarded-Port=123,80
X-Forwarded-Proto=http

Now: http originaldomain.com:123
Before: http originaldomain.com:123
```
Note that the original host and port are the first in the list of the values, that's why they appear in `Now` output. This output shows that the image registry would be sending the token request to the original host and port.

I tried a similar exercise on the image registry from the same cluster:
```bash
$ curl -v -k -H 'X-Forwarded-Host: originaldomain.com' -H 'X-Forwarded-Port: 123' https://default-route-openshift-image-registry.apps.<public-domain>/v2/

> GET /v2/ HTTP/1.1
> Host: default-route-openshift-image-registry.apps.<public-domain>
> User-Agent: curl/7.61.1
> Accept: */*
> X-Forwarded-Host: originaldomain.com
> X-Forwarded-Port: 123
> 

< HTTP/1.1 401 Unauthorized
< content-type: application/json; charset=utf-8
< docker-distribution-api-version: registry/2.0
< www-authenticate: Bearer realm="https://originaldomain.com:123/openshift/token"
< x-registry-supports-signatures: 1
< date: Tue, 07 Sep 2021 09:05:43 GMT
< content-length: 87
< 
{"errors":[{"code":"UNAUTHORIZED","message":"authentication required","detail":null}]}
```
Note the realm in `www-authenticate` header, it's set to the original host and port which I think means that the values appended by HAProxy didn't prevent the image registry from sending the token request to the original client.

So from what I saw, both the ingresscontroller and the imageregistry are implemented with the correct assumptions. If the load balancer which front ends the ingress controller inserts well `X-Forwarded-Host` and `X-Forwarded-Port` headers, they are supposed to be taken into account by the image registry. I assume that the other OpenShift components follow the same logic for the handling of x-forwarded headers as the image registry.

With that said:

- Can you please verify which x-forwarded headers are inserted by the LB you used for the test? Missing `X-Forwarded-Port` header would cause the error you saw (only HAProxy's port would be there in that case).
- You mentioned an e2e test which failed supposedly due to the x-forwarded header problem. I was wondering which e2e you were referring to as the e2e tests from your PRs in `openshift/api` and `openshift/cluster-ingress-operator` were not OKed to be launched yet.

m-yosefpor: Thank you very much @alebedev87 .

> Can you please verify which x-forwarded headers are inserted by the LB you used for the test? Missing X-Forwarded-Port header could cause the error you saw (only HAProxy's port would be there in that case)

Yes I can verify this in our cluster. Actually I used an L4 loadbalancer (hardware Fortigate) for the tests, and there are no `X-Forwarded-..*` headers set by the external LoadBalancer. Setting this header with an L7 LB (or alternatively sending client requests with `X-Forwarded` headers) will resolve the issue.

However the [openshift docs](https://docs.openshift.com/container-platform/4.8/installing/installing_bare_metal/installing-bare-metal.html#installation-load-balancing-user-infra_installing-bare-metal)  mentions that  we should only use L4 LB for app loadbalancing since many other features (such as terminations, etc) might not properly work with L7 ones. Moreover, I suppose other IPI platforms will create some kind of L4 LB setup for `ingressVIP` in `install-config.yaml`, so running ingressController on other host ports will cause this issue for them as well, unless they use proxy-protocol, which is not currently enabled by default ([docs](https://docs.openshift.com/container-platform/4.8/rest_api/operator_apis/ingresscontroller-operator-openshift-io-v1.html) )

> `.spec.endpointPublishingStrategy.hostNetwork` The empty string specifies the default, which is TCP without PROXY protocol. Note that the default is subject to change.


Another point that I should mention, is that image-registry was only an example and many other external apps are using the same mechanism (e.g. KeyCloak) which causes them to misbehaved when deployed on Openshift platforms with this IngressController setup (new non-standard ports).

 
> You mentioned an e2e test which failed supposedly due to the x-forwarded header problem.

I didn't mean the e2e failed already in my PRs in `openshift/api` or `openshift/cluster-ingress-operator` (as even I have not written them yet), but just meant that they will probably fail afterwards.

alebedev87: Indeed, L4 LB won't be acting as a reverse proxy and the workloads which rely on `X-Forwarded-*` headers may be misled.
It's rather clear now that the binding on custom ports may have this downside, thanks again for spotting this. 

Let's document this in the enhancement, I'll work with the OpenShift documentation team to have this reflected in there too.

Talking about the workloads, we'd prefer to keep the current approach: append all (port/host/proto) forwarded headers by default. It may be not perfect for all the use cases but unfortunately there is no universal answer for all the scenarios. Generally we rely on the applications to be careful with the dependency on the forwarded headers as they don't always represent a reliable source of truth. Also we provide the configuration options on different levels: ingress and routes which help to tune the policy to the application needs.

alebedev87: Is this ok with you, @m-yosefpor to update the enhancement with your findings?

m-yosefpor: @alebedev87 Yes it makes sense. Thank you.

Sorry I was busy this week, I try to work on it and add the explanations in this week.
Please feel free to take over it and add/modify the proposal as you will, especially in case if you have some deadline/time pressure.

alebedev87: @m-yosefpor: I created a PR to you fork, it contains mainly the X-Forwarded risk and `bindOptions.ports` field plus some minor changes and rewordings: https://github.com/m-yosefpor/enhancements/pull/1
If you are fine with the changes you can just merge it and they'll appear in this PR.

m-yosefpor: thanks. Merged @alebedev87

alebedev87:
@m-yosefpor : can you squash the commits, please?

m-yosefpor:
@alebedev87 It's ok now.

alebedev87:
@m-yosefpor: there are still 4 commit

m-yosefpor:
> @m-yosefpor: there are still 4 commit

done

m-yosefpor:
@alebedev87 squashing the "Merge branch 'openshift:master' into master" commit results in 57 changed files. It is only fixed by merging openshift:master to my branch!

m-yosefpor:
@alebedev87 NVM. my git alias was staging all changed files after reset soft. Now it's only 1 commit.

alebedev87:
@Miciah, @frobware 
The propsal was updated with `X-Forwarded` headers risk, the PR about using of unix sockets for the router and `bindOptions.port` field.
Can you please review/approve it?

pweil-:
Red Hat sponsored projects and open source communities are subject to and must comply with the U.S. Export Regulations. These regulations prohibit Red Hat software, services, and support from being exported, re-exported or transferred, either directly or indirectly, to a prohibited destination and/or country under U.S. sanctions administered by the U.S. Department of Treasury, Office of Foreign Assets Control ("OFAC"). Because there is indication that your contribution is from an embargoed country and would violate the U.S. Export Regulations to accept, your access has been restricted. You may review OFAC’s Sanctions Programs and Information at: https://www.treasury.gov/resource-center/sanctions/pages/default.aspx and Red Hat’s export compliance policies and guidelines at: https://www.redhat.com/en/about/export-control-product-matrix.

PR #860, opened 2021-08-09
Use conversion webhook for adding repositoryMirrors to ICSP

QiWang19:
In the `repositoryMirrors` enhancement, add conversion webhook as ImageContentSourcePolicy CRD version conversion strategy
for migrating the CRD version from v1alpha1 to v1alpha2.


QiWang19:
@mrunalp @sttts PTAL. Add conversion webhook for version conversion. Is it the way we want it to work?

PR #861, opened 2021-08-10
Add CSI driver metrics ports

jsafrane:
In 4.8 we added metrics to our CSI driver controller pods.

- They run with `hostNetwork: true` to have access to cloud metadata via link-local IP address.
- Each CSI driver controller pod runs up to 6 containers (driver + various sidecars). Each of them has a separate port for metrics.

Using ports 9200 - 9220 to have room for 2 CSI drivers per cloud. On AWS, we already run AWS EBS and AWS EFS CSI drivers and each of them needs its own set of ports.

And add loopback ports 10302-3, as AWS EFS needs a local port on the node for its liveness probe.

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1990662


dhellmann:
It would be good to have someone on the storage team confirm this.

/approve

PR #863, opened 2021-08-10
[OCPNODE-677] Add conversin webhook for ICSP

QiWang19:
Use conversion webhook for adding repositoryMirrors to ICSP.
The ImageContentSourcePolicy CRD needs a conversion webhook from
operator.openshift.io/v1alpha1 to operator.openshift.io/v1alpha2 for
adding new spec `repositoryMirrors` and deprecating `repositoryDigestMirrors`.


QiWang19:
Close the enhancement since we decide to handle both versions of the ICSP inside the MCO and the webhook is not needed. This will avoid taking the complexity of the webhook.

PR #864, opened 2021-08-10
[OCPNODE-555] Add SignedRegistries to Image CR

QiWang19:
Story: https://issues.redhat.com/browse/OCPNODE-555
Epic: https://issues.redhat.com/browse/OCPNODE-538
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1866268
Add `signedRegistries`, under the `registrySources`, to Image CR,
allowing users to use Image CR for Image signature verification.

### DIFF
--- a/enhancements/api-review/add-signedRegistries.md
+++ b/enhancements/api-review/add-signedRegistries.md
@@ -1,0 +1,293 @@
+---
+title: add-signedRegistries
+authors:
+  - "@QiWang19"
+reviewers:
+  - TBD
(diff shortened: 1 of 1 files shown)


jimmyjones2:
What about disconnected clusters using local mirrors? They'll need to use [remapIdentity](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md#signedby). Could those be automatically generated from the ImageContentSourcePolicy objects?

PR #867, opened 2021-08-12
Move PAO to OCP enhancement

MarSik:
This is a preliminary plan describing how to make
the [Performance Addon Operator](https://github.com/openshift-kni/performance-addon-operators) part of OpenShift core.

### DIFF
--- a/enhancements/node-tuning/pao-in-nto.md
+++ b/enhancements/node-tuning/pao-in-nto.md
@@ -1,0 +1,291 @@
+---
+title: pao-as-part-of-core-ocp
+authors:
+  - "@MarSik"
+  - @yanirq
+reviewers:
+  - @fromanirh
+  - @jmencak
+  - @mrunalp
+  - @dhellmann
+  - @sttts
+approvers:
+  - @dhellmann
+creation-date: 2021-08-12
+last-updated: 2021-08-12
+status: provisional
+---
+
+# PAO move into NTO to become part of core OCP
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [x] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The [Performance Addon Operator](https://github.com/openshift-kni/performance-addon-operators) is a component that makes it easier to configure an OCP cluster for low latency and real-time purposes [OCP docs](https://docs.openshift.com/container-platform/4.8/scalability_and_performance/cnf-performance-addon-operator-for-low-latency-nodes.html).
+It is a high level orchestrator that consumes a [Performance Profile](https://github.com/openshift-kni/performance-addon-operators/blob/master/docs/interactions/performance-profile.yaml) and generates multiple manifests that are then processed by core OpenShift components like MCO and NTO.
+We have a simplified interaction diagram here: https://github.com/openshift-kni/performance-addon-operators/blob/master/docs/interactions/diagram.png
+A more in depth description of how a low latency tuned cluster works was presented at DevConf 2021: https://devconfcz2021.sched.com/event/gmJD/openshift-for-low-latency-and-real-time-workloads
+
+The proposal in hand is to move the existing implementation of PAO under
+[cluster node tuning operator (NTO)](https://github.com/openshift/cluster-node-tuning-operator) without adding new features.
+
+## Motivation
+
+The Telco team maintains its own build and release pipelines and releases via a [Telco specific errata](https://red.ht/3boqjeS) that however targets the OpenShift product. So from a customer perspective PAO is part of OCP.
+
+- More and more customers are using PAO to configure their clusters for low latency or are being told to do so. The Telco team has so far managed to release PAO on the same day as OCP to maintain the illusion it is part of the product, but it costs a LOT of time to maintain the downstream build process.
+
+- The OCP [gcp-rt](https://prow.ci.openshift.org/?type=periodic&job=*gcp-rt*) lane tests OCP deployment on top of GCP with the real-time kernel enabled, but it only works [thanks to some pieces of PAO tuning](https://github.com/openshift/release/pull/18870). However the tuning is a pre-rendered snippet, not linked to the operator, because the deployment does not know about PAO.
+
+- PAO is closely tied to some OCP components
+([Cluster Node Tuning Operator](https://github.com/openshift/cluster-node-tuning-operator))
+and depends on the OCP version. So we release PAO for every OCP minor version (since 4.4) and together with some of the OCP Z-streams.
+The upgrade procedure is tricky though, first OCP needs to be upgraded and then PAO via an OLM channel change.
+This is needed to make sure the necessary functionality is present in the running NTO.
+Some tuning changes and bug fixes in NTO/tuned as well as in RHCOS kernel are critical for the advertised functionality.
+
+PAO does not depend on anything version specific with worker nodes.
+NTO takes care of the few pieces that are needed with tuning and MCO installs the proper RT kernel.
+PAO is tightly coupled to NTO/tuned version which is more tight than just OCP minor versions.
+Some tuning changes and bug fixes in NTO/tuned as well as in RHCOS kernel are critical for the advertised functionality.
+
+- The new [Workload partitioning](https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md)
+feature seeks to use PAO as the configuration tool.
+However Workload partitioning needs to be enabled at install phase and PAO CRDs are not yet available at that time.
+[A "hacky" procedure](https://github.com/openshift/enhancements/pull/753) is needed to overcome this.
+
+- Pre-GA availability and testing is tricky in the current way of things. The 4.9 prod OLM index (for example) will not be populated until the operator is shipped.
+At a GA date, all OCP operators are available earlier. PAO has its own separate quay.io bundle/index, but the separation is complicated for testing.
+
+### Goals
+
+We (the telco team) would like to make PAO part of OpenShift. There are multiple paths we can take to achieve this, I am going to describe the one we prefer: **PAO becoming a sub-controller in NTO.**
+
+The initial plan is to keep all logic and custom resources intact, we want to start with simply moving the code over.
+
+### Non-Goals
+
+Logic changes or tighter integration with other MCO controllers are not expected to be part of this initial move. Just the delivery mechanism wrt OCP.
+
+## Proposal
+
+To be really clear here, the Telco team will handle all the necessary changes to both codebases, however we will need support in the form of PR reviews and design consultations.
+
+### Existing Implementation
+
+PAO is being developed under the OpenShift KNI organization (https://github.com/openshift-kni/performance-addon-operators), built by custom builder upstream and cPaaS downstream. PAO is installed via OLM as a second-day operator.
+
+The current deliverables are:
+
+- [performance-addon-operator](https://catalog.redhat.com/software/containers/openshift4/performance-addon-rhel8-operator/5e1f480fbed8bd66f81cbe23?container-tabs=overview&gti-tabs=registry-tokens)
+
+  The operator contains the high level orchestration logic + a [tool to simplify generation of Performance Profiles](https://docs.openshift.com/container-platform/4.8/scalability_and_performance/cnf-create-performance-profiles.html). This tool can be executed locally on admin's laptop using `podman run --entrypoint performance-profile-creator performance-addon-operator:ver`
+
+- [performance-addon-must-gather](https://catalog.redhat.com/software/containers/openshift4/performance-addon-operator-must-gather-rhel8/5ee0bfa4d70cc56cea83827f?container-tabs=overview&gti-tabs=registry-tokens)
+
+  The must gather image [enhances](https://github.com/openshift-kni/performance-addon-operators/tree/master/must-gather) the main OCP one and [collects](https://docs.openshift.com/container-platform/4.8/scalability_and_performance/cnf-performance-addon-operator-for-low-latency-nodes.html#cnf-about-gathering-data_cnf-master) some latency tuning configuration and data that are needed for debugging.
+
+- functional tests are also vendored into
+[cnf-tests](https://catalog.redhat.com/software/containers/openshift4/cnf-tests-rhel8/5e9d81f65a134668769d560b?container-tabs=overview&gti-tabs=registry-tokens)
+that can be used to
+[validate customer deployments](https://docs.openshift.com/container-platform/4.8/scalability_and_performance/cnf-performance-addon-operator-for-low-latency-nodes.html#cnf-performing-end-to-end-tests-for-platform-verification_cnf-master) (this includes other aspects like SCTP, PTP, SR-IOV as well).
+
+### Code
+
+PAO code would be moved into the NTO repository as a separate controller. The Telco team would keep ownership as it has the necessary domain knowledge needed to implement the low latency tuning.
+
+The code currently uses Operator SDK libraries in some of the tests and this might need to be changed.
+Migrating the use of these libraries to [library-go](https://github.com/openshift/library-go) or stripping them to [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime)
+would be options in hand.
+
+PAO also implements a [conversion webhook](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion)
+The hook is currently installed via OLM mechanisms and this might need to be reimplemented or changed to use whatever CVO or ART support.
+PAO must-gather will have the native support with additional arguments and will not require an additional image as currently exists in PAO original repository.
+
+[cnf-tests](https://github.com/openshift-kni/cnf-features-deploy/) are NOT part of this move either, the Telco team will keep maintaining the validation suite in OpenShift KNI.
+
+
+### CI
+
+PAO contains both unit and functional tests implemented using Ginkgo.
+
+Integration with NTO is currently TBD, please advise if anything special is needed.
+
+One of the test suites (4_latency) implements an opinionated latency testing suite to get comparable results from customers and to prevent configuration / command line invocation mistakes. Keeping it in PAO/NTO or moving it into cnf-tests is TBD.
+
+### Build pipeline
+
+PAO would inherit the NTO pipeline and all the images will be released via ART.
+
+### Releases
+
+PAO should be shipped as part of core OCP after NTO will contain the migrated code of PAO.
+PAO already obeys the OCP github PR rules and uses the same PR bot so nothing changes in the lifecycle.
+
+### API stability
+
+The `PerformanceProfile` CR uses an `performance.openshift.io/v2` API group which marks it as stable interface that needs to be maintained for about a year after official deprecation. We do not expect any change to the API as part of this proposal.
+
+### Bugs
+
+PAO bugs are currently tracked in The OpenShift Bugzilla component [Performance Addon Operator](https://bugzilla.redhat.com/buglist.cgi?component=Performance%20Addon%20Operator&list_id=12053428&product=OpenShift%20Container%20Platform)
+
+This should be kept as-is, the Telco team will own all the bugs. We might need to move the component under NTO or rename it.
+
+The bugzilla component is currently not part of OCP metrics and we can include it. TBD
+
+
+### User Stories
+
+Both the NTO and PAO functionality should be kept intact and no regressions are allowed. The user should not notice anything except the fact that OLM is no longer needed to enable the tuning functionality.
+
+### API Extensions
+The existing PAO API Extensions will be moved under the NTO repository.
+
+### Implementation Details/Notes/Constraints
+
+- Existing api compatibility practices:
+PAO API has been treated as Tier 1 for multiple releases already (since 4.4).
+
+- Enumerate list of performance profile owned objects:
+MachineConfig, KubeletConfiguration, Tuned (see [diagram](https://github.com/openshift-kni/performance-addon-operators/blob/master/docs/interactions/diagram.png)).
+All activity happening on the cluster or nodes is indirect via other components (MCO, NTO) reacting to changes in objects owned by PAO.
+
+- Enumerate list of webhooks:
+Validation and conversion webhook (TBD)
+
+- Updates to openshift conformance test suite if included by default:
+All basic checks are doable on all linux platforms except when real-time kernel is needed. That case used to be only supported on GCP (and bare metal).
+Measuring real latency requires bare metal, but is not needed for conformance.
+
+
+### Risks and Mitigations
+
+- Since the performance profile API will be visible for users on all clusters there is a risk that a cluster admin might tune nodes improperly.
+The risk is minor since addressing the performance profile means the admin should find this specific API and deliberately perform tunings on the cluster using it.
+The same risk would be relevant for applying changes for machine and kubelet configurations that already exist in OCP clusters.
+This will mitigated by clearly documenting that admins should proceed with caution.
+
+## Design Details
+
+### Open Questions
+
+- CI: Integration with NTO is currently TBD.
+
+- CI: One of the test suites (4_latency) implements an opinionated latency testing suite to get comparable results from customers and to prevent configuration / command line invocation mistakes. Keeping it in PAO/NTO or moving it into cnf-tests is TBD.
+
+- Maintainability: There are two options:
+  - Maintaining a separate copy of the code in the NTO repository.
+  - Vendoring the current PAO code as a "library" until we can stop supporting the standalone mode (4.6+).
+  From OCP perspective it makes little difference as it would just be part of NTO release process indeed.
+
+- API definitions: They would need to go into openshift/api but at this moment NTO include its own API definitions under its own repository.
+The NTO API definitions transition into openshift/api is TBD and should be considered to be done at the time of PAO/NTO merge.
+
+- List of pertinent topology options (control plane and infrastructure) where this operator is pertinent:
+The assumption is that the addition of PAO to NTO will not add additional requirements for installation at all when control plane topology is "external"
+and the infrastructure topology is not pertinent to the function.
+The resources leveraged by PAO, now introduced into NTO, are all managed externally when in this mode (e.g. node configuration, kubelet configuration).
+PAO control plane component modifies the worker nodes (unless in single node mode where it affects masters).
+NTO currently supports only one cloud installation with external control plane topology (IBM Cloud) with a dedicated manifest.
+
+### Upgrade
+
+PAO has been installed via OLM since 4.4. Once it moves to core we will have to figure out the upgrade procedure, because OLM will not remove the optional operator by itself.
+
+Implementation options:
+- Remove the relevant OLM artifacts (CSV,Subscription,Operator group) via the performance-addon controller and leave the PAO CRDs without the change.
+- In case leader election supports such conditioning, is to keep using the same leadership election method and make sure the newest version always wins. The OLM operator will then stay on the cluster (can be release noted) as an inactive component.
+
+### Test Plan
+
+All tests already exist in OCP CI, the trigger will move to the NTO repository.
+
+Continuous testing is done via [OpenShift CI jobs](https://github.com/openshift/release/tree/daaab71832999c072c60667955637fa3a535d4ba/ci-operator/config/openshift-kni/performance-addon-operators) as well as some downstream specific Jenkins instances (both virtual and bare metal).
+
+### Graduation Criteria
+
+The user facing feature is already considered stable and GA, just shipped as an optional operator.
+
+#### Dev Preview -> Tech Preview
+
+#### Tech Preview -> GA
+
+#### Removing a deprecated feature
+
+### Upgrade / Downgrade Strategy
+
+It is expected that the image will be kept up-to-date as part of regular OCP flows.
+
+API changes are handled by a conversion webhook that PAO already contains to make upgrades transparent to users.
+
+PAO depends on NTO and MCO functionality.
+
+Downgrades should be handled the same way the cluster node tuning operator was handled for minor versions up until now.
+
+If applicable, how will the component be upgraded and downgraded? Make sure this
+is in the test plan.
+
+PAO is an infrastructure orchestrator so the latest available version has to win. It generates and maintains objects for other controllers (MCO, NTO) that then apply them to nodes.
+
+### Version Skew Strategy
+
+NTO behavior will be kept so no addition strategy aspects on NTO counter parts.
+When PAO code base will be moved and used inside NTO it will be version aligned with CRI-O and K8s at the time of release.
+CI should cover the refactored component behavior with its counter parts.
+
+### Operational Aspects of API Extensions
+
+#### Failure Modes
+
+#### Support Procedures
+
+## Implementation History
+
+Major milestones in the life cycle of a proposal should be tracked in `Implementation
+History`.
+
+## Drawbacks
+
+The idea is to find the best form of an argument why this enhancement should _not_ be implemented.
+
+The proposed change adds more code to the NTO component increasing the probability of bugs and possibly a higher resource usage.
+
+However using Performance Profile is optional and so users that do not use this way of tuning should be not affected at all.
+
+## Alternatives
+
+### Alternative upgrade Implementation
+Have the PAO copy in NTO disable itself when the standalone version is installed, updating the ClusterOperator resource with that reason.
+Then the user can remove the old standalone version from their cluster.
+This would be easier to implement reliably, since we would only need to look for the subscription or other indicator that there is a standalone copy installed.
+We also would not need to deliver any changes in the standalone package, so it would not matter which version is installed in the cluster at the start of the upgrade
+
+### PAO as a full OCP operator
+
+The alternative approach is to move PAO into OCP as a fully standalone operator. This would have no impact on the MCO team and would be simpler in terms of code.
+
+However there are disadvantages to this approach:
+
+- OCP is already heavy on infrastructure resource consumption and an additional operator would make that worse
+- implementing a conditional start is possible though, similar to what machine-operator uses, but negates the simpler on coding advantage
+
+### PAO as a sub-controller of Machine Config Operator
+
+PAO could also live inside MCO if NTO cannot accept it. The long term future might be to merge all of them together anyway. All other aspects of this design are the same for this approach.
+Since a new MCO design is under consideration, it make more sense to go with NTO for now.
+## Future
+
+PAO currently uses Ignite heavily and since MCO is [looking at dropping the support](https://hackmd.io/UzgrWuCLQ-u9bMh_JhRKOA#Proposal), it would make it easier to adapt to this change if PAO could directly contribute to the layered Machine Image instead of describing the necessary config changes as an MachineConfig first.
+
+## Infrastructure Needed [optional]
+
+This is just a move of an existing project and all the infrastructure is already in use.


enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,258 @@
...
+
+PAO is being developed under the OpenShift KNI organization (https://github.com/openshift-kni/performance-addon-operators), built by custom builder upstream and cPaaS downstream. PAO is installed via OLM as a second-day operator.
+
+The current deliverables are:
+
+- [performance-addon-operator](https://catalog.redhat.com/software/containers/openshift4/performance-addon-rhel8-operator/5e1f480fbed8bd66f81cbe23?container-tabs=overview&gti-tabs=registry-tokens)
+
+  The operator contains the high level orchestration logic + a [tool to simplify generation of Performance Profiles](https://docs.openshift.com/container-platform/4.8/scalability_and_performance/cnf-create-performance-profiles.html). This tool can be executed locally on admin's laptop using `podman run --entrypoint performance-profile-creator performance-addon-operator:ver`
+
+- [performance-addon-must-gather](https://catalog.redhat.com/software/containers/openshift4/performance-addon-operator-must-gather-rhel8/5ee0bfa4d70cc56cea83827f?container-tabs=overview&gti-tabs=registry-tokens)
```
cynepco3hahue, line 47: probably we should just add the native support for the PAO under the must-gather

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,258 @@
...
+
+  The must gather image [enhances](https://github.com/openshift-kni/performance-addon-operators/tree/master/must-gather) the main OCP one and [collects](https://docs.openshift.com/container-platform/4.8/scalability_and_performance/cnf-performance-addon-operator-for-low-latency-nodes.html#cnf-about-gathering-data_cnf-master) some latency tuning configuration and data that are needed for debugging.
+
+- and the functional tests are also vendored into [cnf-tests](https://catalog.redhat.com/software/containers/openshift4/cnf-tests-rhel8/5e9d81f65a134668769d560b?container-tabs=overview&gti-tabs=registry-tokens) that can be used to [validate customer deployments](https://docs.openshift.com/container-platform/4.8/scalability_and_performance/cnf-performance-addon-operator-for-low-latency-nodes.html#cnf-performing-end-to-end-tests-for-platform-verification_cnf-master) (this includes other aspects like SCTP, PTP, SR-IOV as well).
+
```
cynepco3hahue, line 52: we are missing the performance-profile-creator section

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,258 @@
...
+## Design Details
+
+### Open Questions [optional]
+
+### Upgrade
+
+- PAO has been installed via OLM since 4.4. Once it moves to core we will have to figure out the upgrade procedure, because OLM will not remove the optional operator by itself.
+
+  One of the options is to keep using the same leadership election method and make sure the newest version always wins. The OLM operator will then stay on the cluster (can be release noted) as an inactive component.
```
cynepco3hahue, line 147: I do not think the leader election supports such conditioning. One of the options is to remove the relevant OLM stuff via the performance-addon controller and leave the CRDs without the change.

enhancements/node-tuning/pao-in-nto.md
yanirq: its mentioned in line 45:

> The operator contains the high level orchestration logic + a tool to simplify generation of Performance Profiles

Are you missing some more info about it ?

enhancements/node-tuning/pao-in-nto.md
yanirq: its mentioned in line 45:

> The operator contains the high level orchestration logic + a tool to simplify generation of Performance Profiles

Are you missing some more info about it ?

enhancements/node-tuning/pao-in-nto.md
yanirq: We can add that under the design goals then.  in this section we are going over the existing deliverables.

enhancements/node-tuning/pao-in-nto.md
yanirq: What are the relevant OLM stuff here ?  if we remove the catalog source/csvs/subscriptions  it will probably remove the pao-olm-operator and consequently its owned CRs no ?    
This brings me to another thought, what will happen to the previous version Maching config/tuned generated artifacts.

cynepco3hahue: CSV, subscription, operator group. It will not remove the CRD, so all CR's should be untouched.

enhancements/node-tuning/pao-in-nto.md
cynepco3hahue: nope, if it is mentioned it is good enough for me.

enhancements/node-tuning/pao-in-nto.md
yanirq: So are we aiming for a single version finalizer here ?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,258 @@
...
+The code currently uses Operator SDK libraries in some of the tests and this might need to be changed.
+
+PAO also implements a migration webhook to migrate its custom resources during upgrade. The hook is currently installed via OLM mechanisms and this might need to be reimplemented or changed to use whatever CVO or ART support.
+
+[cnf-tests](https://github.com/openshift-kni/cnf-features-deploy/) are NOT part of this move either, the Telco team will keep maintaining the validation suite in OpenShift KNI.
+
+
+#### CI
+
+PAO contains both unit and functional tests implemented using Gingko.
```
jmencak, line 92: `s/Gingko/Ginkgo/`

jmencak (commented):
Looks reasonable, spotted a typo.

enhancements/node-tuning/pao-in-nto.md
yanirq: done

enhancements/node-tuning/pao-in-nto.md
yanirq: Added the OLM artifacts removal as an option to this doc.

enhancements/node-tuning/pao-in-nto.md
yanirq: done

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Proposal
+
+To be really clear here, the Telco team will handle all the necessary changes to both codebases, however we will need support in the form of PR reviews and design consultations.
+
+### Code
+
+PAO code would be moved into the NTO repository as a separate controller. The Telco team would keep ownership as it has the necessary domain knowledge needed to implement the low latency tuning.
+
+The code currently uses Operator SDK libraries in some of the tests and this might need to be changed.
```
jmencak, line 93: Many of the core operators use either `client-go` directly or moved to https://github.com/openshift/library-go .  You may want to investigate and look into that.  My long-term plan for NTO is to evaluate a transition from directly using `client-go` to using https://github.com/openshift/library-go instead.

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+
+Logic changes or tighter integration with other MCO controllers are not expected to be part of this initial move. Just the delivery mechanism wrt OCP.
+
+## Proposal
+
+To be really clear here, the Telco team will handle all the necessary changes to both codebases, however we will need support in the form of PR reviews and design consultations.
+
+### Code
+
+PAO code would be moved into the NTO repository as a separate controller. The Telco team would keep ownership as it has the necessary domain knowledge needed to implement the low latency tuning.
```
dhellmann, line 91: I think the API definitions would need to go into openshift/api. @sttts or @deads2k can you confirm that?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+This is needed to make sure the necessary functionality is present in the running NTO.
+
+- The new [Workload partitioning](https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md)
+feature seeks to use PAO as the configuration tool.
+However Workload partitioning needs to be enabled at install phase and PAO CRDs are not yet available at that time.
+[A "hacky" procedure](https://github.com/openshift/enhancements/pull/753) is needed to overcome this.
+
+### Goals
+
+We (the telco team) would like to make PAO part of OpenShift. There are multiple paths we can take to achieve this, I am going to describe the one we prefer: **PAO becoming a sub-controller in NTO.**
```
dhellmann, line 77: I know we focus a lot on bare metal for PAO work. Do the APIs implemented by the PAO make sense on all platforms and deployment topologies? Are there any conditions under which we would not want those APIs available?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+Major milestones in the life cycle of a proposal should be tracked in `Implementation
+History`.
+
+## Drawbacks
+
+The idea is to find the best form of an argument why this enhancement should _not_ be implemented.
+
+The proposed change adds more code to the NTO component increasing the probability of bugs and possibly a higher resource usage.
+
+However using Performance Profile is optional and so users that do not use this way of tuning should be not affected at all.
```
dhellmann, line 257: This change means introducing the PerformanceProfile into all clusters, meaning the API is visible to all users. Is there any danger that a cluster admin, experimenting with the APIs, could mis-configure their cluster and break something?

enhancements/node-tuning/pao-in-nto.md
mrunalp: Does it make sense to add a configuration knob to disable it?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+### PAO as a full OCP operator
+
+The alternative approach is to move PAO into OCP as a fully standalone operator. This would have no impact on the MCO team and would be simpler in terms of code.
+
+However there are disadvantages to this approach:
+
+- OCP is already heavy on infrastructure resource consumption and an additional operator would make that worse
+- implementing a conditional start is possible though, similar to what machine-operator uses, but negates the simpler on coding advantage
+
+### PAO as a sub-controller of Machine Config Operator
```
mrunalp, line 270: We can say that currently a new MCO design is under consideration, so it make more sense to go with NTO for now.

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+
+Consider including folks that also work outside your immediate sub-project.
+
+## Design Details
+
+### Open Questions [optional]
+
+### Upgrade
+
+- PAO has been installed via OLM since 4.4. Once it moves to core we will have to figure out the upgrade procedure, because OLM will not remove the optional operator by itself.
```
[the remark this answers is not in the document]
mrunalp: Agree! Also, can we capture addinge2e tests that will exercise the upgrade scenarios?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+- PAO is closely tied to some OCP components
+([Cluster Node Tuning Operator](https://github.com/openshift/cluster-node-tuning-operator))
+and depends on the OCP version. So we release PAO for every OCP minor version (since 4.4) and together with some of the OCP Z-streams.
+The upgrade procedure is tricky though, first OCP needs to be upgraded and then PAO via an OLM channel change.
+This is needed to make sure the necessary functionality is present in the running NTO.
+
+- The new [Workload partitioning](https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md)
+feature seeks to use PAO as the configuration tool.
+However Workload partitioning needs to be enabled at install phase and PAO CRDs are not yet available at that time.
+[A "hacky" procedure](https://github.com/openshift/enhancements/pull/753) is needed to overcome this.
```
dhellmann, line 73: There's a strong desire to reduce the number of components delivered as part of the OCP release payload, and have more of them installed via OLM. There are several benefits to that approach, including decoupling release schedules, shrinking the payload, and making it possible to reduce the API surface area for alternative topologies. The PAO seems like a strong candidate for OLM delivery, especially since that's how it already works.

If we do add it to the payload, NTO seems like a reasonable "home" (although I could also see moving NTO to OLM delivery). But I'm not yet convinced it's necessary to make this change, based on the motivation described here.

Regarding integration with the installer, it would be useful to have additional requirements added to https://issues.redhat.com/browse/OLM-2232, if the plans there don't already cover the use case for PAO.

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+
+- More and more customers are using PAO to configure their clusters for low latency or are being told to do so. The Telco team has so far managed to release PAO on the same day as OCP to maintain the illusion it is part of the product, but it costs a LOT of time to maintain the downstream build process.
+
+- The OCP [gcp-rt](https://prow.ci.openshift.org/?type=periodic&job=*gcp-rt*) lane tests OCP deployment on top of GCP with the real-time kernel enabled, but it only works [thanks to some pieces of PAO tuning](https://github.com/openshift/release/pull/18870). However the tuning is a pre-rendered snippet, not linked to the operator, because the deployment does not know about PAO.
+
+- PAO is closely tied to some OCP components
+([Cluster Node Tuning Operator](https://github.com/openshift/cluster-node-tuning-operator))
+and depends on the OCP version. So we release PAO for every OCP minor version (since 4.4) and together with some of the OCP Z-streams.
+The upgrade procedure is tricky though, first OCP needs to be upgraded and then PAO via an OLM channel change.
+This is needed to make sure the necessary functionality is present in the running NTO.
```
dhellmann, line 68: Can you provide more detail about why is the PAO so tightly coupled to the OCP version?

Does the new work to make it possible for OLM operators to specify the versions of OCP they're compatible with help the upgrade story?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Open Questions [optional]
+
+### Upgrade
+
+- PAO has been installed via OLM since 4.4. Once it moves to core we will have to figure out the upgrade procedure, because OLM will not remove the optional operator by itself.
+
+  One option is to remove the relevant OLM artifacts (CSV,Subscription,Operator group) via the performance-addon controller and leave the PAO CRDs without the change.
+
+  Another option, in case leader election supports such conditioning, is to keep using the same leadership election method and make sure the newest version always wins. The OLM operator will then stay on the cluster (can be release noted) as an inactive component.
```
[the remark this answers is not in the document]
yanirq: @cynepco3hahue  after we check the POC lets see what direction we can move towards here.

enhancements/node-tuning/pao-in-nto.md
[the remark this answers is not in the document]
yanirq: We are still examining this, once decided we will add the conclusions here.

enhancements/node-tuning/pao-in-nto.md
yanirq: Is there an issue with PAO controller being deployed without creating its CRs ?  meaning the controller will exist but will not perform any operations (instead of using a disable knob/feature gate)

enhancements/node-tuning/pao-in-nto.md
yanirq: For experimentations it would be advised to use the performance profile creator to protect the user from mis-configurations. 
Applying directly a performance profile will not be straight forward and there are spec related validations.

enhancements/node-tuning/pao-in-nto.md
yanirq: PAO is dependent on multiple components and releases:  k8s features (cpu manager,memory manager), RHCOS version, kernel version, tuned and MCO - All of which are packed and released in an OCP version where PAO is aligned to. 
We did not have a version that could match a different release up until now (and probably not in the future)

dhellmann: PAO is going to have to be able to handle version skew between the control plane and worker nodes as part of the EUS upgrade strategy. Including the operator in the release payload will not remove that requirement.

Ping me on slack if you need more details about that.

enhancements/node-tuning/pao-in-nto.md
dhellmann: If it's going to be disabled or not in use in a large percentage of clusters, that seems like it could be another reason to deliver it via OLM.

enhancements/node-tuning/pao-in-nto.md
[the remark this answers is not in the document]
dhellmann: Please don't do a lot of development work assuming that this enhancement will be approved. At this point, I think it's better to continue delivering PAO via OLM.

derekwaynecarr:
To help evaluate, can you please provide the following:

- list of pertinent topology options (control plane and infrastructure) where this operator is pertinent.

see:
https://github.com/openshift/api/blob/a6156965faae5ce117e3cd3735981a3fc0e27e27/config/v1/types_infrastructure.go#L92
https://github.com/openshift/api/blob/a6156965faae5ce117e3cd3735981a3fc0e27e27/config/v1/types_infrastructure.go#L103

my assumption is this operator is not installed at all when control plane topology is "external" and the infrastructure topology is not pertinent to the function.  this is because the resources leveraged by the PAO are all managed externally when in this mode (e.g. node configuration, kubelet configuration).

- existing api compatibility practices

for the present state of PAO, would you advertise the API under which compatibility tier
https://docs.openshift.com/container-platform/4.8/rest_api/understanding-api-support-tiers.html
https://docs.openshift.com/container-platform/4.8/rest_api/understanding-api-support-tiers.html#mapping-support-tiers-to-openshift-api-groups_understanding-api-tiers

- enumerate list of performance profile owned objects

just for clarity, if user edits a performance profile, what related resources are modified on their behalf.

- enumerate list of webhooks

is it just a validation webhook or is there something else?

- updates to openshift conformance test suite if included by default

default options should be tested under our conformance suite, what testing would you recommend including if we proceeded on this.  is this testing only possible on particular infrastructure platforms, if so, is this an argument for only enabling this function on those platforms in the node tuning operator itself?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Code
+
+PAO code would be moved into the NTO repository as a separate controller. The Telco team would keep ownership as it has the necessary domain knowledge needed to implement the low latency tuning.
+
+The code currently uses Operator SDK libraries in some of the tests and this might need to be changed.
+Migrating the use of these libaries to [client-go](https://github.com/openshift/library-go)
+would be one of the options.
+
+PAO also implements a migration webhook to migrate its custom resources during upgrade. The hook is currently installed via OLM mechanisms and this might need to be reimplemented or changed to use whatever CVO or ART support.
```
sttts, line 97: what is a migration webhook?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+
+#### Tech Preview -> GA
+
+#### Removing a deprecated feature
+
+### Upgrade / Downgrade Strategy
+
+It is expected that the image will be kept up-to-date as part of regular OCP flows.
+
+API changes are handled by a migration webhook that PAO already contains to make upgrades transparent to users.
```
sttts, line 186: what does this mean? What does it do?

enhancements/node-tuning/pao-in-nto.md
MarSik: The correct term I should have used is conversion webhook: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion

enhancements/node-tuning/pao-in-nto.md
MarSik: Conversion webhook, it makes sure that CRs created with older apiVersion are treated properly keeping the old behavior intact or using proper defaults.

sttts: Do you have plans to deprecate v1 and eventually remove it? Or do you plan to roundtrip through v1 forever?

MarSik:
> To help evaluate, can you please provide the following:
> 
>     * list of pertinent topology options (control plane and infrastructure) where this operator is pertinent.
> 
> 
> see:
> https://github.com/openshift/api/blob/a6156965faae5ce117e3cd3735981a3fc0e27e27/config/v1/types_infrastructure.go#L92
> https://github.com/openshift/api/blob/a6156965faae5ce117e3cd3735981a3fc0e27e27/config/v1/types_infrastructure.go#L103
> 
> my assumption is this operator is not installed at all when control plane topology is "external" and the infrastructure topology is not pertinent to the function. this is because the resources leveraged by the PAO are all managed externally when in this mode (e.g. node configuration, kubelet configuration).

I cannot really answer this, but the assumption you made seems to be reasonable. PAO is a control plane component that modifies the worker nodes (unless in single node mode where it must affect masters of course).

> 
>     * existing api compatibility practices
> 
> 
> for the present state of PAO, would you advertise the API under which compatibility tier
> https://docs.openshift.com/container-platform/4.8/rest_api/understanding-api-support-tiers.html
> https://docs.openshift.com/container-platform/4.8/rest_api/understanding-api-support-tiers.html#mapping-support-tiers-to-openshift-api-groups_understanding-api-tiers

The API has been treated as Tier 1 for multiple releases already (since 4.4).

> 
>     * enumerate list of performance profile owned objects

MachineConfig, KubeletConfiguration, Tuned (see the diagram linked from the description)

https://github.com/openshift-kni/performance-addon-operators/blob/master/docs/interactions/diagram.png

> 
> 
> just for clarity, if user edits a performance profile, what related resources are modified on their behalf.

Potentially all the owned objects mentioned above. Nothing else, all activity happening on the cluster or nodes is indirect via other components (MCO, NTO) reacting to changes in objects owned by PAO.

> 
>     * enumerate list of webhooks
> 
> 
> is it just a validation webhook or is there something else?

Validation and conversion webhook to transparently upgrade PerformanceProfile on apiVersion changes.

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#overview

> 
>     * updates to openshift conformance test suite if included by default
> 
> 
> default options should be tested under our conformance suite, what testing would you recommend including if we proceeded on this. is this testing only possible on particular infrastructure platforms, if so, is this an argument for only enabling this function on those platforms in the node tuning operator itself?

All basic checks are doable on all linux platforms except when real-time kernel is needed. That case used to be only supported on GCP (and bare metal).

Measuring real latency requires bare metal, but is not needed for conformance.

enhancements/node-tuning/pao-in-nto.md
MarSik: IIRC PAO does not depend on anything version specific wrt worker nodes. NTO takes care of the few pieces that are needed wrt tuning and MCO installs the proper RT kernel.

PAO is tightly coupled to NTO/tuned version though and that is more tight than just OCP minor versions. Some tuning changes and bug fixes in NTO/tuned as well as in RHCOS kernel are critical for the advertised functionality.

enhancements/node-tuning/pao-in-nto.md
MarSik: There is a strong coupling to NTO features and sometimes to a specific kernel bug fixes delivered in RHCOS. 

One of the biggest gaps in OLM for such infrastructure operators (apart from the 1st day install described by OLM-2232) is no support for atomic upgrades and proper dependency mechanism that would support multiple Z-streams.

The merge with NTO was proposed to not increase the amount of installed and running processes indeed.

Moving NTO itself to OLM sounds tricky to me without a way to setup at least the openshift-node tuned profile in RHCOS, this is something I consider important for the platform as a whole (@jmencak do you have the original design for NTO?).

enhancements/node-tuning/pao-in-nto.md
MarSik: @dhellmann Indeed, an admin that is not careful could tune the node improperly, however he has more than enough ways to do that with MachineConfigs themselves.

Incorrect kubelet configuration can prevent the node from starting (wrong cpu specification).

enhancements/node-tuning/pao-in-nto.md
MarSik: I second this question. Where would the CRD be defined? NTO's own Tuned is defined in https://github.com/openshift/cluster-node-tuning-operator/blob/master/manifests/20-crd-tuned.yaml

enhancements/node-tuning/pao-in-nto.md
MarSik: Hugepages can be configured nicely via PAO as well as adding of kernel arguments that are currently both done via low level MCO (and documented like that in the OCP docs).

https://docs.openshift.com/container-platform/4.7/scalability_and_performance/what-huge-pages-do-and-how-they-are-consumed-by-apps.html#configuring-huge-pages_huge-pages

If nothing else, PAO would improve the user experience and bring NUMA awareness.

enhancements/node-tuning/pao-in-nto.md
MarSik: Probably yes, we introduced v2 in 4.7 and there were no upgrade incompatible changes to it since. Deprecating v1 is something we can consider indeed, we just haven't had the time to do so yet.

enhancements/node-tuning/pao-in-nto.md
jmencak: @MarSik, the original design document/proposal is a [google doc](https://github.com/openshift/enhancements/pull/867#issuecomment-897607025). I believe the audience for NTO is [a lot wider](https://github.com/redhat-performance/tuned/tree/master/profiles) than for PAO, but PAO definitely helps where low-latency tuning is needed. The main aim behind NTO was to concentrate node tuning into one place and take advantage of years of experience transferred into [workload-specific profiles](https://github.com/redhat-performance/tuned/tree/master/profiles).  `TuneD` has been an integral part of `RHEL` for many years and customers have come to expect it there.

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,280 @@
...
+
+One of the test suites (4_latency) implements an opinionated latency testing suite to get comparable results from customers and to prevent configuration / command line invocation mistakes. Keeping it in PAO/NTO or moving it into cnf-tests is TBD.
+
+### Build pipeline
+
+PAO would inherit the NTO pipeline and all the images will be released via ART.
+
+### Releases
+
+PAO should be shipped as part of core OCP. ART again. PAO already obeys the OCP github PR rules and uses the same PR bot so nothing changes in the lifecycle.
```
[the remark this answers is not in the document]
yanirq: adding the 2 options of maintainability to open questions section.

enhancements/node-tuning/pao-in-nto.md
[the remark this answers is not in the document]
yanirq: the option above will be added to the doc.

enhancements/node-tuning/pao-in-nto.md
jmencak: > Oh, interesting.. @jmencak should NTO move the CRD there as well maybe? Or is there a reason for the separation we need to consider for PAO as well?

Yes, moving into `openshift/api` is definitely on my list of `TODO` tasks for the future.  Whether it is a good idea to move at the time of PAO/NTO merge, possibly, not sure about that.

yanirq: adding the move at the time of PAO/NTO merge under open questions.

enhancements/node-tuning/pao-in-nto.md
yanirq: @MarSik  should we consider v1 deprecation as a design goal here ?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,317 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [x] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
```
dhellmann, line 29: This section is really intended to be a summary of the proposal, not so much a summary of the state of existing code. I would keep the first paragraph, and add a line or two about wanting to add all of the existing implementation of PAO to NTO without adding new features. Then I would move most of the rest of the content down to the `Proposal` or `Implementation Details` section. If you put it in the `Proposal` section, you could add a new subsection there with a name like `Existing Implementation`.

enhancements/node-tuning/pao-in-nto.md
dhellmann: It would be good to capture the detail from this thread in the body of the document.

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,317 @@
...
+
+### API stability
+
+The `PerformanceProfile` CR uses an `performance.openshift.io/v2` API group which marks it as stable interface that needs to be maintained for about a year after official deprecation. We do not expect any change to the API as part of this proposal.
+
+### Bugs
+
+PAO bugs are currently tracked in The OpenShift Bugzilla component [Performance Addon Operator](https://bugzilla.redhat.com/buglist.cgi?component=Performance%20Addon%20Operator&list_id=12053428&product=OpenShift%20Container%20Platform)
+
+This should be kept as-is, the Telco team will own all the bugs. We might need to move the component under NTO or rename it.
```
dhellmann, line 134: I do think we're going to want to consider the bugzilla component question carefully. Moving to a sub-component of NTO seems like it makes sense, since that's the component users will see installed, but it would be good to have the teams decide what to do together.

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,317 @@
...
+- Maintainability: There are two options:
+  - Maintaining a separate copy of the code in the NTO repository.
+  - Vendoring the current PAO code as a "library" until we can stop supporting the standalone mode (4.6+).
+  From OCP perspective it makes little difference as it would just be part of NTO release process indeed.
+
+- API definitions: They would need to go into openshift/api but at this moment NTO does include its own API definitions under openshift/api.
+The NTO API definitions transition into openshift/api should be considered to be done at the time of PAO/NTO merge.
+
+- List of pertinent topology options (control plane and infrastructure) where this operator is pertinent:
+The assumption is that PAO is not installed at all when control plane topology is "external" and the infrastructure topology is not pertinent to the function
```
dhellmann, line 187: If the PAO becomes part of the NTO, do we achieve this by not installing the NTO? Is that already how NTO is handled for external control planes?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,317 @@
...
+since resources leveraged by the PAO are all managed externally when in this mode (e.g. node configuration, kubelet configuration).
+PAO is a control plane component that modifies the worker nodes (unless in single node mode where it affects masters).
+
+### Upgrade
+
+- PAO has been installed via OLM since 4.4. Once it moves to core we will have to figure out the upgrade procedure, because OLM will not remove the optional operator by itself.
+
+  - Implementations options:
```
dhellmann, line 195: We should pick one of these to describe here and document the others in the alternatives section below.

enhancements/node-tuning/pao-in-nto.md
dhellmann: Let's either do the v1 deprecation work before moving or after, but not at the same time.

enhancements/node-tuning/pao-in-nto.md
dhellmann: OK, my point is that it's a risk. It might be small, but it's there, so let's document it.

yanirq: adding the info under the risks section.

enhancements/node-tuning/pao-in-nto.md
yanirq: adding the additional info to this section.

enhancements/node-tuning/pao-in-nto.md
yanirq: @MarSik  @jmencak  Do we want to move this to the open questions section or have a decision on that ?

enhancements/node-tuning/pao-in-nto.md
yanirq: @jmencak   Can you affirm that this is the way NTO is handled for external control planes ?

jmencak: My understanding is that currently, IBM Cloud is the only cloud with external control plane topology.  NTO *is* being installed on IBM Cloud by using a [deployment manifest (annotation)](https://github.com/openshift/cluster-node-tuning-operator/blob/master/manifests/50-operator-ibm-cloud-managed.yaml#L5-L6).  Unless we merge PAO into NTO in a way that we can make use of the above deployment manifest (not what I believe the plan is), similar manifest for including PAO in IBM Cloud would be necessary.  Therefore, as things are planned now, PAO would *not* run on external control planes.

Currently, it is unclear to me what (if anything) needs to be done to include NTO (and PAO for that matter) on other external control plane topologies wrt. efforts such as HyperShift which will make it possible to run external control planes on platforms other than IBM Cloud.

enhancements/node-tuning/pao-in-nto.md
yanirq: v1 deprecation work will happen after moving to NTO. 
In the meanwhile https://github.com/openshift-kni/performance-addon-operators/pull/727 should cover the back and forth conventions between v1 and v2 as mentioned above

yanirq:
@dhellmann  who can we tag from the API team to review the proposal ?

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,323 @@
...
+[cnf-tests](https://catalog.redhat.com/software/containers/openshift4/cnf-tests-rhel8/5e9d81f65a134668769d560b?container-tabs=overview&gti-tabs=registry-tokens)
+that can be used to
+[validate customer deployments](https://docs.openshift.com/container-platform/4.8/scalability_and_performance/cnf-performance-addon-operator-for-low-latency-nodes.html#cnf-performing-end-to-end-tests-for-platform-verification_cnf-master) (this includes other aspects like SCTP, PTP, SR-IOV as well).
+
+### Code
+
+PAO code would be moved into the NTO repository as a separate controller. The Telco team would keep ownership as it has the necessary domain knowledge needed to implement the low latency tuning.
+
+The code currently uses Operator SDK libraries in some of the tests and this might need to be changed.
+Migrating the use of these libaries to [client-go](https://github.com/openshift/library-go)
```
sttts, line 105: library-go

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,323 @@
...
+would be one of the options.
+
+PAO also implements a [conversion webhook](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion)
+to migrate its custom resources during upgrade.
```
sttts, line 109: what does this mean? Why during upgrade?

enhancements/node-tuning/pao-in-nto.md
sttts: s/migration webhook/conversion webhook/

sttts:
@yanirq some wording around conversion is still odd in the text. But overall the migraton plan with a conversion webhook plus https://github.com/openshift-kni/performance-addon-operators/pull/727 looks good now.

enhancements/node-tuning/pao-in-nto.md
yanirq: Added a more elaborate explanation.

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -102,11 +102,11 @@ that can be used to
 PAO code would be moved into the NTO repository as a separate controller. The Telco team would keep ownership as it has the necessary domain knowledge needed to implement the low latency tuning.
 
 The code currently uses Operator SDK libraries in some of the tests and this might need to be changed.
-Migrating the use of these libaries to [client-go](https://github.com/openshift/library-go)
+Migrating the use of these libaries to [library-go](https://github.com/openshift/library-go)
 would be one of the options.
 
 PAO also implements a [conversion webhook](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion)
-to migrate its custom resources during upgrade.
+to convert its custom resources during upgrade to versions that are supported by the openshift version targeted.
```
sttts, line 109: my point is that this not during upgrade, but conversion are done in real-time on every request, even long after upgrade.

yanirq: @sttts  ack,  since this is a section about code in general it would make sense to drop the specific mention about upgrade for that paragraph then (following your comment)

yanirq:
@sttts  Thanks for the feedback.  Comments were addressed. 
Any additional reviews required here?

sttts:
@yanirq lgtm from my side.

sttts:
```
approvers:
  - TBD
```
looks wrong. How owns this topic from the staff engs?

yanirq:
> ```
> approvers:
>   - TBD
> ```
> 
> looks wrong. How owns this topic from the staff engs?

Added @dhellmann  to approvers list and @sttts  to reviewers list.

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,323 @@
...
+[cnf-tests](https://catalog.redhat.com/software/containers/openshift4/cnf-tests-rhel8/5e9d81f65a134668769d560b?container-tabs=overview&gti-tabs=registry-tokens)
+that can be used to
+[validate customer deployments](https://docs.openshift.com/container-platform/4.8/scalability_and_performance/cnf-performance-addon-operator-for-low-latency-nodes.html#cnf-performing-end-to-end-tests-for-platform-verification_cnf-master) (this includes other aspects like SCTP, PTP, SR-IOV as well).
+
+### Code
+
+PAO code would be moved into the NTO repository as a separate controller. The Telco team would keep ownership as it has the necessary domain knowledge needed to implement the low latency tuning.
+
+The code currently uses Operator SDK libraries in some of the tests and this might need to be changed.
+Migrating the use of these libaries to [library-go](https://github.com/openshift/library-go) or stripping them to [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime)
```
jmencak, line 106: `s/libaries/libraries/`

yanirq: done

yanirq:
@fromanirh  @jmencak  @mrunalp  since you are listed as reviewers on this PR in addition to @dhellmann , can you please have a final look on this proposal ?  (we are waiting for final acks from all peers , got an ack only from @sttts so far)  
Thanks.

jmencak:
> @fromanirh @jmencak @mrunalp since you are listed as reviewers on this PR in addition to @dhellmann , can you please have a final look on this proposal ? (we are waiting for final acks from all peers , got an ack only from @sttts so far) Thanks.

lgtm from my side.  Nit: a missing new line in front of `### Test Plan`

fromanirh:
lgt

> @fromanirh @jmencak @mrunalp since you are listed as reviewers on this PR in addition to @dhellmann , can you please have a final look on this proposal ? (we are waiting for final acks from all peers , got an ack only from @sttts so far) Thanks.

lgtm form my side for the general direction and most notably for the idea. There are some details in the upgrade/downgrade flow that I'd like to explore in a bit more details, but that can be postponed I believe.

yanirq:
@dhellmann  I believe we have all the acks (just not in the github label format)

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,333 @@
...
+Validation and conversion webhook (TBD)
+
+- Updates to openshift conformance test suite if included by default:
+All basic checks are doable on all linux platforms except when real-time kernel is needed. That case used to be only supported on GCP (and bare metal).
+Measuring real latency requires bare metal, but is not needed for conformance.
+
+
+### Risks and Mitigations
+
+- Since the performance profile API will be visible for users on all clusters there is a risk that a cluster admin might tune nodes improperly.
```
mrunalp, line 172: nit: As a mitigation we can say that we document clearly that admins proceed with caution.

yanirq: done

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,333 @@
...
+
+- CI: One of the test suites (4_latency) implements an opinionated latency testing suite to get comparable results from customers and to prevent configuration / command line invocation mistakes. Keeping it in PAO/NTO or moving it into cnf-tests is TBD.
+
+- Maintainability: There are two options:
+  - Maintaining a separate copy of the code in the NTO repository.
+  - Vendoring the current PAO code as a "library" until we can stop supporting the standalone mode (4.6+).
+  From OCP perspective it makes little difference as it would just be part of NTO release process indeed.
+
+- API definitions: They would need to go into openshift/api but at this moment NTO does include its own API definitions under openshift/api.
+The NTO API definitions transition into openshift/api should be considered to be done at the time of PAO/NTO merge.
```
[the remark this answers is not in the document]
yanirq: I rephrased it a bit. 
There is no specific migration plan captured at the moment and this is/was considered as a task to be done in NTO which would probably need to move a bit faster now.
This is also the reason I captured this under the open questions section

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,333 @@
...
+  - Deleting any CVO-managed resources added by the new version. The
+    CVO does not currently delete resources that no longer exist in
+    the target version.
+
+### Version Skew Strategy
+
+How will the component handle version skew with other components?
+What are the guarantees? Make sure this is in the test plan.
+
+Consider the following in developing a version skew strategy for this
```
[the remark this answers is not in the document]
yanirq: done

enhancements/node-tuning/pao-in-nto.md
```diff
@@ -0,0 +1,333 @@
...
+  - @dhellmann
+  - @sttts
+approvers:
+  - @dhellmann
+creation-date: 2021-08-12
+last-updated: 2021-08-12
+status: provisional
+---
+
+# PAO as part of core OCP
```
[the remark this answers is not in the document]
yanirq: done

PR #870, opened 2021-08-19
Versioning cincinnati api and json schema

PratikMahajan:


enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,116 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+
+## Summary
+As of today, we declare the schema version in Cincinnati Graph URI. We currently, do not embed the version information
+in json schema. The enhancement will embed the version information in the json schema. This has the benefit of being
```
wking, line 36: nit: [trailing spaces][1] here and elsewhere in the file.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_enhancements/870/pull-ci-openshift-enhancements-master-markdownlint/1428436850845945856#1:build-log.txt%3A18

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,116 @@
...
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+
+## Summary
+As of today, we declare the schema version in Cincinnati Graph URI. We currently, do not embed the version information
+in json schema. The enhancement will embed the version information in the json schema. This has the benefit of being 
+easy to implement and to pass around if we ship Cincinnati JSON over non-HTTP channels. Enhancement will also implement
+media type versioning while requesting the graph e.g. `ContentType: application/vnd.redhat.cincinnati.v1+json`.
```
wking, line 38: nit: "while requesting the graph" would be [the `Accept` header][1].  I like that content negotiation, but for the non-HTTP channel shipping angle, the header is [`Content-Type`][2] (with a hyphen).  Not sure which way you want to go here.  Maybe just step back and say:

> The enhancement will also begin versioning Cincinnati payloads by media type, for example `application/vnd.redhat.cincinnati.v1+json`.

[1]: https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2
[2]: https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,116 @@
...
+easy to implement and to pass around if we ship Cincinnati JSON over non-HTTP channels. Enhancement will also implement
+media type versioning while requesting the graph e.g. `ContentType: application/vnd.redhat.cincinnati.v1+json`. 
+The move to supplying requested version in media type field will require us to move towards a generic un-versioned Graph
+URI, with the move, we'll still support old URI for foreseeable future. 
+
+## Motivation
+
+### Goals
+1. Embed version information in json schema.
+2. Implement media type versioning for clients requesting a specific schema version of cincinnati
```
wking, line 46: nit: when discussing the protocol (in a non-slug form), I think the canonical casing is "Cincinnati" (e.g. [here][1]).

[1]: https://github.com/openshift/cincinnati/blame/d6a64f81a50e68e5feae52df9d5ae23fa7ab6dad/docs/design/cincinnati.md#L19

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,116 @@
...
+3. Move to a generic un-versioned URI for cincinnati
+4. Cincinnati continues to support older clients after adding support for new clients for foreseeable future.
+5. Add metrics to identify hits on URIs and versions.
+
+### Non-Goals
+
```
wking, line 52: add dropping the old URI as a non-goal?

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,116 @@
...
+    "version": 1,
+    "nodes": [],
+    "edges": []
+}
+```
+
+#### Media Type Versioning 
+Currently, the versioning is done by using the versioned URI and does not support media type versioning. The user should 
+be able to request a specific version of cincinnati graph by supplying the version in media type field of the request. 
+for e.g. `ContentType: application/vnd.redhat.cincinnati.v1+json`. Here are some of the
```
wking, line 70: nit: `ContentType` -> `Accept`, per [this thread][1].

[1]: https://github.com/openshift/enhancements/pull/870#discussion_r693265289

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,116 @@
...
+The current URI follows `{URL}/v1/graph`. The new un-versioned URI will follow `{URL}/graph`. The existing URI can be 
+redirected to the new URI. This can be done through cincinnati. In the case where an old URI is being called, cincinnati 
+will by default return the graph that is compatible with `v1`.      
+
+
+#### Version metrics
+Cincinnati should also implement a way to understand how many clients are consuming the various versions. 
+This can be done by implementing metrics for old URI vs new URI. Similarly, metrics should be implemented to understand
+the consumption of various media types (versions). This will help us to make decisions on deprecating the old versions 
+and URI.
```
wking, line 86: do we want to spell this out in more detail?  I'm not sure if the host cluster's Router gives us any metrics we can use out of the box.  But if we need out own, new metrics, probably a single `cincinnati_pe_http_response_total` counter with labels for `uri` and `version`?  Or something...

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,116 @@
...
+### Test Plan
+* cincinnati tests to make sure we are not breaking older versions 
+* cincinnati returns correct resource version when the content type is provided.
+
+### Graduation Criteria
+GA. When it works, we ship it.
+
+
+### Version Skew Strategy
+No special consideration.
```
wking, line 106: we may want to point out that the need to support v1 forever only applies to the URI.  We control all the supported clients, and [set maintenance cutoffs][1], so there will come a time when we no longer care about supporting clients who only speak v1.  The endless maintenance burden is just for clients that can speak v2, etc., but still make their v2 Cincinnati requests via the original, nominally v1, URI because we used to (before openshift/installer#4112, 4.7), stash that in `spec.upstream`, and robots touching `spec` properties to try to auto-recover that for born-before-4.7 clusters by clearing `spec.upstream` when it matches our old default would potentially override an admin who explicitly wanted the cluster pulling from that URI.  And maybe they were on a restricted network with a DNS horizon or some such, and they really meant to stick with that upstream URI?  Anyway, some words about "we are stuck supporting the current URI forever, but do not have to support the v1 payload format forever" would be good.

[1]: https://access.redhat.com/support/policy/updates/openshift/#dates

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Implementation History
+
+## Drawbacks
+
+
+## Alternatives
+We can continue with the current URI based versioning for the API and adding version information in the json schema.     
+**Drawback**: For every minor non-breaking change with a bumped version, the client will have to change the URI with 
+the new version to get upgrades.
```
wking, line 116: Main issue with this is that [proactive negotiation][1] makes compat between simple clients and a smarter server easy, and that's the approach we're aiming for with the Cincinnati service and the cluster-version operator.  Teaching the CVO to walk possible URIs or hit a list-of-available-versions endpoint would be more complicated.

Another alternative is adding content-type negotiation, but sticking with the existing URI.  You should add a subsection on that, and point out that we're avoiding it because folks might find the dissonance confusing while reading docs talking about v2 content from a v1 URI.

[1]: https://datatracker.ietf.org/doc/html/rfc7231#section-3.4.1

enhancements/update/versioning-cincinnati-api.md
PratikMahajan: + wanna keep options open to deprecate it in the future?

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,122 @@
...
+  individual versions will not.
+
+## Implementation History
+
+## Drawbacks
+
+
+## Alternatives
+* We can continue with the current URI based versioning for the API and adding version information in the json schema.  
+  **Drawback**: For every minor non-breaking change with a bumped version, the client will have to change the URI with
```
dhellmann, line 118: Since the default is to return the oldest supported format, won't the client have to explicitly request the new version anyway?

dhellmann (commented):
The linter job is failing because some headers that are required from the template are missing in this doc. Please restore those headers and leave an explicit comment if you believe that section does not apply to this change.

```
enhancements/update/versioning-cincinnati-api.md missing "### User Stories"
enhancements/update/versioning-cincinnati-api.md missing "#### Dev Preview -> Tech Preview"
enhancements/update/versioning-cincinnati-api.md missing "#### Tech Preview -> GA"
enhancements/update/versioning-cincinnati-api.md missing "#### Removing a deprecated feature"
enhancements/update/versioning-cincinnati-api.md missing "### Upgrade / Downgrade Strategy"
```

enhancements/update/versioning-cincinnati-api.md
wking: Yes.  Can we add subsections here to go into more detail?  Something like:

```markdown
## Alternatives

### URI versioning

In addition to adding version information to the JSON Schema for downstream consumers, we can continue with the current URI based versioning for the Cincinnati API.
The benefit to that approach is that we don't need to deprecate the existing v1 URI.
A downside is that clients need to either:

* Poll a series of versioned URIs until they find one the service supports.
  For example, the client could poll the v3 URI, get a 404, poll the v2 URI, get a 404, and then successfully hit the v1 URI if the client preferred v3, but the service they were hitting only supported v1.
* Hit a new URI that lists versions supported by the service.
  For example, `/graph/versions`, which would return "I serve v1 at `$URI_1`, v2 at `$URI_2`, etc.", from which the client selects its favorite.
  The HTTP RFCs call this [reactive negotiation][1].

While these alternatives would work, it's simpler in the client code to construct an `Accept` header and make a single request, and one of the cluster-version operator's design goals is to make the client as simple as possible, because bugs in the cluster-version operator can be hard to recover from.

### Media-type versioning with the existing, versioned URI

We could serve multiple, media-typed versions from the existing add content-type negotiation with the existing `{URL}/v1/graph` location.
The benefit to that approach is that we don't need to deprecate the existing URI.
The drawback is that users might find the dissonance confusing while reading docs talking about v2 content from a v1 URI.

[1]: https://datatracker.ietf.org/doc/html/rfc7231#section-3.4.2
```

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,122 @@
...
+    "version": 1,
+    "nodes": [],
+    "edges": []
+}
+```
+
+#### Media Type Versioning
+Currently, the versioning is done by using the versioned URI and does not support media type versioning. The user should
+be able to request a specific version of cincinnati graph by supplying the version in media type field of the request.
+for example `application/vnd.redhat.cincinnati.v1+json`. Here are some of the
```
wking, line 70: I think you want to call out [the `Accept` header][1] explicitly in this section, to show that this is [proactive negotiation][2].  Something like:

```Markdown
## Media Type Versioning

Currently, the versioning is done by using the versioned URI and does not support media type versioning.
This enhancement allows users to request a specific version of the Cincinnati specification [proactively][proactive-negotiation] with [the `Accept` header][accept].
For example, `Accept: application/vnd.redhat.cincinnati.v1+json` says "I only understand v1 Cincinnati graphs; please give me that format if possible".

In the event where the media type is not provided, or `application/json` is provided, Cincinnati should return a default
response that is compatible with the oldest supported version so as not to break old clusters from upgrading.

In a similar space, the Open Container Initiative Image Format Specification [uses versioned media types][image-spec-media-types] to identify versions of configuration and other formats.

[accept]: https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2
[image-spec-media-types]: https://github.com/opencontainers/image-spec/blob/v1.0.1/media-types.md
[proactive-negotiation]: https://datatracker.ietf.org/doc/html/rfc7231#section-3.4.1
```

[1]: https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2
[2]: https://datatracker.ietf.org/doc/html/rfc7231#section-3.4.1

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,122 @@
...
+
+
+#### Version metrics
+Cincinnati should also implement a way to understand how many clients are consuming the various versions.
+This can be done by implementing metrics for old URI vs new URI in Cincinnati. Similarly, metrics should be implemented
+to understand the consumption of various media types (versions). This will help us to make decisions on deprecating the
+old versions and URI. For example, a single `cincinnati_pe_http_response_total` counter with labels for `uri` and `version`
+
+
+### Risks and Mitigations
```
wking, line 89: The HTTP spec lists some disadvantages in [its proactive negotiation section][1].  We should reference the HTTP spec and bring those over, and explain why we think they are acceptable in this case.  For example:

```Markdown
Another disadvantage listed in the HTTP RFC is inefficiency when  only a small percentage of responses have multiple representations.
But in this case, including the desired/acceptable version(s) in a per-request header is about as efficient as hitting a versioned URI that the service supports, and much more efficient than [the alternatives](#uri-versioning) for supporting versioned URIs when the service does not support the client's preferred version.
```

Where `#uri-versioning` links [the relevant alternatives subsection][2].

[1]: https://datatracker.ietf.org/doc/html/rfc7231#section-3.4.1
[2]: https://github.com/openshift/enhancements/pull/870#discussion_r694366627

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,122 @@
...
+    "nodes": [],
+    "edges": []
+}
+```
+
+#### Media Type Versioning
+Currently, the versioning is done by using the versioned URI and does not support media type versioning. The user should
+be able to request a specific version of cincinnati graph by supplying the version in media type field of the request.
+for example `application/vnd.redhat.cincinnati.v1+json`. Here are some of the
+[examples of media type versioning](https://github.com/opencontainers/image-spec/blob/v1.0.1/media-types.md).
```
LalatenduMohanty, line 71: I think you are trying to say that we will support the `OCI Image Media Types` as media types is more generic than that https://datatracker.ietf.org/doc/html/rfc6838

enhancements/update/versioning-cincinnati-api.md
LalatenduMohanty: This sounds good to me.

LalatenduMohanty:
Mostly looks good to me. Also we should include https://github.com/openshift/enhancements/pull/870#discussion_r694322254

enhancements/update/versioning-cincinnati-api.md
wking: Media types are generic, and RFC 6838 mentions versions in passing a few times.  But many generic media types are not versioned (e.g. image/jpg).  The OCI image types, on the other hand, include some explicit examples of media types being used for versioning.

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,171 @@
+---
+title: versioning-cincinnati-api
+authors:
+- "@pratikmahajan"
+  reviewers:
+- @openshift/openshift-team-cincinnati-maintainers
+- @openshift/openshift-team-cincinnati
+- @openshift/team-cincinnati-operator
```
wking, line 8: nit: teams evolve, so before this lands its probably best to bump it to match the set of folks who have weighed in so far.

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,171 @@
...
+}
+```
+
+#### Media Type Versioning
+Currently, the versioning is done by using the versioned URI and does not support media type versioning.
+This enhancement allows users to request a specific version of the Cincinnati specification
+[proactively][proactive-negotiation] with [the `Accept` header][accept].  
+For example, `Accept: application/vnd.redhat.cincinnati.v1+json` says "I only understand v1 Cincinnati graphs;
+please give me that format if possible".
+[examples of OCI Image Media Types](https://github.com/opencontainers/image-spec/blob/v1.0.1/media-types.md).
```
wking, line 77: nit: narrowly "examples" -> "Examples" to start the sentence.  But probably expand this to explain that we are pointing at these OCI types as examples of existing media types that are used for versioning.  I'd floated an example sentence over [here][1].

[1]: https://github.com/openshift/enhancements/pull/870#discussion_r694371978

wking: oh, wait, you have that sentence down below.  I think you can just drop this one entirely then.

enhancements/update/versioning-cincinnati-api.md
PratikMahajan: fixed

enhancements/update/versioning-cincinnati-api.md
PratikMahajan: added

enhancements/update/versioning-cincinnati-api.md
PratikMahajan: fixed

enhancements/update/versioning-cincinnati-api.md
PratikMahajan: changed reviewers to add specific people

LalatenduMohanty (approved):
/lgtm

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,168 @@
...
+response that is compatible with the oldest supported version so as not to break old clusters from upgrading.
+
+In a similar space, the Open Container Initiative Image Format Specification
+[uses versioned media types][image-spec-media-types] to identify versions of configuration and other formats.
+
+
+#### URI Change
+The current URI follows `{URL}/v1/graph`. The new un-versioned URI will follow `{URL}/graph`. The existing URI can be
+redirected to the new URI. This can be done through cincinnati. In the case where an old URI is being called, cincinnati
+will by default return the graph that is compatible with the oldest supported version.
```
sdodson, line 85: The oldest being v1 specifically or would a backwards compatible v1.1 be available at this endpoint too?

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,168 @@
...
+
+#### Version metrics
+Cincinnati should also implement a way to understand how many clients are consuming the various versions.
+This can be done by implementing metrics for old URI vs new URI in Cincinnati. Similarly, metrics should be implemented
+to understand the consumption of various media types (versions). This will help us to make decisions on deprecating the
+old versions and URI. For example, a single `cincinnati_pe_http_response_total` counter with labels for `uri` and `version`
+
+
+### Risks and Mitigations
+* As there are some customers that have old URI hardcoded while creating the cluster, we will have to support old URI
```
sdodson, line 96: Should we consider installer validation to block new installs using the versioned URI when the client expects a newer version than v1? For example if 4.10 CVO uses a v2 API we don't want to allow them to inadvertantly downrev to v1 and lose targetted edge blocking due to their configuration.

enhancements/update/versioning-cincinnati-api.md
PratikMahajan: we are gonna consider both the endpoints as one. Backward compatible v1.1 will be available at this endpoint as well. Currently, we will be supporting `v1` as that's the oldest graph we know. But going forward we can make decisions when to drop specific versions based on their usage.

enhancements/update/versioning-cincinnati-api.md
PratikMahajan: we are moving away from versioned URI to unversioned. So all requests from `/v1/graph` will be handled as if they're coming to `/graph`. Going forward we will be accepting the version using Media Type Versioning. 
So, if someone hits the old URI with a correct media type version, they'll end up getting the intended graph or if someone hits the new URI without the media type version, they'll end up getting the oldest supported graph.  
Installer validation will speed up the process of shifting to the new URI!

enhancements/update/versioning-cincinnati-api.md
```diff
@@ -0,0 +1,168 @@
...
+The current URI follows `{URL}/v1/graph`. The new un-versioned URI will follow `{URL}/graph`. The existing URI can be
+redirected to the new URI. This can be done through cincinnati. In the case where an old URI is being called, cincinnati
+will by default return the graph that is compatible with the oldest supported version.
+
+
+#### Version metrics
+Cincinnati should also implement a way to understand how many clients are consuming the various versions.
+This can be done by implementing metrics for old URI vs new URI in Cincinnati. Similarly, metrics should be implemented
+to understand the consumption of various media types (versions). This will help us to make decisions on deprecating the
+old versions and URI. For example, a single `cincinnati_pe_http_response_total` counter with labels for `uri` and `version`
```
wking, line 92: nit: line is missing a terminal period `.`.

enhancements/update/versioning-cincinnati-api.md
sdodson: So the conclusion is that we'd never get rid of the legacy URL. That's fine.

PR #871, opened 2021-08-20
Enable Baremetal on other Platforms to support centralized host management

asalkeld:
cluster-baremetal-operator is only enabled when deploying an OpenShift cluster with the baremetal
platform (via the IPI or AI workflow). Having the ability to manage baremetal nodes from
clusters without requiring the cluster to be on baremetal would be beneficial to customers.

https://github.com/openshift/cluster-baremetal-operator/pull/189#issuecomment-892641328
/cc @hardys

### DIFF
--- a/enhancements/baremetal/enable-baremetal-on-other-platforms.md
+++ b/enhancements/baremetal/enable-baremetal-on-other-platforms.md
@@ -4,12 +4,11 @@ authors:
   - "@asalkeld"
   - "@sadasu"
 reviewers:
-  - TBD
   - "@hardy"
   - "@romfreiman"
   - "@dhellman"
 approvers:
-  - TBD
+  - "@hardy"
 creation-date: 2021-08-20
 last-updated: 2021-08-20
 status: implementable
@@ -19,7 +18,7 @@ replaces:
 superseded-by:
 ---
 
-# Enable Baremetal on other Platforms to support ZTP
+# Enable baremetal on other Platforms to support centralized host management
 
 ## Release Signoff Checklist
 
@@ -43,11 +42,15 @@ An initial driver of this feature are the centralized host management use cases
 in edge topologies, which without this feature, is restricted to having the
 central OpenShift cluster deployed on baremetal.
 
+See:
+- https://github.com/openshift/enhancements/blob/master/enhancements/installer/agent-based-installation-in-hive.md
+- https://github.com/openshift/assisted-service/tree/master/docs/hive-integration
+
 ### Goals
 
 The specific goals of this proposal are to:
 
-Support the ZTP (Zero Touch Provisioning) use case by partially enabling Baremetal Host API
+Support the centralized host management use case by partially enabling Baremetal Host API
 on the following platforms:
 - None
 - OpenStack
@@ -55,14 +58,14 @@ on the following platforms:
 
 We will be successful when:
 
-ZTP can deploy clusters when running on the above platforms.
+Centralized host management can deploy clusters when running on the above platforms.
 
 ### Non-Goals
 
 Allow Baremetal Host API to be fully enabled on all platforms.
 
 Note that the scope of this epic is only enablement of the Metal3
-BaremetalHost API and associated controller/services - the edge ZTP flow
+BaremetalHost API and associated controller/services - the centralized host management flow
 currently interacts directly via this API without any Machine API integration.
 
 ## Proposal
@@ -71,7 +74,8 @@ BMO (baremetal-operator) provides the Baremetal Host API, it in turn is configur
 and managed by CBO (cluster-baremetal-operator).
 
 CBO reads the Provisioning CR that is created by the installer on baremetal platforms
-and uses that to configure and deploy BMO.
+and uses that to configure and deploy BMO. In the case of non-baremetal platforms
+the user (or automation) will need to define the Provisioning CR.
 
 Currently CBO checks the platform and if it is not baremetal it will be in a "disabled" state i.e. it will
 1. set status.conditions Disabled=true and
@@ -80,14 +84,22 @@ Currently CBO checks the platform and if it is not baremetal it will be in a "di
 This proposal is to allow CBO to be enabled on the Platforms: None, OpenStack or vSphere.
 
 Further (to restrict the testing matrix) the allowed configuration options
-of the Provisioning CR will be restricted to exactly those required by ZTP.
+of the Provisioning CR will be restricted to exactly those required by centralized host management.
 
 *Only spec.provisioningNetwork=Disabled mode will be accepted in the Provisioning CR.*
 
-If any other provisioningNetwork mode is set, the CBO webhook will refuse the change.
+If any other provisioningNetwork mode is set, the CBO webhook will refuse the change
+in the usual case, but if defined before upgrading the operator, the Reconcile loop
+must always validate the Provisioning CR.
 
-Note: documentation will need to be added to the ZTP docs explaining how to create
-a Provisioning CR. Current doc are here:
+Note:
+
+1. when the Provisioning CR is set to provisioningNetwork=Disabled mode, worker
+nodes would be booted via virtual media. This removes the requirement for the
+Provisioning Network which can be expected to be available only in Baremetal platform types.
+
+2. documentation will need to be added to the centralized host management documentation
+explaining how to create a Provisioning CR. Current documentation is here:
 https://github.com/openshift/assisted-service/blob/8880093ef5ce041d4c1951ffd5ea1096991ec3ee/docs/user-guide/assisted-service-on-openshift.md#configure-bare-metal-operator
 
 
@@ -97,7 +109,7 @@ https://github.com/openshift/assisted-service/blob/8880093ef5ce041d4c1951ffd5ea1
 
 No change.
 
-#### Story 2 - ZTP use case
+#### Story 2 - centralized host management use case
 
 As a user of a hub cluster that performs central infrastructure management, and
 optionally zero-touch provisioning, I need to provision hosts using the k8s-native
@@ -114,10 +126,7 @@ all platforms and with full feature set. However it is still a potential issue.
 
 ### Open Questions [optional]
 
-1. Are these all the platforms that are required?
-2. Is just enabling spec.provisioningNetwork=Disabled mode sufficient?
-3. What will create the Provisioning CR?
-4. Where will the feature be e2e tested?
+1. Where will the feature be e2e tested?
 
 ### Test Plan
 
@@ -127,13 +136,16 @@ We will add unit tests to confirm that cluster-baremetal-operator:
 * is enabled on the required platforms.
 * will restrict functionality on these platforms to ProvisioningNetork=Disabled.
 
-#### e2e Testing
+#### Functional Testing
 
-An e2e test will be written that will:
+An e2e test will be written in the Assisted Installer CI that will:
+1. create one of the platforms above (SNO Platform=None might be the easiest) with Assisted Service.
+2. confirm that CBO is enabled
+3. create a Provisioning CR and confirm that BMO is running
+4. provision a baremetal cluster
 
-1. create one of the platforms above (SNO Platform=None might be the easiest).
-2. Then create the Provisioning CR and wait for CBO to be ready
-3. provision a baremetal host.
+QE will validate the remaining platforms that are supported to reduce the load
+on CI.
 
 ### Graduation Criteria
 
@@ -151,8 +163,8 @@ cluster-baremetal-operator will upgrade as it currently does, this is only a
 minor change in functionality.
 
 On the platforms (None, OpenStack and vSphere) where the operator was in a disabled
-state, after been upgraded it will move into an enabled state. However in all but ZTP
-use cases nothing will change as there is no Provisioning CR.
+state, after been upgraded it will move into an enabled state. However in all but
+centralized host management use cases nothing will change as there is no Provisioning CR.
 
 ### Version Skew Strategy
 
--- a/enhancements/baremetal/enable-baremetal-on-other-platforms.md
+++ b/enhancements/baremetal/enable-baremetal-on-other-platforms.md
@@ -4,16 +4,16 @@ authors:
   - "@asalkeld"
   - "@sadasu"
 reviewers:
-  - "@hardy"
+  - "@hardys"
   - "@romfreiman"
   - "@dhellman"
 approvers:
-  - "@hardy"
+  - "@hardys"
 creation-date: 2021-08-20
 last-updated: 2021-08-20
 status: implementable
 see-also:
-  - "/enhancements/baremetal-provisiong-config.md"
+  - "/enhancements/baremetal/baremetal-provisioning-config.md"
 replaces:
 superseded-by:
 ---
@@ -64,9 +64,12 @@ Centralized host management can deploy clusters when running on the above platfo
 
 Allow Baremetal Host API to be fully enabled on all platforms.
 
-Note that the scope of this epic is only enablement of the Metal3
-BaremetalHost API and associated controller/services - the centralized host management flow
-currently interacts directly via this API without any Machine API integration.
+Allow Machine API integration across platform types.
+- The centralized host management flow currently interacts directly via this API
+  without any Machine API integration.
+- This means further work will be required via a future enhancement to enable the
+  single cluster case where a combination of e.g VM controlplane and Baremetal
+  workers is desired.
 
 ## Proposal
 
@@ -106,8 +109,7 @@ nodes would be booted via virtual media. This removes the requirement for the
 Provisioning Network which can be expected to be available only in Baremetal platform types.
 
 2. documentation will need to be added to the centralized host management documentation
-explaining how to create a Provisioning CR. Current documentation is here:
-https://github.com/openshift/assisted-service/blob/8880093ef5ce041d4c1951ffd5ea1096991ec3ee/docs/user-guide/assisted-service-on-openshift.md#configure-bare-metal-operator
+explaining how to create and update a Provisioning CR for these platforms.
 
 ### User Stories
 
@@ -191,7 +193,6 @@ Customers can instead create a dedicated baremetal cluster to use as the hub
 cluster.
 
 Another alternative is to additionally distribute baremetal-operator as an optional
-operator with OLM. It's probably not a good idea, but worth enumerating as an alternative.
-The main downsides are the complexity of releasing and distributing the same project
-two different ways, and the potential for install-time confusion or conflict over
-which method should be used to install it.
+operator with OLM. The main downsides are the complexity of releasing and
+distributing the same project two different ways, and the potential for install-time
+confusion or conflict over which method should be used to install it.
--- a/enhancements/baremetal/enable-baremetal-on-other-platforms.md
+++ b/enhancements/baremetal/enable-baremetal-on-other-platforms.md
@@ -81,7 +81,14 @@ Currently CBO checks the platform and if it is not baremetal it will be in a "di
 1. set status.conditions Disabled=true and
 2. not read or process the Provisioning CR and thus not deploy baremetal-operator.
 
-This proposal is to allow CBO to be enabled on the Platforms: None, OpenStack or vSphere.
+This proposal is to allow CBO to be enabled on the following platforms only:
+- Baremetal (current)
+- None
+- OpenStack
+- vSphere.
+
+Note: The Hypershift use case will be explicitly disallowed by disabling CBO when
+Infrastructure.Status.ControlPlaneTopology == "External".
 
 Further (to restrict the testing matrix) the allowed configuration options
 of the Provisioning CR will be restricted to exactly those required by centralized host management.
@@ -101,7 +108,6 @@ Provisioning Network which can be expected to be available only in Baremetal pla
 2. documentation will need to be added to the centralized host management documentation
 explaining how to create a Provisioning CR. Current documentation is here:
 https://github.com/openshift/assisted-service/blob/8880093ef5ce041d4c1951ffd5ea1096991ec3ee/docs/user-guide/assisted-service-on-openshift.md#configure-bare-metal-operator
-
 
 ### User Stories
 
--- a/enhancements/baremetal/enable-baremetal-on-other-platforms.md
+++ b/enhancements/baremetal/enable-baremetal-on-other-platforms.md
@@ -51,7 +51,7 @@ See:
 The specific goals of this proposal are to:
 
 Support the centralized host management use case by partially enabling Baremetal Host API
-on the following platforms:
+on the following on-premise platforms:
 - None
 - OpenStack
 - vSphere
@@ -99,8 +99,9 @@ of the Provisioning CR will be restricted to exactly those required by centraliz
 *Only spec.provisioningNetwork=Disabled mode will be accepted in the Provisioning CR.*
 
 If any other provisioningNetwork mode is set, the CBO webhook will refuse the change
-in the usual case, but if defined before upgrading the operator, the Reconcile loop
-must always validate the Provisioning CR.
+in the usual way, but if defined before upgrading the operator, the Reconcile loop must always
+validate the Provisioning CR
+(by setting ClusterOperator/baremetal condition[InvalidConfiguration] = true ).
 
 Note:
 
@@ -130,11 +131,10 @@ There is concern that *random* customers will use this feature out of context
 and create support burden. This is why we have not suggested enabling CBO on
 all platforms and with full feature set. However it is still a potential issue.
 
+Another mitigation for this is to avoid documenting this outside of the CIM/ZTP case.
+For that reason this change won't be documented as a standalone feature, only in the context of CIM/ZTP.
+
 ## Design Details
-
-### Open Questions [optional]
-
-1. Where will the feature be e2e tested?
 
 ### Test Plan
 
--- a/enhancements/baremetal/enable-baremetal-on-other-platforms.md
+++ b/enhancements/baremetal/enable-baremetal-on-other-platforms.md
@@ -1,0 +1,179 @@
+---
+title: enable-baremetal-on-other-platforms
+authors:
+  - "@asalkeld"
+  - "@sadasu"
+reviewers:
+  - TBD
+  - "@hardy"
+  - "@romfreiman"
+  - "@dhellman"
+approvers:
+  - TBD
+creation-date: 2021-08-20
+last-updated: 2021-08-20
+status: implementable
+see-also:
+  - "/enhancements/baremetal-provisiong-config.md"
+replaces:
+superseded-by:
+---
+
+# Enable Baremetal on other Platforms to support ZTP
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Baremetal Host API is only available when deploying an OpenShift cluster with the baremetal
+platform (via the IPI or AI (Assisted Installer) workflow). Having the ability to
+manage baremetal hosts from clusters without requiring the cluster to be on baremetal
+would be beneficial to customers.
+
+## Motivation
+
+An initial driver of this feature are the centralized host management use cases
+in edge topologies, which without this feature, is restricted to having the
+central OpenShift cluster deployed on baremetal.
+
+### Goals
+
+The specific goals of this proposal are to:
+
+Support the ZTP (Zero Touch Provisioning) use case by partially enabling Baremetal Host API
+on the following platforms:
+- None
+- OpenStack
+- vSphere
+
+We will be successful when:
+
+ZTP can deploy clusters when running on the above platforms.
+
+### Non-Goals
+
+Allow Baremetal Host API to be fully enabled on all platforms.
+
+Note that the scope of this epic is only enablement of the Metal3
+BaremetalHost API and associated controller/services - the edge ZTP flow
+currently interacts directly via this API without any Machine API integration.
+
+## Proposal
+
+BMO (baremetal-operator) provides the Baremetal Host API, it in turn is configured
+and managed by CBO (cluster-baremetal-operator).
+
+CBO reads the Provisioning CR that is created by the installer on baremetal platforms
+and uses that to configure and deploy BMO.
+
+Currently CBO checks the platform and if it is not baremetal it will be in a "disabled" state i.e. it will
+1. set status.conditions Disabled=true and
+2. not read or process the Provisioning CR and thus not deploy baremetal-operator.
+
+This proposal is to allow CBO to be enabled on the Platforms: None, OpenStack or vSphere.
+
+Further (to restrict the testing matrix) the allowed configuration options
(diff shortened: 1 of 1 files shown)


enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,179 @@
...
+
+### User Stories
+
+#### Story 1 - Current IPI baremetal platform use case
+
+No change.
+
+#### Story 2 - ZTP use case
+
+If the platform is one of "None, OpenStack or vSphere". cluster-baremetal-operator
```
mhrivnak, line 94: This text describes how to implement a solution, not a user story. This text would be good to re-locate into the "Proposal" section. The user story section should focus on what a user needs to accomplish.

A user story would be: "As a user of a hub cluster that performs central infrastructure management, and optionally zero-touch provisioning, I need to provision hosts using a k8s-native API even when the hub cluster has a platform of None, OpenStack, or vSphere."

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,179 @@
...
+## Implementation History
+
+This PR is the current WIP implementation: https://github.com/openshift/cluster-baremetal-operator/pull/189
+
+## Drawbacks
+
+There is concern that *random* customers will use this feature out of context
+and create support burden.
+
+## Alternatives
```
mhrivnak, line 176: Another alternative is to additionally distribute baremetal-operator as an optional operator with OLM. It's probably not a good idea, but worth enumerating as an alternative. The main downsides are the complexity of releasing and distributing the same project two different ways, and the potential for install-time confusion or conflict over which method should be used to install it.

mhrivnak (changes requested):

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,179 @@
...
+
+An initial driver of this feature are the centralized host management use cases
+in edge topologies, which without this feature, is restricted to having the
+central OpenShift cluster deployed on bare metal.
+
+### Goals
+
+The specific goals of this proposal are to:
+
+Support the ZTP use case by partially enabling cluster-baremetal-operator on the following platforms:
```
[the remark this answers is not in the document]
dhellmann: Is there another enhancement document describing the ZTP use case?

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,179 @@
...
+1. set status.conditions Disabled=true and
+2. not read or process the Provisioning CR and thus not deploy baremetal-operator.
+
+This proposal is to allow cluster-baremetal-operator to be enabled for the
+ZTP use case which is Platforms: None, OpenStack or vSphere.
+
+Further (to restrict the testing matrix) the allowed configuration options
+of the Provisioning CR will be restricted to exactly those required by ZTP.
+
+*Only spec.provisioningNetwork=Disabled mode will be accepted in the Provisioning CR.*
```
dhellmann, line 80: What will the CBO do if another network mode is used?

asalkeld: CBO has a webhook, so we can just fail the validation in the validating webhook.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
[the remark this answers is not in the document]
asalkeld: not sure, I'll look @dhellmann

asalkeld:
thanks for the great feedback @mhrivnak and @dhellmann , I have updated as requested

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### Unit Testing
+
+We will add unit tests to confirm that cluster-baremetal-operator:
+* is enabled on the required platforms.
+* will restrict functionality on these platforms to ProvisioningNetork=Disabled.
+
+#### e2e Testing
+
+It is expected that these use cases will be functionally tested by the Assisted Installer
```
romfreiman, line 126: I do expect Metal to have a CI verifying that the basic functionality works - outside of ZTP scope. AI is agnostic to a platform it runs on, hence AI team can not own multiple platforms to test metal deployment.
Of course it can be done under ZTP ci umbrella, but the expectation on developing the CI part is on the owner of the enhancement.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,170 @@
...
+Note that the scope of this epic is only enablement of the Metal3
+BaremetalHost API and associated controller/services - the edge ZTP flow
+currently interacts directly via this API without any Machine API integration.
+
+## Proposal
+
+BMO (baremetal-operator) provides the Baremetal Host API, it in turn is configured
+and managed by CBO (cluster-baremetal-operator).
+
+CBO reads the Provisioning CR that is created by the installer on baremetal platforms
```
romfreiman, line 73: Can you please elaborate how will it look in SNO case (none platform). Who will create the provisioning CR? What will change in the current behavior?

asalkeld: On baremetal + AI:
assisted-service (in internal/ignition/ignition.go) generates the ignition - part of that is the Provisioning CR
If it is not on baremetal then that will need to be created by what ever system is driving the flow (as the installer will not create it). I'd assume Assisted Installer or Hive?

romfreiman: hmm not sure. if it's a bare metal, then there is no problem already.
for ztp - not sure AI is the place. Who's currently configures the bmo to watch all namesapces? I assume it's not AI but part of the installation guide.  
oc patch provisioning provisioning-configuration --type merge -p '{"spec":{"watchAllNamespaces": true}}'
So i guess it's also up to the user / ztp scritps?

enhancements/baremetal/enable-baremetal-on-other-platforms.md
asalkeld: I understand, however this seems to me to be one small part of a much bigger solution that will get functionally tested - which is why I put that statement in.. 
So what system is creating the Provisioning CR? that might be a good starting point to figuring out where the e2e belongs.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,170 @@
...
+and create support burden. This is why we have not suggested enabling CBO on
+all platforms and with full feature set. However it is still a potential issue.
+
+## Design Details
+
+### Open Questions [optional]
+
+Re ZTP requirements:
+1. Are these all the platforms that are required?
+2. Is just enabling spec.provisioningNetwork=Disabled mode sufficient?
```
asalkeld, line 114: 3. what will create the Provisioning CR?
4. where will the feature be e2e tested?

enhancements/baremetal/enable-baremetal-on-other-platforms.md
dhellmann: That won't help in a case where a customer has created a Provisioning resource before upgrading to the version of CBO that has the validation, right?

enhancements/baremetal/enable-baremetal-on-other-platforms.md
dhellmann: Yes, I think we can say it's part of the installation instructions but that it doesn't have to be automated.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
dhellmann: I don't see a need for e2e testing of metal on every platform, as long as the test uses a configuration that would be compatible with what's being made available on platforms other than baremetal. I think that means configuring the test environment to not use a provisioning network, right?

enhancements/baremetal/enable-baremetal-on-other-platforms.md
asalkeld: Yes, correct. I'll have to remove this check here https://github.com/openshift/cluster-baremetal-operator/blob/master/controllers/provisioning_controller.go#L257-L270 so that that validation always runs in the Reconcile.
Kinda a corner case..

enhancements/baremetal/enable-baremetal-on-other-platforms.md
asalkeld: thanks, I'll add a note to this effect.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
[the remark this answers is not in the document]
hardys: I think it's covered in https://github.com/openshift/enhancements/blob/master/enhancements/installer/agent-based-installation-in-hive.md and also described in https://github.com/openshift/assisted-service/tree/master/docs/hive-integration

I'm not sure if we've used the ZTP acronym elsewhere in those docs though, so it might be more consistent to just refer to running "hub" clusters for multi-cluster management on other platforms (ZTP is only one possible use-case for that design)

enhancements/baremetal/enable-baremetal-on-other-platforms.md
hardys: Not supporting a provisioning network certainly simplifies things and reduces the complexity of testing on additional platforms, but we probably do still need test coverage *somewhere* of every platform we expect to support (even if it's just downstream QE validation of the ZTP flow).

enhancements/baremetal/enable-baremetal-on-other-platforms.md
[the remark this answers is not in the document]
asalkeld: thanks @hardys , I'll update with the links and s/ZTP/centralized host management/

enhancements/baremetal/enable-baremetal-on-other-platforms.md
asalkeld: thanks all, I have updated this now.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,186 @@
...
+authors:
+  - "@asalkeld"
+  - "@sadasu"
+reviewers:
+  - TBD
+  - "@hardy"
+  - "@romfreiman"
+  - "@dhellman"
+approvers:
+  - TBD
```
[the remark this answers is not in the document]
asalkeld: @dhellmann  any suggestions? Who would normally choose the approvers?

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,186 @@
...
+See:
+- https://github.com/openshift/enhancements/blob/master/enhancements/installer/agent-based-installation-in-hive.md
+- https://github.com/openshift/assisted-service/tree/master/docs/hive-integration
+
+### Goals
+
+The specific goals of this proposal are to:
+
+Support the centralized host management use case by partially enabling Baremetal Host API
+on the following platforms:
```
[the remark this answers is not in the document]
asalkeld: AFAIK it would be the same as hub-spoke all on Metal3. CBO will look at the platform purely for validating the Provisioning CR.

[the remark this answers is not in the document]
dhellmann: Normally we run the metal3 pod on the control plane nodes. If the control plane is external, there won't be any control plane nodes. Does that mean metal3 won't schedule, or will it move to worker nodes? If it moves, will the image downloader work in the same way?

enhancements/baremetal/enable-baremetal-on-other-platforms.md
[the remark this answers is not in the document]
sadasu: Doug Hellmann, Steve Hardy, Zane Bitter, Tony Segura Puimedon, Rom Freiman & myself.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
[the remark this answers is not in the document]
sadasu: We have not considered the use case where the control plane is external and metal3 would have to run on worker nodes in this proposal. Currently, metal3 would only run on master nodes. We have limited understanding of this usecase and would like to understand how CBO can distinguish scenarios where metal3 should run on master vs worker nodes. There was also a recent discussion around infrastructure nodes that could be related.

sadasu:
@asalkeld probably worth mentioning that the when the Provisioning CR is set to Disabled mode, workers nodes would be booted via virtual media. This removes the requirement for the Provisioning Network which can be expected to be available only in Baremetal platform types.
Also, when we are restricting metal3 functionality by restricting what is supported in the Provisioning CR, do we also need platform checks. We could enable CBO to run on all platforms and not just None, OpenStack and Baremetal.

We should also mention that in platforms other than Baremetal, centralized machine management (via MAO) would not be available so users would not be able to scale up Baremetal Machinesets. With the cluster-api based machine management expected to come later, we would be able to provide that (and leave a placeholder for the enhancement proposal for that.)

Nit: We can remove "MPINSTALL-7" for the title.

asalkeld:
> @asalkeld probably worth mentioning that the when the Provisioning CR is set to Disabled mode, workers nodes would be booted via virtual media. This removes the requirement for the Provisioning Network which can be expected to be available only in Baremetal platform types.

Sure, I'll add that.

> Also, when we are restricting metal3 functionality by restricting what is supported in the Provisioning CR, do we also need platform checks. We could enable CBO to run on all platforms and not just None, OpenStack and Baremetal.

I think the point is to only enable what we need and nothing else.

> 
> We should also mention that in platforms other than Baremetal, centralized machine management (via MAO) would not be available so users would not be able to scale up Baremetal Machinesets. With the cluster-api based machine management expected to come later, we would be able to provide that (and leave a placeholder for the enhancement proposal for that.)

See the Non-Goals section, I have it there.

> 
> Nit: We can remove "[MPINSTALL-7](https://issues.redhat.com/browse/MPINSTALL-7)" for the title.

:shrug: it was suggested in the template..

enhancements/baremetal/enable-baremetal-on-other-platforms.md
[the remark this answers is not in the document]
dhellmann: If we're worried about avoiding support cases from enabling this feature on the wrong types of clusters, we should make sure that we're covering all of the cases. I don't know the platform type of a hypershift cluster, and I don't see an obvious hypershift type in https://github.com/openshift/api/blob/a6156965faae5ce117e3cd3735981a3fc0e27e27/config/v1/types_infrastructure.go#L193 so I wonder if the type is None?

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,191 @@
...
+This proposal is to allow CBO to be enabled on the Platforms: None, OpenStack or vSphere.
+
+Further (to restrict the testing matrix) the allowed configuration options
+of the Provisioning CR will be restricted to exactly those required by centralized host management.
+
+*Only spec.provisioningNetwork=Disabled mode will be accepted in the Provisioning CR.*
+
+If any other provisioningNetwork mode is set, the CBO webhook will refuse the change
+in the usual case, but if defined before upgrading the operator, the Reconcile loop
+must always validate the Provisioning CR.
```
dhellmann, line 93: What does the controller do if there's bad data found after an upgrade?

enhancements/baremetal/enable-baremetal-on-other-platforms.md
dhellmann: I think this thread can be resolved.

dhellmann (commented):
> what is the overhead of the baremetal operator on these additional platforms when baremetalhost is not used?

We still need an answer to Derek's question. I think there's no additional overhead when the baremetal-operator is not enabled, because cluster-baremetal-operator always runs (see the question inline about cluster profiles, though). So, we would expect users to accept whatever overhead is introduced by enabling the service, but perhaps we need to give them some idea of what that overhead is.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
asalkeld: it sets a clusterCondition[InvalidConfiguration] = true

asalkeld:
> > what is the overhead of the baremetal operator on these additional platforms when baremetalhost is not used?
> 
> We still need an answer to Derek's question. I think there's no additional overhead when the baremetal-operator is not enabled, because cluster-baremetal-operator always runs (see the question inline about cluster profiles, though). So, we would expect users to accept whatever overhead is introduced by enabling the service, but perhaps we need to give them some idea of what that overhead is.

@derekwaynecarr if cbo gets asked to create bmo, but there are no BMHost CRs then I assume it's just the memory overhead.
On my metal cluster (that does have BMH CRs), this is the memory used by all containers deployed by cbo
done with
```
kubectl exec pod/+podName -c  containerName -- cat /sys/fs/cgroup/memory/memory.usage_in_bytes
```
imagecache daemonset
pod/metal3-image-cache-ktn5z -c metal3-httpd: memory 62320640
pod/metal3-image-cache-bflhj -c metal3-httpd: memory 46858240
pod/metal3-image-cache-p5grw -c metal3-httpd: memory 54792192

bmo deployment
pod/metal3-7bd97c69c7-l2dm4 -c metal3-httpd: memory 46096384
pod/metal3-7bd97c69c7-l2dm4 -c metal3-static-ip-manager: memory 2293760
pod/metal3-7bd97c69c7-l2dm4 -c metal3-mariadb: memory 92688384
pod/metal3-7bd97c69c7-l2dm4 -c metal3-ironic-conductor: memory 123207680
pod/metal3-7bd97c69c7-l2dm4 -c metal3-ironic-api: memory 411484160
pod/metal3-7bd97c69c7-l2dm4 -c metal3-baremetal-operator: (doesn't have "cat")
pod/metal3-7bd97c69c7-l2dm4 -c metal3-dnsmasq: memory 1994752
pod/metal3-7bd97c69c7-l2dm4 -c metal3-ironic-inspector: memory 88412160
pod/metal3-7bd97c69c7-l2dm4 -c ironic-inspector-ramdisk-logs: memory 1658880
pod/metal3-7bd97c69c7-l2dm4 -c ironic-deploy-ramdisk-logs: memory 1921024

enhancements/baremetal/enable-baremetal-on-other-platforms.md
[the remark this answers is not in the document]
asalkeld: I think hypershift's control-plane is on aws and workers are on "other" - i think normally aws. If they want to be able to deploy metal workers, then we will need to enable on aws? - I suspect that is a while off tho'..
the other general case is the dev-preview of CAPI, but I think the current proposal is a good start at opening up the scope of metal3. WDYT?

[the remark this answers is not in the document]
mhrivnak: hypershift development has been AWS-centric, but it does not require AWS.

This proposal is for a use case where a management cluster provisions bare metal hosts and joins them to separate managed clusters. I don't think anyone is (yet?) looking to do that with a management cluster whose control plane is external. Unless/until there are ACM stories being prioritized to run ACM on a hypershift control plane, I think it's out of scope to run metal3 components in a hypershift control plane.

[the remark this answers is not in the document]
csrwng: > For example, would we expect to run the BMO in a cluster with an external control plane?

I don't see why not. A cluster with an external control plane is just like any other cluster. If the purpose of deploying BMO is to manage baremetal hosts and join them to *other* clusters, I think that is a perfectly valid use case. As a general principle though in hypershift clusters, we don't manage the cluster's nodes within the same cluster. They are managed by the management cluster that is hosting its control plane. There is no reason though for a hypershift cluster not to be a management cluster itself.

> I don't know the platform type of a hypershift cluster, and I don't see an obvious hypershift type in ...

Hypershift is not a platform, and so clusters with platforms of type AWS, None, etc can be hypershift clusters. It's more about the control plane topology, for which we have the `External` value.

> Normally we run the metal3 pod on the control plane nodes. If the control plane is external, there won't be any control plane nodes. Does that mean metal3 won't schedule, or will it move to worker nodes?

Yes, it would have to move to worker nodes if you want to run it on Hypershift clusters. Currently it would "just work" because we are still labeling worker nodes as both masters and workers, but we will get rid of that with 4.9+ clusters.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,197 @@
...
+title: enable-baremetal-on-other-platforms
+authors:
+  - "@asalkeld"
+  - "@sadasu"
+reviewers:
+  - "@hardy"
+  - "@romfreiman"
+  - "@dhellman"
+approvers:
+  - "@hardy"
```
hardys, line 11: Should be `hardys` here and above

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,197 @@
...
+creation-date: 2021-08-20
+last-updated: 2021-08-20
+status: implementable
+see-also:
+  - "/enhancements/baremetal-provisiong-config.md"
```
hardys, line 16: should be `/enhancements/baremetal/baremetal-provisiong-config.md`

hardys: Or maybe this can be a relative path?  Either way I think the current link is broken :)

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,197 @@
...
+
+Centralized host management can deploy clusters when running on the above platforms.
+
+### Non-Goals
+
+Allow Baremetal Host API to be fully enabled on all platforms.
+
+Note that the scope of this epic is only enablement of the Metal3
+BaremetalHost API and associated controller/services - the centralized host management flow
+currently interacts directly via this API without any Machine API integration.
```
hardys, line 69: We could be even more explicit here to avoid confusion e.g something like "... without any Machine API integration.  This means further work will be required via a future enhancement to enable the single cluster case where a combination of e.g VM controlplane and Baremetal workers is desired"

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,197 @@
...
+We will be successful when:
+
+Centralized host management can deploy clusters when running on the above platforms.
+
+### Non-Goals
+
+Allow Baremetal Host API to be fully enabled on all platforms.
+
+Note that the scope of this epic is only enablement of the Metal3
+BaremetalHost API and associated controller/services - the centralized host management flow
```
sadasu, line 68: Can this be made more specific like "enablement of the Metal3
BaremetalHost API and associated controller/services on a subset of OpenShit platforms"? 
An observation, this statement seems to be specifying a goal in the non-goals section.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,197 @@
...
+
+Note:
+
+1. when the Provisioning CR is set to provisioningNetwork=Disabled mode, worker
+nodes would be booted via virtual media. This removes the requirement for the
+Provisioning Network which can be expected to be available only in Baremetal platform types.
+
+2. documentation will need to be added to the centralized host management documentation
+explaining how to create a Provisioning CR. Current documentation is here:
+https://github.com/openshift/assisted-service/blob/8880093ef5ce041d4c1951ffd5ea1096991ec3ee/docs/user-guide/assisted-service-on-openshift.md#configure-bare-metal-operator
```
sadasu, line 110: This link updates an existing Provisioning CR to start watching all Namespaces. When running on Platforms, OpenStack, vSphere and None, a completely new Provisioning CR needs to be created. Documentation needs to be updated with information from https://github.com/openshift/cluster-baremetal-operator/blob/master/README.md to define a valid Provisioning CR for these platforms.

sadasu: On these platforms, will CBO be able to generate a Provisioning CR with some valid defaults so that it can be updated later by the assisted-service?  This could be handled as an implementation detail that can be worked out later.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,197 @@
...
+There is concern that *random* customers will use this feature out of context
+and create support burden.
+
+## Alternatives
+
+Customers can instead create a dedicated baremetal cluster to use as the hub
+cluster.
+
+Another alternative is to additionally distribute baremetal-operator as an optional
+operator with OLM. It's probably not a good idea, but worth enumerating as an alternative.
```
sadasu, line 194: "It's probably not a good idea, but worth enumerating as an alternative." can be removed.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
asalkeld: cool, i'll add that

enhancements/baremetal/enable-baremetal-on-other-platforms.md
asalkeld: thanks, i'll try to reword it.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,198 @@
...
+
+If any other provisioningNetwork mode is set, the CBO webhook will refuse the change
+in the usual case, but if defined before upgrading the operator, the Reconcile loop
+must always validate the Provisioning CR.
+
+Note:
+
+1. when the Provisioning CR is set to provisioningNetwork=Disabled mode, worker
+nodes would be booted via virtual media. This removes the requirement for the
+Provisioning Network which can be expected to be available only in Baremetal platform types.
```
flaper87, line 109: IIRC, the `ProvisioningOSDownloadURL`, is required. Should/Will it till be required when running on other platforms? 

From the [field's docstring](https://github.com/openshift/cluster-baremetal-operator/blob/cae7b845e8eae10c12e4540e0d83c2bc972d8bb5/api/v1alpha1/provisioning_types.go#L124-L126):

> 	// ProvisioningOSDownloadURL is the location from which the OS
>	// Image used to boot baremetal host machines can be downloaded
>	// by the metal3 cluster.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,198 @@
...
+#### Unit Testing
+
+We will add unit tests to confirm that cluster-baremetal-operator:
+* is enabled on the required platforms.
+* will restrict functionality on these platforms to ProvisioningNetork=Disabled.
+
+#### Functional Testing
+
+An e2e test will be written in the Assisted Installer CI that will:
+1. create one of the platforms above (SNO Platform=None might be the easiest) with Assisted Service.
```
flaper87, line 150: +1 

I have deployed this exact same scenario (with a custom-built 4.8 image and the original PR that sparked this enhancement).

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,198 @@
...
+2. confirm that CBO is enabled
+3. create a Provisioning CR and confirm that BMO is running
+4. provision a baremetal cluster
```
flaper87, line 153: No need to wait for the full cluster to be deployed. Waiting for the agent's discovery phase to be over should be enough proof. This is assuming assisted will be used for this test.

Regardless, enabling CBO and BMO in an SNO node is a good, easy-enough, test

flaper87 (commented):
I like this proposal and I've needed this myself recently as part of an investigation I'm doing for a valid customer use case.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
flaper87: Hold on, my comment is about making `ProvisioningOSDownloadURL` not-required as well, not about `provisioningNetwork` :stuck_out_tongue_closed_eyes:

flaper87: > The ZTP flow however is different/special, in that it only currently deploys ISO images generated by assisted-service, thus it may well be possible to omit ProvisioningOSDownloadURL in that case, provided CBO has sufficient conditionals to just not deploy the download/cache containers in that case?

Chatted with @asalkeld yesterday and it looks like it may be possible to remove the validation that makes this data required. So far, we've been using whatever we find there or whatever matches the version of the cluster where `metal3` is running. However, I think it'd be better if we could not fill in this field with semi-random data.

Thanks for the comment and confirming. I'm unsure on whether there are enough validations in CBO to prevent deployments when this data is not provided but it would be great if there were.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Goals
+
+The specific goals of this proposal are to:
+
+Support the centralized host management use case by partially enabling Baremetal Host API
+on the following platforms:
+- None
+- OpenStack
+- vSphere
```
flaper87, line 57: Could you elaborate on why cloud platforms like AWS, GCP, Azure are left out of this enhancement? 

The motivation is not obvious to me, it may not be obvious to others. 

Maybe it's just to limit the scope for now? 

Noticed this since I had someone asking if it would be possible to enable this use-case on AWS

hardys: @flaper87 the intention was to limit the scope (and test matrix) to on-premise platforms for now - AFAIK the requirements driving this are only for on-premise atm, but it could be expanded in future should the need arise.

Also if we decide to expand the scope of this to enable PXE deployments in future (e.g Provisioning network not `Disabled` ) I could see that being potentially more difficult with provisioning services running on a cloud, vs in the same environment as the target hosts.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,198 @@
...
+As a user of a hub cluster that performs central infrastructure management, and
+optionally zero-touch provisioning, I need to provision hosts using the k8s-native
+API (Baremetal Hosts CR) even when the hub cluster has a platform of None, OpenStack, or vSphere.
+
+
+### Risks and Mitigations
+
+There is concern that *random* customers will use this feature out of context
+and create support burden. This is why we have not suggested enabling CBO on
+all platforms and with full feature set. However it is still a potential issue.
```
hardys, line 131: One mitigation for this we've discussed is to avoid documenting this outside of the CIM/ZTP case.  For that reason this change won't be documented as a standalone feature, only in the context of CIM/ZTP.

enhancements/baremetal/enable-baremetal-on-other-platforms.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Design Details
+
+### Open Questions [optional]
+
+1. Where will the feature be e2e tested?
```
hardys, line 137: It seems this is now answered, in the Functional testing section below?

hardys:
/approve

Overall this looks good to me, couple of minor comments added if you do any further updates, thanks!

asalkeld:
> /approve
> 
> Overall this looks good to me, couple of minor comments added if you do any further updates, thanks!

thanks @hardys , I have updated with the tweaks..

PR #876, opened 2021-08-23
ingress: Add mutable-publishing-scope enhancement

Miciah:
This enhancement defines an approach for allowing users to modify the scope of a service load-balancer for an IngressController that uses the LoadBalancerService endpoint publishing strategy type.

### DIFF
--- a/enhancements/ingress/mutable-publishing-scope.md
+++ b/enhancements/ingress/mutable-publishing-scope.md
@@ -1,0 +1,410 @@
+---
+title: mutable-publishing-scope
+authors:
+  - "@Miciah"
+reviewers:
+  - "@candita"
+  - "@frobware"
+  - "@knobunc"
+approvers:
+  - "@frobware"
+  - "@knobunc"
+creation-date: 2021-08-23
+last-updated: 2021-08-23
+status: implementable
+see-also:
+replaces:
+superseded-by:
+---
+
+# Ingress Mutable Publishing Scope
+
+## Release Signoff Checklist
+
+- [X] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement defines an approach for allowing users to modify the scope of a
+service load-balancer for an IngressController that uses the LoadBalancerService
+endpoint publishing strategy type.
+
+## Motivation
+
+Users sometimes wish to change the scope of the service load-balancer (SLB) of
+an IngressController that uses the "LoadBalancerService" endpoint publishing
+strategy type after the IngressController has been created.  Without this
+enhancement, it is only possible to change the scope by deleting and recreating
+the IngressController, which is disruptive and incompatible with some workflows.
+This enhancement makes it possible to change the scope by updating the
+IngressController's spec.  Then, on platforms that support changing the SLB's
+scope without deleting and recreating it, the operator performs this update.
+For platforms that do not support changing an SLB's scope _in situ_, this
+enhancement guides the user through the process of deleting and recreating the
+SLB.  When completing the process requires user action that would be disruptive,
+the operator clearly communicates this fact and provides instructions for
+canceling the operation if the user wishes to back out of the operation.  This
+enhancement applies to the default IngressController as well as custom
+IngressControllers.
+
+### Goals
+
+1. The cluster administrator can change the scope of the SLB of an IngressController with "LoadBalancerService" endpoint publishing strategy type.
+2. If the cloud platform supports mutating scope on an SLB, the operator will perform this operation when the cluster administrator changes the configured scope.
+3. If the cloud platform does not support mutating scope, the operator will advise the user how to complete the operation.
+
+### Non-Goals
+
+1. The operator does not delete an SLB for an IngressController that is not marked for deletion.
+
+## Proposal
+
+Some platforms (such as Azure and GCP) support changing the scope of a service
+load-balancer between internal and external without deleting and recreating the
+load balancer, by setting [a cloud-provider-specific
+annotation](https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer)
+on the Kubernetes Service object.  On these platforms, the operator merely sets
+the annotation to the desired scope, and Kubernetes's service controller and
+cloud-provider implementation complete the operation of changing the load
+balancer's scope.
+
+Other platforms (such as AWS) require deleting and recreating a load balancer to
+change its scope.  This operation is disruptive: It interrupts ingress traffic
+and may cause the load balancer's address to change.  On these platforms, the
+operator signals that the user must delete the Kubernetes Service object.  Once
+the user performs this step, the operator recreates the load balancer with the
+desired scope to complete the operation.
+
+In order to signal that the user must take action, the operator sets the
+`Progressing` status condition to `True` with a message that indicates the
+action that the user must take:
+
+```yaml
+apiVersion: operator.openshift.io/v1
+kind: IngressController
+metadata:
+  name: custom
+  namespace: openshift-ingress-operator
+spec:
+  replicas: 2
+  domain: mycluster.com
+  endpointPublishingStrategy:
+    type: LoadBalancerService
+    loadBalancer:
+      scope: Internal
+status:
+  availableReplicas: 2
+  conditions:
+  - message: Deployment has minimum availability.
+    reason: MinimumReplicasAvailable
+    status: "True"
+    type: Available
+  - message: "The IngressController scope was changed from \"External\" to \"Internal\".  To effectuate this change, you must delete the service: `oc -n openshift-ingress delete svc/router-default`; the service load-balancer will then be deprovisioned and a new one created.  Alternatively, you can revert the change to the IngressController: `oc -n openshift-ingress-operator patch ingresscontrolle\
+      rs/default --type=merge --patch='{\"spec\":{\"endpointPublishingStrategy\":{\"type\":\"LoadBalancerService\",\"loadBalancer\":{\"scope\":\"External\"}}}}'"
+    reason: ScopeChanged
+    status: "True"
+    type: Progressing
+```
+
+Note that the IngressController remains available, and that the user can revert
+the change by setting `spec.endpointPublishingStrategy.loadBalancer.scope` back
+to its previous value.  That means the user must take one of two actions:
+
+1. Delete the Service referenced in the status condition.
+2. Revert the change to the IngressController.
+
+If the user deletes the Service, the operator recreates it with the desired
+scope.  Alternatively, if the user reverts the change, the `Progressing` status
+condition changes back to `False`, and the state of the IngressController is as
+it was before the user first changed the value of
+`spec.endpointPublishingStrategy.loadBalancer.scope`.
+
+In addition to deleting the Service explicitly, it is possible to annotate the
+IngressController with the newly defined
+`ingress.operator.openshift.io/auto-delete-load-balancer` annotation.  If the
+operator observes that this annotation is set and that its
+`spec.endpointPublishingStrategy.loadBalancer.scope` field has been changed, the
+operator automatically deletes the Service if needed to complete a scope-change
+operation.  This purpose of this annotation is to simplify automation: A tool
+can set the annotation when it creates the IngressController or when it updates
+`spec.endpointPublishingStrategy.loadBalancer.scope` to instruct the operator to
+perform the operation automatically.  This annotation is not intended for
+end-users to use directly.
+
+
+### Validation
+
+This enhancement adds no new API fields or values and requires no new API
+validation.
+
+### User Stories
+
+#### As a cluster administrator, I want my default IngressController to be private (internal)
+
+If the user installs a new cluster without specifying that the cluster is
+"private", the default IngressController is created with "External" scope.  This
+enhancement enables the cluster administrator to change the scope to "Internal"
+as follows:
+
+```shell
+oc -n openshift-ingress-operator patch ingresscontrollers/default --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"type":"LoadBalancerService","loadBalancer":{"scope":"Internal"}}}}'
+```
+
+The cluster administrator can check the status of the IngressController as
+usual:
+
+```shell
+oc -n openshift-ingress-operator get ingresscontrollers/default -o yaml
+```
+
+The `Progressing` status condition will indicate whether the cluster
+administrator must take further action.  The status condition may indicate that
+the cluster administrator needs to delete the Service, like so:
+
+```shell
+oc -n openshift-ingress delete services/router-default
+```
+
+If the cluster administrator deletes the Service, the operator then recreates it
+with internal scope.
+
+
+#### As a cluster administrator, I want to make a private (internal) IngressController public (external)
+
+The cluster administrator can change the scope of an IngressController to be
+external (public) as follows:
+
+```shell
+oc -n openshift-ingress-operator patch ingresscontrollers/private --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"type":"LoadBalancerService","loadBalancer":{"scope":"External"}}}}'
+```
+
+Again, the cluster administrator can check the IngressController's status
+conditions and may need to delete a Service to complete the change in scope.
+
+#### As a cluster administrator, I want to cancel a change to scope before it is completed
+
+Suppose the cluster administrator has changed the scope of an IngressController
+from "External" to "Internal", as follows:
+
+```shell
+oc -n openshift-ingress-operator patch ingresscontrollers/private --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"type":"LoadBalancerService","loadBalancer":{"scope":"Internal"}}}}'
+```
+
+Now the cluster administrator realizes that the scope should not be changed at
+this time, for example due to the disruption that changing the scope would
+cause.  The cluster administrator can cancel the change by changing
+`spec.endpointPublishingStrategy.loadBalancer.scope` back to its original value:
+
+```shell
+oc -n openshift-ingress-operator patch ingresscontrollers/private --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"type":"LoadBalancerService","loadBalancer":{"scope":"External"}}}}'
+```
+
+#### As the provider of a managed service, I want to automate changing an IngressController's scope
+
+A managed service can initiate the operation and tell the operator to complete
+it automatically by annotating the IngressController and changing the scope,
+using an API call that is equivalent to the following commands:
+
+```shell
+oc -n openshift-ingress-operator patch ingresscontrollers/default --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"type":"LoadBalancerService","loadBalancer":{"scope":"Internal"}}}}'
+oc -n openshift-ingress-operator annotate ingresscontrollers/default ingress.operator.openshift.io/auto-delete-load-balancer=
+```
+
+The operator will begin and complete the scope-change operation, including
+deleting and recreating the service load-balancer if needed.
+
+### API Extensions
+
+This enhancement makes the following API changes:
+
+- The semantics of the the IngressController API's `spec.endpointPublishingStrategy.loadBalancer.scope` field changes to allow mutation.
+- A `Progressing` status condition is added to the set of status conditions reported in the IngressController API's `status.conditions` field.
+
+### Implementation Details
+
+The operator's ingress controller reconciles IngressController objects.  When an
+IngressController is created, the controller sets the value of its
+`status.endpointPublishingStrategy.loadBalancer.scope` field to the value of the
+IngressController's `spec.endpointPublishingStrategy.loadBalancer.scope` field.
+With this enhancement, the controller checks whether these two fields differ,
+which would indicate that the user had changed the value of
+`spec.endpointPublishingStrategy.loadBalancer.scope`.  If the controller
+observes a modification to the spec field, the controller updates the status
+field.
+
+Additionally, if the operator is running on Azure or GCP, with this enhancement,
+the controller updates [the cloud-provider-specific
+annotation](https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer)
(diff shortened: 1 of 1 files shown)


Miciah:
Last push abbreviates the status condition message in the example ingresscontroller to satisfy the lint check.

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+
+1. The operator does not delete a service load-balancer for an IngressController that is not marked for deletion.
+
+## Proposal
+
+Some platforms (such as Azure and GCP) support changing the scope of a service
+load-balancer between internal and external without deleting and recreating the
+load balancer, by setting an annotation on the Kubernetes Service object.  On
+these platforms, the operator merely sets the annotation to the desired scope,
+and the operation of changing it scope is complete.
```
candita, line 61:
```suggestion
and the operation of changing its scope is complete.
```

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+    loadBalancer:
+      scope: Internal
+status:
+  availableReplicas: 2
+  conditions:
+  - message: Deployment has minimum availability.
+    reason: MinimumReplicasAvailable
+    status: "True"
+    type: Available
+  - message: Have load balancer with scope "External", want load balancer with scope "Internal".  You can delete the openshift-ingress/router-default service to proceed [...].  Alternatively, you can change the IngressController's spec.endpointPublishingStrategy.loadBalancer.scope field value back to its previous value [...].
```
candita, line 94: The message is really long and probably too much information.  Could we get away with something like: "Scope changed from ___ to ___, requires delete of service ___"

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+
+If the user deletes the Service, the operator recreates it with the desired
+scope.  Alternatively, if the user reverts the change, the `Progressing` status
+condition changes back to `False`, and the state of the IngressController is as
+it was before the user first changed the value of
+`spec.endpointPublishingStrategy.loadBalancer.scope`.
+
+In addition to deleting the IngressController explicitly, it is possible to
+annotate it with the newly defined
+`ingress.operator.openshift.io/auto-delete-load-balancer` annotation.  If the
```
candita, line 115: Is this different than the annotation mentioend earlier in Line 59?  If not, can you add the name of the annotation the first place it appears?

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+1. Delete the Service referenced in the status condition.
+2. Revert the change to the IngressController.
+
+If the user deletes the Service, the operator recreates it with the desired
+scope.  Alternatively, if the user reverts the change, the `Progressing` status
+condition changes back to `False`, and the state of the IngressController is as
+it was before the user first changed the value of
+`spec.endpointPublishingStrategy.loadBalancer.scope`.
+
+In addition to deleting the IngressController explicitly, it is possible to
```
candita, line 113:
```suggestion
In addition to deleting the Service explicitly, it is possible to
```

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+annotate it with the newly defined
+`ingress.operator.openshift.io/auto-delete-load-balancer` annotation.  If the
+operator observes that this annotation is set and that its
+`spec.endpointPublishingStrategy.loadBalancer.scope` field has been changed, the
+operator automatically deletes the Service if needed to complete a scope-change
+operation.  This purpose of this annotation is to simplify automation: A tool
```
candita, line 119: The annotation is sort of awkward. Seems like it would be better specifiied as a flag on the `oc edit` operation, something like `oc edit ... --auto-update` or `--propagate` or similar.

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+can simultaneously update `spec.endpointPublishingStrategy.loadBalancer.scope`
+and set the annotation to instruct the operator to perform the operation
+automatically.  This annotation is not intended for end-users to use directly.
```
candita, line 122: If it's not intended for end-users, then maybe it should not be an annotation?

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+internal (private) as follows:
+
+```shell
+oc -n openshift-ingress-operator patch ingresscontrollers/private --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"type":"LoadBalancerService","loadBalancer":{"scope":"Internal"}}}}'
+```
+
+Again, the cluster administrator can check the IngressController's status
+conditions and may need to delete a Service to complete the change in scope.
+
+#### As a cluster administrator, I want to cancel a change to scope before it is completed
```
candita, line 174: I think this use case will be of little value, and though it is easy to implement and describe, it somewhat muddies an otherwise straightforward documentation.

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+
+#### As the provider of a managed service, I want to automate changing an IngressController's scope
+
+A managed service can initiate the operation and tell the operator to complete
+it automatically by annotating the IngressController and changing the scope,
+using an API call that is equivalent to the following commands:
+
+```shell
+oc -n openshift-ingress-operator patch ingresscontrollers/default --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"type":"LoadBalancerService","loadBalancer":{"scope":"Internal"}}}}'
+oc -n openshift-ingress-operator annotate ingresscontrollers/default ingress.operator.openshift.io/auto-delete-load-balancer=
```
candita, line 200:
```suggestion
oc -n openshift-ingress-operator annotate ingresscontrollers/default ingress.operator.openshift.io/auto-delete-load-balancer=true
```
Or some value

candita: Would you need to add the annotation first, before the change is made?

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+`status.endpointPublishingStrategy.loadBalancer.scope` field to the value of the
+IngressController's `spec.endpointPublishingStrategy.loadBalancer.scope` field.
+With this enhancement, the controller checks whether these two fields differ,
+which would indicate that the user had changed the value of
+`spec.endpointPublishingStrategy.loadBalancer.scope`.  If the controller
+observes a modification to the spec field, the controller updates the status
+field.
+
+Additionally, if the operator is running on Azure or GCP, with this enhancement,
+the controller updates the annotations on the Kubernetes Service object for the
```
candita, line 220: Which annotations, just some pre-existing ones?

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+IngressController.  Kubernetes's service controller and cloud provider
+implementation perform the necessary changes to change the service
+load-balancer's scope, and no further action is needed on the part of the user
+or the operator.
+
+If the operator is running on AWS or IBM Cloud, the controller does not update
```
candita, line 226: Is this the default case?  E.g. for other providers not listed, does it do this action or the previous?

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+the scope that is specified in the IngressController object's status, whereupon
+it sets the IngressController's `Progressing` status condition to `True` with a
+message indicating that the cluster administrator must either revert the change
+to the IngressController or delete the Service.
+
+If the operator observes that the Service does not exist (which would be the
+case once the user deletes the Service), the operator recreates the service with
+the scope indicated in `status.endpointPublishingStrategy.loadBalancer.scope`.
+
+Crucially, by default, the operator *never* deletes the Service as long as the
```
candita, line 238:
```suggestion
Crucially, by default, the operator ***never*** deletes the Service as long as the
```

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+IngressController is not deleted; the operator requires that the user perform
+the disruptive action of deleting the Service in order to complete the
+operation.  The only exception is if the
+`ingress.operator.openshift.io/auto-delete-load-balancer` annotation is set on
+the IngressController, in which case the operator *does* delete the Service.
```
candita, line 243:
```suggestion
the IngressController, in which case the operator ***does*** delete the Service.
```

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+
+If the user modifies `spec.endpointPublishingStrategy.loadBalancer.scope` on a
+version of OpenShift without this enhancement and upgrades to a version of
+OpenShift with this enhancement, the operator annotates the Service or sets
+`Progressing=True` on the IngressController as appropriate.  Thus the operator
+may effectuate a latent scope change, but it does not delete the Service.
+
+If the user changes the scope and then downgrades to a version without this
+enhancement while the operator is reporting `Progressing=True`, the downgraded
+operator does not take any action (other than possibly setting
```
candita, line 305: It might be cleaner if it were guaranteed to properly update the Progessing status in this case (reason and message).

enhancements/ingress/mutable-publishing-scope.md
Miciah: The detailed wording is the result of some discussion here:  https://github.com/openshift/cluster-ingress-operator/pull/582#discussion_r605151494

The explicit and detailed message is intended to prevent mistakes and reduce support cases.  I can try to make this more concise, but I think it is important to communicate the options and their consequences because otherwise users will break their clusters' ingress.

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,348 @@
...
+    status: "True"
+    type: Progressing
+```
+
+Note that the IngressController remains available, and that the user can revert
+the change by setting `spec.endpointPublishingStrategy.loadBalancer.scope` back
+to its previous value.  That means the user must take one of two actions:
+
+1. Delete the Service referenced in the status condition.
+2. Revert the change to the IngressController.
```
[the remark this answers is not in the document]
Miciah: When we introduced an earlier version of this feature (which we since reverted) that automatically deleted and recreated the service, we had at least one user who changed the scope and then realized it would require deprovisioning and recreating the LB and wanted to revert the change.  The option to back out the change is closely tied to the motivation for this enhancement; I'll revise the motivation section to call this out.

enhancements/ingress/mutable-publishing-scope.md
Miciah: It is a different annotation.  I'll try to clarify that on line 59.

enhancements/ingress/mutable-publishing-scope.md
Miciah: Would the `oc patch` command have custom logic to add the `ingress.operator.openshift.io/auto-delete-load-balancer` annotation when the user provided the `--auto-update` command and an IngressController?  I'd rather avoid adding IngressController-specific or (OpenShift-specific) logic to `oc patch`, and anyway, the annotation is intended for automation such as [the cloud-ingress-operator's publishingstrategy controller](https://github.com/openshift/cloud-ingress-operator/blob/f560b7888c0e140cdc4fa4864929638e9982d877/pkg/controller/publishingstrategy/publishingstrategy_controller.go).

enhancements/ingress/mutable-publishing-scope.md
Miciah: Do you have an alternative in mind?  The annotation was requested by Service Delivery; we need something that [cloud-ingress-operator](https://github.com/openshift/cloud-ingress-operator) could use.

enhancements/ingress/mutable-publishing-scope.md
Miciah: It's important because we've had support cases where users changed the scope on the IngressController and then realized that completing the change would cause the service load-balancer's IP address to change and wanted to back out.

enhancements/ingress/mutable-publishing-scope.md
Miciah: The annotation value doesn't actually matter.

> Would you need to add the annotation first, before the change is made?

No, so automation could either add the annotation when creating the IngressController or add it when updating the scope.  I'll try to clarify this point.

enhancements/ingress/mutable-publishing-scope.md
Miciah: Right; I'll add clarify that point.

enhancements/ingress/mutable-publishing-scope.md
Miciah: I added some language to clarify that the downgraded operator will continue updating status conditions.  Does that make things clearer?

Miciah:
Updated per @candita's comments.

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,362 @@
...
+with internal scope.
+
+
+#### As a cluster administrator, I want to make a private (internal) IngressController public (external)
+
+The cluster administrator can change the scope of an IngressController to be
+internal (private) as follows:
+
+```shell
+oc -n openshift-ingress-operator patch ingresscontrollers/private --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"type":"LoadBalancerService","loadBalancer":{"scope":"Internal"}}}}'
```
brandisher, line 181: I'm not sure if its me or the wording but this story seems a little confusing. I readt his has `have: internal IngressController, want external IngressController`. If that's accurate, shouldn't the statement be: `...change the scope of an IngressController to be external (public) ...` along with an update to `External` in the patch command?

enhancements/ingress/mutable-publishing-scope.md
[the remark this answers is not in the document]
candita: Ok

enhancements/ingress/mutable-publishing-scope.md
candita: ok

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,362 @@
...
+    loadBalancer:
+      scope: Internal
+status:
+  availableReplicas: 2
+  conditions:
+  - message: Deployment has minimum availability.
+    reason: MinimumReplicasAvailable
+    status: "True"
+    type: Available
+  - message: The IngressController scope was changed from "External" to "Internal".  To put this change into effect, you must delete the openshift-ingress/router-default service; the service load-balancer will then be deprovisioned and a new one created.  Alternatively, you can change the IngressController's spec.endpointPublishingStrategy.loadBalancer.scope field value back to its previous value.
```
frobware, line 106: Can we also emit oc commands that you can simply cut & paste? I'm assuming we have enough context (e.g., namespace and names, et al). Might be helpful should you want to restore "its previous value" too.

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,362 @@
...
+```shell
+oc -n openshift-ingress-operator get ingresscontrollers/default -o yaml
+```
+
+The `Progressing` status condition will indicate whether the cluster
+administrator must take further action.  The status condition may indicate that
+the cluster administrator needs to delete the Service, like so:
+
+```shell
+oc -n openshift-ingress delete services/router-default
```
frobware, line 168: This is just stating that if you went from External=>Internal and then issued the delete the operator tracks the the scope you wanted and it will come back as Internal - correct?

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,362 @@
...
+`spec.endpointPublishingStrategy.loadBalancer.scope`.  If the controller
+observes a modification to the spec field, the controller updates the status
+field.
+
+Additionally, if the operator is running on Azure or GCP, with this enhancement,
+the controller updates [the cloud-provider-specific
+annotation](https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer)
+on the Kubernetes Service object for the IngressController.  Kubernetes's
+service controller and cloud-provider implementation perform the necessary
+changes to change the service load-balancer's scope, and no further action is
```
frobware, line 237: If this operation fails where would it be reported? Although the Kubernetes service controller makes the changes I'd like to know where we diagnose failures should that fail in any way.

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,362 @@
...
+The operator has end-to-end tests; for this enhancement, the following test is
+added:
+
+1. Create an IngressController with the "LoadBalancerService" endpoint publishing strategy type and "External" scope.
+2. Verify that the operator creates a Service annotated for external scope.
+3. Set the IngressController's endpoint publishing strategy's scope to "Internal".
+4. If the operator is running on AWS and IBM Cloud, verify that the IngressController reports `Progressing=True`.
+5. If the operator is running on Azure or GCP, verify that the Service is annotated for internal scope.
+6. Set the IngressController's endpoint publishing strategy's scope to "External".
+7. Verify that the IngressController reports `Progressing=False`.
```
frobware, line 290: Is it also feasible to have test setups that make progress but ultimately fail so that `Progressing=False` does not occur? I'm asking for both the happy path and the unhappy path.

enhancements/ingress/mutable-publishing-scope.md
```diff
@@ -0,0 +1,362 @@
...
+
+N/A.  We do not plan to deprecate this feature.
+
+### Upgrade / Downgrade Strategy
+
+If the user modifies `spec.endpointPublishingStrategy.loadBalancer.scope` on a
+version of OpenShift without this enhancement and upgrades to a version of
+OpenShift with this enhancement, the operator annotates the Service or sets
+`Progressing=True` on the IngressController as appropriate.  Thus the operator
+may effectuate a latent scope change, but it does not delete the Service.
```
frobware, line 315: How/Why will it not delete the service in this case?

enhancements/ingress/mutable-publishing-scope.md
Miciah: We can emit oc commands, but the patch command alone makes the message too long to pass the markdownlint CI job.  I'll try using double quotes with a backslash-escaped newline to see whether the linter allows that.  I believe that it'll still be valid yaml, although it won't be exactly what `oc get ... -o yaml` would actually output.

enhancements/ingress/mutable-publishing-scope.md
Miciah: More or less.  The main point that I'm trying to convey is that the user must explicitly perform the delete operation if one is needed, but to your question, yes, the operator will recreate the service with the scope specified in the ingresscontroller.  Is the current phrasing clear, or does it need some wordsmithing?

enhancements/ingress/mutable-publishing-scope.md
Miciah: In case of failure, the service controller should emit an event with reason "SyncLoadBalancerFailed", which the ingress operator would observe and reflect in the ingresscontroller's status conditions as `LoadBalancerReady=False` and ultimately `Available=False`.

enhancements/ingress/mutable-publishing-scope.md
Miciah: `Progressing=True` indicates that the operator is in an intermediate state: the user has specified one scope, and the service currently has a different scope.  Once this discrepancy is resolved (by deleting the service or by updating the ingresscontroller to have the same scope as the service), then the operator returns to its baseline of reporting `Progressing=False`.  For reference, here is the proof-of-concept implementation of the status reporting: https://github.com/openshift/cluster-ingress-operator/commit/7daeb6f85305d8f973c039cfafecddc6da60903f#diff-56b131774a926e7a0e30a9be7dac7bf5c5cec11ff709aa6604cecc9ef117ede2R547-R584.  If the service controller raises an error while updating or recreating the service load-balancer, then it will report an event that the operator will observe and report as described in my previous comment.  Is that sufficient?

enhancements/ingress/mutable-publishing-scope.md
Miciah: The how and why are the same in the upgrade case as in the normal case:  If the ingresscontroller and the service specify different scopes and we don't know that the platform supports scope changes _in situ_, then the operator does not delete the service because doing so could disrupt traffic.  The behavior on upgrade or downgrade follows from and is consistent with the logic described in the proposal and implementation details.  Should I add something to that effect to the enhancement text?  I sought to adhere to the enhancement format without being too verbose or repetitive.

Miciah:
The CI job failed with no obvious reason:
```
 + markdownlint-cli2 '**/*.md'
markdownlint-cli2 v0.3.2 (markdownlint v0.24.0)
Finding: **/*.md
Linting: 343 file(s)
Summary: 0 error(s)
++ dirname hack/markdownlint.sh
+ hack/template-lint.sh
Checking enhancements/ingress/mutable-publishing-scope.md
enhancements/ingress/mutable-publishing-scope.md missing "### API Extensions"
enhancements/ingress/mutable-publishing-scope.md missing "### Operational Aspects of API Extensions"
enhancements/ingress/mutable-publishing-scope.md missing "#### Failure Modes"
enhancements/ingress/mutable-publishing-scope.md missing "#### Support Procedures"
{"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"prow/entrypoint/run.go:80","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"2021-11-17T03:22:45Z"} 
```
/test markdownlint

Miciah:
Wait, those are obvious reasons.

Miciah:
c260fc8 → 51e0f70 adds the newly required "API Extensions", "Operational Aspects of API Extensions", "Failure Modes", and "Support Procedures" sections.

enhancements/ingress/mutable-publishing-scope.md
Miciah: I added this information under the new "Support Procedures" section.

enhancements/ingress/mutable-publishing-scope.md
Miciah: The new "Operational Aspects of API Extensions" section should clarify this.

PR #877, opened 2021-08-24
Minor host port registry cleanups

danwinship:
partially split off from #618 

/assign @squeed


danwinship:
/assign @aravindhp 
since you pointed out the vxlan thing

aravindhp (commented):
Thanks for addressing this, @danwinship 

/lgtm

PR #881, opened 2021-08-31
LOG-1514: LokiStack CR extension for the gateway configuration

sasagarw:
This PR:
- Defines the api changes required for LokiStack CR extension for the gateway configuration.

/cc @periklis 
/assign @igor-karpukhin 

Links:
Jira - https://issues.redhat.com/browse/LOG-1514

### DIFF
--- a/enhancements/cluster-logging/loki-gateway-configuration.md
+++ b/enhancements/cluster-logging/loki-gateway-configuration.md
@@ -1,0 +1,512 @@
+---
+title: loki-lokistack-gateway-configuration
+
+authors:
+  - "@sasagarw"
+
+reviewers:
+  - "@periklis"
+
+approvers:
+  - "@periklis"
+  - "@igor-karpukhin"
+
+creation-date: 2021-08-27
+last-updated: 2021-09-27
+status: implementable
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Loki Operator: LokiStack CR extension for the gateway configuration
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+LokiStack Gateway is a component deployed as part of Loki Operator. It provides secure access to Loki's distributor (i.e. for pushing logs) and query-frontend (i.e. for querying logs) via consulting an OAuth/OIDC endpoint for the request subject.
+
+This proposal provides an overview of the required LokiStack API changes to provide a per-tenant OIDC configuration for the LokiStack gateway component.
+
+_Note:_ We are simply providing an API to reconfigure an existing implementation. Most of this context does not need to be implemented as it is already satisfied by [observatorium-api](https://github.com/observatorium/api).
+
+## Motivation
+
+### Goals
+
+The goals of this proposal are:
+
+* Provide required LokiStack API changes to enable a per-tenant OIDC configuration for the gateway component.
+* Declare per-tenant OIDC configuration optional but provide auto-configuration for OpenShift.
+* Leave per tenant OIDC secrets as an administrator task before creating LokiStack CRs.
+
+We will be successful when:
+
+* Deploying a LokiStack CR with per-tenant OIDC configuration enables secure authentication and authorization to the Loki cluster.
+* Deploying a LokiStack CR on OpenShift enables secure authentication and authorization with zero configuration to the Loki cluster.
+
+### Non-Goals
+
+### Risks and Mitigations
+
+In general the key risks to provide the proposed API changes in form of a custom resource definition are:
+
+**Risk:** Extra per-tenant boilerplate.
+
+**Mitigation:** We mitigate this at least in OpenShift Logging mode with secure presets.
+
+**Risk:** Maintaining compatibility to a third-party component `observatorium-api`.
+
+**Mitigation:** Here we are co-maintainers and we certainly want to secure the cluster as Loki does not provide any access security features.
+
+### User Stories
+
+* As a loki cluster admin, I want The LokiStack CR with per-tenant OIDC configuration and static RBAC enables secure authentication and authorization for `static` mode.
+
+* As a loki cluster admin, I want The LokiStack CR with per-tenant OIDC configuration and OPA endpoint enables secure authentication and authorization for `dynamic` mode.
+
+* As a loki cluster admin, I want The LokiStack CR with zero configuration on OpenShift enables secure authentication and authorization for `openshift-logging` mode.
+
+## Design Details
+
+For configuring the LokiStack Gateway, we need information about OIDC provider and OpenPolicyAgent (OPA).
+
+### lokistack_types.go
+
+Add a new field `Tenants` to `LokiStackSpec` struct as below:
+
+```go
+// ModeType is the authentication/authorization mode in which LokiStack Gateway will be configured.
+//
+// +kubebuilder:validation:Enum=static;dynamic;openshift-logging
+type ModeType string
+// PermissionType is a LokiStack Gateway RBAC permission.
+//
+// +kubebuilder:validation:Enum=read;write
+type PermissionType string
+
+// SubjectKind is a kind of LokiStack Gateway RBAC subject.
+//
+// +kubebuilder:validation:Enum=user;group
+type SubjectKind string
+
+const (
+  // Static mode asserts the Authorization Spec's Roles and RoleBindings
+	// using an in-process OpenPolicyAgent Rego authorizer.
+	Static ModeType = "static"
+	// Dynamic mode delegates the authorization to a third-party OPA-compatible endpoint.
+	Dynamic ModeType = "dynamic"
+	// OpenshiftLogging mode provides fully automatic OpenShift in-cluster authentication and authorization support.
+	OpenshiftLogging ModeType = "openshift-logging"
+
+	// Write gives access to write data to a tenant.
+	Write PermissionType = "write"
+	// Read gives access to read data from a tenant.
+	Read PermissionType = "read"
+
+	// User represents a subject that is a user.
+	User SubjectKind = "user"
+	// Group represents a subject that is a group.
+	Group SubjectKind = "group"
+)
+
+// LokiStackSpec defines the desired state of LokiStack
+type LokiStackSpec struct {
+  // Tenants defines the per-tenant authentication and authorization spec for the lokistack-gateway component.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Tenants Configuration"
+	Tenants *TenantsSpec `json:"tenants,omitempty"`
+}
+
+// TenantsSpec defines the mode, authentication and authorization
+// configuration of the lokiStack gateway component.
+type TenantsSpec struct {
+	// Mode defines the mode in which lokistack-gateway component will be configured.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=openshift-logging
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:static","urn:alm:descriptor:com.tectonic.ui:select:dynamic","urn:alm:descriptor:com.tectonic.ui:select:openshift-logging"},displayName="Mode"
+	Mode ModeType `json:"mode"`
+	// Authentication defines the lokistack-gateway component authentication configuration spec per tenant.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Authentication"
+	Authentication []AuthenticationSpec `json:"authentication,omitempty"`
+	// Authorization defines the lokistack-gateway component authorization configuration spec per tenant.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Authorization"
+	Authorization *AuthorizationSpec `json:"authorization,omitempty"`
+}
+
+// AuthenticationSpec defines the OIDC configuration per tenant for lokiStack Gateway component.
+type AuthenticationSpec struct {
+	// TenantName defines the name of the tenant.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Tenant Name"
+	TenantName string `json:"tenantName"`
+	// TenantID defines the id of the tenant.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Tenant ID"
+	TenantID string `json:"tenantId"`
+	// OIDC defines the spec for the OIDC tenant's authentication.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OIDC Configuration"
+	OIDC *OIDCSpec `json:"oidc"`
+}
+
+// OIDCSpec defines the OIDC configuration spec for lokiStack Gateway component.
+type OIDCSpec struct {
+ 	// Secret defines the spec for the clientID, clientSecret and issuerCAPath for tenant's authentication.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Tenant Secret"
+	Secret *TenantSecretSpec `json:"secret"`
+	// IssuerURL defines the URL for issuer.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Issuer URL"
+	IssuerURL string `json:"issuerURL"`
+	// RedirectURL defines the URL for redirect.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Redirect URL"
+	RedirectURL   string `json:"redirectURL"`
+	GroupClaim    string `json:"groupClaim"`
+	UsernameClaim string `json:"usernameClaim"`
+}
+
+// TenantSecretSpec is a secret reference containing name only
+// for a secret living in the same namespace as the LokiStack custom resource.
+type TenantSecretSpec struct {
+	// Name of a secret in the namespace configured for tenant secrets.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:io.kubernetes:Secret",displayName="Tenant Secret Name"
+	Name string `json:"name"`
+}
+
+// AuthorizationSpec defines the opa, role bindings and roles
+// configuration per tenant for lokiStack Gateway component.
+type AuthorizationSpec struct {
+  // OPA defines the spec for the third-party endpoint for tenant's authorization.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OPA Configuration"
+  OPA *OPASpec `json:"opa"`
+  // Roles defines a set of permissions to interact with a tenant.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Static Roles"
+  Roles []*RoleSpec `json:"roles"`
+  // RoleBindings defines configuration to bind a set of roles to a set of subjects.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Static Role Bindings"
+  RoleBindings []*RoleBindingsSpec `json:"roleBindings"`
+}
+
+// OPASpec defines the opa configuration spec for lokiStack Gateway component.
+type OPASpec struct {
+  // URL defines the third-party endpoint for authorization.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OpenPolicyAgent URL"
+    URL string  `json:"url"`
+}
+
+// RoleSpec describes a set of permissions to interact with a tenant.
+type RoleSpec struct {
+	Name        string           `json:"name"`
+	Resources   []string         `json:"resources"`
+	Tenants     []string         `json:"tenants"`
+	Permissions []PermissionType `json:"permissions"`
+}
+
+// RoleBindingsSpec binds a set of roles to a set of subjects.
+type RoleBindingsSpec struct {
+	Name     string    `json:"name"`
+	Subjects []Subject `json:"subjects"`
+	Roles    []string  `json:"roles"`
+}
+
+// Subject represents a subject that has been bound to a role.
+type Subject struct {
+	Name string      `json:"name"`
+	Kind SubjectKind `json:"kind"`
+}
+```
+
+## Proposal
+
+### LokiStack Gateway configuration using _Modes_
+
+LokiStack Gateway can operate in 3 different modes based on user's choice.
+
+* Static Mode - Enables authorization for an OIDC-authenticated subject based on the static lists of Roles and RoleBindings in `RbacSpec`.
+
+* Dynamic Mode - Enables authorization for an OIDC-authenticated subject by delegating the authorization request to a third-party OPA-compatible endpoint.
+
+* Openshift Logging Mode - Enables authorization for an OpenShift-authenticated subject by translating OPA authorization requests via [opa-openshift](https://github.com/observatorium/opa-openshift) to SubjectAccessReviews against the API server. This mode ensures a per-namespace tenant access for authenticated subjects.
+
+### LokiStack CR configuration - _Static Mode_
+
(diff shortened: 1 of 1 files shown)


enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,347 @@
...
+```go
+// LokiTemplateSpec defines the template of all requirements to configure
+// scheduling of all Loki components to be deployed.
+type LokiTemplateSpec struct {
+    // GatewayConfiguration defines the lokistack-gateway component configuration spec.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Gateway Configuration"
+	GatewayConfiguration *LokiGatewayConfigurationSpec `json:"gatewayConfiguration,omitempty"`
```
periklis, line 69: The `LokiTemplateSpec` struct consists of set of fields where each represents a component. `GatewayConfiguration` as field does not represent a component but rather extra configuration for the `Gateway` component. I suggest to use struct embedding here to make the API cleaner, i.e.:
```go
type LokiTemplateSpec struct {
    Gateway *GatewayTemplateSpec
}

type GatewayTemplateSpec struct {
    LokiComponentSpec
    Tenants []TenantSpec
    Rbac *RbacSpec
}
```

An alternative here is to keep `Gateway` as `*LokiComponentSpec` because `LokiTemplateSpec` is all about the deployment/scheduling of component pods on k8s. The tenants/rbac mix could live in a separate struct referenced in the toplevel LokiStackSpec, e.g.:
```go
type TenantsSpec struct {
    Tenants []TenantsSpec `json:"tenants,omitempty"`
    Rbac *RbacSpec `json:"rbac,omitempty"`
}

type LokiStackSpec struct {
    Tenants *TenantsSpec
}
```

periklis (commented):
In total the general course of this PR goes into the right direction to enable authn/authz configuration per tenant for the lokistack-gateway component. Awesome work! :rocket: 

In addition to the present state, I would like to make some suggestions and open the discussion to close down what we want to support initially and what we can push back for later. The suggestions below form the full picture of the gateway's configuration capabilities, which this enhancement proposal is missing to declare in the alternatives section.

## Regarding Authentication
The gateway `tenants.yaml` allows to declare `MTLS` or `OIDC` per tenant. Upon previous discussions we will start with `OIDC` only support. The proposed API is missing the following links:
- How does the user provide input which secret to pick for the `clientID/clientSecret` pair in `OIDCSpec`?

## Regarding Authorization
The gateway `tenants.yaml` and `rbac.yaml` can form together a unit to enable authorization of `OIDC` subjects, but they don't have to do this mandatorily. The two forms of authorization that are possible (assuming only OIDC here) are:
- **Dynamic OPA authorization**: Provide only an `OPA` url to delegate authorization of OIDC extracted subject information (see `UsernameClaim` and `GroupClaim`) to a third-part OPA-compatible endpoint.
- **Static OPA authorization**: Provide a static rego file (e.g. [observatorium-rego](https://github.com/observatorium/api/blob/f04cd85beb511697f53288fd724a27181e0b8c3a/test/config/observatorium.rego)) that can assert the contents of `rbac.yaml` from the OIDC extracted subject information (see `UsernameClaim` and `GroupClaim`) .

## Summary
Looking forward for integrating the gateway as a component in LokiStack we need to support two models:
- **For full OpenShift in-cluster support**: Based on the work in [opa-openshift](https://github.com/observatorium/opa-openshift/pull/1), we will need only **dynamic OPA authorization**, because the sidecar will handle authorization based on SubjectAccessReviews against the API server.
- **For a Simple Config support**: We need to offer **static OPA configuration** with a simple default rego file that can be part of the gateway deployment as a static file. This static rego file needs to understand the `RbacSpec`.

## Discussion
1. We need to settle down if the tenants (authn/authz) structs belong to `LokiStackSpec` or to a `GatewayComponentSpec` as pointed below.
2. We need to give users of the API a choice of which configuration mode they want to choose: a) static opa+rbac authorization, b) dynamic opa authorization, c) auto configured authorization (Only OpenShift). This can be an enum field on `LokiStackSpec`
3. Based on the user selection we need to allow inputs that match the selected mode, i.e.:
3.1 All three modes require a secret selection similar to `ObjectStorageSecretSpec.Name`
3.2 Selecting `Static OPA+RBAC Authz` makes `RbacSpec` mandatory
3.3 Selecting `Dynamic OPA Authz` makes `OPASpec.URL` mandatory and ignores `RbacSpec`
3.4 Selecting `Auto Configuration` renders any inputs obsolete.

WDYT?

periklis:
Based on our online conversation about the API, here is a sample on what aim:
```yaml
apiVersion: loki.openshift.io/v1beta1
kind: LokiStack
metadata:
  name: lokistack-sample
spec:
  size: 1x.small
  replicationFactor: 2
  storage:
    secret:
      name: test
  storageClassName: standard
  tenants:
    mode: static|dynamic|openshift-logging
    authentication:
    - name: tenant-a
      oidc:
        issuerUrl: sso.redhat.com/...
    - name: tenant-b
      oidc:
        issuerUrl: github.com/...
    authorization:
      opa: # Trade off: A single rest opa for all tenants
        url:
      rbac:
        roles: []
        rolebindings: []
```

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+## Implementation History
+
+| Release | Description              |
+| ---     | ---                      |
+| TBD     | **TP** - Initial release |
+| TBD     | **GA** - Initial release |
+
+## Drawbacks
+
+The drawback to not implementing this feature is we are unable to integrate Loki access to our org's OIDC provider and the current multi-tenant featurs of OpenShift Logging
```
bwplotka, line 485:
```suggestion
The drawback to not implementing this feature is we are unable to integrate Loki access to our org's OIDC provider and the current multi-tenant features of OpenShift Logging
```

Why? 🤗

bwplotka (commented):
Thanks for this and thanks @periklis for [pining us](https://coreos.slack.com/archives/C026ZQSB468/p1632298726068800) about this work.

I generally like this enhancement. Amazing opinionated options, easy to understand and configure.

My only concern is about abstracting away from Observatorium API. All of this is really configuring the Observatorium API component and [using this as a gateway for Loki](https://github.com/observatorium/api). Yet this fact is hidden, there is even no mention about this in the enhancement. However it would be beneficial to mention that most of this context does not need to be implemented - it is already satisfied by https://github.com/observatorium/api. 

This brings me to another observation. We create lot's of boilerplate and configuration mechanics on top of reusable components like Observatorium API. IF we instead have common Observatorium API tenancy CR, we can leverage that for any other signal than logging with Loki -> metrics and tracing too!

Just want to ensure we don't build something too custom. Especially I would avoid most of the abstractions like "Loki gateway". Telling that we install Observatorium API would help everyone enormously. WDYT? (: 

I know it would be easier if we have Observatorium API operated already somehow. But really looking on your LokiSpec we could use it in RHOBS for metric purposes already 🤗 (It's so well defined!)

Can we somehow abstract this spec from Loki and logging?

enhancements/cluster-logging/loki-gateway-configuration.md
periklis: @bwplotka The reason is to enable namespace-based tenancy as per OpenShift maintains this world view on resources. The developer persona has access only to namespaces (most of the times only application workload ns) granted by RBAC. The administrator persona has access to all namespaces including audit logs.

@sasagarw I suggest to add a couple of lines explaining this :point_up:

periklis:
> Thanks for this and thanks @periklis for [pining us](https://coreos.slack.com/archives/C026ZQSB468/p1632298726068800) about this work.
> 
> I generally like this enhancement. Amazing opinionated options, easy to understand and configure.
> 
> My only concern is about abstracting away from Observatorium API. All of this is really configuring the Observatorium API component and [using this as a gateway for Loki](https://github.com/observatorium/api). Yet this fact is hidden, there is even no mention about this in the enhancement. However it would be beneficial to mention that most of this context does not need to be implemented - it is already satisfied by https://github.com/observatorium/api.
> 

@sasagarw This is a valid concern to mention that we are simply providing an API to reconfigure an existing implementation i.e. observatorium-api

> This brings me to another observation. We create lot's of boilerplate and configuration mechanics on top of reusable components like Observatorium API. IF we instead have common Observatorium API tenancy CR, we can leverage that for any other signal than logging with Loki -> metrics and tracing too!
> 
> Just want to ensure we don't build something too custom. Especially I would avoid most of the abstractions like "Loki gateway". Telling that we install Observatorium API would help everyone enormously. WDYT? (:
> 
> I know it would be easier if we have Observatorium API operated already somehow. But really looking on your LokiSpec we could use it in RHOBS for metric purposes already hugs (It's so well defined!)
> 
> Can we somehow abstract this spec from Loki and logging?


@bwplotka Thanks for the review and the motivating words! :hugs: 

Assuming an Observatorium API Tenancy CR exists, the  question here is which operator/controller is owning and controlling this resource. Certainly having such a CR requires a different operator/controller dedicated to the observatorium-api rather than using the loki-operator to control it. In turn this means that operating a LokiStack CR requires from a user perspective to install two operators:
- One for observatorium-api controlling the proposed Observatorium API Tenancy CR.
- One for loki controlling the LokiStack CR

This sounds like a lot of deployment boilerplate to get the benefits of securing authn/authz of Loki/Thanos/Jaeger.

Looking forward I can think of two approaches but both have trade-offs:
1. An alternative approach could be to have the CRD + Go Types + Controller code for `observatorium-api` in a simple library which can be embedded and re-used in a loki-operator as well as in any other signal-specific operator, e.g. `observatorium-operator`, `jaeger-operator`, `thanos-operator` etc. However, this would bite every non-expert user when installing multiple operators (e.g. loki-operator, thanos-operator) which both control the same resource. The trade-off here is to provide instructions to use an `observatorium-operator` for the `observatorium-api` and loki/thanos operators for the signals.
2. Another approach is to take the types from this enhancement proposal in a simple library that renders them into the configuration yaml files only. Every operator can re-use and expose the types in its own CRDs, e.g. LokiStack, ThanosStack, etc. This means that every operator controls a custom obs-api instance for the controlled signal. The trade-off here is that we don't have a single observatorium-api securing multiple signals at a time. Thus any authn/authz changes need to be applied on multiple CRs separately.

WDYT?

bwplotka:
Hm, I think I like option ~1. If I understand this right, providing ObservatoriumAPITenancy CR that allow secure logging/metrics/tracing endpoints.

We can build it as lib and provide simple flag to turn-off this logic if Obs operator is deployed (or detect this). But TBH having dependency of Loki operator on some bare bones Observatorium operator that controls API might cleaner way to go.

We could split that in iterations and provide library idea first, that would be close to ideal (:

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+
+### Goals
+
+The goals of this proposal are:
+
+* Provide changes to API to accept OIDC provider in LokiStack CR
+* Make OIDC configuration optional and auto-configuration on OCP
+* Secrets/Credentials/SA-Keys should be provided via a k8s secret that the LokiStack admin provides upfront
+
+We will be successful when our LokiStack CR is able to accept OIDC provider and able to configure LokiStack Gateway
```
periklis, line 50: Suggest to make this a bullet point list to highlight success on OpenShift separately:
```md
We will be successful when:
- Deploying a LokiStack CR with per-tenant OIDC configuration enables secure authentication and authorization to the Loki cluster.
- Deploying a LokiStack CR on OpenShift enables secure authentication and authorization with zero configuration to the Loki cluster.

File: enhancements/cluster-logging/loki-gateway-configuration.md [hunk 4]
```diff
@@ -0,0 +1,517 @@
...
+
+### Non-Goals
+
+### Risks and Mitigations
+
+NA
```
periklis, line 56: I suggest to mention that adding these changes into the present API creates some:
1. Extra per-tenant boilerplate -> At least we mitigate in OpenShift Logging with secure presets.
2. Maintaining compatibility to a third-party component `observatorium-api` -> Here we are co-maintainers and we certainly want to secure the cluster as per Loki does not provide any access security features.

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+
+### User Stories
+
+* As a loki cluster admin, I want The LokiStack CR be able to accept OIDC configuration.
+
+* As a cluster admin, I want the Loki Operator be able to reconcile correctly.
```
periklis, line 62: Let's replace these stories with one story per mode `static`, `dynamic`, `openshift-logging`.

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+
+## Design Details
+
+For configuring the LokiStack Gateway, we need information about oidc provider and opa.
```
periklis, line 66: Identifier always in capital letters, i.e. `OIDC`, `OPA`. AFAICS OPA is mentioned here the first time and deserves a full naming as `OpenPolicyAgent (OPA)`

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+}
+```
+
+## Proposal
+
+### LokiStack Gateway configuration using _Modes_
+
+LokiStack Gateway can operate in 3 different modes based on user's choice.
+
+* Static Mode - It will offer static OPA configuration with a simple default rego file that will be part of the lokiStack gateway deployment as a static file. This static rego file needs to understand the RbacSpec. This makes `RbacSpec` mandatory.
```
periklis, line 259: I suggest to reword this with focus on what it solves, e.g.:
```
Static Mode: Enables authorization for an OIDC-authenticated subject based on the static lists of Roles and RolesBinding in `RbacSpec`.
```

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+
+* Dynamic Mode - It will offer static OPA configuration along with an OPA url to delegate authorization of OIDC extracted subject information to a third-party OPA-compatible endpoint. This makes `OPA URL` mandatory and ignores `RbacSpec`.
+
```
periklis, line 262: Same here, e.g.:
```
Dynamic Mode - Enables authorization for an OIDC-authenticated subject by delegating the authorization request to a third-party OPA-compatible endpoint.

File: enhancements/cluster-logging/loki-gateway-configuration.md [hunk 9]
```diff
@@ -0,0 +1,517 @@
...
+* Openshift Logging Mode - It will not need any information from the user. The lokiStack gateway deployment will configure dynamic OPA authorization itself using [opa-openshift](https://github.com/observatorium/opa-openshift/pull/1), because the sidecar will handle authorization based on SubjectAccessReviews against the API server.
```
periklis, line 263: Same here:
```
OpenShift Logging Mode - Enables authorization for an OpenShift authenticated subject by translating OPA authorization requests via [opa-openshift](https://github.com/observatorium/opa-openshift) to SubjectAccessReviews against the API server. This mode ensures a per-namespace tenant access for authenticated subjects.
```

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+
+### LokiStack CR configuration - _Static Mode_
+
+In this mode, user will need to provide tenants, oidc and rbac configuration madatorily in the LokiStack CR. The lokiStack gateway deployment will itself create a static rego file to understand the rbacSpec.
```
periklis, line 267: wording: `In this mode it is mandatory to provide a static list of per-tenant OIDC configuration as well as static RBAC. The LokiStack gateway deployment will be deployed with access to a static Rego file to match the RbacSpec. The static Rego file represents an in-process OPA endpoint.`

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+          subjects:
+          - kind: user
+            name: admin@example.com
+        roles:
+        - name: read-write
+          permissions:
+          - read
+          - write
+          resources:
+          - metrics
```
periklis, line 296: Suggest to use the `logs` resources for examples here and below as this proposals targets the LokiStack API.

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+          tenants:
+          - tenant-a
+```
+
+The LokiStack Gateway deployment will need to create a static rego file which will include:
+
+```rego
+package observatorium
```
periklis, line 304: Let's remove the references to observatorium here as we did in the implementation

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+}
+
+matched_role_binding[i] {
+  roleBindings[i].subjects[_] == {"name": input.groups[_], "kind": "group"}
+}
+```
+
+### LokiStack CR configuration - _Dynamic Mode_
+
+In this mode, user will need to configure tenants, oidc and an OPA URL madatorily in the LokiStack CR. This OPA url will delegate authorization of OIDC extracted subject information to a third-party OPA-compatible endpoint.
```
periklis, line 332: wording: `In this mode it is mandatory to provide a static list of per-tenant OIDC configuration and a URL to an OPA endpoint. Latter is used to delegate the subject's information to access the requested tenant.`

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+
+During the reconcilation of LokiStack CR, the user provided details will be validated. If anything is missing or provided extra, the Loki operator will give out an error.
```
periklis, line 334: We don't need to mention validation of required fields here. However we can leave a word that mixing static RBAC with dynamic mode results in a LokiStack CR degraded condition.

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+    authentication:
+    - name: tenant-a
+      oidc:
+        issuerURL: https://127.0.0.1:5556/dex
+        redirectURL: https://localhost:8443/oidc/tenant-a/callback
+        usernameClaim: test
+        groupClaim: test
+    authorization:
+      opa:
+        url: http://127.0.0.1:8080/v1/data/observatorium/allow
```
periklis, line 350: Let's remove the references to observatorium here as we did in the implementation

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+```
+
+### LokiStack CR configuration - _Openshift Logging Mode_
+
+In this mode, user will need to configure nothing as everything will be done by Loki operator itself.
+
+During the reconcilation of LokiStack CR, if user provides anything other than `mode`, the Loki operator will give out an error.
```
periklis, line 357: We should rewrite this to explain why and how the auto-configuration works here.

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+  id: 32e45e3e-b760-43a2-a7e1-02c5631e56e9
+  oidc:
+    clientID: test
+    clientSecret: test
+    issuerCAPath: ./tmp/certs/ca.pem
+    issuerURL: https://127.0.0.1:5556/dex
+    redirectURL: https://localhost:8443/oidc/application/callback
+    usernameClaim: name
+  opa:
+    url: http://127.0.0.1:8080/v1/data/observatorium/allow
```
periklis, line 380: Let's remove the references to observatorium here as we did in the implementation

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+  id: 40de0532-10a2-430c-9a00-62c46455c118
+  oidc:
+    clientID: test
+    clientSecret: ZXhhbXBsZS1hcHAtc2VjcmV0
+    issuerCAPath: ./tmp/certs/ca.pem
+    issuerURL: https://127.0.0.1:5556/dex
+    redirectURL: https://localhost:8443/oidc/infrastructure/callback
+    usernameClaim: name
+  opa:
+    url: http://127.0.0.1:8080/v1/data/observatorium/allow
```
periklis, line 391: Let's remove the references to observatorium here as we did in the implementation

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+  id: 26d7c49d-182e-4d93-bade-510c6cc3243d
+  oidc:
+    clientID: test
+    clientSecret: test
+    issuerCAPath: ./tmp/certs/ca.pem
+    issuerURL: https://127.0.0.1:5556/dex
+    redirectURL: https://localhost:8443/oidc/audit/callback
+    usernameClaim: name
+  opa:
+    url: http://127.0.0.1:8080/v1/data/observatorium/allow
```
periklis, line 402: Let's remove the references to observatorium here as we did in the implementation

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+```
+
+### Creating Secrets
+
+The oidc configuration also expects `clientID`, `clientSecret` and `issuerCAPath` which should be provided via a k8s secret that the LokiStack admin provides upfront.
```
periklis, line 407: `oidc` -> `OIDC`, `k8s` -> `Kubernetes`

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+kind: Secret
+metadata:
+  name: lokiStack-gateway-tenant-a
+data:
+  clientID: test
+  clientSecret: ZXhhbXBsZS1hcHAtc2VjcmV0
+  issuerCAPath: /tmp/certs/ca.pem
+```
+
+The above data all should be base64 encoded.
```
periklis, line 420: This is not worth to mention as Kubernetes Secret require this encoding per default.

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+
+The name provided in `metadata.name` should be mentioned in TenantSecretSpec so that correct secrets are picked in OIDCSpec.
+
+If we want to skip the base64 encoding step, we can create the same Secret using the `kubectl create secret` command. For example:
+
+```text
+kubectl create secret generic lokiStack-gateway-tenant-a --from-literal='clientID=test' --from-literal='clientSecret=ZXhhbXBsZS1hcHAtc2VjcmV0' --from-literal='issuerCAPath=/tmp/certs/ca.pem'
+```
```
periklis, line 428: We can skip this too

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,517 @@
...
+  name: lokiStack-gateway-tenant-a
+data:
+  clientID: test
+  clientSecret: ZXhhbXBsZS1hcHAtc2VjcmV0
+  issuerCAPath: /tmp/certs/ca.pem
+```
+
+The above data all should be base64 encoded.
+
+The name provided in `metadata.name` should be mentioned in TenantSecretSpec so that correct secrets are picked in OIDCSpec.
```
periklis, line 422: wording: 
```
Each tenant Secret is required to match:
- `metadata.name` with `TenantsSecretsSpec.Name`
- `metadata.namespace` with `LokiStack.metadata.namespace`

File: enhancements/cluster-logging/loki-gateway-configuration.md [hunk 24]
```diff
@@ -0,0 +1,517 @@
...
+```yaml
+roleBindings:
+{{- range $roleBinding, $spec := .Stack.Template.GatewayConfiguration.Rbac.RoleBindings }}
+- {{ $roleBinding }}: {{ $spec.Name }}
+{{- end -}}
+roles:
+{{- range $role, $spec := .Stack.Template.GatewayConfiguration.Rbac.Roles }}
+- {{ $role }}: {{ $spec.Name }}
+{{- end -}}
+```
```
periklis, line 455: We don't need to provide implementation details of such detail level.

enhancements/cluster-logging/loki-gateway-configuration.md
```diff
@@ -0,0 +1,498 @@
...
+    authorization:
+      opa:
+        url: http://127.0.0.1:8080/v1/data/lokistack/allow
+```
+
+### LokiStack CR configuration - _Openshift Logging Mode_
+
+In this mode, user will need to do zero configuration as everything will be auto-configured by Loki operator itself.
+
+After the new [data model](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/cluster-logging-es-rollover-data-design.md#data-model) was introduced, the logs from individual namespaces share common index. This was done because of multi-tenant feature of Openshift Logging which enables namespace-based tenancy since Openshift maintains this global view on resourses.
```
periklis, line 369: This paragraph looks to be out of context. The new data model is Elasticsearch-only specific. The separation in three big buckets `application`, `infrastructure` and `audit` is more of a conceptual one that guides how to the collector and forwarder in OpenShift Logging categorizes logs. We continue using these three categories as three big tenants in Loki, because using namespaces as tenants will make our chunks too small. We ensure namespace-scoping by adding them as label-selectors, e.g.:

Let's say a user in OpenShift requires access to the logs of pod `my-pod` on namespace `my-project`, the call to lokistack will look like (translated in curl call for easier comprehension): 
```bash
$ curl -G -s  "http://lokistack.svc:8080/api/logs/v1/application/loki/api/v1/query_range" --data-urlencode 'query={pod_name="my-pod"}' --data-urlencode 'step=300' 
```

The lokistack-gateway will ask the authorizer sidecar `opa-openshift` and will retrieve the list of projects the user has access to and in turn amend the query from:
`{pod_name="my-pod"}`
to 
`{pod_name="my-pod", namespace=~"my-project|my-ther-project"}`

Thus the final upstream call to loki looks like (translated in curl call for easier comprehension):
```shell
$ curl -G -s -H "X-Scope-Org-ID: application"  "http://query-frontend.svc:8080/loki/api/v1/query_range" --data-urlencode 'query={pod_name="my-pod", namespace=~"my-project|my-ther-project"}' --data-urlencode 'step=300' 
```

dhellmann:

The template changed while this design doc was being written.

PR #887, opened 2021-09-03
NE-542 Expose router compression variables

candita:
This enhancement proposal extends the IngressController API to allow the cluster administrator
to configure HTTP traffic compression via HAProxy variables.  HTTP traffic compression increases the
performance of a website by reducing the size of data that is passed in HTTP transactions. 

Customers with large amounts of compressible routed traffic need the ability to gzip-compress
their ingress workloads, as described in feature request https://issues.redhat.com/browse/NE-542.

The ability to configure compression was available in 3.x, and removed in 4.x.  This enhancement will bring feature parity back to 4.10.

### DIFF
--- a/enhancements/ingress/haproxy-router-compression.md
+++ b/enhancements/ingress/haproxy-router-compression.md
@@ -53,55 +53,53 @@ their ingress workloads, as described in feature request [NE-542](https://issues
 HAProxy in OpenShift 3.x exposed compression variables, so lack of access to the variables in OpenShift 4.x
 is a migration blocker for customers that were using them in OpenShift 3.x.
 
-The ability to configure compression was available in 3.x, but it applied only to the HAProxy
-`defaults` configuration section, and supported only the `gzip` compression algorithm.  This proposal
-reintroduces and then enhances the previous implementation in these ways:
-- allow the `frontend`, and `backend` sections in addition to the `defaults` section, which would provide for more
-  fine-grain configuration at the route level.
-- allow a choice of `gzip` or `raw-deflate` compression algorithms.
+The ability to configure compression was available in 3.11 and disappeared in 4.x. This proposal
+reintroduces this ability to the extent of feature parity with 3.11.
 
 ### Goals
 
-Enable the configuration of HTTP compression global defaults via the `IngressControllerSpec`.
+Enable the configuration of HTTP compression global defaults via the `IngressControllerSpec`, which will
+expose the two environment variables `ROUTER_ENABLE_COMPRESSION` and `ROUTER_COMPRESSION_MIME`.
 
-Enable the configuration of HTTP compression for route backends/frontends via route annotations.
+Enable HTTP compression on selected MIME types, be they wildcard (e.g. `text/*`, `application/*`, etc.) or
+specific (e.g. `text/html`, `text/plain`, `text/css`, `application/json`, `image/svg+xml`, `font/eot`, etc.).
 
-Enable HTTP compression on selected MIME types, be they wildcard (e.g. text/*, application/*, etc.) or
-specific (e.g. text/html, text/plain, text/css, application/json, image/svg+xml, font/eot, etc.).  Not all MIME
-types benefit from compression, but there is a short list found [here](https://letstalkaboutwebperf.com/en/gzip-brotli-server-config/)
-that will be implemented as a part of this proposal.
-
-By exposing ROUTER_ENABLE_COMPRESSION, any HTTP request with the header "Accept-Encoding: gzip" or
-"Accept-Encoding: raw-deflate" with a MIME type that matches one of the configured ROUTER_COMPRESSION_MIME types
-will have HTTP compression applied to its payload.
+By exposing the environment variables, any HTTP request that terminates within HAProxy, with the header
+**"Accept-Encoding: gzip"**, and with a MIME type that matches one of the configured MIME types, will qualify to have
+HTTP compression applied to its payload.  Note 1: HAProxy can skip compression if it does not have the resources
+available to provide compression at the moment when it is requested.
 
 ### Non-Goals
 
-HAProxy provides for the choice of compression algorithm (e.g. `identity`, `gzip`, `deflate`,
-and `raw-deflate`) but this proposal selectively covers only the `gzip` and `raw-deflate` algorithms.
-The `identity` algorithm is for debugging and doesn't apply any change on data.  The `deflate` algorithm
-is a variation on `gzip` that is not supported on all browsers and lacks universal value.
+Though it is possible to provide the following options, they are not included in this proposal because they
+were not requested in the RFE, which requests only feature parity with OCP 3.11.
 
-HAProxy allows compression to be applied to the `listen` section as well as the `defaults`, `backend`, and `frontend`
-sections of the configuration file.  A `listen` section is used to define a combined `frontend` and `backend`.
-Since the current OpenShift HAProxy template only has a `listen` definition for the `StatsPort`, and
-doesn't allow user configuration of a `listen` section, it is not considered in this proposal.
+* It is possible for us to validate the MIME types in order to preclude pointless compression on data that is already
+  compressed.  This proposal specifically allows admins to choose their own MIME types but will provide guidance in the
+  godoc on how to avoid pointless compression.  It is more likely that we would receive a request to change the allowed
+  MIME types, than to receive a request for support on a HAProxy spending unnecessary resources on compression duties.
 
-HAProxy provides a `compression offload` configuration option that makes HAProxy remove the "Accept-Encoding"
+* HAProxy provides for the choice of compression algorithm (e.g. `identity`, `gzip`, `deflate`,
+and `raw-deflate`).  This proposal covers only the `gzip` algorithm.
+
+* HAProxy allows compression to be applied to the `listen`, `defaults`, `backend`, and `frontend`
+sections of the configuration file.  This proposal covers only the global `defaults` section.
+
+* HAProxy provides a `compression offload` configuration option that makes HAProxy remove the "Accept-Encoding"
 header to prevent backend servers from compressing responses.  This forces all response compression to happen within
 HAProxy, jeopardizing its performance.  This configuration option is not recommended, even by HAProxy guides, so it
 is omitted from this proposal.
 
-A performance tuning variable, `maxcomprate`, sets the maximum per-process input compression rate to the specified
-number of kilobytes per second.  It works in conjunction with `tune.comp.maxlevel` to rate-limit the compression
-level to save data bandwidth. Another, `maxcompcpuusage`, does the same to save limit CPU used on compression.
-Neither variable is currently exposed, and both are omitted from this proposal.  Other per-session performance tuning
+* HAProxy provides a performance tuning variable, `maxcomprate`, to set the maximum per-process input compression rate
+  to the specified number of kilobytes per second.  It works in conjunction with `tune.comp.maxlevel` to rate-limit the
+  compression level to save data bandwidth. Another, `maxcompcpuusage`, does the same to limit CPU used on compression.
+Neither variable is currently exposed, and both are omitted from this proposal.  (Other per-session performance tuning
 variables like `maxconnrate`, which set the maximum per-process number of connections per second, and `maxsslrate`,
-which sets the maximum per-process number of SSL sessions per second, are not exposed either.
+which sets the maximum per-process number of SSL sessions per second, are not exposed either.)
 
 ## Proposal
 
-Currently, the haproxy.cfg contains the lines below in the `defaults` section, but the mechanism for setting these
+Currently, OpenShift router's `haproxy-config.template` file contains the lines below in the `defaults` section, but the mechanism for setting the
 environment variables is not exposed.
 ```go
 defaults
@@ -114,53 +112,6 @@ defaults
 
 ### Global HTTP Compression via IngressControllerSpec and environment variables
 
-To expose the two compression directives in the `defaults` section, the ROUTER_ENABLE_COMPRESSION environment variable
-will be replaced with ROUTER_COMPRESSION_ALGORITHM, and these changes will be made:
-```go
-{{- /* compressionAlgoPattern matches valid options for HTTP compression */}}
-{{- $compressionAlgoPattern := "gzip|raw-deflate" -}}
-
-{{- /* compressionMimeTypes stores valid options for HTTP compression MIME Types */}}
-{{- $compressionMimeTypes := "text/html text/css text/plain text/xml text/x-component text/javascript \
-application/x-javascript application/javascript application/json application/manifest+json application/vnd.api+json \
-application/xml application/xhtml+xml application/rss+xml application/atom+xml application/vnd.ms-fontobject \
-application/x-font-ttf application/x-font-opentype application/x-font-truetype \
-image/svg+xml image/x-icon image/vnd.microsoft.icon font/ttf font/eot font/otf font/opentype" -}}
-...
-defaults
-...
-{{- with  $compressionAlg := firstMatch $compressionAlgoPattern {{ env "ROUTER_COMPRESSION_ALGORITHM"}} -}}
-compression algo {{ $compressionAlg }}
-{{- with $compressionMimeTypes := allMatches $compressionMimeTypes {{ env "ROUTER_COMPRESSION_MIME"}} -}}
-compression type {{ $compressMimeTypes }}
-{{- end }}
-{{- end -}}
-```
-The new function `allMatches` will be added to the `template_helper.go` code:
-```go
-// Compare two white-space delimited strings and return the latter if all its substrings
-// exist in former
-func allMatches(defined string, supplied string) string {
-	log.V(7).Info("allMatches called", "supplied", supplied, "defined", defined)
-	numFound := 0
-	suppliedRange := strings.Fields(supplied)
-	definedRange := strings.Fields(defined)
-	for _, s := range suppliedRange {
-		for _, d := range definedRange {
-            if d == s { // optimize this
-                log.V(7).Info("allMatches validated", "supplied", s)
-                numFound++
-                break
-            }
-        }
-	}
-	if numFound == len(suppliedRange){
-        return supplied
-    }
-	return ""
-}
-```
-
 The proposed change to IngressControllerSpec exposes the HTTP compression configuration variables at the default level:
 ```go
 type IngressControllerSpec struct {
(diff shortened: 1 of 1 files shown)


enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,413 @@
...
+#### As a cluster administrator, I need to enable gzip HTTP compression for only one route and one MIME type, "text/html"
+
+To apply HTTP compression to all routes in the cluster, use a command like this. (This command requires
+an updated `oc` client that supports the `--all-namespaces` flag on the `oc annotate` verb):
+```shell
+$ oc annotate route --all --all-namespaces --overwrite=true "haproxy.router.openshift.io/compressionAlgorithm"="gzip"
+$ oc annotate route --all --all-namespaces --overwrite=true "haproxy.router.openshift.io/compressionMIMETypes"="text/html text/css text/xml"
+
+```
+To apply HSTS to all routes in a particular namespace, use a command like this:
```
frobware, line 286: s/HSTS/HTTP

candita: done

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,421 @@
...
+Enable the configuration of HTTP compression global defaults via the `IngressControllerSpec`.
+
+Enable the configuration of HTTP compression for route backends/frontends via route annotations.
+
+Enable HTTP compression on selected MIME types, be they wildcard (e.g. text/*, application/*, etc.) or
+specific (e.g. text/html, text/plain, text/css, application/json, image/svg+xml, font/eot, etc.).  Not all MIME
+types benefit from compression, but there is a short list found [here](https://letstalkaboutwebperf.com/en/gzip-brotli-server-config/)
+that will be implemented as a part of this proposal.
+
+By exposing ROUTER_ENABLE_COMPRESSION, any HTTP request with the header "Accept-Encoding: gzip" or
```
deads2k, line 74: This will only apply to routes that the router terminates right?  So not pass through.

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,421 @@
...
+	// compressionAlgorithm defines the desired compression algorithm.
+	//
+	// +required
+	CompressionAlgorithm CompressionAlgorithmType `json:"compressionAlgorithm,omitempty"`
+	
+	// compressionMIMETypes is a list of MIME types that should have compression applied
+	//
+	// +kubebuilder:MinItems=1
+	// +required
+	CompressionMIMETypes []string `json:"compressionMIMETypes,omitempty"`
```
deads2k, line 195: My lack of http knowledge shows here, but is this possible to make an enum to avoid accidents?

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,421 @@
...
+  httpCompression:
+    compressionAlgorithm: gzip
+    compressionMIMETypes:
+    - "text/html"
+    - "text/css"
+    - "application/pdf"
+```
+
+### Route-based HTTP Compression via Annotations
+To expose these variables at the route level, the mechanism will be two route annotations that are processed in the
```
deads2k, line 225: The annotation based API for HSTS is a historical mistake.  Knowing what we know now, it is a better idea to modify the route API itself to add a new stanza.  Probably even the same stanza you have for the ingress spec.

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,421 @@
...
+spec:
+  httpCompression:
+    compressionAlgorithm: gzip
+    compressionMIMETypes:
+    - "text/html"
+    - "text/css"
+    - "application/pdf"
+```
+
+### Route-based HTTP Compression via Annotations
```
deads2k, line 224: Can you describe in this section whether the per-route configuration always overrides the per-ingress configuration?  How does a route indicate no compression when the ingress configuration says to compress?

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,421 @@
...
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
+
+### Upgrade / Downgrade Strategy
+
+An upgrade from a previous installation that had the compression directives exposed, will need to change to use the
+IngressController spec.  No previous installation would have any compression directives via route annotations.
+
+If a cluster administrator applied any of the compression configuration options
+in 4.10, then downgraded to 4.9, the compression configuration settings would no longer be in effect unless they
```
deads2k, line 405: Please be sure that there is an explicit test of this capability done by QE on rollback.

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,421 @@
...
+## Implementation History
+n/a
+
+## Drawbacks
+
+By implementing this proposal we are giving the user tools that could threaten the operation of HAProxy.
+Improper configuration of the MIME types or general over-use of compression could allow the HAProxy to
+dedicate more processing time to compression than to its main duties of routing and traffic management.
+
+## Alternatives
```
deads2k, line 420: Because this feature is not security related and because the side-effects are minimal, short-lived, and fixable by the server enforcement by ingress is ok.  

Does HAProxy allow configuration of a default that applies by domain?  If so, is that more natural for customers to configure?  An answer down here in alternatives would help future us remember why the choice was made.

enhancements/ingress/haproxy-router-compression.md
candita: Based on feedback from the team, we will remove route-based HTTP compression options in this release.

enhancements/ingress/haproxy-router-compression.md
candita: Based on feedback from the team, we will not add route-based HTTP compression options in this release.

enhancements/ingress/haproxy-router-compression.md
candita: Based on feedback from the team, we will not add route-based HTTP compression options in this release.  HAProxy does not allow configuration that applies by domain.

enhancements/ingress/haproxy-router-compression.md
candita: Based on feedback from the team, we will not be adding any compressionMIMEType validation or enforcing policy, except maybe a basic regex.

enhancements/ingress/haproxy-router-compression.md
candita: Will add a unit test too.

enhancements/ingress/haproxy-router-compression.md
candita: Now I have an open question about whether my latest solution is too strict.

enhancements/ingress/haproxy-router-compression.md
candita: Thanks, I clarified that.

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -173,38 +125,46 @@ type IngressControllerSpec struct {
...
 
-	// compressionAlgorithm defines the desired compression algorithm.
+	// compressionEnabled defines the desired compression enablement
+	// If true, compression is enabled for all MIME types specified in
+	// CompressionMIMETypes.  If false, compression is disabled for any MIME type.
 	//
 	// +required
-	CompressionAlgorithm CompressionAlgorithmType `json:"compressionAlgorithm,omitempty"`
+	// +kubebuilder:default=false
+	CompressionEnabled bool `json:"compressionEnabled"`
```
frobware, line 143: Do we need this if we have entries in `CompressionMIMETypes`. If `CompressionMIMETypes` is empty then compression is disabled. I don't know whether users would want the convenience of toggling `CompressionEnabled` vis-a-vis just removing the list of mime types. Thoughts?

frobware (commented):
Overall LGTM. Just one question about the api.
Leaving to @Miciah to take another look and/or tag.

enhancements/ingress/haproxy-router-compression.md
candita: I agree, we don't need this. Removed.  @frobware @miciah PTAL at latest change (minor) and tag if lgty.

frobware:
LGTM overall. @candita thanks for the API change. Leaving for @Miciah to take a look and/or tag.

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,322 @@
...
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	CompressionMIMETypes []CompressionMIMEType `json:"compressionMIMETypes,omitempty"`
+}
+
+// CompressionMIMEType defines the format of a single MIME type
+// E.g. "text/css; charset=utf-8", "text/html", "text/*", "image/svg+xml",
+// "application/octet-stream", etc.
+//
+// +kubebuilder:validation:Pattern="[a-zA-Z0-9\\+-;= ]+/([a-zA-Z0-9\\\+-;= ]+|\\*)"
```
Miciah, line 154: This should be anchored.  You shouldn't need to escape `+` in a character class.  Are spaces really allowed in the type?  https://datatracker.ietf.org/doc/html/rfc1341#page-7 indicates that spaces, `;`, and `=` should be prohibited in the type and subtype.

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,322 @@
...
+  namespace: openshift-ingress-operator
+spec:
+  httpCompression:
+    compressionMIMETypes:
+    - "text/html"
+    - "text/css; charset=utf-8"
+    - "application/json"
+```
+
+To expose the HTTP compression variables, further changes will be added to the ingressi router deployment:
```
Miciah, line 174: Looks like this got cut off.  
```suggestion
To expose the HTTP compression variables, further changes will be added to the ingress operator's logic to build the router deployment:
```

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,322 @@
...
+	RouterEnableCompression = "ROUTER_ENABLE_COMPRESSION"
+	RouterCompressionMIMETypes = "ROUTER_COMPRESSION_MIME"
+)
+...
+    // desiredRouterDeployment returns the desired router deployment.
+    func desiredRouterDeployment(ci *operatorv1.IngressController, ingressControllerImage string, ingressConfig *configv1.Ingress, apiConfig *configv1.APIServer, networkConfig *configv1.Network, proxyNeeded bool, haveClientCAConfigmap bool, clientCAConfigmap *corev1.ConfigMap) (*appsv1.Deployment, error) {
+    ...
+      if len(ci.Spec.HTTPCompression.CompressionMIMETypes) != 0 {
+        env = append(env, corev1.EnvVar{Name: RouterEnableCompression, Value: "true"})
+        env = append(env, corev1.EnvVar{Name: RouterCompressionMIMETypes, Value: string.Join(spec.httpCompression.compressionMimeTypes, " ")})
```
Miciah, line 188: Do we need to quote or escape values with spaces (such as `text/css; charset=utf-8` in the example above)?

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,322 @@
...
+[HAProxy compression documentation](https://www.haproxy.com/documentation/hapee/latest/load-balancing/compression/),
+but these are not in scope of this proposal.
+
+### Risks and Mitigations
+
+Overuse of compression can cause the HAProxy to devote more resources to compression than to other operations,
+compromising its operation.  We do not have any evidence that this will occur, and give detailed notes on use of
+necessary MIME types to mitigate pointless compression and overuse of compression.  Should this become a problem
+in the future, we can add default values for the process-level directives mentioned above (`maxcomprate`,
+`maxcompcpuusage`, and `tune.comp.maxlevel`)  The team consensus is that adding more tunable knobs at this stage is
```
Miciah, line 235:
```suggestion
`maxcompcpuusage`, and `tune.comp.maxlevel`).  The team consensus is that adding more tunable knobs at this stage is
```

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,322 @@
...
+HAProxy tries to compress all responses that are not compressed by the server.  Therefore, at least one MIME type is
+required in the `HTTPCompressionPolicy`.  We consider the fact that cluster admins may want to create and use their
+own MIME type, but the allowed set of characters is defined in a kubebuilder validation
+pattern  of `"[a-zA-Z0-9\\+-;= ]+/([a-zA-Z0-9\\\+-;= ]+|\\*)"`, to allow a token of alphanumeric plus "_-;= " followed
+by "/" followed by another alphanumeric token of the same pattern, or "*", the wildcard character.
+
+### Open Questions
+Is the MIME type pattern still too strict?  I reviewed the `mime` golang package at https://golang.org/src/mime/type.go,
+`builtinTypesLower` map, but if we intend to allow any custom MIME type, perhaps a wider net should be cast, as in
+https://golang.org/src/mime/grammar.go.
```
Miciah, line 260: https://datatracker.ietf.org/doc/html/rfc1341#page-7 provides BNF definitions for allowed MIME types.  If we want to make it a little looser (e.g., allowing non-standard types without the `x-` prefix), I don't object.

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,322 @@
...
+#### Dev Preview -> Tech Preview
+n/a
+
+#### Tech Preview -> GA
+n/a
+
+#### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
```
Miciah, line 298: N/A.

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Upgrade / Downgrade Strategy
+
+An upgrade from a previous installation that had the compression directives exposed, will need to change to use the
+IngressController spec.
```
Miciah, line 303: Previous versions did not expose the compression directives.  The API is additive and turned off by default, so there should be no change in behavior or need for migration during upgrades.

enhancements/ingress/haproxy-router-compression.md
candita: "text/css; charset=utf-8" is a valid MIME type, so I included spaces, `;`, and `=` in the pattern.

candita: And will add `"`.

enhancements/ingress/haproxy-router-compression.md
candita: The quotes need to be there already.  No validation done at this point.  I'll add quotes to the pattern.

enhancements/ingress/haproxy-router-compression.md
candita: I kept this pretty loose.  RFC1341 is made obsolete by https://datatracker.ietf.org/doc/html/rfc1521#page-68  

I will anchor and remove escapes from `+`, but if we don't keep this loose, we need to use https://datatracker.ietf.org/doc/html/rfc1521#page-68

enhancements/ingress/haproxy-router-compression.md
candita: https://datatracker.ietf.org/doc/html/rfc1521#page-68 is more current.  Should we go that far?

enhancements/ingress/haproxy-router-compression.md
candita: Fixed.

enhancements/ingress/haproxy-router-compression.md
candita: They need to be quoted, but I would leave that up to the user.  Wdyt?

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,322 @@
...
+
+Enable the configuration of HTTP compression global defaults via the `IngressControllerSpec`, which will
+expose the two environment variables `ROUTER_ENABLE_COMPRESSION` and `ROUTER_COMPRESSION_MIME`.
+
+Enable HTTP compression on selected MIME types, be they wildcard (e.g. `text/*`, `application/*`, etc.) or
+specific (e.g. `text/html`, `text/plain`, `text/css`, `application/json`, `image/svg+xml`, `font/eot`, etc.).
+
+By exposing the environment variables, any HTTP request that terminates within HAProxy, with the header
+**"Accept-Encoding: gzip"**, and with a MIME type that matches one of the configured MIME types will qualify to have
+HTTP compression applied to its payload.  Note 1: this applies only to edge-terminated and reencrypt routes.  Passthrough
```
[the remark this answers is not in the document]
candita: Yes, updated it

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,363 @@
...
+// httpCompressionPolicy turns on compression for the specified MIME types
+//
+// This field is optional, and its absence implies that compression should not be enabled
+// globally in HAProxy.
+//
+// If httpCompressionPolicy exists, compression should be enabled only for the specified
+// MIME types
+type HTTPCompressionPolicy struct {
+	// compressionMIMETypes is a list of MIME types that should have compression applied.
+	// At least one MIME type must be specified.
```
Miciah, line 136:
```suggestion
	// This list can be empty, in which case the ingress controller does not apply compression.
```

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,363 @@
...
+	// +kubebuilder:validation:MinItems=0
+	CompressionMIMETypes []CompressionMIMEType `json:"compressionMIMETypes,omitempty"`
+}
+
+// CompressionMIMEType defines the format of a single MIME type
+// E.g. "text/css;charset=utf-8", "text/html", "text/*", "image/svg+xml",
+// "application/octet-stream", "X-custom/customsub", etc.
+// The format should follow the Content-Type definition in RFC1341 - https://datatracker.ietf.org/doc/html/rfc1341#page-7
+//
+// +kubebuilder:validation:Pattern="^"?(X\-[^ \(\)<>@,;:\\"\/\[\]\?.\=]+|[application|audio|image|message|multipart|text|video])"?\/"?([^ \(\)<>@,;:\\"\/\[\]\?.\=]+(;? *([^ \(\)<>@,;:\\"\/\[\]\?.\=]+=([^ \(\)<>@,;:\\"\/\[\]\?.\=]|".+")+))?)+"?$
```
Miciah, line 154: The quoting with double-quotes looks strange, you don't need to escape things inside brackets, you do need to put any closing bracket first inside brackets (e.g., `[^][ ()<>@,;:"/?.=]`), and you probably want parentheses instead of brackets here: `[application|audio|image|message|multipart|text|video]`.  Do we need to allow double-quotes in the provided MIME type?  Overall, trying to do this validation with a regexp might be more effort than it's worth; it might be better to make things looser, even if that would allow some things that the RFC doesn't strictly allow (such as arbitrary types), but prohibit redundant quoting (if I understand correctly, the yaml syntax might have double-quotes, but the value that is expressed in yaml shouldn't, and the operator should add double-quotes in the generated HAProxy configuration if they are needed there).

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,363 @@
...
+  - that the HTTP compression directive appears in the `defaults` section of the HAProxy template
+  when it is enabled and properly configured on the IngressController spec
+  - that a downgrade to version 4.9 (or version previous to feature implementation version) will not retain the
+  compression configuration of 4.10 (the environment variables should no longer be exposed)
+- E2e tests will validate
+  - that the IngressController spec accepts and enables compression in the `defaults` section of
+  the generated haproxy.cfg
+  - that only requested MIME types receive compression
+
+Note that HAProxy Enterprise **disables** HTTP compression when:
```
Miciah, line 318: We use HAProxy (not HAProxy Enterprise), but I think both have the same behavior.

enhancements/ingress/haproxy-router-compression.md
```diff
@@ -0,0 +1,363 @@
...
+	//
+	// Note: Not all MIME types benefit from compression, but HAProxy will still use resources
+	// to try to compress if instructed to.  Generally speaking, text (html, css, js, etc.)
+	// formats benefit from compression, but formats that are already compressed (pdf, image,
+	// audio, video, etc.) benefit little in exchange for the time and cpu spent on compressing
+	// again. See https://joehonton.medium.com/the-gzip-penalty-d31bd697f1a2
+	//
+	// +required
+	// +kubebuilder:validation:MinItems=0
+	CompressionMIMETypes []CompressionMIMEType `json:"compressionMIMETypes,omitempty"`
```
Miciah, line 146: Maybe we could delete "compression" in the field name to avoid stutter (`spec.httpCompression.compressionMIMETypes` → `spec.httpCompression.mimeTypes`):
```suggestion
	CompressionMIMETypes []CompressionMIMEType `json:"mimeTypes,omitempty"`
```

Miciah (commented):
Thanks!  My remaining questions mainly have to do with the API, so we can discuss those details in the API PR.

/approve
/lgtm

enhancements/ingress/haproxy-router-compression.md
candita: I'm going to go with this pattern: `^"?(X-[^ ()[\]<>@,;:"/?.=]+|(application|audio|image|message|multipart|text|video))"?/"?(.+)+"?$`

PR #890, opened 2021-09-07
Introduce DPU Network Operator

pliurh:
Signed-off-by: Peng Liu <Peng Liu>

### DIFF
--- a/enhancements/network/dpu/dpu-network-operator.md
+++ b/enhancements/network/dpu/dpu-network-operator.md
@@ -1,0 +1,214 @@
+---
+title: DPU Network Operator
+authors:
+  - "@pliurh"
+reviewers:
+  - "@zshi-redhat"
+  - "@dcbw"
+approvers:
+  - 
+creation-date: 2021-09-07
+last-updated: 2021-09-07
+status: implementable
+---
+
+# DPU Network Operator
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+
+## Summary
+
+To facilitate the management of Nvidia BlueField-2 DPU, a two-cluster design is being
+proposed. Under such design, a BlueField-2 card will be provisioned as a worker
+node of the ARM-based infra cluster, whereas the tenant cluster where the normal
+user applications run on, is composed of the X86 servers.
+
+The OVN-Kubernetes components are spread over the two clusters. On the tenant
+cluster side, the Cluster Network Operator is in charge of the management of the
+ovn-kube components. On the infra cluster side, we propose to create a new
+operator to be responsible for the life-cycle management of the ovn-kube components
+and the necessary host network initialization on DPU cards.
+
+## Motivation
+
+Normally, the Cluster Network Operator watches the Network customer resource,
+then provision the ovn-kube components to the Openshift cluster.
+
+In the two-cluster design, some of the ovn-kube components have to run on BF2
+DPUs, but they are still part of the ovn-kube cluster running in the tenant
+cluster. To allow the ovn-kube components on different clusters to talk with
+each other, we need to ensure all the necessary configs and secrets are aligned
+with each other.
+
+### Goals
+
+- Design a new operator that
+  - manages the ovn-kube components running on the DPU that serve the tenant cluster.
+  - enable the switchdev mode on DPUs.
+  - add PF representor to the br-ex bridge, so that the x86 host can connect to
+  the cluster network of the tenant cluster.
+- Define the ovn-kube upgrade mechanism in such a deployment.
+
+### Non-Goals
+
+- Support for network plugins other than OVN-Kubernetes.
+- This enhancement does not address the two-cluster installation.
+- It is possible to have one infra cluster to serve more than one tenant cluster
+in the future. But it is not our initial goal.
+
+## Pre-requisites
+
+- This operator needs to access the apiserver of the tenant cluster, we need to
+  make sure that the pods network in the infra cluster can connect to the
+  apiserver of the tenant cluster in some way.
+- In the infra cluster, one custom MachineConfigPool shall be created for the
+  DPUs on which hardware offloading needs to be enabled.
+
+## Proposal
+
+We will create a namespaced CRD OVNKubeConfig.dpu.openshift.io for the DPU
+OVNKube Operator. One custom resource of this CRD shall be created for the
+tenant cluster in the infra cluster. The operator will take this custom resource
+as input to render and create the objects of the ovn-kube DPU components. The
+operator will also create a MachineConfig CR which injects a systemd service to
+do necessary host configuration on DPUs.
+
+### User Stories
+
+### Risks and Mitigations
+
+## Design Details
+
+The DPU Network Operator will fetch the necessary information from the
+tenant cluster, then create the following objects in the infra cluster:
+
+- The ovnkube-node DaemonSet running in the DPU mode
+- The ServiceAccount and corresponding RBAC definition
+- The ConfigMap that contains ​​ovnkube-config of the tenant cluster
+- The ConfigMap and Secret that contains the root CA and signed certificate of
+  the tenant cluster OVN
+- The ConfigMap that contains the root CA of the tenant cluster API server
+
+The DPU Network Operator will be responsible for the following host
+configuration on the DPUs which we want to enable hardware offloading:
+
+- Add PF representor to the br-ex bridge
+- Enable switchdev mode on DPU nodes
+- Enable OVS hardware offloading: other_config:hw-offload=true
+- Apply udev rules for BF2 interfaces
+
+We assume a custom MachineConfigPool shall be created for all the DPUs, and the
+DPU Network Operator will also create a MachineConfig CR which injects a systemd
+service to do the aforementioned configuration.
+
+### API
+Example configurations:
+
+```yaml
+apiVersion: dpu.openshift.io/v1
+kind: OVNKubeConfig
+metadata:
+  name: cluster
+  namespace: tenant-cluster-1
+spec:
+  kubeConfigFile: tenant-cluster-1-kubeconf
+  poolName: bf2-worker
+status:
+  conditions:
+  - lastTransitionTime: ""
+    status: "True"
+    type: MachineConfigApplied
+  - lastTransitionTime: ""
+    status: "True"
+    type: OvnKubeReady
+  - lastTransitionTime: ""
+    status: "False"
+    type: TenantClusterApiServerUnreachable
+```
+
+1. `kubeConfigFile` stores the secret name of the tenant cluster kubeconfig
+   file. The operator uses this to access the api-server of the tenant
+   cluster.
+2. `poolName` specifies the name of the MachineConfigPool CR which contains all
+   the BF2 nodes in the infra cluster. The operator copies the
+   `spec.nodeSelector` of the MCP to render the ovnkube-node daemonset. A new
+   MachineConfig will be created for this MCP, which injects the systemd
+   service.
+
+### Lifecycle Management
+
+The DPU Network Operator will be delivered as an OLM-managed Operator.
+Users can install this operator from the embedded operator hub of OpenShift,
+after the infra cluster installation is complete.
+
+The OLM shall handle the installation, management, upgrade, and uninstallation
+of this operator.
+
+### Test Plan
+
+TBD
+
+### Graduation Criteria
+
+Graduation criteria follows:
+
+#### Dev Preview -> Tech Preview
+
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, scale)
+- Add CI job at baremetal OCP CI
+- Sufficient time for feedback
+
+#### Removing a deprecated feature
+N/A
+
+### Upgrade / Downgrade Strategy
+No Kubernetes object is used, hence we could say that it is not
+kubernetes-version sensitive.
+
+The version of this operator shall be tied to the version of the ovn-kubernetes
+running in the tenant cluster.
+
+### Version Skew Strategy
+N/A
+
+### API Extensions
+N/A
+
+### Operational Aspects of API Extensions
+N/A
+
+#### Failure Modes
+N/A
+
+#### Support Procedures
+N/A
+
+## Implementation History
+N/A
+
+## Drawbacks
+N/A
+
+## Alternatives
+
+Implement the above-mentioned function in the Cluster Network Operator. Enable
+this feature when the Cluster Network Operator is running in the infra cluster.
+
+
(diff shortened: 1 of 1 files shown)


enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+SmartNIC, but they are still part of the ovn-kube cluster running in the tenant
+cluster. To allow the ovn-kube components on different clusters to talk with
+each other, we need to ensure all the necessary configs and secrets are aligned
+with each other.
+
+### Goals
+
+- Design a new operator that
+  - manages the ovn-kube components that serve the tenant cluster.
+  - enable the swichedev mode on BlueField card.
```
danwinship, line 53:
```suggestion
  - enable the switchdev mode on BlueField card.
```

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+Support for network plugins other than OVN Kubernetes.
+This enhancement does not address the two-cluster installation.
+
+## Pre-requisites
+
+- This operator needs to access the apiserver of the tenant cluster, we need to
+make sure that the pods network in infra cluster can connect to the apiserver of
+the tenant cluster in some way.
+- One custom MachineConfigPool will be created for the BlueField nodes on which
+hardware offloading needs to be enabled.
```
danwinship, line 67: not clear which cluster you're talking about here

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Proposal
+
+We will create a namespaced CRD OVNKubeConfig.smartnic.openshift.io for the
```
danwinship, line 71: figure out naming: "smartnic" vs "dpu"

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+DPU OVNKube Operator. One custom resource of this CRD shall be created for
+the tenant cluster. The operator will take this custom resource as input to
+render and create the objects of ovnkube smart-nic components. The operator will
+also create a MachineConfig CR which injects a systemd service to enable
+switchdev mode on BF-2 node.
+
+### User Stories
+
+As a cluster admin I want to manage the ovn-kubernetes components in the infra
+cluster, and these components can be controlled by the ovn-kubernetes control plane in the tenant cluster.
```
danwinship, line 81: You can just delete the user stories section. This doesn't add any information beyond what's already in Summary/Motivation/Goals.

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Risks and Mitigations
+
+## Design Details
+
+The DPU OVNKube Operator will fetch the necessary information from the
+tenant cluster, then create the following objects in the infra cluster:
+- The ovnkube-node DaemonSet running in the SmartNIC mode
```
danwinship, line 89: "DPU mode"

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+- The ServiceAccount and corresponding RBAC definition
+- The ConfigMap that contains ​​ovnkube-config of the tenant cluster
+- The ComfgMap and Secret that contains the CA and signed certificate of the
```
danwinship, line 92:
```suggestion
- The ConfigMap and Secret that contains the CA and signed certificate of the
```

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+- tenant cluster OVN
+- The ConfigMap CA and signed certificate of the tenant cluster API server
```
danwinship, line 94: what "signed certificate" exactly? (In the OVN case there's only a single cert used by all clients, but I assume the apiserver either has per-node certs, or else doesn't use client certificates.)

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+
+The DPU OVNKube Operator will be responsible for enable the switchdev mode on
+BF2 cards. We assume all the BF2 cards will be put into a custom
+MachineConfigPool,
```
danwinship, line 98: comma should be a period, or else there's something missing at the end?

Who is responsible for the custom MachineConfigPool? When are they supposed to create it?

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+
+Currently, we are only targeting the 1to1 deployment mode, meaning that
+there is 1to1 mapping between the infra cluster and the tenant cluster. However,
+it is possible to have one infra cluster to serve more than one tenant cluster
+in the future. We may want to create one custom resource in a different
+namespace for each tenant cluster.
```
danwinship, line 104: maybe move this to Non-Goals?

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+spec:
+  kubeConfigFile: tenant-cluster-1-kubeconf
+  poolName: bf2-worker
+```
+
+1. `kubeConfigFile` stores the secret name of the tenant cluster kubeconfig
+   file. The operator uses this to access the api-server of the tenant
+   cluster.
+2. `poolName` specify the name of the MachineConfigPool CR which contains all
+   the BF2 nodes. The operator inherits the `spec.nodeSelector` of the MCP to
```
danwinship, line 124: again clarify which cluster?

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+   render the ovnkube-node daemonset.
```
danwinship, line 125: "inherits" implies some hierarchical relationship. I think you just mean that the operator "uses" or "copies" the `spec.nodeSelector` of the MCP?

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,186 @@
...
+make sure that the pods network in infra cluster can connect to the apiserver of
+the tenant cluster in some way.
+- One custom MachineConfigPool will be created for the BlueField nodes on which
+hardware offloading needs to be enabled.
+
+## Proposal
+
+We will create a namespaced CRD OVNKubeConfig.smartnic.openshift.io for the
+DPU OVNKube Operator. One custom resource of this CRD shall be created for
+the tenant cluster. The operator will take this custom resource as input to
```
danwinship, line 73: it's created "for the tenant cluster", but it's created _in_ the infra cluster, which again probably needs to be clarified.

How is the admin expected to create this resource? We should probably provide at least a script or something to grab the necessary information from the tenant cluster and create the resource in the infra cluster...

enhancements/network/dpu-ovnkube-operator.md
pliurh: > Who is responsible for the custom MachineConfigPool? When are they supposed to create it?

That is a good question. In my assumption, the cluster-admin shall label the nodes and create the MCP. WDYT?

enhancements/network/dpu-ovnkube-operator.md
pliurh: No signed certificate here, it should be the root CA of the tenant cluster apiserver.

enhancements/network/dpu-ovnkube-operator.md
pliurh: `User Stories` charpter is required by the markdownlint of this repo, so I just remove the content of it.

enhancements/network/dpu-ovnkube-operator.md
pliurh: I don't quite get the question. This operator only needs the kubeconfig of the tenant cluster, then it can access the apiserver to grab all the info. Do you mean we need a script to generate the kubeconfig file for the tenant cluster?

In my assumption, the whole process shall be like this:

1. After infra cluster installation, the admin labels the DPUs which he wants to enable hardware offload.
2. The admin creates a custom MCP that contains all the selected DPUs.
3. The admin configures the CNO to deprovision the ovn-kube components from the DPUs.
4. The admin installs the DPU operator and creates an OVNKubeConfig CR with the `poolName` but leaves `kubeConfigFile` blank. The operator then does the host network configuration.
5. Install the tenant cluster. Scale up the x86 workers with DPU.
6. The CNO of the tenant cluster puts those x86 hosts into smart-nic host mode.
7. The admin stores the kubeconfig file in the infra cluster then adds `kubeConfigFile` to the existing OVNKubeConfig CR. The operator installs the ovn-kube components on the selected DPUs.

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,200 @@
...
+do necessary host configuration on DPUs.
+
+### User Stories
+
+### Risks and Mitigations
+
+## Design Details
+
+The DPU OVNKube Operator will fetch the necessary information from the
+tenant cluster, then create the following objects in the infra cluster:
```
zshi-redhat, line 90: Shall we add that OVNKube operator will watch those runtime ovn-k8s changes in tenant cluster and reflect that accordingly in the infra clusters? Like whenever the configMap or Secret are changed in tenant cluster.

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,200 @@
...
+operator will also create a MachineConfig CR which injects a systemd service to
+do necessary host configuration on DPUs.
+
+### User Stories
+
+### Risks and Mitigations
+
+## Design Details
+
+The DPU OVNKube Operator will fetch the necessary information from the
```
zshi-redhat, line 89: Does this Operator need to handle kube API compatibility between infra and tenant clusters?

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,200 @@
...
+with each other.
+
+### Goals
+
+- Design a new operator that
+  - manages the ovn-kube components that serve the tenant cluster.
+  - enable the switchdev mode on DPUs.
+  - add PF representor to the br-ex bridge, so that the x86 host can connect to
+  the cluster network of the tenant cluster.
+- Define the ovn-kube upgrade mechanism in such a deployment.
```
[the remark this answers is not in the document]
zshi-redhat: For example, should this OVNKube operator fetch the image information from tenant cluster as well and use the image to spin up ovnkube-node in infra cluster?

enhancements/network/dpu-ovnkube-operator.md
pliurh: Is that necessary? I assume both clusters shall in the same version.

enhancements/network/dpu-ovnkube-operator.md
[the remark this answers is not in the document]
pliurh: Can we assume the two clusters will always be in the same version, and upgrade together? Then we can let the operator use the images which are currently used in the infra cluster.

enhancements/network/dpu-ovnkube-operator.md
pliurh: Yeah, it's a good point. AFAIK, those ovnkube objects can not be changed after cluster installation because CNO doesn't allow them to. But the certificates do expire, we need to watch the change in the tenant cluster.

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,202 @@
...
+The OVN-Kubernetes components are spread over the two clusters. On the tenant
+cluster side, the Cluster Network Operator is in charge of the management of the
+ovn-kube components. On the infra cluster side, we propose to create a new
+operator to be responsible for the life-cycle management of ovn-kube components
+and the necessary host network initialization on DPU cards.
+
+## Motivation
+
+Normally, the Cluster Network Operator watches the Network customer resource,
+then provision the ovn-kube components to the Openshift cluster.
```
Billy99, line 42: Normally CNO manages the ovn-kube components. If this new operator is taking over that role for the infra cluster, do we need to document somewhere that CNO needs to be changed to not manage OVN-Kube on infra clusters and what's the CNO trigger for that?

enhancements/network/dpu-ovnkube-operator.md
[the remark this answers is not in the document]
zshi-redhat: I think we need to define the version compatibility between infra and tenant cluster, like what is supported and not supported.  For example, we may say that 4.x tenant is compatible with 4.x+1 infra or 4.x+n infra.

This would allow new feature be introduced without updating tenant cluster. for example, the update of rhcos image in infra cluster (openvswitch, driver etc) may bring in new features, Operator designed to run on infra cluster would also benefit from it.

Thinking loudly, if we were to move ovn-central to infra cluster, it would further decouple the infra and tenant in terms of cluster or ovnkube versions. which allow more features be added without updating tenant cluster.

enhancements/network/dpu-ovnkube-operator.md
zshi-redhat: If we allow tenant and infra to be in different cluster version (which I think we should if not in the initial delivery), then yes.

enhancements/network/dpu-ovnkube-operator.md
zshi-redhat: One example I had in mind was the ipsec enablement, it can be done after initial cluster installation and ovnkube Operator will need to rendering the ipsec daemonset in infra cluster.

enhancements/network/dpu-ovnkube-operator.md
pliurh: This operator is not taking take over that role for the infra cluster, but managing the ovnkube that servers the tenant cluster. The CNO will still manage the ovnkube of the infra cluster on non-BF2 nodes.

@zshi-redhat Shall we mention these CNO changes in https://github.com/openshift/enhancements/pull/732

zshi-redhat: I think we need a similar patch in CNO like this https://github.com/openshift/cluster-network-operator/pull/1138, to isolate the CNO managed ovnkube from being scheduled on DPU in infra cluster. For example, to introduce the label `network.operator.openshift.io/dpu` which will be used by CNO in infra cluster.

zshi-redhat: @pliurh yes, THE #732 needs to be updated.

enhancements/network/dpu-ovnkube-operator.md
```diff
@@ -0,0 +1,202 @@
...
+## Alternatives
+
+Implement the above-mentioned function in the Cluster Network Operator. Enable
+this feature when the Cluster Network Operator is running in the infra cluster.
+
+
+## Infrastructure Needed
+
+One X86 bare-metal cluster with BlueField-2 DPU installed on worker nodes, and
+one ARM bare-metal cluster.
```
bn222, line 202: I would explicitly that the ARM bare-metal cluster does not have any separate bare-metal worker nodes.

pliurh:
> can you move this into the `enhancements/network/dpu/` directory (which should exist now in master)?
> 
> you may need to add some more dummy sections to make the latest markdownlint job happy too...

Done

PR #894, opened 2021-09-10
authentication/login-logout: initial enhancement

s-urbaniak:


enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,159 @@
...
+
+4. **Request Header identity provider**
+
+Login events can easily be constructed by creating audit events in the [central request authentication method](https://github.com/openshift/oauth-server/blob/4e63f0f350f28171edc4ae2553c7b852e65efbc6/pkg/authenticator/request/headerrequest/requestheader.go#L34).
+
+### Logout events
+
+Logout events are not visible to oauth-server as the infrastructure in OpenShift does not have a central logout handler mechanism. Instead we have to rely on audit events against deletion of `oauthaccesstoken` resources.
+
+It is an open question how we can add information about the underlying user of the deleted `oauthaccesstoken` resource.
```
s-urbaniak, line 139: cc @stlaz any suggestions/ideas? :thinking:

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,159 @@
...
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement describes capturing login, login failure, and logout events in the OpenShift authentication subsystems.
+
+Currently, login and logout events can only be derived from inspecting audit logs. These include creation or deletion of `oauthaccesstokens` resources as described in the API server [Audit Policy Configuration](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/audit-policy.md#logging-of-token-with-secure-oauth-storage).
+
+This approach has a couple of drawbacks:
+1. Login events are captured via creation events of `oauthaccesstokens` in audit logs. This includes successful login events only, not login failures.
```
stlaz, line 38: note, not necessarily actionable: you might be able to audit certain login failures where the client fails to use the `oauthauthorizetoken` by the token timing out

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,159 @@
...
+
+### As a user I want to inspect logout events
+
+Use `oc adm must-gather -- /usr/bin/gather_audit_logs` and inspect apiserver audit logs for deleted `oauthaccesstoken` resources.
+
+### Implementation Details/Notes/Constraints [optional]
+
+What are the caveats to the implementation? What are some important details that
+didn't come across above. Go in to as much detail as necessary here. This might
+be a good place to talk about core concepts and how they relate.
```
stlaz, line 107: remove this paragraph?

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,159 @@
...
+
+	// "WriteLoginEvents" logs login and login failure events.
+  // This is the default.
+	OAuthWriteLoginEventsProfileType AuditProfileType = "WriteLoginEvents"
+)
+```
+
+### User Stories
+
+### As a user I want to inspect login and login failure attempts
```
stlaz, line 95: lang nit: `As a user I want to inspect successful and failed logins`
(I think "login failure attempt" might technically make sense when bruteforcing but I think you meant the above instead :slightly_smiling_face: )

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,159 @@
...
+
+Login failure however cannot be easily detected as there is no guarantee that external oauth providers will invoke callback in the case of a login failure. Hence, only best-effort login failure events can be emitted, if the external OAuth provider provides an `error` response in the [central `HandleRequest` handler](https://github.com/openshift/oauth-server/blob/513a8bb5b6cbd5e8faacbed523eb861aa29a674b/vendor/github.com/RangelReale/osincli/authorize.go#L84).
+
+3. **External OAuth identity providers used as challengers**
+
+External oauth identity providers used as challengers use a [central location](https://github.com/openshift/oauth-server/blob/822478f88514a80f053d9f65cd689d981b6ea4fd/pkg/oauth/external/handler.go#L118) where login and login failures audit events will be emitted.
+
+4. **Request Header identity provider**
+
+Login events can easily be constructed by creating audit events in the [central request authentication method](https://github.com/openshift/oauth-server/blob/4e63f0f350f28171edc4ae2553c7b852e65efbc6/pkg/authenticator/request/headerrequest/requestheader.go#L34).
```
stlaz, line 133: Do you plan to log login failures here anyhow? The situation here is similar to the one in external OAuth identity providers except the login proxy/3rd party login provider will never reach the oauth-server if the login fails there. Probably worth mentioning.

enhancements/authentication/login-logout-events.md
stlaz: I'll have to check the audit code for that, we'll probably need some custom code that would add the user information to the audit event. I'm not sure we can annotate the audit event somehow or whether this information can somehow be injected into a custom field of the event somehow.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,159 @@
...
+
+In all cases above a [Info](https://github.com/openshift/oauth-server/blob/961295a3151da3fff9daa5ea0ab8c0bf92a7d7e6/vendor/k8s.io/apiserver/pkg/authentication/user/user.go#L20) entity is available with user information which can be used to fill audit event data.
+
+### Integration in must-gather
+
+The existing `oc adm must-gather -- /usr/bin/gather_audit_logs` tool will also fetch oauth-server audit logs.
+
+### Open Questions
+
+An initial idea was to add annotations during `oauthaccesstoken` creation and deletion to indicate what underlying user is being logged out. However, that information is already included [in the `oauthaccesstoken` type](https://github.com/openshift/oauth-apiserver/blob/958c914e2b0b2d58f9b51715b2b76aaae774d72b/pkg/oauth/apis/oauth/types.go#L28).
```
stlaz, line 151: the information is missing during deletion though, isn't it?

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Design Details
+
+### Login and Login failure events
+
+To create login and login failure audit events differ slightly depending on the underlying identity provider. There are the following cases to consider:
+
+1. **Password-based identity providers**
+
+All password-based identity providers are using a [central location](https://github.com/openshift/oauth-server/blob/690499e76a0b242adb5ccc73f23cccc8a50b8788/pkg/server/login/login.go#L173) where login events are handled. Both login success and failure events will be emitted here.
```
stlaz, line 117: How do you plan to avoid a DOS attack where your disk resources would be depleted by an attacker forcing failure attempts audit in a loop? The oauth-server currently does not currently implement login attempt limits and adding it might be hard given we don't see the IP of the user but rather the one of the ingress.

enhancements/authentication/login-logout-events.md
s-urbaniak: oh, right, removed

enhancements/authentication/login-logout-events.md
s-urbaniak: yes, that's a great suggestion :+1:  fixed

enhancements/authentication/login-logout-events.md
s-urbaniak: > except the login proxy/3rd party login provider will never reach the oauth-server if the login fails there.

precisely, I swear I wanted to write this down but that information is definitely needed.

Fixed

enhancements/authentication/login-logout-events.md
s-urbaniak: The object does include the necessary information during the in-flight deletion request, it is just not fully printed as it is done in the case of a creation event.

enhancements/authentication/login-logout-events.md
s-urbaniak: I agree, and I believe we briefly discussed this in the arch call as well. AFAIK the current audit event logging subsystem does not have rate limiting in general, I will add this as future work and as a non-goal.

s-urbaniak:
@stlaz addressed comments, where feasible, ptal.

enhancements/authentication/login-logout-events.md
stlaz: so we would need to find a way to add it

s-urbaniak: as discussed, crazy idea: add an in-process admission plugin into oauth-apiserver which intercepts deletions of `oauthaccesstoken` resources and creates a dedicated "logout" event with all necessary information. I will play around with this idea and see if that works.

enhancements/authentication/login-logout-events.md
s-urbaniak: as discussed, we have that already "for free" as part of the apiserver default handler chain: https://github.com/openshift/oauth-server/blob/961295a3151da3fff9daa5ea0ab8c0bf92a7d7e6/vendor/k8s.io/apiserver/pkg/server/config.go#L807

I will add this as a paragraph in the OEP.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,164 @@
...
+### Logout events
+
+Logout events are not visible to oauth-server as the infrastructure in OpenShift does not have a central logout handler mechanism. Instead we have to rely on audit events against deletion of `oauthaccesstoken` resources.
+
+It is an open question how we can add information about the underlying user of the deleted `oauthaccesstoken` resource.
+
+### Event construction
+
+In all cases above a [Info](https://github.com/openshift/oauth-server/blob/961295a3151da3fff9daa5ea0ab8c0bf92a7d7e6/vendor/k8s.io/apiserver/pkg/authentication/user/user.go#L20) entity is available with user information which can be used to fill audit event data.
+
```
EmilyM1, line 143: @s-urbaniak  Just wanted to clairfy: for the logging of failed logins, will the output be:
1. a new event in audit logs
2. an entirely new log from apiserver
3. both
4. other
What logs will we have this event in basically?

enhancements/authentication/login-logout-events.md
s-urbaniak: this has been resolved and the solution is way easier than assumed :)

enhancements/authentication/login-logout-events.md
s-urbaniak: @EmilyM1 it is option 1. "a new event in audit logs"

EmilyM1: Ok, so they are independent. Oauth-server log will not go thru the kube-apiserver which currently outputs audit logs?
So no changes will be needed to kube-apiserver here to enable oauth-server to output audit logs?

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,213 @@
...
+	// templates allow you to customize pages like the login page.
+	// +optional
+	Templates OAuthTemplates `json:"templates"`
+
+	Audit OAuthAudit `json:"audit"`
+}
+
+type OAuthAudit struct {
+	Profile OAuthAuditProfileType `json:"profile,omitempty"`
```
sttts, line 79: Would set the default via kubebuilder marker. It makes it discoverable. Also don't forget to describe the possible values and add an enum marker.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,213 @@
...
+
+2. **External OAuth identity providers**
+
+For login events external OAuth providers must invoke oauth-server callback handler to finalize authorization using [code grant flow](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1).
+
+Login success will be emitted upon successful code exchange with oauth-server in the [central OAuth callback handler](https://github.com/openshift/oauth-server/blob/822478f88514a80f053d9f65cd689d981b6ea4fd/pkg/oauth/external/handler.go#L149-L157). 
+
+Login failure however cannot be easily detected as there is no guarantee that external oauth providers will invoke callback in the case of a login failure. Hence, only best-effort login failure events can be emitted, if the external OAuth provider provides an `error` response in the [central `HandleRequest` handler](https://github.com/openshift/oauth-server/blob/513a8bb5b6cbd5e8faacbed523eb861aa29a674b/vendor/github.com/RangelReale/osincli/authorize.go#L84).
+
+3. **External OAuth identity providers used as challengers**
```
ibihim, line 125: I didn't get that one here, but it might just me. The only thing I could find with regards to "challenge" in the given context was referring to PKCE.

enhancements/authentication/login-logout-events.md
ibihim: I don't know if it is the same thought as mine, but sounds similar: every unfinished login attempt is a login attempt failure. The reasoning remains unclear.
But the implementation would increase significantly and might be a "non goal"?

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,213 @@
...
+
+1. **Password-based identity providers**
+
+All password-based identity providers are using a [central location](https://github.com/openshift/oauth-server/blob/690499e76a0b242adb5ccc73f23cccc8a50b8788/pkg/server/login/login.go#L173) where login events are handled. Both login success and failure events will be emitted here.
+
+2. **External OAuth identity providers**
+
+For login events external OAuth providers must invoke oauth-server callback handler to finalize authorization using [code grant flow](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1).
+
+Login success will be emitted upon successful code exchange with oauth-server in the [central OAuth callback handler](https://github.com/openshift/oauth-server/blob/822478f88514a80f053d9f65cd689d981b6ea4fd/pkg/oauth/external/handler.go#L149-L157).
```
stlaz, line 121: should be "a sucessful code exchange with the OIDC IdP" I believe, or "between the oauth-server and the OIDC IdP"

enhancements/authentication/login-logout-events.md
stlaz: "challenger" is a specific terminology to the oauth-server, it is an IdP posting HTTP challenges (in this specific case the Basic-Auth HTTP Challenge). This section is basically the same as `1. Password-based identity providers`

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,213 @@
...
+
+Login failure however cannot be easily detected as there is no guarantee that external oauth providers will invoke callback in the case of a login failure. Hence, only best-effort login failure events can be emitted, if the external OAuth provider provides an `error` response in the [central `HandleRequest` handler](https://github.com/openshift/oauth-server/blob/513a8bb5b6cbd5e8faacbed523eb861aa29a674b/vendor/github.com/RangelReale/osincli/authorize.go#L84).
+
+3. **External OAuth identity providers used as challengers**
+
+External oauth identity providers used as challengers use a [central location](https://github.com/openshift/oauth-server/blob/822478f88514a80f053d9f65cd689d981b6ea4fd/pkg/oauth/external/handler.go#L118) where login and login failures audit events will be emitted.
+
+4. **Request Header identity provider**
+
+For this case an established trust to the external login proxy exists via means of client certificate trust. This implies that all authentication requests, naturally, have already passed successful authentication in the external 3rd party proxy.
```
stlaz, line 131: it is a mutual TLS trust, the oauth-server trusts the proxy as it's capable to identify it by the CA cert it was supplied with, the proxy should trust the oauth-server as it should trust the OpenShift certs.

I wouldn't say the mTLS trust implies that the authentication requests passed successful authentication in the 3rd party proxy, the successful authentication there is assumed.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,213 @@
...
+		"userName": "exampleuser",
+	  },
+      "userUID": "123",
+      "authorizeToken": "sha256~sometoken"
+    },
+...
+  }
+```
+
+### Event construction
```
stlaz, line 193: Maybe split the above section into `oauth-server` and `oauth-apiserver` subsections so that it's clear which component is being described here?

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,213 @@
...
+
+In all cases above a [Info](https://github.com/openshift/oauth-server/blob/961295a3151da3fff9daa5ea0ab8c0bf92a7d7e6/vendor/k8s.io/apiserver/pkg/authentication/user/user.go#L20) entity is available with user information which can be used to fill audit event data.
```
stlaz, line 195: Can you provide an example of what such an event should look like, then?

enhancements/authentication/login-logout-events.md
ibihim: So what in particular is a challenge? The state that is returned to the callback endpoint? Do you have a link to look it up?

EmilyM1: This is something we add on oauth-server side right? If so, is it similar to PKCE to ensure the IDP has a token we have generated.

enhancements/authentication/login-logout-events.md
s-urbaniak: > Oauth-server log will not go thru the kube-apiserver which currently outputs audit logs?
correct

> So no changes will be needed to kube-apiserver here to enable oauth-server to output audit logs?
correct

enhancements/authentication/login-logout-events.md
s-urbaniak: @ibihim the code location is documented in this OEP, see the first paragraph (https://github.com/openshift/oauth-server/blob/822478f88514a80f053d9f65cd689d981b6ea4fd/pkg/oauth/external/handler.go#L118). If you look at the caller it simply takes credentials from basic auth as @stlaz explained: https://github.com/openshift/oauth-server/blob/822478f88514a80f053d9f65cd689d981b6ea4fd/pkg/authenticator/request/basicauthrequest/basicauth.go#L27

It effectively is https://oauth.net/2/grant-types/password/

s-urbaniak: @EmilyM1 no, PKCE extends code grant flow to make it usable in public clients.

enhancements/authentication/login-logout-events.md
stlaz: I don't see how this could be a non-goal. This is already working, it's just a matter of documenting the behavior - `if an oauthauthorizetoken has not been exchanged for an oauthaccesstoken during the lifetime of an oauthauthorizetoken there were likely issues in the login process`. All of the information mentioned is discoverable from the audit logs.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,213 @@
...
+### As an administrator I want to inspect successful login and login failure attempts
+
+Use `oc adm must-gather -- /usr/bin/gather_audit_logs` and inspect oauth-server audit logs.
+
+### As an administrator I want to inspect logout events and the deleted user
+
+Use `oc adm must-gather -- /usr/bin/gather_audit_logs` and inspect apiserver audit logs for deleted `oauthaccesstoken` resources.
+Audit log entries include the deleted user.
+
+### As an administrator I want to inspect the source IPs
```
stlaz, line 105: This is an incomplete sentence. Source IPs of what?

enhancements/authentication/login-logout-events.md
ibihim: Good catch

enhancements/authentication/login-logout-events.md
ibihim: How can I find that out? `must-gather` and check some similar audit logs and exchange values for what we are expecting?

enhancements/authentication/login-logout-events.md
ibihim: Nice detail.

enhancements/authentication/login-logout-events.md
ibihim: You mean we can deduce it.. Yes, this is definitely true.

enhancements/authentication/login-logout-events.md
ibihim: Yes, the very first paragraph refers the `oauth-server` and the switch to the `oauth-apiserver` is not explicit.

ibihim:
I hope I addressed all comments. PTAL @s-urbaniak , @stlaz , @sttts , @EmilyM1

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,226 @@
...
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement describes capturing login, login failure, and logout events in the OpenShift authentication subsystems.
+
+Currently, login and logout events can only be derived from inspecting audit logs. These include creation or deletion of `oauthaccesstokens` resources as described in the API server [Audit Policy Configuration](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/audit-policy.md#logging-of-token-with-secure-oauth-storage).
+
+This approach has a couple of drawbacks:
+1. Login events are captured via creation events of `oauthaccesstokens` in audit logs. This includes successful login events only, but not all login failures. Login failures can be only deduced by comparing which `oauthauthorizetokens` were created and which were not exchagend for an `oauthaccesstoken`.
```
stlaz, line 38: `Login failures can be only deduced by comparing which oauthauthorizetokens were created and which were not exchagend for an oauthaccesstoken`
->
 `Currently the only login failure that can be deduced from the audit logs is the inability of exchanging the oauthauthorizetoken for an oauthaccesstoken as we log creation of both of these objects`

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,226 @@
...
+	// +listType=atomic
+	IdentityProviders []IdentityProvider `json:"identityProviders,omitempty"`
+
+	// tokenConfig contains options for authorization and access tokens
+	TokenConfig TokenConfig `json:"tokenConfig"`
+
+	// templates allow you to customize pages like the login page.
+	// +optional
+	Templates OAuthTemplates `json:"templates"`
+	// audit specifies what should be audited in the context of OAuthServer.
```
stlaz, line 75: I believe this must be +optional as it's going to be missing during upgrades

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,308 @@
...
+
+All components will be developed with the concern of version skew in mind. So they will handle the missing of the `Audit` property.
+
+### Test Plan
+
+Ideally we test our feature with:
+
+- unit tests,
+- e2e tests in origin and
+- tests at must-gather.
```
ibihim, line 261: Note to future devs: https://github.com/openshift/origin/blob/master/test/extended/cli/mustgather.go

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,308 @@
...
+### Version Skew Strategy
+
+All components will be developed with the concern of version skew in mind. So they will handle the missing of the `Audit` property.
+
+### Test Plan
+
+Ideally we test our feature with:
+
+- unit tests,
+- e2e tests in origin and
```
ibihim, line 260: Note to future devs: https://github.com/openshift/origin/blob/master/test/extended/cluster/audit.go

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,330 @@
...
+
+Use `oc adm must-gather -- /usr/bin/gather_audit_logs` and inspect oauth-server audit logs.
+
+#### As an administrator I want to inspect logout events and the deleted user
+
+Use `oc adm must-gather -- /usr/bin/gather_audit_logs` and inspect apiserver audit logs for deleted `oauthaccesstoken` resources.
+Audit log entries include the deleted user.
+
+#### As an administrator I want to inspect the source IPs of the client triggering an authentication event
+
```
s-urbaniak, line 117: please add how this use case is implemented. namely: Using `oc adm must-gather -- /usr/bin/gather_audit_logs` and inspecting oauth-server audit logs one can see the source IP of the client.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,330 @@
...
+### API Extensions
+
+We introduce a new property called `audit` to the `OAuth` CRD.
+
+This enhancement is not expected to affect any other team or compoment outside of auth.
+
+### Operational Aspects of API Extensions
+
+There is no explicit operational involvement expected. The property is set by default and optional.
```
s-urbaniak, line 126: nit:"and is optional"

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,330 @@
...
+
+It could be turned of by operational interaction with the API extension.
```
s-urbaniak, line 128: nit: "turned off"

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,330 @@
...
+
+#### Support Procedures
+
+If the `audit` can't be set in the CR, support would need to file a bugzilla.
+If there are no audit events for login, login failures or logout events, verify that the `audit` values is set in the `oauth` CR.
+If the `audit` is set to `profile: "WriteLoginEvents"` and there are no audit events for the oauth-server after the login, login failure or logout events, support would need to file a bugzilla.
+
+### Risks and Mitigations
+
+There are no big risks. As with logging, we could always log something that we shouldn't and violate compliance.
```
s-urbaniak, line 142: nit: "no known big risks".

s-urbaniak: actually, there are risks:
- DoS like situations where we have a lot of incoming requests issuing many audit logging events, see https://github.com/openshift/enhancements/pull/894#discussion_r707216430 for details.
- Disk capacity protection: audit log rotation, please find the bits in the apiserver audit logging framework.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,330 @@
...
+
+### Event construction
+
+In all cases above an [Info](https://github.com/openshift/oauth-server/blob/961295a3151da3fff9daa5ea0ab8c0bf92a7d7e6/vendor/k8s.io/apiserver/pkg/authentication/user/user.go#L20) entity is available with user information which can be used to fill audit event data.
+
+### Integration in must-gather
+
+The existing `oc adm must-gather -- /usr/bin/gather_audit_logs` tool will also fetch oauth-server audit logs.
+
+### DoS mitigation
```
s-urbaniak, line 244: can be merged in the "Risks and mitigations" chapter, see comment above.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,330 @@
...
+
+Currently, the default configuration of apiserver's `MaxRequestsInFlight` (i.e. GET requests) and `MaxMutatingRequestsInFlight` (i.e. POST requests) are used as part of the [default configuration](https://github.com/openshift/oauth-server/blob/961295a3151da3fff9daa5ea0ab8c0bf92a7d7e6/vendor/k8s.io/apiserver/pkg/server/config.go#L330-L331) to protect against audit event flooding.
+
+### Graduation Criteria
+
+This feature is planned for OpenShift 4.10.
+
+#### Dev Preview -> Tech Preview
+
+- Audit logging should work.
```
s-urbaniak, line 256: No Tech Preview is planned.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,330 @@
...
+
+#### Tech Preview -> GA
+
+- No sensitive data is logged.
+- Verbosity should be acceptable.
```
s-urbaniak, line 261: No Tech Preview planned, hence N/A

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,330 @@
...
+
+#### E2E testing steps
+
+In order to test the audit logging as mentioned in the porposal:
+
+**Verify audit logging**
+
+1. Create a cluster with OpenShift installed on it.
+2. Verify that the version deployed contains this enhancement.
+3. Verify that the custom resource `OAuth` exists and contains the properties:
```
s-urbaniak, line 293: `oauths.config.openshift.io`

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,330 @@
...
+
+N/A
+
+## Drawbacks
+
+We create more log entries, which increases the space that the logs will occupy.
+
+## Alternatives
+
+enhancements/authentication/login-logout-events.md missing "#### Support Procedures" he alternative is to keep on relying on `oauthaccesstoken` audit events with the aforementioned drawbacks.
```
s-urbaniak, line 330: i think my editor went bonkers. this should be:
> The alternative is to keep on relying on `oauthaccesstoken` audit events with the aforementioned drawbacks.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,330 @@
...
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+As we are introducing a new property with default values, there are not expected risks for upgrading.
+
+Downgrading would not cause any data loss or inconsistencies as the property is a simple feature flag to toggle audit logging.
```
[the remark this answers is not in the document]
sttts: Deserialization after downgrade will not fail. The apiservrer will just prune the fields in the request. The data stays in etcd if it is not rewritten. I think for this kind of feature via a new field, we can add it in `n` without risk.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,335 @@
...
+authors:
+  - "@s-urbaniak"
+reviewers:
+  - "@mfojtik"
+  - "@deads2k"
+  - "@stlaz"
+  - "@sttts"
+approvers:
+- "@Anandnatraj"
+- "@oscardoe"
```
s-urbaniak, line 12: nit: replace @oscardoe with @mfojtik

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,335 @@
...
+Use `oc adm must-gather -- /usr/bin/gather_audit_logs` and inspect apiserver audit logs for deleted `oauthaccesstoken` resources.
+Audit log entries include the deleted user.
+
+#### As an administrator I want to inspect the source IPs of the client triggering an authentication event
+
+Use `oc adm must-gather -- /usr/bin/gather_audit_logs` and inspect oauth-server audit logs for auth events and inspect the IP within the event.
+
+### API Extensions
+
+We introduce a new property called `audit` to the `OAuth` CRD.
```
stlaz, line 122: `OAuth` CRD -> `oauth.config.openshift.io` type

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,335 @@
...
+
+This enhancement is not expected to affect any other team or compoment outside of auth.
+
+### Operational Aspects of API Extensions
+
+There is no explicit operational involvement expected. The property is set by default and is optional.
```
stlaz, line 128: The property is set <to do what> by default?

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,335 @@
...
+
+It could be turned off by operational interaction with the API extension.
```
stlaz, line 130: what's an operational interaction?

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,335 @@
...
+
+#### Failure Modes
+
+N/A
+
+#### Support Procedures
+
+If the `audit` can't be set in the CR, support would need to file a bugzilla.
+If there are no audit events for login, login failures or logout events, verify that the `audit` values is set in the `oauth` CR.
```
stlaz, line 139: The `audit` property does not influence logging of logout events

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,335 @@
...
+If the `audit` is set to `profile: "WriteLoginEvents"` and there are no audit events for the oauth-server after the login, login failure or logout events, support would need to file a bugzilla.
```
stlaz, line 140: ditto on the logout events

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,335 @@
...
+1. Create a cluster with OpenShift installed on it.
+2. Verify that the version deployed contains this enhancement.
+3. Verify that the custom resource `oauths.config.openshift.io` exists and contains the properties:
+
+```yaml
+audit:
+    profile: WriteLoginEvents
+```
+
+4. Create Login failures, Login and Logout events.
```
stlaz, line 305: no logout events?

stlaz (changes requested):

enhancements/authentication/login-logout-events.md
ibihim: Isn't this the goal of the kubebuilder setup?

enhancements/authentication/login-logout-events.md
ibihim: hahaha, wetware in action :smile:

enhancements/authentication/login-logout-events.md
ibihim: It doesn't? What is then the whole purpose of setting `audit`?

Will the `audit` property not tell the `cluster-authentication-operator` to enable audit logging on `oauth-server`?

enhancements/authentication/login-logout-events.md
ibihim: I am mentioning logout events: 
>  or logout events

ibihim: Or do you mean that I should remove the `logout` entry?

enhancements/authentication/login-logout-events.md
ibihim: Shouldn't I mention `Logout events`?

enhancements/authentication/login-logout-events.md
ibihim: Very useful comment.

enhancements/authentication/login-logout-events.md
stlaz: As far as I know, it only influences login events, and also logs some login failures. Hence the `WriteLoginEvents` value that you can configure for the profile.

ibihim: But isn't the goal to turn the audit logging on **and then** add the logging of the other events too?

So there will be a certain moment in time, before we add more logging events to the `oauth-server`, where we will only log a very limited set of events. But once we adjusted `oauth-server`, it will log login, login failures and logout... oh noes... you are right. That is true. The logout events are logged by extending what is being logged on logout and not in particular by the profile. So we have different technical paths to solve the goal.

Thanks a lot for the insight and thorough review :D This enhancement helped me to gain so much more clarity on that topic.

s-urbaniak: @ibihim yes, logout events cannot be captured in oauth-server

s-urbaniak: @ibihim as described in this OEP, logout events are captured only via deletion of oauthaccesstokens, subject auf audit policy configuration in oauth-apiserver.

enhancements/authentication/login-logout-events.md
s-urbaniak: @ibihim yes

enhancements/authentication/login-logout-events.md
ibihim: Created a dedicated entry for the logout events.

enhancements/authentication/login-logout-events.md
ibihim: Ok, I got it.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,344 @@
...
+
+### API Extensions
+
+We introduce a new property called `audit` to the `oauth.config.openshift.io` type.
+
+This enhancement is not expected to affect any other team or compoment outside of auth.
+
+### Operational Aspects of API Extensions
+
+There is no explicit operational involvement expected. The property is optional and set by default to log audit events.
```
stlaz, line 128: nit: `to log login events.`

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,344 @@
...
+
+It could be turned off by a user by removing API extension.
```
stlaz, line 130: That would actually default back to it being turned on.

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,344 @@
...
+2. Verify that the version deployed contains this enhancement.
+3. Verify that the custom resource `oauths.config.openshift.io` exists and contains the properties:
+
+```yaml
+audit:
+    profile: WriteLoginEvents
+```
+
+4. Create Login failures and Login.
+5. Execute `oc adm must-gather`.
```
stlaz, line 307: No, not like that (ditto for all the must-gathers below).

https://docs.openshift.com/container-platform/4.9/security/audit-log-view.html#nodes-nodes-audit-log-basic-viewing_audit-log-view

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,344 @@
...
+
+Ideally we test our feature with:
+
+- unit tests,
+- e2e tests in origin and
+- tests at must-gather.
+
+#### E2E testing steps
+
+In order to test the audit logging as mentioned in the porposal:
```
stlaz, line 293: nit: `s/porposal/proposal/`

enhancements/authentication/login-logout-events.md
ibihim: Do you think so? If there is no `audit` in `OAuthSpec`?

enhancements/authentication/login-logout-events.md
ibihim: Whats wrong with must gather?

enhancements/authentication/login-logout-events.md
ibihim: Let me be more specific to set it to None, just in case we default, without a spec given.

enhancements/authentication/login-logout-events.md
stlaz: You're proposing to create a privileged container in the cluster to pull data from the cluster to get a tar that you need to process instead of doing a streamed API call

ibihim: Sure, for automated things :+1: :+1: :heart_eyes:

enhancements/authentication/login-logout-events.md
```diff
@@ -0,0 +1,344 @@
...
+	// tokenConfig contains options for authorization and access tokens
+	TokenConfig TokenConfig `json:"tokenConfig"`
+
+	// templates allow you to customize pages like the login page.
+	// +optional
+	Templates OAuthTemplates `json:"templates"`
+	// audit specifies what should be audited in the context of OAuthServer.
+	// +optional
+	// +kubebuilder:default:={"profile":"WriteLoginEvents"}
+	Audit OAuthAudit `json:"audit"`
```
mfojtik, line 78: if this is optional, shouldn't it be pointer and `omitempty` ?

mfojtik:
The design looks architecturally sound.

/approve

PR #898, opened 2021-09-15
OWNERS: Prune crawford

wking:
Alex has left Red Hat :(.


OWNERS
```diff
@@ -8,7 +8,6 @@ approvers:
   - celebdor
   - chancez
```
sdodson, line 9: Chance too, we need to clean this up broadly and make DPP responsible for scrubbing OWNERS of the departed.

PR #899, opened 2021-09-15
authentication: pod-security initial commit

s-urbaniak:


enhancements/authentication/pod-security-admission.md
```diff
@@ -0,0 +1,113 @@
+---
+title: pod-security-admission
+authors:
+  - "@s-urbaniak"
+reviewers:
+  - TBD
+approvers:
+  - TBD
```
wking, line 8: `reviewers` and `approvers` probably need populating.

enhancements/authentication/pod-security-admission.md
```diff
@@ -0,0 +1,113 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Motivation
+
+In contrast to PodSecurityPolicy admission the desire is not to disable the new upstream PodSecurity admission plugin subsystem within OpenShift.
```
bparees, line 40: i'm not clear on what the "in contrast to" is contrasting.

enhancements/authentication/pod-security-admission.md
```diff
@@ -0,0 +1,113 @@
...
+The new PodSecurity admissions plugin is a validating plugin only. This implies the opportunity to let it coexist with the existing SCC logic.
+
+### Goals
+
+[1] Enable PodSecurity admission
+[2] Design an architecture that lets PodSecurity coexist with SCC logic
+
+### Non-Goals
+
+[1] Break existing SCC API
```
bparees, line 50:
```suggestion
[1] Deprecate existing SCC API
```

?

break is never an option, but it's important to understand if we still think SCC has its own role to play or not.

enhancements/authentication/pod-security-admission.md
s-urbaniak: `PodSecurityPolicy` is disabled in OpenShift but there is a desire to enable the `PodSecurity` admission. Do you have a suggestion how I can phrase this better (sorry, non-native speaker) :grimacing:

enhancements/authentication/pod-security-admission.md
s-urbaniak: Agreed on the suggestion, this sounds better :+1: SCCs will definitely stay and we must not break that functionality. Do you think it is advisable to add a paragraph that empasizes this?

enhancements/authentication/pod-security-admission.md
bparees: "Unlike the (now deprecated?) PSP admission plugin which is disabled in openshift, the goal here is to enable the PS admission plugin within openshift"

enhancements/authentication/pod-security-admission.md
bparees: i think your existing sentence: 
```
Special interest is put to design a migration scheme which lets PodSecurity coexist with the existing Security Context Constraints (SCC) mechanism.
```

is sufficiently clear that SCCs aren't going anywhere.

enhancements/authentication/pod-security-admission.md
s-urbaniak: This sounds much better, thank you! I will add this sentence instead :+1:

enhancements/authentication/pod-security-admission.md
```diff
@@ -0,0 +1,113 @@
...
+
+Initially, PodSecurity if configured with the following settings:
+
+| Policy  | Profile |
+| --------| --------|
+| enforce | privileged |
+| warn    | baseline |
+| audit   | baseline |
+
+This configures PodSecurity to run in "no-op" mode but will emit warnings and audit log events in case of "baseline" violations.
```
bparees, line 75: "enforce: privileged" doesn't sound no-op?

s-urbaniak: it is counter-intuitive, it means "enforce privileged policy rules in this namespace", aka "all workloads are privileged".

enhancements/authentication/pod-security-admission.md
```diff
@@ -0,0 +1,113 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Motivation
+
+In contrast to PodSecurityPolicy admission the desire is not to disable the new upstream PodSecurity admission plugin subsystem within OpenShift.
+The new PodSecurity admissions plugin is a validating plugin only. This implies the opportunity to let it coexist with the existing SCC logic.
```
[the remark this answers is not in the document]
sttts: PS is only validating and hence the order doesn't play a role. SCC is mutating by design and as such it runs before all validating plugins.

enhancements/authentication/pod-security-admission.md
s-urbaniak: fixed, added @mfojtik as approver

enhancements/authentication/pod-security-admission.md
s-urbaniak: clarification: podsecurity is indeed deprecated as of 1.21 (see https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/)

s-urbaniak: I will add this reference.

mfojtik:
/approve

this is in line with the plan that was debated in control plane theme call and also internally.

enhancements/authentication/pod-security-admission.md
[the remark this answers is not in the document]
s-urbaniak: I think this is good information to add here.

enhancements/authentication/pod-security-admission.md
```diff
@@ -0,0 +1,116 @@
...
+2. Analyze audit logs from CI runs for violations against the "baseline" policy level
+
+From existing CI e2e runs we can introspect the amount of workloads that violate the "baseline" policy.
+This gives us an indicator what workloads need to be annotated as privileged to pass PodSecurity admission.
+
+3. Increase warn and audit profile to "restricted"
+
+Similar to step 2. we will reiterate on existing workloads and asses if there is necessary action.
+
+For existing clusters this can be configured at runtime using an unsupported config override:
```
stlaz, line 89: for existing clusters that already passed step 1. I assume

enhancements/authentication/pod-security-admission.md
```diff
@@ -0,0 +1,116 @@
...
+This configures PodSecurity to run in "no-op" mode but will emit warnings and audit log events in case of "baseline" violations.
+
+2. Analyze audit logs from CI runs for violations against the "baseline" policy level
+
+From existing CI e2e runs we can introspect the amount of workloads that violate the "baseline" policy.
+This gives us an indicator what workloads need to be annotated as privileged to pass PodSecurity admission.
+
+3. Increase warn and audit profile to "restricted"
+
+Similar to step 2. we will reiterate on existing workloads and asses if there is necessary action.
```
stlaz, line 87: what would a necessary action be in this case? What's the end goal? Maybe we could have a "Future work" section that goes into possibilities of why this is being done?

enhancements/authentication/pod-security-admission.md
```diff
@@ -0,0 +1,113 @@
...
+[1] https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement#podsecuritypolicy-migration
+
+For OpenShift, the following high level migration plan is envisioned:
+
+1. (DONE) Enable the PodSecurity admission plugin in no-op mode but with the ability to audit policy violations.
+
+- https://github.com/openshift/api/pull/1001 add `PodSecurity` to default enabled feature set in OpenShift hyperkube
+- https://github.com/openshift/cluster-kube-apiserver-operator/pull/1217 enable `PodSecurity` admission plugin and provide a default configuration
+- https://github.com/openshift/kubernetes/pull/950 vendor api changes in openshift/kubernetes
+- https://github.com/openshift/cluster-kube-apiserver-operator/pull/1231 enable `PodSecurity` feature gate in kube-apiserver-operator
```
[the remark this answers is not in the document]
s-urbaniak: no disagreement, let's remove them then, at the same time leaving them here as a comment:
- https://github.com/openshift/api/pull/1001: add `PodSecurity` to default enabled feature set in OpenShift hyperkube
- https://github.com/openshift/cluster-kube-apiserver-operator/pull/1217: enable `PodSecurity` admission plugin and provide a default configuration
- https://github.com/openshift/kubernetes/pull/950: vendor api changes in openshift/kubernetes
- https://github.com/openshift/cluster-kube-apiserver-operator/pull/1231

enhancements/authentication/pod-security-admission.md
s-urbaniak: fixed

enhancements/authentication/pod-security-admission.md
s-urbaniak: > what would a necessary action be in this case?

readjusting pod security annotations. Added this as it was indeed missing.

> Maybe we could have a "Future work" section that goes into possibilities of why this is being done?

sounds great, will add our current thoughts (controller which synchronizes SCC and Pod Security settings)

PR #901, opened 2021-09-17
BUILD-348: update shared resource via CSI Driver proposal

gabemontero:
/assign @adambkaplan 
/assign @deads2k 
/assign @jsafrane 

@coreydaley FYI

A refresh on out EP stemming from:
- feedback from @deads2k in @coreydaley 's https://github.com/openshift/api/pull/979
- updates I've had in the back of my mind that were needed
- "human cherrypicking" (vs. git)  content from @adambkaplan 's https://github.com/openshift/enhancements/pull/742

If the reference helps, https://github.com/openshift/csi-driver-shared-resource/pull/51 includes changes articulated here (a POC if you will), as well as the latest version of @coreydaley 's https://github.com/openshift/api/pull/979 but applied to the current in-repo API (i.e. the thing we want to migratate to https://github.com/openshift/api as part of getting this enhancement in the OCP payload).

As always, thanks everyone for prior collaboration that led to this PR, and the upcoming comments from you all once you get the cycles to review

### DIFF
--- a/enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
+++ b/enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
@@ -23,7 +23,7 @@ replaces:
 
 creation-date: 2020-03-17
 
-last-updated: 2020-09-01
+last-updated: 2021-09-17
 
 <!-- status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced -->
 status: implementable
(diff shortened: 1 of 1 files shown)


enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md
```diff
@@ -111,116 +128,113 @@ users' workload Pods so that containers in those Pods can utilize those CAs when
...
+```yaml
+rules:
+- apiGroups:
+  - storage.openshift.io
+  resources:
+  - shareresources
+  verbs:
   - list
-  - watch
+  - get
```
adambkaplan, line 207: I think there is an argument to be made that this cluster role should be installed by the operator and aggregated to the user-facing "edit" role [1]:

1. Edit users are assumed to create and manage workloads. By their very nature they are granted a high level of trust.
2. Edit users are the primary audience who need to be able to discover shared resources.
3. The details embedded in a SharedResource are IMO pseduo-sensitive, equivalent to a network IP address. The object can divulge the _potential_ existence of a Secret/ConfigMap in another namespace. However the object does not provide full details of truly sensitive information contained in a Secret.

I really do like this solution with the `use` verb - it succeeds in separating usage of the shared resource with the discovery of the shared resource. I can envision a `SharedResource` to be secured as follows:

1. Admins have full permissions (`*` verb)
2. Editors have (`get`, `list`), thereby can discover the `SharedResources`
3. Editors can request that a _service account_ be granted the `use` permission. Admin configures the appropriate cluster rolebinding.
4. Editors can then deploy a workload which _uses_ the shared resource. 

[1] https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

adambkaplan (approved):
/lgtm

Not sure if we want the api discussion to block merge - IMO we can keep this as an open question and submit a follow-up PR once we've come to agreement.

gabemontero:
@deads2k @jsafrane - just chatted with @adambkaplan on slack, and he did not intend to merge this PR with his lgtm

if you all have cycles to still look at this, I'll take whatever comments you have and craft a follow up PR to incorporate them

thanks

PR #902, opened 2021-09-21
Add swap support to OpenShift

ehashman:
/assign @rphillips 
/cc @mrunalp 

This is a proposal for how to integrate upstream Kubernetes swap memory support into OpenShift, including manifests that can enable swap on a 4.9+ cluster for testing and a plan to add Tech Preview support in 4.10.

### DIFF
--- a/enhancements/kubelet/node-swap.md
+++ b/enhancements/kubelet/node-swap.md
@@ -1,0 +1,314 @@
+---
+title: node-swap
+authors:
+  - "@ehashman"
+reviewers:
+  - "@rphilips"
+  - "@sjenning"
+  - "???"
+approvers:
+  - "@mrunalp"
+creation-date: "2021-06-23"
+status: implementable
+---
+
+# OpenShift Node Swap Support
+
+## Release Signoff Checklist
+
+- [X] Enhancement is `implementable`
+- [X] Design details are appropriately documented from clear requirements
+- [X] Test plan is defined
+- [X] Operational readiness criteria is defined
+- [X] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The upstream Kubernetes 1.22 release introduced alpha support for configuring
+swap memory usage for Kubernetes workloads on a per-node basis.
+
+Now that swap use on nodes is supported in upstream, there are a number of use
+cases that would benefit from OpenShift nodes supporting swap, including
+improved node stability, better support for applications with high memory
+overhead but smaller working sets, the use of memory-constrained devices, and
+memory flexibility.
+
+## Motivation
+
+See [KEP-2400: Motivation].
+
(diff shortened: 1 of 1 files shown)


enhancements/kubelet/node-swap.md
```diff
@@ -0,0 +1,142 @@
...
+
+- We will add `NodeSwap` to the [`TechPreviewNoUpgrade`] feature gate list.
+- We can use [ignition configs](https://coreos.github.io/ignition/configuration-v3_3/) to add swap partitions to worker nodes. (filesystems.format = swap)
+  - Options: partition on the root node, or perhaps provision and mount an NVMe volume? (https://github.com/openshift/machine-config-operator/issues/1619)
+- We will need to ensure the node has `swapon` before kubelet starts
+- The kubelet needs an appropriate KubeletConfiguration (e.g. `NodeSwap` feature flag enabled, `failSwapOn = false`, and [`memorySwap.SwapBehavior` set](https://kubernetes.io/docs/concepts/architecture/nodes/#swap-memory))
+
+TODO: how will we modify MCO to roll this out? Do we need to at all?
+
+### Open Questions [optional]
```
ehashman, line 77: Do we need something to validate that the version of CRI-O installed on the node has support?

enhancements/kubelet/node-swap.md
```diff
@@ -0,0 +1,142 @@
...
+
+- Will we eventually want to enable swap on all OpenShift nodes by default?
```
rphillips, line 79: Probably by Machine Config Pool. A customer can have different usages for each Pool, so it makes sense to turn swap on by Pool.

We would  probably want to enable swap by default... currently that would be with the 'worker' and 'master' pools.

enhancements/kubelet/node-swap.md
```diff
@@ -0,0 +1,142 @@
...
+- Should swap just be limited to worker nodes, or should we consider adding it to control plane nodes too?
```
rphillips, line 80: Maybe just worker nodes, but allow swap to be enabled on masters? We probably need some testing around this.

enhancements/kubelet/node-swap.md
ehashman: Shouldn't be necessary so long as we only work on 4.10+

enhancements/kubelet/node-swap.md
```diff
@@ -0,0 +1,310 @@
...
+Beyond tech preview, we may want to look into adding swap support to the
+installer, and consider adding a partition on the root node, or perhaps
+[provision and mount an NVMe
+volume](https://github.com/openshift/machine-config-operator/issues/1619).
+
+### Open Questions [optional]
+
+- Will we eventually want to enable swap on all OpenShift nodes by default?
+- Should swap just be limited to worker nodes, or should we consider adding it
+  to control plane nodes too?
```
mrunalp, line 200: Could add a line saying that our testing will help answer these questions.

ehashman: Done!

PR #907, opened 2021-09-22
CONSOLE-2894: Multi-cluster console

spadgett:
Proposes an approach for adding basic multi-cluster support to OpenShift console with a cluster selector. The console will discover other clusters in the fleet using the ACM `ManagedCluster` resource.


enhancements/console/multi-cluster-console.md
```diff
@@ -0,0 +1,366 @@
...
+* The OpenShift console will have a cluster selector that allows you to view
+  resources for any cluster in the fleet.
+* Users will be able to create and edit resources on spoke clusters from the
+  console running on the hub.
+* The console will support viewing metrics for workloads running on spoke
+  clusters.
+* Console will support single sign-on (SSO) to avoid logins every time a user
+  switches clusters.
+
+### Non-Goals
```
jhadvig, line 62: Wondering if we also should add that we dont wanna solve any console customizations that is set via console-operator (branding, custom logo, custom product name, ...) since the hub cluster will be only doing API requests to the spoke clusters. right ?

enhancements/console/multi-cluster-console.md
```diff
@@ -0,0 +1,366 @@
...
+
+* For now, we do not plan to remove the OpenShift console from the release
+  payload. Admins can disable the local console on spoke clusters using the
+  existing `managementState` property in the console operator config.
```
jhadvig, line 66:
```suggestion
  existing `spec.managementState` property in the console operator config.
```

enhancements/console/multi-cluster-console.md
```diff
@@ -0,0 +1,366 @@
...
+    name: oauth-serving-cert
+    namespace: openshift-config-managed
+```
+
+For 4.8 and earlier, the console operator will need to use the
+`default-ingress-cert` from the `openshift-config-managed` namespace.
+
+```yaml
+apiVersion: view.open-cluster-management.io/v1beta1
+kind: ManagedClusterView
```
jhadvig, line 149: Just to confirm, this is the CR that console-operator on the spoke cluster will fetch?

enhancements/console/multi-cluster-console.md
```diff
@@ -0,0 +1,366 @@
...
+#### Dev Preview -> Tech Preview
+
+N/A
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default when ACM is installed
+- Ability to connect to non-OpenShift clusters in the fleet
```
jhadvig, line 318: isnt this in the **Non-Goals** section ?

enhancements/console/multi-cluster-console.md
spadgett: Console operator creates this on the hub cluster. ACM will set the status of the `ManagedClusterView` resource on the hub based on the data from the spoke.

enhancements/console/multi-cluster-console.md
spadgett: Yeah, ultimately we want to support this, just not for the tech preview. I'll take it out of non-goals.

enhancements/console/multi-cluster-console.md
spadgett: We plan to use the branding configuration from the hub cluster.

PR #909, opened 2021-09-23
baremetal: Include CoreOS ISO in the release payload

zaneb:
The baremetal platform is switching from a QCOW2 format for its CoreOS image to an ISO. To ensure that existing disconnected clusters can update to this, the ISO image will be included in the release payload. This will be balanced out by removing the RHEL image currently shipped in the release payload. There are many other nice benefits, such that after future upgrades clusters will provision new machines with the latest CoreOS, without requiring an extra reboot.


enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,258 @@
...
+  with, we are able to deliver updates to IPA and the OS it runs on.
+* Stop maintaining and shipping the non-CoreOS, RHEL-based IPA PXE files.
+* Never break existing clusters, even if they are deployed in disconnected
+  environments.
+* Eliminate the extra reboot required to update CoreOS after initial
+  provisioning.
+
+### Non-Goals
+
+* Automatically switch pre-existing MachineSets to deploy with `coreos-install`
```
dtantsur, line 105: nit: `coreos-installer`

dtantsur:
/lgtm

Makes sense from the baremetal perspective

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Summary
+
+The baremetal platform is switching from a QCOW2 format for its CoreOS image to
+an ISO. To ensure that existing disconnected clusters can update to this, the
+ISO image will be included in the release payload. This will be balanced out by
+removing the RHEL image currently shipped in the release payload. There are
+many other nice benefits, such that after future upgrades clusters will
+provision new machines with the latest CoreOS, without requiring an extra
+reboot.
```
hardys, line 47: We might want to be a bit more specific here, e.g we're switching from using the openstack QCOW2 image to using *the* live-iso that's already supported for UPI baremetal deployments.

And relatedly, another benefit we could call out is that this aligns the baremetal IPI flow with what is already supported for UPI.

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,258 @@
...
+proposal](https://github.com/openshift/enhancements/pull/201). Other issues
+specific to the baremetal platform are that boot times for bare metal servers
+can be very long (and therefore the reboot is costly), and that support for
+particular hardware may theoretically require a particular version of CoreOS.
+
+We are changing the deploy disk image to use CoreOS. This takes the form of
+both an ISO (for hosts that can use virtualmedia) and of a kernel + initrd +
+rootfs (for hosts that use PXE). When upgrading an existing disconnected
+cluster, we have no way to acquire these images without the user manually
+intervening to mirror them.
```
hardys, line 77: Again here I'd maybe say "We are changing to use the same CoreOS based images provided for UPI deployments" or similar, just so it's clear we're aligning the flows vs generating any net-new artifacts.

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,258 @@
...
+
+Like the QCOW2 provisioning disk image, the URLs for these images are known by
+the installer, but they point to the cloud by default and would have to be
+customised by the user at install time to allow disconnected installs. IPA
+itself is delivered separately, as a container image as part of the OpenShift
+release payload.
```
hardys, line 83: The terminology here (and above) could perhaps be confusing, e.g "provisioning disk image" vs "deploy disk image" - maybe say "target disk image" or "provisioned disk image" for the provisioning one, so it's clear that's what we're writing to disk, vs using during provisioning?

I'd also perhaps stick to saying e.g "The deploy disk image itself is delivered separately ..." for the same reason - unless you are familiar with IPA it's not super-clear we're talking about the deploy image when saying "IPA" here?

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,258 @@
...
+  environments.
+* Eliminate the extra reboot required to update CoreOS after initial
+  provisioning.
+
+### Non-Goals
+
+* Automatically switch pre-existing MachineSets to deploy with `coreos-install`
+  instead of via QCOW2 images.
+* Update the CoreOS QCOW2 image in the cluster with each OpenShift release.
+* Build CoreOS images for platforms other than baremetal.
```
hardys, line 108: Maybe switch from "Build CoreOS images" to "Provide CoreOS images", since we're only talking about a way to provide existing built artifacts (then optionally embedding some config), not respinning any special/new images for this platform-specific usage?

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,258 @@
...
+
+Build a container image containing the latest CoreOS ISO and the
+coreos-installer binary. This container can be used e.g. as an init container
+to make the ISO available where required. The `coreos-installer iso extract
+pxe` command can also be used to produce the kernel, initrd, and rootfs from
+the ISO, for the purposes of PXE booting.
+
+This image could either replace the existing content of the
+`ironic-ipa-downloader` repo, be built from the new
+`image-customization-controller` repo, or be built from a new repo.
```
hardys, line 120: Personally I think it'd be clearer if we created a new repo/image than re-purposing either of these?

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,258 @@
...
+should now be viable after [fixes to the
+installer](https://github.com/openshift/installer/pull/4413).
+
+The correct version of CoreOS to use is available from the openshift-installer;
+this could be obtained by using the installer container image as a build image.
+
+Because OpenShift container images are built in an offline environment, it is
+not possible to simply download the image from the public Internet at container
+build time. Instead this will be accomplished by *MAGIC*. Uploading the ISOs to
+the lookaside cache in Brew is one possibility.
```
hardys, line 149: Do we have a plan for resolving exactly how this process will work?

hardys:
Overall this looks good to me, and I'm really pleased to see progress on this topic :)

Few minor comments, but the general direction all sounds good to me - it'd be good to get an ack from @cgwalters and @jlebon before approving though

enhancements/baremetal/coreos-image-in-release.md
zaneb: > I'd also perhaps stick to saying e.g "The deploy disk image itself is delivered separately ..." for the same reason - unless you are familiar with IPA it's not super-clear we're talking about the deploy image when saying "IPA" here?

We're not :) What this is saying that IPA is delivered separately to the deploy image, so that (next paragraph) even if you use an out-of-date CoreOS, you will still pull the latest IPA from the registry in the cluster, since it's part of the release payload. I'll rearrange this so it doesn't jump around so much.

IPA and "deploy image" are defined at the top of this section.

enhancements/baremetal/coreos-image-in-release.md
zaneb: We'll need to pull in some folks from the appropriate team (ART?) to tell us the right way to accomplish it.

sadasu:
Overall a pretty detailed enhancement. Would be nice if someone from the ART team helps demystify the *MAGIC*.

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Goals
+
+* Ensure that no matter which version of OpenShift a cluster was installed
+  with, we are able to deliver updates to IPA and the OS it runs on.
+* Stop maintaining and shipping the non-CoreOS, RHEL-based IPA PXE files.
+* Never break existing clusters, even if they are deployed in disconnected
+  environments.
+* Eliminate the extra reboot required to update CoreOS after initial
+  provisioning.
```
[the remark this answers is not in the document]
cgwalters: > Even with #201, my understanding is that the goal there isn't to always have the bootimages match the OSTree version in the machine-os-content, but just making it updateable.

I would say the goal of #201 is indeed avoiding a mandatory reboot.

Anyways agreed to drop this as a goal for now, but as long as we link to #201 I think it's covered.

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,261 @@
...
+build time. Instead this will be accomplished by *MAGIC*. Uploading the ISOs to
+the lookaside cache in Brew is one possibility.
+
+The CoreOS release pipeline may need to be adjusted to ensure that the latest
+ISO is always available in the necessary location.
+
+We will need to ensure that the container image is rebuilt whenever the CoreOS
+image is updated.
+
+OKD uses Fedora CoreOS images instead of actual CoreOS, and this will need to
```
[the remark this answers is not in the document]
cgwalters: I'd s/actual CoreOS/RHEL CoreOS/

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,261 @@
...
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The baremetal platform is switching from the OpenStack QCOW2 CoreOS image to
+the live ISO (as also used for UPI). To ensure that existing disconnected
+clusters can update to this, the ISO image will be included in the release
+payload. This will be balanced out by removing the RHEL image currently shipped
+in the release payload. There are many other nice benefits, such that after
+future upgrades clusters will provision new machines with the latest CoreOS,
```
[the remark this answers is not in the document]
zaneb: Most of this doc applies equally to OKD and Fedora CoreOS. I've changed all the incidences that are RHCOS-specific.

enhancements/baremetal/coreos-image-in-release.md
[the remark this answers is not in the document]
zaneb: Thanks for the extra detail, moved this to non-goal.

jlebon (approved):
Seems reasonable to me!

enhancements/baremetal/coreos-image-in-release.md
sosiouxme: the good news is that there is a mechanism for ART to download and insert binaries like this if that is in fact what we determine to do. we would just need a script written to do it at the time we rebase the source.

the bad news is that it deepens the dependency tree of things that need to align in a release and falls outside of how we currently determine what needs building.

dependency tree: currently we have hyperkube building as an RPM, getting included in an RHCOS build, and that build then getting included in a nightly. this change would mean that for consistency, baremetal would also need to rebuild and be included in a nightly. there are lots of ways for that to go wrong and we would end up with mismatching content. i would say we could validate that before release but i'm actually not sure how we would, would have to think about it. in any case, there would likely be sporadic extended periods where the two were out of sync and we couldn't form a valid release.

detection that the image needs rebuilding: currently we check whether the source has changed, whether the config has changed, or whether RPM contents have changed, which are all things we can easily look up from the image or brew. there's nothing like that for other content. i guess to do this, the process that does the upload could add a label with the RHCOS buildid that's included, and we could then compare that to see if there's a newer build available. i don't look forward to adding that bit of complexity into the scan but it seems solvable.

some other things to consider:
1) how does this work for OKD/CI builds?
2) i assume we need this for all architectures? each arch is a different RHCOS build and that makes it a bit tricky to reuse the same dockerfile for all arches (when the content will need different arch-dependent names).

zaneb: So far baremetal only supports x86_64. But it seems inevitable that _one_ day we'll have to support multiple architectures. Could we have one Dockerfile per arch?

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,260 @@
...
+The RHCOS release pipeline may need to be adjusted to ensure that the latest
+ISO is always available in the necessary location.
+
+We will need to ensure that the container image is rebuilt whenever the RHCOS
+image is updated.
+
+OKD uses (rebuilt) Fedora CoreOS images instead of RHEL CoreOS, and this will
+need to be taken into account. It may be that the build environment for OKD
+allows for a more straightforward solution there (like downloading the image
+directly).
```
zaneb, line 158: @LorbusChris could you comment on how you might imagine this working for OKD?

enhancements/baremetal/coreos-image-in-release.md
cgwalters: I think we should always keep this ISO the same as the one pinned in the installer.

So one idea: have this image derive from the installer.  I think we'd get edge triggering for free then.

cgwalters: > So far baremetal only supports x86_64. But it seems inevitable that one day we'll have to support multiple architectures.

Note that the [stream metadata](https://github.com/openshift/installer/blob/master/data/data/rhcos-stream.json) includes multiple architectures already, and I think this should use that as a source of truth.

enhancements/baremetal/coreos-image-in-release.md
cgwalters: For OKD I think we'd just use the [fcos stream metadata](https://github.com/openshift/installer/blob/master/data/data/fcos-stream.json) there?

enhancements/baremetal/coreos-image-in-release.md
sosiouxme: config has baremetal _building_ for all arches so i assumed those were used but it's simpler if not. we should probably limit those builds to x86...

our build system insists that a single multi-arch build reuse the same dockerfile for all (this is useful for ensuring they all have consistent content). there are ways around it, though, involving either complicating the dockerfile a bit, or (more likely) splitting the builds into a build per arch - we already have some like this for ARM64. so I guess this is not a blocker, just another complication.

sosiouxme: keeping the ISO same as installer seems like a good idea. that would mean it's not changing frequently.

i'm not _quite_ sure how we'd implement that; i'm not sure how we'd determine what to download prior to the build, and we can't download during the build. except maybe from brew... heh.

LorbusChris: > some other things to consider:
> 
> 1. how does this work for OKD/CI builds?
> 
> 2. i assume we need this for all architectures? each arch is a different RHCOS build and that makes it a bit tricky to reuse the same dockerfile for all arches (when the content will need different arch-dependent names).

1: For OKD, the ostree in machine-os-content is different than the one used in the bootimage (which is FCOS). For the okd-machine-os ostree, the hyperkube and client RPMs are extracted from the artifacts image, and layered onto FCOS: https://github.com/openshift/okd-machine-os/blob/master/Dockerfile.cosa#L9. Since the trees differ, a new node in OKD always has to pivot from FCOS to okd-machine-os before it even has a kubelet do run anything with.  

2: This will be needed soon at least for ARM64

enhancements/baremetal/coreos-image-in-release.md
LorbusChris: Yes, I think for OKD we can just download the FCOS version pinned in the stream metadata

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,260 @@
...
+latest bug and security fixes.
+
+## Design Details
+
+### Open Questions [optional]
+
+How will we provide up-to-date images to the container build?
+
+Are there similar internet access restrictions on the build environment for
+OKD?
```
LorbusChris, line 173: No. OKD is built in Prow (and currently also partly in CirrusCI). Builds are currently not done in a disconnected env, and as far as I know it is not planned to do that in future either.

enhancements/baremetal/coreos-image-in-release.md
zaneb: I added a paragraph on multi-arch to the doc.

@sosiouxme what additional information do you think we need to document here to get this to an approvable state?

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,269 @@
...
+
+The target OS disk image is a separate CoreOS QCOW2 image. The URL for this is
+known by the installer. It points to the public Internet by default and may be
+customised by the user to allow disconnected installs. The URL is stored in the
+Provisioning CR at install time and never updated automatically. The image
+itself is downloaded once and permanently cached on all of the master nodes.
+Never updating the image is tolerable because, upon booting, the CoreOS image
+will update itself to the version matching the cluster it is to join. It
+remains suboptimal because new Machines will take longer and longer (and more
+and more bandwidth) to join as the cluster ages. This issue exists on all
```
wking, line 65: this is capped the "the time it takes to download and setup a completely unrelated OS image", right?  E.g. if you install with release A, and later by release B there is now nothing from A that survives, the A->B new-node time is as slow as it gets, and a later release C will have a similar A->C new-node time (assuming C is around the same size as B)?  It's still not great, and it's enough to motivate this enhancement, but maybe worth rewording to something like:

> ... new Machine will take longer (and more bandwidth) to join as the cluster's desired OS diverges from the boot image, capping out when the desired OS is completely different from the boot image.

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,269 @@
...
+the OpenShift release payload, so in any event we will continue to be able to
+update IPA.
+
+We wish to solve the problems with obtaining an up-to-date CoreOS by including
+it in the release payload.
+
+### Goals
+
+* Ensure that no matter which version of OpenShift a cluster was installed
+  with, we are able to deliver updates to IPA and the OS it runs on.
```
wking, line 95: Hmm, if you avoid bumping this image with each release and as a result keep the reboot (both stated non-goals), it was initially unclear to me what the win was from this entry.  Going over the *Motivation* and *User Stories* sections again, I think it might be worth explicitly calling out these sub-cases here:
 
* More convenient experience for folks on restricted networks, who would no longer need to mirror the boot image separately, since it would be mirrored with the rest of the OCP release.
* Support for adding nodes with newer hardware that was not supported by the boot image mirrored when the cluster was installed.
* Avoiding a window of pre-pivot bug/security exposure by sun-setting older boot images.

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,269 @@
...
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+The container registry will always contain an image with latest ISO. This will
+get rolled out to the `image-customization-controller` pod, and future boots of
+the deploy image will be based on the new ISO. The restart of ironic during the
+upgrade should ensure that any BaremetalHosts currently booted into the deploy
+image (i.e. not provisioned) will be rebooted into the new one.
```
wking, line 210: do these pods run on the host cluster's control-plane nodes?  Because you can set compute MachineConfigPools `paused: true` and they won't be rolled during the update or block the update's completion.  And even if your pool is unpaused, the machine-config operator may not drain and reboot nodes if the `machine-os-content` image (holding RHCOS) didn't get bumped by the update (most patch updates bump it, but not all).  I'm not sure where the ironic pod is from, but I'd guess that may or may not get bumped in updates that happen to bump the boot image.  So unless you're running the pod off the boot image itself with a by-digest pullspec, I think expecting something to come around and roll you may be overly optimistic.

zaneb: > Does the CBO need to keep track of the image reference and ensure container that use the image are told when it changes?

It does already.

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,269 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The baremetal platform is switching from the OpenStack QCOW2 CoreOS image to
```
romfreiman, line 42: @zaneb this enhancement will also assist AI. Maybe we should address both cases here?

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,269 @@
...
+
+OKD uses Fedora CoreOS images instead of RHEL CoreOS, but the URLs are
+nevertheless listed in the installer's [stream
+metadata](https://github.com/openshift/installer/blob/master/data/data/fcos-stream.json).
+The build environment for OKD is *not* disconnected, so the container build can
+download the image directly.
+
+### Risks and Mitigations
+
+If the build pipeline does not guarantee the latest RHCOS version gets built
```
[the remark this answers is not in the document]
zaneb: We get to drop the ironic-ipa-downloader image from the payload though, and that is probably even larger.

enhancements/baremetal/coreos-image-in-release.md
sosiouxme: I think my concerns are answered.

ART can define a hook such that each time we build, we check if `installer/data/data/rhcos-stream.json` has changed. If it has, download the ISOs from the locations given there and make them available in the build context with an arch-specific name, probably something like `rhcos.$arch.iso` such that the same Dockerfile can just use the current arch to pick the right one in the arch-specific builds.

enhancements/baremetal/coreos-image-in-release.md
```diff
@@ -0,0 +1,282 @@
...
+the lookaside cache in Brew is one possibility.
+
+Different processor architectures require different ISOs (though all are
+represented in the stream metadata). It may make sense to have a separate
+Dockerfile for each architecture. For the foreseeable future, we only need to
+be able to pull a container for the local architecture that contains an ISO for
+that same architecture. In the future, it is possible that the containers for
+each architecture's ISO may themselves need to be multi-architecture, so that
+it is possible to run a container image containing an ISO for a *different*
+architecture.
```
sosiouxme, line 172: implementation details (I don't think we need to split image builds or anything fancy)

```suggestion
build time. Instead this will be accomplished by ART defining a "modifications"
hook that, each time the source is updated before a build, checks whether
`data/data/rhcos-stream.json` has changed. If it has, ISOs from the locations
given there are downloaded and made available in the build context with an
arch-specific name, something like `rhcos.$arch.iso` so that the Dockerfile can
select the appropriate arch.

Different processor architectures require different ISOs (though all are
represented in the stream metadata). For the foreseeable future, we only need to
be able to pull a container for the local architecture that contains an ISO for
that same architecture. In the future, it is possible that the containers for
each architecture's ISO may themselves need to be multi-architecture, so that
it is possible to run a container image containing an ISO for a *different*
architecture.
```

PR #915, opened 2021-10-01
CONSOLE-2892: Allow dynamic plugins to proxy to services on the cluster

jhadvig:
First draft of the proposed feature for allowing dynamic plugins to proxy to services on the cluster.

Story: https://issues.redhat.com/browse/CONSOLE-2892

/assign @spadgett @vojtechszocs @christianvogt

### DIFF
--- a/enhancements/console/dynamic-plugins.md
+++ b/enhancements/console/dynamic-plugins.md
@@ -147,6 +147,49 @@ spec:
     basePath: '/'
 ```
 
+In case the plugin needs to communicate with some in-cluster service, it can
+declare a service proxy in its `ConsolePlugin` resource using the
+`spec.proxy.services` array field. A service `name`, `namespace` and `port`
+needs to be specified.
+
+Console backend exposes following endpoint in order to proxy the communication
+between plugin and the service:
+`/api/proxy/namespace/<service-namespace>/service/<service-name>:<port-number>/<request-path>?<optional-query-parameters>`
+
+An example proxy request path from plugin to `helm-charts` service,
+in `helm` namespace to list ten helm releases:
+`/api/proxy/namespace/helm/service/helm-charts:8443/releases?limit=10`
+
+Proxied request will use [service CA bundle](https://docs.openshift.com/container-platform/4.8/security/certificate_types_descriptions/service-ca-certificates.html) by default. The service must use HTTPS.
+If the service uses a custom service CA, the `caCertificate` field
+must contain the certificate bundle. In case the service proxy request
+needs to contain logged-in user's OpenShift access token, the `authorize`
+field needs to be set to `true`. The user's OpenShift access token will be
+then passed in the HTTP `Authorization` request header, for example:
+
+`Authorization: Bearer sha256~kV46hPnEYhCWFnB85r5NrprAxggzgb6GOeLbgcKNsH0`
+
+```yaml
+apiVersion: console.openshift.io/v1alpha1
+kind: ConsolePlugin
+metadata:
+  name: my-plugin
+spec:
+  displayName: 'My Plugin'
+  service:
+    name: my-console-plugin
+    namespace: my-operator-namespace
+    port: 8443
+    basePath: '/'
+  proxy:
+    services:
+    - name: helm-charts
+      namespace: helm
+      port: 8443
+      caCertificate: '-----BEGIN CERTIFICATE-----\nMIID....'
+      authorize: true
+```
+
 Plugins are disabled by default. They need to be manually enabled by a cluster
 administrator before console loads any plugin code. Console provides a UI
 for enabling plugins, and the list of enabled plugins is set in the console


enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
     basePath: '/'
 ```
 
+In case the plugin needs to communicate with an in-cluster services, it can
```
vojtechszocs, line 150:
```suggestion
In case the plugin needs to communicate with other in-cluster services, it can
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
...
+declare the communication in it's `ConsolePlugin` manifest, defining
```
vojtechszocs, line 151:
```suggestion
declare such services via `spec.proxyServices` in the `ConsolePlugin` resource manifest.
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
...
+`spec.proxyService` array field entry. In there service `name`,
```
vojtechszocs, line 152:
```suggestion
Each entry in `spec.proxyServices` must refer to an in-cluster service via `name`,
`namespace` and `port`. The console backend ...
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
...
+`namespace`, `port` and `basePath` needs to specified. The console backend
+will proxy the plugin request to the dependent service using the service
```
vojtechszocs, line 154:
```suggestion
will proxy the plugin request to the specified service using the service
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
...
+CA bundle.
+
+After that plugin can make request to the console endpoint, which will
```
vojtechszocs, line 157:
```suggestion
Console backend provides the following endpoint to proxy plugin requests to the relevant
service: `/api/plugin-proxy/<proxy-service-name>/<request-path>`
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
...
+proxy the request to the dependent service. Plugin's request shall use
+following console's endpoint:
+`/api/plugin-proxy/<dependent-service-name>/<request-path>`
+
+In case the service is using a custom service-ca, `caCertificate` field
```
vojtechszocs, line 162:
```suggestion
In case the service is using a custom service CA, `caCertificate` field
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
...
+for the service can be specified. In case the the service proxy request
+shall contain user's authentication token, the `authorize` field needs
```
vojtechszocs, line 164:
```suggestion
needs to contain user's authentication token, the `authorize` field can
```

vojtechszocs (changes requested):

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
     basePath: '/'
 ```
 
+In case the plugin needs to communicate with an in-cluster services, it can
+declare the communication in it's `ConsolePlugin` manifest, defining
+`spec.proxyService` array field entry. In there service `name`,
+`namespace`, `port` and `basePath` needs to specified. The console backend
+will proxy the plugin request to the dependent service using the service
+CA bundle.
```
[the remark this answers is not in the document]
spadgett: Do we require https? It would be good to state that here if so.

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
...
+metadata:
+  name: my-plugin
+spec:
+  displayName: 'My Plugin'
+  service:
+    name: my-console-plugin
+    namespace: my-operator-namespace
+    port: 8443
+    basePath: '/'
+  proxyService:
```
vojtechszocs, line 179:
```suggestion
  proxyServices:
```

enhancements/console/dynamic-plugins.md
spadgett, line 164:
```suggestion
shall contain user's OpenShift access token, the `authorize` field needs
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
...
+
+After that plugin can make request to the console endpoint, which will
+proxy the request to the dependent service. Plugin's request shall use
+following console's endpoint:
+`/api/plugin-proxy/<dependent-service-name>/<request-path>`
+
+In case the service is using a custom service-ca, `caCertificate` field
+for the service can be specified. In case the the service proxy request
+shall contain user's authentication token, the `authorize` field needs
+to be set to `true`, otherwise the token won't be part of the proxy request.
```
spadgett, line 165: I'd mention that the token is passed using the `Authorization` request header.

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
...
+    name: my-console-plugin
+    namespace: my-operator-namespace
+    port: 8443
+    basePath: '/'
+  proxyService:
+  - name: plugin-dependent-service
+    namespace: plugin-dependent-service-namespace
+    port: 8443
+    basePath: '/'
+    caCertificate: '-----BEGIN CERTIFICATE-----\nMIID....'
```
spadgett, line 184: Typically this would be a config map reference instead of specified inline.

spadgett:
cc @jamestalton

enhancements/console/dynamic-plugins.md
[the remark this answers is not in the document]
vojtechszocs: +1 this is consistent with the possibility to host plugin assets via explicit URL, instead of via current in-cluster service definition (name, namespace, port).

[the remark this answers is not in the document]
jhadvig: 👍 I like the idea. 
It also came to my mind that we could have `spec.services` array field which would contain the required fields like `name`, `namespace`, `port` and `basePath` (which are so far common) together with `type` enum field, which would in this case be `Proxy`.   The `spec.services` array field would also contain other optional fields, like mentioned `caCertificate` or `authorize`.
But that might be more clear down the road when we decide to promote `ConsolePlugin` to `v1beta1` or even to `v1`.

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
     basePath: '/'
 ```
 
+In case the plugin needs to communicate with an in-cluster services, it can
+declare the communication in it's `ConsolePlugin` manifest, defining
+`spec.proxyService` array field entry. In there service `name`,
+`namespace`, `port` and `basePath` needs to specified. The console backend
```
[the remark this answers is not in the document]
jhadvig: hmm for some reason I though that the request will need the `basePath` as well but you are right that thats just the case of public assets

enhancements/console/dynamic-plugins.md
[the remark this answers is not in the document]
jhadvig: > Your example has a CA certificate. So I'm assuming we don't always use the service CA bundle?

So I thought we wanna use the custom service-ca if the service is using a custom CA. If not we will default to the  service CA bundle.

> Do we require https? It would be good to state that here if so.

So when we are creating a service endpoint thats constructed from <service-name>-<service-namespace>:<service-port> it needs to be https. So I guess the https is requred otherwise the the service server will respond with `http: server gave HTTP response to HTTPS client`

enhancements/console/dynamic-plugins.md
jhadvig: Right, I just saw that some structs are using inlined CA certs so wanted to give it a try.
Also wanted to avoid granting console || console-operator any additional rights, otherwise the CM would need to be created in the `openshift-config-managed` || `openshift-config` (depends on the implementation). Or we could create a separate NS for this but not sure if thats not an overkill... 
Thoughts ?

jhadvig: route API is inlining the cert https://github.com/openshift/api/blob/master/route/v1/types.go#L234

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,44 @@ spec:
...
+will proxy the plugin request to the dependent service using the service
+CA bundle.
+
+After that plugin can make request to the console endpoint, which will
+proxy the request to the dependent service. Plugin's request shall use
+following console's endpoint:
+`/api/plugin-proxy/<dependent-service-name>/<request-path>`
```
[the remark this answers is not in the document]
spadgett: Maybe

```suggestion
`/api/proxy/namespace/<service-namespace>/service/<service-name>:<port-number>/<request-path>?<optional-query-parameters>`

I'd suggest providing a concrete example URL and what service URL that it maps to.
```

enhancements/console/dynamic-plugins.md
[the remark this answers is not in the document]
spadgett: Right, that's what I was thinking to keep things simple.

enhancements/console/dynamic-plugins.md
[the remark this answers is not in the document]
vojtechszocs: Another option is to use e.g. `/api/proxy/<plugin-name>/<service-name>/<request-path>`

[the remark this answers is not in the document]
spadgett: Service name is not unique... A plugin can proxy to two services with the same name in different namespaces, or the same service, but different ports.

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,48 @@ spec:
     basePath: '/'
 ```
 
+In case the plugin needs to communicate with other in-cluster service, it can
```
vojtechszocs, line 150:
```suggestion
In case the plugin needs to communicate with some in-cluster service, it can
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,48 @@ spec:
...
+declare a service proxy in it's `ConsolePlugin` resource using the
```
vojtechszocs, line 151:
```suggestion
declare a service proxy in its `ConsolePlugin` resource using the
```

jhadvig:
@spadgett @vojtechszocs PTAL

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,48 @@ spec:
...
+`spec.proxy.services` array field. A service `name`, `namespace` and `port`
+needs to specified.
```
vojtechszocs, line 153:
```suggestion
needs to be specified.
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,48 @@ spec:
...
+
+Console backend exposes following endpoint in order to proxy the communication
+between plugin and the service:
+`/api/proxy/namespace/<service-namespace>/service/<service-name>:<port-number>/<request-path>?<optional-query-parameters>`
+
+An example proxy request path from plugin to `helm-charts` service,
+in `helm` namespace to list ten helm releases:
+`/api/proxy/namespace/helm/service/helm-charts:8443/releases?limit=10`
+
+Proxied request will be using using the service CA bundle by default. For
```
vojtechszocs, line 163:
```suggestion
Proxied request will use the service CA bundle by default. For
```

vojtechszocs: I think we should mention here something like "default Console service CA bundle" or similar, not sure how to word it properly.

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,48 @@ spec:
...
+that reason HTTPS communication between plugin and the service is required.
+In case the service is using a custom service CA, `caCertificate` field
+for the service can be specified. In case the the service proxy request
```
vojtechszocs, line 166:
```suggestion
for the service can be specified. In case the service proxy request
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,48 @@ spec:
...
+needs to contain user's OpenShift access token, the `authorize` field needs
```
vojtechszocs, line 167:
```suggestion
needs to contain logged in user's OpenShift access token, the `authorize` field needs
```

enhancements/console/dynamic-plugins.md
```diff
@@ -147,6 +147,48 @@ spec:
...
+to be set to `true`. Token will be then passed using the `Authorization`
```
vojtechszocs, line 168:
```suggestion
to be set to `true`. Access token will be then passed using the HTTP `Authorization`
```

vojtechszocs (changes requested):

spadgett (approved):
/lgtm

PR #919, opened 2021-10-04
Multi-Container-Structured-Logging

alanconway:
Revised proposal based on customer feedback.
Uses annotations to control the indexing of multi-container Pods.
More flexible, simpler and less risky than previous proposal.

/assign @jcantrill
/cc @vimalk78

### DIFF
--- a/enhancements/cluster-logging/multi-container-structured-logging.md
+++ b/enhancements/cluster-logging/multi-container-structured-logging.md
@@ -1,5 +1,5 @@
 ---
-title: output-parameter-templates
+title: Multi-container-structured-logging
 
 authors:
   - "@alanconway"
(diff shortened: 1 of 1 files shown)


enhancements/cluster-logging/multi-container-structured-logging.md
```diff
@@ -22,120 +22,135 @@ see-also:
...
-outputs:
-  - name: myApp
-    type: elasticsearch
-	structuredTypePattern: '{.kubernetes.label.app.myApp}-{.kubernetes.container_name}'
+apiversion: apps/v1
+kind: Deployment
+metadata:
+  name: application-foo
+  annotations:
+	foo.example.com/application-foo: schema-v1
```
jcantrill, line 95: My experience has been we prefer to have well defined, namespaced keys  that describe something for the consumer, in our case cluster logging.  I might, at a minimum think we would prefer at least for this to be NS for logging; we already have a few now:
* `logging.openshift.io`

Why do we need to have the user give us a namespace in CLF instead of just "doing" the routing changes if the annotation exists?

enhancements/cluster-logging/multi-container-structured-logging.md
```diff
@@ -22,120 +22,135 @@ see-also:
...
+	foo.example.com/sidecar-foo: schema-v2
+	foo.example.com/istio-sidecar: istio-schema-v1
+	foo.example.com/default: default
```
jcantrill, line 98: Doe we need to explicitly define default?  Can we not just say... if it doesn't match these other things then it will be indexed in the default manner?

enhancements/cluster-logging/multi-container-structured-logging.md
```diff
@@ -22,120 +22,135 @@ see-also:
...
-    type: elasticsearch
-	structuredTypePattern: '{.kubernetes.label.app.myApp}-{.kubernetes.container_name}'
+apiversion: apps/v1
+kind: Deployment
+metadata:
+  name: application-foo
+  annotations:
+	foo.example.com/application-foo: schema-v1
+	foo.example.com/sidecar-foo: schema-v2
+	foo.example.com/istio-sidecar: istio-schema-v1
```
jcantrill, line 97: I'm interested in the implementation of this:

* What is an example of a concrete index?
* What does the rollover managment look like? 
* Is it falling under `app` and subject to the same criteria?  
* How is it aliased?

enhancements/cluster-logging/multi-container-structured-logging.md
alanconway: I like this idea. I'd go with `container.logging.openshift.io` in case we want to use this for other features.
There is one problem use case - if there are multiple ElasticSearch back-ends with different index set-ups. 
I'll find out if that is a real requirement. If it is, we could use the fixed prefix for common cases and allow over-ride in YAML for multiple ES outputs.

enhancements/cluster-logging/multi-container-structured-logging.md
alanconway: This would be the same way we currently handle `structuredTypeKey`; except that instead of looking up the key as a label, we look up "prefix+container-name" as an annotation.

enhancements/cluster-logging/multi-container-structured-logging.md
alanconway: /hold
waiting for response to question on multi-ES requirement

alanconway: /unhold
Proposal updated to use `container.logging.openshift.io/*` annotations.

alanconway:
@jcantrill quick re-review please, I think all your comments are addressed.

alanconway:
/unhold
@jcantrill can you LGTM this? Your notes are incorporated.

jcantrill (commented):
/lgtm

PR #920, opened 2021-10-05
ETCD-236: etcd: scaling etcd with raft learners

hexfusion:
This PR outlines an enhancement to change scale-up operations of the etcd cluster to use raft learners. This work will assist with `Control Plane Scaling and Recovery`[1].

POC: https://github.com/openshift/cluster-etcd-operator/pull/682

[1] https://issues.redhat.com/browse/OCPPLAN-5712


enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+
+1. As a cluster-admin I want to scale up etcd without fear of quorum loss.
+
+2. As a cluster-admin I want to be able to replace failed control-place nodes without a deep understanding of etcd.
+
+### Critical Alerts
+
+We will add the following critical alerts:
+
+- alert about learner member which has not been promoted or started > 30s
```
lilic, line 88: Is 30 seconds really the right threshold? We every "operator interval" initiate the learner promotion so should be at least "2 x operator interval" I would say. Also, I would not say critical more like a warning alert to help debug issues, we will already alert if the cluster has etcd errors and latency which should cover these issues anyways, no?

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+
+This is where to call out areas of the design that require closure before deciding to implement the
+design.  For instance, > 1. This requires exposing previously private resources which contain
+sensitive information.  Can we do this?
+
+### Test Plan
+
+- e2e with
+
+    1. add dangling etcd learner member which has not been started and not promoted.
```
lilic, line 214: Curious how will we prevent a learner from being promoted? Otherwise sounds great!

lilic (commented):
Looking good! 🎉

enhancements/etcd/scaling-etcd-with-raft-learners.md
hexfusion: Yeah MTTR of 7GB data file could be closer to 2m. So yes I would say this is an arbitrary number (30s) that needs additional thought maybe something longer ~5m. The general observation is that members should quickly be promoted and if not it points to a possible scaling issue.

hexfusion: We might want 2 seperate observations

- learner not started (this should happen as fast as the installer pod can add a new etcd spec)
- learner not promoted (this really can be depending on the size of the state file and the speed of networking)

enhancements/etcd/scaling-etcd-with-raft-learners.md
hexfusion: One idea is to use IPTables to block peer traffic` tcp` port `2380`.

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+### Goals
+
+- provide safe scale up of etcd cluster using raft learners.
+- add additional membership level details such as member ID to the init container validation process for etcd.
+- add observability to allow admin to clearly diagnose scaling failure.
+- reduce divergence from upstream etcd by removing `discover-etcd-initial-cluster`.
+
+### Non-Goals
+
+- implement scale down logic for the `cluster-etcd-operator`
```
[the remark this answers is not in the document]
hexfusion: I think the admin should have this choice as long as performance implications are outlined and proper metrics exposed. For example, a follow-up to this question is what metric exposes the duration leader snapshot takes to scale up learner. During bootstrap this is almost 0 as there is little or no data. But in a production cluster this state can be GBs.

enhancements/etcd/scaling-etcd-with-raft-learners.md
lilic: I see we have the learner_promote_failures and learner_promote_successes, we can just do a rate error alert there, which can be critical, wdyt?

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+1. As a cluster-admin I want to scale up etcd without fear of quorum loss.
+
+2. As a cluster-admin I want to be able to replace failed control-place nodes without a deep understanding of etcd.
+
+### Critical Alerts
+
+We will add the following critical alerts:
+
+- alert about learner member which has not been promoted or started > 30s
+- alert if the number of etcd members is not equal to the number of quorum-guard pods > 30s
```
JoelSpeed, line 89: Is there a defined term (perhaps we can create one) for talking about all members vs only those which are voting members?

Does etcd quorum guard need to cover the learner? Or does that only need to to cover the member once it is promotable?

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+and received a snapshot from the leader including the cluster membership details from the member bucket the new etcd
+member must be able to communicate with its peers. This proposal includes moving population of
+`ETCD_INITIAL_CLUSTER` from etcd itself via `discover-etcd-initial-cluster`[1] to the `etcdenvvar controller` in the
+`cluster-etcd-operator`.
+
+[1] https://github.com/openshift/etcd/tree/openshift-4.10/openshift-tools/pkg/discover-etcd-initial-cluster
+
+### cluster-etcd-operator verify membership command
+
+In 4.9 the `verify` command was added to `cluster-etcd-operator` as way to ensure that before a backup is taken
```
JoelSpeed, line 113: Nit
```suggestion
In 4.9 the `verify` command was added to `cluster-etcd-operator` as a way to ensure that before a backup is taken
```

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+2. `ETCD_MEMBER_IDS`: This variable will be added to the etcd static pod EMV, the mapping of this data enables extra
+precautions during scaling.
+
+- Writer: new function added to `envVarFns`[1].
+
+- Consumer: cluster-etcd-operator verify membership command.
+
+New etcd members on an existing node will need to archive the data-dir of the previous member to allow scale up to
+occur. `verify membership` can observe this by asking the cluster for membership details and using the conditions of
+`Name=""` and the method of `Member` `IsLearner` to conclude this.
```
JoelSpeed, line 150: Is this supported? I was under the impression that we had a 1:1 mapping of etcd member to instance, and that, if removed from the cluster, we expect the node to go away. This could be a misunderstanding on my part, but if this is a change in behaviour from today I think this needs some more explanation on motivation and the risks around this change

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+
+The largest risks to scaling is quorum loss, data loss, and split brain scenario.
+
+- **Quorum Loss**: Because new members are added as non voting members and can not be promoted to voting members unless
+the etcd process starts and the log has been completely replicated and in sync with leader.
+
+
+- **Split brain**: The cluster gates starting of the etcd process on a verification process based on the cluster
+member id. This ensures that each revision has an explicit expected membership. Because quorum guard replicas are
+managed by the operator the cluster topology will remain in a safe configuration (odd number of members).
```
JoelSpeed, line 193: Does `odd number of members` hold true? There was a comment above about ensuring that num etcd replicas == num quorum guard, but this may not always be odd when there's a learner "learning"

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+
+
+- **Data loss**: A concern with rolling scaling of etcd with large data files is possible data loss. The MTTR of log
+replication is directly tied to the size of the state. If members of the cluster were replaced too quickly in
```
JoelSpeed, line 197: > The MTTR of log replication is directly tied to the size of the state.

If you happen to have any figures that show etcd size vs MTTR, this would be useful context for understanding this risk

enhancements/etcd/scaling-etcd-with-raft-learners.md
lilic: Agreed. Rate error alert with % of errors is the standard for these kinds of things, we already have those kinds of alerts for other components.

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+
+## Motivation
+
+Great demand exists to provide flexible automated scaling of the control-plane in the same way the worker
+nodes do today.
+
+### Goals
+
+- provide safe scale up of etcd cluster using raft learners.
+- add additional membership level details such as member ID to the init container validation process for etcd.
```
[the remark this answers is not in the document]
hexfusion: good point will adjust

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+stability issues during scaling. A learner is essentially an etcd member which is non voting thus can not impact 
+quorum but like other members will receive log replications from the leader.
+
+This enhancement proposes:
+- Replacing the default scale up performed by the cluster-etcd-operator to use raft learners.
+- Deprecation and removal of the current `discover-etcd-initial-cluster`[3] command and replacing its
+functionality with the existing `etcdenvvar` controller and subcommands of the `cluster-etcd-operator`. By
+eliminating this code it also removes the teams largest carry against upstream etcd.
+- Ensuring that scale up only happens on control-plane nodes which have a quorum-guard pod monitoring health.
+- Adding a flag to etcd which allows configuration of maximum learners in cluster (--max-learners). Today the max is 1.
```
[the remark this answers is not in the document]
JoelSpeed: So, I'm wondering what happens in a scenario where two new "masters" are added to the cluster at once. Say we have a healthy 3 node control plane, 2 new etcd static pods start. They now both want to become learners, but only one is allowed (due to the max-learners flag). What happens to the one that isn't allowed to be a learner?

enhancements/etcd/scaling-etcd-with-raft-learners.md
JoelSpeed: Cool, that makes sense, might be worth adding that context into the enhancement as it wasn't obvious to me why learners still need quorum guard

enhancements/etcd/scaling-etcd-with-raft-learners.md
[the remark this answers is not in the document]
hasbro17: @JoelSpeed I'm guessing first one past the post wins/gets added as a learner:
https://etcd.io/docs/v3.4/op-guide/runtime-configuration/#error-cases-when-adding-a-learner-member
For the other one I imagine the etcd-operator would probably keep retrying to add it as a learner and failing to do so.

But to your point of 2 static pods, do you mean on a single node? I'm wondering how that scenario might arise.

Since the targetconfigcontroller is just [applying static pod configurations](https://github.com/openshift/cluster-etcd-operator/blob/10c3297019a692f714b7766872b26c67238038d9/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L194-L203) via the configmap, I'm assuming the library-go static pod controller would just create a single manifest in the static pod directory for that node.
But if something goes wrong and we do somehow end up with two identical etcd static pods on a single node, I'm not sure if the etcd-operator can fix that situation by just removing that extra member or try to directly update static pod manifests on the node. That's probably a library-go issue at that point right @hexfusion ?

Although we probably should account for that possibility by recognizing that we've reached the max number of learner members from the error returned.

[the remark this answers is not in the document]
hexfusion: This is a good question I will add an addition of a matrix that describes these conditions and expected results.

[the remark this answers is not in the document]
JoelSpeed: > We'll probably get an alert if the 2nd etcd pod doesn't get added as a learner member after some time, and if the 1st one that is a learner doesn't get promoted in time to make room for the 2nd one.

Probably is good, definite is better 😉  We should check if these alerts exist, check whether that's how we would want to report this, and add the new alerts/make a note they exist already to the proposal as appropriate 😄

enhancements/etcd/scaling-etcd-with-raft-learners.md
JoelSpeed: > islearner for > 5m

Does this not depend on the size of the etcd data?

enhancements/etcd/scaling-etcd-with-raft-learners.md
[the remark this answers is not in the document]
jewzaam: If 2 members were added to an existing 3 member cluster wouldn't it be intentional?  Can the number of expected learners be changed dynamically and what is the impact of that change?  If so, then managing max learners is an operational concern and blocking additional learners and raising an alert is the expected behavior.

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+[3] https://github.com/openshift/etcd/blob/openshift-4.10/openshift-tools/pkg/discover-etcd-initial-cluster/initial-cluster.go
+
+## Motivation
+
+Great demand exists to provide flexible automated scaling of the control-plane in the same way the worker
+nodes do today.
+
+### Goals
+
+- provide safe scale up of etcd cluster using raft learners.
```
jewzaam, line 67: Vertical scaling is a capability we need for managed platforms.  Is this scale up in support of changing instance types for the underlying master?  And can be followed by a graceful and clean scale down of older masters?

enhancements/etcd/scaling-etcd-with-raft-learners.md
```diff
@@ -0,0 +1,343 @@
...
+
+### Non-Goals
+
+- implement scale down logic for the `cluster-etcd-operator`
+
+## Proposal
+
+### User Stories
+
+1. As a cluster-admin I want to scale up etcd without fear of quorum loss.
```
jewzaam, line 80: I'd like to see a distinction between horizontal and vertical scaling and clarity on scope for this enhancement.  For SD and managed platforms we need vertical scaling by way of changing master instance type.  Or a way to scale horizontally up and then down such that we achieve that as the end result.

hexfusion: Understood I will ensure this describes vertical scaling explicitly. I am trying to not make this enhancement too large so I have not included all of the machine specific implementation details in this enhancement.

PR #923, opened 2021-10-06
 [docs] remove invalid web links

selansen:
Few weblink references in Implementation History section are expired or
 removed. so we now get '404 not found error'. Hence removing the
 references from the document.

Signed-off-by: selansen <esiva@redhat.com>

### DIFF
--- a/enhancements/windows-containers/windows-machine-config-operator.md
+++ b/enhancements/windows-containers/windows-machine-config-operator.md
@@ -158,9 +158,7 @@ We plan to maintain kubelet major version parity with the Linux counterpart.
 ## Implementation History
 
 We already have a working [prototype](https://github.com/openshift/windows-machine-config-operator)
-in place where use a custom [CRD](https://github.com/openshift/windows-machine-config-operator/blob/master/deploy/crds/wmc.openshift.io_windowsmachineconfigs_crd.yaml)
-and the [Windows Node Installer](https://github.com/openshift/windows-machine-config-bootstrapper/tree/master/tools/windows-node-installer)
-library to enable Windows workloads.
+in place where use a custom CRD and the Windows Node Installer library to enable Windows workloads.
 
 ## Drawbacks
 


dhellmann:
Are the targets of the links in new locations, so that we could update the links instead of removing them?

enhancements/windows-containers/windows-machine-config-operator.md
```diff
@@ -158,9 +158,7 @@ We plan to maintain kubelet major version parity with the Linux counterpart.
 ## Implementation History
 
 We already have a working [prototype](https://github.com/openshift/windows-machine-config-operator)
-in place where use a custom [CRD](https://github.com/openshift/windows-machine-config-operator/blob/master/deploy/crds/wmc.openshift.io_windowsmachineconfigs_crd.yaml)
-and the [Windows Node Installer](https://github.com/openshift/windows-machine-config-bootstrapper/tree/master/tools/windows-node-installer)
-library to enable Windows workloads.
+in place where use a custom CRD and the Windows Node Installer library to enable Windows workloads.
```
sebsoto, line 161: You've removed the dead links but the information still isnt accurate. We have not had a CRD or used WNI for a long time. Considering this is a history section, I'm not fully sure what should be present here. Perhaps reasoning why we no longer do these things.

selansen: wmc.openshift.io_windowsmachineconfigs_crd.yaml - This file is completely removed by PR https://github.com/openshift/windows-machine-config-operator/pull/81/files
So either we can remove the link or I can leave it as it is ( as this is part of History ).
Same thing will apply to another link https://github.com/openshift/windows-machine-config-bootstrapper/tree/master/tools/windows-node-installer
ABove folder is removed from the repo. It's not move to any other place. So its again a dead link.
This is done in below commit. 
https://github.com/openshift/windows-machine-config-bootstrapper/commit/ed7c5af4b2b1f64832be7bc7285060243e9dff69

Let me know how do you folks want to handle it. I can close this PR as there is no new link that I can use to replace the old link.

aravindhp: As we spoke about before, please link to the older commits in history rather than just removing them.

selansen: will reopen this bug

selansen: /reopen

PR #922, opened 2021-10-05
Allow installer to include/exclude capabilities based on user selections

bparees:

PR #924, opened 2021-10-06
CLI Manager

deejross:
Provide a way to distribute CLI tools and plugins with a krew-compatible API for disconnected environments.


enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,221 @@
...
+cli-artifacts and oc download links that do not require anything outside the cluster.  The goal is to provide the same for other
+binaries/tools/CLIs. Also, these artifacts will be accessible via `oc` as well as through the console via a fileserver.
+
+`oc` will provide the mechanism to provide, list, install, and upgrade OpenShift supported CLIs compatible with   
+a connected cluster. Through a ClusterCLI [Custom Resource](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/),
+each CLI will provide necessary information about its provided artifacts, including its description and location within the image.  The ClusterCLI
+for oc, helm, and odo custom resources as well as a CRD controller to manage these will serve as a reference consumer implementation for other CLIs to 
+follow. Currently. the `cli-artifacts` image stream in the `openshift namespace` is shipped with the release payload. This is where the artifacts
+for `oc` will be extracted from. If other CLIs choose to offer artifacts in this way, they can opt in by supplying an artifacts image to be
+shipped with the release payload.
```
dhellmann, line 45: Placing them in the release payload ties them to the release schedule. Is that what we want?

We're actively trying to limit the growth of the release payload, so if there are ways to achieve the goals of this enhancement that avoids adding anything we should consider those approaches.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,221 @@
...
+Some current examples are oc, kubectl, odo, istio, tekton, and knative.  It is difficult for users to discover what tools exist, 
+where to download them from and which version they should download.  We need to simplify as much as possible the interaction
+of services on OpenShift.  We need a mechanism for providing and consuming tools that is simple to add on to as new CLIs are 
+developed from a variety of sources - and this should be specific for each cluster and available with disconnected installs.
+
+## Requirements:
+
+1. No new form of binary distribution or binary creation will be proposed, because we have an existing structure at Red Hat.
+RPMs or images are the only options, and images must be deployed by the RH pipeline via operators. This proposal is for delivering
+CLIs via images, because this will enable offering CLIs offline through mirroring.
```
dhellmann, line 61: Is the requirement that clients be available using mirroring, or is that an implementation choice?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,221 @@
...
+2. CLI owners must be able to easily distribute their binaries.
+3. The version of the CLI a user is offered is appropriate for the version of the CLI controller installed on the cluster.
+4. Arbitrary binaries not delivered by the in-cluster CRD controller are not important or relevant because of requirement 1 and 2.
+    - Anyone who wants such tools can download them outside of this mechanism.
+5. Users must continue to have the option to download binaries from either the central location where artifacts are published or from the console.
+6. CLI author requirements:
```
dhellmann, line 67: This item sounds like implementation details rather than requirements for the change being proposed.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Motivation
+
+As more services are created on top of OpenShift, more CLIs are introduced to simplify interaction with these services.
+Some current examples are oc, kubectl, odo, istio, tekton, and knative.  It is difficult for users to discover what tools exist, 
+where to download them from and which version they should download.  We need to simplify as much as possible the interaction
+of services on OpenShift.  We need a mechanism for providing and consuming tools that is simple to add on to as new CLIs are 
+developed from a variety of sources - and this should be specific for each cluster and available with disconnected installs.
+
+## Requirements:
```
dhellmann, line 57: The structure of this enhancement is different from our standard format. These requirements should probably move under the Goals heading.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,221 @@
...
+2. CLI owners must be able to easily distribute their binaries.
+3. The version of the CLI a user is offered is appropriate for the version of the CLI controller installed on the cluster.
+4. Arbitrary binaries not delivered by the in-cluster CRD controller are not important or relevant because of requirement 1 and 2.
+    - Anyone who wants such tools can download them outside of this mechanism.
+5. Users must continue to have the option to download binaries from either the central location where artifacts are published or from the console.
+6. CLI author requirements:
+    - Provide CLI image containing binaries
+    - ClusterCLI Custom Resource that describes the binaries (location within the image, metadata, description)
+    - Registering and managing a ClusterCLI Custom Resource through the CRD controller. 
+7. `oc` changes:
```
dhellmann, line 71: This is also an implementation detail.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,221 @@
...
+
+- CLIs are expected to be backwards compatible.  When working with multiple clusters, it's expected that CLI versions will work across cluster versions.
+If this is not the case, CLI owners will provide that information in the Custom Resource description.  
+- A cluster will supply known-good versions of supported CLIs through cli-manager for a particular cluster, but the expectation is that a CLI version
+will be good for multiple cluster versions, and that it is rare for an admin of multiple clusters to require different versions of a single CLI.
+
+## Infrastructure Needed
+
+- CRD controller - for installing ClusterCLI CRD
+- CRD and API [example](https://github.com/deejross/openshift-cli-manager/blob/main/config/crd/bases/config.openshift.io_clitools.yaml)
```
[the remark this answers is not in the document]
deejross: I haven't gotten to distribution yet. I was hoping to get some guidance in this area.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,221 @@
...
+can deliver tools to OpenShift clusters.  A feature is needed to manage various CLIs available for OpenShift and related services.  The goal is for
+ a connected user to discover, install, and upgrade tools that are compatible with the current cluster version easily and from a single location.
+
+Each component is responsible for building and publishing its artifacts and registering information regarding supplied binaries.
+Currently, that location is [index of /pub/openshift-v4/clients](https://mirror.openshift.com/pub/openshift-v4/clients/),
+and this makes it difficult for disconnected installations to mirror them.  
+
+`oc` will retrieve binaries from images that package a CLI's artifacts. Currently, we provide disconnected environments 
+cli-artifacts and oc download links that do not require anything outside the cluster.  The goal is to provide the same for other
+binaries/tools/CLIs. Also, these artifacts will be accessible via `oc` as well as through the console via a fileserver.
```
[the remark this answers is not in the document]
dhellmann: OK, there's a lot of discussion about the console in this enhancement, so it wasn't clear that those changes were not also part of what's being proposed.

enhancements/oc/cli-manager.md
dhellmann: The proposal says

> This proposal is for delivering CLIs via images, because this will enable offering CLIs offline through mirroring.

That combines 2 things: the delivery format and mirroring. Do we have a requirement that CLIs be delivered by images, and as a result of that we get mirroring? Or do we have a requirement that CLI packages be mirrorable, and therefore we're choosing to use images?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,221 @@
...
+    - Anyone who wants such tools can download them outside of this mechanism.
+5. Users must continue to have the option to download binaries from either the central location where artifacts are published or from the console.
+6. CLI author requirements:
+    - Provide CLI image containing binaries
+    - ClusterCLI Custom Resource that describes the binaries (location within the image, metadata, description)
+    - Registering and managing a ClusterCLI Custom Resource through the CRD controller. 
+7. `oc` changes:
+    - Reading metadata from CR and extracting the binaries from the CRD controller's API to local disk (disconnected)
+8. CRD controller for registering CRs with an API for listing, extracting and downloading tools
+    - Krew-compatible index that both krew and oc can consume where the download links are for cluster-local resources
```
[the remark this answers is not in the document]
dhellmann: OK, it would be good if that could be clarified in the text.

enhancements/oc/cli-manager.md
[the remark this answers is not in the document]
dhellmann: We should start by assuming this new operator will be delivered via OLM and see how far we can get with it. Does that impact the planned console integration, for example, or can we use a console plugin? Does it change anything about how we deliver `oc`?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+
+As owner of a CLI tool, I want to publish my CLI tool to users of the cluster. I need to create a CR for my tool, or provide the required information about my tool
+to a cluster-admin for the creation of a CR:
+* Name
+* Short description
+* Long description
+* Caveats
+* Homepage
+* Version
+* Platform/architecture
```
dhellmann, line 156: This field implies that perhaps I could have 2 CRs for the same command for different architectures? Is the name of the CR significant, or does `tool install` look only at the path field?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+* `oc tools list` will access the connected OpenShift cluster, if connected,  and will retrieve information about available CLIs (ClusterCLI CRs)
+* `oc tools install odo` will install `odo` to a user's home directory, as a standalone binary.
+
+##### Example
+```text
+$ oc tools list
+NAME                 DESC                        LATEST       INSTALLED
+-----                -----                       -----        -----
+kubectl              Kubernetes CLI              1.15         1.13
+oc                   OpenShift CLI               4.4          4.3
```
dhellmann, line 122: How will `oc` know which version of the tool is installed locally? Will it ask each of them to report their version? Where does the info from the latest column come from?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+#### Story 2
+
+As a user, I want to access various CLIs from the OpenShift Console Web Terminal so that I can use the
+services available in the cluster. The CLIs versions should match the those of the operators deployed
+on the cluster.
+
+Currently, users can access `consoleclidownloads` custom resources for helm, oc, and odo either from oc with `oc get consoleclidownloads` or through
+the `downloads -n openshift-console` route. This access will still be available, through the updated ClusterCLI custom resources. The ClusterCLI custom
```
dhellmann, line 136: Does this mean that newer versions of oc will use the new CRD instead of the old mechanism? Is there a version skew concern with that change, or will the old mechanism work for old clients?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+### Risks and Mitigations
+
+Distributing undesirable binaries is always a risk. Some mitigations include requiring cluster-admins to maintain the index, and the verification of downloaded
+binaries using SHA256 hashes.
+
+## Design Details
+
+Each CLI will provide an image.
+Each CLI is responsible for creating a CR to hold metadata.  The CR will serve to deliver the metadata and description
+of its deliverable binary.  Initially, CRs will be accessed using `oc tools` command, and `oc tools install <cli>` will extract the binary to a user's home directory.
```
dhellmann, line 169: Will `oc tools install oc` work?

enhancements/oc/cli-manager.md
deejross: I was going to put a note there for clarification, but thought it better to be in a design details section: A single CR supports multiple versions, and each version supports multiple platforms (os/arch). The name of the CR is used for `tool install`.

enhancements/oc/cli-manager.md
deejross: Great questions...keeping track of locally installed version was an interesting problem. Since a SHA256 hash is kept on every version/os/arch binary, `oc` can simply hash the binary locally and query the API with that digest to get the current version. The digests are stored in the CRs status field, so as long as no one removes those, `oc` should always know what version is installed.

In the CR, it is expected that the list of versions provided is in order from oldest to newest, so the latest is always at the end of the list. Attempting to parse version numbers seemed too problematic given how different projects handle versioning.

enhancements/oc/cli-manager.md
deejross: The old mechanism isn't going away yet (maybe never). The current `oc` command to extract binaries from those CRs isn't changing either. This would add a new subcommand to `oc` that will be used exclusively for the new CRD.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+odo                  OpenShift Developer CLI     1.0          Not Installed
+kn                   Knative CLI                 1.0          Not Installed
+tkn                  Tekton CLI                  1.0          Not Installed
+
+```
+
+#### Story 2
+
+As a user, I want to access various CLIs from the OpenShift Console Web Terminal so that I can use the
+services available in the cluster. The CLIs versions should match the those of the operators deployed
```
cfergeau, line 132: > should match the those

There is an extra "the" here, or some missing words

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+
+* Krew and Krew plugins are upstream projects that Kubernetes users are already familiar with
+* The index will be served by the CRD controller with its contents managed by cluster-admins via CRs
+* A Krew-compatible index can provide available CLIs and plugins for a cluster (similar to the current index) and can also provide an easy mechanism for installing through `oc tools`
+* An openshift-krew-index with supported artifacts could hold information about CLIs supported by OpenShift.
+* With `krew index add https://someother-third-party-index` we won't limit cluster-admins from adding their own index with whatever plugins they want. Only the openshift-krew-index will be supported.
+
+Users will continue to have the option to download binaries from either the central location where artifacts are published or from the console.
+This proposal adds the mechanism to install from images built from the same artifacts using `oc`. All options will offer the same binaries,
+ensured through sha256 signatures.
```
cfergeau, line 97: sha256 is a hash/a checksum, I don't think it should/can be called a signature. "SHA256 hashes" is used later in the document.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+* Homepage
+* Version
+* Platform/architecture
+* The image:tag (and registry credentials if required)
+* The path within the image where the binary for the given platform/architecture can be found
+
+### Risks and Mitigations
+
+Distributing undesirable binaries is always a risk. Some mitigations include requiring cluster-admins to maintain the index, and the verification of downloaded
+binaries using SHA256 hashes.
```
cfergeau, line 163: Verifying SHA256 hashes is only good if these hashes can be trusted (for example because they are signed with a RH-owned key). Might be useful to clarify this if this can be done in a few words.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+of services on OpenShift.  We need a mechanism for providing and consuming tools that is simple to add on to as new CLIs are
+developed from a variety of sources - and this should be specific for each cluster and available with disconnected installs.
+
+### Goals
+
+* No new form of binary distribution or binary creation will be proposed, because we have an existing structure at Red Hat.
+RPMs or images are the only options, and images must be deployed by the RH pipeline via operators. This proposal is for delivering
+CLIs via images, because this will enable offering CLIs offline through an existing image registry.
+* CLI owners must be able to easily distribute their binaries
+* The version of the CLI a user is offered is appropriate for the version of the CLI controller installed on the cluster
```
soltysh, line 63: Why a CLI is bound to a controller version? You should be able to ship any CLI at any level you desire.

soltysh: Unless you're talking here about specific CLI, I assume oc, in which case they reflect the OCP version installed, rather than controller and I'd suggest being explicit in such cases.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+* Arbitrary binaries not delivered by the in-cluster CRD controller are not important or relevant because of requirement 1 and 2
+  * Anyone who wants such tools can download them outside of this mechanism
+* CRD controller for registering CRs with an API for listing, extracting and downloading artifacts
+  * Krew-compatible index that both krew and oc can consume where the download links are for cluster-local resources
+  * The index and artifcats will served from a single route and service within the cluster
+
+### Non-Goals
+
+* `oc tools` will not build or serve the binaries.  It will know where to find them.
```
soltysh, line 72: Hmmm.. how does `oc tool` is similar with/different from `oc plugin`?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+developed from a variety of sources - and this should be specific for each cluster and available with disconnected installs.
+
+### Goals
+
+* No new form of binary distribution or binary creation will be proposed, because we have an existing structure at Red Hat.
+RPMs or images are the only options, and images must be deployed by the RH pipeline via operators. This proposal is for delivering
+CLIs via images, because this will enable offering CLIs offline through an existing image registry.
+* CLI owners must be able to easily distribute their binaries
+* The version of the CLI a user is offered is appropriate for the version of the CLI controller installed on the cluster
+* Arbitrary binaries not delivered by the in-cluster CRD controller are not important or relevant because of requirement 1 and 2
```
soltysh, line 64: The "built-in" binaries which are part of cli-artifacts image, I'd call them shipped with OCP, not delivered by the in-cluster CRD controller, this is confusing. The controller you're describing here is a mechanism for delivering those binaries.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+  * Anyone who wants such tools can download them outside of this mechanism
+* CRD controller for registering CRs with an API for listing, extracting and downloading artifacts
+  * Krew-compatible index that both krew and oc can consume where the download links are for cluster-local resources
+  * The index and artifcats will served from a single route and service within the cluster
+
+### Non-Goals
+
+* `oc tools` will not build or serve the binaries.  It will know where to find them.
+* This proposal is not concerned with _which_ binaries will be managed.  This proposal is meant to determine the mechanism only.  Consumers and publishers are clients of the mechanism.
```
soltysh, line 73: That sounds to me more like a goal

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+* `oc tools` will not create or update the ClusterCLI Custom Resources, that will be managed by individual CLIs. However, the CLIs that are
+currently available as `consoleclidownloads`, `oc, odo, and helm`, will be included in the reference implementation of this enhancement.
+
+## Proposal
+
+Each component wishing to provide customers with their binaries will build and publish artifacts via an official channel to a central index.
```
soltysh, line 79: This is a good non-goal, because we don't want to deal with how binaries are shipped through images. We only care about the end result, ie. the image.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+Each component wishing to provide customers with binaries will create a ClusterCLI custom resource to provide an image name and the file path
+within that image for the binaries. The CLIs currently offered via console-cli-downloads will be included in the reference implementation of this enhancement.
+`oc` will gather ClusterCLI from the cluster, and also the fileserver will offer the binaries via the console by mounting the binaries
+from the location/image in each custom resource.
+
+CLIs could be installed in a similar manner to plugins through [krew](https://github.com/kubernetes-sigs/krew).
+CLI images would serve the artifacts to enable disconnected downloads, and OpenShift CLIs will be available as
```
soltysh, line 86: images don't serve artifacts, they are just a shipping mechanism.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+[krew plugins](https://github.com/kubernetes-sigs/krew-index/tree/master/plugins).
+
+* Krew and Krew plugins are upstream projects that Kubernetes users are already familiar with
```
soltysh, line 89: I'd suggest the proposal section be shaped in the following way:
1. introduce what krew is and how it fits the image of delivering plugins based on the index metadata, and most importantly why we picked krew instead of creating our own mechanism 
2. describe how we'll serve the index metadata for 1 inside the cluster
3. describe how a binary author is meant to have his binary served by 2. 

First describe the process from a high level, then create a detailed section for each of them drilling down into implementation details.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
...
+
+### User Stories
+
+#### Story 1
+
+As a user, I want a CLI manager for various CLIs available for OpenShift and related services so that I can discover, install and list them
+The user will invoke the following commands:
+
+* `oc tools list` will access the connected OpenShift cluster, if connected,  and will retrieve information about available CLIs (ClusterCLI CRs)
+* `oc tools install odo` will install `odo` to a user's home directory, as a standalone binary.
```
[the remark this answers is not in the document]
soltysh: this is a good point for krew, iirc it does that too already, so you should use that as argument for krew, as mentioned above.

enhancements/oc/cli-manager.md
soltysh: Either explain the questions popping up in proposal section, or add a questions section towards the end of the proposal, It's fine and good to even leave some questions unaswered

soltysh: Also, doesn't krew track that information for you?

enhancements/oc/cli-manager.md
deejross: Clarified in the next commit

enhancements/oc/cli-manager.md
deejross: I reworded that whole line for the next commit to hopefully make it more clear

enhancements/oc/cli-manager.md
deejross: Moved

enhancements/oc/cli-manager.md
deejross: Will change this to use `oc plugin` instead as that makes more sense

enhancements/oc/cli-manager.md
deejross: That looks like a holdover from the original proposal, removed and clarified

enhancements/oc/cli-manager.md
deejross: Clarified

enhancements/oc/cli-manager.md
deejross: Updated. Thank you.

enhancements/oc/cli-manager.md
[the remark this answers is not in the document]
deejross: Are you saying we should recommend using `krew` over `oc plugin`?

enhancements/oc/cli-manager.md
deejross: I've updated to hopefully clarify that it's the responsibility of the cluster-admins to publish trusted binaries.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
+---
+title: cli-manager
+authors:
+  - "@sallyom"
+  - "@deejross"
+reviewers:
+  - "@soltysh"
+approvers:
```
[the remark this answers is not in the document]
deejross: Done

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,252 @@
+---
+title: cli-manager
+authors:
+  - "@sallyom"
+  - "@deejross"
+reviewers:
```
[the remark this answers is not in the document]
deejross: Done

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,211 @@
...
+Each component wishing to provide customers with their plugins will build and publish images via a trusted image registry
+and create a Plugin CR to provide an image name and the file path within that image for the binaries.
+Clients (i.e. `krew`) will read the index and download binaries from the Controller. The Controller is responsible for building the index from CRs, for pulling the images,
+extracting the binaries, and serving them to clients.
+
+* `krew` and `krew` plugins are upstream projects that Kubernetes users are already familiar with
+* A `krew`-compatible custom index can provide available plugins for a cluster in disconnected environments
+* The index will be served by a Controller with its contents managed by cluster-admins via CRs
+* The binaries will also be served by the Controller that will pull images from a trusted image registry and extract the binaries
+* With `krew index add https://someother-third-party-index` we won't limit users from adding their own index with whatever plugins they want
```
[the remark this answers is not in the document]
soltysh: Yes, we are considering having a set built-in indexes that will be present on all clusters. The only real question is how to deal with those in connected vs disconnected envs, but if we can make krew to fail silently and ignore not available index w/o too much time delay that should work. I'd update this doc with additional finding that will be part of the first POC.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,211 @@
...
+* As `krew` itself is a `kubectl` plugin, it can be invoked using either using `kubectl krew` or `oc krew`
```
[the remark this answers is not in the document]
soltysh: @deejross i guess this is not addressed, but yeah we want to embed krew in oc by default.

enhancements/oc/cli-manager.md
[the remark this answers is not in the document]
soltysh: @deejross while adding final updates, maybe worth mentioning that we'll embed those 3 indexes.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,212 @@
...
+
+### User Stories
+
+#### Story 1
+
+As a user, I want a CLI manager for various CLIs and plugins available for Kubernetes/OpenShift and related services so that I can discover, install, and list them. If `odo` was made available by a cluster-admin, I could install it using:
+* `oc krew install odo`
+
+I could then interact with `odo`:
+* `oc odo --help`
```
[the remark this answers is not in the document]
soltysh: The initial phase will be covering oc plugins, only. I'm envisioning addressing concerns from @deboer-tim and @jwmatthews at later stages.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,211 @@
...
+
+##### Example
+```text
+$ oc krew search odo
+NAME                 DESC                        LATEST       INSTALLED
+-----                -----                       -----        -----
+odo                  OpenShift Developer CLI     1.0          Not Installed
+```
+
+#### Story 2
```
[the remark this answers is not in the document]
soltysh: A good option I considered is that operators would inject their binaries information into that CR during installation. This would allow admins to have a pre-populated list that he can later modify.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,211 @@
...
+* Homepage
+* Version
+* Platform/architecture
+* The image:tag (and registry credentials if required)
+* The path within the image where the binary for the given platform/architecture can be found
+
+### Risks and Mitigations
+
+Distributing undesirable binaries is always a risk. Some mitigations include requiring cluster-admins to maintain the index, and the verification of downloaded
+binaries using SHA256 hashes. Cluster-admins are responsible for publishing only trusted binaries.
```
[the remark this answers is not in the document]
soltysh: Probably something we could help add in krew.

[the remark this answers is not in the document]
soltysh: @deejross please add in followup section.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,212 @@
...
+```text
+$ oc krew search odo
+NAME                 DESC                        LATEST       INSTALLED
+-----                -----                       -----        -----
+odo                  OpenShift Developer CLI     1.0          Not Installed
+```
+
+#### Story 2
+
+As owner of a CLI or plugin, I want to publish it to users of the cluster. I need to create a Plugin CR for my tool, or provide the required information about my tool
```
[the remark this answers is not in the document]
soltysh: @deejross add a followup/further development section and mention it there

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,231 @@
...
+and create a Plugin CR to provide an image name and the file path within that image for the binaries.
+Clients (i.e. `krew`) will read the index and download binaries from the Controller. The Controller is responsible for building the index from CRs, for pulling the images,
+extracting the binaries, and serving them to clients.
+
+* `krew` and `krew` plugins are upstream projects that Kubernetes users are already familiar with
+* A `krew`-compatible custom index can provide available plugins for a cluster in disconnected environments
+* The index will be served by a Controller with its contents managed by cluster-admins via CRs
+* The binaries will also be served by the Controller that will pull images from a trusted image registry and extract the binaries
+* With `krew index add https://someother-third-party-index` we won't limit users from adding their own index with whatever plugins they want
+* The controller could be optional, since connected environments can use the default `krew` index out-of-the box
```
soltysh, line 88: @deejross the biggest missing element in this proposal is the discussion whether we want that controller to be part of every default OCP installation or rather an optional element. I think it's crucial before we make final decision which direction we want to go. 

Pros for built-in:
- available from day 1 for all clusters
- all binaries including the built-ins can be served using single unified mechanism
Cons for built-in
- additional controller running by default

Pros for opt-in:
- less resource usage by default 
Cons for opt-in:
- yet another operator to manage and mirror, especially in disconnected envs. 
- separate mechanism for built-in and 3rd party binaries
- missing plugin CRs for optional elements installed before this controller is installed. 

I'm personally leaning towards buit-in, and if I don't see any objections I'd like to pursuit that one.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,231 @@
...
+* As `krew` itself is a `kubectl` plugin, it can be invoked using either using `kubectl krew` or `oc krew`
+
+Existing methods of downloading binaries (i.e. the console) will not be affected by this proposal. For the initial implementation, supported plugins will create Plugin CRs. The plugins will be downloaded and installed using `krew`.
```
bparees, line 91: will the console be enabled to serve/show this content?  or you'll have to use the cli?

enhancements/oc/cli-manager.md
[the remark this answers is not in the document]
bparees: in order for "oc odo" to work, doesn't that mean "odo" is already on the user's path?  so "odo" would just work, without any aliasing, right?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,231 @@
...
+* Version
+* Platform/architecture
+* The image:tag (and registry credentials if required)
+* The path within the image where the binary for the given platform/architecture can be found
+
+### API Extensions
+
+A new CRD will be created:
+* API: `config.openshift.io/v1`
+* Kind: `Plugin`
```
bparees, line 129: can you flesh out the rest of this CRD spec?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,231 @@
...
+Each plugin will provide an image.
+Each plugin is responsible for creating a CR to hold metadata.  The CR will serve to deliver the metadata and description
+of its deliverable binary. The Controller will use CRs to generate an index for the `oc krew search` command, and `oc krew install <name>` will download the binary from the Controller.
+Users will install OpenShift tools that are known compatible with each cluster version through `oc krew`.
+
+A plugin must provide a Plugin CR. The result of this proposal will be:
+* Plugin Custom Resource Definition compatible with `krew`
+* Must work with image registries that require image pull secrets
+* Use `krew` manage plugins made available via CRs
+* A Controller to manage plugins that will serve binaries from images
```
bparees, line 147: this implies the controller will be pulling images, this introduces the possibility of a delta in image pulling behavior between what happens if you create a pod that refs the image, and what this controller does, with respect to:

* use of the cluster scoped image pull secret
* respecting ICSP rules for pulling from mirrors
* use of CAs to talk to the registry
* use of proxies to talk to the registry

this has been a long struggle for builds (one of the main things that does its own image pulling on the cluster).  So if you're going to do this, it needs to be done right and then kept in sync as crio/pod image pulling behavior may continue to change in the future (as it did when ICSP was introduced)

enhancements/oc/cli-manager.md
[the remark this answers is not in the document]
sferich888: No because of how the name of the plug-in works, the name is 'kube_odo' or 'oc_odo' so some alias/symlink would be needed to get the name you want, but it's a small price to pay for the delivery mechanisms (IMO)

enhancements/oc/cli-manager.md
soltysh: If we go the built-in route, ideally I'd like for console to be able to re-use that. Iow. we'd replace the current container which serves those binaries from cli-artifacts with this mechanism.

enhancements/oc/cli-manager.md
dhellmann: The bar for adding an "always present" component is very very high. Close enough to insurmountable that we are considering making existing always-present components optional #922. I don't think this is going to meet that bar. For example, consider all of the disconnected deployments, workload-only clusters, and clusters consumed by kcp. Users won't be interacting with those to get their clients.

bparees: other discussion from the slack thread(paraphrased):

maciej: this is replacing an existing component (console cli downloader server) so it's not a net-add to the core

bparees: once we have #922 we could make this component optional, so putting it in the core in 4.11 (when we'll have that capability to disable, hopefully) might be an acceptable middle-ground, with a plan to move it to OLM in the future when OLM is capable of the full "cluster-lifecycling/day0" story.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,233 @@
...
+  - "@soltysh"
+  - "@jwmatthews"
+approvers:
+  - "@soltysh"
+  - "@sferich888"
+  - "@deads2k"
+  - "@spadgett"
+creation-date: 2021-10-06
+last-updated: 2021-10-18
+status: implementable
```
soltysh, line 16: nit: update with latest [template](https://github.com/openshift/enhancements/blob/master/guidelines/enhancement_template.md), I'm especially missing:
```
api-approvers:
  - # probably @deads2k here, pls ping him about it 
tracking-link:
  - https://issues.redhat.com/browse/WRKLDS-169
see-also:
  - "/enhancements/installer/component-selection.md"
```
since the latter is connected with this controller having the ability to be turned off, among others.

enhancements/oc/cli-manager.md
soltysh: @deejross I'm missing the section I've mentioned above describing options and which one was picked any why.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,273 @@
...
+---
+
+# OpenShift CLI Manager
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
```
soltysh, line 24: The above two can be checked, given we're planning to start implementation in the next release.

enhancements/oc/cli-manager.md
soltysh: Alternatively the initial version of API can be approved by me as well, and we'll update later on.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,273 @@
...
+* A `krew`-compatible custom index can provide available plugins for a cluster in disconnected environments
+* The index will be served by a Controller with its contents managed by cluster-admins via CRs
+* The binaries will also be served by the Controller that will pull images from a trusted image registry and extract the binaries
+* With `krew index add https://someother-third-party-index` we won't limit users from adding their own index with whatever plugins they want
+* The controller could be optional, since connected environments can use the default `krew` index out-of-the box
+  * We feel as though the controller should be installed by default so that we can replace the existing temporary solution with something more robust
+  * Pros for being installed by default:
+    * Available day one for all clusters
+    * All binaries can be served using single unified mechanism
+    * No user setup required to use in disconnected environments
```
soltysh, line 93: Replace the existing download mechanism.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,273 @@
...
+  * Cons:
+    * Additional controller running by default
+  * Pros for being optionally installed:
+    * Less resource utilization
+  * Cons:
+    * Another operator to manage and mirror for disconnected environments
+    * Separate mechanism for built-in and third-party binaries
+    * Missing plugin CRs for optional elements installed before controller is installed
```
soltysh, line 101: Won't be able to replace the current built-in mechanism used by the console which is serving some binaries.

enhancements/oc/cli-manager.md
soltysh: @deejross can you replace this sentence and below with something like:
```
* There are two possible approaches for this new controller:
1. Controller should be installed by default...
pros... cons...
2. Controller being optionally installed...
pros... cons...
Given above arguments it is reasonable to include the controller by default.
```

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,273 @@
...
+* As `krew` itself is a `kubectl` plugin, it can be invoked using either using `kubectl krew` or `oc krew`\
+* `krew` functionality baked into `oc` by default
+
+Existing methods of downloading binaries (i.e. the console) will not be affected by this proposal. For the initial implementation, supported plugins will create Plugin CRs. The plugins will be downloaded and installed using `krew`.
```
soltysh, line 105: Given we want to go the default route, the existing mechanism will be replaced with this new one, but it should be transparent for the consumers, except for the CR change.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,273 @@
...
+Users will install OpenShift tools that are known compatible with each cluster version through `oc krew`.
+
+A plugin must provide a Plugin CR. The result of this proposal will be:
+* Plugin Custom Resource Definition compatible with `krew`
+* Must work with image registries that require image pull secrets
+* Use `krew` manage plugins made available via CRs
+* A Controller to manage plugins that will serve binaries from images
+* If possible, Controller should be distributed via OLM
+* Controller should be optional, but enabled by default as it will be replacing the existing mechanism which was intented to be temporary
+  * See: https://github.com/openshift/enhancements/pull/922
```
soltysh, line 180: Change the link from PR to `"/enhancements/installer/component-selection.md"` since the other PR is almost merged.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,273 @@
...
+Each plugin is responsible for creating a CR to hold metadata.  The CR will serve to deliver the metadata and description
+of its deliverable binary. The Controller will use CRs to generate an index for the `oc krew search` command, and `oc krew install <name>` will download the binary from the Controller.
+Users will install OpenShift tools that are known compatible with each cluster version through `oc krew`.
+
+A plugin must provide a Plugin CR. The result of this proposal will be:
+* Plugin Custom Resource Definition compatible with `krew`
+* Must work with image registries that require image pull secrets
+* Use `krew` manage plugins made available via CRs
+* A Controller to manage plugins that will serve binaries from images
+* If possible, Controller should be distributed via OLM
```
soltysh, line 178: This doesn't apply since we go the default route.

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,333 @@
...
+A new CRD will be generated based on the types:
+* API: `config.openshift.io/v1`
+* Kind: `Plugin`
+
+```go
+// PluginSpec defines the desired state of Plugin.
+type PluginSpec struct {
+	// ShortDescription of the plugin.
+	// +required
+	ShortDescription string `json:"shortDescription,omitempty"`
```
soltysh, line 161: Nit: required != omitempty

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,333 @@
...
+	// +optional
+	Caveats string `json:"caveats,omitempty"`
+
+	// Homepage of the plugin.
+	// +optional
+	Homepage string `json:"homepage,omitempty"`
+
+	// Version of the plugin.
+	// +required
+	Version string `json:"versions,omitempty"`
```
soltysh, line 177: Similarly required != omitempty

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,333 @@
...
+
+	// Platforms the plugin supports.
+	Platforms []PluginPlatform `json:"platforms,omitempty"`
+}
+
+// PluginPlatform defines per-OS and per-Arch binaries for the given plugin.
+type PluginPlatform struct {
+	// Platform for the given binary (i.e. linux/amd64, darwin/amd64, windows/amd64).
+	// +required
+	Platform string `json:"platform,omitempty"`
```
soltysh, line 187: required != omitempty

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,333 @@
...
+
+	// Image containing CLI tool.
+	// +required
+	Image string `json:"image,omitempty"`
```
soltysh, line 191: required != omitempty

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,333 @@
...
+
+	// ImagePullSecret to use when connecting to an image registry that requires authentication.
+	// +optional
+	ImagePullSecret string `json:"imagePullSecret,omitempty"`
+
+	// Files is a list of file locations within the image that need to be extracted.
+	// +required
+	Files []FileOperation `json:"path,omitempty"`
```
soltysh, line 199: required != omitempty

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,333 @@
...
+	// Bin specifies the path to the plugin executable.
+	// The path is relative to the root of the installation folder.
+	// The binary will be linked after all FileOperations are executed.
+	// +required
+	Bin string `json:"bin,omitempty"`
+}
+
+// PluginFileOperation specifies a file copying operation from plugin archive to the
+// installation directory.
+type FileOperation struct {
```
soltysh, line 210: Maybe `FileLocation` is better

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,333 @@
...
+	// From is the absolute file path within the image to copy from.
+	// Directories and wildcards are not currently supported.
+	// +required
+	From string `json:"from,omitempty"`
```
soltysh, line 214: required != omitempty same below

soltysh (commented):
A few more nits

enhancements/oc/cli-manager.md
deejross: I updated the code for the POC to remove `omitempty` after I discovered that the CRD generator ignores `// +required` if `omitempty` is present. I'll update this with the latest types.

soltysh (approved):
/lgtm
/approve
/hold 
leaving some time for @bparees @dhellmann to comment, if not I'll drop the hold by end of week

enhancements/oc/cli-manager.md
bparees: i'm not clear why it can't still replace the existing mechanism, though it might mean the console needs to have proper handling of the api/CRs not being present in some cases?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,334 @@
...
+  * Cons:
+    * Additional controller running by default
+  2. Controller being optional:
+  * Pros for being optionally installed:
+    * Less resource utilization
+  * Cons:
+    * Another operator to manage and mirror for disconnected environments
+    * Separate mechanism for built-in and third-party binaries
+    * Missing plugin CRs for optional elements installed before controller is installed
+    * Won't be able to replace existing, temporary system used by console which is serving some binaries
```
bparees, line 110: not clear why it couldn't replace this also.

we don't necessarily need to solve this now, but long term i would still expect this component to be a candidate to move to olm, so we'll need to discuss it then

enhancements/oc/cli-manager.md
bparees: for it to be transparent to consumers doesn't that mean the console needs to learn to process the new CR?

is the console team lined up to participate in this change?  or if not what does the transition look like? (will we have both the built-in mechanism and this new one, at least temporarily during the transition period?)

enhancements/oc/cli-manager.md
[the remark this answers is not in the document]
bparees: @soltysh it seems like this thread was never really resolved satisfactorily (or even unsatisfactorily, i.e. saying "yes the ux is degraded but it unification of cli mechanisms is worth it and is the direction we're going")

is there an alternative here?  or are the concerns raised on this thread being rejected as insufficient to change implementation directions?

enhancements/oc/cli-manager.md
```diff
@@ -0,0 +1,334 @@
...
+`krew` is an upstream project to distribute CLI tools (plugins) to Kubernetes users today.
+It works by reading a Git repository of files describing the plugins,
+and providing download links to them for various different OS and architecture combinations.
+Since those download links in the default index are Internet-facing, a Git and file server would need to be
+setup by customers to create their own custom index for use in disconnected environments.
+
+In order to avoid creating a new protocol and tool for this functionality, this proposal aims to leverage the `krew` project and a new custom index feature provided by `krew` in the form of a new Controller. The index will be managed by cluster-admins using
+[Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) (or CRs). An image registry will be used to host images that contain the binaries.
+
+By leveraging an image registry, which is an existing dependency for OpenShift, and allowing the index to be managed
```
bparees, line 50: you say this as if it means the feature is dependent on the OCP internal registry, but i see nothing in the design that requires any images(even for our own plugins) be hosted in the internal registry (i'd expect to get our plugin images from registry.redhat.io, not via/through the internal registry).

in fact i think it is best that this feature *not* take a dependency on the internal registry, as we may want to make that component more optional in the future.

enhancements/oc/cli-manager.md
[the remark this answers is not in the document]
bparees: looks like this was more or less answered in the "alternatives" section below:  we'll move forward w/ this approach and users can use `alias` if they want an easy direct invocation

enhancements/oc/cli-manager.md
bparees: @soltysh i don't really see how this was addressed in the design.....   are you going to use a sidecar approach to allow k8s to handle the image pulling for you and avoid the problem, or is there another plan?

any sort of details about how the images will be pulled/extracted seems like a good thing to have in the design details.

enhancements/oc/cli-manager.md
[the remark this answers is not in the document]
soltysh: @deejross make sure to add followups mentioning that we will want to address the problem of all CLIs as well. This initial proposal is only meant for plugins.

enhancements/oc/cli-manager.md
soltysh: @deejross ^

enhancements/oc/cli-manager.md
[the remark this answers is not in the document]
deejross: Done

enhancements/oc/cli-manager.md
deejross: I have added these points to the Design Details section, as they are important.

enhancements/oc/cli-manager.md
deejross: Due to the nature of krew, downloads are archived into `.tar.gz` files. While it might be possible for the console to implement deflate functionality, I had planned on providing a single-file download option in the REST API specifically for use by the console to simplify integrating with it.

enhancements/oc/cli-manager.md
deejross: If the controller is optional, then the console wouldn't be able to rely on it being present, so the existing mechanism would need to remain.

enhancements/oc/cli-manager.md
deejross: This is mostly saying that you need an image registry to run OpenShift in disconnected environments already, so you won't need to stand up additional infrastructure (unless you want to) to support this functionality.

bparees: got it.

enhancements/oc/cli-manager.md
bparees: > If the controller is optional, then the console wouldn't be able to rely on it being present, so the existing mechanism would need to remain.

no, if it's optional the console would have to be able to handle it not being present and tell the admin "hey if you want cli download functionality, please install/enable this optional feature/capability"

bparees:
@deejross @soltysh can we get a bit of a roadmap here, as there seem to be a few phases of implementation.

e.g.:

phase 1:  add new CRs+component that serves plugin binaries, existing cli download server continues running, serves RH native binaries.  Console is unaware of the new CR/plugin server

phase 2: console is aware of plugin server, serves both the native binaries and the plugin binaries

phase 3: cli manager takes over serving native binaries(as plugins?), existing cli download server goes away

phase 4: cli manager becomes an optional olm component that is installed by default on day 0

that list may not be accurate, please write your own, but i'm interested in understanding mostly:

1) when the console is going to become aware of this thing and use it (basically ensuring the console team is on board w/ this plan/design)
2) when the existing cli server will go away, since part of why we are saying it's ok to add this new component to the core is because we're getting rid of an existing one

afflom:
I'm a bit late to the party here, but can we consider NOT using OCI images for this? I agree that OCI formatted objects are the best way to mirror content, but this can be accomplished without the overhead of handling OCI images. I suggest storing arbitrary content in its original format within an OCI registry as we are seeing with Helm (https://helm.sh/docs/topics/registries/).

deejross:
@bparees I've added a roadmap section to the proposal. Any other concerns or anything I missed?

bparees:
> Controller should be optional, but enabled by default as it will be replacing the existing mechanism which was intented to be temporary 

maybe that was intended to address optionality?  but it stops short of addressing what the console would have to do if this thing is optional (being optional means the CRDs also don't get registered)

again i think making it optional can be a future goal (especially since it needs coordination w/ the console), i just want to see that it's in your plans

deejross:
@bparees fair enough, I have added that to the Followup section.

bparees:
thanks, lgtm.  not sure if i'm the only person you're still waiting on sign-off from

soltysh (approved):
/lgtm
/hold cancel

jupierce:
After reading the proposal and the comments, production distribution is still unclear to me. I see things described as "bring your own" and also references to default/payload based plugins. _If_ any plugins are supposed to be part of the release payload and the release payload only contains images for a single architecture, how would we have multi-arch support as implied by the CR? What is the nature of: 
```
Image string `json:"image"`
```
Is it an imagestreamtag? A pullspec with a floating tag? A pullspec with a sha256 digest?

- If imagestreamtag, it cannot presently support multiarch.
- If pullspec with floating tag, what is the registry, what is the tag, how is is updated in the CR at build time, how is it QE'd before the artifact carrying image is GA'd / released to a known registry?
- If pullspec with sha256, what is the registry, how is it updated in the CR at build time, how is it QE'd before the artifact carrying image is GA'd / released to a known registry?

Optional operators have solved these problems. If all RH plugins are to be delivered via OLM, then I have fewer concerns.

PR #925, opened 2021-10-07
[OCPNODE-716] Kubelet<->Controller Manager communication Profiles

harche:

### DIFF
--- a/enhancements/node-latency-profile/node-latency-profiles.md
+++ b/enhancements/node-latency-profile/node-latency-profiles.md
@@ -1,0 +1,354 @@
+---
+title: worker-latency-profiles
+authors:
+  - "@harche"
+reviewers:
+  - "@rphillips"
+  - "@sttts"
+  - "@soltysh"
+approvers:
+  - "@rphillips"
+  - "@sttts"
+  - "@soltysh"
+creation-date: 2021-09-28
+last-updated: 2021-10-19
+status: implementable
+see-also:
+  - https://github.com/kubernetes-sigs/kubespray/blob/master/docs/kubernetes-reliability.md
+  - https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#controllermanagerconfig
+replaces:
+superseded-by:
+---
+
+# WorkerLatencyProfile
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+
+## Summary
+
+To make sure Openshift cluster keeps running optimally under various circumstances where network latency, API Server latency, etcd latency, control plane load latency, etc. may not always be at their best (e.g. Edge cases), we can tweak the frequency of the status updates done by the `Kubelet` and the corresponding reaction times of the `Controller Manager` and `Kube API Server`.
+
+## Motivation
+
+In order to adjust to different latencies the cluster may experience, we need to adjust how frequently the `Kubelet` updates it's status and how do components such as `Controller Manager` and `Kube API Server`react to those status updates.
+
+In it's more rudimentary form, one can simply edit the relavent arguments to the `Kubelet`, `Controller Manager` and `Kube API Server`to achieve the desired results.
+Upstream `Kubespray` project has excellent [documentation](https://github.com/kubernetes-sigs/kubespray/blob/master/docs/kubernetes-reliability.md#recommendations-for-different-cases) on relavent arguments to each of those components in different scenarios.
+
+But the main motivation of this enhancement is to allow this updates to these critical components in a more controlled manner. Instead of letting the users directly modify those arguments, we want to make sure it happens systematically and methodically. This will eliminate any room for manual errors and would not lead to an unscheduled downtime.
+
+### Goals
+
+* Enable Openshift cluster to fine tune the reliability in various latency scenarios using simple `WorkerLatencyProfile`.
+
+### Non-Goals
+
+* Modify the existing `Kubelet`, `Controller Manager` or `Kube API Server` code in any way.
+* Allow any kind of different latency scenarios between masters
+
+### User Stories
+
+* User wants to fine tune the cluster reliability for their node latency scenario while making sure they are protected from setting parameters that could potentially break the cluster.
+
+## Proposal
+
+* A new Custom Resource `WorkerLatencyProfile` should be created.
+* This CR should be monitored by [MCO](https://github.com/openshift/machine-config-operator), [KCMO](https://github.com/openshift/cluster-kube-controller-manager-operator) and [CKAO](https://github.com/openshift/cluster-kube-apiserver-operator).
+* `KubeletConfigController` of MCO will modify Kubelet specific arguments
+* KCMO will modify `Controller Manager` specific arguments using `ConfigObserver`
+* CKAO will modify `Kube API Server` specific arguments using `ConfigObserver`
+
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+* Succeffully update the relavent arguments of the `Kubelet`, `Controller Manager` and `Kube API Server`
+* End user documentation
+
+#### Tech Preview -> GA
+* More testing (upgrade, downgrade, scale)
+* Optinally make it available during installation
+
+#### Removing a deprecated feature
+
+N/A
+
+## Design Details
+
+### API Extensions
+
+We propose a new following [config v1 API](https://github.com/openshift/api/tree/master/config/v1) which will be consumed by the `WorkerLatencyProfile`,
+
+
+```go
+type WorkerLatencyProfile struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// spec holds user settable values for configuration
+	// +kubebuilder:validation:Required
+	// +optional
+	Spec WorkerLatencyProfileSpec `json:"spec"`
+	// status holds observed values from the cluster. They may not be overridden.
+	// +optional
+	Status WorkerLatencyProfileStatus `json:"status"`
+}
+```
+
+Please refer to the section [WorkerLatencyProfile Scenarios](#workerLatencyProfile-scenarios) for more details on different scenarios.
+```go
+type WorkerLatencyProfileType string
+
+const (
+    // Medium Update and Average Reaction
+    MUAR WorkerLatencyProfileType = "MediumUpdateAndAverageReaction"
+
+    // Low Update and Slow Reaction
+    LUSR WorkerLatencyProfileType = "LowUpdateAndSlowReaction"
+
+    // Default values of relavent Kubelet and Controller Manager
+    Default WorkerLatencyProfileType = "Default"
+)
+```
+
+```go
+type WorkerLatencyProfileSpec struct {
+  // ProfileType determins the type of WorkerLatencyProfile to set on the cluster.
+  // Only follow values "FastUpdateAndFastReaction", "MediumUpdateAndAverageReaction"
+  // "LowUpdateAndSlowReaction" or "Default" are allowed. If the user does not explicitly
+  // set the value, "Default" will be automatically assigned.
+  ProfileType WorkerLatencyProfileType `json:"WorkerLatencyProfileType"`
+}
+```
+```go
+// WorkerLatencyProfileStatus provides information about the status of the
+// WorkerLatencyProfile rollout
+type WorkerLatencyProfileStatus struct {
+  // conditions describes the state of the WorkerLatencyProfile and related components
+  // (Kubelet or Controller Manager or Kube API Server)
+  // +patchMergeKey=type
+  // +patchStrategy=merge
+  // +optional
+  Conditions []NodeLatencyStatusCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+  // relatedObjects is a list of objects that are "interesting" or related to this WorkerLatencyProfile. e.g. KubeletConfig object used for updating Kubelet arguments
+  // +optional
+  RelatedObjects []ObjectReference `json:"relatedObjects,omitempty"`
+}
+```
+```go
+// NodeLatencyStatusConditionType is an aspect of WorkerLatencyProfile state.
+type NodeLatencyStatusConditionType string
+
+const (
+  // Progressing indicates that the updates to component (Kubelet or Controller
+  // Manager or Kube API Server) is actively rolling out, propagating changes to the
+  // respective arguments.
+  WorkerLatencyProfileProgressing NodeLatencyStatusConditionType = "Progressing"
+
+  // Complete indicates whether the component (Kubelet or Controller Manager or Kube API Server)
+  // is successfully updated the respective arguments.
+  WorkerLatencyProfileComplete NodeLatencyStatusConditionType = "Complete"
+
+  // Degraded indicates that the component (Kubelet or Controller Manager or Kube API Server)
+  // does not reach the state 'Complete' over a period of time
+  // resulting in either a lower quality or absence of service.
+  // If the component enters in this state, "Default" WorkerLatencyProfileType
+  // rollout will be initiated to restore the respective default arguments of all
+  // components.
+  WorkerLatencyProfileDegraded NodeLatencyStatusConditionType = "Degraded"
+)
+```
+```go
+
+type ConditionStatus string
+
+// These are valid condition statuses. "ConditionTrue" means a resource is in the condition.
+// "ConditionFalse" means a resource is not in the condition. "ConditionUnknown" means kubernetes
+// can't decide if a resource is in the condition or not.
+const (
+	ConditionTrue    ConditionStatus = "True"
+	ConditionFalse   ConditionStatus = "False"
+	ConditionUnknown ConditionStatus = "Unknown"
+)
+```
+```go
+
+type NodeLatencyStatusCondition struct {
+	// type specifies the aspect reported by this condition.
+	// +kubebuilder:validation:Required
+	// +required
+	Type NodeLatencyStatusConditionType `json:"type"`
+
+	// status of the condition, one of True, False, Unknown.
+	// +kubebuilder:validation:Required
+	// +required
+	Status ConditionStatus `json:"status"`
+
+	// lastTransitionTime is the time of the last update to the current status property.
+	// +kubebuilder:validation:Required
+	// +required
+	LastTransitionTime metav1.Time `json:"lastTransitionTime"`
+
+	// reason is the CamelCase reason for the condition's current status.
+	// +optional
+	Reason string `json:"reason,omitempty"`
+
+	// message provides additional information about the current condition.
+	// This is only to be consumed by humans.  It may contain Line Feed
+	// characters (U+000A), which should be rendered as new lines.
+	// +optional
+	Message string `json:"message,omitempty"`
+}
+```
+```go
+// ObjectReference contains enough information to let you inspect or modify the referred object.
+type ObjectReference struct {
+	// group of the referent.
+	// +kubebuilder:validation:Required
+	// +required
+	Group string `json:"group"`
+	// resource of the referent.
+	// +kubebuilder:validation:Required
+	// +required
+	Resource string `json:"resource"`
+	// namespace of the referent.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+	// name of the referent.
+	// +kubebuilder:validation:Required
+	// +required
+	Name string `json:"name"`
+}
+
+```
+#### WorkerLatencyProfile Custom Resource
+
+
+```yaml
+kind: WorkerLatencyProfile
+metadata:
+  name: medium-update-and-average-reaction
+spec:
+  profileType: "MediumUpdateAndAverageReaction" # Possible values, "MediumUpdateAndAverageReaction", "LowUpdateAndSlowReaction" or "Default"
+```
+
+### Operational Aspects of API Extensions
+
+Openshift always assumes low latency with 100% reliable network connections amongst master nodes. However, this may not be always true for the worker nodes (e.g Edge cases). In this section we will discuss how to keep the cluster healthy when the worker nodes have relatively high latency with not so reliable network connections to reach the master nodes.
+
+When the worker nodes have slightly higher than usual latency or/and relatively worse network connection, we can tweak the frequency at which the `Kubelet` updates the status and the corresponding reaction time of the `Controller Manager` to avoid considering the node unhealthy/unreachable.
+We will also need to tweak `--default-not-ready-toleration-seconds` and `--default-unreachable-toleration-seconds` of the `Kube API Server` to make sure pod eviction happens in timely manner.
+
+#### Default Update And Default Reaction
+Openshift ships with following default configuration which works quite well in most cases.
+
+By default `Kubelet` updates it's status every 10 seconds (`--node-status-update-frequency`), while `Controller Manager` checks the statuses of `Kubelet` every 5 seconds (`--node-monitor-period`).
+Before considering the `Kubelet` unhealthy `Controller Manager` will wait for 40 seconds (`--node-monitor-grace-period`) to hear from the `Kubelet`. Once the `Kubelet` is considered unhealthy the node is given `node.kubernetes.io/not-ready` or `node.kubernetes.io/unreachable` taints.
+Pods with `NoExecute` taint will get executed as per [tolerationSeconds](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions), but pods without such taint will get evicted in 300 seconds (`--default-not-ready-toleration-seconds` and `--default-unreachable-toleration-seconds` settings of the `Kube API Server`)
+
+#### Medium Update And Average Reaction
+
+While the default configuration works well in most cases, sometimes the worker nodes may find themselves in a network with slightly higher than usual latency and/or degraded reliability.
+
+In this scenario, we will reduce the frequency of `Kubelet` updates to every 20 seconds (`--node-status-update-frequency`) and change the node monitor grade period of the `Controller Manager` to 2 minutes (`--node-monitor-grace-period`). `--default-not-ready-toleration-seconds` and `--default-unreachable-toleration-seconds` of the Kube API Server are set to 60 seconds.
+
+`Controller Manager` will wait for 2 minutes to consider unhealthy status for the node and since `--default-not-ready-toleration-seconds` and `--default-unreachable-toleration-seconds` of the Kube API Server are set to 60 seconds the total time will be 3 minutes before eviction process starts.
+
+#### Low Update and Slow reaction
+
+Worker nodes may find themselves in a network with extremely high latency and/or bad reliability.
+In this scenario, we will reduce the frequency of `Kubelet` updates further to every 1 minute (`--node-status-update-frequency`) and change the node monitor grade period of the `Controller Manager` to every 5 minutes (`--node-monitor-grace-period`). `--default-not-ready-toleration-seconds` and `--default-unreachable-toleration-seconds` of the Kube API Server are set to 60 seconds.
+
+`Controller Manager` will wait for 5 minutes to consider unhealthy status for the node and since `--default-not-ready-toleration-seconds` and `--default-unreachable-toleration-seconds` of the Kube API Server are set to 60 seconds the total time will be 6 minutes before eviction process starts.
+
+
+#### Coordination between the Kubelet, Controller Manager and Kube API Server
+
+Since the updates to the `Kubelet`, `Controller Manager` and `Kube API Server` arguments cannot be simultaneous, we will have to take 3-phase approach.
+
+MCO and KCMO will have necessary observers to consume `WorkerLatencyProfile`.
+
+1. **Kubelet** -
+MCO will initialize updates to the Kubelet as a first step by submitting a `KubeletConfig` with relavent value of `nodeStatusUpdateFrequency` which will be processed by existing `KubeletConfigController` within MCO.
+MCO will observe the [status](https://github.com/openshift/machine-config-operator/blob/85f9f9583451cb1e4c25893598eb5d086fd7e4ba/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L349) of this `KubeletConfig` and fill out `NodeLatencyStatusCondition` appropriately.
+
+2. **Controller Manager** -
+Once the Kubelet arguments are successfully updated, KCMO will start rolling out relavent updates to the arguments `--node-monitor-period` and/or `--node-monitor-grace-period` of the Controller Manager.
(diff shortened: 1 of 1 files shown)


enhancements/reliability/kubernetes-reliability.md
```diff
@@ -0,0 +1,274 @@
...
+
+* Enable Openshift cluster to fine tune the reliability in various latency scenarios using simple Kubelet<->Controller communication profiles.
+
+### Non-Goals
+
+* Modify the existing `Kubelet`, `Controller Manager` or `Kube API Server` code in any way.
+
+### User Stories
+
+* User wants to fine tune the cluster reliability for their scenario.
```
sttts, line 51: * User wants to be protected from setting parameters that break the cluster?

enhancements/reliability/kubernetes-reliability.md
```diff
@@ -0,0 +1,274 @@
...
+```
+
+Please refer to the `Kubespray` [documentation](https://github.com/kubernetes-sigs/kubespray/blob/master/docs/kubernetes-reliability.md#recommendations-for-different-cases) for more details on different scenarios.
+```go
+type ProfileType int
+
+const (
+    // Fast Update and Fast Reaction.
+    // See - https://github.com/kubernetes-sigs/kubespray/blob/af04906b518efe8d6dcfc9e2ff57b17b49c20295/docs/kubernetes-reliability.md#fast-update-and-fast-reaction
+    FUFR ProfileType = iota //
```
sttts, line 105: these must be strings

harche: Thanks, updated.

enhancements/reliability/kubernetes-reliability.md
```diff
@@ -0,0 +1,274 @@
...
+    // Low Update and Slow Reaction
+    // See - https://github.com/kubernetes-sigs/kubespray/blob/af04906b518efe8d6dcfc9e2ff57b17b49c20295/docs/kubernetes-reliability.md#low-update-and-slow-reaction
+    LUSR
+)
+```
+
+```go
+type ReliabilityProfileSpec struct {
+    // ReliabilityProfileType determins the type of ReliabilityProfile to set on the cluster.
+	ReliabilityProfileType ProfileType `json:"profileType"`
```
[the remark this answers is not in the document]
sttts: why do we need these cryptic abbrevations? Some more characters are free of charge.

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,276 @@
...
+    // See - https://github.com/kubernetes-sigs/kubespray/blob/af04906b518efe8d6dcfc9e2ff57b17b49c20295/docs/kubernetes-reliability.md#low-update-and-slow-reaction
+    LUSR ProfileType = "LUSR"
+)
+```
+
+```go
+type NodeLatencyProfileSpec struct {
+  // NodeLatencyProfileType determins the type of NodeLatencyProfile to set on the cluster.
+  // Only follow values "FUFR", "MUAR" or "LUSR" are allowed.
+	NodeLatencyProfileType ProfileType `json:"profileType"`
```
sttts, line 121: set a default

enhancements/reliability/kubernetes-reliability.md
```diff
@@ -0,0 +1,274 @@
...
+4. Even though KCMO has updated relavent controller manager arguments, it will keep monitoring  `ReliabilityProfile.Status.KubeAPIServerUpdateStatus.State`. In case it gets set to `FailedState` by CKAO, indicating that CKAO wasn't able to update relavent Kube API Server arguments successfully, it will initiate the rollback. After successful rollback, it will mark `ReliabilityProfile.Status.ControllerManagerUpdateStatus.State` to `RollBackState` so that MCO can initiate the rollback of the `KubeletConfig`
+
+#### CKAO
+
+1. The relavent observer in CKAO watching `ReliabilityProfile` wait until `ReliabilityProfile.Status.ControllerManagerUpdateStatus.State` to get set to `CompleteState`. This will ensure that the updates to Controller Manager and Kube API Server are not carried out simultaneously.
+2. This CKAO observer will read `ReliabilityProfile.Spec.ProfileType` and depending upon the `ProfileType` it will create a Config Observer with appropriate value for the arguments `--default-not-ready-toleration-seconds` and `--default-unreachable-toleration-seconds` of the Kube API Server.
```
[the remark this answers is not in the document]
harche: Thanks, updated.

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,276 @@
...
+const (
+    NotStartedState CurrentState = "NotStarted"
+    InProgressState CurrentState = "InProgress"
+    CompleteState   CurrentState = "Complete"
+    FailedState     CurrentState = "Failed"
+    RollBackState   CurrentState = "RollBacked"
+)
+
+type ProfileUpdateStatus struct {
+    State   CurrentState `json:"currentState"`
```
sttts, line 156: this looks like an anti-pattern in our APIs. Use conditions.

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,276 @@
...
+    Message string       `json:"message,omitempty"`
+}
+
+type KubeletArgsUpdateStatus struct {
+    // KubeletConfigName is the name of the KubeletConfig used for updating the kubelet arguments
+    KubeletConfigName string `json:"KubeletConfigName"`
```
sttts, line 162: what is this and why does it matter?

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,276 @@
...
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+
+## Summary
+
+To make sure Openshift cluster keeps running optimally under various circumstances where network latency, API Server latency, etcd latency, control plane load latency, etc. may not always be at their best (e.g. Edge cases), we can tweak the frequency of the status updates done by the `Kubelet` and the corresponding reaction times of the `Controller Manager` as well as `Kube API Server`.
```
sttts, line 31: My assumption was that this enhancement is just about worker nodes. Now I read about API and etcd latency. This puts this enhancement on a completely different level. The question becomes: do we want to support use-cases where etcd connectivity is slow? I am not convinced. This is a business and support question, not only technical.

cc @mfojtik

mfojtik: can we get a clarification on what is meant with "etcd latency" and "API server latency" and "control plane latency" ? the summary sounds a little vague and generic and open a pandora box to allow unbound customization of control plane performance flags :-) 

enhancement down does not mention etcd at all and the kubespray does not mention API server flags.

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,276 @@
...
+
+## Motivation
+
+In order to adjust to different latencies the cluster may experience, we need to adjust how frequently the `Kubelet` updates it's status and how do components such as `Controller Manager` and `Kube API Server` react to those status updates.
```
mfojtik, line 35: how does KAS react to node updates? (other than maybe updating the host network endpoints?)

enhancements/node-latency-profile/node-latency-profiles.md
harche: References to etcd and other latencies are borrowed from kubespray [documentation](https://github.com/kubernetes-sigs/kubespray/blob/master/docs/kubernetes-reliability.md#overview) on this subject.

enhancements/node-latency-profile/node-latency-profiles.md
harche: Please refer to the section `Coordination between MCO, KCMO, CKAO` which talks in detail about flags and coordination between these components.

enhancements/node-latency-profile/node-latency-profiles.md
harche: It has been documented in section `Coordination between MCO, KCMO, CKAO` subsection `MCO`. In short, MCO needs to know which `KubeletConfig` to track (out of many pre-existing) used for updating Kubelet flags relevant to the given `NodeLatencyProfile`

enhancements/node-latency-profile/node-latency-profiles.md
harche: Ok, let me look in into it. Thanks.

enhancements/node-latency-profile/node-latency-profiles.md
harche: Considering how much of a broad impact these settings have, I am not in favour of setting any defaults. Let user explicitly tell us which `NodeLatencyProfile` they want to set. I don't think it's safe for us to assume any default here because we could just end up breaking their cluster. 

The default value to me only makes sense when there could be a reasonable value of a parameter that could be assumed in absence of user provided one. But in this case, user `must` tell us which latency they will like the cluster to operate in. Assuming anything incorrect, like via default, could even bring down the cluster. 

IMO, user should explicity specific either `FastUpdateAndFastReaction` or `MediumUpdateAndAverageReaction` or `LowUpdateAndSlowReaction`. If nothing is provided, we just error out without taking any action.

harche: Although, we could introduce another profile type which has current default values for args of Kubelet, CM and KAS. Something like `DefaultUpdateAndDefaultReaction`. @sttts does that make sense?

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,276 @@
...
+```
+
+Please refer to the `Kubespray` [documentation](https://github.com/kubernetes-sigs/kubespray/blob/master/docs/kubernetes-reliability.md#recommendations-for-different-cases) for more details on different scenarios.
+```go
+type ProfileType int
+
+const (
+    // Fast Update and Fast Reaction.
+    // See - https://github.com/kubernetes-sigs/kubespray/blob/af04906b518efe8d6dcfc9e2ff57b17b49c20295/docs/kubernetes-reliability.md#fast-update-and-fast-reaction
+    FUFR ProfileType = "FUFR" //
```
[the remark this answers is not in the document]
harche: Updated.

enhancements/node-latency-profile/node-latency-profiles.md
mfojtik: how can we protect from following scenario:

1. customer install cluster with 3 workers nodes (default latency profile)
2. customer switch the node latency profile to LUSR (low update/slow reaction)
3. customer add 10000 worker nodes (edge?)
4. customer run for some time
5. customer switch the latency profile to "fast" for some reason
6. customer bricked their cluster, API server is under DDoS, and nodes are set as not ready

enhancements/node-latency-profile/node-latency-profiles.md
sttts: just call it `Default`

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,280 @@
...
+5. Even though `KubeletConfig` is successfully reconciled, MCO observer will keep monitoring `NodeLatencyProfile.Status.ControllerManagerUpdateStatus.State`. If it gets set to `FailedState` or `RollBackState` which indicates that controller manager arguments update didn't go as planned and MCO observer will rollback Kubelet arguments by deleting the `KubeletConfig`.
+
+#### KCMO
+
+1. The relavent observer in KMCO watching `NodeLatencyProfile` wait until `NodeLatencyProfile.Status.KubeletUpdateStatus.State` to get set to `CompleteState`. This will ensure that the updates to Kubelet and Controller Manager are not carried out simultaneously.
+2. This KMCO observer will read `NodeLatencyProfile.Spec.ProfileType` and depending upon the `ProfileType` it will output appropriate value for the extended arguments `--node-monitor-period` and/or `--node-monitor-grace-period` of the controller manager.
+3. If KMCO is successful in updating the given arguments of the controller manager, it will update  `NodeLatencyProfile.Status.ControllerManagerUpdateStatus.State` to `CompleteState`.
```
sttts, line 220: this is misplaced. kcm does not push back status, but only observes. If you want status about kcm/kube-apiserver, you have to do that from another place, e.g. MCO.

enhancements/node-latency-profile/node-latency-profiles.md
soltysh: The initial idea as described in https://issues.redhat.com/browse/WRKLDS-351 and https://issues.redhat.com/browse/OCPNODE-638 was only touching kube-controller-manager and kubelet. If you're planning to expand this to all control-plane components then this requires both @sttts and @hexfusion approval since they own api and etcd, respectively. Most importantly, though, it wasn't never announced to them that this will affect them so they couldn't account for that work in 4.10 nor in future versions.

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Summary
+
+To make sure Openshift cluster keeps running optimally under various circumstances where network latency, API Server latency, etcd latency, control plane load latency, etc. may not always be at their best (e.g. Edge cases), we can tweak the frequency of the status updates done by the `Kubelet` and the corresponding reaction times of the `Controller Manager` as well as `Kube API Server`.
+
+## Motivation
+
+In order to adjust to different latencies the cluster may experience, we need to adjust how frequently the `Kubelet` updates it's status and how do components such as `Controller Manager` and `Kube API Server` react to those status updates.
+
+In it's more rudimentary form, one can simply edit the relavent arguments to the `Kubelet`, `Controller Manager` and `Kube API Server` to achieve the desired results. Upstream `Kubespray` project has excellent [documentation](https://github.com/kubernetes-sigs/kubespray/blob/master/docs/kubernetes-reliability.md#recommendations-for-different-cases) on relavent arguments to each of those components in different scenarios.
```
soltysh, line 41: Not to mention the fact that touching these values today makes your cluster immediately unsupported, since none of these flags is exposed and you need to touch `unsupportedConfigOverrides`

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,311 @@
...
+
+But the main motivation of this enhancement is to allow this updates to these critical components in a more controlled manner. Instead of letting the users directly modify those arguments, we want to make sure it happens systematically and methodically. This will eliminate any room for manual errors and would not lead to an unscheduled downtime.
+
+### Goals
+
+* Enable Openshift cluster to fine tune the reliability in various latency scenarios using simple `NodeLatencyProfile`.
```
soltysh, line 47: I'd imagine that one of the goals is to explicitly call out the profiles and values that will affect kcm and kubelet, with the impact it will have on the entire cluster, before we will go ahead and start implementing this.

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,276 @@
...
+
+* User wants to fine tune the cluster reliability for their scenario while making sure they are protected from setting parameters that could potentially break the cluster.
+
+## Proposal
+
+* A new Custom Resource `NodeLatencyProfile` should be created.
+* This CR should be monitored by [MCO](https://github.com/openshift/machine-config-operator), [KCMO](https://github.com/openshift/cluster-kube-controller-manager-operator) and [CKAO](https://github.com/openshift/cluster-kube-apiserver-operator).
+* `KubeletConfigController` of MCO will modify Kubelet specific arguments
+* KCMO will modify `Controller Manager` specific arguments using `ConfigObserver`
+* CKAO will modify `Kube API Server` specific arguments using `ConfigObserver`
```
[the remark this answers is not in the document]
soltysh: You're writing this proposal, so I'd prefer the actual values being called out in this proposal, rather than constantly referring to kubespray docs.

enhancements/node-latency-profile/node-latency-profiles.md
harche: @soltysh It's correct that the initial idea was only touching kube-controller-manager and kubelet. However, as soon as we started writing this enhancement we realized that if you don't modify the relevant arguments of kube API server, it may have undesirable consequence on the pods that do no explicitly set taint based eviction timers. More on that in detail is described [here](https://github.com/kubernetes-sigs/kubespray/blob/master/docs/kubernetes-reliability.md#failure).

So in order to set the taint based eviction timers globally (and not limited to just the pods that explicitly set them), it's suggested that we can modify `--default-not-ready-toleration-seconds` & `--default-unreachable-toleration-seconds` arguments of the Kube API Server. This way we get a double protection by making sure the toleration values are set globally and since we are setting them, we can make sure they are set to their correct values. This could avoid unplanned cluster downtime for customers who might accidently (and manually) put wrong values (per pod). 

But I totally agree with your point that, since it was not announced to the API server team about this (in our defence, the moment we started looking into it we put those requirements as a part of this enhancement right away) so they may not able to account for this work in 4.10. But we can go ahead with just the changes in controller-manager and kubelet in 4.10 and cover in the documentation that customers must set taint based eviction timers manually (at least in 4.10). How does that sound @soltysh ?

harche: Although, since @sttts is suggesting we restrict the enhancement strictly to worker nodes, I think we might as well just drop the changes Kube API Server arguments which affect global toleration values, because I fear they will also affect pods running on the master nodes. 

In this case, we just have to restrict ourselves with the changes to controller manager and kubelet only but, make sure we document that user will have to set pod taint based eviction timers manually.

sttts: The changes to kube-apiserver flags are fine. That's not what I meant. In other words, in my view there are two topics:

1. configure kube-apiserver (and kcm and kubelet) to cope with slowly and unreliably connected worker nodes
2. to cope with slow and unreliably connected master nodes.

I was arguing against (2). It must be clear in the enhancements that master nodes are connected to eachother as in any other non-edge scenario, with low latency and nearly 100% reliability. Only the workers are distant and hence we need (1).

harche: If there is any chance changing `--default-not-ready-toleration-seconds` & `--default-unreachable-toleration-seconds` args of the kube API server will affect any control plane pods then it's a strict no go. The more I think about it, more I am getting inclined to drop it. 

I had added the changes to Kube API server to minimize the error arising from user manually setting eviction taints for every pod, but that should not happen at the cost of destabilizing existing control plane pods. Let the user set those taints manually per pod, or as I have mentioned in some other comment we could explore adding an admission controller to inject those taints for all the pods heading for worker nodes.

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,311 @@
...
+
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+* Succeffully update the relavent arguments of the `Kubelet`, `Controller Manager`, `Kube API Server`
```
staebler, line 69:
```suggestion
* Successfully update the relevant arguments of the `Kubelet`, `Controller Manager`, `Kube API Server`
```

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,330 @@
...
+
+#### Phase 2 - Controller Manager
+Once the Kubelet arguments are successfully updated, KCMO will start rolling out relavent updates to the arguments `--node-monitor-period` and/or `--node-monitor-grace-period` of the Controller Manager. MCO will observe the Controller Manager to come back up successfully with updated arguments and will update `NodeLatencyStatusCondition` appropriately to mark the completion of the update to the arguments of Controller Manager.
+
+#### Phase 3 - Kube API Server
+After the Controller Manager arguments have been updated, CKAO will start rolling out relevant updates to the arguments `--default-not-ready-toleration-seconds` and `--default-unreachable-toleration-seconds` of the Kube API Server. MCO will observe the Kube API Server to come back up successfully with updated arguments and will update `NodeLatencyStatusCondition` appropriately to mark the completion of the update to the arguments of Kube API Server.
```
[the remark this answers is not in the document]
harche: Just to try this out, I executed `oc describe co kube-apiserver` 

```
Spec:
Status:
  Conditions:
    Last Transition Time:  2021-11-09T09:11:41Z
    Message:               NodeControllerDegraded: All master nodes are ready
    Reason:                AsExpected
    Status:                False
    Type:                  Degraded
    Last Transition Time:  2021-11-09T09:28:47Z
    Message:               NodeInstallerProgressing: 3 nodes are at revision 6
    Reason:                AsExpected
    Status:                False
    Type:                  Progressing
    Last Transition Time:  2021-11-09T09:12:17Z
    Message:               StaticPodsAvailable: 3 nodes are active; 3 nodes are at revision 6
    Reason:                AsExpected
    Status:                True
    Type:                  Available
    Last Transition Time:  2021-11-09T09:06:54Z
    Message:               KubeletMinorVersionUpgradeable: Kubelet and API server minor versions are synced.
    Reason:                AsExpected
    Status:                True
    Type:                  Upgradeable
  Extension:               <nil>
```
I can see some information related to node status, but I wasn't able to find `status.nodeStatus` there.

@sttts am I looking at the right object?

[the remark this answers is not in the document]
sttts: I meant `kubectl get kubeapiservers -o yaml`, but it looks like we don't have `status.observedGeneration` today, and with all the different controllers operating on that resource I am not sure it is feasible to add it.

This leaves you with checking `spec.observedConfig` for your flags and then wait for the `Progressing` condition.

rphillips:
I reviewed from the Node Team.

/lgtm

I think this is ready for lgtm's from @soltysh and @sttts. Once this merges we, can plan Jira stores for 4.11 on the Node, API, and Workloads team.

enhancements/node-latency-profile/node-latency-profiles.md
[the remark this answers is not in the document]
harche: Updated. Thanks.

harche:
It looks like recently some new sections have been made mandatory, I realized it after rebasing to latest master and running `make lint` 
```
Checking enhancements/node-latency-profile/node-latency-profiles.md
enhancements/node-latency-profile/node-latency-profiles.md missing "### API Extensions"
enhancements/node-latency-profile/node-latency-profiles.md missing "### Operational Aspects of API Extensions"
enhancements/node-latency-profile/node-latency-profiles.md missing "#### Failure Modes"
enhancements/node-latency-profile/node-latency-profiles.md missing "#### Support Procedures"
```

I will try to understand the new [enhancement template](https://github.com/sttts/enhancements-1/blob/8d075202751690ff578bf1f6bf93b3469dc044fd/guidelines/enhancement_template.md) to see what those sections are for.

harche:
> I reviewed from the Node Team.
> 
> /lgtm
> 
> I think this is ready for lgtm's from @soltysh and @sttts. Once this merges we, can plan Jira stores for 4.11 on the Node, API, and Workloads team.

Thanks @rphillips for the review.

harche:
I checked if this feature has any impact on [Pod Disruption Budget (PDB)](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets). 

If the user rolls out any profile other than `Default`, the reaction time of the cluster to mark the node unavailable will increase to accommodate anticipated increase in the node latency. This point is already described in point 3 of the section `### Risks and Mitigations`. 

This feature doesn't directly interfere with PDBs. But having said that, any increase in reaction time of the cluster will mean there will be corresponding increase in the terminating and then recreating the replacement pods affected by the unavailability of the node.

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,354 @@
...
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+
+## Summary
+
+To make sure Openshift cluster keeps running optimally under various circumstances where network latency, API Server latency, etcd latency, control plane load latency, etc. may not always be at their best (e.g. Edge cases), we can tweak the frequency of the status updates done by the `Kubelet` and the corresponding reaction times of the `Controller Manager` and `Kube API Server`.
```
sttts, line 36: this summary does not reflect the content. The enhancement is **only** about kube-apiserver<->worker-kubelet latency, not about etcd latency or latency among control plane nodes.

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,354 @@
...
+In order to adjust to different latencies the cluster may experience, we need to adjust how frequently the `Kubelet` updates it's status and how do components such as `Controller Manager` and `Kube API Server`react to those status updates.
+
+In it's more rudimentary form, one can simply edit the relavent arguments to the `Kubelet`, `Controller Manager` and `Kube API Server`to achieve the desired results.
+Upstream `Kubespray` project has excellent [documentation](https://github.com/kubernetes-sigs/kubespray/blob/master/docs/kubernetes-reliability.md#recommendations-for-different-cases) on relavent arguments to each of those components in different scenarios.
+
+But the main motivation of this enhancement is to allow this updates to these critical components in a more controlled manner. Instead of letting the users directly modify those arguments, we want to make sure it happens systematically and methodically. This will eliminate any room for manual errors and would not lead to an unscheduled downtime.
+
+### Goals
+
+* Enable Openshift cluster to fine tune the reliability in various latency scenarios using simple `WorkerLatencyProfile`.
```
sttts, line 49: be very clear which scenarios and which config changes

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,354 @@
...
+
+### Non-Goals
+
+* Modify the existing `Kubelet`, `Controller Manager` or `Kube API Server` code in any way.
+* Allow any kind of different latency scenarios between masters
```
sttts, line 54: the summary says this is a goal

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,354 @@
...
+
+### User Stories
+
+* User wants to fine tune the cluster reliability for their node latency scenario while making sure they are protected from setting parameters that could potentially break the cluster.
+
+## Proposal
+
+* A new Custom Resource `WorkerLatencyProfile` should be created.
```
sttts, line 62: There is another, parallel enhancement that introduces a `Node` resource. These settings here look like a good fit to moved there too.

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,354 @@
...
+
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+* Succeffully update the relavent arguments of the `Kubelet`, `Controller Manager` and `Kube API Server`
+* End user documentation
+
+#### Tech Preview -> GA
+* More testing (upgrade, downgrade, scale)
```
sttts, line 76: which kind of testing in detail? Manual? Automatic?

Who does the scale tests? What might be an acceptence criteria?

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,354 @@
...
+
+#### Removing a deprecated feature
+
+N/A
+
+## Design Details
+
+### API Extensions
+
+We propose a new following [config v1 API](https://github.com/openshift/api/tree/master/config/v1) which will be consumed by the `WorkerLatencyProfile`,
```
sttts, line 87: consumed? What does that mean?

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,354 @@
...
+}
+
+```
+#### WorkerLatencyProfile Custom Resource
+
+
+```yaml
+kind: WorkerLatencyProfile
+metadata:
+  name: medium-update-and-average-reaction
```
sttts, line 238: I remember having commented here before: why this name? Isn't this a singleton object in the cluster like with every other config/v1 API ?

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,354 @@
...
+spec:
+  profileType: "MediumUpdateAndAverageReaction" # Possible values, "MediumUpdateAndAverageReaction", "LowUpdateAndSlowReaction" or "Default"
+```
+
+### Operational Aspects of API Extensions
+
+Openshift always assumes low latency with 100% reliable network connections amongst master nodes. However, this may not be always true for the worker nodes (e.g Edge cases). In this section we will discuss how to keep the cluster healthy when the worker nodes have relatively high latency with not so reliable network connections to reach the master nodes.
```
sttts, line 245: this is a good discussiion (please keep it), but it is not really what belongs into this section. The section is about what to do when a cluster with a custom profile is running into problems. How can one rule out that this feature is the cause of the problem? Can one go back to the normal profile and try without dataloss and without outage?

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,354 @@
...
+
+Since this feature is controlled using the `KubeletConfig` and `ConfigObserver`, upgrade/downgrade strategies applicable for the `KubeletConfig` and `ConfigObserver` are applicable here too.
+
+## Drawbacks
+
+This solution is fairly complex due to the synchronization it requires between the control plane and the kubelet. Unfortunately, there isn't any way around this.
+
+## Alternatives
+
+Since we don't want users to manually modify the arguments for the components involved, there isn't any viable alternative that could safely work.
```
sttts, line 350: I see alternatives for the API. Why is it not part of the to-be-introduced Node object?

enhancements/node-latency-profile/node-latency-profiles.md
```diff
@@ -0,0 +1,354 @@
...
+
+## Motivation
+
+In order to adjust to different latencies the cluster may experience, we need to adjust how frequently the `Kubelet` updates it's status and how do components such as `Controller Manager` and `Kube API Server`react to those status updates.
+
+In it's more rudimentary form, one can simply edit the relavent arguments to the `Kubelet`, `Controller Manager` and `Kube API Server`to achieve the desired results.
+Upstream `Kubespray` project has excellent [documentation](https://github.com/kubernetes-sigs/kubespray/blob/master/docs/kubernetes-reliability.md#recommendations-for-different-cases) on relavent arguments to each of those components in different scenarios.
+
+But the main motivation of this enhancement is to allow this updates to these critical components in a more controlled manner. Instead of letting the users directly modify those arguments, we want to make sure it happens systematically and methodically. This will eliminate any room for manual errors and would not lead to an unscheduled downtime.
```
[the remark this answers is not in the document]
soltysh: I thought about it more and before diving more into how to solve the coordination problem I'm missing the entire point why we need one in the first place? What kind of coordination is needed, does it have to be kubelet->kcm->kas or it can be kubelet->kcm+kas? What happens if there's no coordination? The answers to these questions will allow you to choose the most appropriate and effective solution to the problem. I'd like you to come to Tuesday's arch call with these questions answered and we can discuss how to best address them only then.

harche:
closing this in favour of https://github.com/openshift/enhancements/pull/977

/close

PR #928, opened 2021-10-08
tools: allow the upstream git remote to have any name

dhellmann:
Instead of requiring that the local clone 'origin' point to the
openshift/enhancements fork, allow any name. Look for the remote with
the right URL when it is needed. Avoid looking for the remote for
every PR by creating a new Summarizer type that holds the remote
information in a way that can be reused for most operations that need
it, other than the initial repository configuration.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,5 +1,12 @@
 # Enhancement Tools
 
+## Local Repository
+
+The tools in this directory are designed to run out of a local clone
+of the `openshift/enhancements` git repository. The repository can be
+cloned from any source, but one of the remotes must point to the
+`openshift/enhancements` repo on GithHub.
+
 ## Configuring
 
 You need to create a `~/.config/ocp-enhancements/config.yml`


tools/enhancements/summary.go
```diff
@@ -24,29 +24,54 @@ func UpdateGitRepo() error {
 	return nil
 }
 
-const desiredRefSetting = "+refs/pull/*/head:refs/remotes/origin/pr/*"
+// findOpenShiftRemote looks through the remotes for the current git
+// repository to find one with "openshift/enhancements" in the URL
```
aravindhp, line 28: I suggest adding a requirement to have an upstream remote in the README.

tools/report/report.go
```diff
@@ -148,9 +148,9 @@ func showOnePR(prd *stats.PullRequestDetails, withDescription bool, withLabels b
 
 // ShowPRs prints a section of the report by formatting the
 // PullRequestDetails as a list.
-func ShowPRs(name string, prds []*stats.PullRequestDetails, withDescription bool, withLabels bool) {
+func ShowPRs(summarizer *enhancements.Summarizer, name string, prds []*stats.PullRequestDetails, withDescription bool, withLabels bool) error {
```
aravindhp, line 151: Is there a reason to return `error`? I don't see an actual error being returned nor is it checked for in the call site.

aravindhp (commented):
Thanks for fixing this, @dhellmann 

Tested-by: Aravindh Puthiyaparambil <aravindh@redhat.com>

tools/enhancements/summary.go
dhellmann: Good idea. I added another commit to this PR to do that.

tools/report/report.go
dhellmann: Ah, one iteration had this function creating the summarizer and then I moved that and didn't clean up. Fixed.

PR #937, opened 2021-10-19
monitoring: support configuration of alerting notifications by application owners

simonpasquier:
This document describes a solution that allows OpenShift users to route alert
notifications without cluster admin intervention. It complements the existing
[user workload monitoring](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/user-workload-monitoring.md) stack, enabling a full self-service experience for
workload monitoring.


enhancements/monitoring/alert-routing-for-user-workload-monitoring.md
```diff
@@ -0,0 +1,417 @@
...
+As an OpenShift cluster admin, I want to exclude certain user namespaces from
+modifying the Plaform Alertmanager configuration so that I can recover in case
+of breakage or bad behavior.
+
+#### Story 7
+
+As an OpenShift cluster admin, I don't want to support AlertmanagerConfig
+resources for application owners so that the configuration of the Platform
+Alertmanager cluster is completely under my control.
+
```
PhilipGough, line 140: Does this equate to full opt-out of the feature at a cluster wide level?

enhancements/monitoring/alert-routing-for-user-workload-monitoring.md
```diff
@@ -0,0 +1,417 @@
...
+As a UWM admin, I don't want to send user alerts to the Platform Alertmanager
+cluster because these alerts are managed by an external system (off-cluster Alertmanager for
+instance).
+
+### Story 9
+
+As a UWM admin, I want to send user alerts to the Platform Alertmanager
+cluster because UWM alerts are managed by an external system (off-cluster Alertmanager for
+instance).
+
```
PhilipGough, line 152: If managed by an external system, why would we want to send to platform AM?

enhancements/monitoring/alert-routing-for-user-workload-monitoring.md
simonpasquier: yes

simonpasquier: The CRD will always exist but at least the Platform Alertmanager won't gather any user `AlertmanagerConfig`s.

enhancements/monitoring/alert-routing-for-user-workload-monitoring.md
simonpasquier: typo :)

enhancements/monitoring/alert-routing-for-user-workload-monitoring.md
```diff
@@ -0,0 +1,417 @@
...
+As an OpenShift cluster admin, I don't want AlertmanagerConfig resources
+defined by application owners to interfere with the routing of platform alerts.
+
+#### Story 6
+
+As an OpenShift cluster admin, I want to exclude certain user namespaces from
+modifying the Plaform Alertmanager configuration so that I can recover in case
+of breakage or bad behavior.
+
```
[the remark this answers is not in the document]
PhilipGough: Got it, that makes sense. So it is a reactive solution essentially. Drifting off slightly to implementation, would they have any means other than logging to be alerted on what namespace has broken the conf?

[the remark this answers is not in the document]
PhilipGough: I see this is covered later with the `AlertmanagerBadConfig `alert

enhancements/monitoring/alert-routing-for-user-workload-monitoring.md
```diff
@@ -0,0 +1,411 @@
...
+#### AlertmanagerConfig resources not being reconciled
+
+An `AlertmanagerConfig` resource might require credentials (such as API keys)
+which are referenced by secrets. If the Platform Prometheus operator doesn't
+have permissions to read the secret or if the reference is incorrect (wrong
+name or key), the operator doesn't reconcile the resource in the final
+Alertmanager configuration.
+
+Mitigation
+* The Prometheus operator should expose a validating admission webhook that should prevent invalid configurations.
```
PhilipGough, line 333: This would require that the webhook and the operator share the same RBAC to read cluster wide secrets, correct?

simonpasquier: the operator implements the webhook endpoint so it should have the proper RBAC but maybe it's better if the webhook doesn't try to resolve external references?
More a discussion for https://github.com/prometheus-operator/prometheus-operator/pull/4316 though ;)

bwplotka (approved):
Amazing! Some suggestions, otherwise looking good 👍🏽

enhancements/monitoring/alert-routing-for-user-workload-monitoring.md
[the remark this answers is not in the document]
PhilipGough: I think that is the goal from the `prometheus-operator` perspective yes. We will discard the invalid CR and it will be excluded from the merge.

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,494 @@
...
+routes. We plan to simulate environments with thousands of `AlertmanagerConfig`
+resources and measure the impact on notification delivery.
+
+### Risks and Mitigations
+
+#### Disruption of the platform Alertmanager
+
+Even though the Prometheus operator prevents it as much as it can, it may be
+possible for users to create an `AlertmanagerConfig` resource that triggers the
+Prometheus operator to generate an invalid Alertmanager configuration, leading
```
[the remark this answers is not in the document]
simonpasquier: The operator will do its best to reject invalid configuration but bugs happen and it may happen that the operator generates a configuration that couldn't be parsed by Alertmanager. In short, unlikely to happen but still possible.

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -226,22 +227,26 @@ metadata:
...
 * Processing of user alerts may slow down the alert notification pipeline.
 * Cluster admins don't want to deal with delivery errors for user notifications.
 
+At the same time, application owners want to configure their alert
+notifications without requesting external intervention.
+
 In this case, UWM admins have the possibility to deploy a dedicated
-Alertmanager. The configuration options will be equivalent to the options
-exposed for the Platform Alertmanager and exposed under the `alertmanager` key
+Alertmanager. The configuration options are to the options
```
PhilipGough, line 255: "are equivalent"?

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -517,6 +524,17 @@ of the monitoring components.
 
 N/A
 
+### Operational Aspects of API Extensions
+
+#### Failure Modes
+
+The validating webhook is configured with `failurePolicy: Ignore` to not block
```
PhilipGough, line 531: @wking - this will hopefully address your concern that we don't have a situation like https://github.com/openshift/cluster-monitoring-operator/pull/1203

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,573 @@
...
+instance).
+
+### API Extensions
+
+This enhancement proposal leverages the `AlertmanagerConfig` CRD which is
+exposed by the `coreos.monitoring.com/v1alpha1` API group and defined by the
+upstream Prometheus operator.
+
+The cluster monitoring operator deploys a `ValidatingWebhookConfiguration`
+resource to validate the `AlertmanagerConfig` resources.
```
sttts, line 157: for what?

enhancements/monitoring/multi-tenant-alerting.md
sttts: that means invalid data is accepted when the operator is down? That's not good.

sttts: Note: you can restrict the webhook to create and update.

PhilipGough: The upstream webhook implementation is already restricted to create and update.

sttts: What's the issue with CVO exactly? Why should it be able to create or update these objects without the webhook?

enhancements/monitoring/multi-tenant-alerting.md
sttts: What must be validated that cannot be expressed with OpenAPI ?

PhilipGough: Well, perhaps we are missing something and it can be done, but here is one example of an issue that we hit https://github.com/prometheus-operator/prometheus-operator/pull/4413#discussion_r752067596

enhancements/monitoring/multi-tenant-alerting.md
simonpasquier: To be honest, when I wrote this section, I copied-pasted what CMO does currently for PrometheusRules resources (which are the only "monitoring" resources that go through a webhook right now) and I didn't put more thoughts into it. But I agree that you have good points @sttts.

The issue with CVO was [BZ1949711](https://bugzilla.redhat.com/show_bug.cgi?id=1949711). Whenever someone deleted the openshift-monitoring namespace, CVO couldn't recreate the monitoring resources and got stuck. The "workaround" was to change the `failurePolicy` setting. Now I don't remember the exact root cause and I realize that deleting manually the validating webhook might have unblock CVO as well. The other reason why we chose `failurePolicy: Ignore` was to avoid blocking other operators when the prometheus operator is down (during upgrades for instance since it's a single replica).

In retrospect the reasons aren't solid and we should rather deploy a highly-available service for monitoring resources validation webhook.

enhancements/monitoring/multi-tenant-alerting.md
PhilipGough: Just to close this one out, this was a single example, but having combed through, there are several examples of things we cannot validate directly in OpenAPI, at least not generated by kubebuilder annotations. So we are resigned to doing what we can in the spec and filling the gaps with the validating webhook.

sttts: list-type set will fix prometheus operator's issue.

In general, try to avoid building APIs that cannot be validated via OpenAPI. Chance is high that those APIs won't be ideomatic anyway. There must be very strong reasons to cope with the operational complexity of a webhook.

PhilipGough: > list-type set will fix prometheus operator's issue.

That feels like a far-reaching assumption to be honest. For the very example [you suggest](https://github.com/prometheus-operator/prometheus-operator/pull/4413#discussion_r762766473) that it will fix, it will solve the duplicate issues, but if you look a couple of [lines down](https://github.com/prometheus-operator/prometheus-operator/pull/4413/files#diff-76b0f0103ac8d11eececdae9064b60fb4ee14845fdf6c295d48e43e00d4860b7R229), you will see that we have other issues to contend with, so it is not so straightforward.

We also have [other complex issues](https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1alpha1/alertmanager_config_types.go#L126-L135) which do not fit well into OpenAPI validation.

I am all for reducing complexity where we can, but in the end, we are trying to provide a mechanism to support https://prometheus.io/docs/alerting/latest/configuration/ which is not an API or config that we can change to suit OpenAPI validation, it is us who must design the API around the nuances there, and I think it is fair to say that we have not decided to reach for the webhook without valid reasons.

The alternative is we support only the simplest of use cases (which would probably discard this enhancement), or we risk accepting bad configuration and bring down alert manager in the process.

PhilipGough: @sttts - I fully understand and take on board your point by the way. Right now (and in the past it seems), we have been keen to provide a 1:1 mapping between upstream config(s) and CRD and perhaps we need to rethink that approach here to simplify validation and make the API more idiomatic.

I will raise that point in the community meeting for the Prometheus operator this week.

sttts: the JSON values make sense, I see that point. Take my comment above as a rule of thumb, not as a law :-)

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,494 @@
...
+
+* Cluster users can configure alert notifications for applications being
+monitored by the user-workload monitoring without requesting intervention from
+cluster admins.
+* Cluster admins can grant permissions to users and groups to manage alert
+routing scoped to individual namespaces.
+* Namespace owners should be able to opt-out from Alertmanager
+configuration (similar to what exist for service/pod monitors and rules using the
+`"openshift.io/user-monitoring: false"` label on the namespace).
+* Cluster admins should be able to opt-out from supporting `AlertmanagerConfig`
```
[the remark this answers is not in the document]
simonpasquier: no the cluster admin will have to enable it explicitly.

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,494 @@
...
+
+Even though the Prometheus operator prevents it as much as it can, it may be
+possible for users to create an `AlertmanagerConfig` resource that triggers the
+Prometheus operator to generate an invalid Alertmanager configuration, leading
+to a potential outage of the Platform Alertmanager cluster.
+
+Mitigations
+* The `AlertmanagerBadConfig` alert fires when Alertmanager can't reload its configuration.
+* Cluster admins can turn off the support for `AlertmanagerConfig` globally so that the Platform Alertmanager cluster can process platform alerts again and the cluster admins have time to identiy the "rogue" `AlertmanagerConfig` resource(s).
+* Cluster admins can exclude specific user namespaces (once the "rogue" `AlertmanagerConfig` resource(s) have been identified) to restore UWM functionality for good citizens.
```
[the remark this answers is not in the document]
simonpasquier: Eventually the operator will update the status field for AlertmanagerConfig resources that the operator doesn't reconcile (see `#### AlertmanagerConfig resources not being reconciled`) below.

simonpasquier:
@dofinn  @wanghaoran1988 could you have another look. The proposal now encompasses the ability to have distinct Alertmanagers: one for platform alerts and one for user alerts.

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,601 @@
...
+(e.g. OSD), it might not be acceptable for cluster admins to let users
+configure the Platform Alertmanager with `AlertmanagerConfig` resources because:
+* User configurations may break the Alertmanager configuration.
+* Processing of user alerts may slow down the alert notification pipeline.
+* Cluster admins don't want to deal with delivery errors for user notifications.
+
+At the same time, application owners want to configure their alert
+notifications without requesting external intervention.
+
+In this case, UWM admins have the possibility to deploy a dedicated
```
dofinn, line 261: This conflicts with https://github.com/openshift/enhancements/pull/937/files#diff-02ac51e89242574f67f8eb856975563d598d4b8bbdc18f7ce848ed135a9a8de5R87 ?

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,601 @@
...
+* The `AlertmanagerBadConfig` alert fires when Alertmanager can't reload its configuration.
+* Cluster admins can exclude specific user namespaces (once the "rogue" `AlertmanagerConfig` resource(s) have been identified) to restore UWM functionality for good citizens.
+* When alerts are sent to the Platform Alertmanager, cluster admins can turn off the support for `AlertmanagerConfig` in the CMO configmap so that the Platform Alertmanager cluster can process platform alerts again and the cluster admins have time to identiy the "rogue" `AlertmanagerConfig` resource(s).
+
+#### Misconfiguration of receivers
+
+Users may provide bad credentials for the receivers, the system receiving the
+notifications might be unreachable, or the system might be unable to process
+the requests. These situations would trigger the
+`AlertmanagerFailedToSendAlerts` and/or `AlertmanagerClusterFailedToSendAlerts`
```
dofinn, line 449: In the context of external AlertManager and/or UWM/AlertManager - these alerts routed via openshift-monitoring/alertmanger thus alerting SRE that a customer has miss-configured their own AlertManager stack ? That is ok, as we can short circuit this use-case to service log and remove SRE from the loop... just wanted to clarify.

dofinn (commented):
@simonpasquier great job on this :)

Left some minor notes for clarifications. One thing that is not clear for me is how the UWM/prom and/or UWM/thanos are configurable to send to an external AM? I dont see anything like that available in the UWM/user-workload-monitoring-config yet it is mentioned as an option.

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,601 @@
...
+
+The configuration of Alertmanager is done via a single configuration file that
+only cluster admins have permissions to modify. If the cluster is shared by
+multiple tenants and each tenant has different requirements to receive their
+notifications then each tenant needs to ask and wait for the cluster admins to
+adjust the Alertmanager configuration.
+
+To streamline the process and avoid cluster admins being the bottleneck,
+application owners should be able to configure alert routing and notification
+receivers in the Plaform Alertmanager without cluster admin intervention.
```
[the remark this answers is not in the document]
simonpasquier: sure!

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,601 @@
...
+
+We plan to leverage the `AlertmanagerConfig` custom resource definition already
+exposed by the Prometheus operator so that application owners can configure how
+and where their alert notifications should be routed.
+
+### User Stories
+
+Personas:
+* Application owners: manage a project with sufficient permissions to define monitoring resources.
+* UWM admins: manage the configuration of the UWM components (edit permissions on the `openshift-user-workload-monitoring/user-workload-monitoring-config` configmap).
```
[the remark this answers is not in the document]
simonpasquier: Thanks for the info, I wasn't aware of this issue. I don't see an easy solution either...

enhancements/monitoring/multi-tenant-alerting.md
simonpasquier: correct I will remove it from the non-goals.

enhancements/monitoring/multi-tenant-alerting.md
simonpasquier: this is correct: these alerts will be routed to the Platform Alertmanager.

simonpasquier (commented):
@dofinn since OpenShift 4.9, it's possible to configure external Alertmanagers for both UWM Prometheus and Thanos Ruler :arrow_right:  https://docs.openshift.com/container-platform/4.9/monitoring/configuring-the-monitoring-stack.html#monitoring-configuring-external-alertmanagers_configuring-the-monitoring-stack

enhancements/monitoring/multi-tenant-alerting.md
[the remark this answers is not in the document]
simonpasquier: Are you talking about prometheus-operator only? Or UWM Prometheus and Thanos Ruler?
Aren't the control plane's nodes tainted to avoid such situation?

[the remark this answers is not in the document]
dofinn: It has tolerations by default :) https://github.com/openshift/managed-cluster-config/pull/990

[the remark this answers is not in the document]
simonpasquier: Thanks for the details, I understand better now :)

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,606 @@
...
+
+Users are able to silence alerts from the Dev Console and the console backend
+assumes that the API is served by the
+`alertmanager-main.openshift-monitoring.svc` service. To support the UWM
+Alertmanager configuration, CMO should provide to the console operator the name
+of the Alertmanager service managing the user alerts (either
+`alertmanager-main.openshift-monitoring.svc` or
+`alertmanager.openshift-user-workload-monitoring.svc`). Based on the presence
+of the `openshift_io_alert_source` label, the console backend can decide which
+Alertmanager service should be queried.
```
simonpasquier, line 508: @kyoto @spadgett i'd welcome your inputs on this part. If we allow users to deploy a dedicated Alertmanager for user-defined alerts then this has an impact on the console since the API endpoint to manage the silences depends on the origin of the alert (platform vs. user). Do you see a problem with this?
Implementation-wise, CMO could reuse the `monitoring-shared-config` configmap in the `openshift-managed-config` namespace but I'm open to suggestions.

kyoto: If I understand correctly, this means the Dev Console frontend code will barely change and the Console backend will decide which Alertmanager to proxy the create silence calls to.

When listing the existing silences, both Alertmanagers will need to be queried and the results merged. I guess this merging could be done either by the backend or the frontend, but if the backend is already deciding which Alertmanager to proxy calls to, then maybe it makes sense for the backend to also do the merging.

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,646 @@
...
+- source_matchers: ['namespace="foo"', ...]
+  target_maTCHERS: ['namespace="foo"', ...]
+  equal: ['namespace', ...]
+receivers:
+# The next item is generated from an AlertmanagerConfig resource named alertmanagerconfig1 in namespace foo.
+- name: foo-alertmanagerconfig1-my-receiver
+  ...
+```
+
+#### Console integration
```
simonpasquier, line 390: @kyoto can you have a look at this section? Does it make sense to you?

simonpasquier:
/remove-lifecycle stale

The proposal is in good shape now, I think all comments have addressed. 
@jeremyeder @dofinn can you give it a final review pls?

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,636 @@
...
+
+The validating webhook points to
+the prometheus operator's service in the `openshift-user-workload-monitoring`
+namespace (path: `/admission-alertmanagerconfigs/validate`).
+
+### Implementation Details/Notes/Constraints
+
+#### Deployment models
+
+##### Leveraging the Platform Alertmanager
```
dofinn, line 173: Inhibition is a global setting... is it possible that multi tenant AlertManager CRs could unknowingly inhibit each other?

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,494 @@
...
+
+#### Disruption of the platform Alertmanager
+
+Even though the Prometheus operator prevents it as much as it can, it may be
+possible for users to create an `AlertmanagerConfig` resource that triggers the
+Prometheus operator to generate an invalid Alertmanager configuration, leading
+to a potential outage of the Platform Alertmanager cluster.
+
+Mitigations
+* The `AlertmanagerBadConfig` alert fires when Alertmanager can't reload its configuration.
```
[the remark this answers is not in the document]
dofinn: cc @mrbarge another candidate for ocm-agent :)

enhancements/monitoring/multi-tenant-alerting.md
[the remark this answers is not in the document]
dofinn: @simonpasquier will this render the co/monitoring degraded?

enhancements/monitoring/multi-tenant-alerting.md
```diff
@@ -0,0 +1,636 @@
...
+metadata:
+  name: user-workload-monitoring-config
+  namespace: openshift-user-workload-monitoring
+data:
+  config.yaml: |-
+    alertmanager:
+      enabled: true
+      enableUserAlertmanagerConfig: true
+      logLevel: info
+      nodeSelector: {...}
```
dofinn, line 281: @simonpasquier will this be workers by default?

simonpasquier: Yes, it should be similar to what happens for the other monitoring components. When nodeSelector and tolerations aren't defined, all worker nodes are eligible and master nodes are excluded (because they are tainted).

enhancements/monitoring/multi-tenant-alerting.md
simonpasquier: yes because the operator will always append a namespace matcher to the inhibition rules, e.g.

```
inhibit_rules:                                                                                      
- target_matchers:                                                                                  
  - alertname="HttpErrors"                                                                          
  source_matchers:                                                                                  
  - alertname=~"DatabaseErrors"                                                                       
  equal:                                                                                            
  - service                                                                                            
```

is turned into

```
inhibit_rules:                                                                                      
- target_matchers:                                                                                  
  - alertname="HttpErrors"
  - namespace="foo"                                                                          
  source_matchers:                                                                                  
  - alertname=~"DatabaseErrors"                                                                       
  - namespace="foo"                                                                          
  equal:                                                                                            
  - service                                                                                            
```

enhancements/monitoring/multi-tenant-alerting.md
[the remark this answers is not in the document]
simonpasquier: If a dedicated Alertmanager cluster is deployed for user-defined alerts, the platform SREs can decide where the `AlertmanagerBadConfig` alert for `namespace="openshift-user-workload-monitoring"` should be routed. I would guess that something similar exists today for alerts about the UWM components (Prometheus + Thanos Ruler).

enhancements/monitoring/multi-tenant-alerting.md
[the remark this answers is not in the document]
simonpasquier: As of today, CMO doesn't check for alerts so Alertmanager (or Prometheus) failing to reload the configuration doesn't change the co/monitoring conditions.

enhancements/monitoring/multi-tenant-alerting.md
dofinn: Excellent, the current prom-operator from UWM schedules itself on masters :)

```
$ oc -n openshift-user-workload-monitoring get pods prometheus-operator-68fd48b89d-6tp7s -o json | jq '.spec.tolerations[]'
{
  "effect": "NoSchedule",
  "key": "node-role.kubernetes.io/master",
  "operator": "Exists"
}
{
  "effect": "NoExecute",
  "key": "node.kubernetes.io/not-ready",
  "operator": "Exists",
  "tolerationSeconds": 300
}
{
  "effect": "NoExecute",
  "key": "node.kubernetes.io/unreachable",
  "operator": "Exists",
  "tolerationSeconds": 300
}
{
  "effect": "NoSchedule",
  "key": "node.kubernetes.io/memory-pressure",
  "operator": "Exists"
}
```

simonpasquier: yes this is following the OpenShift convention for operators -> https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#taints-and-tolerations

dofinn:
> /remove-lifecycle stale
> 
> The proposal is in good shape now, I think all comments have addressed. @jeremyeder @dofinn can you give it a final review pls?

Done ! Awesome work on this @simonpasquier ! /lgtm

PR #939, opened 2021-10-19
[OCPNODE-725] Control Group v2 Enablement on New Clusters

rphillips:
[OCPNODE-725](https://issues.redhat.com/browse/OCPNODE-725)

This enhancement describes enabling cgroup v2 within OpenShift.

### DIFF
--- a/enhancements/machine-config/mco-cgroupsv2-support.md
+++ b/enhancements/machine-config/mco-cgroupsv2-support.md
@@ -1,0 +1,219 @@
+---
+title: Control Group v2 Enablement
+authors:
+  - "@rphillips"
+reviewers:
+  - "@mrunalp"
+  - "@kikisdeliveryservice"
+  - "@sinnykumari"
+  - "@yuqi-zhang"
+  - "@cgwalters"
+approvers:
+  - "@mrunalp"
+  - "@sinnykumari"
+api-approvers:
+  - "@sttts"
+creation-date: 2021-10-19
+last-updated: 2021-10-20
+status: implementable
+---
+
+# Control Group v2 Enablement on New Clusters
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [x] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Definitions and References
+
+- cgroup: control group and is never capitalized
+- cgroups: multiple control cgroups
+- cgroup v1: references cgroup version 1 implementation
+- cgroup v2: references cgroup version 2 implementation
+
+[Upstream Docs](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html)
+
+## Summary
+
+Control Group v2 (cgroup v2) enablement in Kubernetes has progressed to beta
+[upstream](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2254-cgroup-v2).
+The underlying runtime (cri-o) and supporting subsystems are now ready for
+customers to begin their own testing with it. Not all workloads will be
+compatible with cgroup v2, so it will *not* be enabled by default within
+OpenShift at this time.
+
+Note: This enhancement is focusing on `pure` mode cgroup v2. Mixed mode environments
+may behave differently (metrics, vpa, hpa, etc) since cgroup v1 is not
+compatible with cgroup v2.
+
+## Motivation
+
+Migrating to cgroup v2 will bring in many new features and fixes not found in
+cgroup v1. cgroup v1 is considered 'legacy' and migrating to cgroup v2 is
+necessary since RHEL8 ships with cgroup v2 on by default. (OpenShift 4.x
+currently disables cgroup v2 in favor of v1).
+
+Some features of cgroup v2 include:
+
+* IO enhancements
+* User based OOM killer
+* cgroup aware OOM killer
+
+[Kubernetes On Cgroup v2 - Video](https://www.youtube.com/watch?v=u8h0e84HxcE)
+
+### Goals
+
+- [ ] Enable cgroup v2 within the Openshift API
+- [ ] Add kernel flags to MCO to enable cgroup v2 on nodes
+
+### Non-Goals
+
+Mixed mode cgroup modes are not 100% compatible with each other. We need data
+around how cgroup v2 runs in a pure mode before we can allow mixed mode
+environments. Since, Red Hat is steering upstream cgroup v2 adoption, and we do
+not have data around the pure mode environment yet, there needs to be a platform
+to gather data from.
+
+## Proposal
+
+The option to enable cgroup v2 will have to reside in a centralized location.
+The [OpenShift Infrastructure config
+object](https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L28)
+contains information describing how a cluster functions including cloud config  
+and platform specification for each cloud. Setting the cgroup mode is an
+infrastructure setting.
+
+### API Extensions
+
+Create an additional module within openshift/api as `config/v1/types_node.go`:
+
+```go
+type CgroupMode string
+
+const (
+  CgroupMode_Empty = "" // Empty string will always use the Default
+  CgroupMode_v1 = "v1"
+  CgroupMode_v2 = "v2"
+  CgroupMode_Default = CgroupMode_v1
+)
+
+type Node struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// +required
+	Spec NodeSpec `json:"spec"`
+}
+
+type NodeSpec struct {
+  CgroupMode CgroupMode `json:"cgroupMode,omitempty"`
+
+  // an eventual additional option might be crun in the future. This explains
+  //   why a new struct may be necessary
+  // CrunEnabled bool ...
+}
+```
+
+### Operational Aspects of API Extensions
+
+Once the previous API is defined, the MCO will read the configured object and
+set the appropriate kernel options (on bootstrap). The MCO will report an error
+if a user tries to modify/add cgroup related kargs within a MachineConfig.
+
+
+The following kernel command line arguments would be set when `CgroupMode_v2` is enabled:
+```yaml
+  kernelArguments:
+    - systemd.unified_cgroup_hierarchy=1
+    - cgroup_no_v1="all"
+    - psi=1 
+```
+
+#### Failure Modes
+
+N/A
+
+#### Support Procedures
+
+### User Stories
+
+#### 1. As a user, I would like to install a cluster that uses cgroup v2
+
+### Risks and Mitigations
+
+The primary risk of cgroup v2 is some workloads not supporting the changed
+cgroup filesystem paths. This is the reason why it is off by default and has to
+be enabled at install time.
+
+## Design Details
+
+### Test Plan
+
+Testing should be thoroughly done at all levels, including unit, end-to-end, and
+integration.
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+
+cgroup v2 will be dev preview on its initial release. Internal and customer
+usage will be critical to gather information on bugs and enhancements to the
+underlying subsystem.
+
+Graduation requirements to Tech Preview are:
+
+* No regressions from cgroup v1 to cgroup v2
+* Processes are correctly OOMed
+* CPU management works as expected
+* No performance issues - PSAP and QE teams will be asked to test their suites for regressions
+* Metrics are accurate and correctly submitted to monitoring
+
+#### Tech Preview -> GA
+
+With sufficient internal testing and customer feedback the feature will graduate
+to Tech Preview.
+
+Upon graduation to GA the feature will still be turned off by default until we define another enhancement to specify how cgroup v2 is enabled by default within OpenShift.
+
+Graduation requirements to GA:
+* Upstream graduation to GA
+* Internal stakeholders are using cgroup v2 without issue
+* Tech Preview Graduation requirements are still good
+* Add blocking cgroup v2 upgrade jobs
+* CI OpenShift cgroup v2 upgrade jobs pass percentage is similar or better than the OpenShift cgroup v1 upgrade job pass percentage
+
+The following jobs will be run against cgroup v2 periodically and with a minimum of 100 runs. The test pass percentage for each test/job tuple must not be demonstrably worse for any of the following jobs:
+
+- periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upgrade
+- periodic-ci-openshift-release-master-ci-4.10-e2e-azure-ovn-upgrade
+- periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-gcp-ovn-upgrade
+- periodic-ci-openshift-release-master-ci-4.10-e2e-aws-ovn-upgrade
+- periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-aws-ovn-upgrade
+- periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-azure-upgrade
+- periodic-ci-openshift-release-master-ci-4.10-e2e-gcp-upgrade
+- periodic-ci-openshift-release-master-nightly-4.10-upgrade-from-stable-4.9-e2e-metal-ipi-upgrade-ovn-ipv6
+- periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-vsphere-upgrade
+
+### Upgrade / Downgrade Strategy
+
+Downgrading a cluster to an OpenShift version not containing cgroup v2 support
+is unsupported.
+
+### Version Skew Strategy
+
+A cluster installed with cgroup v2 will abide by the usual skew upgrade path.
+
+#### Removing a deprecated feature
+
+N/A
+
+## Implementation History
+
+## Alternatives
+
+## Drawbacks


enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,121 @@
...
+## Design Details
+
+### Test Plan
+
+Testing should be thoroughly done at all levels, including unit, end-to-end, and
+integration.
+
+### Graduation Criteria
+
+#### Tech Preview -> GA
```
kikisdeliveryservice: What are the actual graduation criteria for tech preview? ie what do you expect to deliver

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,123 @@
...
+```yaml
+  kernelArguments:
+    - systemd.unified_cgroup_hierarchy=1
+    - cgroup_no_v1="all"
+    - psi=1 
+```
+
+### User Stories
+
+#### Story 1
```
kikisdeliveryservice, line 81: what is the user story here?

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,123 @@
...
+
+Add Cgroupsv2 flag to Machine Config Pool type
+
+This story would also require wiring up the kernel commandline options to enable
+or disable Cgroupsv2 on the necessary pool.
+
+Acceptance Criteria
```
kikisdeliveryservice, line 88: this section seems misplaced in user stories?

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Upgrade
+
+Downgrading a cluster to an Openshift version not containing Cgroupsv2 support
+with a CgroupsV2 enabled MCP would cause the Machine Config Pool to enable
+Cgroupsv1. We can call this out in documentation.
+
+### Version Skew Strategy
+
+Version skew should not have an impact on this feature. The change would only
```
kikisdeliveryservice, line 124: if this is only applicable to new nodes can we call that out above as well

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,123 @@
...
+
+* IO enhancments
+* User based OOM killer
+* Cgroup aware OOM killer
+
+[Kubernetes On Cgroupsv2 - Video](https://www.youtube.com/watch?v=u8h0e84HxcE)
+
+### Goals
+
+- [ ] Enable Cgroupsv2 on a Machine Config Pool
```
kikisdeliveryservice, line 52: This is only for new nodes (mentioned below) if so can we call that out a bit clearer here?

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: Updated with graduation requirements

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Proposal
+
+[Machine Config
+Pools](https://github.com/openshift/machine-config-operator/blob/85f9f9583451cb1e4c25893598eb5d086fd7e4ba/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L191)
+have a FIPS flag on them to enable `FIPS` within the system. Adding a `Cgroupsv2
+bool` flag to the Machine Config Pool, when true would enable Cgroupsv2 on the
+pool.
+
+The installer passes down the `FIPS` flag to MCO on bootstrap. Similar
```
yuqi-zhang, line 66: Again, if we want to draw parallel to FIPs, the questions would be:

1. does this need to happen on firstboot in the early boot phase, or is post-first-pivot (booting into the right configuration) enough
2. can this be configured per pool or as a cluster level config
3. can this be changed after the fact, or is this purely install time

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,150 @@
...
+
+cgroupsv2 enablement in Kubernetes has progressed to the point that it is
+usable. The underlying runtime (cri-o) and supporting subsystems are now ready
+for customers to begin their own testing with it. Not all workloads will be
+compatible with cgroupsv2, so it will *not* be enabled by default within
+Openshift at this time.
+
+## Motivation
+
+Migrating to Cgroupsv2 will bring in many new features and fixes not found in
```
cgwalters, line 37: The [upstream kernel docs](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html) seem to use "cgroup v2".  The capitalization here is also inconsistent.  Let's both link to the upstream kernel docs, and consistently say "cgroup v2" to match?  (i.e. not "cgroupsv2" or "Cgroupsv2")

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,150 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+cgroupsv2 enablement in Kubernetes has progressed to the point that it is
+usable. The underlying runtime (cri-o) and supporting subsystems are now ready
+for customers to begin their own testing with it. Not all workloads will be
+compatible with cgroupsv2, so it will *not* be enabled by default within
+Openshift at this time.
```
cgwalters, line 33: s/Openshift/OpenShift/g

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,150 @@
...
+### Non-Goals
+
+- Support mixed cgroups modes
+
+## Proposal
+
+[Machine Config
+Pools](https://github.com/openshift/machine-config-operator/blob/85f9f9583451cb1e4c25893598eb5d086fd7e4ba/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L191)
+have a FIPS flag on them to enable `FIPS` within the system. Adding a `Cgroupsv2
+bool` flag to the Machine Config Pool, when true, would enable Cgroupsv2 on the
```
cgwalters, line 63: Pools don't have properties like this.  FIPS is a flag in a MachineConfig object, which applies to a pool.

I'd bikeshed this to match [the systemd terminology](https://systemd.io/CGROUP_DELEGATION/) and have
`controlGroupHierarchy: unified|legacy` where `legacy` is the current default.

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: I've updated the doc with a references and definitions section. good idea.

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,192 @@
...
+
+### User Stories
+
+#### 1. Modify the Openshift API with the agreed upon spec
+
+#### 2. Modify the MCO to accept the new Openshift API and enable cgroup v2 kernel options (bootstrap)
+
+#### 3. Modify the MCO to accept the new Openshift API and enable cgroup v2 kernel options (day 2)
+
+### Risks and Mitigations
```
sinnykumari, line 134: Although this is an implementation detail for MCO, but with adding cgroup option MCO will need to ensure that user is not also adding any cgroup v2 specific kargs via a MachineConfig. Otherwise, I think this can lead to undesired behavior when cluster infraSpec has cgroup v1.

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,192 @@
...
+point that it is usable. The underlying runtime (cri-o) and supporting
+subsystems are now ready for customers to begin their own testing with it. Not
+all workloads will be compatible with cgroup v2, so it will *not* be enabled by
+default within OpenShift at this time.
+
+## Motivation
+
+Migrating to cgroup v2 will bring in many new features and fixes not found in
+cgroup v1. cgroup v1 is considered 'legacy' and migrating to cgroup v2 is
+considered necessary since RHEL ships with cgroup v2 on by default. (OpenShift
```
cgwalters, line 48: s/RHEL/RHEL9/

(Also of note related to this, current Fedora defaults to v2, and I believe OKD is still reverting that - hopefully we can also validate this flag on OKD)

When we're thinking about RHEL9 - would one expect that upgrading to that automatically inherits the OS default to cgroup v2, or would the admin need to flip on this flag in addition?  I would hope the former.

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,192 @@
...
+ // +genclient
+ // +genclient:nonNamespaced
+ // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+@@ -45,6 +52,13 @@ type InfrastructureSpec struct {
+ 	// platformSpec holds desired information specific to the underlying
+ 	// infrastructure provider.
+ 	PlatformSpec PlatformSpec `json:"platformSpec,omitempty"`
++
++	// cgroupSpec holds the desired cgroup version to use for a cluster
++	CgroupSpec CgroupSpec `json:"cgroupSpec,omitempty"`
```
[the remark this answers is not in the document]
kikisdeliveryservice: looking below seems to suggest it will be off by defaul in GA as well? (see line 167)

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,192 @@
...
+[Kubernetes On Cgroup v2 - Video](https://www.youtube.com/watch?v=u8h0e84HxcE)
+
+### Goals
+
+- [ ] Enable cgroup v2 within the Openshift API
+- [ ] Add kernel flags to MCO to enable cgroup v2 on nodes
+
+### Non-Goals
+
+- Support mixed cgroups modes
```
mrunalp, line 66: Thinking about this more, maybe hypershift may want to run control planes and worker nodes differently.

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Proposal
+
+The option to enable cgroup v2 will have to reside in a centralized location.
```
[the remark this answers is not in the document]
cgwalters: I'm not advocating specifically to emphasize this API at a granular level.

I'm more saying two things:

- The hard reality is that clusters will be mixed during upgrades, at least for some period of time, so it needs to at least not actively fail
- We know there are people out there that are e.g. pausing the `worker` pool for long periods of time, so again those situations will be mixed.  Unless we try to *actively block* updates based on this (in the same way there's some MCO work to block upgrades on having too-old workers)
- Internally in the MCO, pools roll out machineconfig.  Expressing this as a machineconfig feels natural.

That said, the more we add things like this, the more need I see for making it easy for a `MachineConfig` object to apply to all node types.  

But today higher level flags in the install config like `hyperThreading` and `fips` are represented as a generated `MachineConfig` for `master` and `worker`.

rphillips:
Updated the enhancement to clarify this is for new clusters and added a new API.

enhancements/machine-config/mco-cgroupsv2-support.md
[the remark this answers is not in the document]
rphillips: So far, we do not have any restrictions on which clusters cgroup v2 can be enabled on. We will need to gather bug reports and feedback on what is working (and not working).

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: The intent is to enable this by flag at install time.

rphillips:
I believe this enhancement is ready for a final review.

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,199 @@
+---
+title: Control Group v2 Enablement
+authors:
+  - "@rphillips"
+reviewers:
+  - "@mrunalp"
```
kikisdeliveryservice, line 6: can you update the reviwers section?

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,199 @@
...
+be enabled within OpenShift at a later time.
+
+Graduation requirements to GA:
+* Upstream graduation to GA
+* Internal stakeholders are using cgroup v2 without issue
+* Tech Preview Graduation requirements are still good
+
+### Upgrade / Downgrade Strategy
+
+Downgrading a cluster to an OpenShift version not containing cgroup v2 support
```
kikisdeliveryservice, line 182: Are these 2 sentences redundant?

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,199 @@
...
+### Test Plan
+
+Testing should be thoroughly done at all levels, including unit, end-to-end, and
+integration.
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+
+cgroup v2 will be dev preview on it's initial release. Internal and customer
```
kikisdeliveryservice, line 155:
```suggestion
cgroup v2 will be dev preview on its initial release. Internal and customer
```

minor (sorry)

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,199 @@
...
+usage will be critical to gather information on bugs and enhancements to the
+underlying subsystem.
+
+Graduation requirements to Tech Preview are:
+
+* No regressions from cgroup v1 to cgroup v2
+* Processes are correctly OOMed
+* CPU management works as expected
+* No performance issues
```
kikisdeliveryservice, line 164: q: are we comparing this to some benchmarking done with v1? or just relying on bz?

kikisdeliveryservice (commented):
Looks good, just a few last comments :)

sinnykumari:
Since this enhancement impacts MCO as well, wanted to know in which OpenShift version we are expecting this to be get implemented(dev preview)?

rphillips:
@sinnykumari @kikisdeliveryservice thank you. updated the PR and mentioned we can target 4.10 for the changes.

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: updated with a blurb about QE and PSAP asks.

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: edited. thanks!

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: done!

sinnykumari:
> @sinnykumari @kikisdeliveryservice thank you. updated the PR and mentioned we can target 4.10 for the changes.

umm, not sure but shouldn't a enhancement be merged before the release planning of a particular release?

rphillips:
@sinnykumari that does make sense to me... removed it from the doc.

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,202 @@
...
+authors:
+  - "@rphillips"
+reviewers:
+  - "@mrunalp"
+  - "@kikisdeliveryservice"
+  - "@sinnykumari"
+  - "@yuqi-zhang"
+  - "@cgwalters"
+approvers:
+  - "@yuqi-zhang"
```
kikisdeliveryservice, line 12: As TL I think @sinnykumari should be the MCO approver here

rphillips: updated

rphillips:
@sttts could you take a pass through this?

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,206 @@
...
+
+### API Extensions
+
+Create an additional module within openshift/api as `config/v1/types_node.go`:
+
+```go
+type CgroupMode string
+
+const (
+	CgroupMode_v1 = ""
```
sttts, line 99: make this explicit, i.e. `v1` and use a default for the field below.

sttts: Looks like you might want to have 3 values? v1, v2, "" (= I want the default).

Will `v1` as explicit opt-out of v2 be a thing at some point?

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,206 @@
...
+* No performance issues - PSAP and QE teams will be asked to test their suites for regressions
+* Metrics are accurate and correctly submitted to monitoring
+
+#### Tech Preview -> GA
+
+With sufficient internal testing and customer feedback the feature will graduate
+to Tech Preview.
+
+Upon graduation to GA the feature will still be turned off by default, but may
+be enabled within OpenShift at a later time.
```
sttts, line 181: do you expect a period where it is on by default, but can be disabled?

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,206 @@
...
+Testing should be thoroughly done at all levels, including unit, end-to-end, and
+integration.
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+
+cgroup v2 will be dev preview on its initial release. Internal and customer
+usage will be critical to gather information on bugs and enhancements to the
+underlying subsystem.
```
sttts, line 165: this means it will be under a feature gate? What if the gate is not enabled, but the user opts into v2?

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,206 @@
...
+```go
+type CgroupMode string
+
+const (
+	CgroupMode_v1 = ""
+	CgroupMode_v2 = "v2"
+	CgroupMode_Default = CgroupMode_v1
+)
+
+type NodeOptions struct {
```
sttts, line 104: `<bike-shedding>`Why not just `Node`? Everything in config/v1 are options.`</bike-shedding>`

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: I updated this sentence in f69f9c7 to address that another enhancement will be created to specify how the feature will be turned on by default. The original sentence here was not clear with my other modifications to the document.

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: Fixed this in f69f9c7.

v1 might be an explicit opt out due to customer workloads not supporting cgroup v2.

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: The feature is not under a feature gate at the moment. cgroup v2 will only be enabled at install time.

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,206 @@
...
+type Node struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// +required
+	Spec NodeSpec `json:"spec"`
+}
+
+type NodeSpec struct {
+  CgroupMode CgroupMode `json:"cgroupMode,omitempty"`
```
deads2k, line 114: This field makes sense to me.  In the actual api PR, you'll need to fill in godoc for users.

sttts, line 114: are we sure it stays at a string? Or should we make it a struct from the beginning?
```
cgroup:
  mode: v2
```

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,206 @@
...
+type CgroupMode string
+
+const (
+  CgroupMode_Empty = "" // Empty string will always use the Default
+  CgroupMode_v1 = "v1"
+  CgroupMode_v2 = "v2"
+  CgroupMode_Default = CgroupMode_v1
+)
+
+type Node struct {
```
deads2k, line 105: @sttts you devoted some thought to the top level object name, does this match your preference?  I have not devoted enough think-time to have a clear opinion.

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: Other parts of the API do the same thing... Making it a struct seems overkill for something that is likely not going to change again for a long time.

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: I started with NodeOptions... someone mentioned on this thread Node makes more sense. I agree with simplifying it to Node.

sttts: will the latency settings (in some other enhancement) go here too?

rphillips: Latency is a multi component api addition. I do not believe it goes in this object.

rphillips:
@deads2k I added two blurbs in 54031d4: adding a blocking cgroup v2 upgrade job, and a comment about how the pass percentage should be the same or better with the cgroup v2 job.

enhancements/machine-config/mco-cgroupsv2-support.md
sttts: It's only about kubelet, and others to cope with the fallout. Seem to be very Node'ish.

rphillips:
* non-blocking for this PR @deads2k to provide a list of openshift jobs for GA

rphillips:
```
periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upgrade
periodic-ci-openshift-release-master-ci-4.10-e2e-azure-ovn-upgrade
periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-gcp-ovn-upgrade
periodic-ci-openshift-release-master-ci-4.10-e2e-aws-ovn-upgrade
periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-aws-ovn-upgrade
periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-azure-upgrade
periodic-ci-openshift-release-master-ci-4.10-e2e-gcp-upgrade
periodic-ci-openshift-release-master-nightly-4.10-upgrade-from-stable-4.9-e2e-metal-ipi-upgrade-ovn-ipv6
periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-vsphere-upgrade
```

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,220 @@
...
+- periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-vsphere-upgrade
+
+### Upgrade / Downgrade Strategy
+
+Downgrading a cluster to an OpenShift version not containing cgroup v2 support
+is unsupported.
+
+### Version Skew Strategy
+
+A cluster installed with cgroup v2 would not be impacted by the version skew
```
deads2k, line 209: you will be when "Add blocking cgroup v2 upgrade jobs" is done for your GA criteria.  You plan to revisit?

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,220 @@
...
+[Kubernetes On Cgroup v2 - Video](https://www.youtube.com/watch?v=u8h0e84HxcE)
+
+### Goals
+
+- [ ] Enable cgroup v2 within the Openshift API
+- [ ] Add kernel flags to MCO to enable cgroup v2 on nodes
+
+### Non-Goals
+
+Mixed mode cgroup modes are not 100% compatible with each other. We need data
```
deads2k, line 76: this precludes migrating from cgroups v1 to cgroups v2.  Can you explode that into a  top level "non-goal" so we know it this enhancement will need revisiting to remove cgroups v1.

enhancements/machine-config/mco-cgroupsv2-support.md
```diff
@@ -0,0 +1,220 @@
...
+Upon graduation to GA the feature will still be turned off by default until we define another enhancement to specify how cgroup v2 is enabled by default within OpenShift.
+
+Graduation requirements to GA:
+* Upstream graduation to GA
+* Internal stakeholders are using cgroup v2 without issue
+* Tech Preview Graduation requirements are still good
+* Add blocking cgroup v2 upgrade jobs
+* CI OpenShift cgroup v2 upgrade jobs pass percentage is similar or better than the OpenShift cgroup v1 upgrade job pass percentage
+
+The following jobs will be run against cgroup v2 periodically and with a minimum of 100 runs:
```
deads2k, line 190: In addition to this sentence.

The test pass percentage for each test/job tuple must not be demonstrably worse for any of the following jobs.

enhancements/machine-config/mco-cgroupsv2-support.md
rphillips: fixed.

PR #945, opened 2021-11-01
readme: describe how to handle linter failures caused by template changes

dhellmann:


aravindhp (approved):
/approve

README.md
```diff
@@ -99,3 +99,22 @@ Ideally pull requests with enhancement proposals will be merged before
...
+From time to time the template for enhancement proposals is modified
+as we refine our processes. When that happens, open pull requests may
+start failing the linter job that ensures that all documents include
+the required sections.
+
+If you are working on an enhancement and the linter job fails because
+of changes to the template (not other issues with the markdown
+formatting), handle it based on the maturity of the enhancement PR:
+
+* If you have a PR that is ready for approval, override the job to
```
kikisdeliveryservice, line 114: one thought: ready for approval seems to rely on the authors whereas the next bp relies on more people. could we make something similar? 

something like: If you have a PR that has reached a point of consensus and been reviewed by (most of the???) affected parties...

kikisdeliveryservice (commented):
one thought/clarification

README.md
dhellmann: How about this: "If the only reason to update your PR is to make the linter job accept it after a template change and there are no substantive content changes needed for approval."

kikisdeliveryservice: I think that works!

dhellmann: Updated

PR #947, opened 2021-11-03
Insights pulling SCA certs - graduation criteria update

tremes:


enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -138,7 +138,7 @@ This feature is planned as a technical preview in OCP 4.9 and is planned to go G
 - this new functionality is documented in the Insights Operator documentation as tech preview
 
 #### Tech Preview -> GA
-- `Share` resource object is automatically created by the Insights Operator
+- ~~`Share` resource object is automatically created by the Insights Operator~~ - `Share` object will still be a tech preview in 4.10. Build team can enhance/create it later
```
bparees, line 141: i don't think it's important to reference version numbers in EPs, the roadmap is intended to define the technical milestones that the feature will go through/requirements at each milestone, not the specific release when we expect to achieve them.

Since this is still the section listing tech preview to GA requirements, are you trying to say that the Share object creation will now *not* be a GA requirement?  If so, just say that (along with why, if you want, and how/when the Share object will be introduced).

tremes: Yes :) OK let me try to update. Thanks

PR #953, opened 2021-11-08
template: remove status field and replace with jira field

kikisdeliveryservice:


guidelines/enhancement_template.md
```diff
@@ -12,7 +12,8 @@ api-approvers: # in case of new or modified APIs or API extensions (CRDs, aggreg
   - TBD
 creation-date: yyyy-mm-dd
 last-updated: yyyy-mm-dd
-status: provisional|implementable|implemented|deferred|rejected|withdrawn|replaced|informational
+jira link: # link to the Jira Feature or Epic ticket that corresponds to this enhancement
```
dhellmann, line 15: Let's call this something more generic, since teams outside of Red Hat may use other tools for tracking. Maybe "tracking_link"?

Would it make sense to move this out of the "header" and into the body of the document somewhere?

guidelines/enhancement_template.md
```diff
@@ -74,6 +75,7 @@ around the enhancement process.
 ## Release Signoff Checklist
 
 - [ ] Enhancement is `implementable`
```
kikisdeliveryservice, line 77: @dhellmann feels like we want to change the above line a bit to reflect what implementable is to us - any suggestions? maybe something like: 

`have agreement from all stakeholders and enhancement will be implemented in the near future`

guidelines/enhancement_template.md
```diff
@@ -74,6 +75,7 @@ around the enhancement process.
 ## Release Signoff Checklist
```
dhellmann, line 75: As part of this change, we can also remove this entire section. We should make sure the items listed here are represented in the rest of the template, but we don't want people to feel obligated to come back and update the checklist and we won't merge enhancements until all of the necessary parts are filled in.

guidelines/enhancement_template.md
kikisdeliveryservice: tracking link wfm

kikisdeliveryservice: let me play around with placement locally for a bit to see.

guidelines/enhancement_template.md
kikisdeliveryservice: that makes sense to me, the checklist always ends up falling behind.

guidelines/enhancement_template.md
dhellmann: If we're going to remove this section (see my earlier comment) we would want to say that somewhere else. Maybe in the instructions around line 26 or 28?

kikisdeliveryservice: make sense to me!

guidelines/enhancement_template.md
kikisdeliveryservice: I think the User Docs is the one missing, I see 2 places we can add it template:
1)  between test plan and graduation criteria as its own section to highlight or within graduation criteria
2) as a question in Graduation Criteria

Something like: Does this enhancement require an update to user-facing documentation in openshift-docs?

dhellmann: Let's start with a question in the graduation criteria for now. We're going to review that section in a later update and we can decide then if we want to leave the question there or add a new section to talk about doc updates.

guidelines/enhancement_template.md
```diff
@@ -23,6 +24,9 @@ superseded-by:
 
 Start by filling out this header template with metadata for this enhancement.
 
+* Enhancements should have agreement from all stakeholders prior to merging and
+  target implementation in the near future.
```
dhellmann, line 28: This reads a little choppy. What do you think about splitting it into 2 items like this:

```suggestion
* Enhancements should be related to work to be implemented in the near future.

* Enhancements should have agreement from all stakeholders prior to merging.
```

kikisdeliveryservice: Good call - done!

dhellmann:
Looks good!

/approve

PR #955, opened 2021-11-12
Enhancement for OAuth2 Authorization Grant Login for oc

arjunrn:
Add enhancement for OAuth2 Authorization Grant login for oc with PKCE. Only contains Summary, Motivation and Implementation sections.

### DIFF
--- a/enhancements/authentication/improved-login-workflow.md
+++ b/enhancements/authentication/improved-login-workflow.md
@@ -1,0 +1,242 @@
+---
+title: oc-authorization-grant-login
+authors:
+  - "@arjunrn"
+reviewers:
+  - "@stlaz"
+  - "@s-urbaniak"
+  - "@soltysh"
+approvers:
+  - "@soltysh"
+  - "@mfojtik"
+creation-date: 2021-11-08
+last-updated: 2021-11-08
+status: provisional
+---
+
+# oc Authorization Grant Login
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement proposes a new workflow for the `oc login` command which requires fewer logins and no copy-pasting of
+sensitive credentials. The new workflow is based on implementing some functionality in
+the [oauth-server](https://github.com/openshift/oauth-server) and [oc](https://github.com/openshift/oc) as specified
+in [RFC-8252](https://datatracker.ietf.org/doc/html/rfc8252).
+
+## Motivation
+
+The current CLI login process involves multiple interactive steps where the user has to enter credentials multiple times
+in the browser. This process can be simplified so that the `oc` tool can be used to login directly without having to
+first login to the cluster console, the OAuth server and finally copying the login command. The `oc` tool can be
+modified to act as a public client to the Openshift OAuth server and received an access token
+through [Authorization Code Grant](https://datatracker.ietf.org/doc/html/rfc6749#section-1.3) flow
+and [PKCE](https://datatracker.ietf.org/doc/html/rfc7636).
+
+### Goals
+
+Provide a single-step login in `oc` for identity providers that don't post HTTP challenges.
+
+### Non-Goals
+
+None
+
+## Proposal
+
+### User Stories
+
+__Story 1__ As a user I want to be able to login to the cluster with `oc` and authenticating/providing credentials at
+most once.
+
+__Story 2__ As a user I want the login without having to copy and paste credentials in plain-text.
+
+### API Extensions
+
+None
+
+### Implementation Details [optional]
+
+#### _oc login_ command
+
+The oc login command will have a new flag `--browser` which indicates to `oc` that it should use the OAuth2
+Authorization Code Grant Flow in a browser to login. Currently, when a user enters `oc login $CLUSTER_API_URL`, `oc`
+attempts to login on the cluster if the cluster has identity provider which
+supports [challenge based authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate).
+The new login flow cannot be a fallback if challenge based authentication is unavailable because this may break existing
+workflows where the user expects `oc` to login with the supplied credentials or for `oc` to prompt for credentials. For
+example, if a user runs `oc login` in a script on a remote headless machine, with the assumption that challenge based
+authentication is available, launching a browser would not be a reasonable fallback. Consequently, a new flag which
+indicates to `oc` to start the browser login is required.
+
+[RFC-8252](https://www.rfc-editor.org/rfc/rfc8252.html) described how native apps(any app which does not run completely
+in a browser) can safely authenticate the user against a know OAuth server. `oc` is such an application and any changes
+will be done as prescribed by the RFC.
+
+#### OAuth server
+
+The internal OAuth server uses [OSIN](https://github.com/openshift/osin) as a dependency to provide OAuth2 Authorized
+Code Grant Flow for a registered client to fetch tokens. It also has support for PKCE(Proof Key for Code Exchange) as
+specified in [RFC-7636](https://datatracker.ietf.org/doc/html/rfc7636). _RFC-7636_ is an enhancement for public clients,
+i.e. clients which cannot hold client credentials to fetch an access token securely. Similarly `oc` is also a public
+client because it cannot be distributed with client credentials to retrieve the access token and PKCE is designed for
+use cases like this one.
+
+The OSIN library will have to be modified so that when the redirect URI is a loopback address of the
+form `http://127.0.0.1:{port}/{path}` or `http://[::1]:{port}/{path}` the redirect URL verification should ignore the
+port value. This is required because one particular port may not always be available on all machines where `oc` is run.
+This change has been prescribed as per [RFC-8253](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3)
+
+#### cluster-authentication-operator
+
+The [cluster-authentication-operator](https://github.com/openshift/cluster-authentication-operator) provisions the
+default [OAuthClient](https://docs.openshift.com/container-platform/4.9/rest_api/oauth_apis/oauthclient-oauth-openshift-io-v1.html)
+resources. It already provisions a client for `oc` but sets the `respondWithChallenges` to _true_ which is unsuitable
+for OAuth Grant Flow. Therefore, another client is required where the `respondWithChallenges` is set to `false` which
+should also be provisioned by the operator.
+
+### Risks and Mitigations
+
+1. There are no changes to existing behavior for users of `oc`. Previous login workflows continue working with the same
+   behavior as in previous versions.
+2. The new login workflow in `oc` will need to listen on a port for the callback from the OAuth server. This port cannot
+   be fixed because the port may be in-use on the user's machine and the user may not have the permissions to resolve
+   this. Instead, the user should be able to specify an alternate port which is free on the system. If the user does not
+   specify a port `oc` can pick any available port at random.
+4. Some OSes under certain configurations might prevent the callback server from listening on the loopback interface. In
+   this case the server would not start and `oc` would print a message that it cannot listen on that interface.
+5. If `127.0.0.1` on the user's machine is not the machine itself but has been compromised and points to another
+   location then the attacker can obtain the security code. But this code cannot be exchanged for a token because the
+   attacker does not have access to the challenge key which is only in memory on the users machine.
+
+## Design Details
+
+### Open Questions
+
+### Test Plan
+
+The modifications in _osin_ will have corresponding unit tests. No end-to-end test will be added here.
+
+Because this login workflow relies on a browser being launched and some user interaction in the browser it is not
+trivial to implement an end-to-end test which tests the entire workflow. However, the existing unit tests for the `oc
+login` command can be extended to verify the following scenarios:
+
+1. The browser is launched with the correct path and parameters.
+2. If the browser cannot be launched then a suitable error message is displayed.
+3. When the callback is received oc processes the callback and shuts down the server listening for callback even in
+   error cases.
+
+### Graduation Criteria
+
+**Note:** *Section not required until targeted at a release.*
+
+Define graduation milestones.
+
+These may be defined in terms of API maturity, or as something else. Initial proposal should keep this high-level with a
+focus on what signals will be looked at to determine graduation.
+
+Consider the following in developing the graduation criteria for this enhancement:
+
+- Maturity levels
+  - [`alpha`, `beta`, `stable` in upstream Kubernetes][maturity-levels]
+  - `Dev Preview`, `Tech Preview`, `GA` in OpenShift
+- [Deprecation policy][deprecation-policy]
+
+Clearly define what graduation means by either linking to
+the [API doc definition](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning), or by redefining
+what graduation means.
+
+In general, we try to use the same stages (alpha, beta, GA), regardless how the functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+**Examples**: These are generalized examples to consider, in addition to the
+aforementioned [maturity levels][maturity-levels].
+
+#### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+- Enumerate service level indicators (SLIs), expose SLIs as metrics
+- Write symptoms-based alerts for the component(s)
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+
+**For non-optional features moving to GA, the graduation criteria must include end to end tests.**
+
+#### Removing a deprecated feature
+
+NA
+
+### Upgrade / Downgrade Strategy
+
+NA
+
+### Version Skew Strategy
+
+#### Compatibility of new oc login flow with previous versions of OCP
+
+When the new login flow is used `oc` will open a new browser window which will make a request on the `/authorize`
+endpoint. The client-id needs to be passed here, and it will be the name of the new `OAuthClient` resource created by
+the `cluster-authentication-operator`.
+
+_Newer oc with older API server_: In older versions of OCP where the `cluster-authentication-operator` has not created
+this new client the request to the OAuth server will fail because the client ID is invalid. The redirect to the endpoint
+on loopback will not happen and an error message will be displayed in the browser window.
+
+_Older oc with newer API server_: The new client does not change existing behavior and older `oc` clients will continue
+to work as they have previously.
+
+### Operational Aspects of API Extensions
+
+None
+
+#### Failure Modes
+
+1. `cluster-authentication-operator` If the operator cannot provision the new `OAuthClient` it will transition to
+   the `Degraded` state. Cluster administrators or the API team will then have to determine why this occurred. However,
+   the operator currently provisions 2 other clients in the same fashion and a failure here due to misconfiguration
+   would already be visible.
+2. `oc` If the new login mechanism does not work users will still be able to login with the existing workflow through
+   the console.
+
+#### Support Procedures
+
+NA
+
+## Implementation History
+
+Major milestones in the life cycle of a proposal should be tracked in `Implementation History`.
+
+## Drawbacks
+
+TBD
+
+## Alternatives
+
+None
+
+## Infrastructure Needed [optional]
+
+Use this section if you need things from the project. Examples include a new subproject, repos requested, github
+details, and/or testing infrastructure.
+
+Listing these here allows the community to get the process for these resources started right away.


enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+Grant Flow for a registered client to fetch tokens. It also has support for PKCE(Proof Key for Code Exchange) as
+specified in [rfc7636](https://datatracker.ietf.org/doc/html/rfc7636). _rfc7636_ is an enhancement for public clients,
+i.e. clients which cannot hold client credentials to fetch an access token securely. The new `oc login` is also a public
+client because it cannot contain client credentials to retrieve the access token and PKCE is designed for use cases like
+this one.
+
+Another change which may be required is to allow the port on the callback URL to be variable when the callback is to
+localhost. This is required because the webserver might not always be able to use the same port to listen on for the
+callback webserver.
+
```
lmzuccarelli, line 98: Do we need to consider token expiry ? -  I would imagine that the token is stored in memory within the "oc client" and not persisted.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+2. Login to console with the configured identity provider.
+3. Click on the **"Copy Login Command"** button which redirects the page to the OAuth application.
+4. Perform login again like in step 2.
+5. Copy the displayed login command.
+6. Run the copied command to configure the client to configure the client and access the cluster.
+
+This process can be simplified to the following steps:
+
+1. User runs `oc login --auto $CLUSTER_API_URL`
+2. `oc` starts a webserver to receive a callback from the OAuth server. It then opens the authorization endpoint with a
```
lmzuccarelli, line 51: I would imagine that the webserver will have minimum impact on resource usage , do we need to profile it i.e memory and cpu foot print or is it not necessary ?

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
+---
+title: oc-authorization-grant-login
+authors:
+  - "@arjunrn"
+reviewers:
+  - "@stlaz"
+  - "@s-urbaniak"
+approvers:
+  - TBD
```
tjungblu, line 9: that's probably @mfojtik

enhancements/authentication/improved-login-workflow.md
tjungblu: the embedded go server is pretty lightweight, I think it would be negligible and not noticeable. Especially when you open a multi-gig browser like chrome.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+Currently, when logging in with the `oc` client a user has to perform the following steps:
+
+1. Open the console in the browser.
+2. Login to console with the configured identity provider.
+3. Click on the **"Copy Login Command"** button which redirects the page to the OAuth application.
+4. Perform login again like in step 2.
+5. Copy the displayed login command.
+6. Run the copied command to configure the client to configure the client and access the cluster.
+
+This process can be simplified to the following steps:
```
tjungblu, line 48: nit: you're describing a solution already in the motivation. I reckon you can just say that this is unnecessarily cumbersome and can be improved using the auth grant flow.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+
+1. User runs `oc login --auto $CLUSTER_API_URL`
+2. `oc` starts a webserver to receive a callback from the OAuth server. It then opens the authorization endpoint with a
+   PKCE code challenge.
+3. The OAuth application performs the authentication either directly or by redirecting the user.
+4. Once authentication is successful the OAuth application redirects the user back to the callback URL of `oc` along
```
tjungblu, line 54: does the oauth server allow us to redirect back to localhost?

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+
+### User Stories
+
+TBD
+
+### Implementation Details [optional]
+
+#### _oc login_ command
+
+The oc login command will have new 2 new flags, `--auto` and `--oauth-url`. The first flag indicates to `oc` that it
```
tjungblu, line 80: why call it "auto"?

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+should use the Oauth2 Authorization Flow to login. The second flag is optional and should be specified when the domain
```
tjungblu, line 81: I also remember we had this discussion about multi-IDPs, would the oauth server just present you all the different options to login? OC should stay agnostic of the actual IDP used.

enhancements/authentication/improved-login-workflow.md
tjungblu: the token / code that comes back from the callback has a very short expiry usually, that's something we should clarify on how long.

As for the actual access token: doesn't it already have an expiration? This is controlled by the oauth server normally.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+#### OAuth server
+
+The internal OAuth server uses [OSIN](https://github.com/openshift/osin) as a dependency to provide OAuth2 Authorized
+Grant Flow for a registered client to fetch tokens. It also has support for PKCE(Proof Key for Code Exchange) as
+specified in [rfc7636](https://datatracker.ietf.org/doc/html/rfc7636). _rfc7636_ is an enhancement for public clients,
+i.e. clients which cannot hold client credentials to fetch an access token securely. The new `oc login` is also a public
+client because it cannot contain client credentials to retrieve the access token and PKCE is designed for use cases like
+this one.
+
+Another change which may be required is to allow the port on the callback URL to be variable when the callback is to
```
tjungblu, line 95: I'd pick any free port, you're supplying the callback as part of the URL including port anyway. The question (similar to the previous one above) is whether we need to explicitly whitelist the localhost domain - or in addition also have the port.

tjungblu: azure has port 50342 btw: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/profile/__init__.py#L56

tjungblu: they also don't really make use of that port funnily enough since they switched to msal: https://github.com/Azure/azure-cli/blob/c6afbd09efe35cebd72beeb47ba6d1893d28434e/src/azure-cli-core/azure/cli/core/auth/identity.py#L119-L123

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+
+### Implementation Details [optional]
+
+#### _oc login_ command
+
+The oc login command will have new 2 new flags, `--auto` and `--oauth-url`. The first flag indicates to `oc` that it
+should use the Oauth2 Authorization Flow to login. The second flag is optional and should be specified when the domain
+of the internal OAuth server has
+been [customized](https://docs.openshift.com/container-platform/4.9/authentication/configuring-internal-oauth.html#customizing-the-oauth-server-url_configuring-internal-oauth)
+. It will be the URL of the modified OAuth server.
```
tjungblu, line 84: also one thing is that we should whether a browser is available - imagine if you're on a headless server in which case we should tell the user it's not possible to login that way.

here's how you can do it in python:
https://github.com/Azure/azure-cli/blob/c6afbd09efe35cebd72beeb47ba6d1893d28434e/src/azure-cli-core/azure/cli/core/util.py#L693

they have some funny hacks around detecting wsl and such, probably also useful for us to look into...

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+Another change which may be required is to allow the port on the callback URL to be variable when the callback is to
+localhost. This is required because the webserver might not always be able to use the same port to listen on for the
+callback webserver.
+
+#### cluster-authentication-operator
+
+The [cluster-authentication-operator](https://github.com/openshift/cluster-authentication-operator) provisions the
+default [OAuthClient](https://docs.openshift.com/container-platform/4.9/rest_api/oauth_apis/oauthclient-oauth-openshift-io-v1.html)
+resources. It should also provision the `OAuthClient` for the `oc` command.
+
```
tjungblu, line 104: another alternative approach I've seen (DANGER DO NOT USE) is to supply username and password on the commandline and pass it along to the IDP. Maybe we should note that we definitely do not go that route for reasons.

tjungblu (commented):
looks great already, thanks for writing it up

enhancements/authentication/improved-login-workflow.md
tjungblu: which also brings up the question on how to even open a browser, there seems to be one popular framework: https://github.com/pkg/browser

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+4. Perform login again like in step 2.
+5. Copy the displayed login command.
+6. Run the copied command to configure the client to configure the client and access the cluster.
+
+This process can be simplified to the following steps:
+
+1. User runs `oc login --auto $CLUSTER_API_URL`
+2. `oc` starts a webserver to receive a callback from the OAuth server. It then opens the authorization endpoint with a
+   PKCE code challenge.
+3. The OAuth application performs the authentication either directly or by redirecting the user.
```
stlaz, line 53: which OAuth application?

enhancements/authentication/improved-login-workflow.md
stlaz: and @soltysh

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+1. Open the console in the browser.
+2. Login to console with the configured identity provider.
+3. Click on the **"Copy Login Command"** button which redirects the page to the OAuth application.
+4. Perform login again like in step 2.
+5. Copy the displayed login command.
+6. Run the copied command to configure the client to configure the client and access the cluster.
+
+This process can be simplified to the following steps:
+
+1. User runs `oc login --auto $CLUSTER_API_URL`
```
stlaz, line 50: `oc login --code-grant` maybe?

Why would I have to supply $CLUSTER_API_URL?

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+   with a code.
+5. `oc` can then use this code along with the code challenge to request an access token. On receiving the token `oc`
+   then performs any configuration required to use the token.
+
+Using the above flow the user has to perform a login only once and the `oc` tool is configured automatically without any
+user input.
+
+### Goals
+
+Provide a one-step login for users which can be specified by the cluster operator.
```
stlaz, line 64: "single step login for identity providers that don't post HTTP challenges"

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+TBD
+
+### Implementation Details [optional]
+
+#### _oc login_ command
+
+The oc login command will have new 2 new flags, `--auto` and `--oauth-url`. The first flag indicates to `oc` that it
+should use the Oauth2 Authorization Flow to login. The second flag is optional and should be specified when the domain
+of the internal OAuth server has
+been [customized](https://docs.openshift.com/container-platform/4.9/authentication/configuring-internal-oauth.html#customizing-the-oauth-server-url_configuring-internal-oauth)
```
stlaz, line 83: you don't need the second flag, the data provided by the kube-apiserver get updated

enhancements/authentication/improved-login-workflow.md
stlaz: +1 on the question

enhancements/authentication/improved-login-workflow.md
stlaz: Any free port outside the of the reserved ports range, you don't want to have to run `oc` with root privileges just because of the port choice.

What happens on Windows and Mac OS? I hear Mac OS might attempt to prevent you from opening localhost ports.

enhancements/authentication/improved-login-workflow.md
stlaz: the code currently has 5min expiry by default today, the actual access token expires in a day by default

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+
+Another change which may be required is to allow the port on the callback URL to be variable when the callback is to
+localhost. This is required because the webserver might not always be able to use the same port to listen on for the
+callback webserver.
+
+#### cluster-authentication-operator
+
+The [cluster-authentication-operator](https://github.com/openshift/cluster-authentication-operator) provisions the
+default [OAuthClient](https://docs.openshift.com/container-platform/4.9/rest_api/oauth_apis/oauthclient-oauth-openshift-io-v1.html)
+resources. It should also provision the `OAuthClient` for the `oc` command.
```
stlaz, line 103: IT does provision the `OAuthClient` for the `oc` command:
https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/controllers/oauthclientscontroller/oauthclientscontroller.go#L129

enhancements/authentication/improved-login-workflow.md
stlaz: I don't think I understand, this enhancement should be targeting **exactly** those IdPs for which you can't just pass the username and password on the command line.

stlaz (changes requested):

enhancements/authentication/improved-login-workflow.md
tjungblu: the other day in a different context I saw "--web", maybe that's a little more user friendly. @Anandnatraj wdyt?

enhancements/authentication/improved-login-workflow.md
arjunrn: Yes, the redirect URI is actually supplied by the client. This is verified by the OAuth server to be one of the previously registered callback URIs of the client.

ibihim: OAuth2 supports Native Apps. They have localhost-like URIs.

The callback URIs can be a private-use URI scheme (`com.example.app`), claimed "https" scheme URI (`https://app.example.com/oauth2redirect/example-provider` or loopback redirect (`http://127.0.0.1:51004/oauth2redirect/example-provider`). [RFC 8252](https://datatracker.ietf.org/doc/html/rfc8252#section-7).

The IdP were we are registered must be also happy to store a redirect URI that contains `localhost:port`.

enhancements/authentication/improved-login-workflow.md
arjunrn: I just went with `--auto` as short for automatic. I don't feel strongly about it and this can be changed.

enhancements/authentication/improved-login-workflow.md
arjunrn: > would the oauth server just present you all the different options to login
IIRC that's what the OAuth server does when a user opens the `/authorize` endpoint.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,348 @@
...
+
+This process can be simplified to the following steps:
+
+1. User runs `oc login --auto $CLUSTER_API_URL`
+2. `oc` starts a webserver to receive a callback from the OAuth server. It then opens the authorization endpoint with a
+   PKCE code challenge.
+3. The OAuth application performs the authentication either directly or by redirecting the user.
+4. Once authentication is successful the OAuth application redirects the user back to the callback URL of `oc` along
+   with a code.
+5. `oc` can then use this code along with the code challenge to request an access token. On receiving the token `oc`
```
ibihim, line 56: Did you mean `code verifier` ([RFC 7636#4.5](https://datatracker.ietf.org/doc/html/rfc7636#section-4.5))?

To verify same lingo:
`code challenge == sha256sum(code verifier)`

enhancements/authentication/improved-login-workflow.md
arjunrn: The way I've seen it work is that when there is no browser available the CLI just prints the URL and says open this. And people trying to run this on a server should recognize that the callback wont work. They could copy the URL and do the login on a machine with the browser and when they are done copy the callback URL and call that on the headless machine.

enhancements/authentication/improved-login-workflow.md
ibihim: I would make we wonder if IdP allow dynamic range of redirect ports, do they?

enhancements/authentication/improved-login-workflow.md
ibihim: You can limit the life time of an access token by client type / trust:
- highest: client with credentials
- medium: native apps
- lowest: SPAs

enhancements/authentication/improved-login-workflow.md
tjungblu: > And people trying to run this on a server should recognize that the callback wont work  

IMHO it should fail and just tell you the option isn't available rather than misleading you that opening it in your browser and copying a curl back into your shell. How would that even work? The browser will display an empty page because there is nothing listening on the port and the post body with the credentials is lost.

@Anandnatraj wdyt?

enhancements/authentication/improved-login-workflow.md
ibihim: I would be worried about the spectrum of bugs that could arise when a local server runs on different operating systems, but I assume the audience is pretty tech savvy, when using this.

enhancements/authentication/improved-login-workflow.md
tjungblu: interesting, so all IdPs must be reconfigured for this feature to work?

enhancements/authentication/improved-login-workflow.md
tjungblu: @ibihim what do you mean? it's not like we want to run a fully blown tomcat and the go server has been pretty portable from my experience on windows, linux - x86 and arm at the very least.

enhancements/authentication/improved-login-workflow.md
arjunrn: Callback's a GET with the token as parameter AFAIK.

enhancements/authentication/improved-login-workflow.md
arjunrn: > I would make we wonder if IdP allow dynamic range of redirect ports, do they?
It doesn't currently. However the server can be modified to ignore the ports if the callback is to `localhost` or `127.0.0.1`. I don't think the implications of this would be severe since there's also PKCE. So even if there is a hijack the attacker would not be able to do anything with the temp token.

enhancements/authentication/improved-login-workflow.md
arjunrn: No, the IDPs don't have to be reconfigured. The `OAuthClient` resource has the [redirectURIs](https://docs.openshift.com/container-platform/4.9/rest_api/oauth_apis/oauthclient-oauth-openshift-io-v1.html#oauthclient-oauth-openshift-io-v1) which needs to be set to to the localhost URL.

enhancements/authentication/improved-login-workflow.md
arjunrn: When opening the authorization URL the client(oc) would need the domain and the path. When the path and the domain are not modified it can derive that from the base domain. If it is it needs the extra parameter. Also this is similar to how `oc login --token $TOKEN $CLUSTER_API_URL` works right now.

enhancements/authentication/improved-login-workflow.md
arjunrn: Could you elaborate how? Because when the client is not logged in it cannot query and get the customized URL of the OAuth server.

enhancements/authentication/improved-login-workflow.md
tjungblu: I only vaguely remember my time working with keycloak on things like that, but just checking their docs it is a redirect with parameters. So the copy&paste would probably work if people know what they're doing.

enhancements/authentication/improved-login-workflow.md
arjunrn: But this one has `respondWithChallenges` set to `true`. I believe we need it to be set to false when using the grant flow.

enhancements/authentication/improved-login-workflow.md
arjunrn: Yep, that's what I meant. I've dropped these lines in the enhancement for simplicity.

enhancements/authentication/improved-login-workflow.md
stlaz: > Do you know anyone that has a mac around here to attest?

I do. Reach me on #forum-apiserver on Slack, we may be able to figure someone out.

enhancements/authentication/improved-login-workflow.md
stlaz: I see, so this is just an optional argument, please have it in `[]`

enhancements/authentication/improved-login-workflow.md
stlaz: Yes it can, that's how the kube-apiserver's `/.well-known/oauth-authorization-server` works.

enhancements/authentication/improved-login-workflow.md
stlaz: I see what you mean. Try to reword the sentence so that it involves these specifics.

enhancements/authentication/improved-login-workflow.md
arjunrn: So just to clarify the `$CLUSTER_API_URL` is required to do the first login. On subsequent logins the URL can be retrieved from the existing kubeconfig. Correct?

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,341 @@
...
+
+#### cluster-authentication-operator
+
+The [cluster-authentication-operator](https://github.com/openshift/cluster-authentication-operator) provisions the
+default [OAuthClient](https://docs.openshift.com/container-platform/4.9/rest_api/oauth_apis/oauthclient-oauth-openshift-io-v1.html)
+resources. It already provisions a client for `oc` but sets the `respondWithChallenges` to _true_ which is unsuitable
+for OAuth Grant Flow. Therefore, another client is required where the `respondWithChallenges` is set to `false` which
+should also be provisioned by the operator.
+
+### Risks and Mitigations
```
soltysh, line 96: What are the risks, especially when we go the default in oc vs an opt-in.

soltysh: MacOS and Windows limitations?

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,341 @@
...
+- If an `N->N+1` upgrade fails mid-way through, or if the `N+1` cluster is
+  misbehaving, it should be possible for the user to rollback to `N`. It is
+  acceptable to require some documented manual steps in order to fully restore
+  the downgraded cluster to its previous state. Examples of acceptable steps
+  include:
+  - Deleting any CVO-managed resources added by the new version. The
+    CVO does not currently delete resources that no longer exist in
+    the target version.
+
+### Version Skew Strategy
```
soltysh, line 221: This is a must for oc since we support +/- 1 version compatibility

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,341 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The current CLI login process involves multiple interactive steps where the user has to enter credentials multiple
```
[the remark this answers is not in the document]
arjunrn: `oc` may not require it but the user has to perform login at least twice to get the token after which they can login/configure `oc`. Yes `oc` does support login with username/password but only if a suitable identity provider is present. In most cases the identity provider may not support such a login. I will update the Summary to highlight this.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Motivation
+
+Currently, when logging in with the `oc` client a user has to perform the following steps:
+
+1. Open the console in the browser.
+2. Login to console with the configured identity provider.
+3. Click on the **"Copy Login Command"** button which redirects the page to the OAuth application.
+4. Perform login again like in step 2.
+5. Copy the displayed login command.
```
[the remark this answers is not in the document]
arjunrn: These step in themselves are not problems, just steps. The proposal aims to provide a way to side-step them for a faster login process.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,341 @@
...
+
+### API Extensions
+
+None
+
+### Implementation Details [optional]
+
+#### _oc login_ command
+
+The oc login command will have a new flag `--auto` which indicates to `oc` that it should use the Oauth2 Authorization
```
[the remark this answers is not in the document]
arjunrn: IIUC making this the default would change existing behavior. The existing behavior is to call the authorization endpoint and set the redirect URI to the implicit token endpoint and if there is an identity provider which supports challenge based authentication it takes the username and password passed along with the `login` command and performs authorization. @stlaz please correct me if I'm wrong.
That said it might be possible for the server to indicate how to proceed and then use that to determine behavior. Let me investigate further.

enhancements/authentication/improved-login-workflow.md
arjunrn: I've tested this on Windows and when started from the command line in Windows an application can listen even on ports like 8080.

enhancements/authentication/improved-login-workflow.md
arjunrn: No limitations. Tested by cross compiling a simple server which listens on port 8080 and running it on Windows and MacOS.

enhancements/authentication/improved-login-workflow.md
arjunrn: Added some details here.

enhancements/authentication/improved-login-workflow.md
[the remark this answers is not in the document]
soltysh: That's why I asked for alternatives and their pros & cons, so that we know in the future what was considered and why we did what we did :)

enhancements/authentication/improved-login-workflow.md
soltysh: Say so in the enhancement, don't leave empty sections for second guessing.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,319 @@
...
+
+### Risks and Mitigations
+
+## Design Details
+
+### Open Questions [optional]
+
+This is where to call out areas of the design that require closure before deciding to implement the design. For
+instance,
+> 1. This requires exposing previously private resources which contain sensitive information. Can we do this?
```
soltysh, line 104: That looks like a security risk to me.

enhancements/authentication/improved-login-workflow.md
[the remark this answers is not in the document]
arjunrn: I"ve explained why the new flow cannot be the default behavior and why a new flag is required.

enhancements/authentication/improved-login-workflow.md
arjunrn: I've added how callback to localhost can be a risk and how it's mitigated with PKCE.

enhancements/authentication/improved-login-workflow.md
arjunrn: Added details about how if an OS prevents `oc` from opening a listener on a certain port an error message will be displayed.

Anandnatraj:
Not able to figure out how do I pass Identity Provider flags and values through`oc login` command ?

arjunrn:
> Not able to figure out how do I pass Identity Provider flags and values throughoc login command ?
@Anandnatraj I looked into this and there's no straight-forward way to specify the identity provider when calling the Openshift oauth-server. Since the oauth-server manages the identity providers internally and only exposes endpoints with parameters which is specified in the [authorization endpoint spec](https://connect2id.com/products/server/docs/api/authorization). It then displays a list of identity providers to the user.

So after this enhancement is implemented the user will be shown the list of configured IDPs and they can select one and proceed. But maybe there's a way to specify that with the `oc login` command which @stlaz know.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,255 @@
...
+### API Extensions
+
+None
+
+### Implementation Details [optional]
+
+#### _oc login_ command
+
+The oc login command will have a new flag `--auto` which indicates to `oc` that it should use the Oauth2 Authorization
+Flow to login. Currently, when a user enters `oc login $CLUSTER_API_URL`, `oc` attempts to be login on the cluster if
```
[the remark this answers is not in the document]
stlaz: I believe it's `oc login --server $CLUSTER_API_URL`

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,255 @@
...
+the cluster has identity provider which
+support [challenge based authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate).
+The new login flow cannot be a fallback if challenge based authentication is unavailable because the user might
+explicitly want to use it to login through code flow with PKCE. For this reason a new flag which indicates to `oc` to
+start the code flow is required.
+
+#### OAuth server
+
+The internal OAuth server uses [OSIN](https://github.com/openshift/osin) as a dependency to provide OAuth2 Authorized
```
stlaz, line 86: does "Authorized Grant Flow" mean "Authorization Code Grant" flow here or is it just "OAuth2-authorized flows"?

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,255 @@
...
+   behavior as in previous versions.
+2. The new login workflow will need to listen on a port for the callback from the OAuth server. This port cannot be
+   fixed because the port may be in-use on the user's machine and the user may not have the permissions to resolve this.
+   Instead, the user should be able to specify an alternate port which is free on the system. In the new client we will
+   allow redirects on the following URLs:
+   * `http://localhost:8280/callback`
+   * `http://localhost:8281/callback`
+   * `http://localhost:8282/callback`
+   * `http://localhost:8283/callback`
+   * `http://localhost:8284/callback`
```
stlaz, line 117: As per https://www.rfc-editor.org/rfc/rfc8252.html#section-7.3, you should not use localhost but IPv{4,6} addresses of loopback devices.

stlaz, line 117: Also, the loopback device ports can vary, you should not use a closed set. See https://www.rfc-editor.org/rfc/rfc8252.html#section-7.3 for how the redirect URI validation should work.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,255 @@
...
+Consider the following in developing a test plan for this enhancement:
+
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+- What additional testing is necessary to support managed OpenShift service-based offerings?
+
+No need to outline all of the test cases, just the general strategy. Anything that would count as tricky in the
+implementation and anything particularly challenging to test should be called out.
+
+All code is expected to have adequate tests (eventually with coverage expectations).
```
stlaz, line 147: I expect this section to be filled.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,255 @@
...
+
+#### Compatibility of new oc login flow with previous versions of OCP
+
+When the new login flow is used `oc` will open a new browser window which will make a request on the `/authorize`
+endpoint. The client-id needs to be passed here, and it will be the name of the new `OAuthClient` resource created by
+the `cluster-authentication-operator`.
+
+_Newer oc with older API server_: In older versions of OCP where the `cluster-authentication-operator` has not created
+this new client the request to the OAuth server will fail because the client id is invalid. The redirect to the endpoint
+on localhost will not happen and an error message will be displayed in the browser window.
```
stlaz, line 216: Could we detect what we're allowed to do by using something like https://datatracker.ietf.org/doc/html/rfc8414, which is somewhat built in to the kube-apiserver even today?

stlaz, line 216: Could we detect what we're allowed to do by using something like https://datatracker.ietf.org/doc/html/rfc8414, which is somewhat built in to the kube-apiserver even today?

stlaz (changes requested):
please read https://www.rfc-editor.org/rfc/rfc8252.html and apply it to this enhancement

enhancements/authentication/improved-login-workflow.md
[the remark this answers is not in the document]
arjunrn: The URL is not passed as a flag.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,255 @@
...
+#### OAuth server
+
+The internal OAuth server uses [OSIN](https://github.com/openshift/osin) as a dependency to provide OAuth2 Authorized
+Grant Flow for a registered client to fetch tokens. It also has support for PKCE(Proof Key for Code Exchange) as
+specified in [rfc7636](https://datatracker.ietf.org/doc/html/rfc7636). _rfc7636_ is an enhancement for public clients,
+i.e. clients which cannot hold client credentials to fetch an access token securely. Similarly `oc` is also a public
+client because it cannot be distributed with client credentials to retrieve the access token and PKCE is designed for
+use cases like this one.
+
+Another change which may be required is to allow the port on the callback URL to be variable when the callback is to
```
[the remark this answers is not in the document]
arjunrn: The first change is the new flag for `oc login`

enhancements/authentication/improved-login-workflow.md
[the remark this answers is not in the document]
arjunrn: There already is default behavior in `oc` even when no credentials are provided I don't think it's possible to have the new login flow as default in any scenario.

enhancements/authentication/improved-login-workflow.md
arjunrn: The metadata endpoint would not expose which clients are registered and what callback URLs they have, because this would be confidential. So it's not possible to determine if the the new worklflow can be used from just the metadata itself.

enhancements/authentication/improved-login-workflow.md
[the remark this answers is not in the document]
arjunrn: Ok, I see how this is misleading. Will fix it.

enhancements/authentication/improved-login-workflow.md
arjunrn: I've added some details about the tests.

enhancements/authentication/improved-login-workflow.md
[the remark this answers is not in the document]
stlaz: But it is.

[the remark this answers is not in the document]
stlaz: Never mind, I see that now, we have two very similar use-cases, one of them using the flag, the other one not :man_facepalming:

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,245 @@
...
+5. If `127.0.0.1` on the user's machine is not the machine itself but has been compromised and points to another
+   location then the attacker can obtain the security code. But this code cannot be exchanged for a token because the
+   attacker does not have access to the challenge key which is only in memory on the users machine.
+
+## Design Details
+
+### Open Questions
+
+1. ~~Should we modify the OAuth server library that it allows any port for the redirect URL when the url is localhost?~~
+   Yes to be in line with RFC-8252.
```
stlaz, line 121: it's not an open question, then :slightly_smiling_face:

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,245 @@
...
+2. ~~Does receiving a callback on `127.0.0.1` pose a security risk? Or in other words what happens when `127.0.0.1` on
+   the users machines actually points to a location other than the machine itself.~~ As per RFC-8252,
```
stlaz, line 123: this does not seem to be an open question, either :slightly_smiling_face:

stlaz (commented):
Almost there I think! :+1:

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,245 @@
...
+Code Grant Flow in a browser to login. Currently, when a user enters `oc login $CLUSTER_API_URL`, `oc` attempts to be
+login on the cluster if the cluster has identity provider which
+supports [challenge based authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate).
+The new login flow cannot be a fallback if challenge based authentication is unavailable because this may break existing
+workflows where the user expects _oc_ to login with the supplied credentials or for `oc` to prompt for credentials. For
+this reason a new flag which indicates to `oc` to start the code flow is required.
+
+[RFC-8252](https://www.rfc-editor.org/rfc/rfc8252.html#section-7.3) described how native apps or any app which is not a
+browser can safely authenticate the user against a know OAuth server. `oc` is such an application and any changes will
+be done as prescribed by the RFC.
```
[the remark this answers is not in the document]
arjunrn: The definition of the "native app" from the RFC:
```
"native app"  An app or application that is installed by the user to
      their device, as distinct from a web app that runs in the browser
      context only.  Apps implemented using web-based technology but
      distributed as a native app, so-called "hybrid apps", are
      considered equivalent to native apps for the purpose of this
      specification.
```
This is what I wanted to convey. I should probably rephrase it.

enhancements/authentication/improved-login-workflow.md
arjunrn: Actually the double tildes at the beginning is a strike-through. It's not visible in preview. I will remove them but I wanted to keep them in some form in the enhancement.

enhancements/authentication/improved-login-workflow.md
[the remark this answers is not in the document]
arjunrn: > It'd be nice to be able to default to this in case there are no HTTP challenges/redirects from the oauth-server, instead of printing a message for the users to go to login to the oauth-server

What about the inverse case. When there are identity providers which support challenges but the user wants to log in with the browser because their identity is present in an identity provider which doesn't support challenges. This would would mean a flag is required. So if we support the fallback case and the flag case that means there are 2 ways launch a browser for login. And one if them is conditional. I'm not sure that is the best user experience.

enhancements/authentication/improved-login-workflow.md
stlaz: Ah, true, I ignored the tildes :slightly_smiling_face:

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Implementation Details [optional]
+
+#### _oc login_ command
+
+The oc login command will have a new flag `--auto` which indicates to `oc` that it should use the Oauth2 Authorization
+Code Grant Flow in a browser to login. Currently, when a user enters `oc login $CLUSTER_API_URL`, `oc` attempts to be
+login on the cluster if the cluster has identity provider which
+supports [challenge based authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate).
+The new login flow cannot be a fallback if challenge based authentication is unavailable because this may break existing
```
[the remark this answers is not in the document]
stlaz: I see, can you work that in into to the document as an example to the possibly broken existing workflows?

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,240 @@
...
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement proposes a new workflow for the `oc login` command which requires fewer logins and no copy-pasting of
+sensitive credentials. The new workflow is based on implementing some functionality in
+the [oauth-apiserver](https://github.com/openshift/oauth-apiserver) and [oc](https://github.com/openshift/oc) as
```
stlaz, line 32: did you mean "oauth-server"?

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,242 @@
...
+
+None
+
+### Implementation Details [optional]
+
+#### _oc login_ command
+
+The oc login command will have a new flag `--browser` which indicates to `oc` that it should use the Oauth2
+Authorization Code Grant Flow in a browser to login. Currently, when a user enters `oc login $CLUSTER_API_URL`, `oc`
+attempts to be login on the cluster if the cluster has identity provider which
```
soltysh, line 71:
```suggestion
attempts to login on the cluster if the cluster has identity provider which
```

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,242 @@
...
+supports [challenge based authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate).
+The new login flow cannot be a fallback if challenge based authentication is unavailable because this may break existing
+workflows where the user expects _oc_ to login with the supplied credentials or for `oc` to prompt for credentials. For
```
soltysh, line 74: Nit: use `oc` or _oc_ consistently

enhancements/authentication/improved-login-workflow.md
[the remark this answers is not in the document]
soltysh: I'm not sure if `--browser` is the most suitable name, but I can't think of anything better yet, so let's go with that for now. If I come up with something better, we can always change that during implementation.

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,242 @@
...
+
+### Open Questions
+
+### Test Plan
+
+The modifications in _osin_ will have corresponding unit tests. No end-to-end test will be added here.
+
+Because this login workflow relies on a browser being launched and some user interaction in the browser it is not
+trivial to implement an end-to-end test which tests the entire workflow. However, the existing unit tests for the _oc
+login_ command can be extended to verify the following scenarios:
```
soltysh, line 129: Make sure QA adds a manual test case to ensure this flow always works.

soltysh (approved):
/lgtm
/approve
/hold
for any remaining reviews

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,242 @@
...
+The current CLI login process involves multiple interactive steps where the user has to enter credentials multiple times
+in the browser. This process can be simplified so that the `oc` tool can be used to login directly without having to
+first login to the cluster console, the OAuth server and finally copying the login command. The `oc` tool can be
+modified to act as a public client to the Openshift OAuth server and received an access token
+through [Authorization Code Grant](https://datatracker.ietf.org/doc/html/rfc6749#section-1.3) flow
+and [PKCE](https://datatracker.ietf.org/doc/html/rfc7636).
+
+### Goals
+
+Provide a single-step login in `oc` for identity providers that don't post HTTP challenges.
```
ibihim, line 46: For me the phrasing "single-step" confused me completely. As far as I can tell there are more steps for OAuth2 for Native Apps:

- Native App triggers flow (oc login)
- Browser opens
- if not logged in, user needs to enter credentials
- User needs to approve the resources to be shared
- redirect

enhancements/authentication/improved-login-workflow.md
```diff
@@ -0,0 +1,242 @@
...
+
+### API Extensions
+
+None
+
+### Implementation Details [optional]
+
+#### _oc login_ command
+
+The oc login command will have a new flag `--browser` which indicates to `oc` that it should use the Oauth2
```
ibihim, line 69: `s/Oauth2/OAuth2/`

enhancements/authentication/improved-login-workflow.md
arjunrn: What I meant to say in there was - provide a way for users to login such that they only have to authenticate once. Currently users have to open the console, then click on the login command button which opens the token page where they have to login all over again before they copy the command and paste it into the console. The new approach instead means that if they user wants to login again to a cluster they have previous logged into again all they have to do is run `oc login --browser` and authenticate if required. The `oc` client will then be reconfigured automatically with the new token. 

If you can think of a good way of describing this in the goal I'm don't mind changing it. I understand it's not the best description.

PR #958, opened 2021-11-17
enhancements/monitoring: User-defined alerting rules proposal

bison:
This adds an enhancment proposal for supporting user-defined alerting
rules and limited alert overrides in the OpenShift monitoring stack.

### DIFF
--- a/enhancements/monitoring/alert-overrides.md
+++ b/enhancements/monitoring/alert-overrides.md
@@ -1,0 +1,338 @@
+---
+title: alert-overrides
+authors:
+  - "@bison"
+reviewers:
+  - TBD
+  - "@openshift/openshift-team-monitoring"
+approvers:
+  - TBD
+api-approvers:
+  - TBD
+creation-date: 2021-11-02
+last-updated: 2021-11-02
+tracking-link: https://issues.redhat.com/browse/MON-1985
+status: implementable
+---
+
+# User-Defined Overrides for Platform Alerting Rules
+
+## Release Signoff Checklist
+
+- [X] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The core OpenShift components ship a large number of Prometheus alerting rules.
+A 4.10-nightly cluster on AWS currently has 175 alerts defined by default, and
+enabling additional add-ons increases that number.  Still, users have repeatedly
+asked for a supported method of adding additional custom rules, and adjusting
+thresholds, or adding labels and annotations to existing platform alerting
+rules.
+
+These modifications are currently not possible, because the alerts are managed
+by operators and cannot be modified, and the addition of custom rules is
+explicitly unsupported at the moment.  This proposal outlines a solution for
+allowing additional user-defined rules and limited overrides of existing
+platform alerting rules.
+
+## Motivation
+
+OpenShift is an opinionated platform, and the default alerting rules have of
+course been carefully crafted to match what we think the majority of users need.
+Nevertheless, users have repeatedly asked for the ability to modify alerting
+rules to adjust thresholds, or to simply add labels and annotations to aide in
+routing alert notifications, as well as a supported method of adding additional
+alerting rules.
+
+Users currently have the ability to add additional rules from a technical
+standpoint -- by simply creating new `PrometheusRule` objects -- but this is
+[explicitly unsupported][1] at the moment.  The reason being that these
+resources are handled directly by the prometheus-operator, making it difficult
+to perform OpenShift specific defaulting and validation.  A concrete example
+being that there is no way to prevent users from creating new recording rules
+that may overwhelm the platform Prometheus instance.
+
+We've also encountered bugs and unexpected interactions that cause alerts to
+fire erroneously.  A supported method of patching platform alerting rules
+provides a more flexible method of applying temporary fixes in such cases.
+
+The goal of this proposal is to enable the addition of custom alerting rules,
+and limited ability to override existing platform alerting rules via a resource
+under the direct control of the cluster-monitoring-operator.
+
+### Goals
+
+- Give users a single, supported, and documented place to add additional rules,
+  in a way that allows the cluster-monitoring-operator to provide its own
+  defaulting, validation, etc.
+- Allow users to apply overrides to alerting rules shipped by the platform.
+- Allow users to drop unwanted platform alerts.
+- User-supplied overrides are merged with platform rules -- updates to fields
+  not modified by users continue to get updates.
+- Resources shipped by operators remain unmodified.  No action needed from
+  OpenShift developers.
+
+### Non-Goals
+
+- Only alerting rules are available for modification.  Not recording rules or
+  other monitoring configuration.
+
+## Proposal
+
+The cluster-monitoring-operator will be extended with a new controller that
+watches an instance of a new resource type with the following structure:
+
+```yaml
+apiVersion: v1alpha1
+kind: alertoverrides.monitoring.openshift.io
+metadata:
+  name: alert-overrides
+  namespace: openshift-monitoring
+spec:
+  # List of overrides for alerts in platform namespaces.
+  overrides:
+  - selector: # required
+      alert: "KubeAPIErrorBudgetBurn" # required
+      matchLabels: # optional, but multiple matches is an error reported in status.
+        severity: "critical"
+        long: "1h"
+    action: "patch"
+    labels:
+      patched: "true"
+
+  - selector:
+      alert: "KubeAPIErrorBudgetBurn"
+      matchLabels: # not a unique match, this is an error.
+        severity: "critical"
+    action: "patch"
+    labels:
+      patched: "true"
+
+  - selector:
+      alert: "Watchdog"
+    action: "drop"
+
+  # List of new Prometheus alerting rules.
+  rules:
+  - alert: "CustomAlert"
+    expr: "vector(1)"
+    for: "15m"
+    labels:
+      custom: "true"
+
+status:
+  conditions:
+  - type: "OverrideError"
+    status: "True"
+    reason: "MultipleMatches"
+    message: "Multiple alerts found with name KubeAPIErrorBudgetBurn and labels: [severity: critical]"
+    lastTransitionTime: "2021-11-26T15:42:19Z"
+```
+
+The controller maintains an index of alerting rule names to their corresponding
+`PrometheusRule` objects, ignoring any in non-platform namespaces, i.e. any
+namespace without the `openshift.io/cluster-monitoring=true` label.  When the
+configuration changes, all `PrometheusRule` resources containing an alerting
+rule with the given name are found, the rules extracted, and the static labels
+are matched against the `matchLabels` map supplied by the user.  If an exact
+match is found, this rule is targeted for the override.  Multiple matches is an
+error condition, this override will be skipped, and an error will be reported in
+the status conditions.
+
+For each override a new alerting rule is generated with the overrides applied on
+top of the existing rule, and this new rule is added to the "alert-overrides"
+`PrometheusRule` object.  Additional user-supplied rules are also added here.
+All generated alerts will be given an identifying label -- currently:
+`monitoring_openshift_io__alert_override=true`.  This happens after applying
+user-supplied overrides, meaning the user cannot modify this label.  All rules
+can be validated individually using [the same code][2] the prometheus-operator
+and Prometheus itself use.
+
+This leaves the original alerting rules present and untouched.  These, as well
+as any rules listed with the `drop` action, must not be forwarded to the
+platform Alertmanager.  A `Secret` containing [alert_relabel_configs][3] data
+for Prometheus is generated to accomplish this:
+
+```yaml
+# Alerts to be dropped are matched against name, the set of labels the user
+# supplied to match on, and an empty override label.  The empty override label
+# differentiates the original and generated alerting rules.
+
+- source_labels: [alertname, severity, long, monitoring_openshift_io__alert_override]
+  regex: "KubeAPIErrorBudgetBurn;critical;1h;"
+  action: drop
+- source_labels: [alertname, severity, monitoring_openshift_io__alert_override]
+  regex: “Watchdog;none;”
+  action: drop
+```
+
+Note that this prevents these alerts, whether overridden or dropped, from being
+sent to Alertmanager, but they will still show up as active alerts in the
+Prometheus API.  The OpenShift web console will have to be made aware of the
+labeling in order to hide these by default.
+
+The generation of a new `PrometheusRule` object for overrides, and dropping the
+original alerts via relabeling is a consequence of the fact that the existing
+rules are managed by various operators.  They cannot be modified directly, as
+the operators would revert the changes.  Keeping the existing alerts in place,
+and only preventing them from being sent to Alertmanager via relabeling also
+allows us to continue to collect telemetry on whether these platform alerts are
+firing -- even if a user has chose to completely "drop" them.
+
+Summary:
+
+- Watch for a user-defined list of overrides and new rules.
+  - Processing of overrides is also triggered by `PrometheusRule` object
+    changes, and changes to the set of platform namespaces.
+- For each override, find the existing rule matching the name and labels, and
+  apply the action specified by the user, i.e. `patch` or `drop`.
+  - If the action is `patch`, generate a new rule with supplied overrides.
+  - If the action is `drop`, simply add the alert to the alert relabel config.
+  - If multiple alerts match the name and labels, notify the user and skip
+    this override.
+- Generate new `PrometheusRule` objects for any new user-defined rules.
+- Generate the [alert_relabel_configs][1] to drop original alerts that have been
+  replaced with a patched version, and any that were explicitly dropped.
+
+### User Stories
+
+- As an OpenShift user, I want to change the severity of an existing alert,
+  choosing between, e.g. from `Warning` to `Critical`.
+- As an OpenShift user, I want to permanently disable an alert that I know does
+  not pertain to my deployment.
+- As an OpenShift user, I want to add labels or annotations to alerts to aide in
+  routing and triage.
+- As an OpenShift user, I want to rewrite the PromQL expression or duration for
+  an alert in order to adjust thresholds to suit my environment.
+- As an OpenShift user, I want a supported way to add custom alerts using
+  metrics from components in the `openshift-*` namespaces.
+
+### API Extensions
+
+We will be adding a new CRD for alert overrides. Initially:
+`alertoverrides.monitoring.coreos.com/v1alpha1`
+
+### Implementation Details/Notes/Constraints
+
+TBD
+
+### Risks and Mitigations
+
+The primary risk is that this provides users with a method to permanently
+modify, including to completely drop, alerting rules, which could lead to
+unintentionally missed alerts in the case of user error.  Careful consideration
+should be given to UX.  Information on which overridden or dropped alerts are
+firing is still available in the Prometheus API.  Perhaps this can be leveraged
+and surfaced in an unobtrusive way.
+
+## Design Details
+
+### Open Questions
+
+TBD
+
+### Test Plan
+
+This should be fairly straight forward to both unit and e2e test.
+
+### Graduation Criteria
+
+Plan to release as Tech Preview initially.
+
+#### Dev Preview -> Tech Preview
+
+N/A
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+- User facing documentation created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+**For non-optional features moving to GA, the graduation criteria must include
+end to end tests.**
+
+#### Removing a deprecated feature
+
+- Announce deprecation and support policy of the existing feature
+- Deprecate the feature
+
+### Upgrade / Downgrade Strategy
+
+Upgrade and downgrade are not expected to present a significant challenge.  The
+overridden rules are reconciled against those shipped by OpenShift in real time,
+so any updates are applied as needed.  On downgrade to a version not supporting
+the feature, the newly generated `PrometheusRule` resource and the `Secret`
+containing the relabeling configs can simply be deleted.
+
+The biggest concern is removal, renaming, or change in severity of OpenShift
+shipped rules as this may prevent a previously overridden rule from being found.
+This will be surfaced to the user in the form of status conditions on the CRD.
+
+### Version Skew Strategy
+
+This shouldn't be an issue as rules are reconciled against OpenShift rules and
+regenerated as necessary as changes occur.
+
+### Operational Aspects of API Extensions
+
+The new CRD is not expected to have significant operational impact.  It is in
+fact a "singleton" CRD, in that there is only expected to be a single instance
+of it in the `openshift-monitoring` namespace.
+
+#### Failure Modes
+
+TBD
+
+#### Support Procedures
+
+TBD
+
+## Implementation History
+
+- Initial Proof-of-Concept:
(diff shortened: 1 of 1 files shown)


bison:
cc: @openshift/openshift-team-monitoring @openshift/sre-alert-sme

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,280 @@
...
+- source_labels: [alertname, severity, monitoring_openshift_io__alert_override]
+  regex: “Watchdog;none;”
+  action: drop
+- source_labels: [alertname, severity, monitoring_openshift_io__alert_override]
+  regex: “AlertmanagerReceiversNotConfigured;Warning;”
+  action: drop
+```
+
+Note that this prevents these alerts, whether overridden or dropped, from being
+sent to Alert Manager, but they will still show up as active alerts in the
```
kyoto, line 143:
```suggestion
sent to Alertmanager, but they will still show up as active alerts in the
```

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Alternatives
+
+- Instead of dropping alerts via relabel config, it may be possible to have the
+  operator generate silences for them.
+  - Pros:
+    - No / fewer changes to console.  Overridden alerts show as silenced.
+  - Cons:
+    - Additional complexity in cluster-monitoring-operator.
+    - Additional load on Alert Manager.
```
kyoto, line 262:
```suggestion
    - Additional load on Alertmanager.
```

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,280 @@
...
+kind: ConfigMap
+metadata:
+  name: alert-overrides
+  namespace: openshift-monitoring
+data:
+  overrides.yaml: |
+    overrides: # List of platform rule overrides.
+    - key:
+        name: "Watchdog" # Alerts are identified by name and severity tuple.
+        severity: "none"
```
kyoto, line 98: The `name` & `severity` together are not enough to uniquely identify rules (for example 2 rules in different rule groups with the same name and severity).

In order to uniquely identify rules in the Console UI, we use all of: (group file, group name, rule name, rule duration, rule query, rule labels). I think we will need to do the same or something similar here.

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,280 @@
...
+`PrometheusRule` objects, ignoring any in non-platform namespaces, i.e. any
+namespace without the `openshift.io/cluster-monitoring=true` label.  When the
+configuration changes, the `PrometheusRule` containing the targeted alert is
+found, and the alert extracted.
+
+For each override a new alerting rule is generated with the overrides applied on
+top of the existing rule, and this new rule is added to the "alert-overrides"
+`PrometheusRule` object.  Additional user-supplied rules are also added here.
+All generated alerts will be given an identifying label -- currently:
+`monitoring_openshift_io__alert_override=true`
```
kyoto, line 122: Could you explain the reasons for keeping the overridden rules?

Conceptually it seems simpler to actually remove an overridden rule and replace it with the new rule. That would avoid needing to add a magic label for dropped and patched rules and should reduce load on the system too. What are the benefits that make it worth keeping the old rules?

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,280 @@
...
+
+This leaves the original alerting rules present and untouched.  These, as well
+as any rules listed with the `drop` action, must be ignored by the platform
+Prometheus instance.  A `Secret` containing [alert_relabel_configs][1] data for
+Prometheus is generated to accomplish this:
```
kyoto, line 127: Could you explain the reasons for keeping the overridden rules?

Conceptually it seems simpler to actually remove an overridden rule and replace it with the new rule. That would avoid needing to add a magic label for dropped and patched rules and should reduce load on the system too. What are the benefits that make it worth keeping the old rules?

enhancements/monitoring/alert-overrides.md
bison: Fixed.

enhancements/monitoring/alert-overrides.md
bison: Fixed.

enhancements/monitoring/alert-overrides.md
bison: We want to be able allow limited overrides of existing rules without the operators having to do anything. We can't modify or remove existing alerts shipped by any OpenShift operator, because the operator will just revert the changes. So the only option is to generate a new alert and somehow ignore the original one.

enhancements/monitoring/alert-overrides.md
jan--f: To extend on this a bit: I think the enhancement proposal could benefit from explicitly stating whether overrides must uniquely identify a single alert or not. Overriding or dropping multiple alerts with one spec could certainly be useful.
However this adds significant complexity from the UI perspective. How can a user ensure their override does what they intend.
If we want an explicit one to one matching lets call that out and sketch how a user interaction would look like if they provide an insufficient spec (i.e. one that doesn't uniquely identify one alert).

jan--f:
Nice work!
I'm wondering if it makes sense to call out the upgrade scenario where an upgrade changes an overridden alert more explicitly. What happens if the spec no longer matches an alert or now matches multiple? Even listing this as "can happend, considered out of scope" would be beneficial I think.

enhancements/monitoring/alert-overrides.md
bison: Yeah, this is a point that absolutely needs clarification, thanks.

On a fresh 4.10-nightly cluster on AWS, I only see two alerts with this problem. One is `AlertmanagerClusterFailedToSendAlerts` which is due to cluster-monitoring-operator explicitly bringing the `critical` version of the alert down to `warning` and the other is `KubeAPIErrorBudgetBurn` which is a bit more complicated. That one legitimately seems to have two versions each at the `warning` and `critical` severities. Of course there could be other instances shipped by add-ons.

I suppose this is a trade-off between preventing an override matching multiple rules, and requiring the user to specify a bunch of fields to match against -- ones that often aren't the most visible to users. Given the above, I don't think either of these options are unreasonable:

1) Consider multiple matches of name + severity as an error condition. Can't override that, sorry. We could couple this with amending the style-guide to disallow this, adding e2e tests for it in `origin`, and fixing the offending alerts.

2) Just apply the override to all matching alerts. Though, like you said, that may present UX challenges.

I guess I would lean towards the first option personally.

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,280 @@
...
+## Implementation History
+
+- Initial Proof-of-Concept:
+  https://github.com/openshift/cluster-monitoring-operator/pull/1473
+
+## Drawbacks
+
+TBD
+
+## Alternatives
```
dhellmann, line 254: It would be useful to describe here why a ConfigMap is being used, instead of defining a new CRD for a regular API.

bison: Yeah, to be honest I'm not sure there's a reason other than that's what we prototyped with. It's called out as an open question. If the consensus is to just go with a CRD from the start, we can do that.

enhancements/monitoring/alert-overrides.md
jan--f: Having a clear an concise failure mode definitely beats a murky UI, so +1 for 1. Though it might be valuable to at least mention both approaches in this document?

enhancements/monitoring/alert-overrides.md
kyoto: Thanks. I think it would be good to briefly explain that background in the proposal.

enhancements/monitoring/alert-overrides.md
dofinn: another example here is Kube* rules that are generated with a `namespace=` field. SRE use this currently to determine what should/could be ignored.

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+The controller maintains an index of alerting rule name-and-severity pairs to
+`PrometheusRule` objects, ignoring any in non-platform namespaces, i.e. any
+namespace without the `openshift.io/cluster-monitoring=true` label.  When the
+configuration changes, the `PrometheusRule` containing the targeted alert is
+found, and the alert extracted.
+
+For each override a new alerting rule is generated with the overrides applied on
+top of the existing rule, and this new rule is added to the "alert-overrides"
```
dofinn, line 119: I am having a hard time grok'ing this. Could you please clarify?

- CM exists with identifiers and actions
- Existing PrometheusRules are identified using the identifiers found in the CM
- Actions are applied (Patch,Drop?) to these found PrometheusRules
- Replacement rules are added to a PrometheusRules object called "alert-overrides"
- label is add to each new replacement rule
?

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+Note that this prevents these alerts, whether overridden or dropped, from being
+sent to Alertmanager, but they will still show up as active alerts in the
+Prometheus API.  The OpenShift console will have to be made aware of the
+labeling in order to hide these by default.
+
+### User Stories
+
+- As an OpenShift user, I want to raise the severity of an alert from `Warning`
```
dofinn, line 149:
```suggestion
- As an OpenShift user, I want to manage the severity of an alert choosing between ("Info", "Warning", "Critical")
```

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,280 @@
...
+- As an OpenShift user, I want to add labels or annotations to alerts to aide in
+  routing and triage.
+- As an OpenShift user, I want to rewrite the PromQL expression or duration for
+  an alert in order to adjust thresholds to suit my environment.
+- As an OpenShift user, I want a supported way to add custom alerts using
+  metrics from components in the `openshift-*` namespaces.
+
+### API Extensions
+
+None, but this is an open question.  Is there good reason to make this a CRD
```
dofinn, line 162: +1 to CRD, could build status into suppress/changed alerts so as a user I have a summary of what I have changed?

dofinn (commented):
Nice job here @bison ! this stuff seemingly gets really complicated quickly. Maybe too many options here become un-productive. I know for OSD we simply do:

- Silence an alert
- Create an alert (if the silence needs replacing/customizing)

This has served us well thus far and we do this via https://github.com/openshift/configure-alertmanager-operator/blob/master/pkg/controller/secret/secret_controller.go#L136-L249 . This is not pretty :)  but gives you an example of what SRE is silencing based on X,Y,Z values that are avaialbe within alert metadata. 

For usability + console, I think two functionalities. 

- silence alert
- create alert (the create could enable the user to pick an current alert to template from, but this in no way links them together)
- created alerts could be stored in PrometheusRule called 'custom-admin-alerts'

enhancements/monitoring/alert-overrides.md
bison: Yeah, that's basically it. For each entry in the `overrides` section of the `ConfigMap`, find the matching rule in the existing `PrometheusRule` objects, then apply the action -- that is either `drop` or `patch`.

If the action is `patch`, apply the user's changes on top of the existing rule, and add the result to the "alert-overrides" `PrometheusRule`. These always get a static label identifying them as a generated override rule. That label is what allows us to drop the original rule in the relabel config, because we can match only rules that *don't* have the label.

So, the last step is to then generate the `alert_relabel_configs` which drops the original version of any overridden alerts and any alerts the user specified the `drop` action for.

enhancements/monitoring/alert-overrides.md
bison: > another example here is Kube* rules that are generated with a namespace= field. SRE use this currently to determine what should/could be ignored.

Sorry, not sure I follow. Do you mean that you'd like to be able to match on the `namespace` label in addition to name and severity when overriding an alert?

bison: I've added some content about this failure mode.

enhancements/monitoring/alert-overrides.md
bison: Added a note about this to the proposal section.

enhancements/monitoring/alert-overrides.md
bison: I've added a bit trying to summarize this better in the proposal section.

enhancements/monitoring/alert-overrides.md
bison: Thanks, I made this a bit more general.

bison:
> I'm wondering if it makes sense to call out the upgrade scenario where an upgrade changes an overridden alert more explicitly. What happens if the spec no longer matches an alert or now matches multiple? Even listing this as "can happend, considered out of scope" would be beneficial I think.

@jan--f, yeah, this is definitely something to think about. I guess in the case of OpenShift simply removing an alert, it's no worse than it would have been. The original alert and the generated alert with the overrides both go away.

I think the UX is a little stranger if OpenShift renames an alert or changes its severity. In that case, my overrides are no longer applied, and I guess I have to go look at the cluster-monitoring-operator (or the CRD if we do that) to know why. In the worst case, that could lead to a critical alert not being routed the way I expect. Though, again, that's maybe not too different from the case without this feature.

bison:
> Nice job here @bison ! this stuff seemingly gets really complicated quickly. Maybe too many options here become un-productive. I know for OSD we simply do:
> 
> Silence an alert
> Create an alert (if the silence needs replacing/customizing)

Thanks for the feedback, @dofinn. So, does this feature sound useful to you? Or are you saying you think silences and copying / creating new alerts is enough, and this is too complicated?

dofinn:
> > Nice job here @bison ! this stuff seemingly gets really complicated quickly. Maybe too many options here become un-productive. I know for OSD we simply do:
> > Silence an alert
> > Create an alert (if the silence needs replacing/customizing)
> 
> Thanks for the feedback, @dofinn. So, does this feature sound useful to you? Or are you saying you think silences and copying / creating new alerts is enough, and this is too complicated?

I think the ability to silence and create alerts combined with ease of management:
* what alerts have I silenced?
* what alerts have I created?
has great merit.

As mentioned previously, SRE already does this using workarounds. These workarounds are not 
necessarily bad but Im thinking this enhancement could/would provide a better way to acheive this. 
Silences/modificiations - > https://github.com/openshift/configure-alertmanager-operator/blob/master/pkg/controller/secret/secret_controller.go#L136
Creates -> https://github.com/openshift/managed-cluster-config/tree/master/deploy/sre-prometheus  

Managing "whats replacing what" or "what alert have i modified" (especially across releases) seems like it would introduce large complexities.

enhancements/monitoring/alert-overrides.md
fpetkovski: A CRD also provides better UX since it has built-in validation and documentation mechanisms

enhancements/monitoring/alert-overrides.md
bison: This makes sense to me. I've updated the proposal with a CRD structure that allows matching against an arbitrary set of static labels, but requires an exact match.

enhancements/monitoring/alert-overrides.md
bison: Added a note about telemetry. Thanks.

enhancements/monitoring/alert-overrides.md
bison: We ended up deciding a CRD is best. I've added some detail about the proposed structure.

enhancements/monitoring/alert-overrides.md
bison: Agreed. I've added a proposed structure for a CRD now.

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,310 @@
...
+OpenShift is an opinionated platform, and the default alerting rules have of
+course been carefully crafted to match what we think the majority of users need.
+Nevertheless, users have repeatedly asked for the ability to modify alerting
+rules to adjust thresholds, or to simply add labels and annotations to aide in
+routing alert notifications, as well as a supported method of adding additional
+alerting rules.
+
+Users currently have the ability to add additional rules from a technical
+standpoint -- by simply creating new `PrometheusRule` objects -- but this is
+explicitly unsupported at the moment.  The reason being that these resources are
```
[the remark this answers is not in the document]
bison: Done.

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,310 @@
...
+
+For each override a new alerting rule is generated with the overrides applied on
+top of the existing rule, and this new rule is added to the "alert-overrides"
+`PrometheusRule` object.  Additional user-supplied rules are also added here.
+All generated alerts will be given an identifying label -- currently:
+`monitoring_openshift_io__alert_override=true`
+
+This leaves the original alerting rules present and untouched.  These, as well
+as any rules listed with the `drop` action, must be ignored by the platform
+Prometheus instance.  A `Secret` containing [alert_relabel_configs][1] data for
```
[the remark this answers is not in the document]
bison: Fixed. Thanks.

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,310 @@
...
+    this override.
+- Generate new `PrometheusRule` objects for any new user-defined rules.
+- Generate the [alert_relabel_configs][1] to drop original alerts that have been
+  replaced with a patched version, and any that were explicitly dropped.
+
+A note on the failure mode when multiple alerting rules match a name and
+severity pair: This doesn't seem to be a big problem in practice for the current
+OpenShift platform alerting rules -- there seem to be only two cases.  We may
+want to add the requirement that these be unique to the [style-guide][2]
+however, and enforce this in e2e tests going forward.
```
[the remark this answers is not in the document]
bison: Have changed this to a CRD that allows matching an arbitrary set of static labels, but requires a unique match.

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,331 @@
...
+
+```yaml
+apiVersion: v1alpha1
+kind: alertoverrides.monitoring.openshift.io
+metadata:
+  name: alert-overrides
+  namespace: openshift-monitoring
+spec:
+  # List of overrides for alerts in platform namespaces.
+  overrides:
```
fpetkovski, line 98: I think this might be redundant because the resource name is already `alertoverrides`

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,331 @@
...
+  - selector: # required
+      alert: "KubeAPIErrorBudgetBurn" # required
+      matchLabels: # optional, but multiple matches is an error reported in status.
+        severity: "critical"
+        long: "1h"
+    action: "patch"
```
fpetkovski, line 104: Instead of having the action as a value, have we considered having it as a key? This would allow us to define different parameters for different actions in the future if such a need arises. E.g.
```
spec:
  patch:
    ...
  drop:
    ...
```

enhancements/monitoring/alert-overrides.md
bison: Yeah, it's basically just there because it differentiates the list of overrides from the list of new additional rules. Not sure what else to call it.

enhancements/monitoring/alert-overrides.md
bison: I don't have a super strong opinion on this. The way it is now somehow feels like a better middle ground between being Kubernetes-ish, and looking a bit like the existing Prometheus / Alertmanager configs to me. But if the consensus is to have a separate key for each action, that works too.

Good point on changing the parameters. Though, I guess we could already add new optional parameters in the existing version, and anything else probably needs a bump to the API resource version anyway, right?

enhancements/monitoring/alert-overrides.md
fpetkovski: Ah ok, I forgot about the additional rules. Then this makes sense :+1:

bison:
I also wanted to ping @openshift/team-cnv and @openshift/openshift-team-storage -- I think there were people on those teams that had expressed interest in this.

enhancements/monitoring/alert-overrides.md
fpetkovski: I think the advantage of separate keys is that we can have different configuration options for each operation. This makes it more visible to the user which option can be used for each action, and also adds some "typing" to action parameters.

But anyway, it's just an option to consider. I don't think it's really a show stopper to have the action as a value.

raptorsun: Agree. This is easier to read when we divide the list into patch, drop, add sections.

jsafrane:
From storage team perspective, we're interested for sure. This proposal looks OK. It would be nice to have if users could see in some "advanced view":

* What alerts definitions are patched / dropped.
* What alerts would be firing right now if they were not patched / dropped.

I think both are really hard to achieve, just leaving here for consideration.

raptorsun:
It would be great if more details can be given on validation of the rule specification. for example
- how to validate the expression is valid, every referenced rule exists
- when dropping a rule, other rules are not referencing to it

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,331 @@
...
+    labels:
+      patched: "true"
+
+  - selector:
+      alert: "KubeAPIErrorBudgetBurn"
+      matchLabels: # not a unique match, this is an error.
+        severity: "critical"
+    action: "patch"
+    labels:
+      patched: "true"
```
jan--f, line 114: Afaiu any patched rule will have a automatic label add to it, so that we can identify its as a patched alert. If a patch directive attempts to overwrite this generated label, we'll run into issues where users can mess with the ability to discern whether an alert stems from a shipped rule or from an overwritten one. For example assume here the generated labelname is `patched` and a patch rule adds `patched: "false"`.
The proposal should specifically call out this (error) scenario.

bison: This is a good point, but having thought about it again, we add the label identifying the alert as patched to the newly generated rule, and we do it after applying the user-supplied overrides. So in practice, it's not possible to change it. 

Also, in the relabel config, we match the original alerts by testing for the absence of the label. We don't actually care what the label's value is. So I don't think this presents a problem in the current design.

bison: Added a note about the label being added after patches are applied.

bison:
> It would be great if more details can be given on validation of the rule specification. for example
> 
> * how to validate the expression is valid, every referenced rule exists
> * when dropping a rule, other rules are not referencing to it

Good points, @raptorsun. I added a note about validation with a link to the code Prometheus uses. Should be easy to reuse and allow validation of each generated rule.

I'm not sure I fully understand what you mean about dropped rules being referenced. Though, keep in mind dropped rules are only prevented from being sent to Alertmanager. They still exist in Prometheus. Can you maybe give an example?

enhancements/monitoring/alert-overrides.md
```diff
@@ -0,0 +1,336 @@
...
+- Watch for a user-defined list of overrides and new rules.
+- For each override, find the existing rule matching the name and labels, and
+  apply the action specified by the user, i.e. `patch` or `drop`.
+  - If the action is `patch`, generate a new rule with supplied overrides.
+  - If the action is `drop`, simply add the alert to the alert relabel config.
+  - If multiple alerts match the name and labels, notify the user and skip
+    this override.
+- Generate new `PrometheusRule` objects for any new user-defined rules.
+- Generate the [alert_relabel_configs][1] to drop original alerts that have been
+  replaced with a patched version, and any that were explicitly dropped.
```
jan--f, line 199: I like the operator approach. I wonder if a scenario is missing here:
Should the operator not also watch `PrometheusRule` objects for changes? The original object might change requiring the generated object to be regenerated? I'm not really sure if the `PrometheusRule` object that CMO creates is an input to the object generated by this new operator.

In any case it might be worth calling this out either way here.

bison: Good point, that is already happening in the proof-of-concept branch. I added a note here about it:
> Processing of overrides is also triggered by `PrometheusRule` object changes, and changes to the set of platform namespaces.

bison:
I've updated https://github.com/openshift/cluster-monitoring-operator/pull/1473 to match the implementation described here -- using a CRD and matching on label sets.

bison:
I started a possible alternative proposal here: https://github.com/openshift/enhancements/pull/1030 -- Not sure if that's the best way to go about that, but I didn't know what else to do.

PR #960, opened 2021-11-22
Insights pulling SCA certs - graduation criteria update

tremes:
This is an update to the graduation criterion for informing user about the error state - especially when Insights receives HTTP 404, which means that the SCA/entitlements are not enabled in the OCM.

### DIFF
--- a/enhancements/insights/pulling-sca-certs-from-ocm.md
+++ b/enhancements/insights/pulling-sca-certs-from-ocm.md
@@ -139,7 +139,7 @@ This feature is planned as a technical preview in OCP 4.9 and is planned to go G
 
 #### Tech Preview -> GA
 - ~~`Share` resource object is automatically created by the Insights Operator~~ - This is no longer the GA requirement. `Share` object will still be a tech preview and can be added later by different component (from Build team)
-- inform a cluster user about the error state (problem with pulling the certificates)
+- inform a cluster user about the error state (problem with pulling the certificates) - The OCM API returns HTTP 404 if the entitlement certificates are not enabled for the corresponding organization. To inform user/cluster admin about this fact, we can create a new ClusterOperator condition (e.g "SCANotAvailable") and create a new info alert based on this new condition.
 - the documentation is revisited and updated
 - the use of SCA certs in the `Build` is documented in the Build API documentation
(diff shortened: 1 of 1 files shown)


enhancements/insights/pulling-sca-certs-from-ocm.md
```diff
@@ -139,7 +139,7 @@ This feature is planned as a technical preview in OCP 4.9 and is planned to go G
 
 #### Tech Preview -> GA
 - ~~`Share` resource object is automatically created by the Insights Operator~~ - This is no longer the GA requirement. `Share` object will still be a tech preview and can be added later by different component (from Build team)
-- inform a cluster user about the error state (problem with pulling the certificates)
+- inform a cluster user about the error state (problem with pulling the certificates) - The OCM API returns HTTP 404 if the entitlement certificates are not enabled for the corresponding organization. To inform user/cluster admin about this fact, we can create a new ClusterOperator condition (e.g "ScaNotEnabled") and create a new info alert based on this new condition.
```
tremes, line 142: Another option (instead of the ClusterOperator condition) would metric (probably counting number of 404 responses).....but I think I like the condition approach more.

bparees: What about setting the Degraded=true condition with an appropriate Reason and Message?

bparees: hm, so essentially you're saying this is an optional functionality of the Insights operator, and getting 404s retrieving certs is potentially "to be expected" even when the cluster is running+configured correctly?

In that case your original proposal seems ok.  Though i'd make it "SCANotEnabled" since i'm not a fan of lowercased acronyms, but if you've got a precedent to cite, i won't stand in the way.

tremes: Yes that's what I am trying to say :-)

I am fine with "SCANotEnabled"...I don't have any precedent. Thanks

tremes: Updated the name suggestion.

bparees:
@tremes someone from your team should lgtm this

PR #963, opened 2021-11-23
Proposal to allow MTU changes with rolling reboots

jcaamano:
This supersedes #926 and expands on the alternative of performing the MTU change through rolling reboots.

/cc @mccv1r0
For MTU changes
/cc @trozet
For OVN kubernetes
/cc @danwinship @knobunc @dcbw
For general thoughts

### DIFF
--- a/enhancements/network/allow-mtu-changes.md
+++ b/enhancements/network/allow-mtu-changes.md
@@ -1,0 +1,363 @@
+---
+title: allow-mtu-changes
+authors:
+  - "@juanluisvaladas"
+  - "@jcaamano"
+reviewers:
+  - "@danwinship"
+  - "@dcbw"
+  - "@knobunc"
+  - "@msherif1234"
+  - "@trozet"
+  - "@yuqi-zhang"
+approvers:
+  - TBD
+creation-date: 2021-10-07
+last-updated: 2021-12-02
+status: provisional
+---
+
+# Allow MTU changes
+
+This covers adding the capability to the cluster network operator of changing
+the MTU post installation.
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Customers may need to change the MTU post-installation. However these changes
+aren't trivial and may cause downtime, hence Cluster Network Operator currently
+forbids them.
+
+This enhancement proposes an automated procedure launched on demand and
+coordinated by the Cluster Network Operator. This procedure is based on draining
+and rebooting the nodes sequentially after performing the required configuration
+changes to minimize service disruption and to ensure that the  
+nodes have a healthy configuration and that workloads preserve connectivity
+during and after the procedure.
+
+## Motivation
+
+While cluster administrators usually set the MTU correctly during the
+installation, sometimes they need to change it afterwards for reasons such as
+changes in the underlay or because they were set incorrectly at install time.
+
+### Goals
+
+* Allow to change MTU post install on OVN-Kubernetes or OpenshiftSDN.
+
+### Non-Goals
+
+* Other safe or unsafe configuration changes.
+
+## Proposal
+
+Cluster Network Operator (CNO from now on) will drive the MTU change when it
+detects the change on the operator network configuration. The procedure is based
+on two rolling reboot sequences that will be coordinated through Machine Config
+Operator (MCO from now on). On the first reboot, routes MTU (`routable-mtu` from
+now on) will be used to pin the effective MTU in use to the lower value allowing
+changes to the MTU set on interfaces (`mtu` from now on) without disruption. On
+a second reboot, once all the interfaces are set to the final MTU value, route
+MTU is unset to make that interface MTU the one in effect. In more detail:
+
+1. Set `mtu` to the higher MTU value and `routable-mtu` to the lower MTU value
+   of the current and target MTU values.
+2. Do a rolling reboot. As a node restarts, `routable-mtu` is set on routes
+   while interfaces have `mtu` configured. The node will effectively use the
+   lower `routable-mtu` for outgoing traffic, but be able to handle incoming
+   traffic up to the higher `mtu`.
+4. Set `mtu` to the target MTU value and unset `routable-mtu`.
+5. Do a rolling reboot. As a node restarts, the target MTU value is set on the
+   interfaces and the routes are reset to default MTU values. Since the MTU
+   effectively used in yet to be restarted nodes of the cluster is the lower one
+   but able to handle the higher one, this has no impact on traffic.
+
+Due to the reboots it is adequate to consider this an MTU migration rather than
+just simply an MTU change. Given the impact that the reboots have on the
+cluster, this operation will be safeguarded by a specific migration annotation
+or field.
+
+The procedure will include handling the MTU of the cluster pod network and the
+MTU of the host itself as is critical to do so in a coordinated manner to avoid
+service disruptions and more reboots than the minimum necessary.
+
+The procedure will optionally leverage Node Feature Discovery operator (NFD from
+now on) through witch information about MTU capabilities of the nodes as well as
+information of MTU migration status will be annotated on the nodes. This
+information can be used to validate and track progress of the procedure.
+
+Changes will be required on openshift-api, OpenshiftSDN, OVN-Kubernetes, CNO and
+MCO.
+
+### OVN-Kubernetes & OpenshiftSDN: new `routable-mtu` configuration parameter
+
+A new `routable-mtu` setting for the CNI, additional to the already existing
+`mtu` one, is introduced to be set as MTU in the following routes:
+* default route in pods, covering traffic from a pod to any destination that is
+  not a pod on the same host.
+* non link scoped host routes of interfaces managed by the CNI covering traffic
+  from host network to pods hosted on different nodes
+* host routes of interfaces managed by the CNI covering traffic to the service
+  network.
+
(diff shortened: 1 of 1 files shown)


jcaamano:
Not ready for review yet, working through some corner cases.

jcaamano:
/cc @yuqi-zhang 
For any possible comments related to MCO usage.

enhancements/network/allow-mtu-changes.md
```diff
@@ -0,0 +1,364 @@
...
+If that is the case:
+* CNO status should be progressing if MachineConfig pool has not yet sourced a
+  MachineConfig or is otherwise pending for managed nodes to be updated and
+  ready.
+* CNO status should be degraded if MachineConfig pool status is degraded.
+
+The CNO should clean up MachineConfig objects that are no longer needed in the
+procedure which deserves a closer look since this might trigger unwanted
+reboots. This happens at two times in the procedure:
+* `cno-set-target-mtu-<TARGET_MTU>-<pool>` MachineConfig object is removed when
```
yuqi-zhang, line 187: The MCO doesn't have a batch changes concept. If you really want, you could just modify the existing MachineConfig from A to B (so instead of creating `cno-unset-target-mtu-<pool>`, you can simply modify `cno-set-target-mtu-<TARGET_MTU>-<pool>` to the new desired state

enhancements/network/allow-mtu-changes.md
```diff
@@ -0,0 +1,364 @@
...
+  `cno-unset-target-mtu-<pool>` MachineConfig object is applied which overall
+  causes one intended reboot.
+* `cno-unset-target-mtu-<pool>` MachineConfig object is removed when the
```
yuqi-zhang, line 190: Hmm, wait if this originally existed, why not just delete the above machineconfig to roll out the second change?

yuqi-zhang (commented):
A few questions from the MCO perspective:

1. Is there a configuration change required from the MCO perspective? You do mention creating a dropin, but does does that actually affect anything? Or is it just being used as a reboot marker? The later comments of removing all the MachineConfigs seem to indicate no.

And some other questons inline:

enhancements/network/allow-mtu-changes.md
```diff
@@ -0,0 +1,372 @@
...
+  MachineConfig objects without a machine pool label. The administrator could
+  then add the label to manually trigger the reboots. We could also offer the
+  option to set which MachineConfig pools to work with, to provide more control
+  over the reboot process. The counterpart is that if reboots are manual,
+  unintended reboots unrelated to this procedure before the label is applied to
+  the Machine configs might lead to a degraded system.
+* We might not be aware at any given time of br-ex link scoped routes if these
+  are dynamically assigned through IPv6 RA. NetworkManager dispatcher does not
+  seem to trigger an event for route modifications resulting from IPv6 RA.
+* Should the host MTU change outlive OVN-K? For example, if migrating back to
```
trozet, line 280: if we migrate back to openshift-sdn, br-ex and the associated connections will be deleted, so whatever MTU the host interfaces connections were configured with will be used. Right?

trozet (changes requested):
overall looks good. Couple of minor comments and would add a detailed procedure about the manual alternative method that may be used in the short term (due to time constraint) and then later the more complex automated CNO and machine config bits can be added. Summary of manual steps:
1. configure CNO with routable-mtu and mtu
2. reboot the nodes manually
3. unset routable-mtu
4. reboot the nodes again

jcaamano:
> A few questions from the MCO perspective:
> 
> 1. Is there a configuration change required from the MCO perspective? You do mention creating a dropin, but does does that actually affect anything? Or is it just being used as a reboot marker? The later comments of removing all the MachineConfigs seem to indicate no.
> 
> And some other questons inline:

@yuqi-zhang  The dropin causes configure-ovs to use specified MTU parameters within, so if does affect things. The MTU itself persists in nmconnection files that configure-ovs handles (or elsewhere if we need to) so the actual dropin or MCO doe snot need to persist so it is removed at the end.

enhancements/network/allow-mtu-changes.md
jcaamano: @yuqi-zhang I agree that it is preferable to just modify the contents but...

How can we know when a node has been updated with a MachineConfig object with specific contents?? The only way I know is to parse the rendered MachineConfig object and look there for the contents of the source MachineConfig. This is not very practical.

Whereas if we use different MachineConfig objects with deterministic names, we can just check if that name is listed in the source of the MachineConfigPool.

enhancements/network/allow-mtu-changes.md
jcaamano: We only want to derive CNO status from MCP status when machine state is changing due to CNO actions.  If we simply remove the MachineConfig then we don't know if the MCP is being updated because of that or any other change unrelated to CNO.

enhancements/network/allow-mtu-changes.md
```diff
@@ -0,0 +1,372 @@
...
+### MCO: configurable MTU for configure-ovs service
+
+configure-ovs is a systemd service deployed through MCO that prepares host
+networking for OVN-K. The proposal is to introduce a systemd dropin for this
+service with a configuration parameters `TARGET_MTU`, `MTU` and `ROUTABLE_MTU`
+that might be set during the MTU migration:
+* `MTU`, if set, shall be configured as MTU on br-ex and ovs-if-phys0
+  interfaces.
+* `ROUTABLE_MTU`, if set, shall be configured on br-ex link scoped routes.
+* If `TARGET_MTU` is unset, the previous `TARGET_MTU` should be configured as
```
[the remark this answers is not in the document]
jcaamano: Nothing. `TARGET_MTU` is just included to be aware it's previous value when it is unset.

enhancements/network/allow-mtu-changes.md
```diff
@@ -0,0 +1,372 @@
...
+The CNO monitors changes on the operator configuration and coordinates the MTU
+change. When it detects a MTU change in the operator configuration:
+1. Check if an annotation `mtu-migration` is set in the operator. If not,
+   consider the MTU change unsafe as done currently and don't proceed.
+2. Check if annotation `mtu-migration` has a value, and if so, that it is a
+   valid target host MTU value with respect to the target MTU value set in the
+   operator configuration. If not, consider the MTU change unsafe and don't
+   proceed. Otherwise, consider this to be the target host MTU value or assume a
+   default target host MTU value of the target MTU value + 100.
+3. Check that the target host MTU values are within the minimum and maximum
```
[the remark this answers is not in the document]
jcaamano: These are the maximum and minimum MTU values the node's physical nic can be configured with. This are annotated on the node through NFD and are obtained for example from `maxmtu` and `minmtu` of `ip -d link show`

enhancements/network/allow-mtu-changes.md
```diff
@@ -0,0 +1,372 @@
...
+annotations through NFD hooks. Values to be reported are current, minimum and
+maximum MTU of ovs-if-phys0, as well as target and routable MTU in use if a MTU
+migration is taking place. CNO will use this information to validate and track
+progress of an MTU migration.
+
+### CNO: automated coordination of MTU change
+
+The CNO monitors changes on the operator configuration and coordinates the MTU
+change. When it detects a MTU change in the operator configuration:
+1. Check if an annotation `mtu-migration` is set in the operator. If not,
```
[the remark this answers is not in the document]
jcaamano: Yes, for the host. While I do mention the annotation before without specifics, it is introduced here. But if you follow through the procedure I think it's clear what it is for.

enhancements/network/allow-mtu-changes.md
jcaamano: Yes, that would be a rollback to the old MTU and I guess that's what I am asking, would that be desirable or maybe not?

trozet: hmm I noticed the title of this talks about MTU change generically which makes me thing openshift-sdn is included, but the details of the proposal talk about ovnk changes. Are we trying to support this for openshift-sdn and ovnk or ovnk only?

enhancements/network/allow-mtu-changes.md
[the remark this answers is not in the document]
trozet: ah OK. It would be good to add that to the doc I think.

enhancements/network/allow-mtu-changes.md
trozet: the epic says openshift-sdn as well. We met offline and came up with a plan to make it work for openshift-sdn as well.

enhancements/network/allow-mtu-changes.md
```diff
@@ -0,0 +1,319 @@
...
+  CNI MTU migration.
+
+This will result in the required rolling reboots as `TARGET_MTU` is set and then
+unset during the MTU migration procedure.
+
+### CNO: allowing MTU migration
+
+The CNO monitors changes on the operator network configuration. When it detects
+a change in `Network.Spec.Migration`:
+1. Check that `MTU.Network.From` and `MTU.Network.To` are not independently set
```
trozet, line 162: why does this say "or different". Shouldn't what you are migrating From and migrating To be different?

enhancements/network/allow-mtu-changes.md
```diff
@@ -0,0 +1,319 @@
...
+
+Host tooling should write a file for NFD source hook to setup annotations for
+current MTU, minimum MTU, maximum MTU and routable MTU if any.
+
+### openshift-api: add MTU migration specific fields
+
+New `MTU.Network.From`, `MTU.Network.To` and `MTU.Machine.To` fields will be
+added in the existing `NetworkMigration` type which is used in CNO's cluster and
+operator network configurations and will contain the CNI and host MTU values to
+migrate from and to to respectively.
```
trozet, line 139: nit: to to

enhancements/network/allow-mtu-changes.md
jcaamano: Fixed. Actually To and From can be equal to attempt to rollback a MTU migration that is in progress when something goes wrong.

trozet (approved):
/approve

PR #965, opened 2021-11-23
CONVENTIONS: Include 33% option for DaemonSet maxUnavailable

wking:
Borrowing [Clayton's][1] [language][2] from the covering test code.

[1]: https://github.com/openshift/origin/blob/e945cb88da780e21c021b6c8b430454bcfb881cf/test/extended/operators/daemon_set.go#L29-L38
[2]: https://github.com/openshift/origin/pull/25928/commits/757edeb9d0b4a6175b15c2520f73e54f04add157#diff-78f020f621e098ac4cbf468ad03b87bbb7900797fa95e70c0877338cab243ab5R29-R38

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -199,17 +199,13 @@ To avoid disruption and mass pod-death, it is important to
   * Components that support workloads directly must not disrupt end-user workloads during upgrade or reconfiguration
     * E.g. the upgrade of a network plugin must serve pod traffic without disruption (although tiny increases in latency are allowed)
     * All components that currently disrupt end-user workloads must prioritize addressing those issues, and new components may not be introduced that add disruption
-* All dameonsets of OpenShift components should use the maxUnavailable rollout strategy, allowing up to 10% of instances to be taken down at once.
-Any bugs that block that should be fixed.
-`10%` is an arbitrary ratio that ensures that most of the cluster is up and
-serving clients, while a big-enough chunk of the cluster is upgradable.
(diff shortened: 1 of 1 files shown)


CONVENTIONS.md
```diff
@@ -199,17 +199,13 @@ To avoid disruption and mass pod-death, it is important to
...
-`10%` is an arbitrary ratio that ensures that most of the cluster is up and
-serving clients, while a big-enough chunk of the cluster is upgradable.
-There wasn't a huge amount of analysis beyond the fact that it significantly
-improved upgrade times and hit our targets. Additionally, 10% is roughly in
-line with the minimum disruption experienced by the default sized cluster
-whenever we reboot nodes and that's mandatory in order to complete an upgrade.
-Keeping the default of `1` makes an operator upgrade unacceptably slowly on
-big clusters because only one host running the daemonset can be drained (and upgraded) at a time.
-
+* All daemonsets of OpenShift components, especially those which are not limited to control-plane nodes, should use the `maxUnavailable` rollout strategy to avoid slow updates over large numbers of compute nodes.
```
dhellmann, line 202: Do we need a caveat here to not do this for single node topologies?

Do we need different guidance for external control plane topologies?

soltysh (approved):
This lgtm

CONVENTIONS.md
sdodson: If this were a problem for SNO it would be a problem with any non zero value including the default of 1.
I can't think of any problems w/ external control planes here.

PR #967, opened 2021-11-24
CONSOLE-3005: Provide alias for dynamic plugin's proxy

jhadvig:
Story: https://issues.redhat.com/browse/CONSOLE-3005

Changes are based on the [API change PR](https://github.com/openshift/api/pull/1067)

/assign @spadgett @christianvogt

### DIFF
--- a/enhancements/console/dynamic-plugins.md
+++ b/enhancements/console/dynamic-plugins.md
@@ -137,28 +137,38 @@ and base path used to access all of the plugin's assets.
 apiVersion: console.openshift.io/v1alpha1
 kind: ConsolePlugin
 metadata:
-  name: my-plugin
+  name: acm
 spec:
-  displayName: 'My Plugin'
+  displayName: 'Advanced Cluster Management'
   service:
-    name: my-console-plugin
-    namespace: my-operator-namespace
+    name: acm
+    namespace: open-cluster-management
     port: 8443
     basePath: '/'
 ```
 
 In case the plugin needs to communicate with some in-cluster service, it can
-declare a service proxy in its `ConsolePlugin` resource using the
-`spec.proxy.services` array field. A service `name`, `namespace` and `port`
-needs to be specified.
+declare a service proxy in its `ConsolePlugin` resource using the `spec.proxy` array.
+Each entry needs to specify type and alias of the proxy, under the `type` and `alias` field.
+For the `Service` proxy type, service `name`, `namespace` and `port` needs to be specified,
+to which the request will be proxied.
+
+```yaml
+spec:
+  proxy:
+  - type: Service
+    alias: <proxy-alias>
+    name: <service-name>
+    namespace: <service-namespace>
+    port: <service-port>
+```
 
 Console backend exposes following endpoint in order to proxy the communication
 between plugin and the service:
-`/api/proxy/namespace/<service-namespace>/service/<service-name>:<port-number>/<request-path>?<optional-query-parameters>`
+`/api/proxy/plugin/<plugin-name>/<proxy-alias>/<request-path>?<optional-query-parameters>`
 
-An example proxy request path from plugin to `helm-charts` service,
-in `helm` namespace to list ten helm releases:
-`/api/proxy/namespace/helm/service/helm-charts:8443/releases?limit=10`
+An example proxy request path from `acm` plugin with a `search` service is:
+`/api/proxy/plugin/acm/search/pods?namespace=openshift-apiserver`
 
 Proxied request will use [service CA bundle](https://docs.openshift.com/container-platform/4.8/security/certificate_types_descriptions/service-ca-certificates.html) by default. The service must use HTTPS.
 If the service uses a custom service CA, the `caCertificate` field
@@ -173,21 +183,22 @@ then passed in the HTTP `Authorization` request header, for example:
 apiVersion: console.openshift.io/v1alpha1
 kind: ConsolePlugin
 metadata:
-  name: my-plugin
+  name: acm
 spec:
-  displayName: 'My Plugin'
+  displayName: 'Advanced Cluster Management'
   service:
-    name: my-console-plugin
-    namespace: my-operator-namespace
+    name: acm
+    namespace: open-cluster-management
     port: 8443
     basePath: '/'
   proxy:
-    services:
-    - name: helm-charts
-      namespace: helm
-      port: 8443
-      caCertificate: '-----BEGIN CERTIFICATE-----\nMIID....'
-      authorize: true
+  - type: Service
+    alias: search
+    name: search
+    namespace: open-cluster-management
+    port: 8443
+    caCertificate: '-----BEGIN CERTIFICATE-----\nMIID....'
+    authorize: true
 ```
 
 Plugins are disabled by default. They need to be manually enabled by a cluster


christianvogt:
content changes lgtm

@jhadvig you have lint issues

enhancements/console/dynamic-plugins.md
```diff
@@ -149,16 +149,27 @@ spec:
...
+      <proxy-alias>:
+        name: <service-name>
+        namespace: <service-namespace>
+        port: <service-port>
+```
 
 Console backend exposes following endpoint in order to proxy the communication
 between plugin and the service:
-`/api/proxy/namespace/<service-namespace>/service/<service-name>:<port-number>/<request-path>?<optional-query-parameters>`
+`/api/proxy/namespace/<plugin-name>/<proxy-alias>/<request-path>?<optional-query-parameters>`
```
spadgett, line 168: Remove namespace from the URL?

I'd give it another path element so we can add something else under proxy/ if we want later.

```suggestion
`/api/proxy/plugin/<plugin-name>/<proxy-alias>/<request-path>?<optional-query-parameters>`
```

enhancements/console/dynamic-plugins.md
```diff
@@ -149,16 +149,27 @@ spec:
...
 
-An example proxy request path from plugin to `helm-charts` service,
-in `helm` namespace to list ten helm releases:
-`/api/proxy/namespace/helm/service/helm-charts:8443/releases?limit=10`
+An example proxy request path from `acm-console-plugin` plugin with an
+`acm-proxy` alias is:
+`/api/proxy/acm-console-plugin/acm-proxy/releases?limit=10`
```
spadgett, line 172:
```suggestion
An example proxy request path from the `acm` plugin with a
`search` service is:
`/api/proxy/plugin/acm/search/pods?namespace=openshift-apiserver`
```

enhancements/console/dynamic-plugins.md
```diff
@@ -149,16 +149,27 @@ spec:
...
-needs to be specified.
+`spec.proxy.services` map. The key of the map is the proxy alias
+of the service and value is an object in which service `name`, `namespace` and `port`
+needs to be specified and to which the request will be proxied.
+
+```yaml
+spec:
+  proxy:
+    services:
+      <proxy-alias>:
```
spadgett, line 160: Nit: this is really an alias for the service rather than for the proxy itself.

enhancements/console/dynamic-plugins.md
```diff
@@ -173,21 +184,22 @@ then passed in the HTTP `Authorization` request header, for example:
 apiVersion: console.openshift.io/v1alpha1
 kind: ConsolePlugin
 metadata:
-  name: my-plugin
+  name: acm-console-plugin
```
spadgett, line 187: I'd like to discourage plugin authors from putting `-plugin` at the end of all the plugin names :) Let's try to remove that from the examples.

```suggestion
  name: acm
```

enhancements/console/dynamic-plugins.md
```diff
@@ -173,21 +184,22 @@ then passed in the HTTP `Authorization` request header, for example:
...
 spec:
-  displayName: 'My Plugin'
+  displayName: 'ACM console plugin'
```
spadgett, line 189:
```suggestion
  displayName: 'Advanced Cluster Management'
```

enhancements/console/dynamic-plugins.md
```diff
@@ -173,21 +184,22 @@ then passed in the HTTP `Authorization` request header, for example:
...
-      namespace: helm
-      port: 8443
-      caCertificate: '-----BEGIN CERTIFICATE-----\nMIID....'
-      authorize: true
+      acm-proxy:
+        name: helm-charts
+        namespace: helm
+        port: 8443
+        caCertificate: '-----BEGIN CERTIFICATE-----\nMIID....'
+        authorize: true
```
spadgett, line 202: Let's give this a more realistic example.

```suggestion
      search:
        name: search
        namespace: open-cluster-management
        port: 8443
        caCertificate: '-----BEGIN CERTIFICATE-----\nMIID....'
        authorize: true
```

enhancements/console/dynamic-plugins.md
```diff
@@ -173,21 +184,22 @@ then passed in the HTTP `Authorization` request header, for example:
...
-  name: my-plugin
+  name: acm-console-plugin
 spec:
-  displayName: 'My Plugin'
+  displayName: 'ACM console plugin'
   service:
-    name: my-console-plugin
-    namespace: my-operator-namespace
+    name: acm-console-plugin
+    namespace: acm-console-plugin-namespace
```
spadgett, line 192:
```suggestion
    namespace: open-cluster-management
```

enhancements/console/dynamic-plugins.md
```diff
@@ -149,16 +149,27 @@ spec:
...
-`spec.proxy.services` array field. A service `name`, `namespace` and `port`
-needs to be specified.
+`spec.proxy.services` map. The key of the map is the proxy alias
+of the service and value is an object in which service `name`, `namespace` and `port`
+needs to be specified and to which the request will be proxied.
+
+```yaml
+spec:
+  proxy:
+    services:
```
spadgett, line 159: I wonder if we need `services` here now. If we want to add other types of proxies, we want the alias to be unique. I'd suggest:

```yaml
spec:
  proxy:
    my-alias:
      service:
         name: my-name
         namespace: my-namespace
         port: my-port
      caCertificate: ...
      authorize: true
```

Then later on we could do something like

```yaml
spec:
  proxy:
    my-alias:
       url: https://example.com/
       caCertificate: ...
```

spadgett (changes requested):

enhancements/console/dynamic-plugins.md
jhadvig: sorry wanted to remove the `namespace`

enhancements/console/dynamic-plugins.md
jhadvig: then I guess it should be `service-proxy-alias` ?

enhancements/console/dynamic-plugins.md
jhadvig: So after talking to @stlaz on the API design here we agreed that best would be to have the `spec.proxy` structured as an array, which would contain `alias`, `authorize`, `caCertificate` and  enum `type` field. 
Based on the `type` a proxy config struct would be used.

eg.
```go
type ConsolePluginSpec struct {
	DisplayName string `json:"displayName"`
	Service ConsolePluginService `json:"service"`
	Proxy []ConsolePluginProxy `json:"proxy"`
}


type ConsolePluginProxy struct {
	Type ProxyType `json:"type"`
	Service ProxyServiceConfig `json:"service"`
	URL ProxyTypeURLConfig `json:"url"`
	CACertificate string `json:"caCertificate"`
	Authorize bool `json:"authorize"`
}

type ProxyType string

const (
	ProxyTypeService ProxyType = "Service"
	ProxyTypeURL ProxyType = "URL"
)

type ProxyTypeServiceConfig struct {
	Name string `json:"name"`
	Namespace string `json:"namespace"`
	Port int32 `json:"port"`
}

type ProxyTypeURLConfig struct {
	URL string `json:"url"`
}
```

This is the idiomatic way how the API should be structured... check[ indentityProviders API for example
](https://github.com/openshift/api/blob/9b3afd95c93af476bec246b2c60f64235f078198/config/v1/types_oauth.go#L164-L239)

spadgett: @jhadvig That looks good to me.

enhancements/console/dynamic-plugins.md
```diff
@@ -493,6 +504,22 @@ None
 
 None
 
+### API Extensions
```
spadgett, line 507: Let's override the job instead of adding these sections per the guidance here:

https://github.com/openshift/enhancements/#template-updates

spadgett (approved):
/approve

Updates look good to me.

spadgett:
Overriding per guidance here: https://github.com/openshift/enhancements/#template-updates

/override markdownlint

spadgett (approved):
/lgtm

PR #968, opened 2021-11-30
add audit spec for password authenticator flow

ibihim:
## What

Adding specification for `audit` events in the case that the authentication happens on the `oauth-server`.

## Why

We didn't specify how the `audit` event should look like and I would like to take the opportunity to discuss this with the team.

## Note

It doesn't yet contain the specification for IdentityProvider cases.

@s-urbaniak, @stlaz, @slaskawi, @EmilyM1

### DIFF
--- a/enhancements/authentication/login-logout-events.md
+++ b/enhancements/authentication/login-logout-events.md
@@ -102,6 +102,123 @@ const (
 )
 ```
 
+We propose to add additional information that helps to identify login and login failures to the audit log.
+
+In case that the authentication happens through the `oauth-server`, we suggest to add:
+
+- `authentication.openshift.io/username`, which is the username for the authentication attempt.
+- `authentication.openshift.io/decision`, which is an enum that can be `allow`, `deny` or `error`.
+
+An audit event for an unsuccesful authentication event would look like so:
+
+```JavaScript
+{
+  "kind": "Event",
+  "apiVersion": "audit.k8s.io/v1",
+  "level": "RequestResponse",
+  "auditID": "${ auditID }",
+  "stage": "ResponseComplete",
+  "requestURI": "${ requestURI }",
+  "verb": "get",
+  "user": {
+    "username": "system:anonymous",
+    "groups": [
+      "system:unauthenticated"
+    ]
+  },
+  "sourceIPs": [
+    "${ sourceIP1 }"
+  ],
+  "userAgent": "Go-http-client/1.1",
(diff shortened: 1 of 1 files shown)


enhancements/authentication/login-logout-events.md
```diff
@@ -102,6 +102,92 @@ const (
 )
 ```
 
+We propose to add additional information that helps to identify login and login failures to the audit log.
+
+In case that the authentication happens through the `oauth-server`, we suggest to add:
+
+- `authentication.openshift.io/username`, which is the username for the authentication attempt.
+- `authentication.openshift.io/authn-method`, which mentions the OAuth2 authentication method in use.
```
stlaz, line 110: This can be implied based on the request URL, no?
Why is this important?
Should it be `oauth-method` instead of `authn-method`, then?

enhancements/authentication/login-logout-events.md
```diff
@@ -102,6 +102,92 @@ const (
...
+- `authentication.openshift.io/decision`, which is an enum that can be `allow` or `deny`.
+- `authentication.openshift.io/reason`, which defines a reason in case that the `authentication.openshift.io/decision` is set to `deny`. In case that it is `allow`, the value is an empty `string`.
```
stlaz, line 112: Aren't these two going to be redundant to the response status and message?

enhancements/authentication/login-logout-events.md
ibihim: Yeah, this is really a thing that makes me wonder:

- UX for "just reading the annotations" during a "wake up call in the night"
vs
- redundant information that could (Murphy's Law) contradict itself.

The method and the decision can be deduced from the other values.

ibihim: I didn't want say "oauth-method" as I would rather think in terms of implicit grant or code grant when thinking of oauth-methods. So as we pick "just" the authentication method, I went for authn.

enhancements/authentication/login-logout-events.md
ibihim: More or less... the reason might have more context.

enhancements/authentication/login-logout-events.md
stlaz: I think that for now we can just go with username and decision to keep things minimal. We can add more later if there's such a request

stlaz (changes requested):

PR #971, opened 2021-12-01
Insights Operator pulling and updating pull-secret (transferring uncl…

tremes:
…assified clusters)

### DIFF
--- a/enhancements/insights/pulling-pull-secret-from-ocm.md
+++ b/enhancements/insights/pulling-pull-secret-from-ocm.md
@@ -1,0 +1,191 @@
+---
+title: pulling-and-updating-pull-secret-from-ocm
+authors:
+  - "@tremes"
+reviewers:
+  - "@mrunalp"
+  - "@aravindhp"
+  - "@inecas"
+  - "@rawsyntax"
+  - "@markturansky"
+approvers:
+  - "@mrunalp"
+  - "@aravindhp" 
+creation-date: 2021-12-01
+last-updated: 2021-12-16
+status: implementable
+see-also:
+  - "https://issues.redhat.com/browse/CCXDEV-6529"
+  - "https://issues.redhat.com/browse/CCX-185"
+---
+
+# Insights Operator pulling and updating `pull-secret` from the OCM API
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [x] User-facing documentation is created/updated in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement will enable the Insights Operator to pull the updated `pull-secret` (from the OCM API)
+and that will enable cluster ownership transfer. Currently, the ownership transfer is a two step process
+which involves initiating the transfer from the OpenShift Cluster Manager (OCM) API
+and then updating the global `pull-secret` on the cluster with the new owner’s `pull-secret`.
+If the `pull-secret` is not changed on the cluster within 5 days, the ownership transfer is canceled
+and needs to be initiated again.
+This process applies to all scenarios that require cluster ownership transferred across Red Hat accounts.
+
+The intent is to simplify the cluster ownership process to a one step process.
+
+## Motivation
+
+The entry point to many OpenShift subscriptions is for a user to start with a 60 day trial from the OCM with personal credentials
+and then migrate it over to a company or organization account. In addition, reports from the field indicate that customers in general when transferring ownership, either follow one of the steps in the process and do not follow through with the other step required to complete the transfer. This enhancement will make the transfer frictionless and provide a better user experience.
+
+
+### Goals
+The goal of this enhancement is to describe the API and the interaction between OCM and the cluster that is being transferred to update the `pull-secret` on that cluster:
+
+- Extend the Insights Operator config with a new OCM API endpoint to be able to query the new `pull-secret` content
+- Periodically (once a day) query the new endpoint and if there is a new or updated content then update the `pull-secret`.
+  
+### Non-Goals
+- The enhancement will not describe the OCM user experience.
+- The enhancement will not provide a generic way to pull information from OCM into the cluster.
+- The process will not apply to disconnected clusters.
+- This process will work only on the clusters where the feature was released as part of the Insights operator and will not be backported to older versions.
+
+## Proposal
+
+The Insights operator already communicates with the OCM API (see [Insights Operator pulling and exposing SCA certs from the OCM API](../insights/pulling-sca-certs-from-ocm.md). This is the proposed flow for updating the `pull-secret`:
+
+1. OCM provides an API endpoint that the cluster can call to check the current global updated `pull-secret`.
+2. User starts the migration process in the OCM by providing the `pull-secret` associated with the account to which the ownership will be transferred to.
+   * The details of whether the user provides the pull secret as a means of verification, or whether a request/approval workflow is implemented, is left to the OCM/UI flow. What is pertinent here is that there is an endpoint available to a cluster entity to fetch the new `pull-secret`.
+3. OCM validates the `pull-secret` for the new account and makes it available to the cluster for some time period i.e. the new API will now return the new `pull-secret` provided by the user.
+4. The Insights operator calls this API periodically, fetches the updated `pull-secret` and compares it with the current global `pull-secret` on the cluster (only entries available in both pull-secrets are compared).
+5. If there is a mismatch, the Insights operator applies the new `pull-secret`.
+   * The updated `pull-secret` from the OCM API is an authoritative source. Insights operator replaces and adds (There should not be any new entries, but this may change in the future) the entries provided in the updated `pull-secret` from the OCM API (This means that the Insights operator cannot delete anything in the existing secret).
+   * Note: MCO updating the `pull-secret` on the nodes will not result in node reboots. This will require an update of the documentation in https://docs.openshift.com/container-platform/4.9/updating/update-using-custom-machine-config-pools.html#update-using-custom-machine-config-pools-about_update-using-custom-machine-config-pools.
+   We should emphasize the need of unpausing the machine config pools to succesfully apply the new `pull-secret` (credentials) to all the nodes.
+6. The Insights operator itself starts using the new `pull-secret`
+7. Cluster comes up with the new `pull-secret` and starts reporting back to OCM into the new account.
+
+### User Stories
+
+As an OCM user, I would like to transfer cluster ownership from my account to another user's account in one step.
+- [CCXDEV-6529 Insights Operator pulling updated pull-secret to transfer unclassified cluster](https://issues.redhat.com/browse/CCXDEV-6529)
+- [CCX-185 Transfer of Unclassified clusters - provide pull-secret](https://issues.redhat.com/browse/CCX-185)
+### API Extensions
+
+No API extensions.
+
+### Risks and Mitigations
+
+Risk: The OCM API is down or returning HTTP response code >= 500.
+
+Mitigation: The Insights Operator is marked as degraded in such case.
+
+Risk: The `pull-secret` does not work or is otherwise corrupted.
+
+Mitigation: We must refer to the OCP documentation, which describes the manual fix procedure.
+
+Risk: Users don't want to wait 24 hours for the pull secret to be updated on the cluster
+
+Mitigation: User can trigger the update faster by deleting the Insights operator Pod (This needs to be documented and may change in the future - e.g we can introduce a new CRD to configure this).
+
+## Design Details
+
+### Authorization
+
+The Insights Operator is able to pull the updated `pull-secret` content from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config` namespace).
+
+The Insights Operator must provide a cluster ID as an identifier of the cluster.
+
+### OCM API
+
+The OCM API provides an endpoint that will serve the updated `pull-secret` definition when the migration window is opened. The endpoint will be defined in a new `ClusterTransfer` API - more details are in https://gitlab.cee.redhat.com/service/uhc-account-manager/-/blob/master/docs/cluster_transfer_api.md.
+The migration window will be opened for certain period of time (e.g 5 days) so the Insights Operator can safely update the secret.
+
+When the migration window is closed, the OCM API returns another HTTP status code and the Insights operator does nothing in such case.
+
+### Monitoring
+
+The OCM API will return time information providing the remaining time until the end of the current migration window. There will be an alert triggered when a certain number of days (1-2) remain in the current migration window.
+
+There is also plan to create an alert about pending `ClusterTransfer` on the OCM side - see https://issues.redhat.com/browse/SDB-2506,
+
+### Configuration
+
+The OCM API endpoint will be configurable in the Insights operator via the `support` secret (this is curenlty the only way to configure the Insights operator).
+
+### Disconnected clusters
+
+This feature will not work on disconnected clusters because there is no corresponding token (in the `pull-secret`) to communicate with Red Hat. We should emphasize this in https://docs.openshift.com/container-platform/4.9/support/remote_health_monitoring/opting-out-of-remote-health-reporting.html.
+
+
+### Test Plan
+
+- `insights-operator-e2e-tests` suite can verify that the Insights Operator is able to download and update the `pull-secret` content
+
+### Graduation Criteria
+
+There are no graduation criteria. This feature is planned as a fast track feature.
+
+#### Dev Preview -> Tech Preview
+
+N/A
+
+#### Tech Preview -> GA
+
+N/A
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+There is no upgrade/downgrade strategy needed.
+
+### Version Skew Strategy
+
+The Insights operator expects the `pull-secret` in the `openshift-config` namespace. If there are any changes to the `pull-secret` location or content, then the update of the Insights operator will be required.
+
+The OCM API is versioned and is expected to be backward compatible. A major update of the OCM API will probably require an update of the Insights operator.
+
+### Operational Aspects of API Extensions
+
+No API extension.
+
+#### Failure Modes
+
+There is no plan to provide direct confirmation of a successful `pull-secret` update. The problem is that the Insights operator is unable to provide such confirmation, because the update is followed up by the MachineConfig operator work. Some potential failures/problems are:  
+
+- There are some paused `MachineConfigPools`, which failed to pick up the updated `pull-secret` definition. The old original secret will
+still be valid for some period of time, but this can happen regardless of the period. We can create a new Insights rule (checking paused MCPs) to mitigate this.
+
+- Attacker who can forge an X.509 cert and impersonate OCM (or who gains server-side access to OCM) recommends an invalid pull secret.
+  This is basically same situation as described in [Risks and mitigations](#risks-and-mitigations) section above. There was a discussion about some pre-apply validation check of the new secret on the cluster side, but the output was that we can't do much about it (in the Insights operator) and we may revisit this in the future.
+
+#### Support Procedures
+
+A new Insights rule can be created based on the alerts described in the [Monitoring](#monitoring) section.
+
+## Implementation History
+
+There are no major milestones in the implementation history.
+
+## Drawbacks
+
+There should not be any significant drawbacks. The possible drawback is that the `pull-secret` update will not take place during the migration window (possible problems are described above). The original secret will continue to work and we will notify user (via alert as described above) and the procedure will need to be repeated.
+
+## Alternatives
+
+- In the future, we may have more flow of data from the OCM API to a cluster and we will revisit if we need a dedicated operator for such flows. However, we do not want to add an entirely new operator for the `pull-secret` flow today.
+


enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,168 @@
...
+
+No API extension.
+
+#### Failure Modes
+
+TBD
+
+#### Support Procedures
+
+TBD
```
tremes, line 155: @mrunalp, @aravindhp What about these two chapters? Do we need to mention anything? WDYT pls?

aravindhp: I think we need to call out how we are going to handle scenarios where the cluster is not able to update the secret in the cluster during Window for whatever reason. Maybe send an event so that the user is aware?

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,168 @@
...
+
+The Insights operator already communicates with the OCM API (see [Insights Operator pulling and exposing SCA certs from the OCM API](../insights/pulling-sca-certs-from-ocm.md). This is the proposed flow for updating the `pull-secret`:
+
+1. OCM provides an API that the cluster can call to check the current global `pull-secret`.
+2. User starts the migration process in the OCM by providing the `pull-secret` associated with the account to which the ownership will be transferred to.
+   * The details of whether the user provides the pull secret as a means of verification, or whether a request/approval workflow is implemented, is left to the OCM/UI flow. What is pertinent here is that there is an endpoint available to a cluster entity to fetch the new `pull-secret`.
+3. OCM validates the `pull-secret` for the new account and makes it available to the cluster for some time period i.e. the new API will now return the new `pull-secret` provided by the user.
+4. The Insights operator calls this API periodically, fetches the `pull-secret` and compares it with the current global `pull-secret` on the cluster.
+5. If there is a mismatch, the Insights operator applies the new `pull-secret`.
```
wking, line 69: This makes me nervous.  The cluster's global pull secret is [explicitly doc'ed as admin-editable][1].  And removing pull-secret entries is [explicitly doc'ed as a way to disable remote monitoring][2].  This line will have the Insights operator and the local cluster-admin sharing responsibility for the global pull secret, and sharing is hard ;).  How will Insights distinguish entries that the local admin has appended?  Perhaps by assuming that RH-supplied pull secrets will never drop entries for a particular domain?  How will Insights distinguish entries that the local admin has explicitly removed?  Perhaps by using the OCM-API token alone as a marker for "there's a mismatch" and then only updating tokens for domains that exist in the current global pullspec?  Even if all that gets sorted, this is still granting the Insights operator huge power to unilaterally break the cluster (e.g. if someone manages to compromise or MitM the OCM-API and recommend a pivot to a garbage secret).  I think this needs a very detailed entry in the risks and mitigation section.

[1]: https://docs.openshift.com/container-platform/4.9/openshift_images/managing_images/using-image-pull-secrets.html#images-update-global-pull-secret_using-image-pull-secrets
[2]: https://docs.openshift.com/container-platform/4.9/support/remote_health_monitoring/opting-out-of-remote-health-reporting.html#insights-operator-new-pull-secret_opting-out-remote-health-reporting

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,168 @@
...
+   * Note: MCO updating the `pull-secret` on the nodes will not result in node reboots.
```
wking, line 70: Which is great, unless the pools are paused ;).  If this lands, we'll need doc updates [here][1], and maybe in other places, that say "if you want your cluster to survive a cred handoff, be very sure to unpause and level all of your nodes before you disable the outgoing secret" or some such.

[1]: https://docs.openshift.com/container-platform/4.9/updating/update-using-custom-machine-config-pools.html#update-using-custom-machine-config-pools-about_update-using-custom-machine-config-pools

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,168 @@
...
+### OCM API
+
+The OCM API provides an endpoint that will serve the complete `pull-secret` definition when the migration window is opened. 
+The migration window will be opened for certain period of time (e.g 5 days) so tha the Insights Operator can safely update the secret.
+
+Open question: What exactly will the API return outside of the given migration window?  
+
+### Monitoring
+
+There will be a new metric registered by the Insights Operator, which will count the number of responses (according the HTTP response status code) from the OCM API.
```
wking, line 113: And then what happens to this metric?  Is it covered by alerting?  Or exported via Telemetry/Insights?  Or...?

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,168 @@
...
+
+The Insights Operator is able to pull the updated `pull-secret` content from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config` namespace).
+
+The Insights Operator must provide a cluster ID as an identifier of the cluster.
+
+### OCM API
+
+The OCM API provides an endpoint that will serve the complete `pull-secret` definition when the migration window is opened. 
+The migration window will be opened for certain period of time (e.g 5 days) so tha the Insights Operator can safely update the secret.
```
wking, line 107: nit: "tha the" -> "the"

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,168 @@
...
+
+Removing this feature will require updating the Insights operator code base.
+
+### Upgrade / Downgrade Strategy
+
+There is no upgrade/downgrade strategy needed.
+
+### Version Skew Strategy
+
+There is no Skew strategy needed. This work should have no impact on the upgrade. It doesn't require any coordinated behavior in the control plane. No other components will change.
```
wking, line 142: Assuming the in-cluster pull secret location doesn't move around, this seems safe.  But possibly worth talking through skew between different versions of the Insights operator and the OCM API service.  Or is the expectation that we get that API right on the first try and never have to update it going forward?

enhancements/insights/pulling-pull-secret-from-ocm.md
wking: Alerting, in both a public Insights rule and an on-cluster warning alert escalating to critical in the last day or something would be the loudest shouting for "the window is closing, didn't you want this handoff?!".  Might be worth giving folks a n OCM-side way to abort the handoff to silence the alarms if they'd changed their minds, although I guess OCM-UX is out-of-scope here.

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,168 @@
...
+
+TBD
+
+## Implementation History
+
+There are no major milestones in the implementation history.
+
+## Drawbacks
+
+The performance of the OCM API can be a possible drawback.
```
wking, line 163: How often is Insights expected to poll the API?  If it's hourly or slower, it's hard to imagine this being a significant load on the OCM service side.  Are there latency constraints on the UX that make hourly polling for slow responses too slow?  Worth talking about latency constraints or lack thereof in the enhancement?

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: I think most likely alerting....and then maybe an Insights rule

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: Well we have been talking about daily interval AFAIK..... but @abhgupta can share his ideas/concerns

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: I think we won't be able to count the days within the window, but I think we can count the number of unsuccessful responses within the window and then fire an alert when a threshold is reached. WDYT?

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: My understanding is that the API will serve only the updated `pull-secret` (the update is a user action in the OCM IMO) and only within the given window. This should minimize the potential diffs between the secrets content. But yes, I agree that this should be described carefully. 
The only mitigation (in similar disaster scenarios) I can think of is to refer to the documentation and manual steps how to try again or update the secret manually. Are there any better ideas?

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: Yes I think we should say that the OCM API should preserve the backport compatibility and that the new version of the API will likely require an update of the Insights operator code. Would it be sufficient?

iNecas: I agree the probability of the API changing during the lifetime of the OCP version being supported should be very small and we should demand on the OCM API preserve the compatibility through the life-time of the versions depending on that particular API.

I can imagine having additional logic on specific extra status code form the API (let's say 410 Gone) and a follow-up alert from the insights-operator that it's outdated and the user needs to update for it to work properly: but personally, given its probability and the other ways there are in place notifying the user on using out-dated cluster, I would vote for simple logic and just assume the API will work on supported versions.

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: OK so the API will provide time information so we should be able to inform user about the last day or something...

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: OK makes sense. Do we need to update the enhancement to highlight this? @mrunalp can you help herer please.

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: I would resolve this conversation thread, but what about the `Failure Modes` chapter?

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: I updated this chapter. Please check.

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,176 @@
...
+4. The Insights operator calls this API periodically, fetches the updated `pull-secret` and compares it with the current global `pull-secret` on the cluster (only entries available in both pull-secrets are compared).
+5. If there is a mismatch, the Insights operator applies the new `pull-secret`.
+   * Insights operator only replaces the entries provided in the updated `pull-secret` from the OCM API.
+   * Note: MCO updating the `pull-secret` on the nodes will not result in node reboots.
+6. The Insights operator itself starts using the new `pull-secret`
+7. Cluster comes up with the new `pull-secret` and starts reporting back to OCM into the new account.
+
+Note: This process will work only on the clusters where the feature was released as part of the Insights operator and will not be backported to older versions.
+
+### User Stories
```
aravindhp, line 77: You can link to the existing Jira epic / stories here

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Authorization
+
+The Insights Operator is able to pull the updated `pull-secret` content from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config` namespace).
+
+The Insights Operator must provide a cluster ID as an identifier of the cluster.
+
+### OCM API
+
```
aravindhp, line 106: @mrunalp @markturansky should we add API details here once MR 2716 has merged?

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,176 @@
+---
+title: pulling-and-updating-pull-secret-from-ocm
+authors:
+  - "@tremes"
+reviewers:
+  - "@mrunalp"
+  - "@aravindhp"
+  - "@inecas"
+  - "@rawsyntax"
```
aravindhp, line 9: Please @markturansky here as a reviewer.

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,176 @@
...
+
+- `insights-operator-e2e-tests` suite can verify that the Insights Operator is able to download and update the `pull-secret` content
+
+### Graduation Criteria
+
+Thera are no graduation criteria. This feature is planned as a fast track feature.
+
+#### Dev Preview -> Tech Preview
+
+Empty
```
aravindhp, line 133: Nit: `NA` might be more applicable than `Empty`

aravindhp (commented):
Thanks for working on this @tremes

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: I will add the link to the `see also` section at the top

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: I mentioned the docs update requirement

enhancements/insights/pulling-pull-secret-from-ocm.md
wking: b55db2b012f9 looks good to me; thanks :)

enhancements/insights/pulling-pull-secret-from-ocm.md
wking: are we confident in the docs, or do we want a robotic guard in the Insights operator that blocks on and complains about paused pools?  Not clear from this doc is how long the outgoing pull-secret will continue to work.  If we're just passing a cluster between users, sticking with the outgoing secret for a while doesn't seem like a big deal.  If we're rotating because someone leaked the old secret, your paused-pool nodes might be break.  Also not clear to me is how gracefully nodes recover from stale pull secrets; is there anything in reboot/pivot code that could lock up?  Maybe all the in-cluster logic should live in the MCO and be about waving alarms if there's a paused pool and a pending pull-secret change?  Thoughts?

enhancements/insights/pulling-pull-secret-from-ocm.md
wking: is there a clear boundary for "the global pull secret provided by the OCM"?  There's a single global pull secret in-cluster, and the bits we hand out as an OCP pull secret could be some, all, or none of the cluster's current pull secret.  For example, maybe the cluster admin added a docker.io pull secret to the cluster's global secret.  We wouldn't want to clobber that.

enhancements/insights/pulling-pull-secret-from-ocm.md
wking: for Failure Modes we just dream up and discuss as many potential issues as we can think of:

* Some `paused` compute pool fails to pick up the new secret and continues attempting to use the old secret.  No problems if the old secret keeps working, but if it is revoked... FIXME: MCO/workloads folks might be able to clarify what will happen.
* Admin never confirms the handoff.  In this case, the window expires; no issues in-cluster.  If they open a support case "I thought I did transfer that over", support will `${STEPS}` to get logs showing that the transfer was initiated, alerted upon, but never confirmed in cluster. (possibly by visiting Telemetry, or maybe this would be summarized in an Insights rule?)
* Attacker who can forge an X.509 cert and impersonate OCM (or who gains server-side access to OCM) recommends an invalid pull secret.  Do we attempt to guard against this with some sort of pre-apply validation checks?  Do we mark it down as unlikely and have a generic "recovering your cluster from a broken pull secret" section coupled with our "remove `${KEY}` from your pull secret to tell Insights to stop polling upstream" section?
* More?

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: Thanks too.

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: @wking I added the following sentence there:

> Insights operator only replaces the entries provided in the updated `pull-secret` from the OCM API.

It means that the Insights operator cannot add or delete anything from the secret (I can try to emphasize this even more). It can only replace existing entries with new ones and if the entries (entry names) don't match then we should inform use (probably alert?)

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: Yep. Good points. Thanks.

- ad `#1` - We must mention the fact that the old secret will work for some time. Who can help us with the clarification (on the MCO side) please?
- ad `#2` - There's no such confirmation planned. We've discussed that the Insights operator is unable to provide such confirmation, because the "real" update is then up to the MCO. So we rely on Telemetry here (+ optionally an Insights rule)
- ad `#3` - Can we do some pre-apply validation @aravindhp? I don't think so, but it's definitely worth mentioning.

enhancements/insights/pulling-pull-secret-from-ocm.md
markturansky: @aravindhp  yes, and I think that MR can be wrapped up today. I'll be doing a final review with @abhgupta this afternoon.

enhancements/insights/pulling-pull-secret-from-ocm.md
markturansky: This API call will be for exactly 1 row in 1 table with a well-indexed ID field.  It returns an instance of "ClusterTransfer" with details populated or it's empty.  Either way, it should be very fast. This is a scalable solution.

enhancements/insights/pulling-pull-secret-from-ocm.md
wking: Probably also means that no operator besides Insights can use "remove an entry from the pull secret" as a knob for turning things off.

enhancements/insights/pulling-pull-secret-from-ocm.md
wking: I don't think you need to worry about MCO application for handoff confirmation.  The Insights operator knows and will alert on "pending transition waiting for local admin approval" per your existing proposal, and the Insights operator also knows if it accepts the approval and updates the local pull secret (regardless of whether the MCO then rolls that secret out).  So it should be possible to create an Insights rule that points out the most recent handoff attempt and then whether that handoff went through or not.  But I'm fine if the position is that, while a rule is possible, the risk of confusion is small enough that it's not worth writing the rule up, and we can always re-evaluate later if support bumps into this often enough that we want to revisit our guess at the risk size.

enhancements/insights/pulling-pull-secret-from-ocm.md
mrunalp: Yes, sounds good :+1:

enhancements/insights/pulling-pull-secret-from-ocm.md
iNecas: Yes, is there such a workflow today?

The one I know of is disabling the remote health functionality https://docs.openshift.com/container-platform/4.9/support/remote_health_monitoring/opting-out-of-remote-health-reporting.html.

If the remote health functionality is disabled, this workflow would not work at all, so implementaiton-wise probably not a worry, we should however make sure the docs are updated to ensure people realize this functionality would be lost as well.

Are there other cases where people might edit the relevant parts of pull secret outside of the knowledge of OCM?

markturansky:
ClusterTransfer Spec was merged:  https://gitlab.cee.redhat.com/service/uhc-account-manager/-/merge_requests/2716

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,189 @@
...
+
+The entry point to many OpenShift subscriptions is for a user to start with a 60 day trial from the OCM with personal credentials
+and then migrate it over to a company or organization account. In addition, reports from the field indicate that customers in general when transferring ownership, either follow one of the steps in the process and do not follow through with the other step required to complete the transfer. This enhancement will make the transfer frictionless and provide a better user experience.
+
+
+### Goals
+The goal of this enhancement is to describe the API and the interaction between OCM and the cluster that is being transferred to update the `pull-secret` on that cluster:
+
+- Extend the Insights Operator config with a new OCM API endpoint to be able to query the new `pull-secret` content
+- Periodically (once a day) query the new endpoint and if there is a new or updated content then update the `pull-secret`
```
mrunalp, line 54: Customer should be able to trigger a pull secret update faster if they want to rather than having to wait for the full poll period (like 24 hours).

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,189 @@
...
+
+There are no major milestones in the implementation history.
+
+## Drawbacks
+
+The performance of the OCM API can be a possible drawback.
+
+## Alternatives
+
+- In the future, we may have more flow of data from the OCM API to a cluster and we will revisit if we need a dedicated operator for such flows. However, we do not want to add an entirely new operator for the `pull-secret` flow today.
```
wking, line 188: While poking around in GitHub today, I stumbled across [ocm-agent][1].  Possibly worth talking to those folks about scoping and timing to see if you want to revisit this alternative or not.

[1]: https://github.com/openshift/ocm-agent

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: Yes. Thanks. I hopefully mentioned all the points here. Resolving the conversation.

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: We discussed this yesterday and I updated the enhancement.

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: I mentioned the `ClusterTransfer` API here and added the link to the MR.

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: I added one sentence describing briefly how to do this right now, but we will need to document it as well. I guess there will be a new documentation section/chapter related to this feature in the Insights operator.

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: I updated this section.

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: It's here https://github.com/openshift/enhancements/pull/971/files#diff-fb03640aacfca46208c1d4f791cc3fa8ad33e6644e236b9ddfac8a527296914cR54

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,188 @@
...
+
+The entry point to many OpenShift subscriptions is for a user to start with a 60 day trial from the OCM with personal credentials
+and then migrate it over to a company or organization account. In addition, reports from the field indicate that customers in general when transferring ownership, either follow one of the steps in the process and do not follow through with the other step required to complete the transfer. This enhancement will make the transfer frictionless and provide a better user experience.
+
+
+### Goals
+The goal of this enhancement is to describe the API and the interaction between OCM and the cluster that is being transferred to update the `pull-secret` on that cluster:
+
+- Extend the Insights Operator config with a new OCM API endpoint to be able to query the new `pull-secret` content
+- Periodically (once a day) query the new endpoint and if there is a new or updated content then update the `pull-secret`. User can trigger the update faster by deleting the Insights operator Pod (This needs to be documented and may change in the future - e.g we can introduce a new CRD to configure this).
```
aravindhp, line 54: >User can trigger the update faster by deleting the Insights operator Pod (This needs to be documented and may change in the future - e.g we can introduce a new CRD to configure this).

I suggest calling this out under `Risks and Mitigations`

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,188 @@
...
+   * The updated `pull-secret` from the OCM API is an authoritative source. Insights operator replaces and adds (There should not be any new entries, but this may change in the future) the entries provided in the updated `pull-secret` from the OCM API (This means that the Insights operator cannot delete anything in the existing secret).
+   * Note: MCO updating the `pull-secret` on the nodes will not result in node reboots. This will require an update of the documentation in https://docs.openshift.com/container-platform/4.9/updating/update-using-custom-machine-config-pools.html#update-using-custom-machine-config-pools-about_update-using-custom-machine-config-pools.
+   We should emphasize the need of unpausing the machine config pools to succesfully apply the new `pull-secret` (credentials) to all the nodes.
+6. The Insights operator itself starts using the new `pull-secret`
+7. Cluster comes up with the new `pull-secret` and starts reporting back to OCM into the new account.
+
+Note: This process will work only on the clusters where the feature was released as part of the Insights operator and will not be backported to older versions.
```
aravindhp, line 77: I suggest making this a non-gaol.

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,188 @@
...
+
+### User Stories
+
+As an OCM user, I would like to transfer cluster ownership from my account to another user's account in one step.
+
```
aravindhp, line 82: You can also link to the stories here so that it is clearly called out:
```suggestion
- [CCXDEV-6529 Insights Operator pulling updated pull-secret to transfer unclassified cluster](https://issues.redhat.com/browse/CCXDEV-6529)
- [CCX-185 Transfer of Unclassified clusters - provide pull-secret](https://issues.redhat.com/browse/CCX-185)

File: enhancements/insights/pulling-pull-secret-from-ocm.md [hunk 17]
```diff
@@ -0,0 +1,188 @@
...
+Risk: The OCM API is down or returning HTTP response code >= 500.
+
+Mitigation: The Insights Operator is marked as degraded in such case.
+
+Risk: The `pull-secret` does not work or is otherwise corrupted.
+
+Mitigation: We must refer to the OCP documentation, which describes the manual fix procedure.
+
+
```
aravindhp, line 97:
```suggestion
Risk: Users don't want to wait 24 hours for the pull secret to be updated on the cluster

Mitigation: User can trigger the update faster by deleting the Insights operator Pod (This needs to be documented and may change in the future - e.g we can introduce a new CRD to configure this).

```

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,188 @@
...
+N/A
+
+#### Tech Preview -> GA
+
+N/A
+
+#### Removing a deprecated feature
+
+Removing this feature will require updating the Insights operator code base.
```
aravindhp, line 147: This can also be N/A. The section is mainly for enhancements that involves removing a deprecated feature.

enhancements/insights/pulling-pull-secret-from-ocm.md
aravindhp: Thanks for pointing this out. The OCM folks are reviewing this enhancement, so they can comment. cc: @markturansky

aravindhp (changes requested):
@tremes this looks really good. Just a few comments.

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: OK. Thanks. Done

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: Moved. Thanks

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: OK. Removed. Thank you.

enhancements/insights/pulling-pull-secret-from-ocm.md
markturansky: sounds good. future requirements TBD.

aravindhp (approved):
Thanks for working on this, @tremes 

/approve

@markturansky please review and LGTM

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,191 @@
...
+This feature will not work on disconnected clusters because there is no corresponding token (in the `pull-secret`) to communicate with Red Hat. We should emphasize this in https://docs.openshift.com/container-platform/4.9/support/remote_health_monitoring/opting-out-of-remote-health-reporting.html.
+
+
+### Test Plan
+
+- `insights-operator-e2e-tests` suite can verify that the Insights Operator is able to download and update the `pull-secret` content
+
+### Graduation Criteria
+
+Thera are no graduation criteria. This feature is planned as a fast track feature.
```
mrunalp, line 138: nit: typo `There`

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,191 @@
...
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+There is no upgrade/downgrade strategy needed.
+
+### Version Skew Strategy
+
+The Inisghts operator expects the `pull-secret` in the `openshift-config` namespace. If there are any changes to the `pull-secret` location or content, then the update of the Insights operator will be required.
```
mrunalp, line 158: typo: `Insights`

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,191 @@
...
+
+#### Failure Modes
+
+There is no plan to provide direct confirmation of a successful `pull-secret` update. The problem is that the Insights operator is unable to provide such confirmation, because the update is followed up by the MachineConfig operator work. Some potential failures/problems are:  
+
+- There are some paused `MachineConfigPools`, which failed to pick up the updated `pull-secret` definition. The old original secret will
+still be valid for some period of time, but this can happen regardless of the period. We can create a new Insights rule (checking paused MCPs) to mitigate this.
+
+- Attacker who can forge an X.509 cert and impersonate OCM (or who gains server-side access to OCM) recommends an invalid pull secret.
+  This is basically same situation as described in [Risks and mitigations](#risks-and-mitigations) chapter above. There was a discussion about some pre-apply validation check of the new secret on the cluster side, but the output was that we can't do much about it (in the Insights operator) and we may revisit this in the future.
```
mrunalp, line 174: chapter -> section?

enhancements/insights/pulling-pull-secret-from-ocm.md
```diff
@@ -0,0 +1,191 @@
...
+### Authorization
+
+The Insights Operator is able to pull the updated `pull-secret` content from the OCM API using the existing `cloud.openshift.com` token
+available in the `pull-secret` (in the `openshift-config` namespace).
+
+The Insights Operator must provide a cluster ID as an identifier of the cluster.
+
+### OCM API
+
+The OCM API provides an endpoint that will serve the updated `pull-secret` definition when the migration window is opened. The endpoint will be defined in a new `ClusterTransfer` API - more details are in https://gitlab.cee.redhat.com/service/uhc-account-manager/-/merge_requests/2716.
```
markturansky, line 112: you can change this link from MR to the doc itself:  https://gitlab.cee.redhat.com/service/uhc-account-manager/-/blob/master/docs/cluster_transfer_api.md

markturansky:
LGTM but for one out-of-date link

enhancements/insights/pulling-pull-secret-from-ocm.md
tremes: Updated. Thanks

PR #974, opened 2021-12-03
dev-guide/cluster-version-operator/dev: Doc TechPreview feature gate annotations

wking:
To make them more discoverable.


LalatenduMohanty (changes requested):
Techpreview should also be a separate question/topic. e.g. How to include manifests in CVO which are in techpreview?

wking:
I've added subsections for the two inclusion categories with 2f4757c8 -> 3d4a9f6b.

LalatenduMohanty (approved):
/lgtm

PR #975, opened 2021-12-03
Move NodeSwap to TechPreviewNoUpgrade feature set

ehashman:
Per @deads2k requested API changes.

/cc @rphillips @mrunalp @deads2k


deads2k:
This is in keeping with how we manage our featuregates. Rather than have a customer individually enable them, we curate a set of techpreview featuregates.  This prevents a proliferation of featuresets.

Nearly every component raises the, "what if a customer just wants *my* featuregate", but from a practical perspective supporting a configuration like that expands our test matrix and in every case so far it has not actually been necessary to provide an individual featuregate enablement.  In addition, you benefit from the the techpreview CI by doing this instead of having to create variants of every job for your node-swap setting.

/lgtm

mrunalp (approved):
/approve

PR #978, opened 2021-12-07
CONVENTIONS: additional HA considerations regarding storage

jan--f:
Signed-off-by: Jan Fajerski <jfajersk@redhat.com>

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -264,6 +264,11 @@ Please note that in case of [Single Node OpenShift](https://github.com/openshift
   * Two replicas
   * Hard pod [anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node) on hostname (no two pods on the same node)
   * Use the maxUnavailable rollout strategy on deployments (prefer 25% by default for a value)
+  * If the operator or operand requires persistent storage
+    * Operators should keep their state entirely in CustomResources, using PV's is discouraged.
+    * If and how persistent storage impacts high availability of an operand depends on the cluster configuration, the deployment resource used (e.g. Deployment vs StatefulSet) and available storage classes, e.g. whether [topology-aware volume provisioning](https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode) is available.
+    * Operands that are deployed as a StatefulSet, i.e. where replicas get a separate volume instead of all replicas sharing a volume, typically require code to sync state between replicas causing cross-zone traffic, which may be charged.
+    * Consider that even with a StatefulSet and node-level anti-affinity, persistent storage can create a single point of failure by either provisioning volumes in the same availability zone or preventing some replicas from getting scheduled (e.g. the scheduler finds one good node permutation, but one node can't access sufficient free storage).
   * This is the recommended approach for all the components unless the component needs 3 replicas
 * If the operator or operand requires >= 3 replicas and should be running on worker nodes
   * Set soft pod anti-affinity on the hostname so that pods prefer to land on separate nodes (will be violated on two node clusters)


jan--f:
From our recent experience of moving CMO components from soft to hard anti-affinity, we learned that persistent storage can play a role when one follows the HA conventions. This is a first attempt to incorporate this experience.

Much more could be said of course. For a first attempt I opted to keep it brief.

jan--f:
cc @simonpasquier

CONVENTIONS.md
```diff
@@ -264,6 +264,9 @@ Please note that in case of [Single Node OpenShift](https://github.com/openshift
   * Two replicas
   * Hard pod [anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node) on hostname (no two pods on the same node)
   * Use the maxUnavailable rollout strategy on deployments (prefer 25% by default for a value)
+  * If the the operator or operand requires persistent storage
+    * Prefer to use topology-aware volume provisioning, i.e. prefer using a StorageClass with `volumeBindingMode: WaitForFirstConsumer` so that volume binding is delayed until the pod is scheduled (taking anit-affinty into account)
```
simonpasquier, line 268: In the monitoring operator case, it can't really check or enforce that the requested storage class specifies `olumeBindingMode: WaitForFirstConsumer`. Are we saying that the operator should do something (warn? fail?) when it detects that the condition isn't fulfilled?

CONVENTIONS.md
```diff
@@ -264,6 +264,9 @@ Please note that in case of [Single Node OpenShift](https://github.com/openshift
   * Two replicas
   * Hard pod [anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node) on hostname (no two pods on the same node)
   * Use the maxUnavailable rollout strategy on deployments (prefer 25% by default for a value)
+  * If the the operator or operand requires persistent storage
```
simonpasquier, line 267:
```suggestion
  * If the operator or operand requires persistent storage
```

CONVENTIONS.md
```diff
@@ -264,6 +264,9 @@ Please note that in case of [Single Node OpenShift](https://github.com/openshift
...
+    * Prefer to use topology-aware volume provisioning, i.e. prefer using a StorageClass with `volumeBindingMode: WaitForFirstConsumer` so that volume binding is delayed until the pod is scheduled (taking anit-affinty into account)
+    * Consider that even with node-level anti-affinity and topology-aware volume provisioning, PVs can be provisioned in the same availability zones (creating a single point of failure) or prevent some replicas from getting scheduled (e.g. the scheduler finds one good node permutation, but one node can't access sufficient free storage)
```
simonpasquier, line 269:
```suggestion
    * Consider that even with node-level anti-affinity and topology-aware volume provisioning, PVs can be provisioned in the same availability zone (creating a single point of failure) or prevent some replicas from getting scheduled (e.g. the scheduler finds one good node permutation, but one node can't access sufficient free storage)
```

CONVENTIONS.md
jan--f: Yeah I wasn't sure how to phrase this. Or maybe this should be added somewhere else?
From what I understand, we want to inform users (who have control over which `StorageClass` is used) what impact their choices can have and what works best with the HA conventions.

Maybe its better to note here that storage HA is left to the storage provider (like we do in the [documentation](https://docs.openshift.com/container-platform/4.9/storage/understanding-persistent-storage.html) already)? I think it would still be useful to outline the possible effects. This seems like a good place in terms of visibility, but I agree its somewhat apart from the rest of the content.

simonpasquier: > From what I understand, we want to inform users (who have control over which StorageClass is used) what impact their choices can have and what works best with the HA conventions.

This document is targeted towards OpenShift developers, not users. In practice, someone should be able to configure the monitoring operands with static storage resources if their storage provider doesn't support `WaitForFirstConsumer` binding.

jan--f: Yeah I was trying to phrase if openly enough, but point taken. What do you think of the current iteration?

CONVENTIONS.md
```diff
@@ -264,6 +264,9 @@ Please note that in case of [Single Node OpenShift](https://github.com/openshift
   * Two replicas
   * Hard pod [anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node) on hostname (no two pods on the same node)
   * Use the maxUnavailable rollout strategy on deployments (prefer 25% by default for a value)
+  * If the operator or operand requires persistent storage
+    * Consider that even with node-level anti-affinity, persistent storage can create a single point of failure by either provisioning volumes in the same availability zone or preventing some replicas from getting scheduled (e.g. the scheduler finds one good node permutation, but one node can't access sufficient free storage)
```
dgrisonnet, line 268: > Consider that even with node-level anti-affinity, persistent storage can create a single point of failure

I wouldn't phrase it that way since even without persistent storage, node-level anti-affinity might still cause SPOF since nodes can be in the same AZ

The issue that we want to highlight here is when the `operator or operand with persistent storage` is migrated from soft anti-affinity to hard anti-affinity as it can cause scheduling issues. We could link the post mortem that was written in the past to detail the problem. In that case, the best we can recommend is to have the operator handle the rebalancing.

jan--f: > I wouldn't phrase it that way since even without persistent storage, node-level anti-affinity might still cause SPOF since nodes can be in the same AZ

Fair point, then intention here was to highlight that even with good pod-level affinity specs, persistent storage can create a SPOF. Will rephrase.

> The issue that we want to highlight here is when the `operator or operand with persistent storage` is migrated from soft anti-affinity to hard anti-affinity as it can cause scheduling issues. We could link the post mortem that was written in the past to detail the problem. In that case, the best we can recommend is to have the operator handle the rebalancing.

My understanding is that this is not just an artifact of moving from soft to hard anti-affinity. Even starting with a hard affinity spec can end up in the situation outlined above, can't it?

dgrisonnet: Ah yes, you are correct. The issue happens when there is more than 1 replica before the migration to hard anti-affinity.

wking (approved):

CONVENTIONS.md
ingvagabund: > when the operator or operand with persistent storage is migrated from soft anti-affinity to hard anti-affinity as it can cause scheduling issues

From the scheduler POV this situation is specific to the volume plugins and how a new storage piece/claim is requested on a new node. Assuming there are two replicas with a persistent storage on the same node, after migrating to the hard anti-affinities, one of the persistent storage pieces will need to be deleted and created on a different node. I am unaware  of how this workflow is expected to proceed. Worth asking the storage folks for the details.

jsafrane: I think it's nearly impossible to run an *operator* in HA with persistent storage. Below we require "an operator deployed by the CVO should run on control plane nodes". We typically run each master in a different availability zone and a storage volume can be usually available only in a single zone. With Deployment used for the operator and hard pod anti-affinity, the operator pods must run in different AZs and one of the replicas will not be able to mount the storage and hence it won't run at all. With StatefulSet, each replica will get its own volume and can run, but then they require some code in the operator to sync storage of each replica so both replicas have the same data. This code is usually not trivial.

I would strongly recommend *operators* not to require any PVs and store all their state in CustomResources. The API server will handle HA of this data.

For *operands*, the text makes sense. But it has the same issue - the operand must have some protocol to sync data between different PVs. Which is typical for databases running in some HA mode. In addition, depending on the underlying cloud, they might want to add anti-affinity rules to their pods to run each operand replica in a different zone. Such application could be tolerant to catastrophic failure of a single zone, on the other hand it would increase cross-zone traffic, which can be charged extra.

gnufied: >  With Deployment used for the operator and hard pod anti-affinity, the operator pods must run in different AZs and one of the replicas will not be able to mount the storage and hence it won't run at all

I guess with the exception of RWX volume types, this is mostly true. A RWX obviously implies that it has inbuilt mechanism to sync update between different highly available pods.

jan--f: @jsafrane You are correct on all point.
Prometheus has a somewhat rare HA setup where the replicas do not need synchronization, i.e. run with separate PVs. As mentioned in https://github.com/openshift/enhancements/pull/978#discussion_r781989100 we still wrote quite a bit of code to handle this when switching from soft to hard anti-affinity.

This PR intends to raise awareness about the impact that persistent storage can have on HA setups, _not_ to provide the solution.

jsafrane: Therefore the guide should say that it's not recommended to use PVs for **operators**, AFAIK no OCP operator uses PVs (they may set up PVCs for the operands, of course)

For operands, the text is sort of OK and pod scheduling requirements (anti-affinity) and storage (availability zones, local storage) must be taken into account.

jan--f: Ahh yes fair point re. operators vs operands. I'll update accordingly, thank you.

jan--f: @jsafrane I updated the PR a bit, would appreciate if you could take another look.

CONVENTIONS.md
```diff
@@ -264,6 +264,11 @@ Please note that in case of [Single Node OpenShift](https://github.com/openshift
   * Two replicas
   * Hard pod [anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node) on hostname (no two pods on the same node)
   * Use the maxUnavailable rollout strategy on deployments (prefer 25% by default for a value)
+  * If the operator or operand requires persistent storage
+    * Operators should keep their state entirely in CustomResources, using PV's is discouraged.
+    * If and how persistent storage impacts high availability of an operand depends on the cluster configuration, the deployment resource used (e.g. Deployment vs StatefulSet) and available storage classes, e.g. whether [topology-aware volume provisioning](https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode) is available
+    * Operands that are deployed as a StatefulSet, i.e. where replicas get a separate volume instead of all replicas sharing a volume, typically require code to sync state between replicas. THis can cause significant cross-zone traffic.
```
simonpasquier, line 270:
```suggestion
    * Operands that are deployed as a StatefulSet, i.e. where replicas get a separate volume instead of all replicas sharing a volume, typically require code to sync state between replicas causing cross-zone traffic which may be charged.
```
wouldn't the same warning apply when multiple replicas in different AZs read from a single volume?

CONVENTIONS.md
```diff
@@ -264,6 +264,11 @@ Please note that in case of [Single Node OpenShift](https://github.com/openshift
   * Two replicas
   * Hard pod [anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node) on hostname (no two pods on the same node)
   * Use the maxUnavailable rollout strategy on deployments (prefer 25% by default for a value)
+  * If the operator or operand requires persistent storage
+    * Operators should keep their state entirely in CustomResources, using PV's is discouraged.
+    * If and how persistent storage impacts high availability of an operand depends on the cluster configuration, the deployment resource used (e.g. Deployment vs StatefulSet) and available storage classes, e.g. whether [topology-aware volume provisioning](https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode) is available
```
simonpasquier, line 269:
```suggestion
    * If and how persistent storage impacts high availability of an operand depends on the cluster configuration, the deployment resource used (e.g. Deployment vs StatefulSet) and available storage classes, e.g. whether [topology-aware volume provisioning](https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode) is available.
```

CONVENTIONS.md
```diff
@@ -264,6 +264,11 @@ Please note that in case of [Single Node OpenShift](https://github.com/openshift
...
+    * Operands that are deployed as a StatefulSet, i.e. where replicas get a separate volume instead of all replicas sharing a volume, typically require code to sync state between replicas. THis can cause significant cross-zone traffic.
+    * Consider that even with a StatefulSet and node-level anti-affinity, persistent storage can create a single point of failure by either provisioning volumes in the same availability zone or preventing some replicas from getting scheduled (e.g. the scheduler finds one good node permutation, but one node can't access sufficient free storage)
```
simonpasquier, line 271:
```suggestion
    * Consider that even with a StatefulSet and node-level anti-affinity, persistent storage can create a single point of failure by either provisioning volumes in the same availability zone or preventing some replicas from getting scheduled (e.g. the scheduler finds one good node permutation, but one node can't access sufficient free storage).
```

CONVENTIONS.md
jsafrane: lgtm, I can see @simonpasquier found the same nits that I would report too.

CONVENTIONS.md
jan--f: If two replicas can mount the same volume, they are both in the AZ of the volume iiuc.

dhellmann:
/test markdownlint

I think the linter job issue was fixed with #1011

jan--f:
@jsafrane ptal

PR #980, opened 2021-12-08
enhancements/authentication: detect invalid certificates

s-urbaniak:
/cc @sttts @stlaz @deads2k

### DIFF
--- a/enhancements/authentication/invalid-certs.md
+++ b/enhancements/authentication/invalid-certs.md
@@ -1,0 +1,204 @@
+---
+title: invalid-certs
+authors:
+  - "@s-urbaniak"
+reviewers:
+  - "@sttts"
+  - "@stlaz"
+  - "@deads2k"
+approvers:
+  - "@mfojtik"
+creation-date: 2021-12-08
+last-updated: 2022-03-10
+---
+
+# Detecting invalid HTTPS server certificates
+
+## Summary
+
+OpenShift 4.10 is going to be rebased against Kubernetes 1.23. This requires using Go 1.17.
+However, starting with Go 1.17 the support for the CommonName field of server HTTPS certificates is going to be removed.
+
+Formally, the temporary `GODEBUG=x509ignoreCN=0` flag [has been removed](https://go.dev/doc/go1.17#crypto/x509).
+This implies that starting from OpenShift 4.10 invalid certificates will not be trusted any more as they will fail verification.
+
+Example:
+
+Given the following certificate:
+```plaintext
+Certificate:
+    Data:
+        ...
+        Subject: CN=foo-domain.com
+        X509v3 extensions:
+            X509v3 Key Usage: critical
+                Digital Signature, Key Encipherment
+            X509v3 Extended Key Usage: 
+                TLS Web Server Authentication
+            X509v3 Basic Constraints: critical
+                CA:FALSE
+```
+
+Verification against the `foo-domain.com` hostname of such certificate will fail with the following error in Go 1.17:
+```plaintext
+x509: certificate relies on legacy Common Name field, use SANs instead
+```
+
+Verification of server certificates is executed during the TLS Handshake procedure.
+A TLS (https) client observing an invalid certificate will reject the connection attempt.
+
+Cluster internal issued certificates are not affected,
+however custom certificates can be configured in various cases:
+- custom serving certificates for kube-apiserver
+- custom API webhooks
+- custom aggregated API endpoints
+- custom certificates for route endpoints
+- certificates of external auth identity providers
+
+This will lead to broken connections to critical core parts of OpenShift and thus to a degraded cluster if invalid custom certificates are configured.
+
+This enhancement suggests a solution to prevent core workloads involving
+invalid certificates, functional on OpenShift 4.9, to break upon the upgrade to
+OpenShift 4.10. It therefore proposes:
+- a way to detect invalid certificates in kube-apiserver and oauth-server
+- a way to prevent upgrades from OpenShift 4.9 to OpenShift 4.10 in the face of invalid certificates
+
+## Motivation
+
+Broken workloads must be prevented due to invalid certificates
+when upgrading from OpenShift 4.9 to OpenShift 4.10.
+
+### Goals
+
+A means to detect invalid certificates for TLS connections between:
+- kube-apiserver and API webhooks
+- kube-apiserver and aggregated API endpoints
+- custom serving certificates for kube-apiserver
+- oauth-server and its external route
+- oauth-server and external identity providers
+
+### Non-Goals
+
+* Other core workloads are out of scope for this enhancement.
+* Alerting, although appropriate and desirable, is left outside the scope of
+  this very enhancement.
+
+## Proposal
+
+### User Stories
+
+1. As a cluster admin I want to be prevented from upgrading OpenShift 4.9 to OpenShift 4.10
+if I configured invalid certificates.
+
+2. As a cluster admin I want to get feedback in OpenShift 4.9 if invalid certificates are being detected.
+
+### API Extensions
+
+N/A
+
+### Detecting invalid certificates
+
+#### kube-apiserver
+
+In Kubernetes detection of invalid certificates has been added as part of https://github.com/kubernetes/kubernetes/pull/95396.
+The introduced Prometheus `apiserver_webhooks_x509_missing_san_total` and `apiserver_kube_aggregator_x509_missing_san_total` metrics provides the means to detect invalid certificates against API webhooks and aggregated API endpoints.
+
+To detect invalid certificates this enhancement proposes to add a new controller in `cluster-kube-apiserver-operator` that:
+
+1. Regularly executes queries against the cluster internal Thanos instance
+2. Verifies if above metrics have a value >0
+3. If invalid certificates are detected, the controller sets a `InvalidCertsUpgradeable=false` condition on the `cluster` `KubeAPIServer.operator.openshift.io` resource which will yield the `Upgradeable=false` status on the `kube-apiserver` `config.openshift.io.ClusterOperator` resource.
+
+#### oauth-server
+
+For oauth-server this enhancement proposes the introduction of the following new metric:
+
+- `openshift_auth_external_x509_missing_san_total`: a metric capturing the count of invalid certificates
+presented by an external identity provider.
+
+To detect invalid certificates this enhancement proposes to add a new controller in `cluster-authentication-operator` that:
+
+1. Regularly executes queries against the cluster internal Thanos instance
+2. Verifies if above metrics have a value >0
+3. If invalid certificates are detected, the controller sets a `InvalidCertsUpgradeable=false` condition on the `cluster` `Authentication.operator.openshift.io` resource which will yield the `Upgradeable=false` status on the `authentication` `config.openshift.io.ClusterOperator` resource.
+
+#### cluster-authentication-operator
+
+In case of having invalid custom certificates configured for the external OAuth route
+the existing `RouterCertsDomainValidationController` is extended.
+Formally additional logic is added in `validateRouterCertificates` https://github.com/openshift/cluster-authentication-operator/blob/ff156ab2bdfbdd68b49d76547fea5ec28d9c3639/pkg/controllers/routercerts/controller.go#L129.
+
+Additionally, custom routes controller in `openshift-authentication-operator` will reject invalid certificates in https://github.com/openshift/cluster-authentication-operator/blob/d2f6218a6ab2daccbec43f71a495de1c80529ef6/pkg/controllers/customroute/custom_route_controller.go#L55.
+
+#### library-go
+
+A library-go function will be implemented that takes a `*x509.Certificate` and returns a `bool` value
+indicating if it is an invalid certificate. The function will check whether the given certificate relies
+on the Common Name field rather than SANs.
+
+This function will be used to assert validity of invalid certificates in the external OAuth route case.
+
+The higher level controller which asserts metrics for a given Prometheus client will be provided in libary-go.
+
+### Risks and Mitigations
+
+This enhancement provides a means to protect against broken clusters when upgrading from OpenShift 4.9 to OpenShift 4.10.
+
+## Design Details
+
+### Open Questions [optional]
+
+N/A
+
+### Test Plan
+
+N/A
+
+### Graduation Criteria
+
+N/A
+
+#### Dev Preview -> Tech Preview
+
+N/A
+
+#### Tech Preview -> GA
+
+N/A
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+The above proposed changes **must** be backported to OpenShift 4.9.
+Otherwise, a detection of invalid certificates will not be possible before an upgrade to OpenShift 4.10.
+
+### Version Skew Strategy
+
+N/A
+
+### Operational Aspects of API Extensions
+
+N/A
+
+#### Failure Modes
+
+N/A
+
+#### Support Procedures
+
+N/A
+
+## Implementation History
+
+N/A
+
+## Drawbacks
+
+N/A
+
+## Alternatives
+
+N/A


enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,152 @@
...
+when upgrading from OpenShift 4.9 to OpenShift 4.10.
+
+### Goals
+
+A means to detect invalid certificates for TLS connections between:
+- kube-apiserver and API webhooks
+- kube-apiserver and aggregated API endpoints
+- oauth-server and its external route
+- oauth-server and external identity providers
+
```
sttts, line 44: custom serving certs for kube-apiserver

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,152 @@
...
+#### oauth-server
+
+For oauth-server we propose the introduction of the following new metric:
+
+- `openshift_auth_external_x509_missing_san_total`: a metric capturing the count of invalid certificates
+presented by an external identity provider.
+
+In case of having invalid custom certificates configured for the external OAuth route
+we propose to extend the existing `RouterCertsDomainValidationController`.
+Formally we propose to add additional logic in `validateRouterCertificates` https://github.com/openshift/cluster-authentication-operator/blob/ff156ab2bdfbdd68b49d76547fea5ec28d9c3639/pkg/controllers/routercerts/controller.go#L129.
```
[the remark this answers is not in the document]
stlaz: You should also extend the https://github.com/openshift/cluster-authentication-operator/blob/d2f6218a6ab2daccbec43f71a495de1c80529ef6/pkg/controllers/customroute/custom_route_controller.go#L55 to not accept "invalid" certificates

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,152 @@
...
+
+#### library-go
+
+The above mentioned logic to validate a `*x509.Certificate` will be provided as a library-go function.
```
stlaz, line 86: Which logic? What does it do? Maybe extend it in this section a bit.

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,152 @@
...
+---
+
+# Detecting invalid CA certificates
+
+## Summary
+
+OpenShift 4.10 is going to be rebased against Kubernetes 1.23. This requires using Go 1.17.
+However, starting with Go 1.17 support for legacy invalid is going to be removed, see https://go.dev/doc/go1.17.
+More formally, the temporary `GODEBUG=x509ignoreCN=0` flag has been removed.
+This implies that starting from OpenShift 4.10 invalid certificates will not be trusted any more
```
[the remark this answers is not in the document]
ibihim: I don't know if it is just me, but the first thing that pops up in my head is: why did OpenShift use such in the first place. Maybe its worth to mention.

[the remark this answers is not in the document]
s-urbaniak: i don't have the historical context on why that env var was used. also, we don't want to solve all the bits in openshift here, just the api/oauth workload.

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,152 @@
...
+
+A means to detect invalid certificates for TLS connections between:
+- kube-apiserver and API webhooks
+- kube-apiserver and aggregated API endpoints
+- oauth-server and its external route
+- oauth-server and external identity providers
+
+### Non-Goals
+
+Other core workloads are out of scope for this enhancement.
```
bparees, line 47: do we at least know what the other workloads are that are likely impacted by this?

does every component that exposes a TLS metrics endpoint need to do work, for example?

enhancements/authentication/invalid-certs.md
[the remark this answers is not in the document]
stlaz: The env var was added to Golang exactly for the reason to buy people time to deal with the situation and prevent products from breaking.

enhancements/authentication/invalid-certs.md
s-urbaniak: @sttts do you have a place where I can hook into the verification of those? Currently, I think we'd need a separate controller in `cluster-kube-apiserver-operator` that checks `config.openshift.io.APIServer#servingInfo.namedCertificates`, parses them, and verifies for invalid certs.

sttts: I believe @sanchezl worked on that. He can link the code here.

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,152 @@
...
+More formally, the temporary `GODEBUG=x509ignoreCN=0` flag has been removed.
+This implies that starting from OpenShift 4.10 invalid certificates will not be trusted any more
+as they will fail verification.
+
+This enhancement suggests a solution to prevent broken workloads from OpenShift 4.9 to OpenShift 4.10
+if invalid certificates are detected.
+
+This enhancment therefore proposes:
+- a way to detect invalid certificates in kube-apiserver and oauth-server
+- a way to prevent upgrades from OpenShift 4.9 to OpenShift 4.10 in the face of invalid certificates
```
wking, line 30: are we planning to backport an `Upgradeable=False` guard for this, or rely on the alert, or start with the alert and maybe follow up with `Upgradeable=False` if we see a lot of alerts, or...?

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,152 @@
...
+last-updated: 2022-12-08
+---
+
+# Detecting invalid CA certificates
+
+## Summary
+
+OpenShift 4.10 is going to be rebased against Kubernetes 1.23. This requires using Go 1.17.
+However, starting with Go 1.17 support for legacy invalid is going to be removed, see https://go.dev/doc/go1.17.
+More formally, the temporary `GODEBUG=x509ignoreCN=0` flag has been removed.
```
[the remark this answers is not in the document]
s-urbaniak: fixed

enhancements/authentication/invalid-certs.md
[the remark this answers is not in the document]
s-urbaniak: I added a concrete example in openssl text format.

enhancements/authentication/invalid-certs.md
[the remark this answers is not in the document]
s-urbaniak: augmenting https://github.com/openshift/cluster-authentication-operator/blob/d2f6218a6ab2daccbec43f71a495de1c80529ef6/pkg/controllers/customroute/custom_route_controller.go#L55 is a good idea.

However we have to be careful: this concrete logic must not be backported. If users configured legacy certificates in the previous releases we have to block upgrades.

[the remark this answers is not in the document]
s-urbaniak: > a degraded condition on an "invalid" certificate?

most likely yes, a degraded condition of i.e. authentication-operator will appear, as the connection will fail to verify.

enhancements/authentication/invalid-certs.md
s-urbaniak: fixed

enhancements/authentication/invalid-certs.md
s-urbaniak: I am not aware that all components expose a TLS metric like this. Our finding refers concretely to a contribution that was made upstream to Kubernetes.

An initial search on the openshift org reveals quite a few usages: https://github.com/search?q=org%3Aopenshift+GODEBUG%3Dx509ignoreCN%3D0&type=code

maybe @eparis has more insight as I believe he had touch points with the `GODEBUG=x509ignoreCN=0` setting.

enhancements/authentication/invalid-certs.md
s-urbaniak: @wking the suggestion is to backport an `Upgradeable=False` guard for this. An alert would be feasible if, after the upgrade, would be functional. However, in the presence of invalid certificates, a cluster will simply be dysfunctional after the upgrade to 4.10.

s-urbaniak:
Also submitted https://bugzilla.redhat.com/show_bug.cgi?id=2031839 to track urgency with respect to 4.10.

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,200 @@
...
+            X509v3 Basic Constraints: critical
+                CA:FALSE
+```
+
+Verification against the `foo-domain.com` hostname of such certificate will fail with the following error in Go 1.17:
+```
+x509: certificate relies on legacy Common Name field, use SANs instead
+```
+
+Verification of server certificates is executed during TLS client handshakes,
```
slaskawi, line 46: You probably mean during "TLS Handshake procedure" or when a client verified "Server Hello".

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,200 @@
...
+N/A
+
+### Detecting invalid certificates
+
+#### kube-apiserver
+
+In Kubernetes detection of invalid certificates has been added as part of https://github.com/kubernetes/kubernetes/pull/95396.
+The introduced Prometheus `apiserver_webhooks_x509_missing_san_total` and `apiserver_kube_aggregator_x509_missing_san_total` metrics provides the means to detect invalid certificates against API webhooks and aggregated API endpoints.
+
+To detect invalid certificates this enhancement proposes to add a new controller in `cluster-kube-apiserver` that:
```
soltysh, line 102: cluster-kube-apiserver-operator

soltysh (approved):
One nit, rest lgtm

enhancements/authentication/invalid-certs.md
deads2k: > a way to prevent upgrades from OpenShift 4.9 to OpenShift 4.10 in the face of invalid certificates
For cases where the platform is a client, this is probably valid.  For cases where the platform is a server, but not a client, I think it's too strong to block upgrades. Can you separate the cases?

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,200 @@
...
+
+1. Regularly executes queries against the cluster internal Thanos instance
+2. Verifies if above metrics have a value >0
+3. If invalid certificates are detected, the controller sets a `InvalidCertsUpgradeable=false` condition on the `cluster` `KubeAPIServer.operator.openshift.io` resource which will yield the `Upgradeable=false` status on the `kube-apiserver` `config.openshift.io.ClusterOperator` resource.
+
+#### oauth-server
+
+For oauth-server this enhancement proposes the introduction of the following new metric:
+
+- `openshift_auth_external_x509_missing_san_total`: a metric capturing the count of invalid certificates
```
deads2k, line 112: What clients of this external server do we have?

deads2k: > What clients of this external server do we have?
If it is just the operator, could the operator build a custom verifier for the check?

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,200 @@
...
+### API Extensions
+
+N/A
+
+### Detecting invalid certificates
+
+#### kube-apiserver
+
+In Kubernetes detection of invalid certificates has been added as part of https://github.com/kubernetes/kubernetes/pull/95396.
+The introduced Prometheus `apiserver_webhooks_x509_missing_san_total` and `apiserver_kube_aggregator_x509_missing_san_total` metrics provides the means to detect invalid certificates against API webhooks and aggregated API endpoints.
```
deads2k, line 100: alerting on this as well seems appropriate

enhancements/authentication/invalid-certs.md
deads2k, line 112: adding the metric for an alert seems valid regardless

enhancements/authentication/invalid-certs.md
s-urbaniak: > For cases where the platform is a client, this is probably valid. For cases where the platform is a server, but not a client, I think it's too strong to block upgrades. Can you separate the cases?

afaiu we have the following separation:

**platform is a client**
- custom API webhooks
- custom aggregated API endpoints
- certificates of external auth identity providers

**platform is a server**
- custom certificates for route endpoints
- custom serving certificates for kube-apiserver

However, if users configured an invalid custom serving cert for kube-apiserver in 4.9 and we don't block upgrades, communication to kube-apiserver will be broken in 4.10. We don't want to block upgrades in this case?

enhancements/authentication/invalid-certs.md
s-urbaniak: > If it is just the operator, could the operator build a custom verifier for the check?

The client, in this case, is oauth-server, not authentication-operator.

s-urbaniak: indeed, good idea.

enhancements/authentication/invalid-certs.md
s-urbaniak: i think all the cases enumerated here are platform clients:
- custom serving certificates for kube-apiserver: openshift workloads
- custom API webhooks: kube-apiserver being the client
- custom aggregated API endpoints: kube-apiserver being the client
- custom certificates for route endpoints: openshift oauth-proxy workloads being the client
- certificates of external auth identity providers: oauth-server being the client

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,200 @@
...
+Verification of server certificates is executed during TLS client handshakes,
+a TLS (https) client observing an invalid certificate will reject the connection attempt.
+
+Cluster internal issued certificates are not affected,
+however custom certificates can be configured in various cases:
+- custom serving certificates for kube-apiserver
+- custom API webhooks
+- custom aggregated API endpoints
+- custom certificates for route endpoints
+- certificates of external auth identity providers
```
pierreprinetti, line 55: Shall we add certificates securing infrastructure API as a target case here?

For example, an OpenStack cloud exposing its API endpoints with a custom certificate. We have several platform-specific touchpoints (eg cluster-api-provider-openstack) that would need to expose the metric, if I understand this enhancement correctly

s-urbaniak: i'd consider this as a separate enhancement proposal or separate bugzilla. This proposal was not meant to be catch-all to focus on scope for the auth team.

pierreprinetti: OK. Opened https://bugzilla.redhat.com/show_bug.cgi?id=2038166 , thanks

enhancements/authentication/invalid-certs.md
```diff
@@ -0,0 +1,200 @@
...
+N/A
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+The above proposed changes **must** be backported to OpenShift 4.9.
+Otherwise, a detection of invalid certificates will not be possible before an upgrade to OpenShift 4.10.
```
pierreprinetti, line 172: I would also be interested in what happens in future releases.

Scenario: a new 4.11 feature triggers calls to a new HTTPS endpoint. The feature, that doesn't exist in 4.10, is active in 4.11 independently of user intervention. The new HTTPS endpoint is served with what you defined as "invalid certificate".

In this scenario, upgrading from 4.10 to 4.11 potentially brings the cluster to an unstable state.

In general terms: for each new HTTPS endpoint we call in 4.y, do we commit on adding an HTTPS-cert check in 4.y-1.z? Do we commit to doing that until a specific date?

For context, [Chrome has dropped support for CN in v58](https://developers.google.com/web/updates/2017/03/chrome-58-deprecations#remove_support_for_commonname_matching_in_certificates) in April 2017.

s-urbaniak: Starting from 4.10 we're not having a go runtime any more that is capable to accept legacy cert based endpoints in the first place.

pierreprinetti: It seemed to me that [library-go's IsHostnameError](https://github.com/openshift/library-go/pull/1275/files#diff-02272a0bc61d7d6f395f4226c37361bad47d9a5acc37f6fc28519e424be29e40R25-R40) was specifically designed for checking invalid certificates under Go v1.17+. What am I missing?

s-urbaniak: No, `IsHostnameError` will only work with <=1.16.

sdodson:
/reopen
It's my understanding that this has already been implemented. Can we do a quick review to see where this diverges from implementation and get this merged?

pierreprinetti:
@slaskawi @soltysh @deads2k @s-urbaniak 
Based on the absence of angry negative feedback, I am going to squash so that the PR is ready to merge.

pierreprinetti:
Minor fix: correcting the "created" and "updated" dates at the top of the document.

pierreprinetti:
I believe this PR is ready for you to consider LGTM'ing, @mfojtik.

PR #981, opened 2021-12-08
Added proposal for HyperShift monitoring.

bwplotka:
Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>


enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Goals
+
+The goal is to provide a solution that maintains similar monitoring capabilities we offer with single box monitoring using CMO in HyperShift for our three [personas](#personas).
+
+> NOTE: The definition of “monitor” in the bullet list below means ability to graph, query and alert on near-real-time metrics about involved components.
+
+* HSRE can monitor management clusters and control planes running there.
+  * Ability to use AlertManager and PagerDuty workflow
+  * Ability to monitor Tugboats themselves
```
gregsheremeta, line 71: avoid the IBM nomenclature, which we're not using
```suggestion
  * Ability to monitor Management Clusters themselves
```

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+* HyperShift is managed first, but form factor will be offered for self-management by 4.11.
+* There is tight budget on computation for control planes (outside of MVP).
+
+## Motivation
+
+This document enhancement proposes potential solution for Admin and Customer metric monitoring within those new topologies, since we cannot apply easily our current platform monitoring products using within cluster monitoring operator (CMO) easily (see [Alternatives](#use-pure-platform-monitoring-solution) on why).
+
+### Goals
+
+The goal is to provide a solution that maintains similar monitoring capabilities we offer with single box monitoring using CMO in HyperShift for our three [personas](#personas).
```
gregsheremeta, line 65: > similar monitoring capabilities we offer with single box monitoring using CMO

Where could a person who is unfamiliar with the current capabilities (such as me) go to read more about them? Could we link to that from here?

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+* CUS: Customers. 
+* BU: Business Unit. 
+* MS-SRE/MT-SRE: Managed Service / Managed Tenant Site Reliability Engineer.
+* RHOBS: Red Hat Observability Service.
+* [Observatorium](http://observatorium.io/): Project for scalable SaaS based monitoring and observability.
+
+For additional acronyms, see [Red Hat Dictionary](https://source.redhat.com/groups/public/red-hat-dictionary-or-lexicon/red_hat_abbreviations_acronyms_and_initialisms_dictionary_wiki)
+
+## Background
+
```
enxebre, line 46: nit: This domain/url is subject to change, let's reference the source repo https://github.com/openshift/hypershift which already contains the link to the docs.

mmazur:
A general comment – I'm having trouble reading the diagrams, because the terminology isn't consistent with https://hypershift-docs.netlify.app/reference/concepts/
Granted, there's not a term for "hosted cluster worker node part" and you use "dataplane". Not sure if that's intuitive/optimal.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+> NOTE: The definition of “monitor” in the bullet list below means ability to graph, query and alert on near-real-time metrics about involved components.
+
+* HSRE can monitor management clusters and control planes running there.
+  * Ability to use AlertManager and PagerDuty workflow
+  * Ability to monitor Tugboats themselves
+  * Ability to monitor Customer control planes
+  * Ability to view fleet-level metrics and drill-downs in a single pane of glass
+  * Ability to keep metrics in-region
+* HSRE can monitor control-plane related (managed) software on guest clusters:
```
jeremyeder, line 75: To me this means any on-HostedCluster operators that an SRE team owns, but we don't conclusively know if this is required.  Did you have other use-cases in mind?

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+  * For example Ingress/Router/CNI/CSI 
+  
+> Exact list of what is scraped on guest clusters for HSRE does not matter. All we need to know: some service monitors.
+
+* HSRE can query metrics data over SLO period (usually 28 days).
+  * Longer durations nice to have, but technically not required. There is also option that allows HSRE to persist SLO from platform to [different backend if needed](https://docs.google.com/document/d/1BJarERppgiJ8esc6d8anbJMOQ0AflFBQea-Zc9wAp0s/edit?disco=AAAAQTFsbwM).
```
jeremyeder, line 81: We can't miss a step here.  28 days is bare minimum but really we do want longer.  I'd argue one year.  Realize that discussion is out of scope for this proposal.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+* HSRE have an easy way to monitor dozens of HyperShift clusters in multiple geo regions and thousands of control planes in total. This means in particular:
+  * Ability to perform PromQL across fleet metrics
+
+> NOTE(Mariusz): Alerts can be region wide only, and query in different tab per region might be good enough for now.
+
```
jeremyeder, line 86: We will need to have a discussion about the SRE personas and their desired dashboard points of view based on the use-cases and roles.  It would probably help to have some mockups supplied by those folks.  Don't need it in the context of this proposal.  Will need to be decided in near future.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+  * Ability to see logs across the fleet to see a pattern
```
jeremyeder, line 87: Glad to see this in the design.  I know we're talking about monitoring metrics and alerts first.  But logs is a very close second, and I would like others to chime in on priority of that.  @mmazur @lisa

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+  * No need to tunnel to specific control plane to query metrics or see logs
+  * Centralized alerting
+* CUS can monitor / use HPA/VPA / vis with local console using metrics FROM their workloads on guest clusters or from nodes on guest clusters
+  * Container metrics (about their containers).
+  * This includes part of Kube State Metrics (Kubernetes metrics) that relate to their workloads.
+* BU can gather required telemetry using the Telemeter pipeline.
```
jeremyeder, line 93: DDIS team, too.  I'll reach out to them.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Non-Goals
+
+* Solve Logging and Tracing needs.
```
jeremyeder, line 97: logging is a non-goal only in the context of this initial doc.  it has to fast-follow.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+* Provide scalable storage solution for customer metrics. We just want to ensure UWM, their DIY,
+console or vendor monitoring does not have major blockers when integrating with this
+architecture.
+* Describe Managed Services (Addons) monitoring in detail. We assume Managed Services are just “Customer workloads”, where the MTSRE is just a customer. See [What About Addons FAQ](#what-about-layered-service--addons-monitoring) for more details.
+* Solve observability challenges brought by KCP. KCP has different trade-offs.
+* Provide any other UI than Grafana for our Managed HSRE needs.
+* Provide any other querying APIs than Prometheus based on PromQL
+* Give customers access to any control plane metrics unrelated to their workloads (this is similar in GKE).
+* Provide a solution for non-HyperShift OSD. In fact, the solution for OSD observability is planned to be really similar to HyperShift.
```
jeremyeder, line 106: Some convergence here might be logical.  We can see how it goes though, no rush.  A decision by end of Internal Preview phase would be desirable.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Personas
+
+* HyperShift Admin SRE (HSRE): People that will operate the Hypershift management cluster and its control planes. In practice, this means the SD org on our managed solutions.
+* Customers (CUS) People that attach data planes to HyperShift and run workloads on data planes.
+* Business Unit (BU): Red Hat business unit who is interested in telemetry and data from OpenShift deployments.
+
+### User Stories
+
+TBD
```
jeremyeder, line 120: This has to be fleshed out before this enhancement should be merged.

enhancements/monitoring/hypershift-monitoring.md
mmazur: OCM Agent Operator is OSD–specific I believe, so not a thing here. As for DVO–like functionality, Jeremy might be right that it's fine with only the data available from control plane. Would need to be verified.

enhancements/monitoring/hypershift-monitoring.md
lisa: What come to mind first is that "Solve Logging and Tracing needs" is a non-goal. How does that non-goal  inform _this_ goal? How are we seeing logs across the fleet to see a pattern?

enhancements/monitoring/hypershift-monitoring.md
bwplotka: Do you mind helping us with that? @jeremyeder @csrwng

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+![h](assets/hypershift-details.png)
+
+Let’s take a look at all parts:
+
+1. In terms of "Observability Product", we already have a growing open-source [Observatorium](http://observatorium.io/) project that packages and offers opinionated deployments of popular open-source projects for observability needs (e.g Thanos and Loki). What matters is that the Observability Group has a strong impact, deep knowledge and understanding of those projects. We also have ongoing production experience in running and scaling them.
+
+Anyone can install Observatorium, and deploy this service on-premise, but Observability Group also provides [RHOBS](https://rhobs-handbook.netlify.app/projects/observability/rhobs/) as a managed instance of observatorium, that is currently only available to specific internal teams.
+
+Observability Product will be a central pane of view of observability data for SRE needs from multiple clusters. Long term, it will also run critical Prometheus Alerting rules and alert routing for notification purposes.
```
alvaroaleman, line 191:
```suggestion
Observability Product will be a central pane of view of observability data for SRE needs from multiple clusters. 
```

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+2. Since the Management cluster is like any other OCP/OSD, we can leverage Platform Monitoring. In near future, we plan to allow [MonitoringStack](https://github.com/openshift/enhancements/pull/866) project to allow scraping platform related metrics, but for the time being, this is done via Platform Monitoring that can remote write relevant metrics to Observability Product.
```
alvaroaleman, line 193: Can we generalize this a bit, something like 

>Metrics are scraped by a component running in the management cluster that uses Prometheus remote_write to write them into the observability product. Initially, this will be done by the Platform Monitoring stack but this might change in the future.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+3. [MonitoringStack](https://github.com/openshift/enhancements/pull/866) project offer similar functionality as UWM, but it aims for more features, flexibility and scalability due to different support models (outside of OpenShift Core) and less tight integration to OpenShift itself. It’s being developed in https://github.com/rhobs/monitoring-stack-operator. We plan to add support to [the agent mode of Prometheus](https://twitter.com/bwplotka/status/1454146102240944134) for efficient metric forwarding.
```
alvaroaleman, line 195: I don't think this passage is relevant, we can cover this in the previous point by mentioning there that we need $something that does prometheus remote_write and which will initially be Platform Monitoring, but might be replaced later on. The Hypershift product doesn't actually need to know any of this, does it?

```suggestion
```

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+4. A similar pipeline can be used for a portion of HSRE owned components living in each data plane.
```
alvaroaleman, line 196: Please expand this. What change in the Hyprshift product is needed to get this done? How will the "similiar pipeline" know what is a HSRE owned component that it needs to scrape metrics from?

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+The typical workflows:
+* HPA/VPA using Prometheus Adapter
+* Monitoring and alerting on kube-state metrics, node exporter, cadvisor and application metrics
+* Local console access with OCP metrics and dashboard
+
+We don’t see any blockers for the proposed model if we would follow the same path as GKE, EKS or any other Managed Kubernetes with hidden control plane offers. By design, HyperShift offers managed Kube API. This means that any metrics related to the things running in the data plane and around API can be obtained by Customers on their own. It could work like this:
+
+![h](assets/hypershift-cus.png)
+
+This is not different to GKE/EKS cases, where, the customer has to deploy on their own kube-state-metrics, node-exporter, cadvisor etc and configure Open Metrics compatible solution to scrape those.
```
alvaroaleman, line 213: Here you say `This is not different to GKE/EKS cases, where, the customer has to deploy on their own kube-state-metrics` and below you write `As the Red Hat data plane, we used to offer this out-of-the-box ,so we should maintain this practice`. Those two are direct contradictions. We should probably remove the passage here?

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+As the Red Hat data plane, we used to offer this out-of-the-box ,so we should maintain this practice. In simple form we could use CMO without PM, but with all exporters and optional UWM. There is a slight tweak where UWM has to know scrape those data-plane KSM, node exporter, cadvisor.
+
+Our team also develops [Monitoring Stacks](https://github.com/openshift/enhancements/pull/866) which aims to do similar yet allows greater scalability and flexibility due to independent versioning and OLM installation model.
```
alvaroaleman, line 217: This is a Hypershift proposal, so saying "Our team" and not referring to the Hypershift team is very confusing. I also don't understand why it needs to be mentioned here, as the existence of Monitoring stack does not result in any change to the Hypershift product in the scope of this enhancement?

```suggestion
```

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+Yet, CUS can use any other DIY or just simple Prometheus/Prometheus Operator or vendor (e.g Grafana Cloud or anything else too) to do similar.
```
alvaroaleman, line 219: This whole passage is very confusing to me. Please clearly outline what the Hypershift product needs to do in order to provide customers with metrics, alerting and enable HPA/VPA to work. Today, we simply deploy the monitoring operator. Will this change or is there anything else that needs to be done?

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Risks and Mitigations
+
+Let's describe the worst case. What if the Observability Team or/and AppSRE fail to Provide a Managed Observability Service with Required Scalability and reliability (in time)?
```
alvaroaleman, line 223: This describes risks related to the Obvervability product, not risks related to the changes this enhancement proposes for Hypershift, IMHO this doesn't belong here.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Test Plan
+
+TBD
+
+## Implementation History
+
+If all parties will agree on the proposed direction we need heavy focus on the actions required:
+
+1. Add Alerting on top of Observatorium, then deploy it on RHOBS.
```
alvaroaleman, line 296: This is an implementation history of an obvervability product, not of monitoring in Hypershift.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+7. Provide efficient forwarding metrics collection solution (aka Prometheus agent).
+
+## Drawbacks
+
+* Different than OSD team effort to maintain critical Observability dependency with high SLO.
+* No native local storage solution
+
+## Alternatives
+
+#### Pay for Observability Vendor
```
alvaroaleman, line 319: This is not an alternative to the proposed changes in Hypershift, this is an alternative to the in-house observability product.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Open Questions
+
+### FAQ
+
+Some common questions gathered during initial proposal discussions.
+
+#### What About Layered Service / Addons Monitoring?
+
+This was discussed and solved by [Monitoring Stacks](monitoring-stack-operator.md). For HyperShift context, Addons requirements are no different to what we assume Customer would want to do / use. So in the context of this enhancement Addons are just another CUS.
```
david-martin, line 271: I can see why this would be a non-goal above. It's easier to say they are just "Customer workloads". 
However, I don't think this is the right attitude if we want addons to succeed as a whole.
I'm thinking internal (RH) here, where there's this RHOBS thing that is solving a lot of problems.
Maybe this isn't the right forum, but why wouldn't addons use that service?

enhancements/monitoring/hypershift-monitoring.md
bwplotka: It is, but I think we are fine with one year out of box for you

enhancements/monitoring/hypershift-monitoring.md
bwplotka: 👍🏽

enhancements/monitoring/hypershift-monitoring.md
bwplotka: Ups sorrry, looks like it's left -over. `* Solve Logging and Tracing needs.` is non Goal, however I would try to poke Observability Group about this (:

enhancements/monitoring/hypershift-monitoring.md
bwplotka: Thanks added this

enhancements/monitoring/hypershift-monitoring.md
bwplotka: 👍🏽

enhancements/monitoring/hypershift-monitoring.md
bwplotka: Hm, I think you statement is a bit conflicting here:

You mentioned:

>  leave any and all requirements for a central metrics aggregation solution out

and

>  IMHO from the POV of hypershift it is a blackbox service the hypershift components shouldn't need to know anything about except how to communicate with it.

But in order to allow HyperShift to treat it as black-box we need to change architecture to centralized approach. That is what simplifies client (from perspective of observability) workload. 

Still, while NON-HyperShift clusters are out of the discussion for this enhancement, I believe aiming for both short term (enabling Hypershift) as well as long term (all clusters have uniform observability) with one go is pretty tempting here! (:

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+- [ ] User-facing documentation is created.
+  - See [Documentation Required](#documentation-required) for further details.
+
+## Summary
+
+We propose to extract metric storage, alerting and querying to a separate service that can be scaled, deployed and managed in a decoupled mode. This will allow robust multi-cluster, fleet monitoring with best experience possible for cluster admins. This means SD org, running managed HyperShift will use RHOBS managed by Observability Group and AppSRE. 
+
+## Glossary
+
+* HSRE: HyperShift Admin SRE: People that will operate the Hypershift management cluster and its control planes. In practice, this means the SD org for ou managed solutions.
```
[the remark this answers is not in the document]
bwplotka: Makes sense 👍🏽

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+* CUS: Customers.
```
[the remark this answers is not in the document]
bwplotka: I don't know what you mean behind `"Owner of a hypershift-managed cluster`, but will try what we mean from our perspective here!

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+This document enhancement proposes potential solution for Admin and Customer metric monitoring within those new topologies, since we cannot apply easily our current platform monitoring products using within cluster monitoring operator ([CMO](https://github.com/openshift/cluster-monitoring-operator)) easily (see [Alternatives](#use-pure-platform-monitoring-solution) on why).
+
+### Goals
+
+The goal is to provide a solution that maintains similar monitoring capabilities we offer with single box monitoring using CMO in HyperShift for our three [personas](#personas).
+
+> NOTE: The definition of “monitor” in the bullet list below means ability to graph, query and alert on near-real-time metrics about involved components.
+
+* HSRE can monitor management clusters and control planes running there.
```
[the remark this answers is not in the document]
bwplotka: Hm, I think that is an important decision. 

I would argue that we need more than just `Integrate with a cross-management-cluster metrics solution`. Rationales:

* HyperShift means by default MANY clusters (dynamic) per single cluser, so we need scalable ingestion in place as well as ideally single query API so managing them is easy.
* We will have many HyperShift clusters and it's really troublesome (like existing OSD) to NOT have a central monitoring place.

It indeed feels like doing many things in one basket, but it truly looks like a good opportunity. Let's discuss this tomorrow on our call: https://docs.google.com/document/d/1crLlWZ9iefuayIcCb3Cwjb2rZrSm-xji9o3c98uEJe4/edit#heading=h.n1ea7eazt5zs

I will try to embed your idea as alternative we can discuss.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+console or vendor monitoring does not have major blockers when integrating with this
+architecture.
+* Describe Managed Services (Addons) monitoring in detail. We assume Managed Services are just “Customer workloads”, where the MTSRE is just a customer. See [What About Addons FAQ](#what-about-layered-service--addons-monitoring) for more details.
+* Solve observability challenges brought by KCP. KCP has different trade-offs.
+* Provide any other UI than Grafana for our Managed HSRE needs.
+* Provide any other querying APIs than Prometheus based on PromQL
+* Give customers access to any control plane metrics unrelated to their workloads (this is similar in GKE).
+* Provide a solution for non-HyperShift OSD. In fact, the solution for OSD observability is planned to be really similar to HyperShift. 
+This is because OSD observability struggles on similar points--inability to aggregate observability data in a cheap way on a global view.
+However, since HyperShift has no solution at all, for now starting from HyperShift sounds like a great opportunity.
```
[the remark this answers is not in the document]
bwplotka: Is there a way to improve language, but keep the point there? I think it matters if this enhancement proposed solution is incompatible or non-uniform to other OSD deployments.

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+## Proposal
+
+### Personas
+
+* HyperShift Admin SRE (HSRE): People that will operate the Hypershift management cluster and its control planes. In practice, this means the SD org on our managed solutions.
+* Customers (CUS) People that attach data planes to HyperShift and run workloads on data planes.
```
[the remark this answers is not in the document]
bwplotka: 👍🏽

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+> NOTE: We will focus on monitoring specific components in this document, but you will see logging parts on some diagrams. This is because logging can follow exactly the same pattern. Still, we will focus on monitoring (metrics) for now as it’s a bit unique (e.g. alerting part).
+
+#### The Big Picture
+
+We propose to send all relevant metrics to the "Centralized Observability Product” using well-defined and stable APIs. At this point we call it product because by design this is replaceable component that can be mixed and matched according to the needs.
+
+For example:
+
+* For our Managed HSRE, this can be our RHOBS service.
+* For CUS it can be locally deployed Observatorium or 3rd-party vendor like Grafana Cloud, Amazon or Google Managed Prometheus, Logz.io and others.
```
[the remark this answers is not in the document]
bwplotka: 👍🏽

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,392 @@
...
+
+* We can have tighter control on observability budget and spending
+* It’s our common, Red Hat responsibility to offer managed OpenShift. If Observability Group + AppSRE will not provide their support and time, the same (or more) effort and money has to be spent on OSD side (or vendor)
+* The monitoring Team already manages this service.
+  * We already work closely together with AppSRE, tightly integrated with them and familiar with AppInterface.
+  * We have devs on-call and SRE mindset.
+
+## Design Details
+
+We don’t want anyone to wait years to achieve this. As the Observability group, we want to give immediate value. Let’s see how we all can look when deployed. We proposed exact solutions that will have the required functionality, the quickest readiness and optimal stability and efficiency.
```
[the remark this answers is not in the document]
bwplotka: Agree, pretty mouthful.

enhancements/monitoring/hypershift-monitoring.md
bwplotka: IMO it very relates as it is an critical aspect of this enhancement.

enhancements/monitoring/hypershift-monitoring.md
bwplotka: I think it's a viable alternative we already use for logging AFAIK, no?

I feel it's integral to the solution we propose.

enhancements/monitoring/hypershift-monitoring.md
bwplotka: Thanks, I see your point, I don't think that is the message that is here. I never said Addons are never part of RHOBS. You are already part of it and using it, so there is definitely such direction. 

We only try to descope too many topics from one proposal. We like it or not our requirements are really not different than customer ones for addons observability (by what we gathered so far). Addons are first citizens as SD org in RHOBS, but AFAIK you can deploy the same way to HyperShift + this enhancement AND to normal OSD and that is the point here, no? (: 

Sorry if you got it as "addons" are less important. They are not. Is there a way I could improve the language here?

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,390 @@
...
+* Monitoring and alerting on kube-state metrics, node exporter, cadvisor and application metrics
+* Local console access with OCP metrics and dashboard
+
+We don’t see any blockers for the proposed model if we would follow the same path as GKE, EKS or any other Managed Kubernetes with hidden control plane offers. By design, HyperShift offers managed Kube API. This means that any metrics related to the things running in the data plane and around API can be obtained by Customers on their own. It could work like this:
+
+![h](assets/hypershift-cus.png)
+
+This is not different to GKE/EKS cases, where, the customer has to deploy on their own kube-state-metrics, node-exporter, cadvisor etc and configure Open Metrics compatible solution to scrape those.
+
+As the Red Hat data plane, we used to offer this out-of-the-box ,so we should maintain this practice. In simple form we could use CMO without PM, but with all exporters and optional UWM. There is a slight tweak where UWM has to know scrape those data-plane KSM, node exporter, cadvisor.
```
csrwng, line 213: 3 options are presented here:
- Use CMO without PM
- Use Monitoring Stacks Operator
- Allow customer to install DIY Prometheus

Which are we recommending? Are we offering all 3?
For option #1, use CMO, there are changes that need to happen to the CMO to do this. Can we add more detail on what that is?

enhancements/monitoring/hypershift-monitoring.md
david-martin: That makes sense.
Maybe all that's needed here is saying that RHOBS would still be the intended metrics & alerting solution for RH managed services on Hypershift as per https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/monitoring-stack-operator.md) ?

simonpasquier (commented):
Epic work! Similar to what @alvaroaleman already mentioned I think, I'd like to see more details about what needs to be changed in the cluster monitoring operator (CMO) to support the HyperShift deployment model. Describing how CMO works currently in HyperShift would be valuable too (I assume that @csrwng has the best knowledge here).
It would good to detail which components depend on the monitoring interfaces today (console, metrics API, HPA, ...) and how they will be impacted (for instance, do HyperShift customers have access to the OCP console? if yes from where would the console retrieve the metrics/alerts?).

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,390 @@
...
+This document enhancement proposes potential solution for Admin and Customer metric monitoring within those new topologies, since we cannot apply easily our current platform monitoring products using within cluster monitoring operator ([CMO](https://github.com/openshift/cluster-monitoring-operator)) easily (see [Alternatives](#use-pure-platform-monitoring-solution) on why).
+
+### Goals
+
+The goal is to provide a solution that maintains similar monitoring capabilities we offer with single box monitoring using CMO in HyperShift for our three [personas](#personas).
+
+> NOTE: The definition of “monitor” in the bullet list below means ability to graph, query and alert on near-real-time metrics about involved components.
+
+* HSRE can monitor management clusters and control planes running there.
+  * Ability to use AlertManager and PagerDuty workflow
```
kyoto, line 70:
```suggestion
  * Ability to use Alertmanager and PagerDuty workflow
```

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,390 @@
...
+
+### Important HyperShift Assumptions
+
+* One-way communication. HyperShift management cluster can “talk to” the data plane, but not the opposite (except Kube API, Ignition and OIDC endpoint).
+* HyperShift is managed first, but form factor will be offered for self-management by 4.11.
+* There is tight budget on computation for control planes (outside of MVP).
+
+## Motivation
+
+This document enhancement proposes potential solution for Admin and Customer metric monitoring within those new topologies, since we cannot apply easily our current platform monitoring products using within cluster monitoring operator ([CMO](https://github.com/openshift/cluster-monitoring-operator)) easily (see [Alternatives](#use-pure-platform-monitoring-solution) on why).
```
kyoto, line 61:
```suggestion
This document enhancement proposes a potential solution for Admin and Customer metric monitoring within those new topologies, since we cannot easily apply our current platform monitoring products using cluster monitoring operator ([CMO](https://github.com/openshift/cluster-monitoring-operator)) (see [Alternatives](#use-pure-platform-monitoring-solution) on why).
```

bwplotka:
Hi All!

Please PTAL latest update to this enhancement. Changes:

* Changed data-plane details given feedback from the in-cluster team. Right now we propose deploying an unchanged CMO that will be also used to remote write metrics tiny amount metrics for HSRE use to RHOBS.
* Removed details around RHOBS. Given the feedback we got, the proposal now focuses on changed in OpenShift in HyperShift topology, without mentioning RHOBS platform details.

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -29,7 +29,11 @@ status: implementable
...
+This proposal aims to figure our monitoring case of HyperShift topology. There are three important parts of this proposal:
+
+For HyperShift Admin, we propose to extract metric storage, alerting and querying for Clusters managed by HyperShift to a separate service that can be scaled, deployed and managed in a decoupled mode. This will allow robust multi-cluster, fleet monitoring with best experience possible for cluster admins. This means SD org, running managed HyperShift will use the RHOBS managed by Observability Group and AppSRE. Metrics from control planes of hosted clusters will be shipped thanks to Monitoring Stacks.
+
+The data plane part of guest clusters requires changes too. In terms of feature parity for users custom or UWM monitoring and HPA we propose to deploy unchanged CMO for smallest changes possible. We also propose using CMO path to gather and send needed monitoring data for HyperShift Admin to RHOBs as well as data-plane telemetry through the usual path.
```
jeremyeder, line 36: Just so I'm clear, this is an unchanged CMO running on the management clusters, as opposed to the Prometheus Agent running on management clusters?  Nothing running on the guest workers, in either scenario.  Confirm? @ianbillett

csrwng: My understanding is that the proposal is to run CMO unchanged on the guest workers so that things like HPA can keep working and collectors like the node exporter and kube state metrics are laid down by the CMO.

dofinn: So two CMOs per "customer deployment"?

1. Standard CMO -> control plane
2. Standard CMO -> data plane

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,384 @@
...
+## Design Details
+
+In details the full design looks as follows:
+
+![h](assets/hypershift-details.png)
+
+Let’s take a look at all parts:
+
+1. Hosted control planes does not deploy any own monitoring, other than relevant Service Monitors to be scraped.
+2. For all control planes we propose to deploy [Monitoring Stack Operator](https://github.com/rhobs/monitoring-stack-operator) that will deploy set of Prometheus/Prometheus Agents that will be sending data off cluster to RHOBS. No local alerting and local querying will be allowed. This path will forward both monitoring data as well as part of telemetry relevant for Telemeter.
```
dofinn, line 185: What is the requirement for prometheus here? How does the agent not suffice?

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,384 @@
...
+
+* HyperShift client monitoring and how to configure metrics remote write (push API) to desired backend.
+
+Internal docs:
+
+* RHOBS entry-points and usage.
+
+### Open Questions
+
+* Who will configure CMO to allow remote writing metrics to RHOBS?
```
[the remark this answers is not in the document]
dofinn: - mgmt CMO is standard OSD/ROSA by SREP?

- we should consider solving for guest clusters with no egress because that solves everything?

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,384 @@
...
+
+* Who is responsible for deploying Monitoring Stacks on Management Clusters for Control planes to use?
```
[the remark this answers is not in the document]
dofinn: this is https://github.com/rhobs/monitoring-stack-operator ?

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,384 @@
...
+
+Let’s take a look at all parts:
+
+1. Hosted control planes does not deploy any own monitoring, other than relevant Service Monitors to be scraped.
+2. For all control planes we propose to deploy [Monitoring Stack Operator](https://github.com/rhobs/monitoring-stack-operator) that will deploy set of Prometheus/Prometheus Agents that will be sending data off cluster to RHOBS. No local alerting and local querying will be allowed. This path will forward both monitoring data as well as part of telemetry relevant for Telemeter.
+3. On data-plane we propose running roughly unchanged CMO stack with platform monitoring and optional UWM. This allows feature parity for customers as well as ability to remote write few metrics interersting for HSRE use (as specified in Goal). The unique part is that Platform Monitoring will know nothing about master nodes and control plane resources (etcd, control plane operators, Kube services etc). It will only provide "platform" metrics for worker nodes and containers running there. Part of telemetry related to data-plane will be sent through Telemeter client as usual. We assume CMO can be deployed on Management cluster for easier maintenance.
```
csrwng, line 186: CMO for the data plane is currently deployed on the worker nodes. Are you proposing that we move it to the control plane? What do we gain?

enhancements/monitoring/hypershift-monitoring.md
ianbillett: I think @jan--f / @fpetkovski / @bwplotka will be able to help with this question?

simonpasquier: IIUC this is how HyperShift is deployed today:
* All monitoring components runs on the guest workers (unlike "traditional" installations of OCP where  the cluster monitoring operator runs on the control plane nodes).
* No service monitors, rules and dashboards are deployed for the control plane components (Kube/OpenShift API, scheduler, etcd, ...).

My understanding of the current proposal is the following:
* It introduces a new monitoring stack on the hosted control plane to collect metrics from the control plane components and forward them to RHOBS. The implementation is yet to be defined but it will eventually be based on the Prometheus agent.
* It doesn't change anything for the guest workers. The platform components running there are still monitored by CMO (required for HPA, console dashboards, telemetry, ...). A selected number of platform metrics could be forwarded to RHOBS if needed via remote write (already supported).
* It doesn't change anything for the management clusters which are standard OSD clusters.

jewzaam (changes requested):
A few points discussed on the sync call this morning that need incorporated if not already.
1. need consistent key on all metrics from a cluster to identify the cluster
2. this includes all SRE and managed service metrics
3. running telemeter-client to scrape and push worker, SRE, and managed service metrics might be a good solution

There would be changes needed I think.  We need to have a way to configure what metrics get shipped beyond the standard ones.  We need a way to set the frequency for scraping and remote writing.  I don't know that these need to be different.  Probably doesn't make sense to be different actually.

@bwplotka 
fyi @dofinn

dofinn:
> A few points discussed on the sync call this morning that need incorporated if not already.
> 
> 1. need consistent key on all metrics from a cluster to identify the cluster
> 2. this includes all SRE and managed service metrics
> 3. running telemeter-client to scrape and push worker, SRE, and managed service metrics might be a good solution
> 
> There would be changes needed I think. We need to have a way to configure what metrics get shipped beyond the standard ones. We need a way to set the frequency for scraping and remote writing. I don't know that these need to be different. Probably doesn't make sense to be different actually.
> 
> @bwplotka fyi @dofinn

Regardless if its telemeter that ships or prom/remoteWrite that ships. SRE will need `_id` field on all exported metrics. This is a problem we have not solved in OSD/ROSA https://issues.redhat.com/browse/OSD-6573 as we dont have the ability to configure external_labels dynamically per clusters/remoteWrite into obs-mst.

jan--f:
CMO adds the[ `_id` label to the telemeter-client config](https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/manifests/manifests.go#L1499) today. We can certainly do the same for any remote_write config as well.

simonpasquier:
> We need a way to set the frequency for scraping and remote writing. I don't know that these need to be different. Probably doesn't make sense to be different actually.

Remote write would just push samples as they get ingested by Prometheus. Though you can control the size of the shards, the time to wait before sending a batch, what's the max number of samples per batch, ... (see `queue_config` in https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write).

bwplotka:
I'm back from long weeks of knee post-surgery recovery. Let's see what is missing for this to merge.

> A few points discussed on the sync call this morning that need incorporated if not already.
> need consistent key on all metrics from a cluster to identify the cluster

`_id` is the common thing, but we need to discuss what are the ways to set it. 


> this includes all SRE and managed service metrics

This means what @dofinn? This proposal? I don't believe this proposal addresses manage service metrics which has to be collected from data planes (I presume).

> running telemeter-client to scrape and push worker, SRE, and managed service metrics might be a good solution

We don't need a telemeter client. Collectors (e.g Prometheus or Prometheus Agent) should be capable of doing all of this without telemetry client complexity.

enhancements/monitoring/hypershift-monitoring.md
bwplotka: > We assume CMO can be deployed on the Management cluster for easier maintenance.

"Easier maintenance" was the sentence. We don't see any other strong motivation in either direction, so up to you.

dofinn:
> `_id` is the common thing, but we need to discuss what are the ways to set it.

@bwplotka - solution is being designed in https://issues.redhat.com/browse/MON-2235 
 
> > this includes all SRE and managed service metrics
> 
> This means what @dofinn? This proposal? I don't believe this proposal addresses manage service metrics which has to be collected from data planes (I presume).

@jewzaam could you please help to clarify? Are MST planning on piggybacking on this monitoring flow or are they responsible for gathering and exporting what ever metrics from the data plane to their tenant in MST? 

> > running telemeter-client to scrape and push worker, SRE, and managed service metrics might be a good solution
> 
> We don't need a telemeter client. Collectors (e.g Prometheus or Prometheus Agent) should be capable of doing all of this without telemetry client complexity.

+1

jewzaam:
> > > this includes all SRE and managed service metrics
> > 
> > This means what @dofinn? This proposal? I don't believe this proposal addresses manage service metrics which has to be collected from data planes (I presume).
> 
> @jewzaam could you please help to clarify? Are MST planning on piggybacking on this monitoring flow or are they responsible for gathering and exporting what ever metrics from the data plane to their tenant in MST?

@dofinn  no plans are written down for observability with respect to ARO and HyperShift yet.  Historically only ALERTS metric are consumed from Prometheus.  But that doesn't mean we'll continue with that pattern and we will look at this sometime Q2.

enhancements/monitoring/hypershift-monitoring.md
bwplotka: Let's provide EPIC for it.

enhancements/monitoring/hypershift-monitoring.md
bwplotka: Amazingly explained @simonpasquier 

I updated the enhancement with some words from your explanation.

enhancements/monitoring/hypershift-monitoring.md
bwplotka: 👍🏽

enhancements/monitoring/hypershift-monitoring.md
```diff
@@ -0,0 +1,398 @@
...
+
+Probably the most important label to ensure consistency across HyperShift metrics is a label indicating hosted cluster. This is because:
+
+* Telemeter uses `_id`. If we maintain same convention we have easier way to joining data across telemeter and other tenants if needed.
+* If we push metrics from both data plane and hosted control planes it is critical to associate those with same cluster, so we can join/view this data uniformly. 
+
+We propose reusing `_id` which is what Telemeter uses. This is also by default propagated in the in-cluster platform monitoring (it takes `_id` from [ClusterVersionSpec](https://github.com/openshift/api/blob/master/config/v1/types_cluster_version.go#L41)), so data from data plane should be already instrumented correctly.
+
+When it comes to hosted control planes we propose for hosted control plane service monitors (or pod monitors) to include relabeling that adds `_id` with corresponding cluster id to each metric. When scraped this label will be attached and propagated to RHOBS.
```
dofinn, line 198: @bwplotka is this available in ServiceMonitor? I can not see it 

```
oc explain servicemonitor.spec
```

csrwng:
```
oc explain servicemonitor.spec.endpoints.metricRelabelings
```

dofinn: doh! thanks @csrwng :) if hypershift ENG is going to standardize on this ? then we should to iterate/ripple this through SD 
 for consistency :)

bwplotka:
Anything else on this in order to get it approved and merged here on enhancements? (: cc @dofinn 

I no, then I will fix CI lint and we should be good to go on this small documentation piece, thanks!

bwplotka:
Thanks, resolving the CI

PR #984, opened 2021-12-10
Update OWNERS

aravindhp:
- Remove ex-Red Hatters :-(
- Add Elango Sivanandam @selansen (Windows Containers team lead) :-)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,13 +1,10 @@
 approvers:
-  - abhinavdahiya
   - adambkaplan
   - aravindhp
   - ashcrow
   - bparees
   - celebdor
-  - chancez
   - cooktheryan
-  - danehans
   - deads2k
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -1,11 +1,9 @@
 approvers:
-  - abhinavdahiya
   - adambkaplan
   - aravindhp
   - ashcrow
   - bparees
   - celebdor
-  - chancez
   - cooktheryan
   - danehans
```
sdodson, line 8: @aravindhp Daneyon left too.

aravindhp: Removed

sdodson:
/lgtm
I think everyone who no longer works here is removed, we can look into other strategies to make sure people are assigned who are more likely to review via other means.

PR #987, opened 2021-12-14
Add the DNS-over-TLS enhancement initial draft

brandisher:


enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+`TLSSecurityProfile` encapsulates TLS specific configuration. Including this 
+allows customers to set their desired TLS version, ciphers, etc.
+
+`CABundle` references a ConfigMap that contains a single CA certificate 
+or a CA bundle.
+
+`ServerName` contains an FQDN to validate when establishing the connection. 
+This is relevant in particular with cases where a customer has multiple 
+upstreams behind one serving certificate (e.g. `1.1.1.1` and `2.2.2.2` 
+behind `dns.myorg.com`)
```
Miciah, line 67: `ServerName` serves two purposes:
* The client specifies the server name in the TLS handshake, which tells the remote end which server the client wants.  (I don't know whether virtual servers are common with DNS, but theoretically the remote end might need this.)
* The client uses the server name to verify the serving certificate that the server provides.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+    // +kubebuilder:validation:Optional
+    TLSSecurityProfile v1.TLSSecurityProfile `json:tls,omitempty`
+
+    // CABundle references a ConfigMap that must contain either a single
+    // CA Certificate or a CA Bundle in the case of multiple upstreams signed
+    // by different CAs.
+    //
+    // +optional
+    // +kubebuilder:validation:Optional
+    CABundle *corev1.ConfigMap `json:caBundle,omitempty`
```
Miciah, line 90: What happens if the user specifies a reference to a configmap that doesn't exist?

What happens if `caBundle` is specified but no server name is specified?

Miciah (commented):
This is detailed and well written.  Looks good overall.  I've pointed out a few places where we need to tweak the proposed API or fill in some details.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+### Goals
+
+1. A cluster administrator can specify a CA certificate or CA bundle, TLS configuration, and an upstream FQDN for the default zone in the DNS configuration.
+2. A cluster administrator can specify a CA certificate or CA bundle, TLS configuration, and an upstream FQDN for a non-default zone.
+3. A cluster administrator can opt-in to DNS-over-TLS (using the configured CA/bundle) for each listed upstream in the default zone.
+4. A cluster administrator can opt-in to DNS-over-TLS (using the configured CA/bundle) for a set of upstreams on a non-default zone.
+
+### Non-Goals
+
+* Allowing configuration of DNS-over-TLS for use cases other than OpenShift's default DNS provider to upstream resolvers is out of scope.
```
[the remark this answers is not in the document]
brandisher: I think I see the miss here; my intention with the non-goals was to make all other use cases for the configuration not applicable (e.g. this won't let you configure forwarding for user-deployed CoreDNS, static CoreDNS pods, etc). Would something like this work?

```
* This enhancement applies only to the CoreDNS Operator shipped as part of OpenShift. Anything outside of that scope would be not applicable to this enhancement including the CoreDNS static pods managed by the `machine-config-operator`.
```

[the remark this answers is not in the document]
Miciah: Yeah, that looks fine.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Proposal
+
+This proposal covers changes to two APIs. The first is a change to 
+`UpstreamResolvers`.  To enable a cluster administrator to configure 
+DNS-over-TLS, the `UpstreamResolvers` type will get three new optional members:
+
+`TLSSecurityProfile` encapsulates TLS specific configuration. Including this 
+allows customers to set their desired TLS version, ciphers, etc.
```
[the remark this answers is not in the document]
brandisher: That is correct, my mistake! I will rework this.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+  * CoreDNS will log an error if the TLS negotiation with the upstream
+    resolver fails.
+* **Accidental opt-in for a given upstream resolver.**
+  * Cluster administrators can revert the configuration.
+  * CoreDNS will log an error if the TLS negotiation with the upstream
+    resolver fails.
+
+#### Support Procedures
+
+_TBD_
```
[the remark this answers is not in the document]
brandisher: I've added a note to my working commits that captures this input.

> I don't think it's possible to get the TLS session key from CoreDNS, is it? So the only way to decrypt packet captures when DNS-over-TLS is used would be to get the server's key?

As far as I can tell, that's correct. The only references I can see with regards to a configurable key in the forward plugin is if you're doing client authentication which would be outside the scope of this enhancement. In this case, all we'll have is a cabundle for verification of the upstream resolver.

Having packet captures be encrypted will indeed make support motions very difficult. Is there another way to approach that aspect without requiring the server key?

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+    // +optional
+    // +kubebuilder:validation:Optional
+    CABundle *corev1.ConfigMap `json:caBundle,omitempty`
+	
+    // Server name to validate in the case of one or more upstreams behind one
+    // FQDN.
+    //
+    // +optional
+    // +kubebuilder:validation:Optional
+    ServerName string
```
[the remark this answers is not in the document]
brandisher: I had one clarifying question on this; the correct validation of this would be a pattern match (`+kubebuilder:validation:Pattern`) for an fqdn, correct?

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+
+`ServerName` contains an FQDN to validate when establishing the connection.
+This is relevant in particular with cases where a customer has multiple
+upstreams behind one serving certificate (e.g. `1.1.1.1` and `2.2.2.2`
+behind `dns.myorg.com`)
+
+Because the list of upstreams in the `ForwardPlugin` is a `[]string`, 
+DNS-over-TLS will be attempted for all upstreams configured for a zone by 
+default and fallback to regular DNS if a TLS connection cannot be 
+established. This will generate a log record indicating that a TLS
```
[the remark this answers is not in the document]
brandisher: Agreed. I've reached out to PM to double check whether this is acceptable.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+* **Misconfiguration of the TLS certs**
+  * Cluster administrators can revert the configuration.
+  * TLS certs will be validated prior to admission using existing code or
+    libraries.
+  * CoreDNS will log an error if the TLS negotiation with the upstream
+    resolver fails.
+* **Accidental opt-in for a given upstream resolver.**
+  * Cluster administrators can revert the configuration.
+  * CoreDNS will log an error if the TLS negotiation with the upstream
+    resolver fails.
```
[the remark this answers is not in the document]
brandisher: This will be TBD for now as the implementation needs to be done but I will add them as soon as they're available.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+```go
+type Upstream struct {
+    // ...snip
+	
+    // UseTLS allows cluster administrators to opt-in to using the TLS config in
+    // UpstreamResolvers for upstreams that support DNS-over-TLS.
+    //
+    // +optional
+    // +kubebuilder:validation:Enum=;true;false
+    // +kubebuilder:default=false
```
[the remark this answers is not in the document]
brandisher: I switched this to `EnableTLS` as that makes sense to me. Seems like this should be explicitly opt-in since regular DNS queries should work by default but TLS queries may not work by default unless that configuration is correct and the upstream supports it.

enhancements/dns/forward-dns-over-tls.md
brandisher: I've updated this with the configmap reference feedback but can you shed some light on what the action item is in this comment? I'm not sure if you're suggesting that I adjust the kubebuilder validation or if these are godoc adjustments (or something else entirely).

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+
+**UX:** Given that this enhancement is only accessible via configuration 
+editing, no UX testing is required.
+
+## Design Details
+
+### Open Questions [optional]
+
+* Should the cluster administrator supplied certificates be stored in the 
+  openshift-dns or openshift-dns-operator namespaces?
```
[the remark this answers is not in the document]
brandisher: ack, is it customary to include a commit with an answer to the question or should the question be removed now that its been answered?

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,266 @@
...
+or a CA bundle.
+
+`ServerName` contains an FQDN to validate when establishing the connection. 
+This is relevant in particular with cases where a customer has multiple 
+upstreams behind one serving certificate (e.g. `1.1.1.1` and `2.2.2.2` 
+behind `dns.myorg.com`)
+
+Since `UpstreamResolvers` doesn't allow for configuration of a custom zone, 
+it's assumed that the same TLS configuration will apply to _n_ `Upstream` 
+objects.
```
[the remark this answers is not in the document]
brandisher: I think we're at an early enough stage that the design could be adjusted pending your approval. I've kept `upstreamResolvers` consistent in heirarchy with `forwardPlugin` in that you can define a cabundle and servername at the top level forwarding/upstream block. However, since Upstream is a little more flexible we could conceivably allow cabundle and servername to live at the `Upstream` level. That would eliminate the need for the `EnableTLS` field and answer some functional questions.

Example:
```
spec:
  upstreamResolvers:
    upstreams:
    - type: SystemResolvConf
      address: 1.1.1.1
      port: 5353
      caBundle: dns-tls-configmap
      servername: dns.bar.com
```
Then the assumption would be that all of the resolvers in `/etc/resolv.conf` would use TLS mostly because there's no way to split those out further.

What do you see as the action items on this one? Should this be a design change or a godoc change?

enhancements/dns/forward-dns-over-tls.md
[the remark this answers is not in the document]
Miciah: If we can't get the server key, an alternative would be to have CoreDNS log keys using something like <https://pkg.go.dev/crypto/tls#example-Config-KeyLogWriter>.  However, that would require an upstream change and would be a bit beyond the scope of this enhancement (unless you're feeling especially ambitious!).  If nothing else, we can document how using TLS will impact debuggability and mention `KeyLogWriter` as a possible future change.

enhancements/dns/forward-dns-over-tls.md
Miciah: These are things that need to be addressed in godoc so the semantics of these new fields is clear.

enhancements/dns/forward-dns-over-tls.md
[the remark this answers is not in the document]
Miciah: Is it a requirement that it be possible to enable TLS for nameservers that are specified in `/etc/resolv.conf`?  I'm gradually coming around to two realizations:

* For `type: SystemResolvConf`, we really shouldn't deviate from `/etc/resolv.conf` semantics because `/etc/resolv.conf` is coming from the node host's `/etc/resolv.conf`, and in that context, the nameservers are non-TLS, so we should assume that they are non-TLS in general.  
* `Upstream` probably should have been a union type.

enhancements/dns/forward-dns-over-tls.md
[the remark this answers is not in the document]
Miciah: I'm not sure what's customary, so use whatever style you prefer: FAQ-style question and answer, or just put the design we landed in the appropriate section.

enhancements/dns/forward-dns-over-tls.md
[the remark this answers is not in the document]
brandisher: I've reached out to PM about this one also; I want to get clarity on your question regarding nameservers in `/etc/resolv.conf`.

I could use some direction on where to go with this item next. From what I understand, there's probably a use case for defining `CABundle` and `ServerName` in the `Upstream` instead of in `UpstreamResolvers` since it may not apply to all of the upstreams.

The other action item is addressing the `SystemResolvConf` case. That would have to be handled by a conditional statement in the implementation, correct? I don't see anything in kubebuilder that allows for conditional definition on a field.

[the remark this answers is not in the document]
Miciah: Correct.  I'm thinking it might have been better to have made the `Upstream` type that we use for `spec.upstreamResolvers.upstreams[*]` a union type, as defined in <https://github.com/kubernetes/enhancements/blob/0e4d5df19d396511fe41ed0860b0ab9b96f46a2d/keps/sig-api-machinery/1027-api-unions/README.md>.  That would have enabled us to specify a group of parameters that could be specified specifically when `spec.upstreamResolvers.upstreams[*].type` were set to "Network".  With feature freeze approaching, though, I don't want to go and try to revise the API now.  So we really just need to leave it up to the operator to ignore fields that are inapplicable for the specified type.  

That aside, in the Corefile, a server block can have a `forward` block with a list of upstreams (which may include any combination of `/etc/resolv.conf` and explicit addresses) andparameters for the TLS cert and server name, and it is better for the API not to allow the user to express something that cannot be expressed in the Corefile, so I think `caBundle` and `servername` belong under `upstreamResolvers`, not under `upstreams`.  We can keep `enableTLS`, which would belong under `upstreams` but would be inapplicable for `SystemResolvConf`.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,318 @@
...
+  //
+  // +optional
+  // +kubebuilder:validation:Optional
+  CABundle corev1.ConfigMapNameReference `json:caBundle,omitempty`
+  
+  // serverName is the upstream server to connect to for DNS resolution. 
+  // This is required when a CABundle is supplied. Populating the 
+  // ServerName without a CABundle configured will have no effect.
+  //
+  // +optional
```
candita, line 93:
```suggestion
  // +required
```

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,318 @@
...
+  // +kubebuilder:validation:Pattern=^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$
```
candita, line 94: Each label (the portion between dots) can be up to 63, not 61, and a hyphen is allowed in any label.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,317 @@
...
+  // referenced does not exist, DNS-over-TLS will not be enabled.
+  //
+  // 1. The configmap must contain a `cabundle.crt` key
+  // 2. The value must be a PEM encoded certificate or certificate bundle.
+  // 3. The administrator must create this configmap in the openshift-config namespace.
+  // 4. The ServerName value must be supplied when CABundle is configured.
+  //
+  // +optional
+  // +kubebuilder:validation:Optional
+  CABundle v1.ConfigMapNameReference `json:"caBundle,omitempty"`
```
deads2k, line 87: this looks like the wrong level for this setting.  I choose servers inside the individual Upstreams right?

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,317 @@
...
+
+```go
+type UpstreamResolvers struct {
+  // ...<snip>
+
+  // caBundle references a ConfigMap that must contain either a single
+  // CA Certificate or a CA Bundle (in the case of multiple upstreams signed
+  // by different CAs). Populating a configmap name in this field
+  // effectively enables DNS-over-TLS. If this is empty or the configmap
+  // referenced does not exist, DNS-over-TLS will not be enabled.
```
deads2k, line 78: Can't we still allow enableTLS and say that we use system certs to verify?

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,317 @@
...
+type Upstream struct {
+  // ...snip
+
+  // enableTls allows cluster administrators to opt-in to using the TLS config in
+  // spec.upstreamResolvers for upstreams that support DNS-over-TLS.
+  //
+  // +optional
+  // +kubebuilder:validation:Enum=true;false
+  // +kubebuiler:default=false
+  EnableTLS string `json:"enableTls,omitempty"`
```
deads2k, line 115: nothing is a bool, how about a transport enum with TLS and ClearText

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,317 @@
...
+  // referenced does not exist, DNS-over-TLS will not be enabled.
+  //
+  // 1. The configmap must contain a `cabundle.crt` key
+  // 2. The value must be a PEM encoded certificate or bundle.
+  // 3. The administrator must create this configmap in the openshift-config namespace.
+  // 4. The ServerName value must be supplied.
+  //
+  // +optional
+  // +kubebuilder:validation:Optional
+  CABundle v1.ConfigMapNameReference `json:"caBundle,omitempty"`
```
deads2k, line 139: Same comment about granularity here.  Looks like the upstreams here may be a []string, so me have to create a new upstreams to represent this properly.

enhancements/dns/forward-dns-over-tls.md
brandisher: `UpstreamResolvers` is only for the root zone so we're limited by what can be represented in the Corefile. The limitation in this case is that we can only have one server block for the root zone `.:5353`.

enhancements/dns/forward-dns-over-tls.md
brandisher: I like that idea! Implemented locally.

enhancements/dns/forward-dns-over-tls.md
brandisher: Similarly here as with `UpstreamResolvers` in that we're limited by what we can represent in the Corefile. [These docs](https://docs.openshift.com/container-platform/4.9/networking/dns-operator.html#nw-dns-forward_dns-operator) have an example of the resulting Corefile when forwardplugin is configured but the main "gotcha" is that CoreDNS only allows one forward block per server so we can't configure separate TLS settings per upstream, they apply per server block (e.g. per set of zones that need forwarding).

enhancements/dns/forward-dns-over-tls.md
brandisher: That's a good point; we can do that with CoreDNS so I'll update this as well.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,334 @@
...
+  // ...<snip>
+  
+  // transport allows cluster administrators to opt-in to using a DNS-over-TLS 
+  // connection between cluster DNS and an upstream resolver(s). Configuring 
+  // TLS as the transport at this level without configuring a CABundle will 
+  // result in the system certificates being used to verify the serving 
+  // certificate of the upstream resolver(s).
+  //
+  // +optional
+  // +kubebuiler:default:="cleartext"
```
deads2k, line 97: this makes it impossible to distinguish between, "I don't care, choose a good default" and "I really want cleartext".  We can litigate this detail on the openshift/api PR.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,334 @@
...
+  Transport DNSTransport `json:"transport,omitempty"`
+  
+  // caBundle references a ConfigMap that must contain either a single
+  // CA Certificate or a CA Bundle (in the case of multiple upstreams signed
+  // by different CAs). This allows cluster administrators to provide their 
+  // own CA or CA bundle for validating the certificate of upstream resolvers.
```
deads2k, line 103: describe what happens when unset and transport is TLS.  can be done in the API PR instead of here.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,334 @@
...
+  //
+  // +optional
+  // +kubebuilder:validation:Pattern=`^([a-zA-Z0-9\p{S}\p{L}]((-?[a-zA-Z0-9\p{S}\p{L}]{0,62})?)|([a-zA-Z0-9\p{S}\p{L}](([a-zA-Z0-9-\p{S}\p{L}]{0,61}[a-zA-Z0-9\p{S}\p{L}])?)(\.)){1,}([a-zA-Z0-9-\p{L}]){2,63})$`
+  ServerName string `json:"serverName,omitempty"`
+}
+```
+
+Because the list of upstreams in the `ForwardPlugin` is a `[]string`, 
+DNS-over-TLS will be attempted for all upstreams configured for a zone by 
+default and fallback to regular DNS if a TLS connection cannot be
```
deads2k, line 163: This doesn't make sense to me.  If I want TLS, I want to be able to specify it in a way that ensures that TLS Is actually used.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,334 @@
...
+}
+```
+
+Because the list of upstreams in the `ForwardPlugin` is a `[]string`, 
+DNS-over-TLS will be attempted for all upstreams configured for a zone by 
+default and fallback to regular DNS if a TLS connection cannot be 
+established. This will generate a log record indicating that a TLS 
+connection could not be established.
+
+A cluster administrator can set only `Transport` and `ServerName` to allow
```
deads2k, line 167: this bit will need to be represented in the godoc so people reading it know what to expect.

deads2k:
https://github.com/openshift/enhancements/pull/987/files#r813350736

looks like a blocker. If I want TLS connections, I'm not going to be happy if my secure configuration actually uses an insecure connection.

enhancements/dns/forward-dns-over-tls.md
Miciah: @brandisher, can we revise the semantics here?  Is this a limitation of CoreDNS?

brandisher: @Miciah this is a limitation of CoreDNS as I understand it. Here's a quote from the documentation for the forward plugin: `Also note the TLS config is “global” for the whole forwarding proxy if you need a different tls-name for different upstreams you’re out of luck.`

That coupled with this limitation `This plugin can only be used once per Server Block.` means that you could only ever define one forward block with one TLS config that applies to all upstreams per server block.

enhancements/dns/forward-dns-over-tls.md
brandisher: Understood, can this be handled in the API PR as well?

enhancements/dns/forward-dns-over-tls.md
brandisher: ack, I'll fix this in the API PR.

enhancements/dns/forward-dns-over-tls.md
brandisher: ack, will do!

brandisher:
> https://github.com/openshift/enhancements/pull/987/files#r813350736
> 
> looks like a blocker. If I want TLS connections, I'm not going to be happy if my secure configuration actually uses an insecure connection.

Totally agree. I'll adjust this as it'd require a workaround on our part to cater to the failed TLS connection so this makes the work simpler and the end result more secure.

enhancements/dns/forward-dns-over-tls.md
brandisher: @deads2k After chatting with @Miciah I realized that I misunderstood your comment. A connection will fail if the TLS negotiation fails but we'll be sure to cover that case with tests in the implementation.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,317 @@
...
+2. Allow a cluster administrator to define a `ServerName` which matches the 
+   server certificate of the upstream resolver(s) for forwarded zones.
+3. Allow a cluster administrator to optionally configure a custom CA that 
+   will serve as a verification CA for the server certificate of the 
+   upstream resolver(s).
+
+### Non-Goals
+
+* This enhancement applies only to the CoreDNS Operator shipped as part of OpenShift. Anything outside that scope would be not applicable to this enhancement including the CoreDNS static pods managed by the `machine-config-operator`.
+* This enhancement does not create or enable a general purpose DNS-over-TLS server.
```
Miciah, line 57:
```suggestion
* This enhancement does not create or enable a general purpose DNS-over-TLS server.  In particular, enabling TLS for communication between in-cluster resolvers and CoreDNS is out of scope for the enhancement.  
```

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,317 @@
...
+```yaml
+spec:
+  upstreamResolvers:
+    transport: tls
+    tls:
+      caBundle: dns-tls-configmap
+      servername: dns.bar.com
+    upstreams:
+    - type: Network
+      address: 1.1.1.1
```
Miciah, line 172: Is it possible for the cluster administrator to specify TLS while using the upstreams from `/etc/resolv.conf`?  If so, does doing so have well defined semantics?  
```yaml
spec:
  upstreamResolvers:
    transport: tls
    tls:
      caBundle: dns-tls-configmap
      servername: dns.bar.com
    upstreams:
    - type: SystemResolvConf
```
Related discussion: https://github.com/openshift/enhancements/pull/987#discussion_r792184198

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,317 @@
...
+### Upgrade / Downgrade Strategy
+
+On upgrade, the default configuration and existing forward configuration 
+will work without issue.
+
+On downgrade, this functionality will revert to sending plaintext 
+DNS queries since the downgraded operator will have the older CRD 
+and logic that does not include these new API fields. Cluster 
+administrators should remove configuration specific to this enhancement 
+after downgrade.
```
Miciah, line 266:
```suggestion
before downgrade.
```
Better to revert to cleartext _before_ downgrading because otherwise the downgrade may force the DNS configuration into an unexpected (and nonworking) state.

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,317 @@
...
+### Operational Aspects of API Extensions
+
+N/A
+
+#### Failure Modes
+
+* **Misconfiguration of the TLS certs**
+  * Cluster administrators can revert the configuration.
+  * TLS certs will be validated prior to admission using existing code or
+    libraries.
```
Miciah, line 281: The phrase "validated prior to admission" suggests to me that you intend to implement an API admission check, but the user creates the configmap using an arbitrary name and may create the configmap before adding a reference to it from the cluster dns config.  What kind of admission check do you mean, and how do validation errors get reported?

enhancements/dns/forward-dns-over-tls.md
```diff
@@ -0,0 +1,317 @@
...
+  // The default value is "" (empty) which results in a standard cleartext connection being used when forwarding DNS
+  // requests to an upstream resolver.
+  //
+  // +optional
+  TransportConfig DNSTransportConfig `json:"transportConfig,omitempty"`
+}
+```
+
+Because of the limitation in CoreDNS that you can only have one `forward` 
+block per server and the TLS configuration for a forward block is global,
```
Miciah, line 142:
```suggestion
block per server and the TLS configuration for a forward block is global for the entire block,
```
or
```suggestion
block per server and the TLS configuration for a forward block applies to all the servers in the block,
```

enhancements/dns/forward-dns-over-tls.md
brandisher: I've removed this as it's not necessary. CoreDNS logs cert related errors when trying to establish a connection.

enhancements/dns/forward-dns-over-tls.md
brandisher: @Miciah this has been fixed in the latest implementation. Any additional changes needed with regards to this comment?

Miciah: Are you saying the API does allow the user to specify an invalid configuration?  Is it feasible to revise the API not to allow that?  Alternatively, we could document this in the godoc and [add an admission check](https://github.com/openshift/kubernetes/blob/master/openshift-kube-apiserver/admission/customresourcevalidation/dns/validate_dns.go) to reject the invalid configuration.

brandisher: > Are you saying the API does allow the user to specify an invalid configuration?
Yes, `UpstreamResolvers` and `ForwardPlugin` use the same union type for defining TLS configuration so the API allows this configuration. As far as I know, there isn't a way to refine the API to not allow this because `UpstreamResolvers` allows individual upstreams to use `/etc/resolv.conf` or an IP:port combination. This block was added to throw an error if TLS is configured for `UpstreamResolvers` when one of the `Upstream` objects is configured with `SystemResolvConf`: https://github.com/openshift/cluster-dns-operator/pull/314/files#diff-167ddcf106346ad9d8ce57a77bc468db4c4f41eeace963fb8cf01952f7e18adbR223-R233

Is the method we're using to address this insufficient? If so, is the next step to add an admission check to validate the configuration ahead of time?

Miciah: It's preferable if the API doesn't allow invalid configuration to be expressed in the first place.  The next best thing is an admission check to reject invalid configuration at admission time (per my previous comment).  If the validation cannot be done in the API and absolutely must be done on admitted objects, then we can do that (and we do that with the ingress operator), but it makes for an unfortunate UX.  Significantly changing the API at this point may not be an option, so I suggest we add an admission check.  The good news is that we already have the admission plugin for `dnses.operator.openshift.io`, so we can add the additional admission check into that plugin.

brandisher: Admission check added here: https://github.com/openshift/kubernetes/pull/1247

Miciah:
Excellent! Thanks!
/approve
/lgtm

https://github.com/openshift/kubernetes/pull/1247 might warrant a follow-up update to the enhancement.

Miciah:
The linter has a new requirement:
```
 enhancements/dns/forward-dns-over-tls.md missing "### Workflow Description" 
```
The requirement wasn't there when the PR was posted, and you do provide a user story and plenty of detail, so I'll override the linter CI job.

PR #990, opened 2021-12-16
Windows Health Management enhancement

sebsoto:


enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+```json
+[
+  {
+    "name": "name of the Windows service",
+    "command": "the command that will be ran",
+    "nodeVariablesToSubstitute": {
+      "name": "string within command field that will be substituted",
+      "jsonPathNodeObject": "value that will be substituted, taken from the provided jsonPath of the Node object"
+    },
+    "powershellVariablesToSubstitute": {
```
aravindhp, line 96: Please explain why this is needed.

How is this a `powershellVariablesToSubstitute`? You are saying that the daemon will execute a PS script and substitute a command line variable with the output of the script. Or am I misunderstanding?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+      "name": "string within command field that will be substituted",
+      "path": "value that will be substituted, taken from the result of the powershell script at this path"
+    }
+  }
+]
```
aravindhp, line 101: Please provide an example.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+  {
+    "name": "name of the Windows service",
+    "command": "the command that will be ran",
+    "nodeVariablesToSubstitute": {
+      "name": "string within command field that will be substituted",
+      "jsonPathNodeObject": "value that will be substituted, taken from the provided jsonPath of the Node object"
+    },
+    "powershellVariablesToSubstitute": {
+      "name": "string within command field that will be substituted",
+      "path": "value that will be substituted, taken from the result of the powershell script at this path"
```
aravindhp, line 98: The explanation is confusing and needs to be reworded. `path` is just the location of the PS script.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+    }
+  }
+]
+```
+
+The files key contains the path and checksum of files copied to the VM by WMCO. This will be used to validate that
```
aravindhp, line 104: `files`

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+essential files have not been tampered with.
+```json
+[
+  {
+    "path": "filepath",
+    "checksum": "checksum of file"
```
aravindhp, line 110: Please mention how this is going to be generated

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+```
+
+### Node configuration: WMCO responsibilities
+
+1) For a given Machine or BYOH VM, WMCO copies the payload and WMCD binary to the VM.
+2) WMCO runs WMCD with the bootstrap command, this results in the node object being created.
+3) WMCO annotates the Node object with the `desiredVersion` annotation.
+4) The value of this is set to the current WMCO version.
+5) WMCO starts WMCD as a service.
+6) Reconciliation can successfully return at this point.
```
aravindhp, line 122: Are you sure we want to do this? How do we authoritatively inform the user that an instance has not been configured?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+  {
+    "path": "filepath",
+    "checksum": "checksum of file"
+  }
+]
+```
+
+### Node configuration: WMCO responsibilities
+
+1) For a given Machine or BYOH VM, WMCO copies the payload and WMCD binary to the VM.
```
aravindhp, line 117: A Windows instance needn't be a VM. Please replace VM with instance everywhere or some other generic verbiage.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+For node configuration WMCD will have two separate responsibilities:
+
+#### Bootstrap command
+
+When run with the bootstrap command WMCD will do the steps required to ensure that Node is created
+for the VM. This functionality is currently implemented within WMCB with the `init-kubelet` command.
+
+#### Running as a Windows service
+
+WMCD will have an informer for both ConfigMap and Node events. When it is detected that either the desired version
```
aravindhp, line 135: - How will the daemon know the desired version?
- Why does need to know about the ConfigMap changes given it is going to be immutable for a particular version?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+has been set for the Node, or the ConfigMap specified by the desired version annotation has been created, WMCD will
+configure the Windows services to the specifications listed in relevant the ConfigMap. If an error occurs, an event
+will be generated against the Node object.
+
+When configuring a service, WMCD will ensure that the service exists as specified in the ConfigMap. In order to
```
aravindhp, line 140: So does every service need the `nodeVariablesToSubstitute` entry? That is being implied in this paragraph.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+properly do this, it must evaluate the variables listed within the ConfigMap. This will be done using the Node object,
+which can be achieved by searching for a Node with an internal IP equivalent to the IP of the VM, and the powershell
+scripts copied to the VM by WMCO.
+
+On a separate thread, WMCD will be periodically polling the state of Windows services. If there is any change in the
+expected state, the services will be created or modified in order to restore expected state. As multiple threads can
+be modifying Windows service state, the appropriate thread safety procedures must be followed.
```
aravindhp, line 147: Why would there be multiple threads watching the service state?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+If it is deleted, WMCO will re-create it. If it is not present when WMCO is kicking off the configuration of a VM,
+it will be created.
+
+The services ConfigMap contains two keys: `services`, and `files`. The values for both keys are JSON objects.
+The services key contains all data required to configure the required Windows services on any VM that is to be
+added to the cluster as a Node. The proposed schema is as follows:
+```json
+[
+  {
+    "name": "name of the Windows service",
```
aravindhp, line 90: How are you expressing service dependency?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+On a separate thread, WMCD will be periodically polling the state of Windows services. If there is any change in the
+expected state, the services will be created or modified in order to restore expected state. As multiple threads can
+be modifying Windows service state, the appropriate thread safety procedures must be followed.
+
+### Node de-configuration: WMCO responsibilies
+
+When a user indicates to WMCO that it wishes to remove a BYOH node from the cluster, the following procedure will be
+kicked off.
+
+1) WMCO stops the WMCD service on the VM. The dependent services will be automatically stopped as well.
```
aravindhp, line 154: What is the service that depends on the daemon? You need to call this out in the design details.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+2) WMCO runs the deconfigure command of WMCD.
```
aravindhp, line 155: How will WMCO know when to run the deconfigure command? Is it going to wait for all services to stop? Or just the daemon?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+3) WMCO removes the files it copied to the VM.
+
+### Node de-configuration: WMCD responsibilities
+
+When the cleanup command is ran, WMCD will remove the services listed in the ConfigMap used to configure the VM.
+It will then delete the VM's Node object.
```
aravindhp, line 161: You need to call out that this responsibility is being removed from WMCO then.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+When configuring a service, WMCD will ensure that the service exists as specified in the ConfigMap. In order to
+properly do this, it must evaluate the variables listed within the ConfigMap. This will be done using the Node object,
+which can be achieved by searching for a Node with an internal IP equivalent to the IP of the VM, and the powershell
+scripts copied to the VM by WMCO.
+
+On a separate thread, WMCD will be periodically polling the state of Windows services. If there is any change in the
+expected state, the services will be created or modified in order to restore expected state. As multiple threads can
+be modifying Windows service state, the appropriate thread safety procedures must be followed.
+
+### Node de-configuration: WMCO responsibilies
```
aravindhp, line 149: `BYOH Node...`

Given this please add a note for Machine nodes.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+When a user indicates to WMCO that it wishes to remove a BYOH node from the cluster, the following procedure will be
+kicked off.
+
+1) WMCO stops the WMCD service on the VM. The dependent services will be automatically stopped as well.
+2) WMCO runs the deconfigure command of WMCD.
+3) WMCO removes the files it copied to the VM.
+
+### Node de-configuration: WMCD responsibilities
```
aravindhp, line 158: `BYOH Node...`

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+User stories can be found within the [health management epic](https://issues.redhat.com/browse/WINC-657).
+
+### API Extensions
+
+N/A
+
+### Risks and Mitigations
+
+* If WMCD has an unrecoverable error that results in the process continuously crashing, the services that it manages
+  will go down, due to them being set as dependents. With the kubelet down, the Node will be unreachable and the Ready
```
aravindhp, line 174: Do you mean the `NotReady` or whatever state that is?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+  status will reflect this. If the VMs are created through the machine-api, the user should set up MachineHealthChecks,
+  which will correct the state by recreating the Machine. For BYOH VMs, user intervention is required at this point.
+  Events and alerts will ensure that an admin is made aware of the issue.
+* There is a possibility that a cluster admin could mistakenly remove the services ConfigMap for an older WMCO version,
+  before Node upgrade is completed. This would result in WMCD not knowing what services it has configured on the VM.
+  In order to prevent this, WMCD can set the Windows service description to "openshift-managed", so that it
```
aravindhp, line 180: A cluster-admin could potentially change the service description also. How do you handle that scenario?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+  existing ability to properly configure Windows VMs to join the cluster as Windows Nodes.
+
+
+### Non-Goals
+
+* Diagnosing or remediating generic Windows issues, unrelated to changes made by WMCO.
+* Cluster-level resource management, such as deleting or re-creating WMCO-configured Machines that enter an unrecoverable state (this is a responsibility of [MachineHealthChecks](https://docs.openshift.com/container-platform/latest/machine_management/deploying-machine-health-checks.html)).
+
+
+## Proposal
```
aravindhp, line 59: The proposal should state the overall workflow and not go into design details which is why the `Design` section is empty.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+tested through the existing tests. New tests will be added to test functionality that wasn't already present, such
+as validating the state of the services ConfigMap.
+
+### Graduation Criteria
+
+This enhancement is targeted for OpenShift 4.11/ WMCO version 6.0.0.
+
+#### Dev Preview -> Tech Preview
+
+N/A
```
aravindhp, line 199: Do you want to release this for community WMCO 5.y.z?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+N/A, as the changes made in the enhancement should be invisible to the user.
+
+### Upgrade / Downgrade Strategy
+
+When WMCO is upgraded, it will create the new Windows service ConfigMap, and then for each existing Windows Node:
+1) Stop the WMCD service
+2) Copy over all new and modified files in the payload, including WMCD.
+3) Update the desired version annotation on the Node, to the current WMCO version.
+4) Start the WMCD service.
```
aravindhp, line 217: Shouldn't you deconfigure also like we are doing today?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+WMCD, triggered by a mismatch between desired node version and current node version labels will correct the state
```
aravindhp, line 219: Won't it be easier to deconfigure so that the path is similar for removal of BYOH entries from the ConfigMap and upgrade?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+by removing the Services specified in the previous version's ConfigMap, and using the ConfigMap of the version it is
+upgrading to, will configure the VM normally. As part of this process, the Node version label is patched from the
+previous version to the desired version.
+
+Downgrades are not supported through OLM, and thus are not supported by WMCO.
+
+### Version Skew Strategy
+
+This enhancement has the potential to reduce issues related to version skew with regard to Nodes, as the services
```
aravindhp, line 228: What are the version skew issues with regard to Nodes today?

This section needs to talk about how you are going to handle situations (or why these situations won't arise) like having a WMCO which is not daemon aware but the daemon is running on the nodes etc.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+N/A
+
+#### Failure Modes
+
+This remains the same, on failure, the state of Windows nodes will not be managed properly. With improved
+event generation problems will be easier to debug.
+
+#### Support Procedures
+
+The support procedures for WMCO will remain the same.
```
aravindhp, line 243: Won't you have daemon logs to gather in addition? IOW, does this enhancement require a Windows specific change to `must-gather`?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Implementation History
+
+v1: Initial Proposal
+
+## Drawbacks
+
+The only drawback to this is that complexity increases by distributing responsibilities between WMCO and WMCD.
```
aravindhp, line 251: Isn't that the case even today?

aravindhp (changes requested):
Thanks for working on this, @sebsoto. Please address my comments.

enhancements/windows-containers/health-management.md
selansen: As you mentioned earlier that WMCD cleans up all the configurations during deconfigure process, I feel we should stop all dependent services first and then stop WMCD.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+which can be achieved by searching for a Node with an internal IP equivalent to the IP of the VM, and the powershell
+scripts copied to the VM by WMCO.
+
+On a separate thread, WMCD will be periodically polling the state of Windows services. If there is any change in the
+expected state, the services will be created or modified in order to restore expected state. As multiple threads can
+be modifying Windows service state, the appropriate thread safety procedures must be followed.
+
+### Node de-configuration: WMCO responsibilies
+
+When a user indicates to WMCO that it wishes to remove a BYOH node from the cluster, the following procedure will be
```
selansen, line 151: Shouldn't we also mention how the MachineSet node is handled?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+kicked off.
+
+1) WMCO stops the WMCD service on the VM. The dependent services will be automatically stopped as well.
+2) WMCO runs the deconfigure command of WMCD.
+3) WMCO removes the files it copied to the VM.
+
+### Node de-configuration: WMCD responsibilities
+
+When the cleanup command is ran, WMCD will remove the services listed in the ConfigMap used to configure the VM.
```
selansen, line 160: is run or is ran?
who invokes this cleanup command?

enhancements/windows-containers/health-management.md
selansen: WMCD is a support service and not related to essential Kubernetes services. If it crashes but Kubernetes services are still up and running, we should let them run.  If you make them a dependent service, they will be stopped if there is an issue with WMCD.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+### API Extensions
+
+N/A
+
+### Risks and Mitigations
+
+* If WMCD has an unrecoverable error that results in the process continuously crashing, the services that it manages
+  will go down, due to them being set as dependents. With the kubelet down, the Node will be unreachable and the Ready
+  status will reflect this. If the VMs are created through the machine-api, the user should set up MachineHealthChecks,
+  which will correct the state by recreating the Machine. For BYOH VMs, user intervention is required at this point.
```
selansen, line 176: BYOH can also have Windows instance.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+  Events and alerts will ensure that an admin is made aware of the issue.
+* There is a possibility that a cluster admin could mistakenly remove the services ConfigMap for an older WMCO version,
+  before Node upgrade is completed. This would result in WMCD not knowing what services it has configured on the VM.
+  In order to prevent this, WMCD can set the Windows service description to "openshift-managed", so that it
+  can search for services and remove them. In order to facilitate upgrades from a version without WMCD, this string
+  should be added to all existing services in 4.10, one major version ahead of WMCD's target release.
```
selansen, line 182: The config map will contain all the services names. Can't we iterate them and stop all the services instead of adding "openshift-managed" description to services? adding them in 4.10 for future release seems unnecessary to me.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Design Details
```
selansen, line 184: I think we clearly call it out what should go in the proposal and what should go in the Design Details. 
This section should cover most of the details that are written in the proposal section.

selansen: I think we should talk about what are all the issues WMCD will address as part of this enhancement.
Ex: restarting service if is it not running or fixing config params If there is a mismatch.
I understand we have a general statement that says "we will fix issues and try to keep the node in desired state" but some details or examples should be discussed and how WMCD will handle the situations.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+### Test Plan
+
+Most functionality described within this enhancement is already tested as part of the extensive end-to-end
+tests in the WMCO repo. As the different commands of WMCD are developed, they can be integrated into WMCO, and
+tested through the existing tests. New tests will be added to test functionality that wasn't already present, such
+as validating the state of the services ConfigMap.
```
selansen, line 191: Are we not going to write any test case that injects some failures and validate if they get fixed by WMCD?
Existing e2e tests will fail if they run into any issues/failures. But this feature is mainly how we recover Windows nodes if there is an issue. 
We should bring up the Windows node and make it ready state. After that, we inject some failure and try to see if the node is automatically recovered by WMCD.  Also need to verify if events are triggered on a failed condition.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+of WMCO will be followed. That being a community operator release for early preview, followed by the official Red Hat
+operator release.
+
+#### Removing a deprecated feature
+
+N/A, as the changes made in the enhancement should be invisible to the user.
+
+### Upgrade / Downgrade Strategy
+
+When WMCO is upgraded, it will create the new Windows service ConfigMap, and then for each existing Windows Node:
```
selansen, line 213: Shouldn't we mention how this is handled in the MachineSet instance upgrade?

enhancements/windows-containers/health-management.md
selansen, line 243: we have to specify where WMCD logs will be stored.

selansen (changes requested):

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+Currently, Windows Node configuration process is split between two components, WMCO and [WMCB](https://github.com/openshift/windows-machine-config-operator/).
+WMCB is a one shot program, which is copied to and subsequently ran on Windows VMs, and in order to configure and
+start the Kubelet Windows Service.
+
+In order to accomplish the goals listed in this enhancement, I am proposing that WMCB be converted into a daemon,
+Windows Machine Config Daemon (WMCD). WMCD will have the following responsibilities:
+* WMCD maintains current responsibilities of bootstrapping kubelet through the `bootstrap` command.
+* Using the configuration provided by a ConfigMap created by WMCO, WMCD maintains the state of Windows Services on the VM.
+* WMCD cleans up all changes it has made to a VM when ran with the `deconfigure` command.
```
jrvaldes, line 69:
```suggestion
* WMCD `deconfigure` command revert all changes made to the Windows VM.
```

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+The Windows Node creation workflow for the end user should not be changed.
+
+### Services ConfigMap
+
+WMCO will create and maintain an immutable ConfigMap with the naming scheme:
+`services-<MajorVersion>-<MinorVersion>-<PatchVersion>-<Commit>`. For example, if WMCO was built from commit a7b5
+of WMCO version 5.0.0, the created ConfigMap would be named `config-5-0-0-a7b5`.
```
jrvaldes, line 77:
```suggestion
of WMCO version 5.0.0, the created ConfigMap would be named `services-5-0-0-a7b5`.
```
Given above pattern:
`services-<MajorVersion>-<MinorVersion>-<PatchVersion>-<Commit>`

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+The purpose of this ConfigMap is to provide WMCD with the specifications required to configure a Node. The ConfigMap
+will be watched via the existing ConfigMap controller. If it is created with incorrect values, WMCO will delete it.
+If it is deleted, WMCO will re-create it. If it is not present when WMCO is kicking off the configuration of a VM,
+it will be created.
+
+The services ConfigMap contains two keys: `services`, and `files`. The values for both keys are JSON objects.
+The services key contains all data required to configure the required Windows services on any VM that is to be
```
jrvaldes, line 85: s/services/`services`

jrvaldes (commented):
@sebsoto  This is good, thanks for working on it. Comments added PTAL.

enhancements/windows-containers/health-management.md
jrvaldes: Interesting. @selansen are you saying that WMCD should not control `kubelet` service state in general or just while facing an error ?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+### API Extensions
+
+N/A
+
+### Risks and Mitigations
+
+* If WMCD has an unrecoverable error that results in the process continuously crashing, the services that it manages
+  will go down, due to them being set as dependents. With the kubelet down, the Node will be unreachable and the Ready
+  status will reflect this. If the VMs are created through the machine-api, the user should set up MachineHealthChecks,
```
jrvaldes, line 175:
```suggestion
  status will reflect this. For VMs created through the machine-api, the user must set up MachineHealthChecks,
```

enhancements/windows-containers/health-management.md
jrvaldes, line 176: > user intervention

Need reference or definition.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+  which will correct the state by recreating the Machine. For BYOH VMs, user intervention is required at this point.
+  Events and alerts will ensure that an admin is made aware of the issue.
```
jrvaldes, line 177:
```suggestion
  Events and alerts will ensure that the cluster administrator is made aware of the issue.
```
Don't see how "Events and alerts" can ensure that.

"Events and alerts" just provide information to the cluster administrator.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+#### Tech Preview -> GA
+
+As the functionality described in this enhancement will be integrated into WMCO, the normal release process
+of WMCO will be followed. That being a community operator release for early preview, followed by the official Red Hat
+operator release.
+
+#### Removing a deprecated feature
+
+N/A, as the changes made in the enhancement should be invisible to the user.
```
jrvaldes, line 209: Should we call out WMCB's repo and submodule removal?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+The purpose of this enhancement is to ensure that all Windows nodes managed by the Windows Machine Config Operator (WMCO)
+maintain a usable state. With health management implemented, WMCO will automatically ensure that OpenShift related services 
+running on the underlying Windows VMs have a status and runtime configuration inline with the expectations of the installed operator version.
+
+## Motivation
```
aravindhp, line 33: Added motivation is to ease the addition of newer services. I also realize this will make enabling 3rd party networks easier. If we detect a supported 3rd party network, the immutable ConfigMap changes to reflect that i.e. hybrid-overlay service will not be present etc.

enhancements/windows-containers/health-management.md
sebsoto: Yes, but I do feel that the responsibilities of WMCB were extremely small compared to WMCO.

enhancements/windows-containers/health-management.md
saifshaikh48: Machine nodes aren't deconfigured, their associated machines are simply deleted, so there would be no change here

enhancements/windows-containers/health-management.md
saifshaikh48: Is this implying moving the discussion around combining WMCO/WMCB into the Design Details section?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+    "dependencies": [
+      "name of a service that this service is dependent on"
+    ],
+    "nodeVariablesToSubstitute": {
+      "name": "string within command field that will be substituted",
+      "jsonPathNodeObject": "The jsonPath of a field within the instance's Node object"
+    },
+    "powershellVariablesToSubstitute": {
+      "name": "string within command field that will be substituted",
+      "path": "location of the PowerShell script to be ran, in order to get it's output"
```
saifshaikh48, line 133: nit: `it's` --> `its`

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+name of the instance's Node, and the output of the PowerShell script located at `C:\k\scripts\get_net_ip.ps`, as the
+value of the argument --variable-arg2. WICD will be aware of the proper Node object, as it searching for a Node with an
+internal IP equivalent to the IP of the instance.
+This service has no dependencies, and should be created only after all services with priority 1 have been created.
+
+The `files` key contains the path and checksum of files copied to the instance by WMCO. This will be used to validate
+that essential files have not been tampered with.
+```json
+{
+  "path": "filepath",
```
aravindhp, line 174: Is this the path on the Windows node? What about the source location within the WMCO image?

enhancements/windows-containers/health-management.md
aravindhp: >At this point WMCO has done everything required of it so it seems right for it to finish reconciliation.
It can ssh into the instance, correctly copy files, has created the Node, and successfully started the daemon.
Re-queuing should not cause a change in state.

How does it create the Node? Doesn't that happen as a result of the kubelet running?

>This would impact our ability to scale, with the trade off that the error would be logged to WMCO's logs. At the moment I'm not convinced that that's worth it.

We shouldn't worry about scaling in this enhancement as improving scaling is not one of the end goals.

My concern is that if we are doing a fire and forget approach, how do we guarantee that only one node is going to be upgraded at one time like we have today?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+  point. Events and alerts will ensure that a cluster administrator is made aware of the issue.
+* There is a possibility that a cluster admin could mistakenly remove the services ConfigMap for an older WMCO version,
+  before Node upgrade is completed. This would result in WICD not knowing what services it has configured on the instance.
+  In order to prevent this, WICD can set the Windows service description to "openshift-managed", so that it
+  can search for services and remove them. In order to facilitate upgrades from a version without WICD, this string
+  should be added to all existing services in 4.10, one major version ahead of WICD's target release.
+
+## Design Details
+
+### Services ConfigMap
```
aravindhp, line 104: With services like hybrid-overlay, WMCO waits for HNS networks and annotations on node objects to be present before proceeding. How will that be handled / expressed in the ConfigMap?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+1) For a given Machine or BYOH instance, WMCO copies the payload to the instance.
+2) WMCO runs WICD with the bootstrap command, this results in the node object being created.
+3) WMCO annotates the Node object with the `desiredVersion` annotation.
+4) The value of this is set to the current WMCO version.
+5) WMCO starts WICD as a service.
+6) Reconciliation can successfully return at this point.
+
+### Node configuration: WICD responsibilities
+
+For node configuration WICD will have two separate responsibilities:
```
aravindhp, line 190: `Node`

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+
+#### Bootstrap command
+
+When run with the bootstrap command WICD will do the steps required to ensure that Node is created
+for the instance. This functionality is currently implemented within WMCB with the `init-kubelet` command.
+The Node will be cordoned, as the Node is not usable at this point.
```
aravindhp, line 196: How will the Node be cordoned if it has not been created?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+
+#### Running as a Windows service
```
aravindhp, line 198: What will cause WICD to run as a service? If you are introducing another command it should be called out.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+6) Reconciliation can successfully return at this point.
+
+### Node configuration: WICD responsibilities
+
+For node configuration WICD will have two separate responsibilities:
+
+#### Bootstrap command
+
+When run with the bootstrap command WICD will do the steps required to ensure that Node is created
+for the instance. This functionality is currently implemented within WMCB with the `init-kubelet` command.
```
aravindhp, line 195: Are you going to rename `init-kubelet` to something else?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+The Node will be cordoned, as the Node is not usable at this point.
+
+#### Running as a Windows service
+
+WICD will have an informer for both ConfigMap and Node events. When it is detected that either the desired version
+annotation has been set for the Node, or the ConfigMap specified by the desired version annotation has been created,
```
aravindhp, line 201: >When it is detected that either the desired version annotation has been set for the Node

I don't follow why the desired annotation is required. Please elaborate on that.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+  "path": "filepath",
+  "checksum": "checksum of file, generated at compile time"
+}
+```
+
+### Node configuration: WMCO responsibilities
+
+1) For a given Machine or BYOH instance, WMCO copies the payload to the instance.
+2) WMCO runs WICD with the bootstrap command, this results in the node object being created.
+3) WMCO annotates the Node object with the `desiredVersion` annotation.
```
aravindhp, line 183: Please explain why this `desiredVersion` is required. 

How do you handle a user stomping on the `desiredVersion`?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+1) Stop the WICD service
+2) Run the WICD `cleanup` command with the `--upgrade` flag to stop and remove all services without removing the Node
+   object.
+2) Copy over all new and modified files in the payload, including WICD.
+3) Update the desired version annotation on the Node, to the current WMCO version.
+4) Start the WICD service.
+
+WICD, triggered by a mismatch between desired node version and current node version labels will correct the state
+by using the ConfigMap of the version it is upgrading to, in order to configure the instance normally. As part of this
+process, the Node version label is patched from the previous version to the desired version.
```
aravindhp, line 265: You need to at least briefly call this out when you first introduce the desired version annotation.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+
+### Version Skew Strategy
+
+This enhancement has the potential to reduce issues related to having Windows Nodes which were configured using an
+older, now uninstalled WMCO version, on a more recent OpenShift cluster. When a compatible WMCO version is installed,
+the services ConfigMap provides a way for WICD to upgrade a Node cleanly from one version to another. This is true when
+upgrading only from WMCO 5.0.0 onwards, as WICD will have no way to cleanly upgrade without either a service ConfigMap
+from the previous version, or the `openshift-managed` label on the Windows services themselves.
+
+Any Windows Nodes from before 5.0.0 should have been deconfigured with the associated WMCO version. At this point, the
```
aravindhp, line 279: This is a breaking change and is not acceptable. Why can't WMCO do the deconfiguration as it is doing today for this version? We can remove the WMCO deconfiguration in 6.0.0.

aravindhp (changes requested):

enhancements/windows-containers/health-management.md
selansen: >> Interesting. @selansen are you saying that WMCD should not control kubelet service state in general or just while facing an error ?
Just saying when WMCD crashes and other services like kubelet are running without any issue, why should we stop them. Then node will go to `not ready` state.

enhancements/windows-containers/health-management.md
selansen: coming back from containerd document comments, it is better to mention that info in the document.

enhancements/windows-containers/health-management.md
sebsoto: I misworded this. I meant going into 6.0.0 any Nodes from before 5.0.0 should have been deconfigured. I will reword it.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+
+### Services ConfigMap
+
+WMCO will create and maintain an immutable ConfigMap with the naming scheme:
+`services-<MajorVersion>-<MinorVersion>-<PatchVersion>-<Commit>`. For example, if WMCO was built from commit a7b5
+of WMCO version 5.0.0, the created ConfigMap would be named `services-5-0-0-a7b5`.
+
+The purpose of this ConfigMap is to provide WICD with the specifications required to configure a Node. The ConfigMap
+will be watched via the existing ConfigMap controller. WMCO is the owner of the ConfigMap and the source of the data
+present within it. If an entity attempts to create a services ConfigMap with incorrect values, WMCO will delete it.
```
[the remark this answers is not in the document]
aravindhp: What I don't understand is how would there be a ConfigMap with incorrect values given that to create such an incorrect ConfigMap it would have to be deleted first?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,258 @@
...
+
+The services ConfigMap contains two keys: `services`, and `files`. The values for both keys are JSON objects.
+The services key contains all data required to configure the required Windows services on any VM that is to be
+added to the cluster as a Node. The proposed schema is as follows:
+```json
+[
+  {
+    "name": "name of the Windows service",
+    "command": "the command that will be ran",
+    "nodeVariablesToSubstitute": {
```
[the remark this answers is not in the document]
aravindhp: This has not been addressed.

I also think `nodeVariablesInCommand` but might more appropriate. It provides a some linkage back to `command`. IOW, some mention of `command` might help make the connection easier. WDYT?

enhancements/windows-containers/health-management.md
aravindhp: Again `powershellVariablesInCommand` might be more appropriate.

enhancements/windows-containers/health-management.md
aravindhp: The confusion here is who is the consumer of the `files` section. Is WICD going to use this to perform checksum validation before it configures the service? That needs to be called out here. 

I am assuming WMCO will continue doing checksum validation upon copying the files.

enhancements/windows-containers/health-management.md
aravindhp: Please call that out here.

enhancements/windows-containers/health-management.md
aravindhp: I prefer the `daemon` or `service` given WICD is a Windows service.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,325 @@
...
+
+#### Bootstrap command
+
+When run with the bootstrap command WICD will do the steps required to ensure that Node is created
+for the instance. This functionality is currently implemented within WMCB with the `init-kubelet` command.
+WICD will wait for the Node to be created, and cordon it, as the Node is not ready for workloads at this point.
+
+#### Running as a Windows service
+
+WICD will fullfil its responsibilities as a Windows service controller when ran with the `controller` command.
```
aravindhp, line 208: Lets stick with `when run`

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,325 @@
...
+    "dependencies": [
+      "name of a service that this service is dependent on"
+    ],
+    "nodeVariablesToSubstitute": {
+      "name": "string within command field that will be substituted",
+      "jsonPathNodeObject": "The jsonPath of a field within the instance's Node object"
+    },
+    "powershellVariablesToSubstitute": {
+      "name": "string within command field that will be substituted",
+      "path": "location of the PowerShell script to be ran, in order to get its output"
```
aravindhp, line 133: `to be run`

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,325 @@
...
+
+This enhancement has the potential to reduce issues related to having Windows Nodes which were configured using an
+older, now uninstalled WMCO version, on a more recent OpenShift cluster. When a compatible WMCO version is installed,
+the services ConfigMap provides a way for WICD to upgrade a Node cleanly from one version to another. This is true when
+upgrading only from WMCO 5.0.0 onwards, as WICD will have no way to cleanly upgrade without either a service ConfigMap
+from the previous version, or the `openshift-managed` label on the Windows services themselves.
+
+Any Windows Nodes from before 5.0.0 should have been deconfigured with the associated WMCO version. If WMCO 6.0.0+ is
+installed, the stale instances should be removed manually and should be freshly recreated before attempting to add
+them to the OpenShift cluster.
```
aravindhp, line 295: Are you saying that when WMCO 5.y.z is upgraded to 6.y.z, the instances configured by 5.y.z are considered stale? What is the definition of stale instances here?

aravindhp (changes requested):

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,311 @@
...
+
+### Risks and Mitigations
+
+* It is possible that issues with Windows instances will occur which WICD is unable to resolve, causing the Node to
+  enter a `NotReady` state. For instances created through the machine-api, the user must set up MachineHealthChecks,
+  which will correct the state by recreating the Machine. For BYOH instances, user intervention is required at this
+  point. Events and alerts will ensure that a cluster administrator is made aware of the issue.
+* There is a possibility that a cluster admin could mistakenly remove the services ConfigMap for an older WMCO version,
+  before Node upgrade is completed. This would result in WICD not knowing what services it has configured on the instance.
+  In order to prevent this, WICD can set the Windows service description to "openshift-managed", so that it
```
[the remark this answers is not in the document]
aravindhp: I am fine with that. Please call that out.

enhancements/windows-containers/health-management.md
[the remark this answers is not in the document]
aravindhp: Thank you for the explanation. This bit of detail will be useful to have in the enhancement.

enhancements/windows-containers/health-management.md
aravindhp: I still don't understand how you will handle [waiting for HNS networks](https://github.com/openshift/windows-machine-config-operator/blob/master/pkg/windows/windows.go#L431) which is done by WMCO today.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,329 @@
...
+]
+```
+
+The nodeVariablesToSubsitute and powershellVariablesinCommand fields will be used by WICD when processing the
+ConfigMap. They are needed as some Windows services have arguments that require values sourced from either the
+Node object associated with the instance, or data from the instance itself. These fields allow for a ConfigMap that is
+generic enough to apply to all Windows instances being configured, by providing WICD instructions for how to retrieve
+certain values. For these fields, a retry period will be necessary for when the information is not available yet. For
+example, when waiting for an annotation to be applied to the Node object. When making use of a PowerShell script,
+retries should be present within the script, if necessary.
```
aravindhp, line 147: What about with `nodeVariablesCommand`? Won't we have to wait for those too?

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,329 @@
...
+
+This enhancement has the potential to reduce issues related to having Windows Nodes which were configured using an
+older, now uninstalled WMCO version, on a more recent OpenShift cluster. When a compatible WMCO version is installed,
+the services ConfigMap provides a way for WICD to upgrade a Node cleanly from one version to another. This is true when
+upgrading only from WMCO 5.0.0 onwards, as WICD will have no way to cleanly upgrade without either a service ConfigMap
+from the previous version, or the `openshift-managed` label on the Windows services themselves.
+
+Any Windows Nodes from before 5.0.0 should have been deconfigured with the associated WMCO version. If WMCO 6.0.0+ is
+installed, any stale instances from 4.y.z or below should be removed manually and should be freshly recreated before
+attempting to add them to the OpenShift cluster.
```
aravindhp, line 299: 4.y.z or below??? For one you need to mention the version explicitly. I still don't understand how there will be stale instances given we don't allow skipping major WMCO upgrades.

aravindhp (changes requested):

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,329 @@
...
+maintain an expected state. With health management implemented, WMCO will automatically ensure that OpenShift related
+services running on the underlying Windows instances are configured, and have a state inline with the
+expectations of the installed WMCO version.
+
+## Motivation
+
+Currently [WMCO](https://github.com/openshift/windows-machine-config-operator/) properly fulfills its goal of
+configuring Windows instances into worker Nodes. Once the Node object exists, barring specific conditions, WMCO does
+not have the functionality to detect and correct changes in Node state. Windows nodes should be resilient enough to not
+require manual intervention when an issue occurs, who's cause is within the scope of WMCO's control. This work serves
```
alinaryan, line 38: who's ----> whose

enhancements/windows-containers/health-management.md
sebsoto: I call that out: `For example, when waiting for an annotation to be applied to the Node object`

enhancements/windows-containers/health-management.md
sebsoto: I understand now, that is a command that we run which does not have an output used in a Windows service.

That wait will covered as are retrying for the `distributed-router-gateway-mac` Node annotation, which is only placed on the Node [once the networks have been created](https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/hybrid-overlay/pkg/controller/node_windows.go#L309).
Additionally we are [creating an endpoint](https://github.com/openshift/windows-machine-config-operator/blob/bb5f6d48a340569fd203d2de12237df4d445fb57/pkg/windows/windows.go#L830) which requires the HNS network, so that check should be built into the PowerShell script which covers that. I'll make a note of that in the enhancement.

enhancements/windows-containers/health-management.md
aravindhp: I don't follow how you are going to be upgrading one node at a time with this fire and forget approach.

aravindhp (changes requested):

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,337 @@
...
+
+The purpose of this ConfigMap is to provide WICD with the specifications required to configure a Node. The ConfigMap
+will be watched via the existing ConfigMap controller. WMCO is the owner of the ConfigMap and the source of the data
+present within it. If the ConfigMap is deleted, WMCO will re-create it.
+
+If an entity attempts to create a services ConfigMap with incorrect values, WMCO will delete it.
+This will resolve a scenerio in which the ConfigMap controller is busy configuring Windows instances, and a user
+deletes and recreates the service ConfigMap with incorrect values, before the controller has a chance to re-create
+the ConfigMap.
+
```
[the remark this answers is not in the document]
sebsoto: Eventual consistency, when it is no longer busy it will reconcile to the correct state.

enhancements/windows-containers/health-management.md
sebsoto: Added more detail in the upgrage section
```
 3) Update the desired version annotation on the Node, to the current WMCO version. This should be done one Node as a
-   time, in order to ensure that workloads can continue to run with minimal interruption.
+   time, in order to ensure that workloads can continue to run with minimal interruption. WMCO is okay to change a
+   Node's desired version annotation if: there are zero Nodes with a mismatched desired and current version; there
+   are zero NotReady Nodes which have a current version matching WMCO's version.
```

enhancements/windows-containers/health-management.md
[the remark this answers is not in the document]
aravindhp: But what if WICD (on a Node  which is not being processed) which also looks at this ConfigMap reacts to the incorrect changes?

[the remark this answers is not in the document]
sebsoto: The same result, the instance could be temporarily set up with an incorrect configuration, but it will be reverted when the ConfigMap is changed, and there will be eventual consistency.

There shouldn't be an expectation of WMCO to prevent these scenarios.

What can be done, is we can reduce the impact by allowing parallel reconciliations within the ConfigMap controller. This will reduce the reaction time of WMCO, allowing for the BYOH ConfigMap events to be processed at the same time as service events. This doesn't reduce the chance of this happening to zero, and there is nothing stopping a user from setting up an infinite loop of delete and create API calls to continuously attempt to set the ConfigMap how they want. RBAC is the proper way to deal with this, users should not be granted permissions to change the ConfigMap. In the end, if the cluster Admin wishes to mess with OpenShift's functionality with this level of disruption, that's on them.

Would allowing parallel reconciliations within the ConfigMap controller be enough of a mitigation in your opinion?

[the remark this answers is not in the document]
aravindhp: >What can be done, is we can reduce the impact by allowing parallel reconciliations within the ConfigMap controller. 

Why should it be parallel reconciliations? You can have a semaphore that controls the number of reconciliations. Or have a separate thread outside of the controller that watches for deletion events? Either way this needs to be called out as a risk with a mitigation.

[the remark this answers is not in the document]
dhellmann: Are there other examples of controllers deleting user-provided configuration instead of ignoring it and surfacing the error?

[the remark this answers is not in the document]
aravindhp: CVO 😃 The services `ConfigMap` is similar to a CVO managed manifest. We expect it to look a certain way and will stomp on any user introduced changes.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,349 @@
...
+N/A, as the changes made in the enhancement should be invisible to the user.
+
+### Upgrade / Downgrade Strategy
+
+When WMCO is upgraded, it will create the new Windows service ConfigMap, and then for each existing Windows Node:
+1) Stop the WICD service
+2) Run the WICD `cleanup` command with the `--upgrade` flag to stop and remove all services without removing the Node
+   object.
+2) Copy over all new and modified files in the payload, including WICD.
+3) Update the desired version annotation on the Node, to the current WMCO version. This should be done one Node as a
```
aravindhp, line 297: `one Node at a`

aravindhp (approved):
Nice work, @sebsoto @saifshaikh48 

/approve

aravindhp:
/hold

to allow other folks to review and prevent the next lgtm causing it to merge.

enhancements/windows-containers/health-management.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Non-Goals
+
+* Diagnosing or remediating generic Windows issues, unrelated to changes made by WMCO.
+* Cluster-level resource management, such as deleting or re-creating WMCO-configured Machines that enter an unrecoverable state (this is a responsibility of [MachineHealthChecks](https://docs.openshift.com/container-platform/latest/machine_management/deploying-machine-health-checks.html)).
+
+
+## Proposal
+
+Currently, Windows Node configuration process is split between two components, WMCO and [WMCB](https://github.com/openshift/windows-machine-config-operator/).
```
alinaryan, line 61: This links to WMCO

alinaryan:
Great work!
/lgtm

selansen:
Awesome work @sebsoto @saifshaikh48 
/lgtm

PR #992, opened 2021-12-23
WIP: auth: add enhancement for direct kube-apiserver oidc config

stlaz:
Adds an enhancement that describes how OIDC could be configured directly for the kube-apiserver as a part of `TechPreviewNoUpgrade` feature set.

/assign @deads2k 

### DIFF
--- a/enhancements/authentication/direct-use-of-oidc.md
+++ b/enhancements/authentication/direct-use-of-oidc.md
@@ -1,0 +1,450 @@
+---
+title: direct-use-of-oidc
+authors:
+  - "@stlaz"
+reviewers:
+  - "@s-urbaniak"
+  - "@slaskawi"
+  - "@deads2k"
+approvers:
+  - "@deads2k"
+api-approvers: []
+creation-date: yyyy-mm-dd
+last-updated: yyyy-mm-dd
+tracking-link: []
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Direct use of OIDC as an identity provider
+
+## Summary
+
+When users authenticate to OpenShift, the authentication goes through the `oauth-server`
+which serves as middle-man actor that unifies access to multiple kinds of
+identity providers, such as LDAP, OIDC providers, providers returning HTTP Basic
+Authentication challenges, and similar.
+
+The above differs from vanilla Kubernetes where authentication is dealt directly
+by the `kube-apiserver`. This has limitations especially when it comes to being
+able to login in a uniform manner. On the other hand, `kubectl`, the binary for
+accessing Kubernetes APIs, has evolved through the years to simplify the login
+process to Kubernetes clusters, and communities built useful toolings around
+the binary.
+
+## Motivation
+
+Due to the `oauth-server` being the authentication middle man minting
+OpenShift-specific access tokens, it is impossible to use tokens minted by a
+3rd party OIDC provider directly to authenticate to OpenShift clusters as such
+a token is forwarded directly to the `kube-apiserver` and even if it weren't and
+went directly to the `oauth-server`, there exists no such mechanism to for the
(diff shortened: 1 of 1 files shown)


enhancements/authentication/direct-use-of-oidc.md
```diff
@@ -0,0 +1,450 @@
...
+own OIDC provider.
+
+### Goals
+
+- allow directly configuring OIDC-related flags in `kube-apiserver`, guarded
+  by a  `TechPreviewNoUpgrade` feature gate
+- describe expectations that an OIDC provider needs to fulfill in order for its
+  integration with OpenShift to be seamless
+
+### Non-Goals
```
deads2k, line 60: is co-existence with the built-in oauth server a goal or a non-goal?  I think suspect a non-goal and in fact we explicitly want to disable the internal oauth-server.  If that's the case, list the disabling as a goal and co-existence as a non-goal

enhancements/authentication/direct-use-of-oidc.md
```diff
@@ -0,0 +1,450 @@
...
+- If the `KubeOIDCConfiguration` is disabled, any previous `kube-apiserver` OIDC-related
+  flags get removed from the configuration and the `KubeOIDC` field of the 
+  `authentication.config.openshift.io/cluster` resource is ignored
+
+Additionally to the above, the configmap from `authentication.config.openshift.io/cluster`
+`KubeOIDC.CAFile` should get synchronized from the `openshift-config` namespace
+to `openshift-kube-apiserver` namespace if the `KubeOIDCConfiguration` feature
+gate is enabled.
+
+// TODO: decide how involved should the `oauth-server` be when OIDC is configured
```
deads2k, line 158: this cannot be a TODO, it must be resolved before we mark this as implementable. We *will* face questions.  Given previously stated goals, I suggest co-existence as an anti-goal and disabling the existing as a goal.

enhancements/authentication/direct-use-of-oidc.md
```diff
@@ -0,0 +1,450 @@
...
+Authentication challenges, and similar.
+
+The above differs from vanilla Kubernetes where authentication is dealt directly
+by the `kube-apiserver`. This has limitations especially when it comes to being
+able to login in a uniform manner. On the other hand, `kubectl`, the binary for
+accessing Kubernetes APIs, has evolved through the years to simplify the login
+process to Kubernetes clusters, and communities built useful toolings around
+the binary.
+
+## Motivation
```
deads2k, line 36: I think one of our primary motivation is allowing oidc providers to see the gaps in the power they provide to openshift clusters as opposed to considering this a viable replacement. This is necessary to provide in order for the gaps to become obvious to the oidc providers so they can address the shortcomings.

enhancements/authentication/direct-use-of-oidc.md
```diff
@@ -0,0 +1,450 @@
...
+  "userInfo.Extra" fields of the authentication response so even if the 3rd party
+  OIDC identity provider allows custom scopes, there is currently no way for the
+  OpenShift-specific `scopeAuthorizer` to learn of these.
+- Any component that relied on an OpenShift-specific retrieval of user information
+  (==> `get` on `User` with a given accesstoken) would need to be rewritten to
+  either use the OIDC token's claims or retrieve that information from the UserInfo
+  endpoint of the OIDC identity provider.
+- In order for `oc`, `oauth-proxy` and OpenShift web console to work with the given
+  3rd party OIDC identity provider, the user must configure the `oauthMetadata` of
+  the `authentication.config.openshift.io/cluster` resource so that these clients are
```
slaskawi, line 193: It's a matter of setting up proper documentation, right?

slaskawi (commented):
Great idea and nice design. The only bit that I miss is the migration. We need to be careful here.

PR #994, opened 2021-12-23
WIP: auth: add enhancement about pinning SCCs to pods

stlaz:
Adds an enhancement that should help with pinning SCCs to core workloads in order to prevent these from breaking.

/assign @s-urbaniak 

### DIFF
--- a/enhancements/authentication/scc-pinning.md
+++ b/enhancements/authentication/scc-pinning.md
@@ -1,0 +1,327 @@
+---
+title: scc-pinning
+authors:
+  - "@stlaz"
+reviewers:
+  - "@s-urbaniak"
+  - "@deads2k"
+  - "@sttts"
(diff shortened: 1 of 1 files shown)


enhancements/authentication/scc-pinning.md
```diff
@@ -0,0 +1,327 @@
...
+## Proposal
+
+### User Stories
+
+1. I am an owner of a core OpenShift workload and I want to be able to hint which
+   SCC should the SCC admission use in order to evaluate my workload's security context
+
+### API Extensions
+
+The SCC admission is currently using the "openshift.io/scc" annotation to denote which
```
s-urbaniak, line 59: this could be potentially racy. shouldn't we have a separate dedicated annotation which expresses the intent to use an SCC?

PR #996, opened 2022-01-05
Enhancement for enabling single-node-openshift day2 workers expansion

omertuc:


enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,282 @@
...
+
+# Single Node OpenShift Day-2 Workers
+
+## Summary
+
+This enhancemnet aims to enable adding workers to a single-node cluster by
+dealing with a "floating ingress" issue encountered in UPI-installed single
+control-plane node clusters which get worker nodes added to them. It does so by
+adjusting the installer to pin the default `IngressController` to the master
+pool when installing single-node bootstrap-in-place clusters.
```
eranco74, line 33: Why limit to bootstrap-in-place?

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,290 @@
...
+  The baremetal platforms solves this issue with virtual IP addresses/keepalived.
+- Deal with users who want their ingress traffic to not go through the single
+  control-plane node.
+
+## Proposal
+
+Modify the installer's behavior during Single Node bootstrap-in-place
+installation to create an `IngressController` CR installer manifest targetting
+control-plane nodes, which will prevent the OpenShift Ingress Operator from creating
+the default `IngressController` CR which targets worker nodes. This will ensure
```
eranco74, line 125: Perhaps the Ingress Operator should look at the controlPlaneTopolgy and create the `IngressController` CR with the right label selector in case of SingleReplica?

enhancements/single-node/single-node-openshift-day2-workers.md
omertuc: It's mentioned a few times in the rest of the enhancement, but when installing IPI on e.g. AWS you don't suffer from this problem since you have a LoadBalancer type service that follows the node holding router-default automatically

enhancements/single-node/single-node-openshift-day2-workers.md
omertuc: This is addressed in the alternatives section, looking at the topology is insufficient

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,290 @@
...
+replaces:
+superseded-by:
+---
+
+# Single Node OpenShift Day-2 Workers
+
+## Summary
+
+This enhancemnet aims to enable adding workers to a single-node cluster by
+dealing with a "floating ingress" issue encountered in UPI-installed single
```
tsorya, line 30: is it only "floating ingress" issue? other operators are fine with it? Shouldn't we just set all the "control-plane" operators to run on master in case of SNO?

omertuc: If other operators get pods on the worker node, why would that be a problem?

enhancements/single-node/single-node-openshift-day2-workers.md
eranco74: I see, so it's not about how you do the bootstrap it's about the platform.
Consider rephrasing `bootstrap-in-place` to None platform

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,290 @@
...
+  Any future important additions to the default `IngressController` created by
+  the OpenShift Cluster Ingress Operator would also have to be added to the
+  installer code where it generated the manifest for single-node
+  bootstrap-in-place installations.
+
+- TBD
+
+## Alternatives
+
+- Implement this change in the OpenShift Cluster Ingress Operator rather than the installer.
```
eranco74, line 262: It's already looking at it in order to decide the replica count
https://github.com/openshift/cluster-ingress-operator/pull/519

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,290 @@
...
+  I'm not sure how trivial it is for the OpenShift Cluster Ingress Operator to know that we're
+  doing a single-node bootstrap-in-place installation. It couldn't just inspect the
+  control-plane and infrastructure topology fields, because those have the same value
+  whether you're on AWS (which doesn't suffer from the problems described in this enhancement)
+  or a UPI bootstrap-in-place installation. Maybe it could also look at the "platform"
+  field in the infrastructure CR, then only do this if the platform is "none".
```
eranco74, line 268: Beside the fact that SNO on AWS platform isn't supported, we should probably strive to the same Ingress behavior in all platforms.
So if we decide to pin the ingress to the control plane node in none platform we should do the same for other platforms.

enhancements/single-node/single-node-openshift-day2-workers.md
omertuc: Right, but as mentioned, it's not sufficient to just look at the replica count, we need to look at the platform as well. Because in cloud with a LB or on baremetal with keepalived there's no point in pinning to the master pool.

Anyway, whether the IngerssController change happens in the installer or in the ingress operator is pretty arbitrary, I don't really care either way - I'm mostly interested to discuss whether the idea makes sense in general

enhancements/single-node/single-node-openshift-day2-workers.md
omertuc: > Beside the fact that SNO on AWS platform isn't supported

It might be supported in the future so I wouldn't like to dismiss it from the discussion entirely

> we should probably strive to the same Ingress behavior in all platforms.
So if we decide to pin the ingress to the control plane node in none platform we should do the same for other platforms.

Not sure that I agree with this. I'd rather ingress to be as flexible as possible when the circumstances allow it, and only force it to a smaller subset of nodes (the SNO node) when it's necessary. 

I'd argue that it's more important that SNO remains similar to multi-node as much as possible, rather than remain similar to itself across different platforms. And by pinning just when it's necessary you're closer to the former than the latter

omertuc:
>I see, so it's not about how you do the bootstrap it's about the platform.
Consider rephrasing bootstrap-in-place to None platform

Yes. I'll revise the enhancement to make this clearer

omertuc:
Enhancement updated

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,290 @@
...
+title: single-node-openshift-day2-workers
+authors:
+  - "@omertuc"
+reviewers:
+  - "@romfreiman"
+  - "@eranco74"
+  - "@tsorya"
+  - TBD
+approvers:
+  - TBD
```
dhellmann, line 11: It would be good to have someone from the cluster infra team included as reviewer at least. @JoelSpeed can help identify someone.

Which other teams are affected by this work? @sinnykumari do you anticipate any MCO impact?

romfreiman: @dhellmann of course. We wanted to go over it first internally (even I didnt have time to review it so far) and than share it broadly for feedback and comments.
I will affect also networking and installer teams. We'll ask for more feedback early next week.

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,290 @@
...
+
+This enhancement does not modify/add any API
+
+### Implementation Details/Notes/Constraints
+
+This enhancement can be easily implemented by adjusting the installer's
+`generateDefaultIngressController` method such that when `config.platform.none
+!= nil`, it will return an `IngressController` CR struct with
+`.spec.nodePlacement.nodeSelector.matchLabels` targetting
+`node-role.kubernetes.io/master`.
```
JoelSpeed, line 150: Is there a way to also detect that it's single node, at the moment I think this would also affect HA clusters that are platform none wouldn't it?

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,290 @@
...
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+In the non-goals section it's mentioned that this enhancement does not apply to
+clusters which have been installed prior to the enhancement, so their upgrade
+is not discussed.
```
JoelSpeed, line 216: Is there a way that existing clusters can modify their cluster to match the outcome of this enhancement?

enhancements/single-node/single-node-openshift-day2-workers.md
omertuc: You're right, introduced this error during my last enhancement commit. Will fix

enhancements/single-node/single-node-openshift-day2-workers.md
omertuc: Yes, see note in non-goals

>(they can be taken care of with a simple documentation on how it can be
  enabled with a few manual steps)

omertuc: All they would have to do is adjust their IngressController from `worker` to `master` I believe

JoelSpeed: Is it worth including a short example of how this can be done within the enhancement?

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,291 @@
+---
+title: single-node-openshift-day2-workers
```
romfreiman, line 2: @omertuc I think we should structure this enhancement slightly different:
1. It should focus on adding workers as day2, and the challenges with it (user has to configure external dns/load balancer in advance
2. Then we can suggest the ingress change as an alternative
3. Also, we should add to non-goals an installation of master + worker, wdyt?
4. I think we should also cover here the infrastructure topology change that will always be single replica in such flow?

omertuc: 1/2. But then that wouldn't be an enhancement, needing a load balancer to deal with the floating ingress issue is already the state of how things work currently when you add workers to a single node cluster
3. This is already the third bullet in the non-goals section
4. Nothing has to change with how the topology variables currently work if we focus on just day2. It will only be an issue if we try to deal with day-1 installation, in that case if there's more than one worker then the infra topology will change to highly available and we need to think how that's going to affect things

fakebanana:
A SNO cluster's control plane can't scale beyond a single node - how do we ensure that adding additional worker nodes (and presumedly more pods to manage) won't overload the control plane? What test do we have to identify the failure modes which obtain when the resources allocated to the control plane are exceeded. Do we have tests that exercise the systems which recover the control plane when these limits are exceeded, and the alerts which should fire when these systems fail to operate as intended (and do the alerts fire if the control plane is down?)

omertuc:
> A SNO cluster's control plane can't scale beyond a single node - how do we ensure that adding additional worker nodes (and presumedly more pods to manage) won't overload the control plane? 

That's a good point. 

For now according to https://issues.redhat.com/browse/MGMT-8414 (specifically "Intent is to focus on small scale sites where adding 1-3 worker nodes would be beneficial.") the goal is addressing the addition of a small number of worker nodes (this should have been and will be clarified in the enhancement goals / non goals section). 

I think we'll have to test those limits in practice, and maybe set a limit on how many pods are allowed on these additional nodes (either hard limit or documentation limit, i.e. "You may add additional worker nodes but the total number of pods across all nodes in the cluster should not exceed current advertised single node limits"). I'll give it some thought and address it in the enhancement. 

>What test do we have to identify the failure modes which obtain when the resources allocated to the control plane are exceeded. 

The current plan is simply launching a single control-plane node cluster, adding a few worker nodes to it, and then running conformance tests.

I'll search for prior art with regards to testing the control-plane by pushing it to its limit, and see if we can do something similar for this kind of topology.

>Do we have tests that exercise the systems which recover the control plane when these limits are exceeded, 

I'm not aware of any such systems. I'll check to see how things are currently handled in standard multi-node clusters.

>and the alerts which should fire when these systems fail to operate as intended 

I'm not aware of specific alerts relating to control plane being overloaded, I'm assuming if such alerts exist, there's no reason they wouldn't apply to the SNO + workers topology

>(and do the alerts fire if the control plane is down?)

Depends on what exactly goes wrong and the state of the cluster - prometheus / alertmanager can arbitrarily be running on the control plane node or one of the other additional worker nodes. As long as they both are able to keep running I don't see a reason why alerts wouldn't fire.

Thanks for your comment, it gave me some things I'll need to think about, not sure all of them will be addressed in this enhancement though

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,453 @@
...
+- Deal with non-cloud, non-"none" platforms such as baremetal, vSphere, etc.
+
+## Proposal
+
+Set the "Infrastructure Topology" value to "SingleReplica" in almost all 
+circumstances where the "Control Plane Topology" is also "SingleReplica".
+
+Make sure the default `IngressController` CR created by the Cluster Ingress
+Operator targets the "master" pool rather than the "worker" pool whenever the
+"Infrastrcture Topology" is set to "SingleReplica". This will ensure that the
```
bparees, line 195:
```suggestion
"Infrastructure Topology" is set to "SingleReplica". This will ensure that the
```

global typo fix

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,453 @@
...
+
+### User Stories
+
+- As an OpenShift cluster administrator, I want to install a single control-plane
+node cluster with additional worker nodes, so that my cluster will be able to
+handle larger computation demands.
+
+- As an OpenShift cluster administrator, I want to add worker nodes to my
+existing single control-plane node cluster, so that it'll be able to meet
+growing computation demands.
```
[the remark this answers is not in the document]
omertuc: I personally don't have answers to these questions, I suggest taking a look at MGMT-8414 and the documents linked in it. 

Some things that I think may make this make more sense -

- If my understanding of the motivations behind the requirements is correct, the most common use case is only a single additional worker being added, and in that case it's not enough to form a HA control plane. Situations with 2 or 3 more workers would be the exception.

- 3 machines behaving as 1 control plane + 2 workers are not equivalent to 3 machines that all act as control plane  + workers (compact cluster) - you pay for the HA of the control plane with, for example, a lot of memory for maintaining 3 k8s api servers rather than just 1. 

- I wouldn't be surprised if users actually *did* want to simply transform their 1 control plane + 2 workers cluster into a simple HA compact cluster if given the option to, but this is not something we support and I believe is a harder technical challenge to tackle - users would like to avoid re-installing their existing 1 + 2 cluster into a compact cluster - so for now, 1 + 2 is a good simpler middle-ground to answer user needs before we begin supporting live control plane expansion.

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,453 @@
...
+
+Set the "Infrastructure Topology" value to "SingleReplica" in almost all 
+circumstances where the "Control Plane Topology" is also "SingleReplica".
+
+Make sure the default `IngressController` CR created by the Cluster Ingress
+Operator targets the "master" pool rather than the "worker" pool whenever the
+"Infrastructure Topology" is set to "SingleReplica". This will ensure that the
+`router-default` deployment created by the Cluster Ingress Operator will always
+run on the single control plane node, and as a result any
+`*.apps.<cluster>.<base>` DNS entries which originally pointed at the single
```
tsorya, line 198: How worker will have those dns values? or we don't need them for workers?

enhancements/single-node/single-node-openshift-day2-workers.md
```diff
@@ -0,0 +1,453 @@
...
+circumstances where the "Control Plane Topology" is also "SingleReplica".
+
+Make sure the default `IngressController` CR created by the Cluster Ingress
+Operator targets the "master" pool rather than the "worker" pool whenever the
+"Infrastrcture Topology" is set to "SingleReplica". This will ensure that the
+`router-default` deployment created by the Cluster Ingress Operator will always
+run on the single control plane node, and as a result any
+`*.apps.<cluster>.<base>` DNS entries which originally pointed at the single
+control plane node will remain correct even in the face of newly added worker
+nodes.
```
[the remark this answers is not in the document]
romfreiman: It's a good point, but Im not sure whether it's different from multi node, right? We dont have a way to block in in multi node scenario, mostly it's a documentation. I would assume we'll have to do it here as well. Also, I dont think we should recomend to go ever handfull number of workers. If people want scale, they should go with HA deployment. I would suggest to add it to the docs.

romfreiman:
@omertuc I think lets add non-goal to deal with the solution scalability. The goal of this type of deployment is not to stress the cluster, but more of provide some extra capacity for the workload. wdyt?

enhancements/single-node/single-node-openshift-day2-workers.md
omertuc: We do need them for workers. The user is supposed to make sure DNS is configured "globally" and pointing to the control-plane node before trying to add workers. 
In the case of Assisted, we'll do our best to do it automatically whenever possible with dnsmasq tricks.

omertuc:
Linter failures due to
https://github.com/openshift/enhancements/pull/1011

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,482 @@
+---
+title: single-node-openshift-with-workers
+authors:
+  - "@omertuc"
+reviewers:
+  - "@romfreiman"
+  - "@eranco74"
+  - "@tsorya"
+  - TBD
```
bparees, line 9: you can please make it clear from this section what teams are going to be impacted, either by having to implement something for this now, or support the implementation in the future?  the teams should be tagged for review with an indication of what aspect of it they will be involved in.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,482 @@
...
+the future it may be worth revisiting making this value configurable.
+
+The "None platform SNO day1 single/multiple worker" scenario rows aren't really
+possible today as there is no real way to install this kind of cluster. That's
+why their CIT / CTP have been left empty.
+
+It's proposed that the only scenarios in which the "Infrastrcture Topology"
+parameter will be "HighlyAvailable" are when installing a single control-plane
+node cluster in the cloud with one or more day-1 workers. Cloud deployments
+have a load balancer so they can benefit from having highly-available ingress.
```
dhellmann, line 285: We don't currently offer formal support for single-node deployments in the cloud. Is that expected to change as part of this work?

If the reason for using the HA setting is the presence of a load balancer, would we want to make it possible to configure the infrastructure topology for HA on metal (or platform none) deployments if there is also a load balancer there?

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,482 @@
...
+attempting to support the expansion of a single-node control-plane into a
+three-node control-plane, but I think they would not be trivial.
+
+- Adjust the "baremetal" platform to support single-node installations and make
+users and the Assisted-Installer use that platform instead of the "none"
+platform for single control-plane node cluster installations. The baremetal
+platform solves the issue described in this enhancement with virtual IP
+addresses/keepalived. This approach was dismissed due to much higher
+development efforts and additional processes that would need to run on the
+already resource constrained single control-plane node. Furthermore, even once
```
dhellmann, line 462: Which additional processes would be running? What other implementation work is needed?

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,482 @@
...
+to reduce their costs, and one of the conclusions is that getting rid of the load
+balancer installed by default by the IPI installer results in major cost savings.
+
+A user installing Single-Node OpenShift on "none"-platform will be tempted to
+simply point their DNS entries directly at the IP address of the single node
+that they just installed.
+
+Similarly, in the Assisted Installer, the user is able to complete the
+installation without needing to define any DNS entries. This is currently made
+possible by injecting a `MachineConfig` manifest targeting the "master" pool.
```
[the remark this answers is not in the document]
romfreiman: only SNO

enhancements/single-node/single-node-openshift-with-workers.md
romfreiman: > We don't currently offer formal support for single-node deployments in the cloud. Is that expected to change as part of this work?
nope. Cloud support is out of scope. But we did want to address the flow regardless.

Regrading HA for infrastructure topology - I personally think that it should always be SingleReplica. I think if customers need any kind of HA, they should move to 3nodes + workers cluster. Also, I dont think we should differentiate between on prem and cloud envs unless it's a real must

enhancements/single-node/single-node-openshift-with-workers.md
romfreiman: I assume @omertuc meant keepalived, haproxy, .... all the on prem networking.

enhancements/single-node/single-node-openshift-with-workers.md
dhellmann: > Regrading HA for infrastructure topology - I personally think that it should always be SingleReplica. I think if customers need any kind of HA, they should move to 3nodes + workers cluster. Also, I dont think we should differentiate between on prem and cloud envs unless it's a real must

Being consistent is going to make it easier to reason about the system, so I support that as well, especially if we're still considering formal cloud support out of scope.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,482 @@
...
+schedulable, and as a result a lot of infrastructure workloads fail to schedule
+and operators degrade. User intervention is required to fix this.
+
+- Deal with expansion of the single-node control-plane by adding more control-plane nodes
+
+- Deal with expansion of clusters that have been installed before the
+implementation of this enhancement (if possible, their expansion may be
+addressed with documentation outlining the steps that the user has to take to
+enable expansion, e.g. patch the default `IngressController` to have node
+selector targetting the master pool rather than the worker pool).
```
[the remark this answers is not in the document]
omertuc: It's just that in this case an upgrade will not give the users the new feature - the ingress operator does not modify existing IngressController CRs (it only creates one if one doesn't already exist) - so the router-default deployment will still be in the worker pool

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: I addressed this point in the third bullet in the "Alternatives section".

Some caveats -

- Setting SingleReplica in those scenarios would mean Control plane topology == infrastructure topology under all circumstances, these two parameters would always be identical unless we one day allow users to choose / change them.

- Making this consistent across the types of single node deployments means we're losing consistency with regular HA OCP deployments

- Cloud deployments would lose the HA ingress benefits they can get from the load balancer

Anyway I don't think the decision here is too critical, I don't mind either way. I'll update the enhancement to go with SingleReplica everywhere if you confirm you're okay with the above caveats

enhancements/single-node/single-node-openshift-with-workers.md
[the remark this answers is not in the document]
omertuc: Is your suggestion to make the ingress operator "smarter" and have it realize an upgrade happened and the node selector on the ingress controller needs updating, or simply to ignore it completely and tell users to just reinstall the new version if they want additional workers?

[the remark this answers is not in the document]
dhellmann: It's OK if there is manual work to enable the feature after an upgrade, but we should not claim support for the feature in older releases.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,482 @@
...
+generated during installation", but these are not currently generated in such
+bootstrap-in-place installation.
+
+### Goals
+
+- Solve some of the current known problems related to adding few workers to a
+single control-plane node cluster.
+
+- Define a sensible value for the "Infrastructure Topology" parameter in the
+various configurations a single control-plane node cluster may be deployed in.
```
dhellmann, line 153: Based on something @derekwaynecarr said on slack, I think Hypershift uses this same API. Maybe @enxebre can confirm that and help us ensure that the changes here won't affect Hypershift deployments.

enhancements/single-node/single-node-openshift-with-workers.md
dhellmann: The control plane topology can also be "external", in the Hypershift case. So the infrastructure and control plane fields wouldn't always be the same.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,482 @@
...
+|None platform SNO day1 single worker    |None    |1  |1  |-  |SR |-  |M  |Yes, ingress floats                              |No, ingress pinned to master pool             |
+|Cloud platform SNO day1 single worker   |Cloud   |1  |1  |SR |HA |W  |W  |Yes - provided, ingress floats                   |Yes - provided, there are two ingress replicas|
+|None platform SNO day1 multiple workers |None    |2+ |2+ |-  |SR |-  |M  |Yes, ingress is on two different nodes           |No, one ingress and is pinned to master pool  |
+|Cloud platform SNO day1 multiple workers|Cloud   |2+ |2+ |HA |HA |W  |W  |Yes - provided, ingress is on two different nodes|Yes - provided, there are two ingress replicas|
+|None platform SNO only day2 workers     |None    |0  |1+ |SR |SR |W  |M  |Yes, ingress floats                              |No, ingress pinned to master pool             |
+|Cloud platform SNO only day2 workers    |Cloud   |0  |1+ |SR |SR |W  |M  |Yes - provided, ingress floats                   |Yes - provided, ingress floats                |
```
bparees, line 265: Some questions on this table.....

1) can we denote which of these scenarios are already supported today?  And which "work" today?
2) the first two rows show "zero" workers, but presumably the single node serving as the master is a worker, right?  So it's zero additional workers?  I'm trying to understand how it differentiates from the 3rd and 4th rows which show "one" worker  (Is the master also treated as a worker, when you have a dedicated worker?)
3) It feels like we may be trying to shoehorn the use of the InfrastructureTopology field here.  The fact that we'd have scenarios where we have multiple workers capable of scheduling multiple replicas of infrastructure components, but we will not because of loadbalancer/ingress concerns, makes me think we either need a new field, or a new value for the existing field (e.g. "HighlyAvailableInfraWithSingleNodeIngress" or "HAWithMasterIngress").  Otherwise users looking at the field are likely to be confused by the value it shows.
4) what does the `-` mean in the CIT and CTP columns?  does that mean it's not a scenario that's possible today?
5) in discussing this with @dhellmann we were wondering if maybe even the Cloud scenarios should always be single ingress (targets master).  Yes in theory getting a load balancer is easy, but is it necessary or beneficial, vs the benefits of keeping the SNO experience consistent?  Especially since using SNO is in part a cost decision, and adding the LB in increases cost again.

romfreiman: >in discussing this with @dhellmann we were wondering if maybe even the Cloud scenarios should always be single ingress (targets master). Yes in theory getting a load balancer is easy, but is it necessary or beneficial, vs the benefits of keeping the SNO experience consistent? Especially since using SNO is in part a cost decision, and adding the LB in increases cost again.

Defintely +1 from me. I think it should be 1 everywhere. IMO no real need to provide HA solution in such scenarios - if users need ha, they can use ha control plane as a starting point

enhancements/single-node/single-node-openshift-with-workers.md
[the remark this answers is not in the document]
omertuc: Oh I see what you mean now - I didn't realize my enhancement text implies that I'm suggesting the users should make those changes without upgrading. I'll rephrase it to clarify

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: > * can we denote which of these scenarios are already supported today?  And which "work" today?
> * what does the `-` mean in the CIT and CTP columns?  does that mean it's not a scenario that's possible today?

See [this paragraph](https://github.com/openshift/enhancements/blob/278909ed0e513d86445c3320cb928c3a7da819a5/enhancements/single-node/single-node-openshift-with-workers.md?plain=1#L279-L281) about unsupported scenarios. I've also updated the PR with a new column indicating current works/supported/tested for each scenario to clarify this.

> * the first two rows show "zero" workers, but presumably the single node serving as the master is a worker, right?  So it's zero additional workers?  I'm trying to understand how it differentiates from the 3rd and 4th rows which show "one" worker  (Is the master also treated as a worker, when you have a dedicated worker?)

The word "worker" throughout the document refers to nodes which are strictly just workers. So to answer your question - rows 1-2 are just a single node throughout the entire life of the cluster. Rows 3-4 are clusters which get installed with exactly 2 nodes (1 control-plane node and another dedicated worker). Rows 5-6 are clusters which get immediately installed with 3 or more nodes (1 control-plane node and 2 or more dedicated workers). Rows 7-8 are clusters which get installed with just 1 control plane node but then have one or more dedicated worker nodes added to them post-installation.

> * It feels like we may be trying to shoehorn the use of the InfrastructureTopology field here.  The fact that we'd have scenarios where we have multiple workers capable of scheduling multiple replicas of infrastructure components, but we will not because of loadbalancer/ingress concerns, makes me think we either need a new field, or a new value for the existing field (e.g. "HighlyAvailableInfraWithSingleNodeIngress" or "HAWithMasterIngress").  Otherwise users looking at the field are likely to be confused by the value it shows.

Yes I agree with you. I've kinda addressed this [here](https://github.com/openshift/enhancements/blob/9a485478aa068f55d7383393dda0abbecf577800/enhancements/single-node/single-node-openshift-with-workers.md?plain=1#L296-L304). I didn't consider adding a new field / new enum values as a possible solution. It definitely sounds reasonable. I'll think about it

> * in discussing this with @dhellmann we were wondering if maybe even the Cloud scenarios should always be single ingress (targets master).  Yes in theory getting a load balancer is easy, but is it necessary or beneficial, vs the benefits of keeping the SNO experience consistent?  Especially since using SNO is in part a cost decision, and adding the LB in increases cost again.

I want to point out a potential misunderstanding - "getting a load-balancer" is not entirely accurate, a load-balancer is already automatically provided by the IPI installer today - everything there already "just works". Today if you set your install-config.yaml to 1 control plane and n>0 workers on AWS and run the installer, you get a working SNO + workers cluster that behaves correctly even when the ingress floats and you have highly available ingress. Changing the infra topology to SingleReplica in this enhancement will be actively making such deployments worse than they are today for the sake of consistency. Changing the infra topology parameter to SingleReplica would make more sense the day we remove the load balancer from IPI, but with a load-balancer already provided today, we'd just be making things worse. 

Note that I'm not saying that consistency doesn't have a value / is the incorrect decision here, I just want to make sure the situation is well understood before we decide consistency is the way to go. See also [this](https://github.com/openshift/enhancements/pull/996#discussion_r787127690) comment

bparees: > Today if you set your install-config.yaml to 1 control plane and n>0 workers on AWS and run the installer, you get a working SNO + workers cluster that behaves correctly even when the ingress floats and you have highly available ingress.

but technically this is not an install topology/platform we support today, right?

> Note that I'm not saying that consistency doesn't have a value / is the incorrect decision here

consistency is one piece, cost savings is another piece (keeping SNO's footprint as minimal as possible)

i'm not necessarily trying to push hard for this approach, but i think we should at least make sure we've had the conversation w/ PM about whether it makes sense, for the cloud customers we have in mind for this topology if/when we decide to declare it to be supported

omertuc: >but technically this is not an install topology/platform we support today, right?

That's correct. We do not support it. But it works.

>consistency is one piece, cost savings is another piece (keeping SNO's footprint as minimal as possible)

The point I'm trying to get across is that consistency *is* the only piece

This is because cost saving is out of scope for this enhancement - the load balancer will exist regardless of whether we set the infra topology to HA or not, no money is saved (if I understand correctly, I might have a knowledge gap here then feel free to correct me).

This enhancement does not get rid of the load balancer, so might well keep the infra topology HA.

If future enhancements want to do away with the load balancer in single control-plane node clusters in the cloud for whatever reason - those enhancement should re-define the value of the infra topology field to be SingleReplica.

bparees: > This is because cost saving is out of scope for this enhancement - the load balancer will exist regardless of whether we set the infra topology to HA or not, no money is saved

thanks, I did not realize it was being created regardless in this case.

At a minimum i think we need to ensure that answering the question of what LB behavior we want, is defined before we declare support for SNO on cloud.  That's outside the scope of this EP, but probably needs to be called out in this EP.  Removing the cloud entries from the table (as i think was suggested elsewhere) might help to reduce the confusion.  Or at least making it clear that the behavior defined for cloud entries is being described as "how it works today, but it's not supported and we would consider changing how it works before making it a supported topology"

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,483 @@
...
+
+- `SNO` = Single control-plane node cluster
+
+|Scenario                                |Platform|D1W|D2W|CIT|PIT|CTP|PTP|Is a load balancer currently needed?             |Will a load balancer be needed after proposal?|Current Support                     |
+|----------------------------------------|--------|---|---|---|---|---|---|-------------------------------------------------|----------------------------------------------|------------------------------------|
+|None platform SNO no workers            |None    |0  |0  |SR |SR |W  |M  |No                                               |No                                            |Works, supported, tested            |
+|Cloud platform SNO no workers           |Cloud   |0  |0  |SR |SR |W  |M  |No                                               |No                                            |Works, unsupported, tested          |
+|None platform SNO day1 single worker    |None    |1  |1  |-  |SR |-  |M  |Yes, ingress floats                              |No, ingress pinned to master pool             |Impossible to install               |
```
staebler, line 261: I am missing what the limitation is that makes a SNO cluster with day-1 workers an impossibility for the "none" platform. I see mention above about the worker ignition file not being generated when doing a bootstrap-in-place installation. Is that what the "impossibility" means here? If so, then that assume thats (1) the SNO installation is a bootstrap-in-place installation and (2) that the user does not know how to run `openshift-install create ignition-configs` to generate the worker ignition.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,483 @@
...
+means an optimal value for this parameter must be determined during
+installation. You can see the consequences of that, for example, in the last
+row of the table - even though we add workers in day 2 to the cluster, the
+"Infrastrcture Topology" stays "SingleReplica" even though that cluster could
+benefit from the additional worker nodes for the purposes of highly-available
+infrastructure (as it's running in the cloud, so it has a load-balancer). In
+the future it may be worth revisiting making this value configurable.
+
+The "None platform SNO day1 single/multiple worker" scenario rows aren't really
+possible today as there is no real way to install this kind of cluster. That's
```
staebler, line 280: I am missing what the limitation is that makes a SNO cluster with day-1 workers an impossibility for the "none" platform. I see mention above about the worker ignition file not being generated when doing a bootstrap-in-place installation. Is that what the "impossibility" referes to here? If so, then that assume that (1) the SNO installation is a bootstrap-in-place installation and (2) that the user does not know how to run `openshift-install create ignition-configs` to generate the worker ignition.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,483 @@
...
+
+It's proposed that the only scenarios in which the "Infrastrcture Topology"
+parameter will be "HighlyAvailable" are when installing a single control-plane
+node cluster in the cloud with one or more day-1 workers. Cloud deployments
+have a load balancer so they can benefit from having highly-available ingress.
+
+Today, if you only have 1 day-1 worker in a single control-plane node cluster
+in the cloud, then the "Infrastrcture Topology" is set to "SingleReplica" - I
+don't believe there's a good reason for that, as both the control-plane node
+and the worker node are actually in the worker pool (TODO: make sure this is
```
staebler, line 291: Why is the control-plane node in the worker pool in this scenario? The installer would not mark the masters as schedulable. Is there something happening in-cluster to make the masters schedulable?

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,483 @@
...
+documentation points the user to use "the worker ignition file that get
+generated during installation", but these are not currently generated in such
+bootstrap-in-place installation.
+
+### Goals
+
+- Solve some of the current known problems related to adding few workers to a
+single control-plane node cluster.
+
+- Define a sensible value for the "Infrastructure Topology" parameter in the
```
staebler, line 152: How is this a goal of this enhancement? To me, this sounds like a by-product of the solution chosen to tackle the subsequent goal.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,483 @@
...
+various configurations a single control-plane node cluster may be deployed in.
+
+- Deal with the "floating ingress" issue encountered in "none"-platform
+single control-plane node clusters which have worker nodes added to them.
+
+- Define how the installer may be modified to generate the worker ignition
```
staebler, line 158: I don't see where a design to satisfy this goal is addresses at all in this enhancement.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,483 @@
...
+control-plane node cluster today using the OpenShift Installer does not provide
+the user with a worker node ignition file. As a result, the user is not able to
+manually add worker nodes to the resulting cluster. All current official OCP
+documentation points the user to use "the worker ignition file that get
+generated during installation", but these are not currently generated in such
+bootstrap-in-place installation.
+
+### Goals
+
+- Solve some of the current known problems related to adding few workers to a
```
staebler, line 149: This goal is ambiguous. The current known problems should be enumerated as goals.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,483 @@
...
+
+When installing day-1 workers on a single control-plane node cluster on the
+"none"-platform, it's unlikely you have a load balancer / virtual IP set up, so
+even if you have multiple workers to hold ingress pods, there would be no load
+balancer to balance between them. So it makes sense to only have one ingress
+replica. That's why we set "Infrastrcture Topology" to "SingleReplica" even in
+the "none"-platform even though we have day-1 workers. This has the unfortunate
+consequence that non-ingress infrastructure workloads (such as monitoring, for
+example) can no longer benefit from spreading over the worker nodes - since
+they will all also have just 1 replica (as a result of the topology). It's
```
staebler, line 304: This consequence is a smell. It is what tells me that forcing the infrastructure topology to single replica for the purposes of ensuring that the router pods run on the masters is not a desirable solution. It seems to me that one and only goal of this entire proposal is getting the router pods pinned to run on the master in SNO. The discussion about topologies is just distracting from that goal. Since the real goal is pinning the router pods, and the router pods are strictly under the purview of the ingress operator, the configuration directing the ingress operator to place the router pods on the master should be in the ingress.config.openshift.io resource. The configuration of the router placement then has no side-effects for the rest of the cluster.

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: Will respond on the duplicate comment

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: Yeah at first the table specified the topology that such clusters would have according to the current installer logic for the number of workers, but I was convinced that this is not even possible, so I left them blank. 

I wasn't aware the installer is able to:

- Do single node "none" platform without bootstrap-in-place (using a bootstrap node). It's not something I ever tried myself / it's not something we're testing periodically so I guess I kinda just assumed it might not work.

- Didn't realize generating the worker configs in bootstrap-in-place is a simple as running this command... but when we're doing bootstrap-in-place how/when would that worker join? That that work in any way? Also, I guess the definition of "day-1" workers becomes a bit blurry. Does day-1 workers simply mean the number of workers specified in the installer-config manifest? I guess that definition could make sense. Would the installer even accept doing bip with workers being a non-0 value today?

The first bullet is convincing enough for me to fix the table to show the current topology rather than `-` though, so I'll do that.

In any case, the important bit in this part of the enhancement was to explicitly discuss / specify what the values *should* be today, rather than what it is now.

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: Yeah it was just some assumption I based this paragraph on that I later realized might likely be completely wrong, this is why the TODO is there - I wanted to try it in practice, haven't got around to it yet.

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: You are correct, it's specified as TBD [here](https://github.com/openshift/enhancements/blob/9a485478aa068f55d7383393dda0abbecf577800/enhancements/single-node/single-node-openshift-with-workers.md?plain=1#L335) - I wanted to get feedback on other more crucial parts of this enhancement before digging into this part, to make sure I'm on the right track.

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: The goal of this enhancement was, among other things like solving the floating ingress issue, was truly to make a formal definition / discuss the value of the infra topology field in all possible SNO/Workers/Day-1/2 scenarios because to the best of my knowledge, since these kind of scenarios weren't really supported, they weren't explicitly discussed / defined too.

You are however 100% right that the new values chosen for the infra topology were to tackle the subsequent goal, and it's completely fair that you do not agree with that approach. If I were to come up with a different approach, I would still like this enhancement to explicitly specify the infra topology value, even if just to say "the value stays the same as it happens to be today".

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: For the record Ben shared similar concerns to which I responded [here](https://github.com/openshift/enhancements/pull/996#discussion_r787631818) (third quote) 

The thought process behind changing the topology was -

- "We need to pin the router pod to master in "none" platform because SNO users typically don't want / need a load balancer / VIP solution by default even when there are workers"

- "It's insufficient to simply pin the router to master because if there are multiple workers in day 1 the installer would set infra topology to HA and then we would have multiple replicas of the router pod which doesn't make sense if they're all pinned to a single node"

- "We need to avoid having multiple router replicas. The only solution: Set infra topology to SingleReplica to avoid multiple routers"

I didn't realize adding new APIs / configuration options would be desirable, so I didn't really consider it as an option at all, so I went with the suboptimal (and only solution I can think of) of "abusing" the infrastructure topology parameter. 

But given your & Ben's response to the proposed solution I guess adding new configuration options is a reasonable thing to do.

_______

I guess what we would want then would be a new way to tell the Ingress Operator (ideally through the IngressController API) to ignore the infrastructure topology altogether when determining the number of replicas of the router deployment and always set the "replicas" parameter to 1 and pin that router pod to master.

This way the installer logic where the number of day 1 workers determines the infra topology stays the same, infrastructure workloads can still spread over all nodes in the worker pool, and ingress would be just 1 replica pinned to master.

This new IngressController configuration, for now, will be applied by the Installer automatically in single control-plane node "none"-platform clusters. In the future if one day the load balancer is removed from cloud IPI deployments of single control-plane node clusters then this configuration could be applied on those platforms as well.

Does this approach make sense?

omertuc: Come to think of it it doesn't necessarily have to be a new configuration option... Maybe simply setting the IngressContrller replicas to 1 and pinning it to master by the installer is all we have to do

staebler: > I guess what we would want then would be a new way to tell the Ingress Operator (ideally through the IngressController API) to ignore the infrastructure topology altogether when determining the number of replicas of the router deployment and always set the "replicas" parameter to 1 and pin that router pod to master.

The Ingress Operator would not always set the replicas to 1. The Ingress Operator would consider the control plane topology rather than the infrastructure topology when configured to place its pods on the masters.

staebler: > Maybe simply having the installer set the IngressContrller replicas to 1 and pin it to master is all we have to do?

Directing the ingress operator to "pin it to master" is what requires an api change, unless I am missing something.

omertuc: >The Ingress Operator would not always set the replicas to 1. The Ingress Operator would consider the control plane topology rather than the infrastructure topology when configured to place its pods on the masters.

Yeah that makes more sense

>Directing the ingress operator to "pin it to master" is what requires an api change, unless I am missing something.

We could simply do something like this:
image


(don't mind the bip case condition)

staebler: > We could simply do something like this

That is the action that the ingress operator would take to pin the routes to the masters. But that is not how the ingress operator would know to take that action. That last bit is what would require a new configuration item.

romfreiman: @Miciah

romfreiman: I do wonder though whether we dont want to have all the infra components pinned to the master - for predictability, ease of debug and keeping the workers for users workloads

omertuc: >That is the action that the ingress operator would take to pin the routers to the masters. But that is not how the ingress operator would know to take that action. That last bit is what would require a new configuration item.

What I'm saying is that the ingress operator doesn't really have to know to take that action, we can simply create this IngressController resource in the installer (the screenshot above is installer code)

staebler: Ah, sorry, I missed that that was the installer code for generating the default ingress controller. The installer generates a manifest for a default ingress controller only if the cluster is using an internal publishing strategy. The publishing strategy can only be set to internal when installing to a cloud platform. Perhaps with this enhancement, the installer would allow an internal publishing strategy for non-cloud platforms. Nevertheless, I still think that it is better for the ingress operator itself to determine the node placement based on configuration that the ingress operator knows. This would allows us to keep consistent behavior regardless of the publishing strategy.

omertuc: I'm not sure I understand how the publishing strategy is relevant: I know the installer currently only generates an IngressController manifest if the publishing strategy is internal, but how does that relate to the discussion? What stops us from also generating an IngressController manifest when the platform is "none" (regardless of publishing strategy)?

staebler: There is nothing that stops the installer from doing that. Is that what we want? Does the network edge team have an opinion on whether they would prefer the installer an IngressController or whether they would prefer the ingress operator create the IngressController? My preference is to have the ingress operator do it since it keeps the domain knowledge located in the operator instead of the installer.

> generating an IngressController manifest when the platform is "none" and the control plane topology is "SingleReplica"

Would this be when the platform is an on-prem platform rather than just a "none" platform?

omertuc: > There is nothing that stops the installer from doing that. Is that what we want? Does the network edge team have an opinion on whether they would prefer the installer an IngressController or whether they would prefer the ingress operator create the IngressController? My preference is to have the ingress operator do it since it keeps the domain knowledge located in the operator instead of the installer.
> 
> > generating an IngressController manifest when the platform is "none" and the control plane topology is "SingleReplica"
> 
> Would this be when the platform is an on-prem platform rather than just a "none" platform?

Okay I've gave it some thought, do you personally see any problem with the following approach then:

In the absence of IngressController node-selector and replicas, if the user is on an on-prem platform (none, vsphere, etc*) IngressOperator will "put ingress on the contrl-plane" meaning:
- It will set number of replicas to equal controlplanetopology (1 if singlereplicas 2 if ha)
- It will set nodeSelector to master rather than worker

No installer changes, no new ingresscontroller API (other than changing documentation in the existing API to reflect that behavior)

* I guess IngressController would also have to maintain a list of platforms it considers "on-prem" to do that

Miciah: > Does the network edge team have an opinion on whether they would prefer the installer an IngressController or whether they would prefer the ingress operator create the IngressController?

My opinion is that the proposed defaulting semantics for the IngressController API are somewhat complex (based on the table under "Infrastructure Topology value"), and this could be rather difficult to explain in the API godoc and difficult for the cluster administrator to understand.  

It seems to me that instead of using some non-obvious inference rules to define a function to map a cluster infrastructure config to a set of nodes on which router pods should run, it would be more obvious and more in the Kubernetes way to label the nodes at install-time (using `node-role.kubernetes.io/infra`, or some novel label if the definitions of "infra node" and "router node" should be independent) and allow the user to adjust labels as desired after installation.  Would using labels in this way be out of the question?

Miciah: > *I guess IngressController would have to maintain a list of platforms it considers "on-prem" to do that

We kind of do that already to decide whether to default the endpoint publishing strategy type to `HostNetwork` or to `LoadBalancerService`.  

To my point in my previous comment, the current defaulting logic is straightforward enough (platform type → endpoint publishing strategy type, infrastructure topology → number of replicas).  The propose logic will be more complex ((platform type, control-plane topology, infrastructure topology) → node selector).  Complex defaulting logic can lead to surprises.

omertuc: >My opinion is that the proposed defaulting semantics for the IngressController API are somewhat complex (based on the table under "Infrastructure Topology value"), and this could be rather difficult to explain in the API godoc and difficult for the cluster administrator to understand.

A lot of that complexity you're describing is coming from me trying to abuse the Infrastructure Topology value to achieve my goals, @staebler's proposal is to admit that infra topology has nothing to do with it - and just tell Ingress Operator to completely disregard infrastructure topology when the platform is not a cloud platform. Instead it should look at the platform and control-plane topology, because when the control-plane topology is "SingleReplica" and the platform is on-prem, the ingress is practically bound to the control-plane due to the likely absence of a load-balancer/VIP, so it shouldn't care about the infra topology.

The infra topology is still important for other things like monitoring, logging, etc - they can still regard it, their behavior is not bound to an existence of a load-balancer

omertuc: Edited my previous [comment](https://github.com/openshift/enhancements/pull/996#discussion_r799388937) since it was a little wrong

omertuc: To summarize the ingress logic, it would look something like this:
- If on-prem platform and single-replica control plane topology, set replicas to 1 and pin to master
- If cloud platform, set replicas according to infrastructure topology and pin to workers

Does that still sound too complicated?

Miciah: As an exercise, I tried to write out my understanding of the proposal for `spec.replicas` as I would expect it to be documented in the API:

> If unset, the operator uses a default value that depends on the platform type, control-plane topology, and infrastructure topology.
> 
> For cloud platforms, the default value depends on the infrastructure topology on the assumption that the cluster has dynamically configured DNS and a cloud load-balancer that can dynamically add worker nodes to its pool.  For on-premise platform, the default value depends on the control-plane topology on the assumption that a single-node cluster may have static DNS and load-balancer configuration (or no load balancer at all) with the ingress controller pinned to the control-plane node, and so adding additional workers should not change the number of ingress controller replicas.
> 
> The following platform types are considered to be cloud platforms: Alibaba, AWS, Azure, GCP, IBMCloud, and PowerVS.  All other platform types (unspecified, BareMetal, OpenStack, etc.) are considered to be on-premise platforms.
> 
> If the platform type is a cloud platform and the infrastructure topology is "SingleNode", then the default number of replicas is 1.  If the platform type is on-premise and the control-plane topology is "SingleNode", then the default number of replicas is 1. Otherwise, the default number of replicas is 2.

We would want similar godoc for the `spec.nodeSelector` field.  

Does my example capture the full complexity?  Do you anticipate any future changes that could make the defaulting logic still more complex?  

I'm still a little reluctant to put this much complexity behind API field defaulting, but if it passes API review, we can include it in the operator.

staebler: I do not think that the ingress operator should be inferring from platform type where to put the router pods. My suggestion is that we add a field to the configuration of the ingress operator to direct the pods to either the workers or the masters.
The logic would be the following.
* If the ingress operator is configured for routing on masters, the ingress operator will use a node selector for the pods to place them on the masters and will set the replica count based on the ControlPlaneTopology.
* If the ingress operator is configured for routing on workers, the ingress operator will use a node selector for the pods to place them on the workers and will set the replica count based on the InfrastructureTopology. My assumption there is that is the current behavior. Please correct my naive understanding of what the ingress operator does under the covers.
* For capability with upgrades, if there is no configuration set, then the infgess operator will default to workers for the configuration.

omertuc: > add a field to the configuration of the ingress operator to direct the pods to either the workers or the masters.

I see. I guess I just find that idea a bit strange because, in a way, the IngressController already has an API to do this - it's called the nodeSelector.

You suggest we would introduce another, more explicit "master vs workers" field, that will control the defaulting behavior of both the replicas and the nodeSelector. Sounds reasonable.

Here's another idea I want to float, it's a bit hacky, let me know what you think -
Instead of introducing a new  API parameter, the ingress operator would adjust its defaulting `replicas` behavior depending on whether the IngressController nodeSelector has particular labels or not (master/worker labels).

It's a bit ugly because as a user I would find it surprising that my choice of node selector labels is somehow able to affect other fields of the IngressController, but on the other hand it would get rid of a bit of redundancy in the IngressController API

staebler: > I see. I guess I just find that idea a bit strange because, in a way, the IngressController already has an API to do this - it's called the nodeSelector.

Yes, the IngressController has an API to do this. And I would expect the ingress operator to use that. I prefer that the ingress operator create the IngressController rather than having the installer create it. As far as I know, ingress is the only operator for which the installer is creating controllers. I don't know the history of that, but unless there is strong desire by the ingress team to double down on that, then I would rather not add to the criteria in which the installer needs to create an IngressController.

staebler: > Instead of introducing this new boolean parameter to IngressController

I am not suggesting adding a field to IngressController. I am suggesting adding a field to the [Ingress CRD](https://github.com/openshift/api/blob/master/config/v1/types_ingress.go#L16).

omertuc: Oh, didn't realize that you're talking about the Ingress CRD. sgtm.

omertuc: >Add a high-level "ingress topology" field analogous to the control-plane topology and infrastructure topology fields; the ingress operator could use this high-level description of intent in its defaulting logic.
Add both a nodePlacement field and a replicas field to the ingress config; the ingress operator could directly use these values, if provided, as default values for the corresponding fields in ingresscontrollers.

I'm leaning towards the first option, since the second one would put the burden of understanding the complexities involved on the installer, because it would be the one who would have to set the correct values on that ingress config CR.

With the first option all the installer has to do is set the "IngressTopology" to "ControlPlane"

>One question: Would these values only be used when generating a default ingresscontroller, or would they be used when defaulting the spec.replicas and spec.nodePlacement fields on any ingresscontroller that doesn't specify these fields?

I think the answer that makes most sense is "both", is there any concerns that you have with that? I can't come up with any personally

omertuc: >Add a high-level "ingress topology" field analogous to the control-plane topology and infrastructure topology fields; the ingress operator could use this high-level description of intent in its defaulting logic.

Ran this by @dhellmann , he proposed to have this topology field under the Ingress CRD rather than be a "high level" API as you mentioned, since it's component-specific.

I think that makes sense, doing that would be kind of a hybrid between your two proposed solutions - we make it abstract (a boolean topology) - but we put it in the Ingress config CR rather than in a high level cluster configuration.

The installer can then not care about the exact defaulting details (exact label values and number of replicas)

Miciah: > I think the answer that makes most sense is "both", are there any concerns that you have with that? I can't come up with any personally

I have no concerns with this answer; this approach makes sense.  

> Ran this by @dhellmann , he proposed to have this topology field under the Ingress CRD rather than have it be a "high level" API as you mentioned, since it's component-specific.
> 
> I think that makes sense, doing that would be kind of a hybrid between your two proposed solutions - we make it abstract (a boolean topology) - but we put it in the Ingress config CR rather than in a high level cluster configuration.

This is actually what I meant.  Sorry that I wasn't clearer.  We are in complete agreement.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,496 @@
...
+reviewers:
+  - "@romfreiman"
+  - "@eranco74"
+  - "@tsorya"
+  - "@dhellmann"
+  - "@Miciah"
+  - "@bparees"
+  - "@JoelSpeed"
+  - "@staebler"
+  - "@derekwaynecarr"
```
romfreiman, line 14: @sinnykumari PTAL

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,496 @@
...
+approvers:
+  - TBD
```
romfreiman, line 16: @dhellmann @sdodson who should we add as approvers?

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,496 @@
...
+api-approvers: # in case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers)
```
romfreiman, line 17: @Miciah who's the approver from network-edge side?

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,496 @@
...
+This enhancemnet aims to enable adding workers to a single control-plane node
+cluster by attempting to tackle multiple issues that arise when worker nodes
+are added to such cluster.
+
+## Motivation
+
+Today OCP supports deploying a cluster which only has a single control-plane
+node and no additional workers.
+
+There's been recent demand from OCP to also formally support deployment of
```
dhellmann, line 44:
```suggestion
There's been recent demand for OCP to also formally support deployment of
```

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,496 @@
...
+
+### API Extensions
+
+Introduce a new topology field in the Ingress config CR
+(`config.openshift.io/v1/ingresses`) called `IngressTopology`.
+
+### Implementation Details/Notes/Constraints
+
+Unlike the existing `config.openshift.io/v1/infrastructures` topology fields, this
+new field will have one of these values - `ControlPlane` or `Workers`.
```
dhellmann, line 224: If it's describing the location of the ingress controllers, rather than the "topology", maybe the new field should have a different name.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,496 @@
...
+Introduce a new topology field in the Ingress config CR
+(`config.openshift.io/v1/ingresses`) called `IngressTopology`.
+
+### Implementation Details/Notes/Constraints
+
+Unlike the existing `config.openshift.io/v1/infrastructures` topology fields, this
+new field will have one of these values - `ControlPlane` or `Workers`. 
+
+In addition, allow the `.spec.replicas` and `.spec.nodePlacement` parameters in
+`operator.openshift.io/v1/ingresscontrollers` CRs to be ommitted.
```
dhellmann, line 227: If there are already fields that control how many ingress controller replicas there are and where they run, why do we need the new field at all? The description below only mentions default behaviors. Are those defaults different from what a user would see today?

enhancements/single-node/single-node-openshift-with-workers.md
dhellmann: There's an API change here, so we need an API approver to look at it. @mfojtik can you help identify someone with time to review this?

I would expect to see someone from the networking team familiar with ingress, too. @knobunc can you help find a reviewer?

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: I've raised this concern and @staebler had [this](https://github.com/openshift/enhancements/pull/996#discussion_r799552984) to say about it

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: Fair point, renamed to `IngressPlacement`

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: >The description below only mentions default behaviors. Are those defaults different from what a user would see today?

See quote from enhancement:
>When the value of `IngressPlacement` is `Workers`, the defaulting behavior of
`.spec.replicas` and `.spec.nodePlacement` will be the same as it is today

dhellmann: > > If there are already fields that control how many ingress controller replicas there are and where they run, why do we need the new field at all?
> 
> I've raised this concern and @staebler had [this](https://github.com/openshift/enhancements/pull/996#discussion_r799552984) to say about it

That argument explains why the new field is on the Ingress CRD, but I still don't understand why we need it at all.

> > The description below only mentions default behaviors. Are those defaults different from what a user would see today?
> 
> See quote from enhancement:
> 
> > When the value of `IngressPlacement` is `Workers`, the defaulting behavior of
> > `.spec.replicas` and `.spec.nodePlacement` will be the same as it is today

So the new field allows us to make the other fields optional and can be used to control the defaults of the other fields.

We can we not use the existing fields, and set their values using the same logic that would be used to set the value of the new field?

omertuc: >We can we not use the existing fields, and set their values using the same logic that would be used to set the value of the new field?

Setting just those fields would require the installer to create an entire IngressController CR (typically it's the Ingress operator's job to create that CR when it's missing), and this is what @staebler wants to avoid

dhellmann: Ah, I missed that the fields were in different CRs. That makes much more sense now, thank you.

enhancements/single-node/single-node-openshift-with-workers.md
romfreiman: @tkashem

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,495 @@
...
+```
+
+(Note that the `kubernetes.io/os: linux` label is mentioned just because it's
+the current behavior, it has no importance in this enhancement)
+
+The installer will detect situations in which it's unlikely the user will want
+to set up a load-balancer. Those situations currently include installation of
+single control-plane node cluster deployments on "on-prem" platforms such as
+"none" or "vSphere" (although today single control-plane node clusters are only
+possible on the "none" platform). In those situations, the installer will set
```
Miciah, line 272: The wording here, especially the phrase "currently include", is imprecise.  Does this include *all* on-premise platforms (when configured with a single control-plane node)?  Are there currently any other situations in which the installer would set `IngressPlacement: ControlPlane`?  Are there any other situations that we can foresee in which the installer might reasonably set `IngressPlacement: ControlPlane`?

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,495 @@
...
+
+The value of the `IngressPlacement` field will affect the defaulting behavior
+of the `IngressController`'s `.spec.replicas` and `.spec.nodePlacement`
+parameters. i.e. in the absence of an `IngressController` resource created by
+the user/installer, or when the user/installer creates an `IngressController`
+with these two parameters ommited, the Cluster Ingress Operator will choose the
+default values for those parameters based on the value of `IngressPlacement`.
+
+If the value of `IngressPlacement` itself is ommited, it is defaulted to equal
+to `Workers`.
```
Miciah, line 236: This isn't really true though, is it?  The actual defaulting behavior is more complicated, as described with lines 268 through 273 ("The installer will detect situations in which it's unlikely the user will want to set up a load-balancer. [...] In those situations, the installer will set `IngressPlacement` to be `ControlPlane`.").  Am I misunderstanding?

Miciah (commented):
I want to make sure I understand the defaulting behavior correctly.  Otherwise, I just have a few stylistic suggestions and typo fixes.  The overall approach looks good to me.  
/approve

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,495 @@
...
+the worker pool (the reason the control-plane node was also in the worker pool
+is that when during installation there are no worker nodes, the OpenShift
+installer sets the Scheduler CR `.spec.mastersSchedulable` to `true` and as a
+result the control-plane node is in both the "master" and "worker" pools).
+
+This pod floating between the nodes is a problem because user ingress traffic
+has to be directed at the node currently holding the `router-default` pods, and
+since the DNS entries have been "naively" pointed at the original control-plane
+node's IP address (which may no longer be running those pods), ingress traffic
+can no longer work. This can be temporarily solved if DNS is adjusted to point
```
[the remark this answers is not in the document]
omertuc: right

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -0,0 +1,495 @@
...
+amount of workers/pods is desired the user would have to re-install the cluster
+as a regular multi control-plane node cluster.
+
+- Deal with clusters that have less nodes than they had during day-1
+installation. The rationale for this being a non-goal is that it seems that
+removing the "initial" nodes of a cluster causes a lot of issues even in a
+regular, multi-node installation. So dealing with similar issues in single
+control-plane node clusters is also out of scope for this enhancement. For
+example, installing a cluster with 3 control-plane nodes and 3 workers results
+in unschedulable control-plane nodes. Scaling down the worker machine-sets from
```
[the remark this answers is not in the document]
omertuc: Your understanding is correct. And I understand the confusion from using the blanket term "unschedulable", however I'm just borrowing the terminology from the Scehduler CR entry that controls this behavior:
https://github.com/openshift/api/blob/8e88d13169d91aa9c64c8b6ed8bba0caf4a40d3a/config/v1/0000_10_config-operator_01_scheduler.crd.yaml#L42-L44

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: Added paragraph

```markdown
In the future, when IPI-installed single control-plane node clusters in the
cloud no longer provision a load-balancer by default, they would also benefit
from having the installer set the `IngressPlacement` to `ControlPlane`.
```

enhancements/single-node/single-node-openshift-with-workers.md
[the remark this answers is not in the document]
Miciah: Yeah, the `mastersSchedulable` field name and documentation suffer from using such a fraught term.  It's probably too much trouble to change the field name now, and fixing the godoc or product documentation is way out of scope for this enhancement.  However, in my opinion, going forward, the terms "schedulable" and "unschedulable" should be qualified or avoided entirely because their various historical uses render the semantics murky and misleading.  That's just my opinion though.  ¯\\\_(ツ)\_/¯

enhancements/single-node/single-node-openshift-with-workers.md
Miciah: Fair enough.  There are many layers of defaulting here!

PR #1002, opened 2022-01-10
CFE-10: Add host-network-bind-options

Elbehery:
This PR add a proposal for adding Binding Options for the IngressAPI

### DIFF
--- a/enhancements/ingress/host-network-configurable-ports.md
+++ b/enhancements/ingress/host-network-configurable-ports.md
@@ -1,0 +1,319 @@
+---
+title: ingress-host-network-configurable-ports
+authors:
+  - "@elbehery"
+  - "@sherine-k"
+reviewers:
+  - "@alebedev87"
+  - "@frobware"
+  - "@Miciah"
+approvers:
+  - "@knobunc"
+  - "@tjungblu"
+creation-date: 2022-01-10
+last-updated: 2022-02-02
+status: implementable
+---
+
+# Ingress Host Network Configurable Ports
+
+## Release Signoff Checklist
+
+- [X] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement aims to enable reuse of the same nodes by 2 or more Ingress Controllers with `HostNetwork` endpoint strategy without running into port reuse conflicts. 
+This is done by allowing configuration of the ports used by the router pods within the Ingress Controller API specification.
+
+## Motivation
+
+In the current implementation, Ingress Controllers in `HostNetwork` strategy use standard ports for HTTP (80), HTTPS (443), and stats (1936) on the underlying nodes. 
+
+As a result, each Ingress Controller needs to be deployed on its own set of cluster nodes. It is impossible to reuse the same nodes for two Ingress Controllers.
+
+For use cases such as multi-tenancy, multi-stage clusters, this results in increasing the number of cluster nodes, not to meet performance requirements, but for the sole purpose of having nodes with available http, https and stats ports for a new Ingress Controller.
+
+Some clients also need to use the same nodes as infra and master combined. In such cases, it won't be possible to use port 443 for master api service as it is used by the router pod(s). In such a case, it could be helpful to give the cluster admin the ability to specify the port used by the ingress controller. 
+
+### Goals
+
+The goal of this enhancement is to allow cluster nodes to host router pods for more than one Ingress Controller with `HostNetwork` endpoint strategy. 
+
+This is achieved by allowing each Ingress Controller to run on a different set of ports, thus solving the issue of concurrency on the same ports of a cluster node.
+
+The Ingress Controller's port number become configurable in the API specification for protocols:
+- HTTP
+- HTTPS
+- Stats
+
+### Non-Goals
+
+- Allowing configuration of custom ports on the ingress controllers that utilise any endpoint publishing strategy other than `HostNetwork`.
+
+## Proposal
+
+Configuring http, https and stats port for the ingress controllers that use the `HostNetwork` endpoint publishing strategy.
+
+
+Update the IngressController API by adding port configuration fields in `HostNetwork` field:
+
+```go
+    // HostNetworkStrategy holds parameters for the HostNetwork endpoint publishing
+    // strategy.
+    type HostNetworkStrategy struct {
+        // ...
+        // HTTPPort is the port on the host which should be used to listen for
+        // HTTP requests.
+        // Define this field when default port 80 is known to be used by another process.
+        // Please avoid specifying a port within the nodeport range, which defaults to 
+        // 30000-32767, but could be configured differently.
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:validation:Maximum=65535
+        // +kubebuilder:validation:Minimum=1
+        // +kubebuilder:default=80
+        HTTPPort int32 `json:"httpPort"`
+
+        // HTTPSPort is the port on the host which should be used to listen for
+        // HTTPS requests.
+        // Define this field when default port 443 is known to be used by another process. 
+        // Please avoid specifying a port within the nodeport range, which defaults to 
+        // 30000-32767, but could be configured differently.
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:validation:Maximum=65535
+        // +kubebuilder:validation:Minimum=1
+        // +kubebuilder:default=443
+        HTTPSPort int32 `json:"httpsPort"`
+
+        // StatsPort is the port on the host where the stats from the router are
+        // published.
+        // Please avoid specifying a port within the nodeport range, which defaults to 
+        // 30000-32767, but could be configured differently
+        // If you configure an external load-balancer to forward 
+        // connections this ingress controller, the load balancer should 
+        // use this port for health checks. 
+        // A load balancer can send HTTP probes to this port on a given 
+        // node with the path /healthz/ready to determine whether the 
+        // ingress controller is ready to receive traffic on the node.
+        // For proper operation, the load balancer must not forward traffic 
+        // to a node until /healthz/ready reports ready, and the load 
+        // balancer must stop forwarding within a maximum of 45 seconds 
+        // after /healthz/ready starts reporting not-ready. 
+        // Probing every 5 or 10 seconds, with a 5-second timeout and with
+        // a threshold of two successful or failed requests to become 
+        // healthy or unhealthy, respectively, are well-tested values.
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:validation:Maximum=65535
+        // +kubebuilder:validation:Minimum=1
+        // +kubebuilder:default=1936
+        StatsPort int32 `json:"statsPort"`
+      }
+```
+
+Here's an example of a public ingress controller and an internal ingress controller with the HostNetwork strategy that run on the same nodes without port conflicts:
+
+```yaml
+    ---
+    apiVersion: operator.openshift.io/v1
+    kind: IngressController
+    metadata:
+      name: public
+      namespace: openshift-ingress-operator
+    spec:
+      domain: foo.com
+      endpointPublishingStrategy:
+        type: HostNetwork
+    # Without any HostNetwork field under endpointPublishingStrategy, the public IngressController will be using the default ports: 
+    # 80, 443, 1936
+    # More spec fields here, configuring selectors and so forth...
+    ---
+    apiVersion: operator.openshift.io/v1
+    kind: IngressController
+    metadata:
+      name: internal
+      namespace: openshift-ingress-operator
+    spec:
+      domain: private.foo.com
+      endpointPublishingStrategy:
+        type: HostNetwork
+        hostNetwork:
+          httpPort: 8080
+          httpsPort: 8443
+          statsPort: 8936
+      # More spec fields here, configuring selectors and so forth...
+
+    ---
+
+```
+### Validation
+
+The port used for `spec.endpointPublishingStrategy.hostNetwork.*Port` should be greater than `1`, and lower than `30000` (to avoid conflicts with 30000-32767 nodePort services range).
+
+The `spec.endpointPublishingStrategy.hostNetwork` should not be set if the `endpointPublishingStrategy` is not of type `HostNetwork`.
+
+The status of the Ingress Controller should report if it wasn't able to run the requested number of replicas because of the unavailability of the requested ports on the available nodes.
+
+### User Stories
+
+#### As a cluster administrator, I need to set binding ports for the IngressController so that I can deploy multiple ingress controllers on the same node for the HostNetwork strategy
+
+The cluster administrator can set the ingress controller's `spec.endpointPublishingStrategy.hostNetwork`.
+
+#### As an OpenShift developer, I want to make sure Ingress Controller documentation is updated with this new feature concerning the HostNetwork strategy
+
+#### As an OpenShift developer, automated e2e tests cover the case of deploying 2 ingress controllers in HostNetwork strategy to the same nodes with different sets of ports, so that I can validate the feature
+
+#### As a Cluster Administrator, I want the IngressController status to report problems provisioning the Ingress Controller pods in HostNetwork strategy when there is a port conflict
+
+### API Extensions
+
+This  proposal edits `IngressController.operator.openshift.io/v1`.
+
+### Operational Aspects of API Extensions
+
+#### External load-balancer health checks setup
+
+An external load balancer can send HTTP probes to the ingress controller's StatsPort on a given node with the path /healthz/ready to determine whether the ingress controller is ready to receive traffic on the node. 
+Sending health check requests every 5 or 10 seconds, with a 5-second timeout and with a threshold of two successful or failed requests to become healthy or unhealthy, respectively, are well-tested values.
+For proper operation, the load balancer must not forward traffic to a node until /healthz/ready reports ready, and the load balancer must stop forwarding within a maximum of 45 seconds after /healthz/ready starts reporting not-ready. 
+
+#### Investigating issues
+
+Cluster administrators should be aware of the host ports that are used by the OpenShift cluster. These ports are referenced in the [host-port registry](https://github.com/openshift/openshift-docs/blob/main/modules/installation-network-user-infra.adoc). 
+
+More generally, they should watch out for the ingress controller status after updates:
+
+In case the router pod cannot be scheduled due to port unavailability (or any other scheduling concerns), the ingress controller's conditions will report the scheduling failure.
+
+In such a case, the cluster administrator might need to inspect other pods and system daemonsets that use the same port, using `ss` or `netstat`. Router pod logs, OVN/SDN logs might also be helpful to identify the root cause for scheduling failure.
+
+Cluster administrators might also encounter issues if an ingresscontroller with HostNetwork strategy is using a port from the nodePort range: if a nodeport service attempts to use the same port, the router pod will not be able to get scheduled correctly. The nodePort ingress controller's conditions should report the scheduling failure.
+
+
+### Implementation Details
+
+Implementing this needs changes in the following repositories:
+
+* openshift/api
+* openshift/cluster-ingress-operator
+
+OpenShift's ingress operator creates a deployment for router pods with environment variables for port bindings, which OpenShift router already respects:
+`ROUTER_SERVICE_HTTP_PORT`, `ROUTER_SERVICE_HTTPS_PORT`, `STATS_PORT`.
+
+Also the deployments should not have ip:port conflicts for `SNI_PORT` and `NO_SNI_PORT` since [4.9](https://github.com/openshift/router/pull/326):
+
+The router gets rid of the loopback hop  by using a Unix domain socket instead of TCP sockets to connect the be_sni backend to the fe_sni frontend and the be_no_sni backend to the fe_no_sni frontend.
+
+### Risks and Mitigations
+
+#### Failure Modes
+
+* A router pod could fail to be scheduled if no node had compatible labels and taints and available ports; in this case, the ingresscontroller's status conditions should report the scheduling failure.
+* A router pod could be scheduled to a node where the configured port were in use by something other than a pod, such as a system daemon (the Kubernetes schedule only checks for conflicts with ports allocated by Kubernetes); in this case, the ingresscontroller's status conditions should reflect the pod's failure to start.
+* A router pod could be scheduled and start listening on a port inside the nodeport range; in this case, a subsequently created nodeport service could by chance be assigned the in-use port and fail to be configured on the host with the router pod.
+
+#### Support Procedures
+
+a. Checking ingresscontroller's status conditions
+
+```sh
+oc get IngressController -n openshift-ingress-operator
+```
+In case of a failure to schedule the router pods, the result should at least show, in the status conditions:
+```yaml
+  - lastTransitionTime: "2022-01-05T12:01:32Z"
+    status: "False"
+    type: PodsScheduled
+```
+
+b. Checking router pod logs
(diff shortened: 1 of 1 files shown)


enhancements/ingress/host-network-bind-options.md
```diff
@@ -0,0 +1,266 @@
...
+While utilising the HostNetwork strategy for ingress controllers, the ports 80, 443, 10443, 10444, 1936 on the host are bound by HAProxy for http, https, no_sni, sni and stats correspondingly. 
+
+Those ports might be occupied by other processes which makes it unfeasible to run multiple ingress controllers on the same node with HostNetwork strategy. 
+
+### Goals
+
+Enable running multiple set of ingress controllers on the same node with `HostNetwork` endpoint strategy. The approach is to configure the bindings for the following ports:
+- HTTP
+- HTTPS
+- Stats
```
lmzuccarelli, line 55: I know this states the obvious but maybe adding the ports and even port ranges might make it more  readable.

enhancements/ingress/host-network-bind-options.md
```diff
@@ -0,0 +1,266 @@
...
+- [X] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement updates the Ingress Controller API to allow specifying binding options in the HostNetwork strategy and allows running many instances of the ingress controller on the same node using the HostNetwork strategy, therefore avoiding port binding conflicts.
+## Motivation
```
lmzuccarelli, line 44: nit: Maybe add a new line

enhancements/ingress/host-network-bind-options.md
```diff
@@ -0,0 +1,266 @@
...
+
+While utilising the HostNetwork strategy for ingress controllers, the ports 80, 443, 10443, 10444, 1936 on the host are bound by HAProxy for http, https, no_sni, sni and stats correspondingly. 
+
+Those ports might be occupied by other processes which makes it unfeasible to run multiple ingress controllers on the same node with HostNetwork strategy.
```
lmzuccarelli, line 48: maybe something like "**These** ports might be occupied by other processes which will make it **impossible** to run multiple..."

enhancements/ingress/host-network-bind-options.md
```diff
@@ -0,0 +1,266 @@
...
+`spec.endpointPublishingStrategy.hostNetwork.bindingOptions` is valid and implies
+the default behavior, which is to use the default ports to bind to. 
+
+The API validates
+that any value provided for the ports defined in
+`spec.endpointPublishingStrategy.hostNetwork.bindingOptions` is greater than `1`, and lower than `30000` (to avoid conflicts with 30000-32767 nodePort services range).
+
+### User Stories
+
+#### As a cluster administrator, I need to set binding ports for the IngressController
```
lmzuccarelli, line 184: maybe add so that I can deploy multiple ingress controllers on the same node for the HostNetwork strategy

enhancements/ingress/host-network-bind-options.md
```diff
@@ -0,0 +1,266 @@
...
+
+* openshift/api
+* openshift/cluster-ingress-operator
+
+OpenShift Cluster Ingress Operator, creates a deployment for Router with environment variables for port bindings which OpenShift Router already respects:
+`ROUTER_SERVICE_HTTP_PORT`, `ROUTER_SERVICE_HTTPS_PORT`, `STATS_PORT`.
+
+Also the deployments should not have ip:port conflicts for `SNI_PORT` and `NO_SNI_PORT`. There are multiple ways to achieve this:
+
+- The cluster ingress operator configures the router to use a unique loopback address for the sni/nosni frontends. This loopback address would be determined deterministically, for example by hashing the ingresscontroller’s name.
```
lmzuccarelli, line 204: nit: not sure about "determined deterministically",  I'm not an English grammar expert but maybe something like "the loopback address is configured automatically , for example ...

lmzuccarelli:
@Elbehery - could you check and fix the lint problem ?

enhancements/ingress/host-network-bind-options.md
```diff
@@ -0,0 +1,280 @@
...
+        // See specific bindingOptions fields for more details.
+        //
+        //
+        // Setting fields within bindingOptions is generally not recommended. The
+        // default values are suitable for most configurations.
+        //
+        // +optional
+        BindingOptions *IngressControllerBindingOptions `json:"bindingOptions,omitempty"`
+    }
+    ```
```
alebedev87, line 89: This and subsequent code blocks got rendered weirdly: no syntax highlight and back tips visible.
Maybe it's the matter of indent in the raw markdown, I think the code block doesn't need the indent.

enhancements/ingress/host-network-bind-options.md
```diff
@@ -0,0 +1,280 @@
...
+        //
+        // +kubebuilder:validation:Optional
+        // +optional
+        Ports *IngressControllerPorts `json:"ports,omitempty"`
+    }
+    // IngressControllerPorts specifies ports on which ingress controller pods will be bound
+    type IngressControllerPorts struct {
+        // http defines the port number which HAProxy process binds for
+        // http connections. Setting this field is generally not recommended. However in
+        // HostNetwork strategy, default http 80 port might be occupied by other processess
```
alebedev87, line 112:
```suggestion
        // HostNetwork strategy, default http 80 port might be occupied by other processes
```

enhancements/ingress/host-network-bind-options.md
```diff
@@ -0,0 +1,280 @@
...
+        //
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:validation:Minimum=1
+        // +kubebuilder:validation:Maximum=30000
+        // +kubebuilder:default:=80
+        // +optional
+        HTTP int32 `json:"http,omitempty"`
+        // https defines the port number which HAProxy process binds for
+        // https connections. Setting this field is generally not recommended. However in
+        // HostNetwork strategy, default https 443 port might be occupied by other processess
```
alebedev87, line 122:
```suggestion
        // HostNetwork strategy, default https 443 port might be occupied by other processes
```

enhancements/ingress/host-network-bind-options.md
```diff
@@ -0,0 +1,280 @@
...
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:validation:Minimum=1
+        // +kubebuilder:validation:Maximum=30000
+        // +kubebuilder:default:=443
+        // +optional
+        HTTPS int32 `json:"https,omitempty"`
+        // stats is the port number which HAProxy process binds
+        // to expose statistics on it. Setting this field is generally not recommended.
+        // However in HostNetwork strategy, default stats port 1936 might
+        // be occupied by other processess
```
alebedev87, line 133:
```suggestion
        // be occupied by other processes
```

enhancements/ingress/host-network-bind-options.md
```diff
@@ -0,0 +1,280 @@
...
+to ensure all the ports defined in the `bindingOptions` section are unique, and return an error with a meaningful message if there are conflicting ports.
+
+#### Usage of X-Forwarded headers
+
+The default forwarded header policy of the ingress controller is to append X-Forwarded headers. This results into the insertion of `X-Forwarded-Port` header on OpenShift router side (HAProxy acts as a reverse proxy).
+If this was the first insertion (L4 load balancer in front of the router) the router's port will be the only one seen by the server (application inside OpenShift).
+And if the server relies on `X-Forwarded-*` headers (like OpenShift image registry for instance) this may give a wrong impression that the client used the port set in `X-Forwarded-Port` which in case of the custom port binding may be different.
+
+A mitigation to this risk is to change the forwarded header policy on the ingress controller or on the route level: [documentation](https://docs.openshift.com/container-platform/4.8/networking/ingress-operator.html#nw-using-ingress-forwarded_configuring-ingress).
```
alebedev87, line 237:
```suggestion
A mitigation to this risk is to change the forwarded header policy on the ingress controller or on the route level: [documentation](https://docs.openshift.com/container-platform/4.9/networking/ingress-operator.html#nw-using-ingress-forwarded_configuring-ingress).
```

enhancements/ingress/host-network-bind-options.md
Elbehery: done

Elbehery:
@alebedev87  thanks for your review, I have adjusted all of them 

cc @lmzuccarelli  @sherine-k

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -0,0 +1,297 @@
...
+            https: 8443
+            stats: 8936
+      # More spec fields here, configuring selectors and so forth...
+
+    ---
+
+```
+### Validation
+
+Setting `spec.endpointPublishingStrategy.type: HostNetwork` and ommitting
```
arjunrn, line 183:
```suggestion
Setting `spec.endpointPublishingStrategy.type: HostNetwork` and omitting 
```

Elbehery: @arjunrn  adjusted 🙏🏽

Elbehery:
@sherine-k  I added you to the authors section 👍🏽

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -0,0 +1,298 @@
...
+
+#### Support Procedures
+n
+N/A.
+
+#### Port Conflict
+
+Users might define conflicting ports which would cause HAProxy process to fail at startup.
+
+A mitigation to this risk is to implement a validation feature in the reconciliation loop of the Cluster Ingress Operator
```
arjunrn, line 246: So we're going to validate in the reconcile loop. It would be nicer to prevent the creation of `IngressController` resources with conflicting ports. For instance when a  __NodePort__ service is created with a port which is already allocated then the creation fails with an error message, like so:

> The Service "service02" is invalid: spec.ports[0].nodePort: Invalid value: 30007: provided port is already allocated

alebedev87: Yeah, would be good to have a validating webhook for the ingress operator. But for the moment the validation is done in [the reconcile loop](https://github.com/openshift/cluster-ingress-operator/blob/0d5f49acfb118d7a6c553709b730ab1e7b2a6f6c/pkg/operator/controller/ingress/controller.go#L284).

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -0,0 +1,298 @@
...
+to ensure all the ports defined in the `hostNetworkParameters` section are unique, and return an error with a meaningful message if there are conflicting ports.
+
+#### Usage of X-Forwarded headers
+
+The default forwarded header policy of the ingress controller is to append X-Forwarded headers. This results into the insertion of `X-Forwarded-Port` header on OpenShift router side (HAProxy acts as a reverse proxy).
```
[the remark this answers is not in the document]
alebedev87: The default router binds on the standard 80, 443 ports which match the ports of the LB which "frontends" the router (`endpointPublishingStrategy == LoadBalancer`).
This works well for the applications like the image registry which needs to send the token retrieval URL back to the client. The token retrieval URL uses:
- OpenShift cluster's `*.apps` wildcard DNS which get resolved to the before mentioned LB
- original port (the one appended to `X-Forwarded-Port` by the router)

So, if the router is bound on a port `10443`  instead of `443`, the image registry will send the following URL:
```
https://registry.apps.<cluster_name>.<base_domain>.:10443/openshift/token?account=my-sa&client_id=docker&offline_token=true
```
which will fail as the "frontend" LB listens on the standard port `443` not `10443`.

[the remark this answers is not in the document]
arjunrn: The default policy is `append`. So each host and port will be appended. The image registry should just use the first host and port in the headers. I don't think having non-standard behavior when one setting is changed is a good idea. It's confusing for a user who is trying to debug an issue.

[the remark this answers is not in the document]
alebedev87: The image registry does take the first host and port, it's just that the first values are the router ones as the LB in front of the OpenShift cluster is L4 one, so it doesn't append any HTTP headers.

> I don't think having non-standard behavior when one setting is changed is a good idea. It's confusing for a user who is trying to debug an issue.

By "non-standard behavior" you mean the disabling of the append of the X-Forwarded headers? And by "one setting is changed" you mean the usage of the custom ports?
If so, the idea was to warn about this potential issue but keep the EP reasonably small moving the responsibility about using X-Forwarded headers to the applications. Do you have any idea of how to avoid the issue at reasonable cost?

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -0,0 +1,298 @@
+---
+title: ingress-host-network-configurable-ports
+
```
tjungblu, line 3: that's extremely picky, but what's up with all those newlines here? this seems to also prevent the markdown renderer to show this as a table

enhancements/ingress/host-network-configurable-ports.md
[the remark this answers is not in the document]
arjunrn: Discussed this offline. This mitigation is not to be implemented in the controller but can be applied by the user to workaround any issues which might arise from components which use the `X-Forwarded-Port` to reconstruct the request URL.

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Proposal
+
+Configuring http, https and stats port for the ingress controllers that use the `HostNetwork` endpoint publishing strategy.
+
+
+Update the IngressController API by adding an optional `HostNetworkParameters` field with type `*HostNetworkParameters` to the `HostNetworkStrategy` struct:
+
+```go
```
arjunrn, line 76: This multiple level of structs for the configuration seems redundant since the structs contain a single field. The _HTTP_, _HTTPS_ and _STATS_ port configuration can be moved into `HostNetworkStrategy` directly.

alebedev87: I agree with the idea of the simplification. But I think it can be done in many ways.

1) What Arjun proposed:
```yaml
endpointPublishingStrategy:
  type: HostNetwork
  hostNetwork:
    httpPort: 10080
    httpsPort: 10443
    statsPort: 11936
```
2) Similar to Arjun's one but with 1 level more to avoid `Port` suffix:
```yaml
endpointPublishingStrategy:
  type: HostNetwork
  hostNetwork:
    ports:
      http: 10080
      https: 10443
      stats: 11936
```
3) What Sherine put into EP:
```yaml
endpointPublishingStrategy:
  type: HostNetwork
  hostNetwork:
     hostNetworkParameters:
       httpPort: 10080
       httpsPort: 10443
       statsPort: 11936
```

sherine-k: Removing the `hostNetworkParameters` level makes sense too.
I vote for # 1

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -0,0 +1,266 @@
...
+
+N/A.
+
+## Drawbacks
+
+N/A.
+
+## Alternatives
+
+N/A.
```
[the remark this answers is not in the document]
alebedev87: How the cluster ingress operator can find an available port from the nodeport range in a race free way?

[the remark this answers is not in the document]
alebedev87: I see, on the level of the contract with the user of the API, specifying host ports or node ports are equally prone to the conflicts.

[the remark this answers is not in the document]
sherine-k: @Miciah , @arjunrn & @alebedev87 
From the information we got from the PMs(@lmzuccarelli + @Deepthidharwar), other customers are still interested in implementing this change for hostNetwork.

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -0,0 +1,297 @@
...
+        // published.
+        // Please avoid specifying a port within the nodeport range, which defaults to 
+        // 30000-32767, but could be configured differently
+        // If you configure an external load-balancer to forward 
+        // connections this ingress controller, the load balancer should 
+        // use this port for health checks. 
+        // A load balancer can send HTTP probes to this port on a given 
+        // node with the path /healthz/ready to determine whether the 
+        // ingress controller is ready to receive traffic on the node.
+        // For proper operation, the load balancer must not forward traffic // to a node until /healthz/ready reports ready, and the load
```
Miciah, line 101:
```suggestion
        // For proper operation, the load balancer must not forward traffic
        // to a node until /healthz/ready reports ready, and the load 
```

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -0,0 +1,297 @@
...
+        // HTTPPort is the port on the host which should be used to listen for
+        // HTTP requests.
+        // Define this field when default port 80 is known to be used by another process.
+        // Please avoid specifying a port within the nodeport range, which defaults to 
+        // 30000-32767, but could be configured differently.
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:validation:Maximum=65535
+        // +kubebuilder:validation:Minimum=1
+        // +kubebuilder:default=80
+        HTTPPort int `json:"HTTPPort"`
```
Miciah, line 78: This field and the other port fields should have type `int32`; see <https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types>.

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -182,7 +173,20 @@ This  proposal edits `IngressController.operator.openshift.io/v1`.
...
+
+An external load balancer can send HTTP probes to the ingress controller's StatsPort on a given node with the path /healthz/ready to determine whether the ingress controller is ready to receive traffic on the node. 
+Sending health check requests every 5 or 10 seconds, with a 5-second timeout and with a threshold of two successful or failed requests to become healthy or unhealthy, respectively, are well-tested values.
+For proper operation, the load balancer must not forward traffic to a node until /healthz/ready reports ready, and the load balancer must stop forwarding within a maximum of 45 seconds after /healthz/ready starts reporting not-ready. 
+
+#### Investigating issues
+
+Cluster administrators should watch out for the ingress controller status after updates:
+In case the router pod cannot be scheduled due to port unavailability (or any other scheduling concerns), the ingress controller's conditions will report the scheduling failure.
+In such a case, the cluster administrator might need to inspect other pods or daemonsets that use the same port. Router pod logs, OVN/SDN logs might be helpful to identify the root cause for scheduling failure.
```
Miciah, line 186: I would say "pods and system daemons"; a likely source of conflicts is running daemons on the host that aren't managed by OpenShift. `ss` or `netstat` would be an easy way to check for conflicts.

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -182,7 +173,20 @@ This  proposal edits `IngressController.operator.openshift.io/v1`.
...
+
+Cluster administrators might also encounter issues if an ingresscontroller with HostNetwork strategy is using a port from the nodePort range: if a nodeport service attempts to use the same port, the router pod will not be able to get scheduled correctly. The nodePort ingress controller's conditions should report the scheduling failure.
```
Miciah, line 188: If a nodeport service attempts to use a port that the ingresscontroller is already using, then the nodeport service may not work properly.  I know that OVN-Kubernetes will log the conflict in its pods; I'm not sure about openshift-sdn's failure mode.

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -202,11 +206,36 @@ The router gets rid of the loopback hop  by using a Unix domain socket instead o
...
+a. Checking ingresscontroller's status conditions
+
+```sh
+oc get IngressController -n openshift-ingress-operator
+```
+In case of a failure to schedule the router pods, the result should at least show, in the status conditions:
+```yaml
+  - lastTransitionTime: "2022-01-05T12:01:32Z"
+    status: "True"
+    type: PodsScheduled
```
Miciah, line 224: In case of failure to schedule pods, the operator would report `PodsScheduled=False` (I hope).

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -202,11 +206,36 @@ The router gets rid of the loopback hop  by using a Unix domain socket instead o
...
+b. Checking router pod logs
+```sh
+oc get pods -n openshift-ingress
+# get the pod name
+oc logs pods/pod_name -n openshift-ingress
+```
+Logs reporting the failure to bind to the requested port on the underlying host should be found.
+
+c. Checking SDN/OVN logs
+Check the logs from the ovn/sdn pods on the nodes where the router pods are running for errors related to port conflicts.
```
Miciah, line 236: It would be helpful to identify the pods to check here.  You can ask the SDN team if you need help figuring out the appropriate namespace or pod names.

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -258,9 +287,11 @@ N/A.
...
 ## Drawbacks
 
-N/A.
+Additional operational complexity, which will increase the needed test matrix, frustration for users, and volume of customer cases.
 
 ## Alternatives
 
-N/A.
+* Advise users to the NodePortService endpoint publishing strategy type. 
+* Augment the NodePortService strategy to allow the user to specify requested ports within the nodeport range.
```
Miciah, line 295: Usually this section will include some explanation of why the alternatives were rejected.  Do we have that information?  Maybe it's as simple as that these alternatives have been rejected by a subset of users who require HostNetwork for unspecified reasons.  If we do know a reason why these alternatives were rejected, this would be the place to record it.

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -0,0 +1,297 @@
...
+        // HTTPSPort is the port on the host which should be used to listen for
+        // HTTPS requests.
+        // Define this field when default port 443 is known to be used by another process. 
+        // Please avoid specifying a port within the nodeport range, which defaults to 
+        // 30000-32767, but could be configured differently.
+        // +kubebuilder:validation:Optional
+        // +kubebuilder:validation:Maximum=65535
+        // +kubebuilder:validation:Minimum=1
+        // +kubebuilder:default=443
+        HTTPSPort int `json:"HTTPSPort"`
```
alebedev87, line 89: It's not the API definition, so it's less important, however to match with the example below (and with the most API examples actually): JSON tag names are better to be lower camel case.

enhancements/ingress/host-network-configurable-ports.md
squeed: We actually have a host-port registry -- https://github.com/openshift/enhancements/blob/master/dev-guide/host-port-registry.md -- as well as user-facing documentation (https://github.com/openshift/openshift-docs/blob/main/modules/installation-network-user-infra.adoc) for which ports we "claim". So end-user documentation should reference those ranges as being off-limits.

enhancements/ingress/host-network-configurable-ports.md
```diff
@@ -0,0 +1,312 @@
...
+
+SDN:
+```sh
+oc logs -f ds/sdn -c sdn -n openshift-sd
+E0228 16:22:00.629556    5050 proxier.go:1303] "can't open port, skipping it" err="listen tcp4 :32634: bind: address already in use" port="\"nodePort for ***\" (:32634/tcp4)"
+```
+OVN:
+oc logs -f ds/ovnkube-node -n openshift-ovn-kubernetes -c ovnkube-node
+W0301 17:17:48.060076    5734 port_claim.go:119] PortClaim for xxxx: xxxx/xxxxxxxx on port: 32634, err: listen tcp :32634: bind: address already in use
+E0301 17:17:48.060097    5734 port_claim.go:145] Error claiming port for service: xxxx/xxxxxxxx: listen tcp :32634: bind: address already in use
```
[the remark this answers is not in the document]
sherine-k: Updated. Thanks

PR #1005, opened 2022-01-12
Non-disruptive upgrades in DPU clusters

danwinship:
In a [cluster with DPUs](https://github.com/openshift/enhancements/blob/master/enhancements/network/dpu/overview.md) (eg, BlueField-2 NICs), the x86 hosts form one OCP cluster, and the DPU ARM systems form a second OCP cluster. This makes upgrades to new OCP releases complicated, because there is currently no way to synchronize upgrades between the two clusters, but rebooting the BF-2 systems as part of the MCO upgrade will cause a network outage on the x86 systems. In order for upgrades to work smoothly, we need some way to ~synchronize the reboots between the two clusters, so that the BF-2 systems are only rebooted when their corresponding x86 hosts have been cordoned and drained~ ensure that tenant nodes do not get rebooted when they are running user workloads..

cc @zshi @pliurh @Billy99

### DIFF
--- a/enhancements/network/dpu/upgrades.md
+++ b/enhancements/network/dpu/upgrades.md
@@ -1,0 +1,318 @@
+---
+title: Non-Disruptive DPU Infra Cluster Upgrades
+authors:
+  - "@danwinship"
+reviewers:
+  - "@dhellmann"
+  - "@sdodson"
+  - "@wking"
+  - "@romfreiman"
+  - "@zshi"
+approvers:
+  - TBD
+api-approvers: # in case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers)
+  - TBD
+creation-date: 2021-01-11
+last-updated: 2021-01-11
+tracking-link:
+  - https://issues.redhat.com/browse/SDN-2603
+see-also:
+  - "/enhancements/network/dpu/overview.md"
+---
+
+# Non-Disruptive DPU Infra Cluster Upgrades
+
+## Summary
+
+Our architecture for [using DPUs (eg, BlueField-2 NICs) in
+OCP](../network/dpu/overview.md) involves creating one ordinary
+"tenant" OCP cluster consisting of x86 hosts, and a second "infra" OCP
+cluster consisting of the ARM systems running on the NICs installed in
+the x86 hosts.
+
+In the current Dev Preview of DPU support, there is no coordination
+between the infra and tenant clusters during upgrades. Thus, when
+upgrading the infra cluster, each NIC will end up rebooting at some
+point, after it has drained _its own_ pods, but without having made
+any attempt to drain the pods of its x86 tenant node. When this
+happens, the x86 node (and its pods) will lose network connectivity
+until the NIC finishes rebooting.
+
+In order to make upgrades in DPU clusters work smoothly, we need some
+way to ensure that tenant nodes do not get rebooted when they are
+running user workloads.
+
+## Glossary
+
+(Copied from the [DPU Overview
+Enhancement](../network/dpu/overview.md).)
+
+- **DPU** / **Data Processing Unit** - a Smart NIC with a full CPU,
+  RAM, storage, etc, running a full operating system, able to offload
+  network processing and potentially other tasks from its host system.
+  For example, the NVIDIA BlueField-2.
+
+- **DPU NIC** - refers to the DPU and its CPU, OS, etc (as opposed to
+  the CPU, OS, etc, of the x86 server that the DPU is installed in). A
+  **DPU NIC node** or **DPU NIC worker** is an OCP worker node
+  running on a DPU NIC.
+
+- **DPU Host** - refers to an x86 server (and its CPU, OS, etc) which
+  contains a DPU. A **DPU Host node** or **DPU Host worker** is an OCP
+  worker node running on a DPU Host.
+
+- **Two-Cluster Design** - any architecture for deploying an OCP
+  cluster with (some) nodes containing DPUs, where the DPUs are nodes
+  in a second OCP cluster. (For purposes of this term, the
+  architecture still counts as "two-cluster" even if HyperShift is
+  involved and there are actually three clusters.)
+
+- **Infra Cluster** - in a Two-Cluster Design, the OCP cluster which
+  contains the DPU NIC workers. (It may also include master nodes
+  and/or non-DPU workers, depending on the particular details of the
+  two-cluster design.)
+
+- **Tenant Cluster** - in a Two-Cluster Design, the OCP cluster
+  containing the DPU Host workers. (It may also include master nodes
+  and/or workers without DPUs.)
+
+## Motivation
+
+### Goals
+
+- Make upgrades of DPU Infra Clusters work without interrupting
+  network connectivity in their corresponding Tenant Clusters.
+
+### Non-Goals
+
+- Requiring Infra and Tenant Clusters to upgrade at the same time;
+  either cluster should be able to do an upgrade without involving the
+  other.
+
+- _Allowing_ Infra and Tenant Clusters to upgrade _efficiently_ at the
+  same time. If an administrator wants to upgrade both clusters at the
+  same time, then we would ideally do this in a way that would only
+  drain each Tenant node once. But this would require substantially
+  more synchronization between the two clusters, and is not handled by
+  this proposal.
+
+## Proposal
+
+### User Stories
+
+#### User Story 1
+
+As the administrator of a cluster using DPUs, I want to be able to do
+y-stream and z-stream upgrades without causing unnecessary network outages.
+
+#### User Story 2
+
+As an infrastructure provider of clusters using DPUs, I want to be
+able to upgrade the DPU Infra Clusters without having to coordinate
+with the administrators of the individual Tenant Clusters.
+
+### API Extensions
+
+TBD
+
+### Implementation Details/Notes/Constraints
+
+#### Setup
+
+We can set some things up at install time (eg, creating credentials to
+allow certain operators in the two clusters to talk to each other).
+
+#### Inter-Cluster Version Skew
+
+There is very little explicit communication between the two clusters:
+
+  1. The dpu-operator in the infra cluster talks to the apiserver in
+     the tenant cluster.
+
+  2. The ovn-kubernetes components in the two clusters do not
+     communicate directly, but they do make some assumptions about
+     each other's behavior (which is necessary to coordinate the
+     plumbing of pod networking).
+
+The apiserver communication is very ordinary and boring and not likely
+to be subject to any interesting version skew problems (even if we
+ended up with, say, a 3-minor-version skew between the clusters).
+
+Thus, other than ovn-kubernetes, no OCP component needs to be
+concerned about version skew between the two clusters, because the
+components in the two clusters are completely unaware of each other.
+
+For ovn-kubernetes, if we ever change the details of the cross-cluster
+communication, then we will need to add proper checks to enforce
+tolerable cross-cluster skew at that time.
+
+## Design Details
+
+If an Infra Node is being cordoned and drained, then we can assume
+that the administrator (or an operator) is planning to do something
+disruptive to it. Anything that disrupts the Infra Node will disrupt
+its Tenant Node as well.
+
+Therefore, if an Infra Node is being cordoned and drained, we should
+respond by cordoning and draining its Tenant Node, and not allowing
+the Infra Node to be fully drained until the Tenant has also been
+drained.
+
+Given that behavior, when an Infra Cluster upgrade occurs, if it is
+necessary to reboot nodes, then each node's tenant should be drained
+before the infra node reboots, so there should be no pod network
+disruption.
+
+### "Drain Mirroring" Controller
+
+In order to prevent MCO from rebooting an Infra node until its Tenant
+has been drained, we will need to run at least one drainable pod on
+each node, so that we can refuse to let that pod be drained until the
+infra node's tenant is drained.
+
+Annoyingly, although we need this pod to run on every worker node, it
+can't be deployed by a DaemonSet, because DaemonSet pods are not
+subject to draining. Thus, we will have to have the DPU operator
+manage this manually. To minimize privileges and cross-cluster
+communication, it would also make sense to have the DPU operator
+handle the actual mirroring of infra state to the tenant cluster,
+rather than having each infra node manage its own tenant node's state.
+
+So, the "drain mirroring" controller of the DPU operator will:
+
+  - Deploy a "drain blocker" pod to each infra worker node. This pod
+    will do nothing (ie, it just runs the "pause" image, or
+    equivalent), and will have a
+    `dpu-operator.openshift.io/drain-mirroring` finalizer.
+
+  - Whenever an infra worker is marked `Unschedulable`, mark its
+    tenant node `Unschedulable` as well, with a `Condition` indicating
+    why it was made unschedulable.
+
+  - If a "drain blocker" pod on an unschedulable infra node is marked
+    for deletion, then drain the corresponding tenant node as well
+    before removing the finalizer from the pod.
+
+  - Whenever an infra worker is marked not-`Unschedulable`:
+
+      - If we were in the process of draining its tenant, then abort.
+
+      - If the tenant node is also `Unschedulable` and has a
+        `Condition` indicating that we are the one who made it
+        `Unschedulable`, then remove the `Condition` and the
+        `Unschedulable` state from the tenant node.
+
+      - If the infra node does not currently have a "drain blocker"
+        pod, then deploy a new one to it.
+
+### Open Questions
+
+Do finalizers block draining/eviction, or do we need to set
+`TerminationGracePeriodSeconds` too? (If the latter, then this gets
+more complicated since we need to control when the "drain blocker"
+actually exits.)
+
+### Risks and Mitigations
+
+TBD
+
+### Test Plan
+
+TBD
+
+### Graduation Criteria
+
+TBD
+
+#### Dev Preview -> Tech Preview
+
+#### Tech Preview -> GA
+
+#### Removing a deprecated feature
+
+### Upgrade / Downgrade Strategy
+
+This is a modification to the upgrade process, not something that can
+be upgrade or downgraded on its own.
+
+TBD, as the details depend on the eventual design.
+
+### Version Skew Strategy
+
+TBD, as the details depend on the eventual design.
+
+### Operational Aspects of API Extensions
+
+TBD
+
+#### Failure Modes
+
+TBD
+
+#### Support Procedures
+
+TBD
+
+## Implementation History
+
+- Initial proposal: 2021-01-11
+- Updated for initial feedback: 2021-01-24
+- Updated again: 2021-03-02
+
+## Drawbacks
+
+This makes the upgrade process more complicated, which risks rendering
+clusters un-upgradeable without manual intervention.
+
+However, without some form of synchronization, it is impossible to
+have non-disruptive tenant cluster upgrades.
+
+## Alternatives
+
+### Never Reboot the DPUs
+
+This implies never upgrading OCP on the DPUs. I don't see how this
+could work.
+
+### Don't Have an Infra Cluster
+
+If the DPUs were not all part of a single OCP cluster (for example,
+they were just "bare" RHCOS hosts, or they were each running
+Single-Node OpenShift), then it might be simpler to synchronize the
+DPU upgrades with the tenant upgrades, because then each tenant could
+coordinate the actions of its own DPU by itself.
+
+The big problem with this is that, for security reasons, we don't want
+the tenants to have any control over their DPUs. (For some future use
+cases, the DPUs will be used to enforce security policies on their
+tenants.)
+
+### Synchronized Infra and Tenant Cluster Upgrades
+
+In [the original proposal], the plan was for infra and tenant cluster
+uprades to be synchronized, with the Infra and Tenant MCOs
+communicating to ensure that workers were upgraded in the same order
+in both clusters, and that reboots of the Infra and Tenant workers
+happened at the same time.
+
+After some discussion, we decided we didn't need that much
+synchronization.
+
+[the original proposal]: https://github.com/openshift/enhancements/commit/8033a923
+
+### Simultaneous but Asynchronous Infra and Tenant Cluster Upgrades
+
+[The second version of the proposal] took advantage of the fact that
+it's presumed safe to reboot infra nodes whenever their tenant nodes
+are idle, but still assumed that the infra and tenant clusters would
+be upgraded at the same time. In this version, the two clusters
+upgraded independently until they reached the MCO upgrade stage. Then
+the infra cluster MCO was supposed to queue up RHCOS changes without
+actually rebooting nodes, and then the tenant cluster MCO would
+proceed with its upgrade, and each infra worker node would watch to
+see when its tenant rebooted, and reboot itself at the same time.
+
+This approach also had various problems, and probably would not have
+worked well.
+
+[The second version of the proposal]: https://github.com/openshift/enhancements/commit/b4650860


enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
+---
+title: Synchronized Upgrades Between Clusters
+authors:
+  - "@danwinship"
+reviewers:
+  - TBD
```
dhellmann, line 6: @sdodson and someone from the OTA team like @LalatenduMohanty or @wking should be included as reviewers

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+This makes the upgrade process more complicated, which risks rendering
+clusters un-upgradeable without manual intervention.
+
+However, without some form of synchronization, it is impossible to
+have non-disruptive tenant cluster upgrades.
+
+## Alternatives
+
+The fundamental problem is that rebooting the DPU causes a network
+outage on the tenant.
```
dhellmann, line 205: If the tenant cluster is not running any workloads, does that network outage cause any real issues? Could this problem be addressed by an outside orchestration tool cordoning and draining both the tenant and DPU clusters, upgrading the tenant cluster, then upgrading the DPU cluster, then removing the cordon from both?

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+This implies never upgrading OCP on the DPUs. I don't see how this
+could work.
+
+### Don't Have an Infra Cluster
+
+If the DPUs were not all part of a single OCP cluster (for example,
+they were just "bare" RHCOS hosts, or they were each running
+Single-Node OpenShift), then it might be simpler to synchronize the
+DPU upgrades with the tenant upgrades, because then each tenant could
+coordinate the actions of its own DPU by itself.
```
dhellmann, line 218: Are the DPUs linked together to form a multi-node cluster across DPUs in a host? Or across multiple hosts?

romfreiman:
@tsorya

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Risks and Mitigations
+
+TBD
+
+## Design Details
+
+### Open Questions
+
+Basically everything...
```
romfreiman, line 72: There is an effort of central upgrade service happening for Telco. I assume we can reuse some bits there.

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+  to upgrade.
+
+  - In particular:
+
+    - There can be no `Upgradeable: False` operators in either cluster.
+
+    - The version to upgrade to must be available to both clusters
+      (ie, it must be available for both x86 and ARM).
+
+  - This could be implemented via some sort of "dpu-cluster-upgrade"
```
romfreiman, line 100: I would not have communications between the clusters, given the fact that the goal of the separation is security. Why not to have an external upgrade-operator (runs on the hub), that can correlate and orchestrate upgrade flows between clusters.

romfreiman: The goal is to provide central mgmt, not a standalone system that should orchestrate itself.

enhancements/update/synchronized-upgrades.md
danwinship: Multiple hosts. Every worker in the tenant cluster has 0 or 1 DPUs, and if they have a DPU, it's their primary network interface.

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+
+# Synchronized Upgrades Between Clusters
+
+## Summary
+
+In a [cluster with DPUs](../network/dpu/overview.md) (eg, BlueField-2
+NICs), the x86 hosts form one OCP cluster, and the DPU ARM systems
+form a second OCP cluster. This makes upgrades to new OCP releases
+complicated, because there is currently no way to synchronize upgrades
+between the two clusters, but rebooting the BF-2 systems as part of
```
wking, line 27: nit: `overview.md` link was helpful, but can we stick to terms from the glossary (like "DPU NIC") instead of using "BF-2" (a specific subset of possible BPU NICs)?

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+    - The Infra MCO will cordon and drain that host's infra node, and
+      the Tenant MCO will cordon and drain that host's tenant node.
+      (This can happen in parallel.)
+
+    - The Infra MCO will then upgrade the infra node (causing it to
+      reboot and temporarily break network connectivity to the tenant
+      node).
+
+    - Once the infra node upgrade completes, the Tenant MCO will
+      reboot and upgrade the tenant node.
```
sdodson, line 133: Does rebooting the tenant node implicitly restart the infra node or does the infra node remain online during a tenant node reboot? Does this vary by either infra node or tenant node hardware or even configuration such as BIOS settings?

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+
+- As part of the DPU security model, the tenant cluster cannot have
+  any power over the infra cluster. (In particular, it can't be
+  possible for an administrator in the tenant cluster to force the
+  infra cluster to upgrade/downgrade to any particular version.) Thus,
+  the upgrade must be initiated on the infra cluster side, and the
+  infra side will tell the tenant cluster to upgrade as well.
+  (Alternatively, the upgrade must be simultaneous-ish-ly initiated in
+  both clusters, if we don't want the infra cluster to have to have a
+  credential that lets it initiate an upgrade in the tenant cluster.)
```
sdodson, line 88: However, tenant node state inhibiting updates of infra node is not a violation of the security model?

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+
+    - Once the infra node upgrade completes, the Tenant MCO will
+      reboot and upgrade the tenant node.
+
+  - (This seems like it will absolutely require MCO changes.)
+
+  - One way to do this would be to have a CRD with an array of hosts,
+    indicating the ordering, and the current status of each host, and
+    the two MCOs could update and watch for updates to monitor each
+    other's progress.
```
sdodson, line 140: How do the infra and tenant nodes known which correspond to one another?

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Drawbacks
+
+This makes the upgrade process more complicated, which risks rendering
+clusters un-upgradeable without manual intervention.
+
+However, without some form of synchronization, it is impossible to
+have non-disruptive tenant cluster upgrades.
+
+## Alternatives
```
sdodson, line 202: If the infra node workload only services the local tenant node then the infra cluster could update including staging machineconfig and RHCOS changes. Once that completes the tenant cluster updates and the MCO reboots tenant nodes which reboots the infra node into the updated RHCOS?

If an unexpected reboot seems like a concern perhaps the MCO on the infra node could be made to only apply the new desired configuration on a soft restart assuming rebooting the tenant triggers a soft reboot of the infra node?

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+### Non-Goals
+
+- Supporting synchronized upgrades of more than 2 clusters at once.
+
+## Proposal
+
+### User Stories
+
+As the administrator of a cluster using DPUs, I want to be able to do
+z-stream upgrades without causing unnecessary network outages.
```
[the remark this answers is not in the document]
danwinship: No, this applies to y-stream upgrades too. It's just that you might be able to convince people that y-stream upgrades are allowed to have a little bit of disruption since they're rare, but you can't make that argument for z-stream upgrades.

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+In a [cluster with DPUs](../network/dpu/overview.md) (eg, BlueField-2
+NICs), the x86 hosts form one OCP cluster, and the DPU ARM systems
+form a second OCP cluster. This makes upgrades to new OCP releases
+complicated, because there is currently no way to synchronize upgrades
+between the two clusters, but rebooting the BF-2 systems as part of
+the MCO upgrade will cause a network outage on the x86 systems. In
+order for upgrades to work smoothly, we need to synchronize the
+reboots between the two clusters, so that the BF-2 systems are only
+rebooted when their corresponding x86 hosts have been cordoned and
+drained.
```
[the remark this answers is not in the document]
danwinship: Currently (in dev preview) the only coupling between the two clusters is with ovn-kubernetes. This enhancement may result in other components becoming coupled too (MCO, "dpu-cluster-upgrade" operator, etc).

I'm just assuming there's no way we can avoid there ever being skewed versions between the clusters, and hence we will just have to cope with that (eg, any API changes between them will have to be eased in over multiple releases).

So yes, everything on the infra cluster could be updated ahead of OVN on the host cluster.

enhancements/update/synchronized-upgrades.md
danwinship: In our lab, rebooting the tenant does not restart the NIC, and AFAIK this does not vary by hardware. (ie, ACPI/EFI say that on reboot you send a PCI reset but you don't cut power to the bus.) IIRC an earlier (prerelease) version of the BF-2 did reboot on reboot (I guess because it was designed to reboot on PCI reset) but the GA ones do not.

If there are situations where the NIC reboots on host reboot (including the possibility of other non-BF-2 DPUs in the future that behave that way), then it would only be a small change to the procedure here; instead of draining both nodes, rebooting the NIC, and then rebooting the host, we would drain both nodes, then reboot the host without explicitly rebooting the NIC.

enhancements/update/synchronized-upgrades.md
danwinship: A good point. We should think about alerts in the case where the tenant node is thwarting the infra node.

(I'm kind of assuming that any human who has cluster-admin in the infra cluster will also have cluster-admin in the tenant cluster, and so the same person would kick off the upgrade in both clusters. In the future there may be reasonable use cases where that isn't true...)

enhancements/update/synchronized-upgrades.md
danwinship: I forget, but it's already implemented. (They need that info to make OVN offload function.) I think currently the tenant nodes annotate their Node with the MAC of the DPU or something, and then the infra nodes can match themselves up. In the future when we have a more synchronized install system as well, then we can just figure it all out at install time.

enhancements/update/synchronized-upgrades.md
danwinship: So as above, rebooting the tenant doesn't reboot the infra node, although I suspect there's probably some way we can detect when the host reboots.

So I guess with what you're suggesting, the only coordination that would be needed would be that you have to run the upgrade to almost-completion in the infra cluster, and then you start the upgrade in the tenant cluster, and as the tenant cluster upgrade completes it will also cause the completion of the infra cluster upgrade. Right? There'd still need to be (small) MCO changes to prevent the automatic reboot in the infra cluster.

sdodson: Yeah that's what I was thinking.

danwinship: It looks like it might not be possible for the BF-2 to directly detect a host reboot, at least when the card is in "isolated" mode.

But the infra node knows what pod-network pods are running on the tenant node, so it can infer that the tenant has been drained when (most of?) the pod-network pods go away. (If necessary, the dpu-operator could confirm this by looking at the Node object in the tenant cluster.) So we could say, if we have a reboot pending, and the tenant gets drained, then reboot. Even if we guessed wrong and the drain wasn't actually because of the MCO upgrade on the tenant, we still would at least avoid most of the disruption.

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+  need to coordinate to complete the upgrade: First, they have to
+  agree on what order the physical hosts will be upgraded in. Second,
+  for each physical host, they have to properly synchronize the
+  upgrades of its infra node and its tenant node.
+
+  - More specifically, for each physical host, in some order:
+
+    - The Infra MCO will cordon and drain that host's infra node, and
+      the Tenant MCO will cordon and drain that host's tenant node.
+      (This can happen in parallel.)
```
zshi-redhat, line 126: This may impose another sync in the application pod level on which dest nodes the drained pod (infra and host) shall be placed to, since the DPU pods need to be rescheduled to the dest DPU node whose x86-host the application pods will be rescheduled to.

zshi-redhat, line 126: When this happens in parallel, will one MCO waits for the other until it completes the drain, then coordinate on the rebooting? especially the Infra MCO needs to wait until Tenant MCO to finish drain, otherwise infra MCO would reboot DPU which disconnect Tenant api network.

I don't see we have such wait mechanism in MCO.

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+  the upgrade must be initiated on the infra cluster side, and the
+  infra side will tell the tenant cluster to upgrade as well.
+  (Alternatively, the upgrade must be simultaneous-ish-ly initiated in
+  both clusters, if we don't want the infra cluster to have to have a
+  credential that lets it initiate an upgrade in the tenant cluster.)
+
+- An upgrade should not be able to start unless both clusters are able
+  to upgrade.
+
+  - In particular:
```
zshi-redhat, line 93: Do we want to mention version compatibility for ovn-k8s or rhcos (driver) during upgrade or is it too detailed to be mentioned in this enhancement?

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,223 @@
...
+
+TBD
+
+### Risks and Mitigations
+
+TBD
+
+## Design Details
+
+### Open Questions
```
zshi-redhat, line 70: Will this apply to hypershift as well?

enhancements/update/synchronized-upgrades.md
danwinship: > I don't see we have such wait mechanism in MCO.

Exactly. That's why we need this enhancement.

danwinship:
Updated to reflect various comments. In particular, it now uses a less-synchronized synchonization as suggested by Scott.

danwinship:
/cc @cgwalters @kikisdeliveryservice @sinnykumari @yuqi-zhang 

Need some MCO input on this.

Someone also pointed me to https://issues.redhat.com/browse/MCO-206 which seems like it could end up being useful here...

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,302 @@
...
+  occur?
+
+- How exactly will infra nodes detect their tenant node rebooting?
+
+- Do we want to allow (z-stream) upgrades of the tenant cluster
+  without also upgrading the infra cluster?
+
+### Risks and Mitigations
+
+The infra cluster could remain stuck at the pre-reboot stage of the
```
kikisdeliveryservice, line 210: The prereboot stage means that the upgrade has been performed except for the final reboot, this doesn't seem like a state we want all nodes to stay in at all?

kikisdeliveryservice (commented):
I have a lot of questions on this (and I'm sure other team members also will once they read the proposal) since I have no background on this feature outside of the enhancment. 

I really think there needs to be a conversation with the MCO as this proposes major changes to fundamental behaviors and needs some scrutiny, discussion and planning (outside of discussion on the actual enhancement PR).

cc: @craychee

kikisdeliveryservice:
Also cc: @aravindhp @mrunalp

danwinship:
> I have a lot of questions on this (and I'm sure other team members also will once they read the proposal) since I have no background on this feature outside of the enhancement.

So, abstractly, there is a "tenant" cluster, which is more-or-less a normal OCP cluster, except that its worker nodes aren't connected directly to the network; instead, each one is stuck behind its own personal firewall that monitors and filters its network access:

```
  +-----------------+    +-----------------+   +-----------------+
  | Tenant Master 1 |    | Tenant Worker A |   | Tenant Worker B |  ...
  +-------+---------+    +--------+--------+   +--------+--------+
          |                       |                     |
          |                       |                     |
          |              +--------+--------+   +--------+--------+
          |              |   Firewall A    |   |   Firewall B    |  ...
          |              +--------+--------+   +--------+--------+
          |                       |                     |         
          |                       |                     |         
----------+-----------------------+---------------------+-----------
```

We need some way to deal with installing and managing and upgrading all of these firewall hosts, and it turns out that OCP is a good way to do that... so, we create a second OCP cluster (the "infra" cluster), and then each firewall host is actually a worker node in that cluster:

```

  +-----------------+    +-----------------+   +-----------------+
  | Tenant Master 1 |    | Tenant Worker A |   | Tenant Worker B |  ...
  +-------+---------+    +--------+--------+   +--------+--------+
          |                       |                     |
          |                       |                     |
          |              +--------+--------+   +--------+--------+    +----------------+
          |              | Infra Worker A  |   | Infra Worker B  |    | Infra Master 1 |  ...
          |              +--------+--------+   +--------+--------+    +------+---------+
          |                       |                     |                    |
          |                       |                     |                    |
----------+-----------------------+---------------------+--------------------+-----------
```

So the problem (for purposes of this enhancement) is that when we want to do an upgrade of the infra cluster, we (generally) have to reboot each of the infra workers. But when an infra worker reboots, its corresponding tenant loses network connectivity until it comes back. So the goal is: **only reboot infra workers when their tenant nodes are also rebooting**. (The infra nodes do not run user workloads, they only run the stuff we deploy on them, so we know it's fine to just randomly reboot them whenever we want to, as long as their tenant is idle.)

(Of course, the infra workers are not actually physically separate hosts; they're ARM systems running on the NICs of the tenant workers. But that doesn't really matter for _this_ enhancement; the important part is just that we have to be careful when we reboot them.)

enhancements/update/synchronized-upgrades.md
danwinship: > Also what happens if the update is bad? All nodes would be scheduled to make the update as opposed to a rolling mechanism that we have now.

Yeah... so if you look at [the first draft of the enhancement](https://github.com/openshift/enhancements/commit/8033a9236774dc115633adebb2610a8fee76599f), I had suggested a more complicated system, where the MCOs of the two clusters would agree on a particular order to upgrade nodes in, and then each individual node upgrade would be synchronized. Scott suggested that we didn't actually need that much synchronization and I rewrote the enhancement to the current form. But maybe that won't work...

dhellmann:
> > I have a lot of questions on this (and I'm sure other team members also will once they read the proposal) since I have no background on this feature outside of the enhancement.
> 
> So, abstractly, there is a "tenant" cluster, which is more-or-less a normal OCP cluster, except that its worker nodes aren't connected directly to the network; instead, each one is stuck behind its own personal firewall that monitors and filters its network access:

Including some version of that explanation & the diagrams in the doc would be good. You could do them as ASCII art, or experiment with GitHub's new support for Mermaid (https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/).

enhancements/update/synchronized-upgrades.md
mrunalp: Are there any files shared between the two?

mrunalp: Reading a bit more doesn't look like there is.

derekwaynecarr:
is there a way we could model this around `Lease`s?

https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/589-efficient-node-heartbeats

could the infra cluster take a lease on the tenant node, and the mco would not reboot a node with an active lease?

other related art around node maintenance holds:
https://github.com/kubevirt/node-maintenance-operator

derekwaynecarr:
just recording input from our discussion:

- ideally we can do this with no mco changes
- run a canary pod on the infra cluster that will cordon the tenant node when its going down
- if the canary goes down, we should project some condition back to the tenant node for NetworkNotReady so new pods dont scheduled to that tenant node.

basically, lets not assume a shared maintenance window for the mvp.

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,302 @@
...
+power over the infra cluster. In particular, it can't be possible for
+an administrator in the tenant cluster to force the infra cluster to
+upgrade/downgrade to any particular version.
+
+Thus, the infra cluster upgrade must be initiated on the infra cluster
+side. In theory, we could have the infra cluster then initiate the
+tenant cluster upgrade, but that would require having a tenant cluster
+cluster-admin admin credential lying around in the infra cluster
+somewhat unnecessarily. It probably makes more sense to require the
+administrator to initiate the upgrade on the tenant side as well.
```
cgwalters, line 122: What credentials does the dpu-operator have?  I'm guessing it's at least somewhat privileged today?  Glancing at e.g. https://github.com/openshift/dpu-network-operator/blob/7750537749e734dfb6edfd25fc3e674398eef548/config/rbac/role.yaml - that's equivalent to cluster admin (although maybe only on the infra cluster).

ISTM it'd be a better user experience if the admin only needs to initiate upgrades in one place, and we can certainly design a system to do that securely.  

If we don't require the admin to do this manually by default (or script it) then we don't need to be debugging failure cases where they didn't do it.

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,302 @@
...
+The upgrade plan takes advantage of two things:
+
+- Other than the reboots, no part of the upgrade needs to be
+  synchronized between the two clusters.
+
+- Each infra node only runs pods to support the pods on its
+  corresponding tenant node. Thus, if a tenant node reboots, it is
+  safe to immediately reboot its infra node as well without needing
+  to cordon and drain it, because it is guaranteed that none of its
+  pods are doing anything important at that point.
```
cgwalters, line 159: OK...I have a potentially dumb question - when the tenant reboots, the infra node can (by default) just continue running?  I would have thought that there'd be some platform initialization (e.g. of the PCI bus) that might break that, but is it really more that the infra node maintains electrical power and just keeps running?

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,302 @@
...
+
+So, both clusters' upgrades can proceed normally up until the MCO
+upgrade.
+
+In the infra cluster, rather than cordoning, draining, and rebooting
```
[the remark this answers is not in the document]
cgwalters: I think the phrase "queue up the new image" is assuming we have https://github.com/openshift/machine-config-operator/issues/1190 but unfortunately we don't today.  So I'd agree we need to do a per pool/node pause approach.

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,302 @@
...
+nodes one-by-one, the MCO will instead queue up the new RHCOS image on
+each node, and then mark the node (somehow) as "waiting for a reboot".
+Then the MCO will watch the nodes and wait for each one to reboot (of
+its own accord), and then finally mark itself fully-upgraded after
+that happens.
+
+On the infra nodes, when they are "waiting for a reboot", they will
+monitor the state of the ovn-kubernetes network; when they determine
+that the tenant node is rebooting, the infra node will then also
+reboot.
```
cgwalters, line 174: Is this reboot also a drain?  Or can we assume there are no non-daemonset workloads on the infra nodes, and we can just kill everything and reboot?

enhancements/update/synchronized-upgrades.md
```diff
@@ -0,0 +1,302 @@
...
+
+### More closely coordinated reboots
+
+In the original proposal, the infra and tenant reboots were more
+closely coordinated, with the infra and tenant MCOs communicating so
+that the infra MCO could reboot each infra node at the same time as
+the corresponding tenant node rebooted. But this (probably) turns out
+to be unnecessary, as it should always be safe to just reboot the
+infra nodes when their tenant nodes reboot, without needing to have
+"planned" the reboot.
```
cgwalters, line 302: OK this is answering my above question - infra nodes don't need to drain (or at least, the drain should be predictable and fast).

But it also seems like it's heading long term to conflict with the above goal of doing more on the DPUs.

enhancements/update/synchronized-upgrades.md
cgwalters: Sorry, I am now more carefully reading earlier conversation and this was already answered at https://github.com/openshift/enhancements/pull/1005#discussion_r785201582

enhancements/update/synchronized-upgrades.md
cgwalters: The two systems don't share a kernel (or a common CPU architecture even), so they can't share files.

enhancements/network/dpu/upgrades.md
```diff
@@ -0,0 +1,318 @@
...
+### "Drain Mirroring" Controller
+
+In order to prevent MCO from rebooting an Infra node until its Tenant
+has been drained, we will need to run at least one drainable pod on
+each node, so that we can refuse to let that pod be drained until the
+infra node's tenant is drained.
+
+Annoyingly, although we need this pod to run on every worker node, it
+can't be deployed by a DaemonSet, because DaemonSet pods are not
+subject to draining. Thus, we will have to have the DPU operator
```
cgwalters, line 175: The daemonset draining is configurable I think; if we wanted to do that we could.  See e.g. https://github.com/openshift/machine-config-operator/blob/62fb49f2d14519e215125faa6b817662cdb7ffd0/pkg/daemon/daemon.go#L316

That said, we'd probably need to change the MCO to opt-in specific daemonsets to being drained.  This also relates to https://github.com/openshift/machine-config-operator/pull/2978 I think.

cgwalters: (And in practice, we probably do need to care about things *other* than the MCO doing draining, so relying on a daemonset drain is probably not viable in the near future)

enhancements/network/dpu/upgrades.md
```diff
@@ -0,0 +1,318 @@
...
+manage this manually. To minimize privileges and cross-cluster
+communication, it would also make sense to have the DPU operator
+handle the actual mirroring of infra state to the tenant cluster,
+rather than having each infra node manage its own tenant node's state.
+
+So, the "drain mirroring" controller of the DPU operator will:
+
+  - Deploy a "drain blocker" pod to each infra worker node. This pod
```
wking, line 183: nit: [linter][1] has some list-indent opinions:

```
enhancements/network/dpu/upgrades.md:183:1 MD007/ul-indent Unordered list indentation [Expected: 0; Actual: 2]
enhancements/network/dpu/upgrades.md:188:1 MD007/ul-indent Unordered list indentation [Expected: 0; Actual: 2]
...
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_enhancements/1005/pull-ci-openshift-enhancements-master-markdownlint/1499031887035764736#1:build-log.txt%3A19

enhancements/network/dpu/upgrades.md
```diff
@@ -0,0 +1,318 @@
...
+Given that behavior, when an Infra Cluster upgrade occurs, if it is
+necessary to reboot nodes, then each node's tenant should be drained
+before the infra node reboots, so there should be no pod network
+disruption.
+
+### "Drain Mirroring" Controller
+
+In order to prevent MCO from rebooting an Infra node until its Tenant
+has been drained, we will need to run at least one drainable pod on
+each node, so that we can refuse to let that pod be drained until the
```
cgwalters, line 170: Let's clarify here and say "infra node"?  In general throughout the text I think we need to be very clear for type of node.  Perhaps a shorthand "inode" and "tnode"?  Or "infra-node" and "tenant-node"?

enhancements/network/dpu/upgrades.md
danwinship: The choice isn't "drain DaemonSets" or "don't drain DaemonSets". It's "bail out with an error if there are DaemonSets since they can't be drained" or "don't bail out with an error if there are DaemonSets". They never get drained.

enhancements/network/dpu/upgrades.md
```diff
@@ -0,0 +1,318 @@
...
+In [the original proposal], the plan was for infra and tenant cluster
+uprades to be synchronized, with the Infra and Tenant MCOs
+communicating to ensure that workers were upgraded in the same order
+in both clusters, and that reboots of the Infra and Tenant workers
+happened at the same time.
+
+After some discussion, we decided we didn't need that much
+synchronization.
+
+[the original proposal]: https://github.com/openshift/enhancements/commit/8033a923
```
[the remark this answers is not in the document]
danwinship: In my experience, github doesn't ever GC revisions that were commented on as part of a PR.

(Yes, there's that "I can't find those revisions anywhere" error but I think that's more about it trying to stop you from _accidentally_ following out-of-date links.)

enhancements/network/dpu/upgrades.md
```diff
@@ -0,0 +1,318 @@
...
+
+If an Infra Node is being cordoned and drained, then we can assume
+that the administrator (or an operator) is planning to do something
+disruptive to it. Anything that disrupts the Infra Node will disrupt
+its Tenant Node as well.
+
+Therefore, if an Infra Node is being cordoned and drained, we should
+respond by cordoning and draining its Tenant Node, and not allowing
+the Infra Node to be fully drained until the Tenant has also been
+drained.
```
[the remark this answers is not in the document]
danwinship: DPU NIC nodes do not currently run any non-DaemonSet pods, and that's not expected to change. So in fact, the drain-blocker pod is actually the _only_ pod that will get drained.

(I've been imprecise with terminology here... I talk about "infra nodes" and "infra workers", but there are actually three machinesets in the infra cluster: "master", "worker", and "dpu-worker", and it's specifically the "dpu-worker" nodes we're talking about here, and they don't run any general-purpose pods. They only run pods that are specifically part of providing functionality to the corresponding DPU Host Node.)

enhancements/network/dpu/upgrades.md
```diff
@@ -0,0 +1,318 @@
...
+
+Given that behavior, when an Infra Cluster upgrade occurs, if it is
+necessary to reboot nodes, then each node's tenant should be drained
```
pliurh, line 162: Whenever we reboot a DPU, we also have to reboot the host. So that VFs can be recreated and VF representers can be discovered by the DPU. Without the VF reps, the ovnkube-node pod cannot start on the DPU.

enhancements/network/dpu/upgrades.md
```diff
@@ -0,0 +1,318 @@
...
+      - If we were in the process of draining its tenant, then abort.
+
+      - If the tenant node is also `Unschedulable` and has a
+        `Condition` indicating that we are the one who made it
+        `Unschedulable`, then remove the `Condition` and the
+        `Unschedulable` state from the tenant node.
+
+      - If the infra node does not currently have a "drain blocker"
+        pod, then deploy a new one to it.
+
```
pliurh, line 207: As I mentioned above, we need to add a step to reboot the tenant node after the DPU reboot. However, I am not quite sure how DPU can do that. One option is to add a reboot node API to MCO and let DPU invoke that.

enhancements/network/dpu/upgrades.md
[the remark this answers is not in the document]
wking: Instead of relying on GitHub's history preservation, we can just inline the relevant bits in this alternative section.

In this case "we didn't need that" is a bit fuzzy.  I think the relevant bits were something like:

> We realized that all we actually needed was "before I drain the Infra Node, drain its Tenant Node", and that we could achieve that fairly simply.  Adding simultaneous, node-for-node matched simultaneous update was going to be extremely complicated, and didn't bring any additional value.

enhancements/network/dpu/upgrades.md
danwinship: Hm... do we actually _need_ to reboot? It seems to me like there are two possibilities here:

1. Reboot the tenant when we reboot the infra
2. Make the tenant notice when the infra reboots, and properly recreate its VFs and stuff when that happens

Right? We don't have the code to do 2 now, but since we don't have the code to do 1 now either, we have to write code either way. And it seems like making the tenant more able to recover from unexpected infra problems would be a good thing in general?

tsorya:
@danwinship can we reopen enhancement ?

PR #1006, opened 2022-01-12
CSI Inline Ephemeral Volume Security

adambkaplan:
This enhancement proposes a new admission plugin to check that the CSI
driver used to provision a `csi` volume has been vetted and deemed
"safe" for a given PodSecurity profile. Components which install
CSIDriver drivers can optionally add a label to the driver object, which
declares that the driver is safe for a given PodSecurity profile. The
admission checks if a pod references a CSI volume on creation, and if
so, looks up the referenced CSIDriver object and checks the value of the
`security.openshift.io/csi-ephemeral-volume` label. If the driver's
profile is more permissive than the PodSecurity profile of the pod's
namespace, admission is denied.

### DIFF
--- a/enhancements/storage/csi-inline-vol-security.md
+++ b/enhancements/storage/csi-inline-vol-security.md
@@ -209,33 +209,11 @@ restrict use of volumes from these drivers to privileged namespaces.
 
 **TODO - more here? Is this sufficient?**
 
-> API Extensions are CRDs, admission and conversion webhooks, aggregated API servers,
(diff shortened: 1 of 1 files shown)


adambkaplan:
@deads2k @jsafrane this effectively the draft KEP I proposed upstream in https://docs.google.com/document/d/1FSHyYrb0_zWL3d1KZRA_6aXfhF8N-v9D-ggYAidHkwE, with bits filled in for what is asked in OpenShift. New for me are the asks for SLI/SLOs, alerting, and supportability. Do we have such things for admission plugins today?

PR #1010, opened 2022-01-18
authentication: add enhancement on PSa autolabeling

stlaz:
Introduces the controller to synchronize Pod Security admission labels on namespaces.

/assign @s-urbaniak 
/cc @soltysh
/cc @deads2k 

### DIFF
--- a/enhancements/authentication/pod-security-admission-autolabeling.md
+++ b/enhancements/authentication/pod-security-admission-autolabeling.md
@@ -1,0 +1,422 @@
+---
+title: pod-security-admission-autolabeling
+authors:
+  - "@stlaz"
+reviewers:
+  - "@deads2k"
+  - "@s-urbaniak"
+  - "@soltysh"
+approvers:
+  - "@sttts"
+  - "@mfojtik"
+api-approvers: # in case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers)
+  - @sttts"
+creation-date: 2022-02-04
+last-updated: 2022-02-04
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - https://issues.redhat.com/browse/AUTH-133
+see-also:
+  - "/enhancements/authentication/pod-security-admission.md"
+replaces: []
+superseded-by: []
+---
+
+# Pod Security Admission Autolabeling
+
+## Summary
+
+This enhancement expands the "PodSecurity admission in OpenShift". It introduces
+an opt-in mechanism that allows users to to keep their workloads running when
+Pod Security Admission plugin gets turned on.
+
+## Motivation
+
+We want to adhere with the upstream pod security standards for our workloads but
+we also want to provide our users access to the Security Context Constraints
+(hereinafter SCCs) API that they are already used to. However, each of these
+admission plugins works a bit differently and so there must be a middle-man
+that synchronizes the privileges SCCs provide into privileges that Pod Security
+admission (hereinafter PSa) understands.
+
+### Goals
+
+1. OpenShift clusters can run with the restricted Pod Security admission level by default
+without breaking pod controller workloads
+2. Privileged users can opt-in/opt-out a namespace to and from the Pod Security admission
+
+### Non-Goals
+
+1. Make sure that bare pods (pods w/o pod controllers) continue working
+    - Both the admission systems are triggered on pod creation, possibly preventing the
+      API server from persisting the pod on admission denial. It is therefore impossible
+      to take the Pod as an input without touching code of both the admission controllers.
+      Such a code change would also very likely end up being rather unreasonable.
+2. Add the ability to tune the cluster-wide PSa configuration
+
+## Proposal
+
+### User Stories
+
+1. I would like my workloads to be admitted based on the current pod security standards
+   with minimal effort or no effort at all
+
+### Pod Security Admission Label Synchronization
+
+This section explains how pod security admission works and later elaborates on how to use
+the pod security labeling in order to synchronize SCC permissions into pod security admission
+restriction levels.
+
+The SCC permissions to PSa levels transformation gets explained at the end of this section.
+
+#### Pod Security Admission Introduction
+
+Pod Security admission validates pods' security according to the upstream
+[pod security standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/)
+and distinguishes three different security levels:
+- privileged - most privileged mode, anything is allowed
+- baseline   -  minimally restrictive policy which prevents known privilege escalations
+- restricted - heavily restricted policy, following current Pod hardening best practices
+
+By default, there is a cluster-global configuration which enforces the configured policies
+on pods and known workloads.
+
+It is possible to override the cluster-global policy enforce configuration on a per-namespace
+basis by using the `pod-security.kubernetes.io/enforce` label on given namespaces. It is also
+possible to exempt certain users, namespace and runtime classes from the admission completely.
+
+#### PSa Label Synchronization Controller
+
+This section describes a controller that is going to be a part of the cluster-policy-controller.
+
+In order to keep SCCs working alongside restricted cluster-wide PSa level, there must be
+a mechanism that would synchronize the PSa policy level for a namespace that workload runs
+in, so that the workload is allowed by both SCC and PSa admissions. For that purpose, a PSa
+Label Synchronization Controller (hereinafter the Controller) shall be built.
+
+The Controller watches changes of:
+- **namespaces**
+  - manual changes to the interesting labels of the namespaces that were
+    opted-in to label synchronization should be reverted
+- **roles** and **rolebindings** (and their **cluster variants**)
+  - to be able to properly assess SA capabilities in controlled namespaces
+- **SCCs**
+  - to properly assess restrictive PSa levels per affected namespaces
+  - to be able to assess SA capabilities in controlled namespace along with the
+    information retrieved from roles and rolebindings (using the SCC `users` and
+    `groups` fields)
+
+The generic control loop of the Controller performs the following actions:
+1. list all controlled (opted-in by a label) namespaces
+2. for each namespace:
+    1. create SCC <-> []SA associations for all SAs present in the namespace
+    1. order observed SCCs most-privileged first (according to PSa privilege levels
+       they map to)
+    1. evaluate the namespace's PSa privilege level based on the most privileged
+       SCC that is allowed to be used by its SA
+    1. modify the namespace's PSa labels based on the enforce level observed in the
+       previous step
+
+There is another control loop reacting only to changes of namespaces which performs
+all the steps in 2.
+
+#### Matching SCCs to Service Accounts
+
+The Controller needs to be able to map SCCs to a given SA in order to be able to
+later use this information to determine the restrictive level for a given SA.
+
+To map SCCs to SAs, the Controller lists all roles and all SCCs, and locally attempts
+to match the SCCs to the roles based on their `rules`. It does that using the [upstream
+RBAC rules evaluation code](https://github.com/kubernetes/kubernetes/blob/c6153a93d0528a3dc6be9dedc2602f140081688d/plugin/pkg/auth/authorizer/rbac/rbac.go#L168-L193).
+
+The local evaluation was chosen over sending SARs to the API server in order to
+prevent periodic surge of requests in clusters with a larger amount of
+namespaces + roles + SCCs.
+
+For mapping the roles to SAs, an indexing should be introduced on top of the
+rolebinding informer cache. The indexing will index rolebindings to SAs,
+SA groups, and the `system:authenticated` group.
+
+To get SCCs that match a given SA, the process is as follows:
+1. list all SCCs
+2. determine SA access to SCCs based on SCC `users` and `groups` fields
+3. get all roles for an SA and its groups from an indexed informer cache
+4. add SCCs for the roles found in 3. to the list of SCCs found in 2.
+
+#### Label Synchronization Opting-in
+
+The Controller described in [PSa Label Synchronization Controller](#psa-label-synchronization-controller)
+works on an opt-in basis, meaning that users need to specifically sign themselves up
+if they wanted to stick with just SCCs.
+
+Namespaces that specifically want to have their PSa labels synchronized
+should be labelled by the `security.openshift.io/scc.podSecurityLabelSync`
+label with the label value set to `“true”`, on the other hand, to specifically
+request no PSa label synchronization, the namespace should set the label's
+value to "false".
+
+By default, namespaces without the synchronization label will be still
+considered for label synchronization. These namespaces are considered
+"no-opinion" and the label synchronization behavior for these may change
+in any future release.
+
+Having the "no-opinion" path helps keeping the older workloads working
+during upgrades. It also gives us space to decide whether we want the
+label synchronization be done by default or whether the users should
+specifically opt into it in the future.
+
+As a consequence of the opt-in/opt-out being dependent on `Namespace`
+labelling, only the privileged users can determine what kind of
+admission mechanism may run on namespaces, non-privileged users
+depend on the platform defaults.
+
+#### SCC to PSa Level Transformation
+
+The automatic PSa label synchronization requires us to map SCCs directly to a PSa
+level.
+
+In order to be able to map SCCs to PSa level automatically, it is necessary to
+introspect each field of such an SCC.
+
+SCC fields each tell the SCC authorizer how to either:
+- default a value of a given pod's `securityContext` field if unspecified
+- restrict the domain of values for a given pod's `securityContext` field
+
+PSa is based on a number of checks, each of which validates a certain `securityContext`
+field per given PSa restrictiveness level.
+
+As per the above, it is possible to introspect the SCC type fields and based on
+possible values and based on introspecting the PSa check functions for the given
+pod's `securityContext` field that's affected by the SCC field values, it is possible
+to create mappings to different PSa levels.
+
+Given the need for human interaction with PSa check functions when creating the
+SCC-\>PSa mapping, and the complexity of this task, each OpenShift version should
+always ever consider the "latest" PSa checks.
+
+#### Deriving PSa Level from SCC fields
+
+The SCC fields perform different tasks for fields of a security context:
+
+1. restrict the values to a certain value domain
+2. modify the field value in case the field is unset
+3. combination of the above
+4. restrict the values of an already allowed set (e.g. `AllowedFlexVolumes` only
+   further restricts flex volumes allowed by `Volumes` but upstream makes no
+   difference there)
+
+For the conversion between SCC and PSa levels, we are only interested in fields
+that do 1. and 3. The conversion itself needs to make sure that the values allowed
+by SCCs are a subset of the values allowed by the given PSa level in order for the
+SCC to match the PSa level.
+
+**Caveats and Gotchas Found During SCC Mapping Investigation**
+
+1. Currently, there exists no default SCC supplied with the platform that would
+match any PSa level more restricted than `baseline`
+    - the `restricted` (and any dependent) SCC needs to be modified to match the
+      pod security standards of today:
+        - `RequiredDropCapabilities` should now contain just one value - `ALL`
+        - `AllowPrivilegeEscalation` should now be `false`
+        - `AllowedCapabilities` should include `NET_BIND_SERVICE` which allows
+          the application to use domain privileged ports (numbers lower than 1024)
+        - `SeccompProfiles` should equal `runtime/default` which only explicitly
+          defaults this value for pods but it is already implicitly used
+2. The SCC to PSa conversion depends on a namespace for the conversion
+    - the `MustRunAs` strategy of `RunAsUser` allows setting an empty range and
+      its evaluation for a given pod retrieves the range from the pod's namespace
+      annotations
+3. The upstream checks can be updated at any time and there is nothing that would
+   allow a simple machine conversion between SCC and PSa levels, this is all human driven
+4. PSa is designed in a way that allows improving security with time if such a need
+   arises
+    - this could mean that our SCCs might slip to less restrictive PSa levels in
+      time
+    - we will need to have a process to tighten SCC permissions that would match
+      the PSa capabilities in this regard
+
+From the above points, we can safely skip 2. as that is just an interesting observation.
+We are evaluating the PSa levels on namespace level anyway when we are checking the
+SCC-related permissions for each of the SA, which are namespaced by nature.
+
+The above changes required **to make 1. work** might cause workloads to break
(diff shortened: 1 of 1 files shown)


enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+
+### Goals
+
+1. OpenShift clusters can run with the restricted Pod Security admission level by default
+without breaking pod controller workloads
+2. Privileged users can opt-in/opt-out a namespace to and from the Pod Security admission
+
+### Non-Goals
+
+1. Make sure that bare pods (pods w/o pod controllers) continue working
```
s-urbaniak, line 49: can we describe why we exclude them?

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+reviewers:
+  - "@deads2k"
+  - "@s-urbaniak"
+  - "@soltysh"
+approvers:
+  - "@sttts"
+  - "@mfojtik"
+api-approvers: # in case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers)
+  - @sttts"
+creation-date: yyyy-mm-dd
```
soltysh, line 14: Fill those in.

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+last-updated: yyyy-mm-dd
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - TBD
```
soltysh, line 17: Jira link?

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+
+### Non-Goals
+
+1. Make sure that bare pods (pods w/o pod controllers) continue working
+
+## Proposal
+
+### User Stories
+
+1. I would like my workloads to be admitted based on the current pod security standards
```
soltysh, line 55: with minimal or no effort at all

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+
+This section explains how pod security admission works and later elaborates on how to use
+the pod security labeling in order to synchronize SCC permissions into pod security admission
+restriction levels.
+
+The SCC permissions to PSa levels transformation gets explained at the end of this section.
+
+#### Pod Security Admission Introduction
+
+Pod Security admission validates pods' security and distinguishes three different levels:
```
soltysh, line 67: Maybe worth linking https://kubernetes.io/docs/concepts/security/pod-security-standards/

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+basis by using the `pod-security.kubernetes.io/enforce` label on given namespaces. It is also
+possible to exampt certain users, namespace and runtime classes from the admission completely.
+
+#### PSa Label Synchronization Controller
+
+This section describes a controller that is going to be a part of the cluster-policy-controller.
+
+In order to keep SCCs working alongside restricted cluster-wide PSa level, there must be
+a mechanism that would synchronize the PSa policy level for a namespace that workload runs
+in so that the workload is allowed by both SCC and PSa admissions. For that purpose, a PSa
```
soltysh, line 85:
```suggestion
in, so that the workload is allowed by both SCC and PSa admissions. For that purpose, a PSa
```

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+1. list all controlled (opted-in by a label) namespaces
+2. for each namespace:
+    1. get all instances of known workload types
+    1. associate SCCs to the workloads based on the SAs running them, create a
+       SCC <-> []SA associations
+    1. order observed SCCs most-privileged first (according to PSa privilege levels
+       they map to)
+    1. evaluate the namespace's PSa privilege level based on the most privileged
+       SCC that is allowed to be used by its SA
+    1. modify the namespace's PSa labels based on the enforce level obsedved in the
```
soltysh, line 112:
```suggestion
    1. modify the namespace's PSa labels based on the enforce level observed in the
```

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+       previous step
+
+There is another control loop reacting only to changes of namespaces which performs
+all the steps in 2.
+
+#### Label Synchronization Opting-in
+
+The Controller described in [PSa Label Synchronization Controller](#psa-label-synchronization-controller)
```
soltysh, line 120: This sentence doesn't make sense to me, if the PSa controller is opt-in it means you voluntarily sign up for that. So that last part `if they wanted to stick with just SCCs` is not clear to me. If I want to stick with SCC I don't sign up, right?

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+works on an opt-in basis, meaning that users need to specifically sign themselves up
+if they wanted to stick with just SCCs.
+
+The opt-in is driven by the `security.openshift.io/scc.podSecurityLabelSync: “true”`
+label, namespaces containing it will have their PSa labels reconciled by
+the Controller.
+
+The opt-in nature of the Controller would cause current workloads to break.
```
soltysh, line 128: The opt-in wouldn't cause breakage, the opt-out would, or more precisely the controller work would. That's why you are making this opt-in to prevent it.

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+The choice of opting-in is therefore only left upon the users that are
+capable of creating the `Namespace` objects directly. Such a user can clearly
+set the appropriate labels themselves.
+
+The rest of the users rely on `Namespace` creation delegated via the OpenShift
```
soltysh, line 133: This is confusing, first you're talking about the mechanism being opt-in, but eventually this is only opt-in to a small, privileged group of users, since you're changing the default. In that case I'd start with this part where you say we make it a default, but only a set of users can turn it off. Or am I misunderstanding something here 😅

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,290 @@
...
+`Project` API. The default `Project` template should therefore be modified so
+that it by default includes the `security.openshift.io/scc.podSecurityLabelSync: “true”`
+label. This ensures that the decision to use PSa as the main authorization
+mechanism is left to privileged users only and the workloads of the rest
+of the users keeps working.
+
+Simply updating the `Project` template is not enough to keep older workloads
```
soltysh, line 140: This is critical information that you should start with.

soltysh: I'd suggest framing the proposal such that it's clear that this will be done in 2 stages, where:
1. stage 1 (n-1) - introduce this functionality mode in non-breaking mode
2. stage 2 (n) - introduce this in breaking mode
is that accurate or am I misreading something?

soltysh: Additionally, I'm missing answers to the following:
1. I'm a current SCC user how that impacts me?
2. I haven't used SCC in the past and I don't care, does this impacts me?
3. I want to use PS over SCC how do I do that? 
Not sure if I missed something else, but it would be good to describe different options.

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: :+1: this finally made me look into English grammar for comma usage, until now I felt like commas were used somewhat random
https://webapps.towson.edu/ows/modulecomma.htm

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: Hm, you're right, I'll have to reword this section somewhat more understandable.

The truth indeed is that the feature is opt-in only for the privileged users, the rest gets the "old" behavior. And understandably so, we don't want just anyone be choosing which admission to use for their pods, that's a consideration for a more privileged user.

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: If you don't sign yourself up, you're going to be limited by the cluster-wide pod security admission.

The controller adjusts the registered namespaces so that the PSa level matches the permissions of the SCCs required to run workloads of the said namespace. That means that if you don't register, PSa might block your workloads from running even though you have the proper SCC permissions.

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: Without you opting in to the sync, the workloads would break.

The controller runs for your namespace - nothing changes, everything works as it used to.
If it does not - you need to handle the PSa labels yourself.

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: I tried to capture that in a bullet point in this section

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,293 @@
...
+- privileged - most privileged mode, anything is allowed
+- baseline   -  minimally restrictive policy which prevents known privilege escalations
+- restricted - heavily restricted policy, following current Pod hardening best practices
+
+By default, there is a cluster-global configuration which enforces the configured policies
+on pods and known workloads.
+
+It is possible to override the cluster-global policy enforce configuration on a per-namespace
+basis by using the `pod-security.kubernetes.io/enforce` label on given namespaces. It is also
+possible to exampt certain users, namespace and runtime classes from the admission completely.
```
deads2k, line 80: exempt

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,293 @@
...
+  - manual changes to the interesting labels of the namespaces that were
+    opted-in to label synchronization should be reverted
+- all known **workload types** in controlled namespaces
+  - new pods may be getting rejected by PSa for some time before the label
+    synchronization happens
+  - the controller should not allow more relaxed PSa admission level than
+    what SCCs allow
+- **roles** and **rolebindings** (and their **cluster variants**)
+  - to be able to properly assess SA capabilities in controlled namespaces
+- **SCCs**
```
deads2k, line 102: if you're doing this based on SCCs, why do you need the workload types?

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,293 @@
...
+  - to properly assess restrictive PSa levels per affected namespaces
+
+The generic control loop of the Controller performs the following actions:
+1. list all controlled (opted-in by a label) namespaces
+2. for each namespace:
+    1. get all instances of known workload types
+    1. associate SCCs to the workloads based on the SAs running them, create a
```
deads2k, line 109: Could we skip the workload type watches by simply iterating over every SA in a namespace?

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,293 @@
...
+the Controller.
+
+The opt-in nature of the Controller would cause current workloads to break.
+The choice of opting-in is therefore only left upon the users that are
+capable of creating the `Namespace` objects directly. Such a user can clearly
+set the appropriate labels themselves.
+
+The rest of the users rely on `Namespace` creation delegated via the OpenShift
+`Project` API. The default `Project` template should therefore be modified so
+that it by default includes the `security.openshift.io/scc.podSecurityLabelSync: “true”`
```
deads2k, line 138: This mechanism for enforcement is the first one that occurs to me as well, but reading the overall KEP and reflecting on how we change the current state, I think we need a cluster-wide consideration of how to handle "no opinion" so that at some future date we can either switch no-opinion to mean "off" or to have a different install mode that is "direct-kube" that wouldn't have this synchronization.

The `label[security.openshift.io/scc.podSecurityLabelSync]` should still be set to have an explicit opinion, but I think having the project template do it by default will constrain our future choices.  If someone wanted to swap their cluster without breaking current namespaces, they could do something like `oc get ns -o name | oc label security.openshift.io/scc.podSecurityLabelSync:“true”`, right?

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,293 @@
...
+
+#### SCC to PSa Level Transformation
+
+The automatic PSa label synchronization requires us to map SCCs directly to a PSa
+level.
+
+In order to be able to map SCCs to PSa level automatically, it is necessary to
+introspect each field of such an SCC.
+
+SCC fields each tell the SCC authorizer how to either:
```
deads2k, line 159: Have you written this mapping down somewhere?

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,293 @@
...
+The Controller watches changes of:
+- **namespaces**
+  - manual changes to the interesting labels of the namespaces that were
+    opted-in to label synchronization should be reverted
+- all known **workload types** in controlled namespaces
+  - new pods may be getting rejected by PSa for some time before the label
+    synchronization happens
+  - the controller should not allow more relaxed PSa admission level than
+    what SCCs allow
+- **roles** and **rolebindings** (and their **cluster variants**)
```
deads2k, line 100: Do you have PR sketching out how you will manage this?

deads2k:
I like the design.  It's straightforward

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,293 @@
...
+
+Pod Security admission validates pods' security according to the upstream
+[pod security standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/)
+and distinguishes three different security levels:
+- privileged - most privileged mode, anything is allowed
+- baseline   -  minimally restrictive policy which prevents known privilege escalations
+- restricted - heavily restricted policy, following current Pod hardening best practices
+
+By default, there is a cluster-global configuration which enforces the configured policies
+on pods and known workloads.
```
s-urbaniak, line 76: We should explicitely say that despite the possibility to override on a namespace level we do not (officially) allow overriding the cluster-global configuration which is defaulting to `restricted`.

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: My point was to only provide the label for what's currently required by the workloads in each namespace. This is to prevent the labels from going "privileged" in every controlled namespace in case a SA is given cluster-wide cluster-admin access. 

\**thinking out loud*\* On a second thought, the situation where an SA could launch privileged pods anywhere they want is only limited to direct pod creation which I ruled out in the non-goals already. So it is in fact reasonable to to remove the workload types :+1: 

I'll make it so that only the SAs permissions matter for each given NS local to the SA.

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: yes, per https://github.com/openshift/enhancements/pull/1010#discussion_r811869490 I think we should

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: I've got a small PoC which shows some SCC shortcomings that need to be addressed (see FIXMEs in that project): https://github.com/stlaz/sccpsaconversion

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: I don't have a PR for this yet

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: added to non-goals

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: I see. So you suggest a namespace without such a label should still be considered for the sync until we say otherwise? I suppose that it would be easier to migrate it either way (on->off) eventually as labelling the NSes is as simple as you describe.

stlaz: I've rewritten this section to use unlabelled namespaces for PSa label sync, too, clarifying that this behavior may change on unlabelled namespaces in the future.

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: I've rewritten this section a bit, hopefully it explains the situation a bit better.

enhancements/authentication/pod-security-admission-autolabeling.md
deads2k: I think a description of how you'll trigger either resource access review of subjectaccessreview (you should make a case for which path is more efficient) is going to be important for your PR reviewer.

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,304 @@
...
+in, so that the workload is allowed by both SCC and PSa admissions. For that purpose, a PSa
+Label Synchronization Controller (hereinafter the Controller) shall be built.
+
+The Controller watches changes of:
+- **namespaces**
+  - manual changes to the interesting labels of the namespaces that were
+    opted-in to label synchronization should be reverted
+- **roles** and **rolebindings** (and their **cluster variants**)
+  - to be able to properly assess SA capabilities in controlled namespaces
+- **SCCs**
```
deads2k, line 102: will users and groups in SCCs be honored or will you only honor RBAC based mappings?

enhancements/authentication/pod-security-admission-autolabeling.md
deads2k: I think describing this mapping here is worthwhile.

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,304 @@
...
+
+A new `Namespace` label is being introduced - `security.openshift.io/scc.podSecurityLabelSync`.
+When set to value `"true"`, the such-labeled `Namespace` gets picked by the
+PSa label synchronization controller and this `Namespace` and its workloads
+will be reconciled according to the [PSa Label Synchronization Controller](#psa-label-synchronization-controller)
+section.
+
+### Risks and Mitigations
+
+Turning on the PSa to restricted level enforcing even with the namespace label
```
deads2k, line 180: only one that doesn't rely on SA permissions.  Builds for instance will need a cluster-wide punch through in the apiserver config.

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,304 @@
...
+It is also likely that pod controllers might happen to fail pod creation before
+the labels are properly synchronized on their namespace. There is no way to
+mitigate that, the pods will eventually succeed to be created if the given
+service account running them has the correct permissions.
+
+## Design Details
+
+### Open Questions
+
+1. Should we have a way to completely disable SCC admission per namespace
```
deads2k, line 198: I think we can worry about this at a later date.

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,304 @@
...
+   to allow only PSa?
+
+### Test Plan
+
+The tests shall include these scenarios:
```
deads2k, line 203: make sure you include checking the build tests for each build strategy.

stlaz:
/remove-lifecycle stale
I was coding on the agreed parts of this enhancement in the meantime

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: I'm going to open a follow-up PR for that

enhancements/authentication/pod-security-admission-autolabeling.md
stlaz: users and groups in SCCs need to be honored, too

stlaz:
@deads2k @s-urbaniak I added a paragraph on how the controller evaluates roles in order to determine access of SAs to SCCs.

deads2k:
> I'm going to open a follow-up PR for that

This is critical for determining if the plan laid out here will work.  As I recall, while determining this, you hit a snag.

stlaz:
> This is critical for determining if the plan laid out here will work. As I recall, while determining this, you hit a snag.

A couple. The seccomp profiles are already fixed (https://github.com/openshift/kubernetes/pull/1218), and it seems that we've been able to drop "ALL" capabilities for some time, only we never enforced that.

The next issue is that with our default SCCs, there is not a single one that would map to restricted, we are "baseline" at best. We'll want to address that. I can include the SCC mapping and its caveats in this PR, then.

stlaz:
Added more information on the SCC->PSa level mapping, the new sections also debate caveats and points of interest of the method used.

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,420 @@
...
+From the above points, we can safely skip 2. as that is just an interesting observation.
+We are evaluating the PSa levels on namespace level anyway when we are checking the
+SCC-related permissions for each of the SA, which are namespaced by nature.
+
+The above changes required **to make 1. work** might cause workloads to break
+if applied directly to the `restricted` SCC. It is impossible to tell which
+workload would break by this step because most of the platform users very likely just
+went with the platform defaults and did not perform further steps in order to
+tighten their workloads to their actual needs. To solve this issue, new
+`*-v2` (e.g. `restricted-v2`, `nonroot-v2`) SCCs with tighter security policies
```
deads2k, line 244: My first blush opinion is that this is ok.

enhancements/authentication/pod-security-admission-autolabeling.md
```diff
@@ -0,0 +1,420 @@
...
+should be introduced for the `restricted` SCC and SCCs derived from it. The `restricted`
+SCC will also drop `system:authenticated` from its `Groups` field in favour of adding
```
deads2k, line 246: only for new installs, correct?

stlaz: Yes, we should not backport this change. Since the current SCCs are create-only, changing this in the CKAO manifests won't perform any changes to the current SCC.

PR #1011, opened 2022-01-18
Ignore new markdownlint rules that fail on existing documents

omertuc:
The lint job on PRs began failing recently on existing documents.

This is because a few new rules have been added to markdownlint: https://github.com/DavidAnson/markdownlint/pull/448

I wasn't sure whether I should be updating the existing documents to
conform to the new rules or just ignore those new rules altogether,
so I went with the latter.

### DIFF
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -48,3 +48,9 @@ config:
   # Trailing spaces don't cause any known issues and are often introduced when 
   # updating text to conform to max line length restrictions.
   no-trailing-spaces: false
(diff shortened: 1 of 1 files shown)


sdodson:
But it only fails on files involved in the PR correct? I don't see the harm in new rules being applied to files as they're modified.

dhellmann:
The markdown lint rules are applied to all files. The template rules are applied to changed files.

dhellmann:
/lgtm
/approve

Thanks, @omertuc!

PR #1017, opened 2022-01-25
helm: fix example in the documentation

zonggen:
Fixes CR example in the Helm enhancement doc, and adds a section with
desired RBAC setup before applying CR.

Signed-off-by: Allen Bai <abai@redhat.com>

### DIFF
--- a/enhancements/helm3/namespace-scoped-helm-repository-crd.md
+++ b/enhancements/helm3/namespace-scoped-helm-repository-crd.md
@@ -64,14 +64,36 @@ kind: ProjectHelmChartRepository
 metadata:
   name: my-enterprise-chart-repo
 spec:
-  url: https://my.chart-repo.org/stable
-
-  # optional and only needed for UI purposes
-  displayName: myChartRepo
-
-  # optional and only needed for UI purposes
-  description: my private chart repo
+  # Optional: human readable repository name, it is used by UI for displaying purposes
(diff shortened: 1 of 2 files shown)


enhancements/helm3/namespace-scoped-helm-repository-crd.md
```diff
@@ -64,14 +64,34 @@ kind: ProjectHelmChartRepository
 metadata:
   name: my-enterprise-chart-repo
 spec:
-  url: https://my.chart-repo.org/stable
-
-  # optional and only needed for UI purposes
-  displayName: myChartRepo
```
dperaza4dustbit, line 70: Instead of deleting can we rename to "name". The meaning is the same: Friendly name to display in the UI

zonggen: Updated.

dhellmann:
You don't need to add the sections from the new template, we can just override the job. The duplicate header problem is real though, so if you fix that I can help with the override.

zonggen:
@dhellmann Thanks for offering help! I've removed the empty sections.

dhellmann:
/override markdownlint

This is an update to an existing file and so the new template sections do not apply.

dperaza4dustbit (commented):
/lgtm

PR #1018, opened 2022-01-25
Add Windows Health Management containerd context

saifshaikh48:
This PR updates the Windows health management enhancement proposal, reconciling it with the concurrent shift from Docker container runtime to containerd. With the switch, we would no longer able to assume the container runtime is already present, running, and configured on Windows instances.

### DIFF
--- a/enhancements/windows-containers/health-management.md
+++ b/enhancements/windows-containers/health-management.md
@@ -25,9 +25,9 @@ status: implementable
 
 ## Summary
 
-The purpose of this enhancement is to ensure that all Windows nodes managed by the Windows Machine Config Operator (WMCO)
-maintain an expected state. With health management implemented, WMCO will automatically ensure that OpenShift related
-services running on the underlying Windows instances are configured, and have a state inline with the
+The purpose of this enhancement is to ensure that all Windows nodes managed by the Windows Machine Config Operator 
+(WMCO) maintain an expected state. With health management implemented, WMCO will automatically ensure that OpenShift 
+related services running on the underlying Windows instances are configured, and have a state inline with the
 expectations of the installed WMCO version.
 
 ## Motivation
@@ -43,7 +43,8 @@ and the node upgrade path from any WMCO version to another is codified through C
 ### Goals
 
 * Maintain the `Ready` state of Windows Nodes to ensure Windows workloads can be run without interruption.
-* Ensuring WMCO-managed Windows services on the instances associated with Windows Node objects are in the expected state and configured with the expected values:
+* Ensuring WMCO-managed Windows services on the instances associated with Windows Node objects are in the expected state
+  and configured with the expected values:
   + Kubernetes components
   + Node-level networking components
   + Container runtime
@@ -53,19 +54,21 @@ and the node upgrade path from any WMCO version to another is codified through C
 ### Non-Goals
 
 * Diagnosing or remediating generic Windows issues, unrelated to changes made by WMCO.
-* Cluster-level resource management, such as deleting or re-creating WMCO-configured Machines that enter an unrecoverable state (this is a responsibility of [MachineHealthChecks](https://docs.openshift.com/container-platform/latest/machine_management/deploying-machine-health-checks.html)).
+* Cluster-level resource management, such as deleting or re-creating WMCO-configured Machines that enter an 
+  unrecoverable state (this is a responsibility of [MachineHealthChecks](https://docs.openshift.com/container-platform/latest/machine_management/deploying-machine-health-checks.html)).
 
 
 ## Proposal
 
-Currently, Windows Node configuration process is split between two components, WMCO and [WMCB](https://github.com/openshift/windows-machine-config-bootstrapper/).
-WMCB is a program which performs a one-shot configuration of a Windows instance, to ensure that it can become a worker
-Node.
+Currently, Windows Node configuration process is split between two components, WMCO and 
+[WMCB](https://github.com/openshift/windows-machine-config-bootstrapper/). WMCB is a program which performs a one-shot 
+configuration of a Windows instance, to ensure that it can become a worker Node.
 
 To accomplish the goals listed in this enhancement, I am proposing that WMCB be converted into a daemon,
 Windows Instance Config Daemon (WICD). WICD will have the following responsibilities:
 * WICD maintains current responsibilities of bootstrapping kubelet through the `bootstrap` command.
-* Using the configuration provided by a ConfigMap created by WMCO, WICD maintains the state of Windows Services on the instance.
+* Using the configuration provided by a ConfigMap created by WMCO, WICD maintains the state of Windows Services on the 
+  instance.
 * WICD reverts all changes made to an instance when run with the `cleanup` command, and also deletes the Node
   object.
 
--- a/enhancements/windows-containers/health-management.md
+++ b/enhancements/windows-containers/health-management.md
@@ -28,7 +28,8 @@ status: implementable
 The purpose of this enhancement is to ensure that all Windows nodes managed by the Windows Machine Config Operator 
 (WMCO) maintain an expected state. With health management implemented, WMCO will automatically ensure that OpenShift 
 related services running on the underlying Windows instances are configured, and have a state inline with the
-expectations of the installed WMCO version.
+expectations of the installed WMCO version. As the shift to [containerd](container-runtime-containerd.md) is imminent,
+this enhancement is written considering containerd as the only supported container runtime.
 
 ## Motivation
 
@@ -66,7 +67,10 @@ configuration of a Windows instance, to ensure that it can become a worker Node.
 
 To accomplish the goals listed in this enhancement, I am proposing that WMCB be converted into a daemon,
 Windows Instance Config Daemon (WICD). WICD will have the following responsibilities:
-* WICD maintains current responsibilities of bootstrapping kubelet through the `bootstrap` command.
+* As part of the node `bootstrap` phase, WICD will configure the containerd runtime and start it as a Windows service
+  based on its definition in the services ConfigMap.
+* WICD will also maintain current responsibilities of starting kubelet as a service through the `bootstrap` command. 
+  WICD will fully configure kubelet in one shot, as everything required will be available in the services ConfigMap.
 * Using the configuration provided by a ConfigMap created by WMCO, WICD maintains the state of Windows Services on the 
   instance.
 * WICD reverts all changes made to an instance when run with the `cleanup` command, and also deletes the Node
@@ -74,10 +78,16 @@ Windows Instance Config Daemon (WICD). WICD will have the following responsibili
 
 To enable this, WMCO will have these changes in responsibility:
 * WMCO will no longer configure the Windows services on Windows instances, except for WICD.
+* WMCO will take over the parsing of the worker ignition file. This is because, as the maintainer of the services
+  ConfigMap, WMCO provide the expected state of the kubelet service to WICD. This requires information found within the
+  ignition file. Note that WMCO will be able to provide all the information needed to fully configure kubelet right away
+  since [CNI configuration is no longer provided to kubelet](https://github.com/kubernetes/kubernetes/pull/106907).
+* WMCO will copy files obtained from parsing the worker ignition onto Windows instances as part of the payload.
+  These files include the bootstrap kubeconfig and the client CA certificate needed to authenticate with the API server.
 * WMCO will create and maintain a ConfigMap which provides WICD with the specifications for each Windows service that
   must be created on a Windows instance.
-* WMCO will invoke the `bootstrap` command of WICD when initially adding a Windows instance to the cluster. This will
-  take the worker ignition file as an input, and will not involve the services ConfigMap.
+* WMCO will invoke the `bootstrap` command of WICD when initially adding a Windows instance to the cluster. 
+  This will involve the services ConfigMap in order to start the containerd runtime and kubelet services.
 * WMCO will invoke the `controller` command to run WICD as a Windows service on the instance.
 * WMCO will invoke the `cleanup` command of WICD when removing an instance from the cluster. WMCO will no longer
   delete the Node object.
@@ -112,6 +122,9 @@ N/A
   To reduce the risk of this, the `MaxConcurrentReconciles` option can be provided when initalizing the ConfigMap
   controller. This will enable the controller to process service ConfigMap events that occur while WMCO is handling
   its BYOH configuration responsibilities.
+* WMCO will not default to containerd as the container runtime until the 6.0.0 timeframe. Therefore, we will be using a 
+  runtime flag `dockerRuntime=false` introduced by the containerd enhancement to develop during the 5.y.z cycle.
+  This flag will be dropped in WMCO 6.0.0.
 
 ## Design Details
 
@@ -129,7 +142,6 @@ If an entity attempts to create a services ConfigMap with incorrect values, WMCO
 This will resolve a scenerio in which the ConfigMap controller is busy configuring Windows instances, and a user
 deletes and recreates the service ConfigMap with incorrect values, before the controller has a chance to re-create
 the ConfigMap.
-
 
 The services ConfigMap contains two keys: `services`, and `files`. The values for both keys are JSON objects.
 The `services` key contains all data required to configure the required Windows services on any instance that is to be
@@ -150,7 +162,8 @@ added to the cluster as a Node. The proposed schema is as follows:
       "name": "string within command field that will be substituted",
       "path": "location of the PowerShell script to be run, in order to get its output"
     },
-    "priority": "integer that will be used to order the creation of the services, priority 1 is created first"
+    "bootstrap": "boolean flag indicating whether this service should be handled as part of node bootstrapping",
+    "priority": "non-negative integer that will be used to order the creation of the services, priority 0 is created first"
   }
 ]
 ```
@@ -168,6 +181,11 @@ When making use of a PowerShell script, retries should be present within the scr
 is waiting for the HNS network to be created in order to create an HNS endpoint. In that specific case, the PowerShell
 script which creates the endpoint should include a retry period which waits for the needed Network to exist.
 
+Services marked with `bootstrap: true` will start before any others. Bootstrap services cannot depend on a non-bootstrap
+service; WICD will throw an error if this is detected. Similarly, each service that has the bootstrap flag set as true
+must have a higher priority than all non-bootstrap services. WICD will throw an error if this is not the case.
+There should be no overlap in the priorities of bootstrap services and controller services.
+
 For example, if the service ConfigMap had an entry with the following data:
 ```json
 {
@@ -182,6 +200,7 @@ For example, if the service ConfigMap had an entry with the following data:
     "path": "C:\k\scripts\get_net_ip.ps"
   },
   "dependencies": [],
+  "bootstrap": false,
   "priority": 2
 }
 ```
@@ -190,7 +209,8 @@ WICD would know to create the service named `new-service` with the value of the 
 name of the instance's Node, and the output of the PowerShell script located at `C:\k\scripts\get_net_ip.ps`, as the
 value of the argument --variable-arg2. WICD will be aware of the proper Node object, as it searching for a Node with an
 internal IP equivalent to the IP of the instance.
-This service has no dependencies, and should be created only after all services with priority 1 have been created.
+This service is not marked as a bootstrap service, so it will be started by WICD when run with the `controller` command.
+It should be created only after all bootstrap services and those with priorites 0 and 1 have been created.
 
 The `files` key contains the path and checksum of files copied to the instance by WMCO. This will be used by WICD to
 validate that essential files have not been tampered with.
@@ -224,8 +244,22 @@ For Node configuration WICD will have two separate responsibilities:
 
 #### Bootstrap command
 
-When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created
-for the instance. This functionality is currently implemented within WMCB with the `init-kubelet` command.
+When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created for the instance. 
+The `bootstrap` command will start all services that have a `bootstrap` value of true, and exclusively these services.
+
+The `bootstrap` command will have the responsibility of starting the containerd service. This is becuase the kubelet
+service has a dependency on the container runtime; the container runtime must have a reachable endpoint
+(i.e. `\\.\pipe\containerd-containerd` is open and usable) when kubelet is initialized (with the `container-runtime`
+and `container-runtime-endpoint` parameters supplied). The containerd service will be configured based on the
+information in the services ConfigMap, a priority 0 service with no dependencies that points to the location of
+CNI plug-ins on the instance. Although CNI config is populated later in the `bootstrap` phase after hybrid-overlay runs,
+pointing the containerd config to this location is enough to ensure that the service picks up networking config changes
+without erroring out on start-up or requiring a restart.
+
+The rest of the functionality is reading the kubelet configuration from the service ConfigMap and starting the kubelet 
+service. A key note is that, since networking configuration is now the container runtime's responsibility, kubelet can 
+be fully configured upon service creation, no longer requiring a separate intitialization stage then a later restart.
+
 WICD will wait for the Node to be created, and cordon it, as the Node is not ready for workloads at this point.
 
 #### Running as a Windows service


enhancements/windows-containers/health-management.md
```diff
@@ -221,8 +222,12 @@ For Node configuration WICD will have two separate responsibilities:
 
 #### Bootstrap command
 
-When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created
-for the instance. This functionality is currently implemented within WMCB with the `init-kubelet` command.
+When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created for the instance. 
+When containerd is the container runtime, the `bootstrap` commnad will have the added responsibility of starting 
+the containerd service. This is becuase the kubelet service has a dependency on the container runtime; the container 
+runtime must have a reachable endpoint when kubelet is initialized. The containerd service will be configured based on
```
aravindhp, line 228: It is unclear what endpoint you are referring to here. Please elaborate.

enhancements/windows-containers/health-management.md
```diff
@@ -221,8 +222,12 @@ For Node configuration WICD will have two separate responsibilities:
 
 #### Bootstrap command
 
-When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created
-for the instance. This functionality is currently implemented within WMCB with the `init-kubelet` command.
+When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created for the instance. 
+When containerd is the container runtime, the `bootstrap` commnad will have the added responsibility of starting 
+the containerd service. This is becuase the kubelet service has a dependency on the container runtime; the container 
+runtime must have a reachable endpoint when kubelet is initialized. The containerd service will be configured based on 
+the information in the services ConfigMap, pointing to CNI plug-ins' (binaries and config files) locations on the instnace.
```
aravindhp, line 229: I thought the CNI config gets populated after hybrid-overlay runs. How are you handling that?

aravindhp (changes requested):
Thanks for the update, @saifshaikh48. Please address my comments.

enhancements/windows-containers/health-management.md
saifshaikh48: Correct, but if we point containerd to the _eventual_ location of CNI config, the service will pick it up once created without 
1. erroring out on service start up
2. having to restart 
Unless I am mistaken, this is the behavior @selansen saw in POC work.

aravindhp: Please call that out here.

aravindhp (approved):
/approve

aravindhp:
/hold

@selansen please remove then hold when you review and lgtm

saifshaikh48:
Would also like to wait for @sebsoto  to look over it before it merges

selansen:
@saifshaikh48  if I am not mistaken, only the last two diffs are related to cotainerd change and the rest are "formatting" changes. Am I correct?

enhancements/windows-containers/health-management.md
```diff
@@ -221,8 +227,17 @@ For Node configuration WICD will have two separate responsibilities:
 
 #### Bootstrap command
 
-When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created
-for the instance. This functionality is currently implemented within WMCB with the `init-kubelet` command.
+When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created for the instance. 
+When containerd is the container runtime, the `bootstrap` command will have the added responsibility of starting
```
selansen, line 231: By the time we implement WHM, the containerd always going to be the runtime.  Do we still need to call this out? This brings other questions like how do we handle other runtimes? Should we mention containerd will be the only supported runtime by WHM?

saifshaikh48: Since we will continue using docker through 5.x and only switching to containerd officially in 6.0.0, I think it is better to call out responsibilities in both cases. I don't want to restrict the scope of the enhancement based on an arbitrary timeline. Assume we finish health management work quickly and release it as part of 5.1 -- then the enhancement would be out of sync and we would have to come back again to update it.

saifshaikh48:
> @saifshaikh48 if I am not mistaken, only the last two diffs are related to cotainerd change and the rest are "formatting" changes. Am I correct?

@selansen there are 4 content related diffs, I think looking commit by commit can be clearer

enhancements/windows-containers/health-management.md
```diff
@@ -77,7 +79,8 @@ To enable this, WMCO will have these changes in responsibility:
 * WMCO will create and maintain a ConfigMap which provides WICD with the specifications for each Windows service that
   must be created on a Windows instance.
 * WMCO will invoke the `bootstrap` command of WICD when initially adding a Windows instance to the cluster. This will
-  take the worker ignition file as an input, and will not involve the services ConfigMap.
+  take the worker ignition file as an input, and will only involve the services ConfigMap if containerd is the container
+  runtime in order to use its defined service configuration.
```
sebsoto, line 83: Lets make sure we think through how the `bootstrap` command is going to recognize containerd from the services ConfigMap.
Some options:
* By name?
* Runtime is recognized by priority zero? Is this overloading the priority field?
* Should we have a separate runtime field in the service object within the services array?
* Do we add a separate `runtime` key which contains a service object on the same level as `services` and `files`? We could also add the endpoint within this. So something like:
```
{
  files: []
  services; []
  runtime: {
    service: {containerd service spec}
    endpont: {\\.\pipe\containerd-containerd}
  }
}
```

I am leaning towards the last option, as it is only possible for there to be one runtime. What do you think?

saifshaikh48: After discussing, we decided that containerd runtime and its config will be represented as just another service in the services array (no new key), and this service will have the highest priority.

mansikulkarni96:
LGTM

enhancements/windows-containers/health-management.md
aravindhp: How will WICD if containerd is the container runtime? Won't it need to look at the ConfigMap to figure that out? IOW, the `bootstrap` command will need to involve / look at the ConfigMap always.

enhancements/windows-containers/health-management.md
```diff
@@ -224,8 +227,20 @@ For Node configuration WICD will have two separate responsibilities:
...
+
+When containerd is the container runtime (rather than pre-installed Docker), the `bootstrap` command will have the added 
+responsibility of starting the containerd service. This is becuase the kubelet service has a dependency on the container 
+runtime; the container runtime must have a reachable endpoint (i.e. `\\.\pipe\containerd-containerd` is open and usable) 
+when kubelet is initialized (with the `container-runtime` and `container-runtime-endpoint` parameters supplied). 
+The containerd service will be configured based on the information in the services ConfigMap, a priority 0 service with
+no dependencies that points to the eventual location of CNI plug-ins on the instance.
+Although CNI config is populated later in the `bootstrap` phase after hybrid-overlay runs, 
+pointing the containerd config to this location is enough to ensure that the service picks up 
+networking config changes without erroring out on start-up or requiring a restart.
```
aravindhp, line 241: Nit: The paragraph formatting looks like a bit strange here.

enhancements/windows-containers/health-management.md
```diff
@@ -224,8 +227,20 @@ For Node configuration WICD will have two separate responsibilities:
 
 #### Bootstrap command
 
-When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created
-for the instance. This functionality is currently implemented within WMCB with the `init-kubelet` command.
+When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created for the instance. 
+The `bootstrap` command will start all services that have a `priority` field value of 0, and exclusively these services.
```
selansen, line 231: Does it mean even if we want bootstrap to start more than 1 service, all the services should have the same priority?
Can we have a range instead of a fixed number to be flexible with our design to accommodate future needs?

aravindhp: In addition if the bootstrap services have an ordering requirement like containerd and kubelet, how is that being expressed?

enhancements/windows-containers/health-management.md
```diff
@@ -66,18 +66,24 @@ configuration of a Windows instance, to ensure that it can become a worker Node.
...
 * Using the configuration provided by a ConfigMap created by WMCO, WICD maintains the state of Windows Services on the 
   instance.
 * WICD reverts all changes made to an instance when run with the `cleanup` command, and also deletes the Node
   object.
 
 To enable this, WMCO will have these changes in responsibility:
 * WMCO will no longer configure the Windows services on Windows instances, except for WICD.
+* WMCO will take over the parsing of the worker ignition file. Note that WMCO will now be able to provide all the 
+  information needed to fully configure the kubelet service upon its creation as 
+  [CNI configuration is no longer provided to kubelet](https://github.com/kubernetes/kubernetes/pull/106907#).
```
sebsoto, line 82: Lets make sure we mention that WMCO will copy over the files contained within the ignition file, such as the bootstrap kubeconfig and the kubelet-ca.crt https://github.com/openshift/windows-machine-config-bootstrapper/blob/b9612c9cdf99acb1184810817b56b4b1a9ef8b95/pkg/bootstrapper/bootstrapper.go#L467-L472

sebsoto (changes requested):

enhancements/windows-containers/health-management.md
saifshaikh48: Good questions, we tried to address them with [this](https://github.com/openshift/enhancements/pull/1018/commits/8995d1bdd14f0f5f2fd9db616bef8a49136ac905#diff-3c7c637b883f288bf20ea8e8e521c4962adef0c81b2d149955787c8baf87273fR179-R181) and [this](https://github.com/openshift/enhancements/pull/1018/commits/8995d1bdd14f0f5f2fd9db616bef8a49136ac905#diff-3c7c637b883f288bf20ea8e8e521c4962adef0c81b2d149955787c8baf87273fR242).

enhancements/windows-containers/health-management.md
saifshaikh48: Sure, added directly below this line.

enhancements/windows-containers/health-management.md
```diff
@@ -168,6 +176,10 @@ When making use of a PowerShell script, retries should be present within the scr
 is waiting for the HNS network to be created in order to create an HNS endpoint. In that specific case, the PowerShell
 script which creates the endpoint should include a retry period which waits for the needed Network to exist.
 
+Each service that has the bootstrap flag set as true must have a higher priority than all non-bootstrap services.
```
sebsoto, line 179: nit: While this probably should be the case, this doesn't necessarily need to be true

sebsoto (commented):
LGTM other than nit

enhancements/windows-containers/health-management.md
selansen: > Good questions, we tried to address them with [this](https://github.com/openshift/enhancements/pull/1018/commits/8995d1bdd14f0f5f2fd9db616bef8a49136ac905#diff-3c7c637b883f288bf20ea8e8e521c4962adef0c81b2d149955787c8baf87273fR179-R181) and [this](https://github.com/openshift/enhancements/pull/1018/commits/8995d1bdd14f0f5f2fd9db616bef8a49136ac905#diff-3c7c637b883f288bf20ea8e8e521c4962adef0c81b2d149955787c8baf87273fR242).

The link takes me to the expired page.
Are you referring to the below statement ?
`Also, when there are multiple services that share the same priority, the tied services' dependencies will be examined to
help determine order of creation.`

enhancements/windows-containers/health-management.md
```diff
@@ -66,18 +66,26 @@ configuration of a Windows instance, to ensure that it can become a worker Node.
 
 To accomplish the goals listed in this enhancement, I am proposing that WMCB be converted into a daemon,
 Windows Instance Config Daemon (WICD). WICD will have the following responsibilities:
-* WICD maintains current responsibilities of bootstrapping kubelet through the `bootstrap` command.
+* As part of the `bootstrap` phase, WICD will configure the [containerd](container-runtime-containerd.md) runtime and 
+  start it as a Windows service, if its configuration is detailed in the services ConfigMap.
+* WICD will also maintain current responsibilities of starting kubelet as a service through the `bootstrap` command. 
+  WICD will fully configure kubelet in one shot, as everything required will be available in the services ConfigMap.
```
aravindhp, line 72: How is this possible when using Docker? Don't you need to configure hybrid-overlay and the reconfigure kubelet for CNI?

enhancements/windows-containers/health-management.md
```diff
@@ -66,18 +66,26 @@ configuration of a Windows instance, to ensure that it can become a worker Node.
...
 * Using the configuration provided by a ConfigMap created by WMCO, WICD maintains the state of Windows Services on the 
   instance.
 * WICD reverts all changes made to an instance when run with the `cleanup` command, and also deletes the Node
   object.
 
 To enable this, WMCO will have these changes in responsibility:
 * WMCO will no longer configure the Windows services on Windows instances, except for WICD.
+* WMCO will take over the parsing of the worker ignition file. Note that WMCO will now be able to provide all the
```
aravindhp, line 80: While this might be needed I don't understand the reasoning behind doing so. So more explanation is needed regarding the why. All that is mentioned here is the how.

enhancements/windows-containers/health-management.md
```diff
@@ -168,6 +176,11 @@ When making use of a PowerShell script, retries should be present within the scr
 is waiting for the HNS network to be created in order to create an HNS endpoint. In that specific case, the PowerShell
 script which creates the endpoint should include a retry period which waits for the needed Network to exist.
 
+For ease of processing, services with the bootstrap flag set to true should have a higher priorities than all 
+non-bootstrap services.
```
aravindhp, line 180: Please add the WICD will throw an error if this is not the case.

enhancements/windows-containers/health-management.md
```diff
@@ -168,6 +176,11 @@ When making use of a PowerShell script, retries should be present within the scr
 is waiting for the HNS network to be created in order to create an HNS endpoint. In that specific case, the PowerShell
 script which creates the endpoint should include a retry period which waits for the needed Network to exist.
 
+For ease of processing, services with the bootstrap flag set to true should have a higher priorities than all 
+non-bootstrap services. 
+Also, when there are multiple services that share the same priority, the tied services' dependencies will be examined to
+help determine order of creation.
```
aravindhp, line 182: While this is fine, I feel this is a unnecessary complication. You could divide the services into two sections, bootstrap and post-bootstrap (or something named to that effect) and have individual priorities within them.

enhancements/windows-containers/health-management.md
```diff
@@ -224,8 +239,22 @@ For Node configuration WICD will have two separate responsibilities:
...
-for the instance. This functionality is currently implemented within WMCB with the `init-kubelet` command.
+When run with the `bootstrap` command, WICD will do the steps required to ensure that Node is created for the instance. 
+The `bootstrap` command will start all services that have a `bootstrap` value of true, and exclusively these services.
+
+When containerd is the container runtime (rather than pre-installed Docker), the `bootstrap` command will have the added
+responsibility of starting the containerd service. This is becuase the kubelet service has a dependency on the container
+runtime; the container runtime must have a reachable endpoint (i.e. `\\.\pipe\containerd-containerd` is open and usable)
+when kubelet is initialized (with the `container-runtime` and `container-runtime-endpoint` parameters supplied). 
+The containerd service will be configured based on the information in the services ConfigMap, a priority 0 service with
+no dependencies that points to the eventual location of CNI plug-ins on the instance. Although CNI config is populated
```
aravindhp, line 250: Why is this the eventual location? Won't it be the location from the get go? It is just that it will be used later by containerd but that should be irrelevant to WICD.

enhancements/windows-containers/health-management.md
saifshaikh48: Whoops. Yes that statement, as well as this: 
`The bootstrap command will start all services that have a bootstrap value of true, and exclusively these services.`

enhancements/windows-containers/health-management.md
saifshaikh48: The idea behind this is that kubelet no longer needs to be restarted with its initial config being overwritten with new CNI options. Since WMCO transfers (and deletes when deconfiguring) all other files onto the node, it should extract the needed info from ignition and handle the transport onto instances. We thought WICD should not be responsible for transferring anything -- it should just manage the state of already present binaries and files.

aravindhp: >We thought WICD should not be responsible for transferring anything

WICD is on the instance. How is it ever going to transfer anything?

sebsoto: The reason that needs to be captured here is that WMCO is the maintainer of the services ConfigMap. WMCO must provide the expected state of the kubelet service, and must parse the ignition file in order to do so.

enhancements/windows-containers/health-management.md
sebsoto: The `tie breaker` logic for multiple services with the same priorities isn't needed. If there's a tie its fine to just go down the list and do them in any arbitrary order. In practice containerd can be priority 0 and kublet priority 1.

sebsoto: To be clear, I'd prefer to drop this:
```
the tied services' dependencies will be examined to
help determine order of creation.
```
Instead of making two service lists

enhancements/windows-containers/health-management.md
sebsoto: Missed this earlier, I don't think this will have a noticeable effect on ease of processing. Its an unnecessary restriction that id prefer to drop.

enhancements/windows-containers/health-management.md
saifshaikh48: CNI config actually gets populated after hybrid-overlay runs, hence the "eventual" wording

aravindhp: Why would the location change after hybrid-overlay runs?

saifshaikh48: It won't. Will remove eventual

enhancements/windows-containers/health-management.md
saifshaikh48: Added `As the shift to containerd is imminent, this enhancement is written considering containerd as the only supported container runtime.` to the summary, as discussed by the team.

enhancements/windows-containers/health-management.md
```diff
@@ -150,7 +159,8 @@ added to the cluster as a Node. The proposed schema is as follows:
       "name": "string within command field that will be substituted",
       "path": "location of the PowerShell script to be run, in order to get its output"
     },
-    "priority": "integer that will be used to order the creation of the services, priority 1 is created first"
+    "bootstrap": "boolean flag indicating whether this service should be handled as part of node bootstrapping",
+    "priority": "non-negative integer that will be used to order the creation of the services, priority 0 is created first"
```
aravindhp, line 163: What if I have `priority=2, bootstrap=true` and `priority=1,boostrap=false`? Basically means you need to say more about the interplay of these two options.

enhancements/windows-containers/health-management.md
aravindhp: You also need to mention somewhere that you will be using the Docker runtime flag introduced by the containerd enhancement to develop during the 5.y.z cycle and this will be dropped in the 6.0.0 timeframe.

aravindhp (changes requested):

enhancements/windows-containers/health-management.md
saifshaikh48: Good point. @sebsoto I think we need to add the restriction back, otherwise I'm not sure how we'd handle this case.
`Each service that has the bootstrap flag set as true must have a higher priority than all non-bootstrap services. WICD will throw an error if this is not the case.`

aravindhp: Or you can sort them separately and add a restriction that only a controller service can depend on a bootstrap service but not the other way around.

saifshaikh48: We will be creating `bootstrap` services before `controller` ones. So using the example above, if we have Service A `priority=2, bootstrap=true` and Service B `priority=1,boostrap=false`, we are basically saying the priority field can be meaningless when compared across WICD stages (since Service A will be started first, despite being lower priority).
This is fine imo but it's just not super clear. If I'm understanding it right, I can add an explanation to the enhancement?

sebsoto: > the priority field can be meaningless

Thats a bit of an overstatement. The priority field allows us to find order in services. Bootstrapping is a special case, that will only happen once for a node, while the `controller` command, will be reconciling through the node's entire lifespan.

For example lets say you have 3 services with priorities {4,0,1}
Lets say the ordered bootstrap subset contains {0,4}. We use the priority field to order these services, just the same as we use it to order the ordered full set {0,1,4}.

Please go ahead and add an explanation

saifshaikh48: Explanation added here: https://github.com/openshift/enhancements/pull/1018/files#diff-3c7c637b883f288bf20ea8e8e521c4962adef0c81b2d149955787c8baf87273fR184-R188

enhancements/windows-containers/health-management.md
```diff
@@ -165,6 +181,11 @@ When making use of a PowerShell script, retries should be present within the scr
 is waiting for the HNS network to be created in order to create an HNS endpoint. In that specific case, the PowerShell
 script which creates the endpoint should include a retry period which waits for the needed Network to exist.
 
+Services marked with `bootstrap: true` will start before any others, using only their subset of priorities to determine
+order. Priorities extend beyond the `bootstrap` stage, meaning that in the `controller` stage, WICD will use priorities 
+of the whole service list to build its ordered list of services. This means that although a bootstrap service 
+starts before a non-bootstrap service, the non-bootstrap service can have a greater priority when WICD is reconciling.
```
aravindhp, line 187: >This means that although a bootstrap service starts before a non-bootstrap service, the non-bootstrap service can have a greater priority when WICD is reconciling.

This does not compute for me. Please provide a scenario when there will be a controller only service that will have a higher priority than a bootstrap service.

You should also call out if a bootstrap service can depend on a controller service. PS: this is not possible.

enhancements/windows-containers/health-management.md
aravindhp: >Lets say the ordered bootstrap subset contains {0,4}. We use the priority field to order these services, just the same as we use it to order the ordered full set {0,1,4}.

This will not work. By definition, `bootstrap` services have a higher priority than the `controller` ones.

saifshaikh48: As discussed in standup:
```
Services marked with `bootstrap: true` will start before any others. Bootstrap services cannot depend on a non-bootstrap
service; WICD will throw an error if this is detected. Similarly, each service that has the bootstrap flag set as true
must have a higher priority than all non-bootstrap services. WICD will throw an error if this is not the case.
```

enhancements/windows-containers/health-management.md
```diff
@@ -168,6 +181,10 @@ When making use of a PowerShell script, retries should be present within the scr
 is waiting for the HNS network to be created in order to create an HNS endpoint. In that specific case, the PowerShell
 script which creates the endpoint should include a retry period which waits for the needed Network to exist.
 
+Services marked with `bootstrap: true` will start before any others. Bootstrap services cannot depend on a non-bootstrap
+service; WICD will throw an error if this is detected. Similarly, each service that has the bootstrap flag set as true
+must have a higher priority than all non-bootstrap services. WICD will throw an error if this is not the case.
+
```
aravindhp, line 187: I would also add that there should be no overlap in the priorities of bootstrap and controller services.

saifshaikh48: I think that's implied already, but sure I can state it explicitly.

aravindhp (approved):
/approve

/hold
so that others can review

sebsoto:
LGTM

enhancements/windows-containers/health-management.md
```diff
@@ -112,6 +122,9 @@ N/A
   To reduce the risk of this, the `MaxConcurrentReconciles` option can be provided when initalizing the ConfigMap
   controller. This will enable the controller to process service ConfigMap events that occur while WMCO is handling
   its BYOH configuration responsibilities.
+* WMCO will not default to containerd as the container runtime until the 6.0.0 timeframe. Therefore, we will be using a 
+  runtime flag `dockerRuntime=false` introduced by the containerd enhancement to develop during the 5.y.z cycle.
+  This flag will dropped come WMCO 6.0.0.
```
selansen, line 127: flag will be dropped

saifshaikh48: Not sure I understand. It already says
`This flag will dropped come WMCO 6.0.0.`

saifshaikh48: oh nvm i see

PR #1019, opened 2022-01-26
Enhancement for improving image pull via shortname security

umohnani8:
This is a proposal to improve the security of image pulls
when the image is refered to by shortname. This is already
implemented in podman and buildah and we would like to
extend this to cri-o so that OCP can benefit from it.

There is no API change happening here. Creating an enhancement
so that we can get feedback on the approach here and whether
this would be helpful for OpenShift. Would also like to get approval
on the list of images in the alias table to be shipped with OCP once
this enhancement is approved.

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

### DIFF
--- a/enhancements/node-tuning/secure-shortname-image-pulls.md
+++ b/enhancements/node-tuning/secure-shortname-image-pulls.md
@@ -1,0 +1,216 @@
+---
+title: secure-shortname-image-pulls
+authors:
+  - "@umohnani8"
+reviewers:
+  - "@mrunalp"
+approvers:
+  - "@mrunalp"
+api-approvers: # in case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers)
+  - N/A
+creation-date: 2022-01-26
+last-updated: 2022-01-26
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - https://issues.redhat.com/browse/RUN-1134
+see-also:
+replaces:
+superseded-by:
+---
+
+# Secure Shortname Image Pulls
+
+## Summary
+
+Pulling images via shortnames is not very secure. There is a possibility of
+spoofing where the image is pulled from as it goes through the list of `unqualified-search-registries`
+checking for a match and pulls the image from the first location it is found in. With this method
+there is no guarantee that the image is actually being pulled from the correct source.
+
+We have made this more secure by adding support for an alias table that has a list of shortname images
+pointing to their fully qualified image name. This alias table is checked for a match whenever a pull
+via shortname is attempted, which guarantees that the image will only be pulled from the location specified
+in the alias table.
+
+## Motivation
+
+We are always trying to make OpenShift more secure. While we do highly discourage the usage of
+shortnames in favour of fully qualified names, there are users who still prefer using shortnames
+so we have updated the pull logic to be more secure than just checking a list of registries and pulling
+from the first location where a match is found.
+
+So, to make the shortname use-case more secure, the c/image library was updated to check an
+alias table when available. An alias table is a list of image names with shortnames pointing to their
+corresponding fully qualified image names. We can use this new feature to ensure that any Red Hat and OpenShift
+images that are referred to by short name will always resolve to our official locations, hence
+guaranteeing that those images will be official.
+
+### Goals
+
+- Secure where the official Red Hat and OpenShift images are pulled from when using shortnames
+- The alias table list will be maintained by us and be specific to the OpenShift use-case
+- The MCO will be used to lay down the alias table
+
+### Non-Goals
+
+- This is not encouraging the use of shortnames. It is instead focused on making the few use-cases
+  that are out there more secure
+
+## Proposal
+
+Enable the shortname alias logic in the cri-o code so that when an image is being pulled and
+referred to by shortname, cri-o will check the alias table first and if there is a match it
+will pull the image from that location. If there is no match it will fallback to check the
+`unqualified-search-registries` list in order and pull the image from the first location where
+a match is found - note, this is what is done currently.
+
+We will maintain the alias table shipped on the OpenShift node via the MCO. We will create an alias
+table for the official Red Hat and OCP images based on a subset of the aliases in the existing tables
+that are shipped in RHEL. We are doing this as the tables shipped with RHEL have a lot of images from
+the community which are not required in the OpenShift use-case.
+
+Note: This proposal is a security enhancement around the shortnames use case and is not an API change.
+
+### User Stories
+
+#### As a user, I would like to use image short names when running my workloads
+
+When the user uses shortname for official Red Hat and OpenShift images that have a matching alias in the
+alias table, the image will be pulled from the matching location specified there. This will ensure
+the security of where these images are being pulled from reducing the risk of spoofing.
+
+### API Extensions
+
+No API changes are needed for this proposal.
+
+### Implementation Details/Notes/Constraints [optional]
+
+Implementing this enhancement requires changes in:
+
+- cri-o/cri-o
+- openshift/machine-config-operator
+
+Update the cri-o code to enable the shortname alias logic to `permissive` mode. This will ensure
+that when an image is referred to by shortname, it will check the alias table first, if there is no
+match, it will then try all the registries in the `unqualified-search-registries` list in order.
+
+Documentation: We will document that while we highly discourage the use of shortnames, there is a way to have
+more secure shortname access for official Red Hat and OpenShift images.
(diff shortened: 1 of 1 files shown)


umohnani8:
@mrunalp PTAL

enhancements/node-tuning/secure-shortname-image-pulls.md
```diff
@@ -0,0 +1,216 @@
...
+So, to make the shortname use-case more secure, the c/image library was updated to check an
+alias table when available. An alias table is a list of image names with shortnames pointing to their
+corresponding fully qualified image names. We can use this new feature to ensure that any Red Hat and OpenShift
+images that are referred to by short name will always resolve to our official locations, hence
+guaranteeing that those images will be official.
+
+### Goals
+
+- Secure where the official Red Hat and OpenShift images are pulled from when using shortnames
+- The alias table list will be maintained by us and be specific to the OpenShift use-case
```
mtrmac, line 50: Will this be 100% automated, or does that involve an on-going manual process to update the lists, that will need to be regularly performed? Is there an established mechanism for doing that?

---

How will the list be shipped? And if it is inside a container image, like the MCO one, are there any concerns WRT relative timing of changes to the “ground truth” (e.g. adding a new official repo) vs. timing of creating/publishing/deploying updates to that list (e.g. on a long-term-support branch of MCO that only rarely receives updates, and those are deployed even more rarely by a customer)?

(The existing CoreOS-shipped list that is built from a RPM has very similar timing issues, of course — it’s a matter of what “specific to OpenShift” means and whether it’s something we can broadly guarantee to be useful in some specific ways, or whether it’s best thought of as a crutch for long-term widely-known images only.)

enhancements/node-tuning/secure-shortname-image-pulls.md
```diff
@@ -0,0 +1,216 @@
...
+No API changes are needed for this proposal.
+
+### Implementation Details/Notes/Constraints [optional]
+
+Implementing this enhancement requires changes in:
+
+- cri-o/cri-o
+- openshift/machine-config-operator
+
+Update the cri-o code to enable the shortname alias logic to `permissive` mode. This will ensure
```
mtrmac, line 92: This seems to refer to the code impacted by https://github.com/cri-o/cri-o/pull/4983/files , and is incorrect about what that option does.

Aliases are _already_ interpreted; compare https://github.com/containers/image/issues/1448#issuecomment-1018501551 and see https://bugzilla.redhat.com/show_bug.cgi?id=2021026 for evidence.

enhancements/node-tuning/secure-shortname-image-pulls.md
```diff
@@ -0,0 +1,216 @@
...
+that when an image is referred to by shortname, it will check the alias table first, if there is no
+match, it will then try all the registries in the `unqualified-search-registries` list in order.
+
+Documentation: We will document that while we highly discourage the use of shortnames, there is a way to have
+more secure shortname access for official Red Hat and OpenShift images.
+
+### Risks and Mitigations
+
+There are no risks involved here as the alias table will be maintained by us in the MCO repo and will only contain aliases for official
+Red Hat and OpenShift images.
```
mtrmac, line 102: There is a known customer broken by us shipping an alias, in https://github.com/containers/image/issues/1448 .

I’m not at all saying that that customer’s setup is a good reason to not ship aliases, I just think we should, at least, acknowledge that changing the aliases _can_ break existing deployments - both _adding_ aliases and _removing_ aliases. And maybe think about what upgrade processes / guarantees we should make (e.g. “we will never drop an alias”, “we will only ever add aliases that match `openshift*`” — I’m not at all suggesting _these_ policies, just that we might need _some_).

enhancements/node-tuning/secure-shortname-image-pulls.md
umohnani8: The MCO can lay down files on the nod during bootstrap, so the plan here is to lay down an `00-alias-table.conf` under `/var/lib/containers/registries.conf.d` similar to how we lay down the the crio config file under crio.conf.d that overrides some of the crio configurations from the defaults we ship upstream with cri-o - https://github.com/openshift/machine-config-operator/blob/master/templates/master/01-master-container-runtime/_base/files/crio.yaml. When OpenShift is updated and the MCO updates, it will update any changes made to the files under templates as well - I am not seeing a timing issue here for that as it has worked for the other config files we lay down via templates, but please let me know if I am missing something.

As for how we will update the alias table, I think it might have to be a manual process initially till we figure out if there is a good way to automate it and where to get the aliases and their matching fully qualified names from for the OCP specific use case.

enhancements/node-tuning/secure-shortname-image-pulls.md
umohnani8: So looks like there is truly no difference between the `permissive` and `disabled` mode when it comes to alias look up? 
The difference seems to be in whether the alias is recorded in the table when a match via `unqualified-search-registries` happen - is that correct?
From what I understood, `permissive` meant look at alias table, if no match is found don't prompt the user asking where to pull from, just check all the registries under `unqualified-search-registries`. And `disabled` meant don't look at any alias table, just check all the registries under `unqualified-search-registries`. This is why we changed cri-o to disabled mode when the first bug occurred https://bugzilla.redhat.com/show_bug.cgi?id=1969599.

If there is truly no difference, then would it matter whether it is in `permissive` or `disabled` mode when it comes to alias look up? We definitely don't want to do `enforcing` mode as that would throw an error when it can't prompt the user to select a location to pull from when an alias is not found.

enhancements/node-tuning/secure-shortname-image-pulls.md
umohnani8: This is a good point. I was assuming that we would keep this list as stable as possible with minimal changes to the aliases already added to the list. Will need to have a discussion on how often we expect to update the alias table and plan to handle upgrades/downgrades with that.

enhancements/node-tuning/secure-shortname-image-pulls.md
mtrmac: The timing issue I worry about is:

- redhat.io/$product/$repo1 , and redhat.io/$product/$repo2 previously existed.
- We add them as aliases $product/$repo1 and $product/$repo2
- Customers start using those aliases, and rely on them existing.

(time passes)

- The product in question introduces redhat.io/$product/repo3 , and announces it.
- Customers, habitually, refer to $product/$repo3 , expecting an alias. This does not work because the alias doesn’t exist. (Customers, presumably, grumpily use a fully-qualified name, or define their own alias, or in the worst case, change USR.)

(time passes)

- A MCO update ships a $product/$repo3 alias. But, by then, no-one needs it to exist and it might even, strictly speaking, break something.

enhancements/node-tuning/secure-shortname-image-pulls.md
mtrmac: > So looks like there is truly no difference between the `permissive` and `disabled` mode when it comes to alias look up?

See https://github.com/containers/image/blob/70982d037a7a006fd3806dfb0882840aac2e2259/pkg/shortnames/shortnames.go#L302 , the mode value is not consulted at all at that point.

> The difference seems to be in whether the alias is recorded in the table when a match via `unqualified-search-registries` happen - is that correct?

Only in interactive use, i.e. not with CRI-O.

> From what I understood, `permissive` meant look at alias table, if no match is found don't prompt the user asking where to pull from, just check all the registries under `unqualified-search-registries`. And `disabled` meant don't look at any alias table, just check all the registries under `unqualified-search-registries`.

No, the mode _only_ governs lookup in USR (and possibly recording aliases for the outcome of that USR lookup).
- disabled = old risky USR lookup.
- enforcing = prompt interactively, fail if impossible.
- permissive = prompt if interactive, old risky USR lookup if not interactive.

> If there is truly no difference, then would it matter whether it is in `permissive` or `disabled` mode when it comes to alias look up?

No.

enhancements/node-tuning/secure-shortname-image-pulls.md
mtrmac: > where to get the aliases and their matching fully qualified names from for the OCP specific use case.

BTW is there already some conceptual idea of what the ”OCP specific use case” list would be? I.e. how does it differ from the RHEL list, if it does differ at all? (Note that the alias in https://bugzilla.redhat.com/show_bug.cgi?id=2021026 is an `openshift3/…` alias; would that be included or excluded?)

Without articulating how that list differs it’s hard to me to think whether we need it.

umohnani8:
We will not be moving forward with adding this to cri-o/ocp. Closing the enhancement.

PR #1020, opened 2022-01-27
Add enhancement for AWS Load Balancer operator

arjunrn:
AWS Load Balancer operator is an operator for running and managing the [aws-load-balancer-controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/). This controller is for Ingress and Service resource and has better integration with AWS services.

### DIFF
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -6,7 +6,7 @@ reviewers:
   - "@Miciah"
   - "@alebedev87"
 approvers:
-  - TBD
+  - "@Miciah"
 api-approvers:
   - TBD
 creation-date: 2022-01-27
@@ -14,7 +14,7 @@ last-updated: 2022-01-27
 tracking-link:
   - https://issues.redhat.com/browse/CFEPLAN-39
 see-also:
-  - "/enhancements/ingress/transition-ingress-from-beta-to-stable.mde"
+  - "/enhancements/ingress/transition-ingress-from-beta-to-stable.md"
 replaces:
 superseded-by:
 ---
@@ -98,6 +98,7 @@ distributed through Operator Hub.
   resources.
 * Support for `ip` mode for Ingress resources
 * Support for NLB backed Services through the lb-controller.
+* Support for the AWS VPC CNI plugin.
 
 ## Proposal
 
@@ -268,7 +269,10 @@ controllers cannot be started with different configurations.
 The lb-controller also requires that the subnets where the load balancers are
 provisioned have certain resource tags present on them. The operator can detect
 the subnets and tag them or the user can do this. This is specified with the
-_AutoTagSubnet_ field.
+_AutoTagSubnet_ field. This field could be dropped if the subnets are already
+tagged with the correct values. This is explained in further detail in the
+section [Changes in the Openshift
+Installer](#changes-in-the-openshift-installer).
 
 The user can also specify the Ingress class which the lb-controller will
 reconcile. This could default to **_“alb”_**. This value is required because if
@@ -300,6 +304,20 @@ on behalf of the lb-controller and mount the minted credentials in the
 _Deployment_ of the controller. This means that the operator will only work on
 OCP/OKD but this is sufficient for the initial release.
 
+#### Changes in the OpenShift Installer
+
+The OpenShift Installer provisions a VPC and private and public subnets within
+it. It currently only tags the private subnets with
+`kubernetes.io/role/internal-elb` but does not tags the public subnets with
+`kubernetes.io/role/elb`. To reduce the functionality of the operator the
+installer could itself tag the public subnets with the appropriate tag. For
+clusters installed into existing VPC the installer could similarly add the tags
+to the subnets specified in the installation manifest.
+
+The VPC ID also has to be passed to the `lb-controller` as a CLI parameter.
+Currently, there is no resource which has the VPC ID. The installer could be
+modified to write the VPC id into the `Infrastructure` resource _Status_ for
+both provisioned and existing VPCs.
 
 ### Risks and Mitigations
 
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -108,8 +108,8 @@ distributed through Operator Hub.
 
 The user wants to route traffic for the domain example.com to pods of Service
 example-service and have TLS termination and DDoS protection on the load
-balancer. The user deploys the aws-alb-operator and creates an instance of the
-resource AWSLoadBalancerController as follows:
+balancer. The user deploys the operator and creates an instance of the
+resource `AWSLoadBalancerController` as follows:
 
 ```yaml
 kind: AWSLoadBalancerController
@@ -117,9 +117,21 @@ group: networking.openshift.io/v1beta1
 metadata:
   name: cluster
 spec:
-  autoTagSubnet: true
+  subnetTagging: auto
   ingressClass: tls-termination
 ```
+
+Then they create an `IngressClass` resource as follows:
+
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: IngressClass
+metadata:
+  name: tls-termination
+spec:
+  controller: ingress.k8s.aws/alb
+```
+
 Then they create an ingress resource with the following schema:
 ```yaml
 apiVersion: networking.k8s.io/v1
@@ -144,10 +156,12 @@ spec:
                 port:
                   number: 80
 ```
-The controller then creates an application load balancer with an HTTPS listener
-which terminates the TLS connection with the certificate ARN which is specified
-in the annotation. If the certificate does not exist, the load balancer is not
-created. It also enables AWS Shield on the Elastic IP associated with the ALB.
+
+The controller then creates an application load balancer with an HTTPS listener.
+The ALB only has an HTTPS listener because a certificate has been specified
+explicitly through the annotations. If the certificate does not exist, the load
+balancer is not created. It also enables AWS Shield on the Elastic IP associated
+with the ALB.
 
 #### Multiple Ingress through single ALB
 
@@ -163,7 +177,7 @@ metadata:
   annotations:
     alb.ingress.kubernetes.io/scheme: internet-facing
     alb.ingress.kubernetes.io/group.name: example
-    alb.ingress.kubernetes.io/group.order: “1”
+    alb.ingress.kubernetes.io/group.order: "1"
 spec:
   ingressClass: alb
   rules:
@@ -173,9 +187,9 @@ spec:
         - path: /blog
           backend:
             service:
-        name: example-1
-        port:
-          number: 80
+              name: example-1
+              port:
+                number: 80
 —--
 kind: Ingress
 metadata:
@@ -183,7 +197,7 @@ metadata:
   annotations:
     alb.ingress.kubernetes.io/scheme: internet-facing
     alb.ingress.kubernetes.io/group.name: example
-    alb.ingress.kubernetes.io/group.order: “2”
+    alb.ingress.kubernetes.io/group.order: "2"
 spec:
   ingressClass: alb
   rules:
@@ -193,9 +207,9 @@ spec:
         - path: /store
           backend:
             service:
-        name: example-2
-        port:
-          number: 80
+              name: example-2
+              port:
+                number: 80
 —--
 kind: Ingress
 metadata:
@@ -203,7 +217,7 @@ metadata:
   annotations:
     alb.ingress.kubernetes.io/scheme: internet-facing
     alb.ingress.kubernetes.io/group.name: example
-    alb.ingress.kubernetes.io/group.order: “3”
+    alb.ingress.kubernetes.io/group.order: "3"
 spec:
   ingressClass: alb
   rules:
@@ -213,9 +227,9 @@ spec:
         - path: /
           backend:
             service:
-        name: example-3
-        port:
-          number: 80
+              name: example-3
+              port:
+                number: 80
 ```
 
 In this example the annotations `alb.ingress.kubernetes.io/group.name` specifies
@@ -231,7 +245,7 @@ lb-controller instance. The resource has the following _Spec_:
 
 ```golang
 type AWSLoadBalancerControllerSpec struct {
-  AutoTagSubnet bool // indicates if the operator should tag the subnets
+  SubnetTagging SubnetTaggingPolicy
   // +optional
   AdditionalResourceTags map[string]string // Default AWS Tags that will be applied to all AWS resources managed by this controller (default [])
   // +optional
@@ -254,6 +268,12 @@ type DeploymentConfig struct {
   // +optional
   Replicas int
 }
+
+type SubnetTaggingPolicy string
+const (
+    AutoSubnetTaggingPolicy SubnetTaggingPolicy = "Auto"
+    ManualSubnetTaggingPolicy SubnetTaggingPolicy = "Manual"
+)
 ```
 
 
@@ -268,8 +288,8 @@ controllers cannot be started with different configurations.
 
 The lb-controller also requires that the subnets where the load balancers are
 provisioned have certain resource tags present on them. The operator can detect
-the subnets and tag them or the user can do this. This is specified with the
-_AutoTagSubnet_ field.
+the subnets and tag them or the user can do this. This can be enabled by setting
+the `SubnetTagging` to `AutoSubnetTaggingPolicy`.
 
 The user can also specify the Ingress class which the lb-controller will
 reconcile. This could default to **_“alb”_**. This value is required because if
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -47,7 +47,7 @@ AWS has three kinds of load balancers:
     route requests based on the HTTP path and host. It also integrates with other
     AWS services like DDoS protection and WAF.
 
-### Openshift Router
+### OpenShift Router
 In OCP/OKD Ingress resources are serviced by the Openshift
 [router](https://github.com/openshift/router). Routers are created by the
 [cluster-ingress-operator](https://github.com/openshift/cluster-ingress-operator/)(CIO)
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -24,15 +24,15 @@ superseded-by:
 ## Summary
 
 Users would like to use [Application Load
-Balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html)(ALBs)
+Balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html) (ALBs)
 for their Ingress resources so that they can use [Amazon Certificate
-Manager](https://aws.amazon.com/certificate-manager/)(ACM) to manage
+Manager](https://aws.amazon.com/certificate-manager/) (ACM) to manage
 certificates for the Ingress domains and also attach [Web Application
-Firewalls](https://aws.amazon.com/waf/) to the ALB for security and traffic
-shaping. The Openshift [router](github.com/openshift/router) uses an NLB which
-passes through all traffic to the router. TLS termination is done on the router.
-A WAF cannot be attached to the NLB and the router cannot be configured to
-perform TLS termination at the NLB.
+Firewalls](https://aws.amazon.com/waf/) (WAFs) to the ALB for security and traffic
+shaping. The OpenShift [router](https://github.com/openshift/router) uses a classic ELB or an NLB, which
+passes through all traffic to the router. TLS termination is done on the router or backend pod.
+A WAF cannot be attached to the ELB or NLB, and the router cannot be configured to
+perform TLS termination at the ELB or NLB.
 
 ## Background
 
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -328,7 +328,7 @@ resources.
 
 #### Parallel operation  of the Openshift Router and lb-controller
 
-The openshift Router manages ingresses which don’t have any ingress class value.
+The OpenShift router manages ingresses that don’t have any ingress class value.
 This is the expected behavior currently and the lb-controller should not attempt
 to reconcile Ingresses which don’t have an ingress class set. If the user wants
 to switch over to using the lb-controller as default they can create an
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -269,10 +269,7 @@ controllers cannot be started with different configurations.
 The lb-controller also requires that the subnets where the load balancers are
 provisioned have certain resource tags present on them. The operator can detect
 the subnets and tag them or the user can do this. This is specified with the
-_AutoTagSubnet_ field. This field could be dropped if the subnets are already
-tagged with the correct values. This is explained in further detail in the
-section [Changes in the Openshift
-Installer](#changes-in-the-openshift-installer).
+_AutoTagSubnet_ field.
 
 The user can also specify the Ingress class which the lb-controller will
 reconcile. This could default to **_“alb”_**. This value is required because if
@@ -303,21 +300,6 @@ The operator will also create a
 on behalf of the lb-controller and mount the minted credentials in the
 _Deployment_ of the controller. This means that the operator will only work on
 OCP/OKD but this is sufficient for the initial release.
-
-#### Changes in the OpenShift Installer
-
-The OpenShift Installer provisions a VPC and private and public subnets within
-it. It currently only tags the private subnets with
-`kubernetes.io/role/internal-elb` but does not tags the public subnets with
-`kubernetes.io/role/elb`. To reduce the functionality of the operator the
-installer could itself tag the public subnets with the appropriate tag. For
-clusters installed into existing VPC the installer could similarly add the tags
-to the subnets specified in the installation manifest.
-
-The VPC ID also has to be passed to the `lb-controller` as a CLI parameter.
-Currently, there is no resource which has the VPC ID. The installer could be
-modified to write the VPC id into the `Infrastructure` resource _Status_ for
-both provisioned and existing VPCs.
 
 ### Risks and Mitigations
 
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -93,7 +93,7 @@ distributed through Operator Hub.
 * Implement operator which can be installed and used in OCP/OKD.
 
 ### Non-Goals
-* Implementing support for ALB for Openshift Routes
+* Implementing support for ALB for OpenShift Routes
 * Support for multiple installations of the controller i.e. multiple custom
   resources.
 * Support for `ip` mode for Ingress resources
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -86,7 +86,7 @@ sake of brevity.
 ## Motivation
 
 Implement an operator, tentatively called `aws-load-balancer-operator`, which
-deploys and manages an instance of the lb-operator. The operator will be
+deploys and manages an instance of lb-controller. The operator will be
 distributed through Operator Hub.
 
 ### Goals
@@ -287,7 +287,7 @@ versions other fields from the
 [PodSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core)
 could also be exposed here.
 
-Finally, WAFs and DDos prevent configurations are attached to the load balancer
+Finally, WAFs and DDoS prevention configurations are attached to the load balancer
 through annotations specified on the Ingress resource. But users might want to
 limit which features are available and the `DisabledAddons` array can be used to
 specify which features/addons have to be disabled. By default all the addons
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -388,7 +388,7 @@ TBD
 
 ## Alternatives
 
-The existing Openshift Router could be modified to provision ALBs for Ingresses
+The existing OpenShift router could be modified to provision ALBs for Ingresses
 which specify it through an annotation. The disadvantage would that the router
 would be replicating a lot of the functionality which is present in the
 lb-controller.
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -326,7 +326,7 @@ to restrict the features supported by the controller to only Ingresses the
 lb-controller will have to be modified so that it does not reconcile any Service
 resources.
 
-#### Parallel operation  of the Openshift Router and lb-controller
+#### Parallel operation  of the OpenShift router and lb-controller
 
 The OpenShift router manages ingresses that don’t have any ingress class value.
 This is the expected behavior currently and the lb-controller should not attempt
--- a/enhancements/ingress/aws-load-balancer-operator.md
+++ b/enhancements/ingress/aws-load-balancer-operator.md
@@ -1,0 +1,394 @@
+---
+title: aws-load-balancer-operator 
+authors:
+  - "@arjunrn"
+reviewers:
+  - "@Miciah"
+  - "@alebedev87"
+approvers:
+  - TBD
+api-approvers:
+  - TBD
+creation-date: 2022-01-27
+last-updated: 2022-01-27
+tracking-link:
+  - https://issues.redhat.com/browse/CFEPLAN-39
+see-also:
+  - "/enhancements/ingress/transition-ingress-from-beta-to-stable.mde"
+replaces:
+superseded-by:
+---
+
+# AWS Load Balancer Operator
+
+## Summary
+
+Users would like to use [Application Load
+Balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html)(ALBs)
+for their Ingress resources so that they can use [Amazon Certificate
+Manager](https://aws.amazon.com/certificate-manager/)(ACM) to manage
+certificates for the Ingress domains and also attach [Web Application
+Firewalls](https://aws.amazon.com/waf/) to the ALB for security and traffic
+shaping. The Openshift [router](github.com/openshift/router) uses an NLB which
+passes through all traffic to the router. TLS termination is done on the router.
+A WAF cannot be attached to the NLB and the router cannot be configured to
+perform TLS termination at the NLB.
+
+## Background
+
+### AWS Load Balancing
+AWS has three kinds of load balancers:
+1. **Classic Load Balancer:** This is a Level 4 load balancer with some Level 7
+    features. It does not have integration with
+    AWS’s newer services.
+2. **Network Load Balancer:** This is also a Level 4 load balancer but has been
+    optimized for performance and cost compared to the Classic Load Balancer.
+3. **Application Load Balancer:** This is a Level 7 load balancer which can
+    route requests based on the HTTP path and host. It also integrates with other
+    AWS services like DDoS protection and WAF.
+
+### Openshift Router
+In OCP/OKD Ingress resources are serviced by the Openshift
+[router](https://github.com/openshift/router). Routers are created by the
+[cluster-ingress-operator](https://github.com/openshift/cluster-ingress-operator/)(CIO)
+for every
+_[IngressController](https://docs.openshift.com/container-platform/4.9/networking/configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-ingress-controller.html)_
+resource. By default on AWS the CIO creates a Service of type
+[LoadBalancer](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer).
+The in-tree Service controller in turn provisions a Network Load
+Balancer with a TCP listener. The CIO then creates a DNS entry for the subdomain
+specified in the IngressController resource with the load balancer created in
+the previous step as the target. 
+
+### aws-load-balancer-controller
+AWS maintains an out-of-tree controller with support for ALBs for Ingress and
+Service type resources. The in-tree Service controller only provisions [classic
+load
+balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html)
+and
+[Network](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html)
+Load Balancers. The out-of-tree controller supports integration of Kubernetes
+resources with more of the AWS services like WAFs and the DDoS prevent service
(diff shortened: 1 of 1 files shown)


enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,384 @@
...
+
+### Goals
+* Limit configurations to those supported on OCP/OKD.
+
+### Non-Goals
+* Implementing support for ALB for Openshift Routes
+* Support for multiple installations of the controller i.e. multiple custom
+  resources.
+
+## Proposal
```
dhellmann, line 96: I can't tell if this is proposing a new operator delivered via OLM, or if the suggestion is to add this to release payload. Can you talk about how the new operator would be added to a cluster?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,384 @@
...
+
+## Motivation
+
+Implement an operator, tentatively called `aws-load-balancer-operator`, which
+deploys and manages an instance of the aws-load-balancer-controller. 
+
+### Goals
+* Limit configurations to those supported on OCP/OKD.
+
+### Non-Goals
```
alebedev87, line 91: Support for `Service` - `NLB` is also non goal.
How about the `ip` traffic mode? We won't be able to support it either.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,384 @@
...
+provisioned have certain resource tags present on them. The operator can detect
+the subnets and tag them or the user can do this. This is specified with the
+_AutoTagSubnet_ field.
+
+The user can also specify the Ingress class which the lb-controller will
+reconcile. This could default to **_“alb”_**. This value is required because if
+an ingress class value is not passed to the controller it attempts to reconcile
+all Ingress resources which don’t have an ingress class annotation or where the
+ingress class field is empty. Passing the value of “alb” does not restrict the
+number of ingress classes which the controller can reconcile.
```
alebedev87, line 271: > does not restrict the number of ingress classes which the controller can reconcile.

Didn't quite get this.
You meant the number of ingress groups (the one you can set with `group.name` annotation)?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,384 @@
...
+on behalf of the lb-controller and mount the minted credentials in the
+_Deployment_ of the controller. This means that the operator will only work on
+OCP/OKD but this is sufficient for the initial release.
+
+
+### Risks and Mitigations
+
+#### Restricting features in the lb-controller
+
+Some changes will have to be made in the lb-controller to ensure that the users
```
alebedev87, line 298: Sometimes you call it `aws-load-balancer-controller` sometimes `lb-controller`. I think the former is clearer as you introduced it with this name in **Background** chapter.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,384 @@
...
+)
+
+type DeploymentConfig struct {
+  // +optional
+  Replicas int
+}
+```
+
+
+### Implementation Details/Notes/Constraints
```
alebedev87, line 252: May be, for the completeness of the picture, it's worth mentioning the CRDs of `aws-load-balancer-controller` and the fact that we leave this configuration up to the end user.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,384 @@
...
+The openshift Router manages ingresses which don’t have any ingress class value.
+This is the expected behavior currently and the lb-controller should not attempt
+to reconcile Ingresses which don’t have an ingress class set. If the user wants
+to switch over to using the lb-controller as default they can create an
+_IngressClass_ resource with the annotation
+`ingressclass.kubernetes.io/is-default-class` set to _true_ and the
+`spec.controller` value should be _ingress.k8s.aws/alb._ The user can also
+create multiple `IngressClass` resources with different configurations and have
+the `spec.controller` set to _ingress.k8s.aws/alb._ All Ingresses which belong
+to these classes will be reconciled by the lb-controller.
```
alebedev87, line 330: How is this possible? `aws-load-balancer-controller` expects only 1 `IngressClass`, so will reconcile only `Ingress`es of this class.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: I've mentioned that I will refer to it as lb-controller in the background section.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: No I mean IngressClasses themselves. If the `IngressClass` has the `spec.controller` set to `ingress.k8s.aws/alb` then the controller will reconcile those Ingresses regardless of the ingress class passed to it through the CLI argument. You can find more details in the "Parallel Operation" section.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: Because that's how it's been implemented in the controller. The CLI parameter is kind of non-obvious in how it works. It the ingress class is specified through the annotation then it limits it to that parameter. However if it is specified through the field in the spec then it ignores that parameter and only verifies that the corresponding `IngressClass` has the `controller` field set to the value mentioned above. Now that I think about it might be confusing for the user. We should probably disable the annotations based Ingress Class altogether.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: The CRDs are mentioned in the background section. I guess we could add more examples to the user stories.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: I've mentioned it in the _Motivation_ section.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: Here's the how the ingress is [classified](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/46668f11bdc548a3df60d1a921b0e70cbf0f91f1/pkg/ingress/group_loader.go#L171). Maybe we should hide the `--ingress-class` parameter altogether from the user. And expect the user to create the `IngressClass` manually.

alebedev87: Interesting. Yes, that kinda makes sense, as now, with `IngressClass` resource there is no need to distinguish by the class name (like in the annotation), `.spec.controller` came into play and many classes can now be implemented by the same controller which creates a new abstraction layer.
I think that the docs are not well updated in some places cause I definitely saw the statements about `--ingress-class` flag limiting the reconciliation logic. Also, I'm still confused about how multiple `IngressClassParams` with conflicting fields are managed.

Anyway, taking this logic into account, I think I agree with you that setting `--disable-ingress-class-annotation` as default and removing `IngressClass` from the API is a nice option.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,393 @@
...
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: example
+  annotations:
+    alb.ingress.kubernetes.io/scheme: internet-facing
+    alb.ingress.kubernetes.io/certificate-arn: arn:aws:acm:us-west-2:xxxxx
+    alb.ingress.kubernetes.io/shield-advanced-protection: true
+spec:
+  ingressClass: tls-termination
```
alebedev87, line 133:
```suggestion
  ingressClassName: tls-termination
```

Same in the other examples.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,393 @@
...
+  rules:
+  - host: example.com
+    https:
```
alebedev87, line 136:
```suggestion
    http:
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,393 @@
...
+        paths:
+- path: /
+  backend:
+    service:
+name: example-service
+port:
+  number: 80
```
alebedev87, line 143: Indent is missing.

Same in the other examples.

enhancements/ingress/aws-load-balancer-operator.md
alebedev87: Just to finish on this one. I don't know why I thought that multiple `IngressClassParams` would cause a conflict, a brain bug I guess. There is no problem with this as each ingress belongs to only 1 ingress class. The conflict can be only if ingresse belonging to different classes are grouped in a single ingress group, in that case different fields from the `IngressClassParams` may cause the conflict, like here when different schemes are used:
```
{"level":"debug","ts":1643651687.9120367,"logger":"controller-runtime.manager.events","msg":"Warning","object":{"kind":"Ingress","namespace":"hello2","name":"hello-openshift","uid":"0e68ba57-39ea-49a4-af63-90fe1034288b","apiVersion":"networking.k8s.io/v1","resourceVersion":"167518"},"reason":"FailedBuildModel","message":"Failed build model due to conflicting scheme: map[internal:{} internet-facing:{}]"}
```
But this is no different from the situation when a single ingress class is used and multiple ingresses set conflicting settings in the annotations.

enhancements/ingress/aws-load-balancer-operator.md
alebedev87: Yeah, we can add an example of the usage of `IngressClassParams` indeed.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+Users would like to use [Application Load
+Balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html)(ALBs)
+for their Ingress resources so that they can use [Amazon Certificate
+Manager](https://aws.amazon.com/certificate-manager/)(ACM) to manage
+certificates for the Ingress domains and also attach [Web Application
+Firewalls](https://aws.amazon.com/waf/) to the ALB for security and traffic
+shaping. The Openshift [router](github.com/openshift/router) uses an NLB which
+passes through all traffic to the router. TLS termination is done on the router.
+A WAF cannot be attached to the NLB and the router cannot be configured to
+perform TLS termination at the NLB.
```
Miciah, line 35: The ingress operator configures a classic ELB on AWS by default, with NLB as an option.  
```suggestion
Users would like to use [Application Load
Balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html) (ALBs)
for their Ingress resources so that they can use [Amazon Certificate
Manager](https://aws.amazon.com/certificate-manager/) (ACM) to manage
certificates for the Ingress domains and also attach [Web Application
Firewalls](https://aws.amazon.com/waf/) (WAFs) to the ALB for security and traffic
shaping. The OpenShift [router](https://github.com/openshift/router) uses a classic ELB or an NLB, which
passes through all traffic to the router. TLS termination is done on the router or backend pod.
A WAF cannot be attached to the ELB or NLB, and the router cannot be configured to
perform TLS termination at the ELB or NLB.
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+1. **Classic Load Balancer:** This is a Level 4 load balancer with some Level 7
+    features. It does not have integration with
+    AWS’s newer services.
+2. **Network Load Balancer:** This is also a Level 4 load balancer but has been
+    optimized for performance and cost compared to the Classic Load Balancer.
+3. **Application Load Balancer:** This is a Level 7 load balancer which can
+    route requests based on the HTTP path and host. It also integrates with other
+    AWS services like DDoS protection and WAF.
+
+### Openshift Router
```
Miciah, line 50:
```suggestion
### OpenShift Router
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+In OCP/OKD Ingress resources are serviced by the Openshift
+[router](https://github.com/openshift/router). Routers are created by the
+[cluster-ingress-operator](https://github.com/openshift/cluster-ingress-operator/)(CIO)
+for every
+_[IngressController](https://docs.openshift.com/container-platform/4.9/networking/configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-ingress-controller.html)_
+resource. By default on AWS the CIO creates a Service of type
+[LoadBalancer](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer).
+The in-tree Service controller in turn provisions a Network Load
```
Miciah, line 58:
```suggestion
In OCP/OKD, Ingress resources are serviced by the OpenShift
[router](https://github.com/openshift/router). Routers are created by the
[cluster-ingress-operator](https://github.com/openshift/cluster-ingress-operator/) (CIO)
for every
_[IngressController](https://docs.openshift.com/container-platform/4.9/networking/configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-ingress-controller.html)_
resource. By default on AWS, the CIO creates a Service of type
[LoadBalancer](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer).
The in-tree Service controller in turn provisions a classic Elastic Load
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+Balancer with a TCP listener. The CIO then creates a DNS entry for the subdomain
+specified in the IngressController resource with the load balancer created in
+the previous step as the target. 
+
+### aws-load-balancer-controller
+AWS maintains an out-of-tree controller with support for ALBs for Ingress and
+Service type resources. The in-tree Service controller only provisions [classic
```
Miciah, line 65:
```suggestion
AWS maintains an out-of-tree controller with support for ALBs for Ingress and
NLBs for Service type resources. The in-tree Service controller only provisions [classic
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+load
+balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html)
+and
+[Network](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html)
+Load Balancers. The out-of-tree controller supports integration of Kubernetes
+resources with more of the AWS services like WAFs and the DDoS prevent service
+[Shield](https://docs.aws.amazon.com/waf/latest/developerguide/ddos-overview.html).
+It also enables traffic routing from the load balancer directly to the pods of
+an Ingress or Service if the cluster uses the [AWS VPC Container Networking
+Interface](https://github.com/aws/amazon-vpc-cni-k8s). In addition it also
```
Miciah, line 75: Is support for amazon-vpc-cni-k8s in scope or out of scope?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+introduces 2 Custom Resources namely
+[`IngressClassParams`](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/guide/ingress/ingress_class/#ingressclassparams)
```
Miciah, line 77: Is support for `IngressClassParams` in scope?  The same parameters that can be configured using `IngressClassParams` can all be configured using annotations, right?  Will we support both the annotations and the CRD?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+_IngressClassParams_ is used for specifying custom parameters for Ingresses of a
+certain class.  The controller uses _TargetGroupBinding_ internally and it can
+also be used by users to manage the load balancer externally. The
+aws-load-balancer will henceforth be referred to as **“lb-controller”** for the
+sake of brevity.
+
+## Motivation
+
+Implement an operator, tentatively called `aws-load-balancer-operator`, which
+deploys and manages an instance of the lb-operator. The operator will be
```
Miciah, line 89: The "lb-operator" in this paragraph is actually the "lb-controller" defined in the previous paragraph, correct?  
```suggestion
deploys and manages an instance of lb-controller. The operator will be
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+distributed through Operator Hub.
+
+### Goals
+* Implement operator which can be installed and used in OCP/OKD.
+
+### Non-Goals
+* Implementing support for ALB for Openshift Routes
+* Support for multiple installations of the controller i.e. multiple custom
+  resources.
```
Miciah, line 98: Is the API being designed to accommodate a future enhancement adding this capability in case it should be come possible upstream and users should request the capability?  

Actually, I want to ask this question about the other non-goals too, but especially this one.  

On further thought and on reading https://github.com/openshift/enhancements/pull/1020/files#r794420487, does it make sense at all to have multiple instances running?  It seems that every instance will match every ingress that has an ingressclass that has `spec.controller: ingress.k8s.aws/alb`.  I suppose, though, that having multiple instances could be useful in the future if we also added support for managing services or routes, right?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+
+## Proposal
+
+### User Stories
+
+#### TLS Termination and DDoS protection
+
+The user wants to route traffic for the domain example.com to pods of Service
+example-service and have TLS termination and DDoS protection on the load
+balancer. The user deploys the aws-alb-operator and creates an instance of the
```
Miciah, line 110:
```suggestion
balancer. The user deploys the aws-load-balancer-operator and creates an instance of the
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+resource AWSLoadBalancerController as follows:
+
+```yaml
+kind: AWSLoadBalancerController
+group: networking.openshift.io/v1beta1
+metadata:
+  name: cluster
+spec:
+  autoTagSubnet: true
```
Miciah, line 119: Avoid literal Boolean values.  The [upstream API conventions warn against using them](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types), and Yaml makes them extra confusing.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+  ingressClass: tls-termination
```
Miciah, line 120: Is creating the ingressclass object optional?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+        paths:
+          - path: /
+            pathType: Exact
+            backend:
+              service:
+                name: example-service
+                port:
+                  number: 80
+```
+The controller then creates an application load balancer with an HTTPS listener
```
Miciah, line 146: The reason it creates an HTTPS listener only (and not also an HTTP listener) is because `alb.ingress.kubernetes.io/certificate-arn` is specified and `alb.ingress.kubernetes.io/listen-ports` is omitted—correct?  It would be helpful to note this for people who are unfamiliar with aws-load-balancer-controller.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+then creates 3 Ingress resources as follows:
+
+```yaml
+kind: Ingress
+metadata:
+  name: example-1
+  annotations:
+    alb.ingress.kubernetes.io/scheme: internet-facing
+    alb.ingress.kubernetes.io/group.name: example
+    alb.ingress.kubernetes.io/group.order: “1”
```
Miciah, line 165: Smartquotes will foil inattentive copy-pasters.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+  rules:
+  - host: example.com
+    http:
+        paths:
+        - path: /blog
+          backend:
+            service:
+        name: example-1
+        port:
+          number: 80
```
Miciah, line 177: Indentation is wrong for this `service:` stanza, and likewise for the two ingresses below.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+attached to the same ALB.
+
+### API Extensions
+
+This operator uses a cluster-scoped Custom Resource to configure the
+lb-controller instance. The resource has the following _Spec_:
+
+```golang
+type AWSLoadBalancerControllerSpec struct {
+  AutoTagSubnet bool // indicates if the operator should tag the subnets
```
Miciah, line 233: What are the semantics of `AutoTagSubnet`?  What tag does the operator apply to the subnets, how does it determine which subnets to tag, why does it tag the subnets, and does it automatically update tags as the CR or subnets are created, updated, and deleted?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+  // +optional
+  AdditionalResourceTags map[string]string // Default AWS Tags that will be applied to all AWS resources managed by this controller (default [])
```
Miciah, line 235: Are tags specified in OpenShift's cluster infrastructure config used at all by this operator?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+  // +optional
+  IngressClass string   // indicates the class of ingress for which the controller should provision ALBs
```
Miciah, line 237: Use API defaulting (`// +kubebuilder:default=alb`) and make sure the godoc mentions the default value.  The field could also have some validation.  

Oh, from https://github.com/openshift/enhancements/pull/1020/files#r794420487, it seems that this field should be deleted, right?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+  // +optional
+  Config *DeploymentConfig
+  // +optional
+  DisabledAddons []AWSAddon // indicates which AWS addons should be disabled.
```
Miciah, line 241: I don't know the context here; why is this opt-out instead of opt-in?  

The risk of opt-in is that, I would imagine, AWS could add additional addons after the cluster administrator had configured the operator, and then users would have access to addons to which the administrator might not want users to have access, right?  

To me, it would make make more sense to have an opt-in list with a reasonable set of defaults (and mention in the godoc that the default value is subject to change), unless the number of existing addons would make specifying an opt-in list unwieldy.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+
+const (
+  AWSAddonShield AWSAddon = "AWSShield"
+  AWSAddonWAFv1 AWSAddon = "AWSWAFv1"
+  AWSAddonWAFv2 AWSAddon = "AWSWAFv2"
+)
+
+type DeploymentConfig struct {
+  // +optional
+  Replicas int
```
Miciah, line 254: Why does this need to be configurable?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+Router](#parallel-operation-of-the-openshift-router-and-lb-controller) and
+lb-controller section.
+
+The _DeploymentConfig_ field contains any configuration required for the
+deployment. The number of replicas is the only field for now. But in future
+versions other fields from the
+[PodSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core)
+could also be exposed here.
+
+Finally, WAFs and DDos prevent configurations are attached to the load balancer
```
Miciah, line 289:
```suggestion
Finally, WAFs and DDoS prevention configurations are attached to the load balancer
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+specify which features/addons have to be disabled. By default all the addons
+will be enabled. They are disabled by setting the addon CLI flag to false. When
+an addon which was previously enabled is disabled the controller does not remove
+the existing addon attachment from the provisioned load balancers.
+
+The operator will also create a
+[CredentialRequest](https://docs.openshift.com/container-platform/4.9/rest_api/security_apis/credentialsrequest-cloudcredential-openshift-io-v1.html#credentialsrequest-cloudcredential-openshift-io-v1)
+on behalf of the lb-controller and mount the minted credentials in the
+_Deployment_ of the controller. This means that the operator will only work on
+OCP/OKD but this is sufficient for the initial release.
```
Miciah, line 301: Could we document how users could create a compatible secret when the cloud credential operator isn't available to create it?

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+target Ingress resources. These changes may also be contributed back upstream
+since it may benefit other users.
+
+1. The annotation `alb.ingress.kubernetes.io/target-type` can be set to the
+value **_ip_** or **_instance_**. When the value is set to _instance_ the
+created load balancer will be configured to route traffic through a NodePort
+service. The _ip_ value only works when the cluster uses the Amazon VPC CNI and
+hence will not work on OCP/OKD clusters. The controller has to be modified so
+that this annotation is ignored and the controller always uses routing through
+_NodePort_ services.
```
Miciah, line 319: Does this affect source preservation? 

Maybe worth mentioning <https://bugzilla.redhat.com/show_bug.cgi?id=2023681>.  Based on <https://bugzilla.redhat.com/show_bug.cgi?id=2023681#c2>, I think you will be unaffected if you use `ip`.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+"external"` is specified on a _Service_ resource of type _LoadBalancer_ the
+in-tree controller ignores this resource and the lb-controller instead
+provisions a Network Load Balancer with the correct configuration. Since we want
+to restrict the features supported by the controller to only Ingresses the
+lb-controller will have to be modified so that it does not reconcile any Service
+resources.
+
+#### Parallel operation  of the Openshift Router and lb-controller
+
+The openshift Router manages ingresses which don’t have any ingress class value.
```
Miciah, line 331:
```suggestion
The OpenShift router manages ingresses that don’t have any ingress class value.
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+
+Implement an operator, tentatively called `aws-load-balancer-operator`, which
+deploys and manages an instance of the lb-operator. The operator will be
+distributed through Operator Hub.
+
+### Goals
+* Implement operator which can be installed and used in OCP/OKD.
+
+### Non-Goals
+* Implementing support for ALB for Openshift Routes
```
Miciah, line 96:
```suggestion
* Implementing support for ALB for OpenShift Routes
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+the subnets and tag them or the user can do this. This is specified with the
+_AutoTagSubnet_ field.
+
+The user can also specify the Ingress class which the lb-controller will
+reconcile. This could default to **_“alb”_**. This value is required because if
+an ingress class value is not passed to the controller it attempts to reconcile
+all Ingress resources which don’t have an ingress class annotation or where the
+ingress class field is empty. Passing the value of “alb” does not restrict the
+number of ingress classes which the controller can reconcile. This is described
+further in the [Parallel operation of Openshift
```
Miciah, line 279:
```suggestion
further in the [Parallel operation of OpenShift
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+2. The _lb-controller_ has support for _Service_ type resources as
+well. When the annotation `service.beta.kubernetes.io/aws-load-balancer-type:
+"external"` is specified on a _Service_ resource of type _LoadBalancer_ the
+in-tree controller ignores this resource and the lb-controller instead
+provisions a Network Load Balancer with the correct configuration. Since we want
+to restrict the features supported by the controller to only Ingresses the
+lb-controller will have to be modified so that it does not reconcile any Service
+resources.
+
+#### Parallel operation  of the Openshift Router and lb-controller
```
Miciah, line 329:
```suggestion
#### Parallel operation  of the OpenShift router and lb-controller
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+## Implementation History
+TBD
+
+## Drawbacks
+
+TBD
+
+## Alternatives
+
+The existing Openshift Router could be modified to provision ALBs for Ingresses
```
Miciah, line 391:
```suggestion
The existing OpenShift router could be modified to provision ALBs for Ingresses
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,394 @@
...
+the existing addon attachment from the provisioned load balancers.
+
+The operator will also create a
+[CredentialRequest](https://docs.openshift.com/container-platform/4.9/rest_api/security_apis/credentialsrequest-cloudcredential-openshift-io-v1.html#credentialsrequest-cloudcredential-openshift-io-v1)
+on behalf of the lb-controller and mount the minted credentials in the
+_Deployment_ of the controller. This means that the operator will only work on
+OCP/OKD but this is sufficient for the initial release.
+
+
+### Risks and Mitigations
```
alebedev87, line 304: Just to track what we discussed in Slack. Hope it can be resolved, in that case we can just resolve this comment.

I was wondering where we can get `VPCID` needed for the OCP deployment as it doesn't allow the access to AWS metadata server although [it will be needed](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/deploy/installation/#using-metadata-server-version-2-imdsv2).
I could not find it in [Infrastructure resource](https://github.com/openshift/api/blob/8779abfbd5714ea73487699cbc6fc6e8e5f79158/config/v1/types_infrastructure.go#L340-L366). Other 2 things which will be needed by the operator are there though: [region](https://github.com/openshift/api/blob/8779abfbd5714ea73487699cbc6fc6e8e5f79158/config/v1/types_infrastructure.go#L351), [cluster name](https://github.com/openshift/api/blob/8779abfbd5714ea73487699cbc6fc6e8e5f79158/config/v1/types_infrastructure.go#L55).

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: Yes, it's out of scope. OCP doesn't offer it as an option so we cannot detect if it is enabled and then enable it in the controller.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: Added a section on the changes needed in the Installer.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: Both will be supported. @alebedev87 did some digging into this. When there are conflicting configurations or non-reconcilable configuration the controller errors out.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: > Is the API being designed to accommodate a future enhancement adding this capability in case it should be come possible upstream and users should request the capability?
Yes, when this becomes possible upstream users will be able to create multiple controllers. Right now we only reconcile one resourcce called `cluster`.

arjunrn: > does it make sense at all to have multiple instances running?

We don't restrict by namespace in the current implementation but the controller does support it. So I suppose when the multiple instance support does come we will have to support the namespace configuration as well. But we will have to ensure that there is no situation where multiple instances of the controller don't all try to reconcile the same Ingress resource due to overlap.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: In this case, the example is wrong and the user should create the `IngressClass` resource as well. But if the user specifies the ingress class through the annotation then it is not required. I suppose it works this way for older versions of Kubernetes which did not have the `IngressClass` resource. We have discussed disabling the annotation based specification completely.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: The bugs you mention are for NLBs but the Ingresses will always use ALBs which don't preserve the source.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: The cloud-credentials-operator will always be present in OCP/OKD cluster right? Because we are implementing the operator with the assumption that it is only for OpenShift.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: The controller is critical for production services. Users may want to use multiple replicas to ensure high availability.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: For the initial version of the operator it will not use the tags specified in the `Infrastructure` resource. Users will have to specify it explicitly.

enhancements/ingress/aws-load-balancer-operator.md
alebedev87: Almost all `IngressClassParams` settings take precedence over the annotations in `Ingress` resources: e.g. [group](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/guide/ingress/ingress_class/#specgroup), [ipAddressType](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/guide/ingress/ingress_class/#specipaddresstype). Conflicts may come only if ingresses with conflicting settings are added to the same [IngressGroup](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/guide/ingress/annotations/#ingressgroup) which may be done in `IngressClassParams` or in annotations. And yes, [errors are produced](https://github.com/openshift/enhancements/pull/1020/files#r795938025) in such cases.

alebedev87:
As discussed in Slack, LGTM.

enhancements/ingress/aws-load-balancer-operator.md
Miciah: > The cloud-credentials-operator will always be present in OCP/OKD cluster right? 

I think so.

> Because we are implementing the operator with the assumption that it is only for OpenShift.

Is that an assumption only for the initial implementation, or is there no intention to consider upstreaming this operator?  Either way, can you add this to the enhancement proposal somewhere, for example to the summary or as a non-goal?  (Unless you already said it and I forgot or missed it.)

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,415 @@
...
+
+#### Support Procedures
+TBD
+
+## Implementation History
+TBD
+
+## Drawbacks
+
+TBD
```
Miciah, line 408: For the "TBD" sections, do you plan to D them in a follow-up after this PR merges, or do you plan to D them in this PR?

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: This is just speculation but when multi-instance support does come it may be possible to have different instances reconciling Ingresses of the same Ingress class in different namespaces. We will have to be careful in how we allow the user to configure the instances so that there is no contention between the instances.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: I've clarified this in the _Motivation_ section. The operator will only work on OCP/OKD at least for the first release. However there have to be no API/design choices which prevent it from being extended to work on a Kubernetes compliant cluster.

enhancements/ingress/aws-load-balancer-operator.md
arjunrn: I've added details in some sections and will add more as we implement the operator. Some of the sections like Version Skew don't apply to this operator, so I've added an NA there.

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,433 @@
...
+[CredentialRequest](https://docs.openshift.com/container-platform/4.9/rest_api/security_apis/credentialsrequest-cloudcredential-openshift-io-v1.html#credentialsrequest-cloudcredential-openshift-io-v1)
+on behalf of the lb-controller and mount the minted credentials in the
+_Deployment_ of the controller. This means that the operator will only work on
+OCP/OKD but this is sufficient for the initial release.
+
+The lb-controller requires a validating and mutating webhook for correct operation. The operator will have to create the
+webhooks along with the controller deployment. The webhook can be registered with a CA bundle which is used to verify
+the identity of webhook by the API server.
+The [service-ca controller](https://docs.openshift.com/container-platform/4.9/security/certificate_types_descriptions/service-ca-certificates.html)
+can be used to generate certificates and have then injected into the webhook configurations. The operator will also
```
Miciah, line 329:
```suggestion
can be used to generate certificates and have them injected into the webhook configurations. The operator will also
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,433 @@
...
+### Upgrade / Downgrade Strategy
+NA
+
+### Version Skew Strategy
+NA
+
+### Operational Aspects of API Extensions
+
+The new Custom Resource will not affect any existing operations of the cluster. However the webhooks which have to be
+created for the lb-controller will impact the availability and performance and availability of Ingress resource _create_
```
Miciah, line 404:
```suggestion
created for the lb-controller will impact the availability and performance of Ingress resource _create_
```

enhancements/ingress/aws-load-balancer-operator.md
```diff
@@ -0,0 +1,433 @@
...
+and _update_ operations. These webhooks are light-weight and do not perform any complex validation. So the performance
+impact should be minimal. Availability of the webhook is critical because if the webhook is unavailable operations on
+the Ingress resource will not be possible.
+
+#### Failure Modes
+
+1. Operator is unable to determine the private and public subnets in the cluster. In this case the operator will be
+   marked as __Degraded__ in the status.
+2. lb-controller pods are not running. This would mean that the webhook are unavailable and all operations on _Ingress_
```
Miciah, line 413:
```suggestion
2. lb-controller pods are not running. This would mean that the webhook is unavailable and all operations on _Ingress_
```

Miciah (commented):
I noted a couple typos in the most recently added text, but this looks great overall.
/approve

PR #1023, opened 2022-02-02
NE-881: ingress: Add route-subdomain enhancement

Miciah:
This enhancement implements the `spec.subdomain` field of the OpenShift Route API.  When this field is specified, the host name of the Route depends on the domain of the IngressController that exposes the Route.  If multiple IngressControllers expose a Route that specifies `spec.subdomain`, then the Route has a distinct host name for each IngressController that exposes it.  This is particularly useful with sharding where the Route is exposed on multiple shards.  It is also useful generally in situations where the user wants to specify the subdomain part of the Route's host name but allow the IngressController to set the domain part.

* `enhancements/ingress/route-subdomain.md`: New file.

---

WIP implementation:

- [ ] https://github.com/openshift/openshift-apiserver/pull/254
- [ ] https://github.com/openshift/cluster-ingress-operator/pull/674
- [ ] https://github.com/openshift/router/pull/357
- [ ] https://github.com/openshift/origin/pull/27030

---

https://gist.github.com/Miciah/5fd6c88b45803dd3ebb55f63f28ca83e shows an example of how this new feature can be used to expose a route with two different routers and domains.

### DIFF
--- a/enhancements/ingress/route-subdomain.md
+++ b/enhancements/ingress/route-subdomain.md
@@ -1,0 +1,646 @@
+---
+title: route-subdomain
+authors:
+  - "@Miciah"
+reviewers:
+  - "@alebedev87"
+  - "@smarterclayton"
+  - "@gcs278"
+approvers:
+  - "@frobware"
+api-approvers:
+  - "@Miciah"
+creation-date: 2022-02-02
+last-updated: 2022-02-02
+tracking-link:
+  - "https://issues.redhat.com/browse/NE-700"
+see-also:
+replaces:
+superseded-by:
+---
+
+# Route Subdomain
+
+
+## Release Signoff Checklist
+
+- [X] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+This enhancement implements the `spec.subdomain` field of the OpenShift Route
+API.  When `spec.host` is specified or `spec.subdomain` is omitted on a Route,
+then the Route API behaves the same as before this enhancement.  When
+`spec.host` is omitted and `spec.subdomain` is specified, the host name of the
+Route depends on the domain of the IngressController that exposes the Route.  If
+multiple IngressControllers expose a Route that specifies `spec.subdomain`, then
+the Route has a distinct host name for each IngressController that exposes it.
+This is particularly useful with sharding where the Route is exposed on multiple
+shards.  It is also useful generally in situations where the user wants to
+specify the subdomain part of the Route's host name but allow the
+IngressController to set the domain part.
+
+## Motivation
+
+OpenShift router is designed to accommodate multiple router deployments running
+within a single cluster.  Within a cluster with multiple router deployments,
+each Route may be exposed by a single router deployment, or it may be exposed by
+multiple router deployments.  The former is useful for sharding, that is,
+partitioning Routes for isolation or performance.  The latter (having each
+Route exposed by multiple routers) is useful for redundancy or for enabling
+multiple ingress paths for the same application, such as having an internal
+router deployment for in-cluster clients and an externally exposed router
+deployment for clients outside the cluster.
+
+The absence of support for the `spec.subdomain` API complicates both use-cases.
+Without support for `spec.subdomain`, the `spec.host` field must be specified.
+If the user does not specify a value when the Route is created, the API fills in
+a default value using the cluster ingress domain.  Having `spec.host` default to
+the cluster ingress domain complicates sharding because it puts the onus on the
+user to specify `spec.host` with a domain that matches the domain of the shard
+to which the Route will belong.  The defaulting also complicates the use-case
+of having the same Route exposed on multiple router deployments because the
+Route's host will match the domain of only one of those router deployments.
+
+The `spec.subdomain` field solves this problem by allowing `spec.host` to be
+omitted and a subdomain to be specified.  Then, each router deployment that
+exposes the Route constructs a host name from the subdomain specified in
+`spec.subdomain` and the router deployment's domain.  The router deployment
+reports this host name in the Route's status: each router deployment adds an
+entry to `status.ingress[]` with the host name that the router deployment has
+assigned to the Route.
+
+By implementing the `spec.subdomain` field, it will be possible for a Route to
+have one or more host names determined by the router deployment that exposes the
+Route, thus simplifying sharding and other use-cases.
+
+### Goals
+
+* An application administrator can create a Route that is exposed by one or more
+  IngressControllers, and each of the IngressControllers will expose the Route
+  under the IngressController's domain.
+* An application administrator can create a Route with a so-called "vanity"
+  (custom) subdomain without needing to specify a domain.
+* When a user creates a Route with a nonempty value for `spec.subdomain`, the
+  API does not set a value for the Route's `spec.host`.
+* When an IngressController admits a Route with an empty value for `spec.host`
+  and a nonempty value for `spec.subdomain`, the IngressController exposes the
+  Route using a host name comprising the Route's subdomain and the
+  IngressController's domain.
+
+### Non-Goals
+
+* This enhancement does not configure labels or label selectors for sharding.
+* This enhancement does not add any new API fields.
+* This enhancement does not provide any templating mechanism for host names.
+* This enhancement does not allow one Route to have multiple host names for the
+  same router deployment.
+
+## Proposal
+
+Years before this enhancement was written, the `spec.subdomain` field was added
+to the Route API but left unimplemented.  This enhancement implements support
+for `spec.subdomain` in the OpenShift API server and OpenShift router so that
+users can specify `spec.subdomain` instead of specifying `spec.host` to get a
+Route that assumes the particular domain of any and all IngressControllers that
+expose it.  For example, if a router with the domain "bar.tld" admits a Route
+with subdomain "foo", the router exposes the Route using the host name
+"foo.bar.tld", and if a second router with the domain "baz.tld" admits the same
+Route, then this second router exposes the Route using the host name
+"foo.baz.tld".
+
+With this enhancement, the OpenShift API server is modified not to set a default
+value for `spec.host` if `spec.subdomain` is specified, and the ingress operator
+is modified to configure router deployments to use the IngressController's
+`status.domain` as the router deployment's domain.  OpenShift router is modified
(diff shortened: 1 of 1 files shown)


enhancements/ingress/route-subdomain.md
```diff
@@ -0,0 +1,593 @@
...
+  ingress:
+  - conditions:
+    - lastTransitionTime: # ...
+      status: "True"
+      type: Admitted
+    host: hello.shard1.apps.mycluster.com
+    routerCanonicalHostname: router-shard1.shard1.apps.mycluster.com
+    routerName: shard1
+    # ...
+```
```
smarterclayton, line 313: Technically console wanted this, we didn't deliver it, and as a result the default console DNS name in openshift ootb sucks (console-openshift-console).  While we can't necessarily change that, the use case of cluster agnostic (i.e. unaware of the subname) naming is valid (i want to create a name within the route shard's scope) as an infrastructure user.

enhancements/ingress/route-subdomain.md
```diff
@@ -0,0 +1,593 @@
...
+character and contain only alphanumeric characters, possibly with non-initial
+and non-final dashes.  Previously, the OpenShift API server performed no
+validation on the field; however, because the field was unused before this
+enhancement, the stricter validation can be justified as not breaking any
+previously supported use-case; cf. "Risks and Mitigations" below.
+
+The OpenShift API server is also modified not to set a default value for
+`spec.host` if `spec.subdomain` has a nonempty value.  This change enables each
+router deployment to set a host, which is reported in the Route's
+`status.ingress[*].host` field.
```
alebedev87, line 130: Not quite clear which host the router will set. Or you mean the `status.ingress.host`?

alebedev87 (commented):
LGTM, just a couple of clarification questions.

alebedev87:
As discussed in Slack.
Do we need to check the impact on `oc` commands related to the route like `oc describe route`, `oc get route` ?

enhancements/ingress/route-subdomain.md
```diff
@@ -0,0 +1,593 @@
...
+* This enhancement does not provide any templating mechanism for host names.
+* This enhancement does not allow one Route to have multiple host names for the
+  same router deployment.
+
+## Proposal
+
+Years before this enhancement was written, the `spec.subdomain` field was added
+to the Route API but left unimplemented.  This enhancement implements support
+for `spec.subdomain` in the OpenShift API server and OpenShift router so that so
+that users can specify `spec.subdomain` instead of specifying `spec.host` to get
```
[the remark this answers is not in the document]
Miciah: Thanks!  Fixed in the next push.

enhancements/ingress/route-subdomain.md
```diff
@@ -0,0 +1,593 @@
...
+a Route that assumes the particular domain of any and all IngressControllers
+that expose it.
+
+Specifically, the OpenShift API server is modified not to set a default value
+for `spec.host` if `spec.subdomain` is specified, and the ingress operator is
+modified to configure router deployments to use the IngressController's
+`status.domain` as the router deployment's domain.
```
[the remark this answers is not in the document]
Miciah: I've added some examples to help clarify in the next push.

enhancements/ingress/route-subdomain.md
Miciah: A router named "foo" will set the route's `status.ingress["foo"].host` field.  If it will help, I can add the previous sentence to the enhancement document.  Is the `[*]` syntax confusing, is the wording confusing, or something else?

Miciah:
Latest push (1) adds some examples of how routes' hosts would be reflected in status, (2) adds open questions regarding `oc` output and using `spec.subdomain` in core platform routes, and (3) adds an alternative that proposes defaulting `spec.subdomain` when `spec.host` is unset

Miciah:
Added a sentence in the summary to clarify that this enhancement does not change the behavior if `spec.host` is specified.

enhancements/ingress/route-subdomain.md
alebedev87: Now it's clear, I believe that before you didn't mention that the router updates the status so it could have been interpreted as if the router sets `.spec.host`.

enhancements/ingress/route-subdomain.md
```diff
@@ -0,0 +1,642 @@
...
+other factors (cf. [RFE-1467](https://issues.redhat.com/browse/RFE-1467) for an
+exploration of this idea).  Disadvantages of this approach include the
+additional development and operations effort required to implement and manage
+such a controller as well as the fundamental limitation that a Route could still
+only have a single host name and thus could not be simultaneously be exposed by
+multiple shards with different domains.
+
+### Default `spec.subdomain` instead of defaulting `spec.host`
+
+A possible modification to this enhancement would be to change the OpenShift API
```
gcs278, line 622: So my hot take is: I think this alternative is the most correct, as it paves a way to a single domain-specifier solution. Here's my arguments:
1. As a user, the relationship between `spec.host` and `spec.subdomain` (without reading the docs), is not natural to me like I mentioned in an comment above.
2. I just generally think the existing `spec.host` is a little redundant that you have to copy/paste the full domain into the route to match your ingress controller (hope you don't have copy/paste error). I'd think just `spec.subdomain` is a easier way to configure your domains.

So, I'm still new to enhancements and I'm not out-right disagreeing with this one, but I am curious if there is agreement that there is some discontinuity between the two, and if you really actually need both `spec.host` and `spec.subdomain`. 

The questions I have:
- In a perfect world, do we agree that we could really get by with just `spec.subdomain`?
- If we agree above, What's the transition plan to consolidate the two specifiers, if any?

gcs278 (commented):
Overall great and well written enhancement, but I have questions about how this evolves in the future.

enhancements/ingress/route-subdomain.md
```diff
@@ -0,0 +1,642 @@
...
+2. Create four Routes:
+   - One Route that omits both `spec.host` and `spec.subdomain`.
+   - One Route that omits `spec.host` and specifies `spec.subdomain`.
+   - One Route that specifies `spec.host` and omits `spec.subdomain`.
+   - One Route that specifies both `spec.host` and `spec.subdomain`.
+3. Verify that each router deployment admits each Route and puts the expected host in the Route's status.
+   - For a Route that omits both `spec.host` and `spec.subdomain`, the API should set a default value for `spec.host`, and the router should expose the Route using that host name.
+   - For a Route that specifies `spec.host`, the router should expose the Route using that host name.
+   - For a Route that omits `spec.host` and specifies `spec.subdomain`, the router should leave `spec.host` empty, and the router should expose the Route using a host name composed from the Route's subdomain and the router's domain.
+
```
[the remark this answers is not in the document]
Miciah: The second case covers that.  I'll make it more explicit though.

enhancements/ingress/route-subdomain.md
```diff
@@ -0,0 +1,642 @@
...
+Users and admins can also proactively check for instantiated Routes that specify
+`spec.subdomain` on a running cluster:
+
+```shell
+oc get routes --output=go-template --template='{{range .items}}{{if .spec.subdomain}}{{printf "%s %s %s\n" .metadata.namespace .metadata.name .spec.subdomain}}{{end}}{{end}}' --all-namespaces
+```
+
+However, per the Route API's existing semantics, a Route object that has already
+had `spec.host` defaulted on an older OpenShift cluster will continue to be
+served using the same host name, and `spec.subdomain` will be ignored; the
```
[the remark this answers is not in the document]
Miciah: I understand the API is a little confusing.  Unfortunately that arises from the necessity to maintain backwards compatibility.  Moreover the `spec.subdomain` field and these semantics have actually been in openshift/api for many releases—we just haven't implemented the documented semantics—so I'm reluctant to try to change the semantics now.  

It might be worth exploring the possibility of printing a warning when both `spec.host` and `spec.subdomain` are set.  I can look into what would be required to do that.  Maybe (although I'm more dubious of this possibility) we could even add an admission check that rejects _new_ routes (whilst allowing updates to old routes, for compatibility) that specify both `spec.host` and `spec.subdomain`.

enhancements/ingress/route-subdomain.md
Miciah: > So my hot take is: I think this alternative is the most correct, as it paves a way to a single domain-specifier solution. Here's my arguments:
> 
>     1. As a user, the relationship between `spec.host` and `spec.subdomain` (without reading the docs), is not natural to me like I mentioned in an comment above.

Agreed.  

>     2. I just generally think the existing `spec.host` is a little redundant that you have to copy/paste the full domain into the route to match your ingress controller (hope you don't have copy/paste error). I'd think just `spec.subdomain` is a easier way to configure your domains.

Disagree.  See below.  

> So, I'm still new to enhancements and I'm not out-right disagreeing with this one, but I am curious if there is agreement that there is some discontinuity between the two, and if you really actually need both `spec.host` and `spec.subdomain`.
> 
>     * In a perfect world, do we agree that we could really get by with just `spec.subdomain`?

No, the user may need to specify a host name outside of the ingress controller's domain.  It would be up to the user to create a DNS record pointing that custom host name to the ingress controller (typically, the user would use a CNAME record pointing to the `routerCanonicalHostname` from the route's status), but notwithstanding that, use cases do exist for specifying `spec.host` instead of `spec.subdomain`.  

The idea with this alternative is really that the user would still be able to specify `spec.host` or `spec.subdomain`, but if the user specified _neither_, the API would default `spec.subdomain`, rather than defaulting `spec.host`.  The next paragraph describes the tradeoff, which is, in short, that defaulting `spec.subdomain` instead of defaulting `spec.host` would be surprising and could break users' long-established workflows.

enhancements/ingress/route-subdomain.md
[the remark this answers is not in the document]
gcs278: Oh I see, my brain went "4 tests and 3 results....hmmm"

enhancements/ingress/route-subdomain.md
gcs278: Okay I see your point, I was unaware of specific use cases with `spec.host`, but that makes sense now. I generally agree with this position.

I still think the domain-related API is getting a little confusing, which I think you agree with. I'm curious, if we had a chance to redo it, what would be a more natural solution. That can be a future discussion.

Miciah:
Latest push adds a tracking link and adds myself to api-approvers to appease the linter (the enhancement implements an existing API, so there's no API approval needed) as well as updating the reviewers list.

enhancements/ingress/route-subdomain.md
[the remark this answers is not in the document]
Miciah: I added a warning in openshift-apiserver; see <https://github.com/openshift/openshift-apiserver/pull/254#issuecomment-1108076241>.

PR #1024, opened 2022-02-02
Change openshift-managed to OpenShift managed

sebsoto:


aravindhp (approved):
Thanks for updating this, @sebsoto 

/approve

PR #1025, opened 2022-02-04
Introduce detection of stale clusteroperator conditions

mfojtik:
This is a proposal to improve the correctness of clusteroperator conditions in cases when operator processes are stuck or not running.

### DIFF
--- a/enhancements/kube-apiserver/stale-clusteroperators-detection.md
+++ b/enhancements/kube-apiserver/stale-clusteroperators-detection.md
@@ -1,0 +1,108 @@
+---
+title: detection-and-reporting-of-stale-clusteroperators
+authors:
+  - "@mfojtik"
+reviewers:
+  - "@deads2k"
+approvers:
+  - "@deads2k"
+creation-date: 2022-01-31
+last-updated: 2022-01-31
+---
+
+# Detection and reporting of stale cluster operators
+
+## Summary
+
+In OpenShift 4, every active operator included in the core payload (cluster operator) updates its conditions in "
+clusteroperator" resource. Conditions determine whether specific core components are Degraded or Available or if they
+are making progress or if their state causes the cluster to be not upgradeable. These conditions are essential for
+cluster administrators to determine the health of their clusters, and they are usually the first thing Red Hat support
+is going to look at in case there is a problem with the cluster.
+
+Unfortunately, in some edge cases, these conditions might not reflect the current state of the cluster. For example,
+since they depend on the running operator, these conditions might become stale and outdated if the operator is not
+running or cannot contact the API server. And while the API server might still work, the result
+of `oc get clusteroperators` does not represent the cluster's current state.
+
+Therefore, we need a detection mechanism to help detect in-active operators or operators that are not actively updating
+the conditions.
+
+## Motivation
+
+In the past, there were multiple cases when either kubelet or CRI-O failed to run pods, either by KCM malfunctioning or
+some other system error. However, the static pods succeeded in running, making the Kubernetes API server available and
+accessible with `oc`. When support or engineers ran `oc get clusteroperators` against this cluster, they saw all
+operators reporting available, non-degraded and upgradeable status, but the truth was that none of the operators that
+provided these conditions were running.
+
+### Goals
+
+* Detect inactive cluster operators
+* Provide more accurate condition messages and status
+
+### Non-Goals
+
+* Make a new operator that detects other operators status
+* Create a heart-beat mechanism
+
+## Proposal
+
+To implement stale operator detection, we need a controller process that will run close to the Kubernetes API server,
+preferably using a local connection and client-based certificate for authentication to avoid dependencies on working
+auth and network when the cluster is unstable.
+
+The controller will watch the clusteroperator resource and check the `status.[]conditions.lastTransitionTime` field for
+Degraded, Available, Progressing, and Upgradeable conditions. Suppose the "lastTransitionTime" field is not updated for
+10 minutes. In that case, the controller will change the human-readable condition message
+to: `Operator checking for stale status, the active operator will reset this message: <original message>`.
+
+After the message is changed, as it suggests, the active operators should reset this message back to the original
+message, removing the "checking" part. This shows a healthy, active operator.
+
+Suppose the operator is not active and does not reset the message back to the original after 10 minutes. In that case,
+the controller will change the message again
+to: `Operator has not updated this condition for more than 20 minutes, last known condition state was "<False/True>", original message: <original message>`
+. In addition to changing the message, this time, the controller will also change the condition status from False or
+True to **Unknown**.
+
+Since the clusteroperator resource has multiple conditions in status, if **any** of these conditions are updated, we assume
+the operator is still active and able to reflect the operand status into clusteroperator resource.
+
+For cluster administrators, this means the operator is having a problem and is potentially down or unable to make
+updates to conditions. In that case, an alert might be required to trigger a manual investigation of the current
+operator state.
+
+### API Extensions
+
+*No API extensions required*
+
+### Risks and Mitigations
+
+* One risk here is that we find operators that are not correctly reconciling their clusteroperator conditions, which this controller makes obvious. These operators should be fixed as a result.
+* This also creates (2 writes) * (number of cluster operators) every N minutes
+
+## Design Details
+
+### Test Plan
+
+* This feature should be pared with an e2e invariant that ensures a minimum update time for the stale message. Let's take a guess at 30s to start and see how many fail.
+
+### Operational Aspects of API Extensions
+
+No API extension is needed to implement this proposal.
+
+#### Failure Modes
+
+#### Support Procedures
+
+## Implementation History
+
+## Drawbacks
+
+## Alternatives
+
+One alternative discussed was a new API field that will be added to clusteroperators.Status called "lastUpdateTime".
+Instead of relying on
+"lastTransitionTime" will introduce this field and use it as "heart-beat". This will require **all** operators to update
+this field correctly and multiply the writes significantly.


enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Proposal
+
+To implement stale operator detection, we need a controller process that will run close to the Kubernetes API server,
+preferably using a local connection and client-based certificate for authentication to avoid dependencies on working
+auth and network when the cluster is unstable.
+
+The controller will watch the clusteroperator resource and check the `status.[]conditions.lastTransitionTime` field for
+Degraded, Available, Progressing, and Upgradeable conditions. Suppose the "lastTransitionTime" field is not updated for
+10 minutes. In that case, the controller will change the human-readable condition message
```
zaneb, line 57: `lastTransitionTime` is normally only [updated when the status transitions](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go#L43-L50) from once value to another. Futhermore, we probably won't want to change that because it is an important source of information - if the operator is up, this field tells you how long it has been up.

I assume that most operators will update the message again, but using the lastTransitionTime to remember state about when you last checked is problematic.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,87 @@
...
+since they depend on the running operator, these conditions might become stale and outdated if the operator is not
+running or cannot contact the API server. And while the API server might still work, the result
+of `oc get clusteroperators` does not represent the cluster's current state.
+
+Therefore, we need a detection mechanism to help detect in-active operators or operators that are not actively updating
+the conditions.
+
+## Motivation
+
+In the past, there were multiple cases when either kubelet or CRI-O failed to run pods, either by KCM misfunctioning or
```
deads2k, line 33: malfunctioning

enhancements/kube-apiserver/stale-clusteroperators-detection.md
deads2k: > I assume that most operators will update the message again, but using the lastTransitionTime to remember state about when you last checked is problematic.

it's not perfect, but it's not problematic.  Since this a non-destructive write that does not change the condition.status, the human readable message tweak is sufficient when combined with an in-memory recording of, "I touched these".  That will limit the number of writes while providing the indications.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,87 @@
...
+After the message is changed, as it suggests, the active operators should reset this message back to the original
+message, removing the "checking" part. This shows a healthy, active operator.
+
+Suppose the operator is not active and does not reset the message back to the original after 10 minutes. In that case,
+the controller will change the message again
+to: `Operator has not updated this condition for more than 20 minutes, last known condition state was "<False/True>", original message: <original message>`
+. In addition to changing the message, this time, the controller will also change the condition status from False or
+True to **Unknown**.
+
+For cluster administrators, this means the operator is having a problem and is potentially down or unable to make
```
deads2k, line 69: This feature should be pared with an e2e invariant that ensures a minimum update time for the stale message.  Let's take a guess at 30s to start and see how many fail.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,87 @@
...
+### Risks and Mitigations
+
+One risk here is that we find operators that are not correctly reconciling their clusteroperator conditions, which this
+controller makes obvious. These operators should be fixed as a result.
+
+## Alternatives
+
+One alternative discussed was a new API field that will be added to clusteroperators.Status called "lastUpdateTime".
+Instead of relying on
+"lastTransitionTime" will introduce this field and use it as "heart-beat". This will require **all** operators to update
```
deads2k, line 86: this approach would also multiply writes significantly.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
zaneb: I agree that an in-memory recording of the last checked times would work (it could even be made a little more robust by adding a timestamp to the message it sets). 

As currently written though, this implies that it will look at the `lastTransitionTime` and that won't work - if this new controller doesn't change the time, then the operator will set it back immediately to the original state and you will go immediately into an inifinite hot loop. And I don't think we want the new controller to modify the `lastTransitionTime` because that changes the meaning of the field.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,87 @@
...
+## Proposal
+
+To implement stale operator detection, we need a controller process that will run close to the Kubernetes API server,
+preferably using a local connection and client-based certificate for authentication to avoid dependencies on working
+auth and network when the cluster is unstable.
+
+The controller will watch the clusteroperator resource and check the `status.[]conditions.lastTransitionTime` field for
+Degraded, Available, Progressing, and Upgradeable conditions. Suppose the "lastTransitionTime" field is not updated for
+10 minutes. In that case, the controller will change the human-readable condition message
+to: `Operator checking for stale status, the active operator will reset this message: <original message>`.
```
bparees, line 58: for a healthy operator that's not experiencing any changes, aren't we going to be asserting this message every 10 minutes and then waiting for it to be reset back?

I guess the reset back will be quick (the operator should immediately detect the change and reassert the value) but it's definitely new churn.

it also loses the ability to know when the operator *really* last changed a condition(e.g. when did it enter progressing?)

enhancements/kube-apiserver/stale-clusteroperators-detection.md
bparees: more significantly than the proposed mechanism?  in either case isn't a healthy operator going to be doing a write to clusteroperator every 10 minutes?  (either to update the heartbeat, or to re-assert "no i'm not stale")

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,87 @@
...
+* Create a heart-beat mechanism
+
+## Proposal
+
+To implement stale operator detection, we need a controller process that will run close to the Kubernetes API server,
+preferably using a local connection and client-based certificate for authentication to avoid dependencies on working
+auth and network when the cluster is unstable.
+
+The controller will watch the clusteroperator resource and check the `status.[]conditions.lastTransitionTime` field for
+Degraded, Available, Progressing, and Upgradeable conditions. Suppose the "lastTransitionTime" field is not updated for
```
bparees, line 56: it's not clearly stated here, so to avoid any implementation confusion:  I would expect that this behavior would only be triggered if at least one of the above conditions is "stale".  As long as any one condition was updated in the last 10 mins, then no write is needed.  (vs treating each condition independently)

enhancements/kube-apiserver/stale-clusteroperators-detection.md
bparees: (the message will be reassert in short order, ideally, the lastTransitionTime will not be restored)

bparees:
@mfojtik here's an ugly edge case to this proposal:

1) operator goes degraded/unavailable.  updates status message accordingly to warn the admin
2) operator then fails completely
3) 10 mins go by
4) this logic wipes out the the status message from (1) that told the admin what was going wrong
5) admin comes along and all they see is "this operator went stale and we erased all its status info for you, you're welcome"

might be better to preserve the existing messages, but append them with a "btw this might be stale" message.

mfojtik:
We preserving the message in both cases. If operator is totally down, it has no idea if the operand is still down or it actually recovered on its own ;-) So the status is rightfully "unknown" which means "we don't know".On Mon, Feb 7, 2022 at 18:43, Ben Parees ***@***.***> wrote:  
@mfojtik here's an ugly edge case to this proposal:

operator goes degraded/unavailable.  updates status message accordingly to warn the admin
operator then fails completely
10 mins go by
this logic wipes out the the status message from (1) that told the admin what was going wrong
admin comes along and all they see is "this operator went stale and we erased all its status info for you, you're welcome"

might be better to preserve the existing messages, but append them with a "btw this might be stale" message.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
mfojtik: good catch

enhancements/kube-apiserver/stale-clusteroperators-detection.md
mfojtik: @deads2k i realize... should operators change lastTransitionTime when they are not transitioning states, but only update message?

enhancements/kube-apiserver/stale-clusteroperators-detection.md
deads2k: > how is it non-destructive? it destroys the current lastTransitionTime and the current message associated w/ the condition

It should not.  Changing a message does not change the status and the lastTransitionTime is about status, not the message.  If an operator is reseting lastTransitionTime based on the message change, then that operator is bugged: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1497

bparees: got it, this is resolved from my perspective, dunno if @zaneb still has concerns

enhancements/kube-apiserver/stale-clusteroperators-detection.md
bparees: covered above, lastTransitionTime should not change just because the message is updated.

i still have a small concern about churning the messages which may look strange in any sort of dashboard that's reporting the message strings.  It's not a showstopper, but it doesn't feel like an ideal UX for the end-user.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
zaneb: No concerns with what is being discussed in the comments here (store last checked time in-memory; change message at 10-minute intervals unless lastTransitionTime is <10 minutes ago). Slight concern that the actual text of the enhancement could be interpreted to mean something else.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,108 @@
...
+
+In OpenShift 4, every active operator included in the core payload (cluster operator) updates its conditions in "
+clusteroperator" resource. Conditions determine whether specific core components are Degraded or Available or if they
+are making progress or if their state causes the cluster to be not upgradeable. These conditions are essential for
+cluster administrators to determine the health of their clusters, and they are usually the first thing Red Hat support
+is going to look at in case there is a problem with the cluster.
+
+Unfortunately, in some edge cases, these conditions might not reflect the current state of the cluster. For example,
+since they depend on the running operator, these conditions might become stale and outdated if the operator is not
+running or cannot contact the API server. And while the API server might still work, the result
```
wking, line 25: ClusterOperator is primarily an API for operators to report to the cluster-version operator.  The CVO will expose concerning issues like `Available=False` or `Degraded=True` to admins in ClusterVersion.  But the CVO also manages the operator Deployments, so in the case where the operator is actually not running, the deployment controller should be reporting some sad condition, and the CVO should be complaining about that sad deployment in ClusterVersion's `Failing`.  Admins that reach around the CVO and inspect ClusterOperator directly may not be aware of the Deployment context, but :shrug: that's what you get when you imperfectly break through an abstraction layer :wink:.  And there are some cases today, when the CVO is stuck mid-update on some earlier manifest, where even the CVO stops caring about Deployments and ClusterOperators that are later in the update order (at least as far as `Failing` reporting goes).  But I think all of that either works today, or could work tomorrow if we are concerned about it.

The remaining issue, which this `message`-poking approach could help with, is where the operator deployment is up and happy, but the loop in charge of reporting status conditions has wedged.  But not clear to me that `message`-poking is a better way to detect that vs. a liveness probe or similar.  And the benefit of using a liveness probe is that if the operator locks up, Kube will cycle the container for you to help recover, instead of just complaining about the lockup.

Then there's also a possibility that the boring liveness service keeps working, but the more complicated status sync loop locks up.  Maybe `message`-poking is the best way at that?  But by this point, we're down in what seems like a pretty tiny corner, and we may be fine just living with the risk of the occasional partially-locked-operator bug.

deads2k:
there is now an impl at https://github.com/openshift/cluster-kube-apiserver-operator/pull/1423

enhancements/kube-apiserver/stale-clusteroperators-detection.md
bparees: @deads2k @mfojtik this thread never really got resolved, is this the behavior that's being baked into the proposed implementation?  flapping status messages?

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,108 @@
...
+to: `Operator checking for stale status, the active operator will reset this message: <original message>`.
+
+After the message is changed, as it suggests, the active operators should reset this message back to the original
+message, removing the "checking" part. This shows a healthy, active operator.
+
+Suppose the operator is not active and does not reset the message back to the original after 10 minutes. In that case,
+the controller will change the message again
+to: `Operator has not updated this condition for more than 20 minutes, last known condition state was "<False/True>", original message: <original message>`
+. In addition to changing the message, this time, the controller will also change the condition status from False or
+True to **Unknown**.
```
bparees, line 67: @deads2k your slack message says we'd set the operator to "A=unknown, D=true, P=unknown".  Assuming that's the plan, can we get the EP updated to reflect that behavioral mapping?  (And also what it would do for the upgrading condition).

Also i would posit that this new logic should *not* set Available=unknown if Available is currently set to False.  Though i can see arguments for why unknown might be more appropriate, reducing the severity of the status(I would argue unavailable is more severe than unknown, in terms of how admins are likely to treat it/have alerts configured) doesn't seem appropriate to me.  And if the operator isn't updating its status, it seems pretty unlikely that it is actually Available.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,108 @@
...
+## Implementation History
+
+## Drawbacks
+
+## Alternatives
+
+One alternative discussed was a new API field that will be added to clusteroperators.Status called "lastUpdateTime".
+Instead of relying on
+"lastTransitionTime" will introduce this field and use it as "heart-beat". This will require **all** operators to update
+this field correctly and multiply the writes significantly.
```
bparees, line 108: not clear to me why this would increase the number of writes?  In fact i think it would potentially reduce the number of writes.

the current proposal requires 2x writes every 10 minutes, per operator (one to set the message as potentially stale, another to set it back to the correct value)

a new field would only require that each operator poke the field every 10 minutes, so 1x write every 10 minutes (the heartbeat controller would have to watch all the operators, so potentially 1x read every 10 minutes)

but yes, it would require all operators to add new logic to maintain the new field.

Note that this alternate approach would also have the benefit of eliminating the "Flapping status message" behavior that i believe is built into the proposed approach.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
deads2k: > @deads2k @mfojtik this thread never really got resolved, is this the behavior that's being baked into the proposed implementation? flapping status messages?

Yes, the message gets a prefix "Checking for stale status, the active operator will reset this message: " and the existing message.  The Reason, Status, and LastTransitionTime do not change.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
deads2k: Conditions are standardized across in all of kube in apimachinery metav1.  Consumers of messages expect the schemas to match.  Our previous heartbeats were set at the nyquist frequency, so it was always 2x the speed of your desired detection limit..

Keeping the schemas consistent for external handling appears more important than having a stable message.

bparees: > Keeping the schemas consistent for external handling appears more important than having a stable message.

This seems to be a subjective value judgement.  Appears more important based on what, exactly?  Even taken on its face, having a core consistent subset of conditions that all external handlers can expect and must understand seems valuable, but claiming no extensions can be made that external handlers are free to ignore seems a little more difficult to defend.

But also, this alternative proposal doesn't suggest a new condition anyway, it suggests a new status field entirely.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,108 @@
...
+### Non-Goals
+
+* Make a new operator that detects other operators status
+* Create a heart-beat mechanism
+
+## Proposal
+
+To implement stale operator detection, we need a controller process that will run close to the Kubernetes API server,
+preferably using a local connection and client-based certificate for authentication to avoid dependencies on working
+auth and network when the cluster is unstable.
```
tkashem, line 53: > we need a controller process that will run close to the Kubernetes API server
Can you help me understand the goal better - it feels like this controller can be a test that runs in CI only (it does not need to run in production cluster)
The feature it is testing is - operators should write to `status.[]conditions`

The controller can be a test that goes through all `status.[]conditions` with a very aggressive interval, let's say `3m` and sets the message to `[stale marker] + {original message}` and if the message does not get reverted to its original form we can fail the test on this operator

This test establishes the invariant: each `co` backed operator writes to all `status.[]conditions`.

Now, if we see stale conditions in a real life cluster, we don't have to guess whether the operator is buggy in writing to `status.[]conditions`. the staleness is an indication of an underlying condition, like api connectivity.

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Goals
+
+* Detect inactive cluster operators
+* Provide more accurate condition messages and status
+
+### Non-Goals
+
+* Make a new operator that detects other operators status
+* Create a heart-beat mechanism
```
tkashem, line 47: I think this is equivalent to a heart-beat mechanism, the operator will revert the message every `N` minutes to prove to the controller that it's healthy

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,108 @@
...
+10 minutes. In that case, the controller will change the human-readable condition message
+to: `Operator checking for stale status, the active operator will reset this message: <original message>`.
+
+After the message is changed, as it suggests, the active operators should reset this message back to the original
+message, removing the "checking" part. This shows a healthy, active operator.
+
+Suppose the operator is not active and does not reset the message back to the original after 10 minutes. In that case,
+the controller will change the message again
+to: `Operator has not updated this condition for more than 20 minutes, last known condition state was "<False/True>", original message: <original message>`
+. In addition to changing the message, this time, the controller will also change the condition status from False or
```
tkashem, line 66: if we decide to take this controller to production, the flapping messages are an inconvenience to the cluster operators.
I know we can't add new fields to `Condition`, but we could introduce new fields directly under `Status`, like `lastUpdatedTime` which the operator will need to update every `10m` or something - this has the same runtime overhead as flapping the message.

Only trade off is each operator needs to incorporate this heart-beat pattern into its logic, which we could make less of a maintenance burden if every operator in question uses a shared `co` status updater. 
I think it's reasonable to accept this trade-off of additional maintenance burden on the operator in order to give cluster admins a better experience

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Proposal
+
+To implement stale operator detection, we need a controller process that will run close to the Kubernetes API server,
+preferably using a local connection and client-based certificate for authentication to avoid dependencies on working
+auth and network when the cluster is unstable.
+
+The controller will watch the clusteroperator resource and check the `status.[]conditions.lastTransitionTime` field for
+Degraded, Available, Progressing, and Upgradeable conditions. Suppose the "lastTransitionTime" field is not updated for
+10 minutes. In that case, the controller will change the human-readable condition message
```
tkashem, line 57: >  Suppose the "lastTransitionTime" field is not updated for 10 minutes

aren't we changing the semantics of `lastTransitionTime` here if the operator changes it due to an artificial (flapping) reason, it no longer represents the original transition time. Or are you thinking that the controller will change the message every `10m`, irrespective of `lastTransitionTime`? 
and when the operator resets the Condition we don't expect it to change `lastTransitionTime`, right?

tkashem (commented):
In general, how do we assert that our operands are not stale, for example, for kube-apiserver, etcd, scheduler? For these, we look at:
- tail the logs to see if the operand is active
- each operand has health check that kubelet probes
- we can look at metrics to see if they are active

these `co` operators have metrics, health checks - have we thought about a solution that utilizes metrics/health check to warn us if they are stale? For example, these operators could implement `component-base` metrics and they will readily provide client-go metrics and we can tell if they are stale by looking at their API request latency and error rate. 

we could build alerts based on these and have the alerts included in must-gather artifacts, thoughts?

enhancements/kube-apiserver/stale-clusteroperators-detection.md
bparees: > Now, if we see stale conditions in a real life cluster, we don't have to guess whether the operator is buggy in writing to status.[]conditions. the staleness is an indication of an underlying condition, like api connectivity.

the problem is how do we recognize stale conditions in a real life cluster?  If the condition is set to "available=true" and never changes, is that accurate, or is that a stale condition due to a buggy/broken/failed operator?

CI detection isn't enough because we're not trying to prove that "in the happy path/normal case, an operator updates its conditions".  We're trying to detect scenarios where the operator has stopped doing so for unanticipated reasons.

(put another way, the operator could still be buggy, in a way that CI does not expose)

enhancements/kube-apiserver/stale-clusteroperators-detection.md
benluddy: How averse are we to requiring updates to existing cluster operators as part of this proposal?

Without changing the API, or even introducing a new condition type, I could imagine the staleness controller comparing the ObservedGeneration of conditions and probing via something that causes an increment to metadata.generation, say, an inert annotation (set to the metadata.generation observed by the staleness controller at the time of the probe?).

enhancements/kube-apiserver/stale-clusteroperators-detection.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Support Procedures
+
+## Implementation History
+
+## Drawbacks
+
+## Alternatives
+
+One alternative discussed was a new API field that will be added to clusteroperators.Status called "lastUpdateTime".
```
benluddy, line 105: Even the message prefix approach requires an API change: https://github.com/openshift/api/blob/23ce2770bc0e607f7a05a4681542ae593c61b03d/config/v1/types_cluster_operator.go#L136.

:)

enhancements/kube-apiserver/stale-clusteroperators-detection.md
deads2k: > Only trade off is each operator needs to incorporate this heart-beat pattern into its logic, which we could make less of a maintenance burden if every operator in question uses a shared co status updater.
I think it's reasonable to accept this trade-off of additional maintenance burden on the operator in order to give cluster admins a better experience

This is acceptable to me, but will take a significant effort to push.  Since the apiserver team often fields these problems due to every operator reporting available, I'll leave the choice to you
1. flap the message in a non-destructive way and address the problem immediately
2. take ownership of the feature, add a new field, update every operator manage that new field, then write a controller to use this field.

I have no preference, but I do offer a complete solution, mergeable today for option 1.

deads2k: I'll throw out another possibility if someone wanted to build a cleaner thing: operators should have unexpired leases.  Given a relatedResource in status that is a lease, a clever controller could go from clusteroperator->relatedresource->lease->expired or not.  that would be peer to the current heatbeat mechanism for kubelets for instance.

I'm willing to back any of the three ideas listed in this thread so far.  But I think option 2 will have to be led by someone other than me.  option three might be an interesting compromise.

PR #1028, opened 2022-02-08
flow-control-api spec

pmoogi-redhat:
As a cluster admin I want to be able to set a flow control policy or logging rate limiting policy : drop as if logging rate exceeds rate limits provided (bytes/sec) on selected containers, ignore by dropping entire logs from selected group of containers.

*A cluster-wide default for all containers (optional)
*A special rate for containers in named namespaces
*A special rate for containers matching a label selector
*Rates are per-container and measured in bytes/sec,  the default value -1 means no limits are applied (default behaviour), 0 - means ignore entirely that set of selected containers logs belonging to a defined group

Rates are enforced by dropping log data if it exceeds the rate. Cluster Logging Forwarder Spec applies RateLimiting policies to its input sources of logs and also its output endpoints connecting to a given persistent storage.

JIRA Reference [Epic] - https://issues.redhat.com/browse/LOG-884
JIRA Reference [Story] - https://issues.redhat.com/browse/LOG-2207

### DIFF
--- a/enhancements/cluster-logging/flow-control-api.md
+++ b/enhancements/cluster-logging/flow-control-api.md
@@ -185,11 +185,10 @@ Extend the `ClusterLogForwarder` API with an optional `limits` field (list of `l
 The `limit` type has fields:
 
 - `name`: Name used to identify this limit.
-- `maxBytesPerSecond`: ([Quantity][Quantity], optional) second in [k8s resource notation][memory].\
-  Maximum byte rate allowed. If the inbound rate exceeds this limit then the `policy` is applied to keep the outbound rate within the limit.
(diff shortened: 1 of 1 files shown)


pmoogi-redhat:
@alanconway please review

alanconway (changes requested):
Made some suggested changes, you can apply them automatically via github. Important points:
- changed input.limit to input.perContainerLimit
- changed limit.maxBytesPerSecond to input.MaxRecordsPerSecond

alanconway:
@pmoogiThis should be linked to https://issues.redhat.com/browse/LOG-1043 - that task is specifically for designing the API.
 https://issues.redhat.com/browse/LOG-2207 is about implementing it.

alanconway:
I have some updates from recent discussion to be included as soon as I get a chance to write them up.
/hold

alanconway:
I will update in a separate PR.

PR #1029, opened 2022-02-08
OVN-Kubernetes on Hypershift

squeed:
Outlines a plan for deploying OVN-Kubernetes on Hypershift.

### DIFF
--- a/enhancements/network/ovn-on-hypershift.md
+++ b/enhancements/network/ovn-on-hypershift.md
@@ -1,0 +1,294 @@
+---
+title: OVN on Hypershift
+authors:
+  - "@russellb
+  - "@squeed"
+reviewers:
+  - "@dcbw"
+  - @danwinship
+  - @tssurya
+  - @numans
+approvers:
+  - TBD
+creation-date: 2022-02-01
+last-updated: 2022-02-08
+tracking-link: 
+  - "https://issues.redhat.com/browse/SDN-2589"
+see-also:
+  - "https://hypershift-docs.netlify.app/"
+---
+
+
+# OVN on Hypershift hosted
+
+## Summary
+
+[Hypershift](https://hypershift-docs.netlify.app/) is a new deployment model for OpenShift which removes the requirement that clusters be "self-hosted". Rather, a given hosted cluster's control plane actually resides as pods on another "management" cluster.
+
+Adapting the OVN networking plugin to the Hypershift model requires some careful design. Specifically, OVN provides its own control plane and datastore, all of which will be hosted external to the hosted cluster.
+
+Thus, the OpenShift OVN deployment model must be updated to support split control planes. Likewise, the Network Operator must be able to manage components on both the control plane as well as the hosted cluster.
+
+## Motivation
+
+We will elide the full motivation for hosted control planes (i.e. the Hypershift model). However, hosting the OVN control plane off-cluster has the same advantage as hosting the Kubernetes control plane:
+- potentially higher reliability
+- no special master nodes.
+- no chicken-egg problem while bootstrapping (i.e. cannot use most cluster features until networking comes up)
+- More available features (e.g. volumes, stateful sets)
+- etc.
+
+The most significant motivation is the availability of StatefulSets for the OVN databases. In OpenShift,
+we use the ovsdb raft replication mode to provide high availability on master nodes. Managing this
+manually is error prone, especially in the case of lost masters. However, self-hosted OVN cannot use
+StatefulSets, since they require a functioning network. OVN on Hypershift does not suffer this
+chicken-and-egg problem.
+
+### Goals
+
+Hypershift clusters are easily deployed using OVN, with the entire control plane residing off-cluster. OVN is the default networking provider for Hypershift.
+
+### Non-Goals
+
+- Removing the OVN south-bound DB (sbdb) as the primary integration point (for nodes)
+- Full network connectivity between the management and hosted clusters
+## Proposal
+
+The gist of the design is:
+
+1. All OVN control-plane components (ovnkube-master, NB/SB databases, ovn-northd), as well as the CNO, reside on the management cluster
+2. The CNO manages components in both clusters
+3. Communication between the nodes and the OVN control plane is via the southbound OVN database, exposed from the management cluster via a Route
+4. No changes to OVN are required
+
+
+### User Stories
+
+I, a Hypershift user, find running Kubernetes control planes terribly boorish. I would like all the fancy features
+available to me through ovn-kubernetes (EgressIP, EgressFirewall, ICNI, et cetera) without having to
+run that pesky control plane. My cluster networking Just Works without causing me any problems.
+
+### API Extensions
+
+n/a
+
+## Design Details
+
+### Cluster Network Operator (CNO) design
+
+The majority of this proposal's implementation effort will be in CNO. CNO is a complicated operator, with 10 controllers operating in varying levels of the stack.
+The full list of controllers within the CNO can be found [here](https://github.com/openshift/cluster-network-operator/blob/master/docs/operands.md).
+
+Most controllers within the CNO will not need to be aware of the management cluster, and will thus remain unchanged. All of their operands are entirely
+on the hosted cluster and Hypershift will be entirely transparent.
+
+Work needed to CNO includes:
+- Make specific controllers aware of multiple clusters
+- Refactor MTU detection
+- Managing ovsdb PKI across both clusters
+- Un-hard-coding namespaces. The CNO currently assumes certain namespaces.
+
+It is assumed the CNO's configuration objects (CRDs) reside in the hosted cluster. This is to avoid any potential CRD version conflicts while multiple versions of the CNO are running on the management cluster.
+
+### OVN-Kubernetes 
+
+OVN and OVN-Kubernetes have a control plane and node components. By and large, OVN and OVN-kubernetes will not require any code changes. Rather, how they are deployed must be rethought. This will primarily be implemented as changes in the CNO.
+
+```text
+                          │
+  Management Cluster      │       Hosted Cluster
+                          │
+┌──────────────────┐      │      ┌───node─────────┐
+│  Route (public)  │◄─────┼──────┤ ovnkube-node   │
+└────────┬─────────┘      │      │                │
+         │                │      │ ovn-controller │
+┌────────▼───┐            │      │                │
+│ ovs-db-N.. │            │      │ openvswitch    │
+└┬───────────┴───┐        │      │                │
+ │  nbdb | sbdb  │        │      └────────────────┘
+ └───────▲───────┘        │
+         │                │
+         │                │
+         │                │
+┌────────┴───────┐        │
+│ ovnkube-master │        │
+│   ovn-northd   │        │
+└────────────────┘        │
+```
+
+**Control-plane:**
+- ovn-kube master: Watches apiserver, creates objects in nbdb
+- ovn-northd: Translates nbdb objects to sbdb flows
+- nbdb, sbdb: Raft-based highly available configuration databases (northbound and southbound)
+
+**Node:**
+- ovn-kube node: Does some node-level configuration
+- ovn-controller: Programs flows from sbdb to Open vSwitch
+- Open vSwitch (OVS): actual packet handling
+
+The control plane components will be deployed as pods on the management cluster. The node components remain on the hosted cluster. The CNO will be made aware of both of these clusters so it can deploy the right compoments in the right place.
+
+The ovnkube-master pod currently runs using host networking. This must be changed to use the cluster network, instead.
+
+The CNO creates a headless service called `ovnkube-db` for the OVN databases. This needs to be updated to have a ClusterIP, since in this updated architecture, we need to point a `Route` at this `Service`. The CNO will also be responsible for creating this `Route` and configuring `ovn-controller` in the hosted cluster to connect to the associated hostname for southbound DB access.
+
+Anywhere the CNO assumes that `openshift-ovn-kubernetes` is the target namespace for OVN components must be updated. That will be the namespace for Node components in the hosted cluster, but not for the components that run in the hosted control plane.
+
+#### OVN Control plane
+
+Deploying the OVN control plane will require some general refactoring, as they no longer will run as HostNetwork, privileged pods. Rather
+
+**ovsdb-server (nbdb, sbdb)**: *should* not require any changes. This depends on Raft working when nodes are identified by DNS (and IPs change).
+
+**OVN-Kubernetes master:** *may* need some changes to perform leader election via the management cluster. Otherwise, it doesn't need to be aware of the management cluster.
+
+#### OVN DBs
+
+The OVN databases (nbdb, sbdb) will continue to use the internal raft-based replication. However, they will be deployed as a StatefulSet instead. Rather than using HostPath volumes, they will use whatever persistent storage volumes are available in the management cluster.
+In the case of AWS, there are some important details to recognize: EBS volumes cannot be moved between AZs. Thus, where they are first provisioned is critical.
+
+The databases need a N/2+1 quorum to maintain availability. We need to protect against disruption
+- zone-based anti-affinities, to ensure every DB is in a different AZ.
+- Suitable PodDisruptionBudget, to ensure we never take a DB down by maintenance.
+
+There has some been desire to support "swinging" control planes between clusters. As long as the PVs are persisted, this should be possible.
+We will need to test how the raft implementation handles name changes.
+
+
+#### OVN Node components
+**ovn-controller**: itself will not require any changes as it appears to already have support for the new functionality required. It must support configuration of the SB (southbound) database connection as a URL instead of an IP address. Further, when using TLS, it must set SNI (Server Name Indication) so that the management cluster is able to route the TLS connection to the correct database instance.
+
+**OVN-Kubernetes node:** should not require any changes.
+
+#### Node-DB connectivity
+By (ab)using the TLS Passthrough feature of OpenShift Routes, we can expose the southbound database directly to the target cluster.
+
+### Monitoring
+
+The control plane components, as pods in the management cluster, will need to be monitored by the management monitoring infrastructure. We will need to create a ServiceMonitor for each control plane in question.
+
+One complication is that we should create AlertManager alerts on a per-customer basis. That means, inevitably, that we will need some kind of management layer for these alerts. What form this takes is not yet defined.
+
+
+#### MTU detection
+Right now, the CNO assumes every node has the same MTU as the node on which it runs. This is, obviously, not correct anymore. Thus, we
+will need to add a MTU "prober" Job that executes, once, to determine MTU in the target cluster. It will write the result in a ConfigMap.
+The CNO will then consume this as needed.
+
+Setting MTU correctly is critical -- too large, and we suffer hard-to-diagnose connectivity issues. Too small, and performance suffers.
+Given that AWS supports jumbo frames, and GCP nodes have a <1500 MTU, this is not a theoretical problem.
+
+### Risks and Mitigations
+
+**OVSBD Raft and DNS** - the ovsdb team does not currently test Raft with dynamic member IPs. While the team believes this should work, it must be added to their test plan. Also, we have missed the window for any new changes to OVS for the 4.11 release. If any changes to ovsdb are needed, we either need exceptional processes or workarounds.
+
+### Open Questions 
+
+
+What storage / volume types will be available for the OVN DBs to use? Can we assume a functioning CSI environment? Will this storage have sufficient performance? Can we generate a reasonable size prediction? To answer this, we will need to utilize the existing Perf&Scale processes to characterize performance. Given that etcd on Hypershift also uses a PV-backed volume, this should not be an issue.
+
+Can we minimize cross-AZ network traffic? We only need to talk to the leader of each raft cluster, but there's no guarantee that nbdb
+and sbdb will elect the same leader at any point in time. Can we add a mode to ovn-northd that makes this possible? This may need to be
+a follow-up improvement.
+
+What bandwidth will the external sbdb traffic consume? Will the added latency cause any issues? Should we enable ovn-controller conditional monitoring? At the time this document was being drafted, the OVN team is re-examining the performance impacts of ovn-controller's conditional monitor feature.
+
+If the bandwidth or performance is not acceptable, will we have to deploy some kind of ovsdb proxy on the infra nodes? Assuming we cannot *require* infra nodes, will we nevertheless need to make that option available to customers? At present, the plan is to defer this unless
+experience shows this is necessary.
+
+How do we report control-plane status -- which references "invisible" pods -- to the end-user? Does the end-user need to know that ovsdb is somehow degraded? What do we show in the `network` ClusterOperator object?
+
+How do we report control-plane status to the SRE teams? They will need to react to losses of quorum, etc. Can we deploy alerts? How do we handle cross-version alert definitions?
+
+### Test Plan
+
+The general e2e test suites will be acceptable.  We expect OVN to be the default for all hypershift based OpenShift clusters once OVN support is ready, so anywhere we are testing hypershift will be exercising this integration.
+
+Additionally, we will need to write some disruption tests that ensure ovsdb raft plays well with `StatefulSet` rotations.
+
+We will work closely with the Performance & Scale test team to test out the performance of this architecture.
+
+### Graduation Criteria
+
+Once this functionality is complete, we expect to make OVN the default for hypershift managed clusters. At that point, the graduation of this functionality will be tied to the graduation of hypershift more broadly.
+
+#### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+- Enumerate service level indicators (SLIs), expose SLIs as metrics
+- Write symptoms-based alerts for the component(s)
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Document SLOs for the component
+- Conduct load testing
+- User facing documentation created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+#### Removing a deprecated feature
+n/a
+### Upgrade / Downgrade Strategy
+
+This architecture will only apply to new clusters deployed via hypershift. Thus, the generalized Hypershift upgrade model must take OVN in to account.
+
+OVN has somewhat complicated upgrade ordering. On upgrades, nodes must be upgraded before the control plane. For downgrades, the control plane must be rolled out first. While this logic is unchanged in Hypershift, splitting the components across clusters makes this more complicated w.r.t. potential failure cases.
+Where there is an ordering defined, we also need to wait for any Deployments / Daemonsets / StatefuSets to roll out -- it's not sufficient just to wait for the object to be updated. Accordingly, the code that enforces the ordering needs to be multi-cluster aware.
+
+
+### Version Skew Strategy
+
+There should not be any further version skew questions introduced by the hypershift model. Neither OVN-K nor CNO have a tight coupling with the API server or kubernetes control plane. Nevertheless, we (as well as the wider Hypershift team) need to consider the implication of very old control plane components running on an up-to-date control plane. At present, the CNO creates several `v1beta1` objects.
+
+In practice, client-go + kube-api support wide version skews, even if this is not officially supported. This is exploited in the EUS-EUS upgrade strategy.
+
+There is a coupling between OVN-Kubernetes, the machine-configuration-operator, and rollouts of files on the individual nodes ([specifically](https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml)). If end-users delay rollouts to their nodes, then we may trigger bugs not detected by CI or QE (since we do not test delayed rollouts).
+
+### Operational Aspects of API Extensions
+
+#### Failure Modes
+
+Probable failure situations include
+- OVN raft is degraded / broken
+  - Close to losing quorum for too long
+  - Quorum is failed
+- OVN DBs are too slow / overloaded
+  - Too much CPU / Memory
+  - Too slow a transaction rate
+  - Disk is almost full
+
+
+#### Support Procedures
+
+The SD / SRE team will be responsible for operating the control plane. As such, we will need to develop runbooks and alerting.
+
+The most likely problem is slow propagation of flows to the individual nodes. We have alerting to detect
+this, but we should define the procedures and alerts carefully.
+
+## Implementation History
+
+Major milestones in the life cycle of a proposal should be tracked in `Implementation
+History`.
+
+## Drawbacks
+
+It's not entirely clear, ultimately, if OVN sbdb is the right integration point. However, this is ultimately a question as to the design of OVN itself, rather than Hypershift per se. The stretched nature of Hypershift may expose limitations in the ovsdb protocol.
+
+The OVN control plane is "heavy-weight" - that is to say, it has high resource draw and is a complicated piece of software. Taking responsibility for its availability is certainly an operational burden. For end-users, however, it is a distinct advantage.
+
+## Alternatives
+
+### Run two copies of the CNO
+
+An alternative is to run two copies of the CNO for each hosted cluster. One would be in the hosted control plane and would only manage the parts of OVN that run there. The CNO running in the hosted cluster on worker nodes would manage only the parts of OVN running on the worker nodes.
+
+There are disadvantages to this approach and a lack of clear benefits compared to the direction proposed. Separating management of OVN makes upgrade coordination more difficult. We would also like to keep these management workloads off of the worker nodes wherever possible.
+
+### Integrate more directly with hypershift
+
+Instead of reusing the CNO, we briefly considered whether code should be shared with hypershift and have the existing hypershift components manage OVN. This seemed unnecessarily complex and without clear benefit.
+
+Hypershift does include its own implementation of managing some components, but it's focused on things that work in a drastically different way than they do on a standalone cluster. The changes are not as drastic. There's a different in the locations for where pieces run, but the architecture remains fundamentally the same.


enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+**ovsdb-server**: *should* not require any changes. This depends on Raft working when nodes are identified by DNS (and IPs change).
+
+**OVN-Kubernetes:** *may* need some changes to perform leader election via the management cluster. Otherwise, it doesn't need to be aware of the management cluster.
+
+#### OVN DBs
+
+The OVN databases (nbdb, sbdb) will continue to use the internal raft-based replication. However, they will be deployed as a StatefulSet instead. Rather than using HostPath volumes, they will use whatever persistent storage volumes are available in the management cluster.
+
+#### Node-DB connectivity
+By (ab)using the TLS Passthrough feature of OpenShift Routes, we can expose the southbound database directly to the client cluster.
```
knobunc, line 152: FWIW this is a documented and tested usage of the router, it's not really an abuse.  We advertise that it supports TLS passthrough (when the SNI header is present).  It's been a feature of the router since 3.0.

The only concern is that the router will eventually close the connections after a (tunable) period.  It's an hour by default, so as long as the client gracefully reconnects, there's no problem.

knobunc (commented):
This is nicely done.  Thanks Casey

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+
+What bandwidth will the sbdb traffic consume? Will the added latency cause any issues? Should we enable ovn-controller conditional monitoring? At the time this document was being drafted, the OVN team is re-examining the performance impacts of ovn-controller's conditional monitor feature.
+
+If the bandwidth or performance is not acceptable, will we have to deploy some kind of ovsdb proxy on the infra nodes? Assuming we cannot *require* infra nodes, will we nevertheless need to make that option available to customers?
+
+How do we report control-plane status -- which references "invisible" pods -- to the end-user? Does the end-user need to know that ovsdb is somehow degraded? What do we show in the `network` ClusterOperator object?
+
+How do we report control-plane status to the SRE teams? They will need to react to losses of quorum, etc. Can we deploy alerts? How do we handle cross-version alert definitions?
+
+How do we "ship" the CNO to Hypershift? Since it runs in the control plane, it would no longer be present in `/manifests`.
```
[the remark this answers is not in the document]
squeed: It depends, exactly, on how much control users have over cluster version. Assuming that upgrades are manually triggered, we could theoretically have a v4.11 CNO running on a much, much newer cluster.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+The OVN databases (nbdb, sbdb) will continue to use the internal raft-based replication. However, they will be deployed as a StatefulSet instead. Rather than using HostPath volumes, they will use whatever persistent storage volumes are available in the management cluster.
+
+#### Node-DB connectivity
+By (ab)using the TLS Passthrough feature of OpenShift Routes, we can expose the southbound database directly to the client cluster.
+
+### Monitoring
+
+The control plane components, as pods in the management cluster, will need to be monitored by the management monitoring infrastructure. We will need to create a ServiceMonitor for each control plane in question.
+
+One complication is that we should create AlertManager alerts on a per-customer basis. That means, inevitably, that we will need some kind of management layer for these alerts. What form this takes is not yet defined.
```
jewzaam, line 158: We need to solve alerts for other control plane components like apiserver availability.  Plan is to ship the metrics to RHOBS and manage alerts there for the control plane.  Details TBD as we flesh out this space and operationalize things.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Open Questions 
+
+
+What storage / volume types will be available for the OVN DBs to use? Can we assume a functioning CSI environment? Will this storage have sufficient performance? Can we generate a reasonable size prediction? To answer this, we will need to utilize the existing Perf&Scale processes to characterize performance.
+
+What bandwidth will the sbdb traffic consume? Will the added latency cause any issues? Should we enable ovn-controller conditional monitoring? At the time this document was being drafted, the OVN team is re-examining the performance impacts of ovn-controller's conditional monitor feature.
+
+If the bandwidth or performance is not acceptable, will we have to deploy some kind of ovsdb proxy on the infra nodes? Assuming we cannot *require* infra nodes, will we nevertheless need to make that option available to customers?
```
jewzaam, line 171: For Sd, the plan is to _not_ provide/require infra nodes.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+
+How do we report control-plane status -- which references "invisible" pods -- to the end-user? Does the end-user need to know that ovsdb is somehow degraded? What do we show in the `network` ClusterOperator object?
```
jewzaam, line 173: Metrics should report up to the control plane and will be shipped off to RHOBS, at least in the SD use case.  We're drawing the line at the control plane except for what is needed to support HPA / VPA, customers don't get the control plane metrics.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+
+- Removing the OVN south-bound DB (sbdb) as the primary integration point (for nodes)
+- Full network connectivity between the management and target clusters
+## Proposal
+
+The gist of the design is:
+
+1. All OVN control-plane components (ovnkube-master, NB/SB databases, ovn-northd), as well as the CNO, reside on the management cluster
+2. The CNO manages components in both clusters
+3. Communication between the nodes and the OVN control plane is via the southbound OVN database, exposed from the management cluster via a Route
```
tsorya, line 62: Route will work on aws platform but all others will require nodeport support too. Same as it is for other hypershift components.

enhancements/network/ovn-on-hypershift.md
russellb: They have no visibility into the db at all, so it's probably not worth reporting.  The customer visible side effects would be things like pod creation latency, or in the worst case, the inability to start up new pods if the network control plane is locked up completely.

Can an admin of a hosted cluster run `oc get clusteroperators` ?  What will they see?  Will they see any of the things running on the management cluster?  I'd guess they wouldn't see it at all?  I need to get hypershift running again and check I guess ...

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Test Plan
+
+The general e2e test suites will be acceptable.  We expect OVN to be the default for all hypershift based OpenShift clusters once OVN support is ready, so anywhere we are testing hypershift will be exercising this integration.
+
+Additionally, we will need to write some disruption tests that ensure ovsdb raft plays well with `StatefulSet` rotations.
+
+We will work closely with the Performance & Scale test team to test out the performance of this architecture.
+
+### Graduation Criteria
```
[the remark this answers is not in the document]
russellb: Target is GA in 4.11.

enhancements/network/ovn-on-hypershift.md
tsorya: Currently hypershift installs all control plane components services as node ports in case platform is none/agent, in case of OVN you will require Route and it will create inconsistency. 
Another issue is that hypershift can run on non OCP clusters and adding route will actually block it.
And the most important one, using route will require setting dns for this specific route and it is very problematic for many baremetal environments, especially when you try to show poc.
I think it is not lower priority as there are many planned scenarios for using hypershift + ovn and adding support only for route will block us

romfreiman: @russellb on prem with hypershift and OVN are planned to be production as well pretty soon. We actually have many opportunities on this front (not for public github :)
So I dont think we can de-prioritize that flow.
Regrading the challenges - will metal-lb be the solution for no prem ingress/route? 
cc: @derekwaynecarr

hardys: The route is created on the management cluster, but that doesn't mean we can't then deploy none/agent HostedClusters where ingress (on the guest cluster) may not be available by default?

enhancements/network/ovn-on-hypershift.md
squeed: Sounds like we should ensure that ovn-controller uses MonitorCondSince. Does it?

enhancements/network/ovn-on-hypershift.md
russellb: > 1. Currently hypershift installs all control plane components services as node ports in case platform is none/agent, in case of OVN you will require Route and it will create inconsistency.

OK - I don't think that's a big deal if it's useful, but I won't prioritize it equally with what's proposed.

>     2. hypershift can run on non OCP clusters and adding only route for ovn will actually block it.

This is completely OpenShift specific, when the OpenShift cluster-network-operator is used to manage ovn-kubernetes for a guest cluster, so non-OpenShift isn't really relevant for this proposal.  Point taken that running a non-openshift guest cluster would work differently.

>     3. Using route will require setting dns for this specific route and it is very problematic for many baremetal environments, especially when you try to show some proof of concept.

Typically DNS for routes is set up as a wildcard, right?  I get the impression you just don't want to block on external DNS configuration at all?  Again, I'm OK with it, but not for prioritizing it equally with our ideal flow for AWS as a first pass here.

russellb: > Regrading the challenges - will metal-lb be the solution for no prem ingress/route?

metallb is used for Services of type=LoadBalancer only.  It could theoretically be used for Ingress, but it's not set up that way automatically right now.  It's an optional component and requires extra configuration.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+  - @tssurya
+  - @numans
+approvers:
+  - TBD
+creation-date: 2022-02-01
+last-updated: 2022-02-08
+tracking-link: 
+  - "https://issues.redhat.com/browse/SDN-2589"
+see-also:
+  - "/enhancements/this-other-neat-thing.md"
```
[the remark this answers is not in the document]
russellb: yeah - just a copy/paste out of the enhancement template

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+---
+
+
+# OVN on Hypershift
+
+
+## Summary
+
+[Hypershift](https://hypershift-docs.netlify.app/) is a new deployment model for OpenShift which removes the requirement that clusters be "self-hosted". Rather, a given cluster's control plane actually resides as pods on another "management" cluster.
```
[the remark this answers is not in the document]
russellb: I've seen "guest cluster" a bunch ... according to the docs, we should be using "Hosted Cluster" -- https://hypershift-docs.netlify.app/reference/concepts-and-personas/

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+
+Adapting the OVN networking plugin to the Hypershift model requires some careful design. Specifically, OVN provides its own control plane and datastore, all of which will be hosted external to the target cluster.
+
+Thus, the OpenShift OVN deployment model must be updated to support split control planes. Likewise, the Network Operator must be able to manage components on both the control plane as well as the target cluster.
+
+## Motivation
+
+We will elide the full motivation for hosted control planes (i.e. the Hypershift model). However, hosting the OVN control plane off-cluster has the same advantage as hosting the Kubernetes control plane:
+- potentially higher reliability
```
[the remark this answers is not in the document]
russellb: If it's confusing, we can just remove it, because getting into the benefits of hypershift itself really isn't the point of this enhancement.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+- no special master nodes.
+- no chicken-egg problem while bootstrapping
```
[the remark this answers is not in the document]
russellb: chicken-egg, like we can't make use of cluster SDN functionality to run the cluster SDN infrastructure.  In this model, we can, at least for hosting the OVN control plane

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+- More available features (e.g. volumes, stateful sets)
+- etc.
+
+The most significant motivation is the availability of StatefulSets for the OVN databases. In OpenShift,
+we use the ovsdb raft replication mode to provide high availability on master nodes. Managing this
+manually is error prone, especially in the case of lost masters. However, self-hosted OVN cannot use
+StatefulSets, since they require a functioning network. OVN on Hypershift does not suffer this
```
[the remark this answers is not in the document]
russellb: I think the main benefit here is using persistent storage for the OVN databases.  Today they sit on ephemeral storage, which is OK because kube is the source of truth and the DBs can be rebuilt from scratch as needed.  It'd be better to avoid unnecessary database rebuilds though, especially as clusters get larger and/or more dense.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+
+n/a
+
+## Design Details
+
+### Cluster Network Operator (CNO) design
+
+The majority of this proposal's implementation effort will be in CNO. CNO is a complicated operator, with 10 controllers operating in varying levels of the stack.
+
+Most controllers within the CNO will not need to be aware of the management cluster, and will thus remain unchanged.
```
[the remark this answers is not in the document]
russellb: @squeed

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+│ ovs-db-N.. │            │      │ openvswitch    │
+└┬───────────┴───┐        │      │                │
+ │  nbdb | sbdb  │        │      └────────────────┘
+ └───────▲───────┘        │
+         │                │
+         │                │
+         │                │
+┌────────┴───────┐        │
+│ ovnkube-master │        │
+│ ovn-controller │        │
```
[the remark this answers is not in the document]
russellb: Yes, this should be ovn-northd, not ovn-controller.

My expectation is it's still in the same pod, but the arrow shows how services are talking to each other.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+
+The CNO creates a headless service called `ovnkube-db` for the OVN databases. This needs to be updated to have a ClusterIP, since in this updated architecture, we need to point a `Route` at this `Service`. The CNO will also be responsible for creating this `Route` and configuring `ovn-controller` to connect to the associated hostname for southbound DB access.
+
+Anywhere the CNO assumes that `openshift-ovn-kubernetes` is the target namespace for OVN components must be updated. That will be the namespace for Node components in the guest cluster, but not for the components that run in the hosted control plane.
+
+#### OVN Control plane
+
+Deploying the OVN control plane will require some general refactoring, as they no longer will run as HostNetwork, privileged pods. Rather
+
+**ovn-controller**: itself will not require any changes as it appears to already have support for the new functionality required. It must support configuration of the SB (southbound) database connection as a URL instead of an IP address. Further, when using TLS, it must set SNI (Server Name Indication) so that the management cluster is able to route the TLS connection to the correct database instance.
```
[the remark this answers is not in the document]
russellb: This text is for ovn-controller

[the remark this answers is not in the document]
tssurya: why is it under control-plane?

enhancements/network/ovn-on-hypershift.md
russellb: Ok, I got on a hosted cluster and confirmed you see all the cluster operators

```
oc get clusteroperators
NAME                                       VERSION       AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
console                                    4.10.0-rc.1   True        False         False      44h     
csi-snapshot-controller                    4.10.0-rc.1   True        False         False      4d15h   
dns                                        4.10.0-rc.1   True        False         False      4d15h   
image-registry                             4.10.0-rc.1   True        False         False      4d15h   
ingress                                    4.10.0-rc.1   True        False         False      4d15h   
kube-apiserver                             4.10.0-rc.1   True        False         False      4d15h   
kube-controller-manager                    4.10.0-rc.1   True        False         False      4d15h   
kube-scheduler                             4.10.0-rc.1   True        False         False      4d15h   
kube-storage-version-migrator              4.10.0-rc.1   True        False         False      4d15h   
monitoring                                 4.10.0-rc.1   True        False         False      4d15h   
network                                    4.10.0-rc.1   True        False         False      4d15h   
node-tuning                                4.10.0-rc.1   True        False         False      4d15h   
openshift-apiserver                        4.10.0-rc.1   True        False         False      4d15h   
openshift-controller-manager               4.10.0-rc.1   True        False         False      4d15h   
openshift-samples                          4.10.0-rc.1   True        False         False      4d15h   
operator-lifecycle-manager                 4.10.0-rc.1   True        False         False      4d15h   
operator-lifecycle-manager-catalog         4.10.0-rc.1   True        False         False      4d15h   
operator-lifecycle-manager-packageserver   4.10.0-rc.1   True        False         False      4d15h   
service-ca                                 4.10.0-rc.1   True        False         False      4d15h   
storage                                    4.10.0-rc.1   True        False         False      4d15h   
```

So this info is still visible, but you're right that less is actionable by the owner of the hosted cluster. 

@derekwaynecarr @csrwng Do you have any guidance for any changes we should make to how the cluster-network-operator reports its status for hosted clusters?  Or just leave it the same and info that isn't helpful to the target cluster owner may still be helpful if they have to report it back to whoever is running the management cluster?

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
tssurya: ah gotcha thanks Russell

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
tssurya: yeah I don't mind if its guest or target or hosted, as long as its the same one being used in all places.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+## Design Details
+
+### Cluster Network Operator (CNO) design
+
+The majority of this proposal's implementation effort will be in CNO. CNO is a complicated operator, with 10 controllers operating in varying levels of the stack.
+
+Most controllers within the CNO will not need to be aware of the management cluster, and will thus remain unchanged.
+
+Work needed to CNO includes:
+- Make controllers aware of multiple clusters
```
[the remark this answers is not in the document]
squeed: Well, one CNO will manage one HyperShift user. However, imagine that that cluster itself is BlueField, and thus has two control planes in and of itself. Thus, we need support for arbitrary named clusters.

enhancements/network/ovn-on-hypershift.md
squeed: I talked to @dceara, and he confirmed that ovn-controller uses MonitorCondSince, which means it **will not** download the whole DB on a simple connection reset (for most cases).

enhancements/network/ovn-on-hypershift.md
squeed: @jewzaam is there a link / document I can reference for this?

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
csrwng: For operators we deploy on the control plane side, we take the image from the release payload for the hosted cluster, not the management cluster. So you ship the same way. The cno is included in the release payload. We will filter out the manifests for the CVO to not apply them to the guest cluster and we will deploy the operator via the hypershift control plane operator.

enhancements/network/ovn-on-hypershift.md
csrwng: I'd say we can leave it as is now. If necessary we can reconcile the status of the HostedCluster to reflect the status reported inside the guest cluster's cno clusteroperator resource.

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
csrwng: > Most controllers within the CNO will not need to be aware of the management cluster, and will thus remain unchanged.

Does this mean that the CNO will still be able to support a different SDN like Calico by deploying the kube-proxy as usual? (We have this use case with IBM today)

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
squeed: ah, thanks for the pointer to the official verbiage, will update.

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
squeed: FYI, ovn-k currently uses hostpath volumes, so we don't lose the database in general. However, the real bother is that we can't add or replace master nodes without manual intervention.

enhancements/network/ovn-on-hypershift.md
squeed: When you read Route here, you can also think Ingress object. It's essentially the same. This doesn't rely on any networking features on the hosted cluster than the ability to reach the control plane.

How are we exposing the control plane to the hosted cluster if not a Route!?

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
squeed: All this means is that most of the CNO's [operands](https://github.com/openshift/cluster-network-operator/blob/master/docs/operands.md) will need nothing to do with the management cluster, and thus will be blissfully unaware of any of this.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+```text
+                          │
+  Management Cluster      │       Target Cluster
+                          │
+┌──────────────────┐      │      ┌───node─────────┐
+│  Route (public)  │◄─────┼──────┤ ovnkube-node   │
+└────────┬─────────┘      │      │                │
+         │                │      │ ovn-controller │
+┌────────▼───┐            │      │                │
+│ ovs-db-N.. │            │      │ openvswitch    │
```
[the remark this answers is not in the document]
squeed: Nth replica in a stateful set. Not enough room to fit it all.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+- nbdb, sbdb: Raft-based highly available configuration databases (northbound and southbound)
+
+**Node:**
+- ovn-kube node: Does some node-level configuration
+- ovn-controller: Programs flows from sbdb to Open vSwitch
+- Open vSwitch (OVS): actual packet handling
+
+The control plane components will be deployed as pods on the management cluster. The node components remain on the target cluster. The CNO will be made aware of both of these clusters so it can deploy the right compoments in the right place.
+
+The ovnkube-master pod currently runs using host networking. This must be changed to use the cluster network, instead.
```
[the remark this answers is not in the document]
squeed: Fortunately we can use DNS. Based on our first experiments, it should work fine. If necessary, we can hack around this with some service IPs and horrible glue.

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
squeed: yeah yeah this needs a fixup, good catch.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### OVN Control plane
+
+Deploying the OVN control plane will require some general refactoring, as they no longer will run as HostNetwork, privileged pods. Rather
+
+**ovn-controller**: itself will not require any changes as it appears to already have support for the new functionality required. It must support configuration of the SB (southbound) database connection as a URL instead of an IP address. Further, when using TLS, it must set SNI (Server Name Indication) so that the management cluster is able to route the TLS connection to the correct database instance.
+
+**ovsdb-server**: *should* not require any changes. This depends on Raft working when nodes are identified by DNS (and IPs change).
+
+**OVN-Kubernetes:** *may* need some changes to perform leader election via the management cluster. Otherwise, it doesn't need to be aware of the management cluster.
```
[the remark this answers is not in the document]
squeed: yup, good point.

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
squeed: easy enough; removed this.

squeed:
Updated the document based on feedback.

I think this is getting close to merge.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,272 @@
...
+creation-date: 2022-02-01
+last-updated: 2022-02-08
+tracking-link: 
+  - "https://issues.redhat.com/browse/SDN-2589"
+see-also:
+  - "https://hypershift-docs.netlify.app/"
+---
+
+
+# OVN on Hypershif hosted
```
tssurya, line 22:
```suggestion
# OVN on Hypershift hosted
```

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
tssurya: agreed that purpose of hypershift and how that provides reliability is not the scope here, nvm me, we can leave it as is.

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
tssurya: ack thanks both,

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Design Details
+
+### Cluster Network Operator (CNO) design
+
+The majority of this proposal's implementation effort will be in CNO. CNO is a complicated operator, with 10 controllers operating in varying levels of the stack.
+The full list of controllers within the CNO can be found [here](https://github.com/openshift/cluster-network-operator/blob/master/docs/operands.md).
+
+Most controllers within the CNO will not need to be aware of the management cluster, and will thus remain unchanged. All of their operands are entirely
+on the hosted cluster and Hypershift will be entirely transparent.
```
tssurya, line 83: ++

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
tssurya: sounds good :)

tssurya (commented):
/lgtm mostly on the agreed high level design.
I still have the developer itch type questions which I will save for the future CNO PRs

enhancements/network/ovn-on-hypershift.md
squeed: My first plan is to just say something like

- "Control plane upgrade is rolling out, 1 out of 3 pods updated"
- "Control plane is degraded: ovn-db has 1 unready pod"

and discard any pod-specific information.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,271 @@
...
+
+What bandwidth will the sbdb traffic consume? Will the added latency cause any issues? Should we enable ovn-controller conditional monitoring? At the time this document was being drafted, the OVN team is re-examining the performance impacts of ovn-controller's conditional monitor feature.
+
+If the bandwidth or performance is not acceptable, will we have to deploy some kind of ovsdb proxy on the infra nodes? Assuming we cannot *require* infra nodes, will we nevertheless need to make that option available to customers?
+
+How do we report control-plane status -- which references "invisible" pods -- to the end-user? Does the end-user need to know that ovsdb is somehow degraded? What do we show in the `network` ClusterOperator object?
+
+How do we report control-plane status to the SRE teams? They will need to react to losses of quorum, etc. Can we deploy alerts? How do we handle cross-version alert definitions?
```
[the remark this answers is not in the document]
squeed: How do we configure this? Do we create a single ServiceMonitor object for the management cluster, or should we create one per namespace (one per hosted cluster)?

enhancements/network/ovn-on-hypershift.md
squeed: :bow:

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
squeed: But, to bring this together - yes, kube-proxy + calico should work just fine as well.

enhancements/network/ovn-on-hypershift.md
derekwaynecarr: I believe we need to support NodePort and Route for this option based on interviewing interested users.
Is there a technical reason NodePort will not work?

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,275 @@
...
+
+1. All OVN control-plane components (ovnkube-master, NB/SB databases, ovn-northd), as well as the CNO, reside on the management cluster
+2. The CNO manages components in both clusters
+3. Communication between the nodes and the OVN control plane is via the southbound OVN database, exposed from the management cluster via a Route
+4. No changes to OVN are required
+
+
+### User Stories
+
+I, a Hypershift user, find running Kubernetes control planes terribly boorish. I would like all the fancy features
```
derekwaynecarr, line 67: this cracked me up ;-)

thanks for the humor.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,275 @@
...
+
+The majority of this proposal's implementation effort will be in CNO. CNO is a complicated operator, with 10 controllers operating in varying levels of the stack.
+The full list of controllers within the CNO can be found [here](https://github.com/openshift/cluster-network-operator/blob/master/docs/operands.md).
+
+Most controllers within the CNO will not need to be aware of the management cluster, and will thus remain unchanged. All of their operands are entirely
+on the hosted cluster and Hypershift will be entirely transparent.
+
+Work needed to CNO includes:
+- Make specific controllers aware of multiple clusters
+- Refactor controllers off of controller-framework
```
derekwaynecarr, line 87: I am curious why this is a requirement.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,275 @@
...
+- Refactor MTU detection
+- Managing ovsdb PKI across both clusters
+- Un-hard-coding namespaces. The CNO currently assumes certain namespaces.
+
+It is assumed the CNO's configuration objects (CRDs) reside in the hosted cluster. This is to avoid any potential CRD version conflicts while multiple versions of the CNO are running on the management cluster.
```
derekwaynecarr, line 92: it would be good to enumerate which CRDs are projected as "read-only" inside the hosted cluster.

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
derekwaynecarr: this sounds consistent with what we did for etcd statefulset.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,275 @@
...
+
+What bandwidth will the sbdb traffic consume? Will the added latency cause any issues? Should we enable ovn-controller conditional monitoring? At the time this document was being drafted, the OVN team is re-examining the performance impacts of ovn-controller's conditional monitor feature.
+
+If the bandwidth or performance is not acceptable, will we have to deploy some kind of ovsdb proxy on the infra nodes? Assuming we cannot *require* infra nodes, will we nevertheless need to make that option available to customers?
+
+How do we report control-plane status -- which references "invisible" pods -- to the end-user? Does the end-user need to know that ovsdb is somehow degraded? What do we show in the `network` ClusterOperator object?
+
+How do we report control-plane status to the SRE teams? They will need to react to losses of quorum, etc. Can we deploy alerts? How do we handle cross-version alert definitions?
+
+How do we detect guest cluster MTU? We cannot just probe from the CNO as normal. We will either need to require MTU to be configured, or execute a probe pod. We do, however, support changing MTU
```
derekwaynecarr, line 181: i assume this is modifiable via the CNO config today, and we will provide that as referenced config to HostedCluster?

/cc @csrwng

csrwng: My understanding is that mtu is configured in the `network.operator.openshift.io` resource. We currently don't let you specify that from HostedCluster. So far we've only exposed configuration resources in the `config.openshift.io` group, which includes `network.config.openshift.io`. We would need to add explicit support for the operator config and sync it to the guest cluster so that the cno can find it where expected.

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
alvaroaleman: So far, we create one per hosted cluster.

enhancements/network/ovn-on-hypershift.md
squeed: Ah, that is actually something I had misunderstood. So, the CRDs in question live in the management cluster?

enhancements/network/ovn-on-hypershift.md
squeed: That's correct, it's assumed to be a "lower level" knob.

Another option is to probe for the MTU via a Job once the first guest Node appears. We would then persist this.

FWIW, end-users may still wish to modify the `network.operator.openshift.io` if they intend to use Multus or advanced OVN features (e.g. flow exporting, policy auditing).

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,275 @@
...
+- Conduct load testing
+- User facing documentation created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+#### Removing a deprecated feature
+n/a
+### Upgrade / Downgrade Strategy
+
+This architecture will only apply to new clusters deployed via hypershift. Thus, the generalized Hypershift upgrade model must take OVN in to account.
+
+OVN has somewhat complicated upgrade ordering. On upgrades, nodes must be upgraded before the control plane. For downgrades, the control plane must be rolled out first. While this logic is unchanged in Hypershift, splitting the components across clusters makes this more complicated w.r.t. potential failure cases.
```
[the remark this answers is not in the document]
squeed: The CNO always go first, and will manage the upgade ordering. It should Just Work, and we've not had issues in the field so far. Just something to call out.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,275 @@
...
+
+In practice, client-go + kube-api support wide version skews, even if this is not officially supported. This is exploited in the EUS-EUS upgrade strategy.
+
+There is a coupling between OVN-Kubernetes, the machine-configuration-operator, and rollouts of files on the individual nodes ([specifically](https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml)). If end-users delay rollouts to their nodes, then we may trigger bugs not detected by CI or QE (since we do not test delayed rollouts).
+
+### Operational Aspects of API Extensions
+
+#### Failure Modes
+
+Probable failure situations include
```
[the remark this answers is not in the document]
squeed: Another option is to use HostPath volumes and just write to disk. Then, assuming we can get our rollout gates correct, we wouldn't lose data as long as quorum is maintained.

What will we be doing for etcd?

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,275 @@
...
+
+The OVN databases (nbdb, sbdb) will continue to use the internal raft-based replication. However, they will be deployed as a StatefulSet instead. Rather than using HostPath volumes, they will use whatever persistent storage volumes are available in the management cluster.
+
+#### OVN Node components
+**ovn-controller**: itself will not require any changes as it appears to already have support for the new functionality required. It must support configuration of the SB (southbound) database connection as a URL instead of an IP address. Further, when using TLS, it must set SNI (Server Name Indication) so that the management cluster is able to route the TLS connection to the correct database instance.
+
+**OVN-Kubernetes node:** should not require any changes.
+
+#### Node-DB connectivity
+By (ab)using the TLS Passthrough feature of OpenShift Routes, we can expose the southbound database directly to the client cluster.
```
numansiddique, line 156: I think its better to replace "client cluster" with "hosted cluster" because every where else "hosted cluster" is used.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,275 @@
...
+One complication is that we should create AlertManager alerts on a per-customer basis. That means, inevitably, that we will need some kind of management layer for these alerts. What form this takes is not yet defined.
+
+### Risks and Mitigations
+
+**OVSBD Raft and DNS** - the ovsdb team does not currently test Raft with dynamic member IPs. While the team believes this should work, it must be added to their test plan. Also, we have missed the window for any new changes to OVS for the 4.11 release. If any changes to ovsdb are needed, we either need exceptional processes or workarounds.
+
+### Open Questions 
+
+
+What storage / volume types will be available for the OVN DBs to use? Can we assume a functioning CSI environment? Will this storage have sufficient performance? Can we generate a reasonable size prediction? To answer this, we will need to utilize the existing Perf&Scale processes to characterize performance.
```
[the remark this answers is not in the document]
numansiddique: ovsdb-server stores the db in ASCII and the size of the database file keeps increasing with the transactions committed to it and periodically the db gets compacted by ovsdb-server. 

Even I'm not aware of performance characteristics.   @dceara @igsilya - do you have any comments ?

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
numansiddique: If there is a requirement to upgrade control plane first,  then it can be done.  OVN supports the option to PIN ovn-controller and ovn-northd to a specific version.  More details here - https://github.com/ovn-org/ovn/commit/1dd27ea7aea40122c1edbff845e14abaa70c0413

CNO needs to be enhanced if OCP/Hypershift wants to make use of this.

enhancements/network/ovn-on-hypershift.md
```diff
@@ -0,0 +1,275 @@
...
+  - Too slow a transaction rate
+  - Disk is almost full
+
+
+#### Support Procedures
+
+The SD / SRE team will be responsible for operating the control plane. As such, we will need to develop runbooks and alerting.
+
+The most likey problem is slow propagation of flows to the individual nodes. We have alerting to detect
+this, but we should the procedures and alert definitions carefully.
```
numansiddique, line 250: There seems to be typo -- "we should define the procedures and alert ..."

enhancements/network/ovn-on-hypershift.md
squeed: Mostly because of the difficulty of wedging muli-cluster support in to controller-framework. Also, controller-framework has never been a good fit for CNO - we would like the ability to do server-side label scoped watches, for example.

alvaroaleman: If you mean controller-runtime, it does support both multi-cluster controllers and label-filtered watches. But either way, such a refactoring seems to be a bit orthogonal?

enhancements/network/ovn-on-hypershift.md
russellb: There's no technical reason NodePort can't work.  It was just a prioritization question for me, and I want to prioritize dev/test of the more urgent thing (Route).  Otherwise I'm fine supporting both.  In practice I'm sure supporting both isn't that much work, at least at development time.

romfreiman: From what I understand (please correct me if Im wrong, and if so, just ignore my comment), not having NodePort in none platfrom (on prem basically - dpus, late binding, agent based provisioning), will enforce users to create external DNS per hosted cluster that they will spawn. 
@csrwng @derekwaynecarr in current on prem model (none platform, sdn) - I think it's not a must and everything is working based on ips, correct?
If this is the case, I dont think it's reasonable to enforce the user to configure dns per hosted cluster route.

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
derekwaynecarr: @squeed we are always write to a PV backed volume for etcd.

enhancements/network/ovn-on-hypershift.md
jewzaam: @squeed https://docs.google.com/document/d/1YoohUsfarD-c91XFoZ16fbWZfmd_fR9FeH5yr_DmIUQ/edit (RH internal)

dofinn: @squeed https://github.com/openshift/enhancements/pull/981/files :) 
I think for this enhancement, exposing `ServiceMonitor`s for said components is enough. The rest should be up the architecture regarding how to consume the `ServiceMonitor`s. 

This is the current flow for other HOSTCP components as per https://issues.redhat.com/browse/HOSTEDCP-169 - specifically https://issues.redhat.com/browse/HOSTEDCP-293

enhancements/network/ovn-on-hypershift.md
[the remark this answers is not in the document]
squeed: Glad to hear. If it's good enough for etcd, it's good enough for me.

enhancements/network/ovn-on-hypershift.md
squeed: There are a few places where it doesn't really fit our needs. Agreed that this is mostly orthogonal at this point.

enhancements/network/ovn-on-hypershift.md
squeed: Added MTU probe job.

squeed:
@derekwaynecarr @russellb and others: Updated based on feedback. No major changes, just some more details, especially around MTU detection and AntiAffinity / DisruptionBudget for ovsdbs.

I think this is ready to go in as a first run. Once we have the details around how to deliver alerts nailed down, we can update those sections in another PR.

russellb:
> 

Thanks!  It lgtm.  For the approve/lgtm maybe @knobunc (or a delegate from OCP networking) and @derekwaynecarr (or a delegate from hypershift) ?

russellb:
also approved by @derekwaynecarr during a call, so i'm just merging this

/lgtm

PR #1014, opened 2022-01-20
Heterogeneous architecture clusters

Prashanth684:
Introduces support for provisioning and upgrading heterogenous architecture clusters in phases

PR #1032, opened 2022-02-10
OpenShift CoreOS Layering (provisional)

cgwalters:
**NOTE: Nothing in this proposal should be viewed as final.
It is highly likely that details will change.  It is quite possible
that larger architectural changes will be made as well.**

Change RHEL CoreOS as shipped in OpenShift to be a "base image" that can
be used as in layered container builds and then booted.  This will allow
custom 3rd party agents delivered via RPMs installed in a container
build.  The MCO will roll out and monitor these custom builds the same
way it does for the "pristine" CoreOS image today.

This is the OpenShift integration of [ostree native containers](https://fedoraproject.org/wiki/Changes/OstreeNativeContainer) or [CoreOS layering](https://github.com/coreos/enhancements/pull/7) via the MCO.

### DIFF
--- a/enhancements/ocp-coreos-layering.md
+++ b/enhancements/ocp-coreos-layering.md
@@ -201,18 +201,54 @@ For now, updating kernel arguments will continue to happen via the MCD on each n
 
 Ignition will continue to handle the `disks` and `filesystem` sections - for example, LUKS will continue to be applied as it has been today.
 
-Further, it is likely that we will need to ship a targeted subset of the configuration via Ignition too - for example, the pull secret will be necessary to pull the build containers. 
+However, see below.
 
 ##### Per machine state, the pointer config
 
 See [MCO issue 1720 "machine-specific machineconfigs"](https://github.com/openshift/machine-config-operator/issues/1720).
 We need to support per machine/per node state like static IP addresses and hostname.
 
-##### 3 Ignition "levels"
+##### Transitioning existing systems
 
-- Pointer configuration: this stays unchanged
-- Firstboot ignition: Contains the bits needed to perform the switch to the custom image
-- Everything else: This all ends up in the `mco-coreos` container image, e.g. `kubelet.service` systemd unit.
+This is a complex and multi-faceted topic.  First, let's assume the node is already
+running a new enough host stack (most specifically rpm-ostree with container support).
+
+The system is running e.g. 4.11 and will be upgrading to e.g. 4.12 which will transition
+to the new format.  In this case, the user is not customizing the image at all, and we just
+want to perform a seamless transition.
+
+One key aspect we need to handle here is the "per machine state" above.
+
+- Node currently has a large set of files written to e.g. `/etc` and `/usr/local` via
+  Ignition.
+- The MCO asks the node to deploy the new format image
+- The MCD detects it is transitioning from old format to native container format, and
+  *does not perform any filesystem modifications at all*.
+- MCD simply runs `rpm-ostree rebase` + drain + `reboot`
+
+With this approach, any locally-modified files `/etc` will be preserved by this upgrade, such
+as static IP addresses, etc.  This means that e.g. rollback will still work in theory.
+
+However, there are two problems:
+
+First, we will need to be careful to keep e.g. `kubelet.service` in `/etc`, and not as
+part of the container build also migrate it to `/usr` (which we can do now!).  If we
+moved the unit, then we'd have *two* copies.
+
+This risk is greater for helper binaries (things in `/usr/local` i.e. `/var/usrlocal`).
+Those must be moved into `/usr/bin` in the image.  We will need to be careful to ensure
+that the binaries in `/usr/bin` are preferred by scripts.  In general, we should be
+able to handle that by using absolute paths.
+
+#### Ignition (pointer vs provisioning vs in-image)
+
+We will likely keep the pointer configuration as is.  Attempting to scope in
+changing that now brings its own risks.
+
+However, the MCS will need to serve a "provisioning configuration" for example,
+with at least a pull secret and other configuration sufficient to pull images.
+(Alternatively, we may be able to provide a "bootstrap pull secret" that allows
+ doing a pull-through from the in-cluster registry)
 
 #### Drain and reboot
 
--- a/enhancements/ocp-coreos-layering.md
+++ b/enhancements/ocp-coreos-layering.md
@@ -1,0 +1,443 @@
+---
+title: OpenShift CoreOS Layering
+authors:
+  - "@yuqi-zhang"
+  - "@jkyros"
+  - "@mkenigs"
+  - "@kikisdeliveryservice"
+  - "@cheesesashimi"
+  - "@cgwalters"
+  - "@darkmuggle (emeritus)"
+reviewers:
+  - "@mrunalp"
+approvers:
+  - "@sinnykumari"
+  - "@mrunalp"
+creation-date: 2021-10-19
+last-updated: 2022-02-09
+tracking-link:
+  - https://issues.redhat.com/browse/GRPA-4059
+---
+
+# OpenShift Layered CoreOS (PROVISIONAL)
+
+**NOTE: Nothing in this proposal should be viewed as final.  It is highly likely that details will change.  It is quite possible that larger architectural changes will be made as well.**
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Change RHEL CoreOS as shipped in OpenShift to be a "base image" that can be used as in layered container builds and then booted.  This will allow custom 3rd party agents delivered via RPMs installed in a container build.  The MCO will roll out and monitor these custom builds the same way it does for the "pristine" CoreOS image today.
+
+This is the OpenShift integration of [ostree native containers](https://fedoraproject.org/wiki/Changes/OstreeNativeContainer) or [CoreOS layering](https://github.com/coreos/enhancements/pull/7) via the MCO.
+
+## Motivation
+
+1. We want to support more customization, such as 3rd party security agents (often in RPM format).
+1. It is very difficult today to roll out a hotfix (kernel or userspace).
+1. It should be possible to ship configuration with the same mechanism as adding content.
+1. There are upgrade problems today caused by configuration updates of the MCO applied separately from OS changes.
+1. It is difficult to inspect and test in advance what configuration will be created by the MCO without having it render the config and upgrade a node.
+
+### Goals
+
+- [ ] Administrators can add custom code alongside configuration via an opinionated, tested declarative build system and via e.g. `Dockerfile` as an unvalidated generic mechanism
+- [ ] The output is a container that can be pushed to a registry, inspected, and processed via security scanners. 
+- [ ] Further, it can be executed as a regular container (via podman or as a pod) *and* booted inside or outside of a cluster
+- [ ] Gain [transactional configuration+code changes](https://github.com/openshift/machine-config-operator/issues/1190) by having configuration+encode captured atomically in a container
+- [ ] Avoid breaking existing workflow via MachineConfig (including extensions)
+- [ ] Avoid overwriting existing custom modifications (such as files managed by other operators) during upgrades
+
+### Non-Goals
+
+- While the base CoreOS layer/ostree-container tools will be usable outside of OpenShift, this enhancement does not cover or propose any in-cluster functionality for exporting the forward-generated image outside of an OpenShift cluster. In other words, it is not intended to be booted (`$ rpm-ostree rebase <image>`) from outside of a cluster.
+- This proposal does not cover generating updated "bootimages"; see https://github.com/openshift/enhancements/pull/201
+- Don't change existing workflow for RHEL worker nodes
+
+## Proposal
+
+**NOTE: Nothing in this proposal should be viewed as final.  It is highly likely that details will change.  It is quite possible that larger architectural changes will be made as well.**
+
+1. The `machine-os-content` shipped as part of the release payload will change format to the new "native ostree-container" format, in which the OS content appears as any other OCI/Docker container.  
+  (In contrast today, the existing `machine-os-content` has an ostree repository inside a UBI image, which is hard to inspect and cannot be used for derived builds).  For more information, see [ostree-rs-ext](https://github.com/ostreedev/ostree-rs-ext/) and [CoreOS layering](https://github.com/coreos/enhancements/pull/7). 
+2. Documentation and tooling will be available for generating derived images from this base image
+3. This tooling will be used by the MCO to create derived images (more on this in a separate enhancement)
+
+
+### User Stories
+
+#### What works now continues to work
+
+An OpenShift administrator at example.corp is happily using OpenShift 4 (with RHEL CoreOS) in several AWS clusters today, and has only a small custom MachineConfig object to tweak host level auditing.  They do not plan to use any complex derived builds, and expect that upgrading their existing cluster continues to work and respect their small audit configuration change with no workflow changes.
+
+#### Adding a 3rd party security scanner/IDS
+
+example.bank's security team requires a 3rd party security agent to be installed on bare metal machines in their datacenter.  The 3rd party agent comes as an RPM today, and requires its own custom configuration.  While the 3rd party vendor has support for execution as a privileged daemonset on their roadmap, it is not going to appear soon. 
+
+After initial cluster provisioning is complete, the administrators at example.bank supply a configuration that adds a repo file to `/etc/yum.repos.d/agentvendor.repo` and requests installation of a package named `some-3rdparty-security-agent` as part of a container build.
+
+The build is successful, and is included in both the control plane (master) and worker pools, and is rolled out in the same way the MCO performs configuration and OS updates today.
+
+A few weeks later, when Red Hat releases a kernel security update as part of a cluster image, and the administrator starts a cluster upgrade, it triggers a rebuild of both the control plane and worker derived images, which succeed.  The update is rolled out to the nodes.
+
+A month after that, the administrator wants to make a configuration change, and creates a `machineconfig` object targeting the `worker` pool.  This triggers a new image build.  But, the 3rd party yum repository is down, and the image build fails.  The operations team gets an alert, and resolves the repository connectivity issue.  They manually restart the build which succeeds.
+
+#### Kernel hotfix
+
+example.corp runs OCP on aarch64 on bare metal.  An important regression is found that only affects the aarch64 architecture on some bare metal platforms.  While a fix is queued for a RHEL 8.x z-stream, there is also risk in fast tracking the fix to *all* OCP platforms.  Because this fix is important to example.corp, a hotfix is provided via a pre-release `kernel.rpm`.
+
+The OCP admins at example.corp get a copy of this hotfix RPM into their internal data store, and configure a derived build where the kernel RPMs are applied. The MCO builds a derived image and rolls it out.
+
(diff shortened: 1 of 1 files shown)


enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,435 @@
...
+
+After initial cluster provisioning is complete, the administrators at example.bank supply a configuration that adds a repo file to `/etc/yum.repos.d/agentvendor.repo` and requests installation of a package named `some-3rdparty-security-agent` as part of a container build.
+
+The build is successful, and is included in both the control plane (master) and worker pools, and is rolled out in the same way the MCO performs configuration and OS updates today.
+
+A few weeks later, when Red Hat releases a kernel security update as part of a cluster image, and the administrator starts a cluster upgrade, it triggers a rebuild of both the control plane and worker derived images, which succeed.  The update is rolled out to the nodes.
+
+A month after that, the administrator wants to make a configuration change, and creates a `machineconfig` object targeting the `worker` pool.  This triggers a new image build.  But, the 3rd party yum repository is down, and the image build fails.  The operations team gets an alert, and resolves the repository connectivity issue.  They manually restart the build which succeeds.
+
+#### Kernel hotfix
```
danwinship, line 92: @cgwalters yes, this user story exactly covers the problem I was asking about on Slack (subject to the warning down below under "Support Procedures" about making sure it's very clear when a customer is doing this)

cgwalters:
I'd like to proceed with at least adding a new `rhcos` entry to the release payload for 4.11 (it wouldn't actually be *used* at first), and we plan to try to land some preparatory stuff for this to the MCO git main/master.  So if we can get some initial even tentative approval for the broad outlines of this proposal, that'd be great!

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,435 @@
...
+
+example.corp runs OCP on aarch64 on bare metal.  An important regression is found that only affects the aarch64 architecture on some bare metal platforms.  While a fix is queued for a RHEL 8.x z-stream, there is also risk in fast tracking the fix to *all* OCP platforms.  Because this fix is important to example.corp, a hotfix is provided via a pre-release `kernel.rpm`.
+
+The OCP admins at example.corp get a copy of this hotfix RPM into their internal data store, and configure a derived build where the kernel RPMs are applied. The MCO builds a derived image and rolls it out.
+
+(Note: this flow would likely be explained as a customer portal document, etc.)
+
+Later, a fixed kernel with a newer version is released in the main OCP channels.  The override continues to apply, and will currently require manual attention at some point later to remove the override when we can guarantee the non-hotfix kernel has the desired fix.
```
mrunalp, line 100: We can add a note that this could be automated in the future (detection and removal of hotfix with released rpm).

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,435 @@
...
+4. Inside the cluster, there will be an `openshift-machine-config-operator/coreos` object of type `imagestream`, owned by the MCO.
+5. The MCO builds a container image for each `MachineConfigPool` from user input using the cluster build system 
+7. Each machineconfig pool will also support a `custom-coreos` `BuildConfig` object and imagestream.  This build must use the `mco-coreos` imagestream as a base.   The result of this will be rolled out by the MCO to nodes.
+8. Each machineconfig pool will also support a `custom-external-coreos` imagestream for pulling externally built images (PROVISIONAL)
+9. MCD continues to perform drains and reboots, but writes much less configuration per node
+10. The Machine Configuration Server (MCS) will only serve a "bootstrap" Ignition configuration (pull secret, network configuration) sufficient for the node to pull the target container image.
+
+For clusters without any custom MachineConfig at all, the MCO will deploy the result of the `mco-coreos` build.
+
+<!-- current drawing: https://docs.google.com/drawings/d/1cGly8mjYDezVjUQiUEWd63IO_S2ZrgSbv1cT5X1KJIk/edit -->
```
mrunalp, line 131: Could we add a section on upgrade flow? It could cover:
1. How we move existing clusters to the new format.
2. How the imagestream is updated as new OpenShift updates go out.
(The information is there in the doc, but it may be easier to follow if we have a dedicated section.)

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,435 @@
...
+
+There are other things that pull Ignition:
+
+- [openshift-ansible for workers](https://github.com/openshift/openshift-ansible/blob/c411571ae2a0b3518b4179cce09768bfc3cf50d5/roles/openshift_node/tasks/apply_machine_config.yml#L23)
+- [openshift-ansible for bootstrap](https://github.com/openshift/openshift-ansible/blob/e3b38f9ffd8e954c0060ec6a62f141fbc6335354/roles/openshift_node/tasks/config.yml#L70) fetches MCS
+- [windows node for openshift](https://github.com/openshift/windows-machine-config-bootstrapper/blob/016f4c5f9bb814f47e142150da897b933cbff9f4/cmd/bootstrapper/initialize_kubelet.go#L33)
+
+#### Intersection with https://github.com/openshift/enhancements/pull/201
+
+In the future, we may also generate updated "bootimages" from the custom operating system container.
```
cybertron, line 254: Okay, I think that answers my questions above. The boot image customization would be a future feature (that we would be very interested in).

cybertron (commented):
Still working my way through this, but a few initial comments inline from a host networking perspective.

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,435 @@
...
+- [ ] The output is a container that can be pushed to a registry, inspected, and processed via security scanners. 
+- [ ] Further, it can be executed as a regular container (via podman or as a pod) *and* booted inside or outside of a cluster
+- [ ] Gain [transactional configuration+code changes](https://github.com/openshift/machine-config-operator/issues/1190) by having configuration+encode captured atomically in a container
+- [ ] Avoid breaking existing workflow via MachineConfig (including extensions)
+- [ ] Avoid overwriting existing custom modifications (such as files managed by other operators) during upgrades
+
+### Non-Goals
+
+- While the base CoreOS layer/ostree-container tools will be usable outside of OpenShift, this enhancement does not cover or propose any in-cluster functionality for exporting the forward-generated image outside of an OpenShift cluster. In other words, it is not intended to be booted (`$ rpm-ostree rebase <image>`) from outside of a cluster.
+- This proposal does not cover generating updated "bootimages"; see https://github.com/openshift/enhancements/pull/201
```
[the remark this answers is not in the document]
cgwalters: Everything that works today will continue to work.  We've invested a *ton* of time on the CoreOS side into handling all kinds of network configuration, and that all will continue to work.  Specifically, static IP addresses can (and probably should) still be configured via providing Ignition on a per-machine/node basis.

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,443 @@
...
+
+A month after that, the administrator wants to make a configuration change, and creates a `machineconfig` object targeting the `worker` pool.  This triggers a new image build.  But, the 3rd party yum repository is down, and the image build fails.  The operations team gets an alert, and resolves the repository connectivity issue.  They manually restart the build which succeeds.
+
+#### Kernel hotfix
+
+example.corp runs OCP on aarch64 on bare metal.  An important regression is found that only affects the aarch64 architecture on some bare metal platforms.  While a fix is queued for a RHEL 8.x z-stream, there is also risk in fast tracking the fix to *all* OCP platforms.  Because this fix is important to example.corp, a hotfix is provided via a pre-release `kernel.rpm`.
+
+The OCP admins at example.corp get a copy of this hotfix RPM into their internal data store, and configure a derived build where the kernel RPMs are applied. The MCO builds a derived image and rolls it out.
```
c3d, line 96: Could you contrast the "before/after" situation and outline benefits / drawbacks, notably in this scenario? I'm notably concerned with the override continuing to apply. IIUC, that's a "whole kernel" override, not just a given patch, so if there is a second hotfix with a wider distribution, how would the two combine?

enhancements/ocp-coreos-layering.md
cgwalters: OK, added a bit more about 1.

enhancements/ocp-coreos-layering.md
cgwalters: > I'm notably concerned with the override continuing to apply.

I pushed a bit about perhaps using a Bugzilla entry to know whether or not we still need the override.

cgwalters:
@mrunalp @aravindhp  Given this is all provisional, any further concerns before merging this?

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,479 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Change RHEL CoreOS as shipped in OpenShift to be a "base image" that can be used as in layered container builds and then booted.  This will allow custom 3rd party agents delivered via RPMs installed in a container build.  The MCO will roll out and monitor these custom builds the same way it does for the "pristine" CoreOS image today.
```
janosdebugs, line 37: Currently, AFAIK the licensing of CoreOS does not allow reusing it for other purposes (FIXME). Will this change with this proposal?

cgwalters: There is (to the best of my knowledge) no *licensing* restrictions - "RHEL CoreOS" is just not a separate thing from OpenShift.

I think this question is really addressed by https://github.com/openshift/enhancements/pull/1032/files#diff-d56db1c5d442467039e4ac4e6d118edd49185320a947906e8fd6973cbbd5f1abR60

But, the underlying non-OCP functionality of "boot a container" is in https://fedoraproject.org/wiki/Changes/OstreeNativeContainer and https://github.com/coreos/enhancements/blob/main/os/coreos-layering.md and will ship in the `rpm-ostree` in RHEL.  Truly productizing that is a whole separate thing.  I pushed https://github.com/openshift/enhancements/pull/1032/commits/d25534f6ecb1a27d76b7f542ac871fe31599cf75 related to that.

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,485 @@
...
+
+#### What works now continues to work
+
+An OpenShift administrator at example.corp is happily using OpenShift 4 (with RHEL CoreOS) in several AWS clusters today, and has only a small custom MachineConfig object to tweak host level auditing.  They do not plan to use any complex derived builds, and expect that upgrading their existing cluster continues to work and respect their small audit configuration change with no workflow changes.
+
+#### Adding a 3rd party security scanner/IDS
+
+example.bank's security team requires a 3rd party security agent to be installed on bare metal machines in their datacenter.  The 3rd party agent comes as an RPM today, and requires its own custom configuration.  While the 3rd party vendor has support for execution as a privileged daemonset on their roadmap, it is not going to appear soon. 
+
+After initial cluster provisioning is complete, the administrators at example.bank supply a configuration that adds a repo file to `/etc/yum.repos.d/agentvendor.repo` and requests installation of a package named `some-3rdparty-security-agent` as part of a container build.
```
Prashanth684, line 84: is it also possible to do it on an initial cluster install? by just doing an `openshift-install create manifests` and then adding this custom manifest with this config which MCO would rollout during install?

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,485 @@
...
+
+The build is successful, and is included in both the control plane (master) and worker pools, and is rolled out in the same way the MCO performs configuration and OS updates today.
```
Prashanth684, line 86: i'm wondering how this will be supported in the context of heterogeneous clusters as the internal image registry will not support manifestlisted images and builds doesn't support it either: https://github.com/openshift/enhancements/pull/1014/files#r797900391

cgwalters: Pushed a blurb on this

enhancements/ocp-coreos-layering.md
cgwalters: We could clearly use the bootstrap node to perform a build (e.g. `podman build`).  I think performance needs will drive us to do this because otherwise we'd be adding a whole other phase to rollout where we'd need to wait for the cluster to be up to do a build, to generate our container image to roll it out to all the nodes.

cgwalters: I suspect a number of administrators though will end up wanting to provide an *externally* built image; that's alluded to above.  A whole host of tradeoffs with thaat.

Prashanth684: hmm..in that case where would the builds be pushed though ? so that it can be picked up my the individual nodes' MCO?

cgwalters: We'd run a registry on the bootstrap node.

Prashanth684: rigth..externally built images is definitely an option but most of the time we have requests like - "i want to add this systemd unti/file to ignition " or just change some kargs for which this will be useful

cgwalters: Yep definitely, all that works via MachineConfig today and we intend to keep it working.

mrunalp:
/approve

(will keep it open for a couple more days to see if there more comments from others before we merge)

mrunalp (approved):

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,496 @@
...
+
+- While the base CoreOS layer/ostree-container tools will be usable outside of OpenShift, this enhancement does not cover or propose any in-cluster functionality for exporting the forward-generated image outside of an OpenShift cluster. In other words, it is not intended to be booted (`$ rpm-ostree rebase <image>`) from outside of a cluster.
+- This proposal does not cover generating updated "bootimages"; see https://github.com/openshift/enhancements/pull/201
+- Don't change existing workflow for RHEL worker nodes
+
+## Proposal
+
+**NOTE: Nothing in this proposal should be viewed as final.  It is highly likely that details will change.  It is quite possible that larger architectural changes will be made as well.**
+
+1. The `machine-os-content` shipped as part of the release payload will change format to the new "native ostree-container" format, in which the OS content appears as any other OCI/Docker container.
```
sferich888, line 68: @control-d I think this means we can now ship this image on registry.redhat.io (or the container catalog) and not just on Quay.

sferich888: I wonder if this does or would solve some of the CVE reporting issues we have with RHCOS today?

cgwalters: Yes, a big sub-goal here is that the container image (though it has some "ostree stuff" in there) otherwise looks like e.g. the UBI base image, and should be understandable to security scanners like Clair directly.

There's some details though around how we handle extensions - those would not be visible to scanners.

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,496 @@
...
+
+#### Upgrade flow
+
+See [OSUpgrades](https://github.com/openshift/machine-config-operator/blob/master/docs/OSUpgrades.md) for the current flow before this proposal.
+
+1. The "base image" will be part of the release image, the same as today.
+1. The CVO will replace a ConfigMap in the MCO namespace with the OS payload reference, as it does today.
+1. The MCO will update an `imagestream` object (e.g. `openshift-machine-config-operator/rhel-coreos`) when this ConfigMap changes.
+1. The MCO builds a container image for each `MachineConfigPool`, using the new base image
+1. Each machineconfig pool will also support a `custom-coreos` `BuildConfig` object and imagestream.  This build *must* use the `mco-coreos` imagestream as a base.   The result of this will be rolled out by the MCO to nodes.
```
bparees, line 132: so if i'm understanding it correctly, this is the spot at which a user/admin would define the image customizations they want to apply, by defining a buildconfig that points to a dockerfile that contains their additions, more or less?

picking up @fabiand's question from the mailing list:

> How is it imagined to include different packages from different vendors in the same base image?
Let's say we've got a HW vendor providing (proprietary) drivers for their GPU, hardware monitoring tools by the system vendor, and some IDS. Some of this will be provided by an Operator, other elements by a human. How are they expected to create a single new base image?

and @cgwalters answer:
> Hmm; in theory perhaps we could create an interface (like a CRD) that
allowed registering "build fragments" (much like MachineConfig) that
could all be gathered and merged into a single build.

is there another option here that we allow multiple user-provided buildconfigs, with some sort of ordering, where only the final build produces the actual coreos image that will be consumed by the cluster?  I don't know if it's strictly necessary, vs just aggregating all the customization logic into a single buildconfig/dockerfile, but i wanted to continue the thread.

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,496 @@
...
+- This proposal does not cover generating updated "bootimages"; see https://github.com/openshift/enhancements/pull/201
+- Don't change existing workflow for RHEL worker nodes
+
+## Proposal
+
+**NOTE: Nothing in this proposal should be viewed as final.  It is highly likely that details will change.  It is quite possible that larger architectural changes will be made as well.**
+
+1. The `machine-os-content` shipped as part of the release payload will change format to the new "native ostree-container" format, in which the OS content appears as any other OCI/Docker container.  
+  (In contrast today, the existing `machine-os-content` has an ostree repository inside a UBI image, which is hard to inspect and cannot be used for derived builds).  For more information, see [ostree-rs-ext](https://github.com/ostreedev/ostree-rs-ext/) and [CoreOS layering](https://github.com/coreos/enhancements/pull/7).
```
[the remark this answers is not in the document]
cgwalters: We actually today do ship some content in CoreOS that doesn't come from an RPM - systemd units and the like: https://github.com/coreos/fedora-coreos-config/tree/testing-devel/overlay.d

Doing that for binaries...we want to enable that for customers.  For us (e.g. to stop building kubelet via RPM) - definitely *possible* but has other ramifications.

enhancements/ocp-coreos-layering.md
```diff
@@ -0,0 +1,496 @@
...
+Another mitigation here may be to support peer-to-peer upgrades, or have the control plane host a "bootstrap registry" that just contains the pending OS update.
+
+#### Manifest list support
+
+We know we want heterogeneous clusters, right now that's not supported by the build and image stream APIs.
+
+
+#### openshift-install bootstrap node process
+
+A key question here is whether we need the OpenShift build API as part of the bootstrap node or not.  One option is to do a `podman build` on the bootstrap node.
```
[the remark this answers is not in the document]
cgwalters: > (reimplementing a lot of push/pull secrets, proxy configs, CAs, ICSP logic)

We already need to be able to pull container images on the bootstrap node with plain `podman`, and we already do.  So that covers CAs and ICSP too.

The push aspect is interesting, but I think the most likely thing here is to run a basic in-memory registry (possibly the in-cluster one, just in a non-production mode) and we can easily set things up to allow pushing to that.  Or we just use `oc image serve` etc.

enhancements/ocp-coreos-layering.md
cgwalters: I feel confident that: no matter what we pick, it will have feature gaps :smile:

This enhancement specifically didn't yet try to call out the details of how we propose the user interface to work, because I think we simply need time to iterate on it, provide a functioning preview and and learn what works before we can draw a roadmap into that area.

bparees: totally fair, just trying to get some thoughts in place for when you get to that point

mrunalp:
/lgtm

I think we have enough to merge this provisional enhancement. We can keep iterating and working through what the UI will look like and update the KEP.

zhouhao3:
@cgwalters Hi, I want to test this feature in OCP. I noticed that one of the main tasks of this proposal is to build a container image of RHEL CoreOS. Is this image built? I didn't find this image. Thanks!

cgwalters:
It is published but only for OCP developers right now at `registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest`.

Publishing this image officially is tracked by https://url.corp.redhat.com/d4e7ed8

PR #1034, opened 2022-02-15
Agent Installer Development Environment

lranjbar:
This enhancement is looking to define a baseline for the e2e testing framework for the new ephemeral agent installer project.

### DIFF
--- a/enhancements/agent-installer/development-environment.md
+++ b/enhancements/agent-installer/development-environment.md
@@ -94,8 +94,8 @@ Not applicable.
 Which means the risk of breaking things in the central framework now is larger and
 has a larger "splash zone."
 
-We can mitigate this for the new ephemeral agent installer by adding e2e tests into the
-dev-scripts repository that run the agent flows. The purpose of this would be to find 
+We can mitigate this for the new ephemeral agent installer by running e2e tests using the agent flows for changes to the
+dev-scripts repository. The purpose of this would be to find 
 breakages in the tests across the different teams.
 
 ** The dev-scripts repository clones the [metal-dev-env](https://github.com/metal3-io/metal3-dev-env) 
--- a/enhancements/agent-installer/development-environment.md
+++ b/enhancements/agent-installer/development-environment.md
@@ -15,9 +15,9 @@ approvers:
 api-approvers: # in case of new or modified APIs or API extensions (CRD, aggregated api servers, webhooks, finalizers)
   - N/A
 creation-date: 2022-02-10
-last-updated: 2022-03-29
+last-updated: 2022-04-01
 tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
-  - https://issues.redhat.com/browse/AGENT-20
+  - https://issues.redhat.com/browse/AGENT-71
 ---
 
 # Ephemeral Agent Installer Development Infrastructure
@@ -51,9 +51,9 @@ project while limiting the scope of modifications to them to the new project's r
 development and for automating the flows in Openshift CI.
 * Create predefined configurations to deploy Openshift in HA, compact and SNO topologies.
 * Create predefined network configurations for Openshift including but not limited IP type 
-(IPv4, IPv6, dual-stack), L2 segments and DHCP configurations.
-* A reusable environment configuration for development and testing teams to use with
-predefined flows for quicker onboarding.
+(IPv4, IPv6, dual-stack).
+* A reusable and reproducible environment configuration with predefined flows for development 
+and testing teams to use.
 * Convergence of the underlying scripts and environment setup code between Metal Platform (IPI) 
 team and the new Agent Installer teams.
 
@@ -71,11 +71,14 @@ components as needed for the new project.
 
 ### User Stories
 
-* As a developer, I need to set up an environment so that I can install a 
+* As a developer, I need to set up an environment so that I can install an Openshift
 cluster locally using the agent-based installer for development and testing
 
-* As a developer, I need to launch the built ephemeral agent based installer image so that I can 
-test and validate this image during development.
+* As a developer, I need to launch the built ephemeral agent based installer image so that 
+I can test and validate this image during development.
+
+* As a developer, I want my environment configuration to be reusable and reproducible so that 
+I can share it with my colleagues so that we can easily troubleshoot the same problems.
 
 * As a developer, I need to check cluster installation progress of the Openshift cluster being
 installed by agent installer and when this installation is complete to validate the Openshift
@@ -94,9 +97,9 @@ Not applicable.
 Which means the risk of breaking things in the central framework now is larger and
 has a larger "splash zone."
 
-We can mitigate this for the new ephemeral agent installer by running e2e tests using the agent flows for changes to the
-dev-scripts repository. The purpose of this would be to find 
-breakages in the tests across the different teams.
+We can mitigate this for the new ephemeral agent installer by running e2e tests using the 
+agent flows for changes to the dev-scripts repository. The purpose of this would be to find 
+breakages in the tests before they happen.
 
 ** The dev-scripts repository clones the [metal-dev-env](https://github.com/metal3-io/metal3-dev-env) 
 repository for most of its Ansible roles. Which means there is a risk of changes in this 
@@ -104,9 +107,12 @@ repository can break things in our repository as well. We have less control over
 than others inside Openshift.
 
 This is partially mitigated already because the repository is cloned at a specific
-Git SHA. To further this we can add a test in dev-scripts that runs maybe once or twice
-a week that clones the HEAD of this repository. We would be alerted for breakages more
-frequently and we can mitigate them on a case-by-case basis.
+Git SHA. Extra testing and sometimes changes are needed when this Git SHA pin for metal3-dev-env
+moves.
+
+**OPTIONAL SUGGESTION:** Add a test in dev-scripts that runs maybe once a week that clones 
+the HEAD of  metal3-dev-env repository. We would be alerted for breakages more frequently 
+and mitigate them more often.
 
 
 ## Design Details
@@ -117,7 +123,7 @@ frequently and we can mitigate them on a case-by-case basis.
 existing scripts and new functionality required by this new project.
 * New functionality we will write in Ansible, unless it is highly coupled with existing
 shell scripts.
-* We will create Anisble roles to wrap the existing code making it easier to use within
+* We will create Ansible roles to wrap the existing code making it easier to use within
 a pure Ansible solution
 * The existing bash scripts will be left as is. Refactoring them will be done piece by 
 piece as needed.
@@ -125,32 +131,67 @@ piece as needed.
 ### Phase 1 (Crawl): Add development scripts into dev-scripts for Agent Installer
 
 1) Add development scripts into dev-scripts for the new agent installer project using 
-existing conventions.
+existing conventions. [1]
 2) Add Ansible roles and tasks to wrap existing shell scripts to make them easier
 to integrate with an Ansible solution.
-3) Add base e2e test flow to run the ephemeral agent installer image in Openshift CI.
+3) Add base e2e test flow to run the ephemeral agent installer image in Openshift CI. [2]
 4) Write documentation for new developers joining the agent installer team.
 
 In this phase it should also become more clear what scripts and code are overlapping
-between the two teams.
+with the existing code base.
+
+[1]: The development scripts for agent installer will include the following scenarios:
+
+* Building the agent installer from a configurable Git SHA.
+* Setting up the environment for install and stopping. Allowing for a person to manually
+test an install.
+* Setting up the environment for install and automatically running the install using the
+agent installer.
+
+[2]: This will be run against the repository that the agent installer is being worked on.
+The base e2e test flow for agent installer is the following scenario:
+
+* Build agent installer
+* Set up environment for Openshift installation
+* Using the built agent installer start the Openshift installation
+* Report when the Openshift installation is complete
+* Run existing Openshift conformance tests and validations against the newly installed cluster
 
 ### Phase 2 (Walk): Define common environment default configurations for Agent Installer
 
-1) Define an Ansible role that will pipe in values to existing shell scripts and manage
+1) Add basic cluster validations to be ran by a local developer after the Openshift cluster
+is installed. [1]
+2) Define an Ansible role that will pipe in values to existing shell scripts and manage
 the environment variables needed for these scripts.
-2) Add in more configuration examples like 
+3) Add in more configuration examples like 
 [config_example.sh](https://github.com/openshift-metal3/dev-scripts/blob/master/config_example.sh) 
 for dev-scripts that define  common configurations for OCP installs. 
-3) Use the above examples to define default configurations for VMs (HA, Compact, SNO) 
+4) Use the above examples to define default configurations for VMs (HA, Compact, SNO) 
 and networks (IPV4, IPV6, dualstack) in Ansible.
-4) Identify minor refactors to existing scripts deemed helpful for the two projects and
+5) Identify minor refactors to existing scripts deemed helpful for the two projects and
 move them into Ansible.
+
+[1]: These local cluster validations include checks for the following: Kubeconfig is created and accessible, 
+Openshift Control Plane came up successfully, Openshift cluster is in the "Installed" state and
+finally a generic error capture for the errors that output when the Openshift cluster fails installation.
 
 ### Phase 3 (Run): Update existing e2e IPI flows to use the refactored Ansible in dev-scripts
 
 1) Update Agent Installer flows to use the new Ansible roles to configure the flows instead
 of using the environment variables directly.
 2) Update the e2e IPI flows to use this new Ansible configuration roles as needed.
+
+
+At the end of Phase 3 this enhancement should be considered complete. This enhancement is
+defining the base set of things for our development environment. Phase 4 includes the next steps 
+that we've already identified.
+
+### Phase 4 (Fly): Continue adding features, flows and configurations as needed for Agent Installer
+
+1) Add more advanced configurations for Agent Installer flows including: connected / disconnected,
+network configurations using multiple L2 segments and DHCP configurations.
+2) Add new scenarios and flows into the environment as new features require them.
+3) Continue to evaluate the existing code base for improvements.
 
 ### Open Questions [optional]
 
@@ -160,6 +201,14 @@ of using the environment variables directly.
 * On-board the base e2e flow into Openshift CI for agent installer.
 
 ### Graduation Criteria
+
+Not applicable.
+
+#### Dev Preview -> Tech Preview
+
+Not applicable.
+
+#### Tech Preview -> GA
 
 Not applicable.
 
@@ -175,10 +224,8 @@ Not applicable.
 
 * Currently the dev-scripts repository is not versioned. This will stay the same.
 
-* However for the e2e test scenarios themselves it makes sense to version these.
-Since the ephemeral agent based installer is targeted to be released with OCP it would 
-be easiest to version the e2e framework the same as OCP x.y. This is the strategy
-that openshift-tests takes.
+* For our test scenarios these will be versioned in openshift-tests following the existing
+versioning scheme.
 
 ### Operational Aspects of API Extensions
 
@@ -193,6 +240,8 @@ Not applicable.
 Not applicable.
 
 ## Implementation History
+
+TBD
 
 ## Drawbacks
 
--- a/enhancements/agent-installer/development-environment.md
+++ b/enhancements/agent-installer/development-environment.md
@@ -1,5 +1,5 @@
 ---
-title: ephemeral-agent-installer-e2e-tests
+title: development-environment
 authors:
   - "@lranjbar"
 reviewers:
--- a/enhancements/agent-installer/development-environment.md
+++ b/enhancements/agent-installer/development-environment.md
@@ -52,7 +52,7 @@ development and for automating the flows in Openshift CI.
 * Create predefined configurations to deploy Openshift in HA, compact and SNO topologies.
 * Create predefined network configurations for Openshift including but not limited IP type 
 (IPv4, IPv6, dual-stack), L2 segments and DHCP configurations.
-* A reusable Ansible environment configuration for development and testing teams to use with
+* A reusable environment configuration for development and testing teams to use with
 predefined flows for quicker onboarding.
 * Convergence of the underlying scripts and environment setup code between Metal Platform (IPI) 
 team and the new Agent Installer teams.
--- a/enhancements/agent-installer/development-environment.md
+++ b/enhancements/agent-installer/development-environment.md
@@ -1,0 +1,221 @@
+---
+title: ephemeral-agent-installer-e2e-tests
+authors:
+  - "@lranjbar"
+reviewers:
+  - "@andfasano"
+  - "@bfournie"
+  - "@hardys"
+  - "@pawanpinjarkar"
+  - "@rwsu"
+  - "@zaneb"
+approvers:
+  - "@dhellmann"
+  - "@celebdor"
+api-approvers: # in case of new or modified APIs or API extensions (CRD, aggregated api servers, webhooks, finalizers)
+  - N/A
+creation-date: 2022-02-10
+last-updated: 2022-03-29
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - https://issues.redhat.com/browse/AGENT-20
+---
+
+# Ephemeral Agent Installer Development Infrastructure
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Create a development environment for the ephemeral agent based installer image in the 
+[dev-scripts](https://github.com/openshift-metal3/dev-scripts) repository and refactor 
+components as needed for the new project. In addition we will add base case e2e test flows
+integrated into Openshift CI to get the project started.
+
+## Motivation
+
+The new agent based installer project will need a standard development environment for
+the team to use. We will need an e2e testing framework for validation in Openshift CI. 
+In this enhancement we will explain our desire to reuse existing frameworks for the new 
+project while limiting the scope of modifications to them to the new project's requirements.
+
+### Goals
+
+* Create a development environment for the agent installer project that can be used for local
+development and for automating the flows in Openshift CI.
+* Create predefined configurations to deploy Openshift in HA, compact and SNO topologies.
+* Create predefined network configurations for Openshift including but not limited IP type 
+(IPv4, IPv6, dual-stack), L2 segments and DHCP configurations.
+* A reusable Ansible environment configuration for development and testing teams to use with
+predefined flows for quicker onboarding.
+* Convergence of the underlying scripts and environment setup code between Metal Platform (IPI) 
+team and the new Agent Installer teams.
+
+### Non-Goals
+
+* Re-factoring dev-scripts and assisted-test-infra for reasons not related to the new 
+ephemeral agent based installer image project.
+* Overall convergence between all the installer development teams of scripts and e2e test flows.
+
+## Proposal
+
+Create an development environment for the ephemeral agent based installer image in the 
+[dev-scripts](https://github.com/openshift-metal3/dev-scripts) repository and refactor 
+components as needed for the new project.
+
+### User Stories
+
+* As a developer, I need to set up an environment so that I can simulate the installation of a 
+users first cluster using the ephemeral based agent installer image.
+
+* As a developer, I need to launch the built ephemeral agent based installer image so that I can 
+test and validate this image during development.
+
+* As a developer, I need to check cluster installation progress of the Openshift cluster being
+installed by agent installer and when this installation is complete to validate the Openshift
+cluster.
+
+* As a developer, I want to extract logs from agent installer and the Openshift cluster under
+install to troubleshoot and debug.
+
+### API Extensions
+
+Not applicable.
+
+### Risks and Mitigations
+
+** There will be now a third team using dev-scripts for slightly different purposes.
+Which means the risk of breaking things in the central framework now is larger and
+has a larger "splash zone."
+
+We can mitigate this for the new ephemeral agent installer by adding e2e tests into the
+dev-scripts repository that run the agent flows. The purpose of this would be to find 
+breakages in the tests across the different teams.
+
+** The dev-scripts repository clones the [metal-dev-env](https://github.com/metal3-io/metal3-dev-env) 
+repository for most of its Ansible roles. Which means there is a risk of changes in this 
+repository can break things in our repository as well. We have less control over this repository 
+than others inside Openshift.
+
+This is partially mitigated already because the repository is cloned at a specific
+Git SHA. To further this we can add a test in dev-scripts that runs maybe once or twice
+a week that clones the HEAD of this repository. We would be alerted for breakages more
+frequently and we can mitigate them on a case-by-case basis.
+
+
+## Design Details
+
+### General Strategy
+
+* Using the scope of the agent installer project we will identify possible refactors to 
+existing scripts and new functionality required by this new project.
+* New functionality we will write in Ansible, unless it is highly coupled with existing
+shell scripts.
+* We will create Anisble roles to wrap the existing code making it easier to use within
+a pure Ansible solution
+* The existing bash scripts will be left as is. Refactoring them will be done piece by 
+piece as needed.
+
+### Phase 1 (Crawl): Add development scripts into dev-scripts for Agent Installer
+
+1) Add development scripts into dev-scripts for the new agent installer project using 
+existing conventions.
+2) Add Ansible roles and tasks to wrap existing shell scripts to make them easier
+to integrate with an Ansible solution.
+3) Add base e2e test flow to run the ephemeral agent installer image in Openshift CI.
+4) Write documentation for new developers joining the agent installer team.
+
+In this phase it should also become more clear what scripts and code are overlapping
+between the two teams.
+
+### Phase 2 (Walk): Define common environment default configurations for Agent Installer
+
+1) Define an Ansible role that will pipe in values to existing shell scripts and manage
+the environment variables needed for these scripts.
+2) Add in more configuration examples like 
+[config_example.sh](https://github.com/openshift-metal3/dev-scripts/blob/master/config_example.sh) 
+for dev-scripts that define  common configurations for OCP installs. 
+3) Use the above examples to define default configurations for VMs (HA, Compact, SNO) 
+and networks (IPV4, IPV6, dualstack) in Ansible.
+4) Identify minor refactors to existing scripts deemed helpful for the two projects and
+move them into Ansible.
+
+### Phase 3 (Run): Update existing e2e IPI flows to use the refactored Ansible in dev-scripts
+
+1) Update Agent Installer flows to use the new Ansible roles to configure the flows instead
+of using the environment variables directly.
+2) Update the e2e IPI flows to use this new Ansible configuration roles as needed.
+
+### Open Questions [optional]
+
+### Test Plan
+
+* Test locally while creating the development environment.
+* On-board the base e2e flow into Openshift CI for agent installer.
+
+### Graduation Criteria
+
+Not applicable.
+
+#### Removing a deprecated feature
+
+Not applicable.
+
+### Upgrade / Downgrade Strategy
+
+Not applicable.
+
+### Version Skew Strategy
+
+* Currently the dev-scripts repository is not versioned. This will stay the same.
+
+* However for the e2e test scenarios themselves it makes sense to version these.
+Since the ephemeral agent based installer is targeted to be released with OCP it would 
+be easiest to version the e2e framework the same as OCP x.y. This is the strategy
+that openshift-tests takes.
+
+### Operational Aspects of API Extensions
+
+Not applicable.
+
+#### Failure Modes
+
+Not applicable.
+
+#### Support Procedures
+
+Not applicable.
+
+## Implementation History
+
+## Drawbacks
+
+The biggest argument against it is that by using dev-scripts as the base is we don't
+inherit the work that was done previously in validating Assisted Installer in 
+assisted-test-infra.
+
+## Alternatives
+
+The choices for making a new development environment for the agent installer project were:
+
+1) Make an entirely new framework
+2) Extend assisted-test-infra for agent-installer
+3) Extend dev-scripts for agent-installer (Chosen)
+
+Making a new framework (Option #1) was not chosen due to the amount of progress
+that would be lost by starting over. In existing frameworks we have solved a lot
+of problems for example: image registry mirroring, proxies, upgrades, etc. Redoing
+this work is not trivial. The risks were deemed likely to happen with this approach.
+
+As far as extending assisted-test-infra (Option #2) for agent installer was not pursued. 
+This is because the flow of the e2e tests in this framework first setup a minikube cluster.
+From the perspective of the agent-installer this is not needed. The beginning of the
+e2e test flow used in dev-scripts is significantly closer to the flow we desire for the
+agent installer.
+
--- a/enhancements/agent-installer/development-environment.md
+++ b/enhancements/agent-installer/development-environment.md
@@ -71,8 +71,8 @@ components as needed for the new project.
 
 ### User Stories
 
-* As a developer, I need to set up an environment so that I can simulate the installation of a 
-users first cluster using the ephemeral based agent installer image.
+* As a developer, I need to set up an environment so that I can install a 
+cluster locally using the agent-based installer for development and testing
 
 * As a developer, I need to launch the built ephemeral agent based installer image so that I can 
 test and validate this image during development.


enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
+---
+title: ephemeral-agent-installer-e2e-tests
```
romfreiman, line 2: So you're basically suggesting, that this time, this test framework, unlike all others, will be error prone, super easy to read and follow, debugable, easy to extend, collect result and operate. Of course not like others.
BTW, test infra started with the same goal. And dev scripts. And prow. And many others test frameworks as well.
The problem comes when you have to add features to it - static ips, ipxe, pxe?, vsphere, add workers, use cluster0 as a hub.... I wonder how then it will suit the suggested design pattern.
And, what will happen in a year, when someone else will come and suggest 4th test framework, because the other 3 are not readable, debugguble, too complex....

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
```
romfreiman, line 31: The advantage of test-infra that it being used also by the QE, which reduces the duplication of efforts. 
The advantage of devscripts is that it has virtual media integration and disconnected env (which is not relevant at this point of the agent, but might be in the future?)

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
...
+
+Create a new e2e testing framework for the new agent based installer project. 
+Currently, for this project there would be potentially two existing e2e frameworks 
+that could be used, dev-scripts and assisted-test-infra. However, both of these have 
+significant challenges in re-purposing them for this new project. 
+
+The main challenge with either of these existing frameworks is that the environment
+setup, installation and subsequent tests they run are all highly coupled together.
```
hardys, line 39: I'm not sure this is accurate - in dev-scripts these steps are all in different scripts, the ansible roles are decoupled and used from upstream, etc - what kind of coupling did you have in mind here?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
...
+
+#### Setup Environment for Install Without Installing OCP or k8s
+
+As a developer, I need to set up VMs so that I can simulate the installation of a users 
+first cluster using the ephemeral based agent installer image.
+
+Both existing frameworks today set up VMs using libvirt today to test a users OCP install.
+However, they also both immediately use those VMs to start the installation process. There
+isn't an easy way to stop the current frameworks from using IPI and Assisted Installer
+after the VM creation.
```
hardys, line 127: This is not accurate, in dev-scripts you can simply run [01_install_requirements.sh](https://github.com/openshift-metal3/dev-scripts/blob/master/01_install_requirements.sh) then [02_configure_host.sh](https://github.com/openshift-metal3/dev-scripts/blob/master/02_configure_host.sh)

From that point you can do whatever you want with the VMs, for example testing some new installer workflow (which could be added as a new script)

Another option if you really don't want to use the same scripts is to just consume the upstream [metal3-dev-env ansible roles](https://github.com/metal3-io/metal3-dev-env/tree/master/vm-setup) which we use inside dev-scripts - this might be preferable if you're aiming for a pure-ansible solution.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
...
+The first priority will be to run these playbooks locally but how to understand this in the CI
+is being taken into consideration.
+
+### Open Questions [optional]
+
+### Test Plan
+
+* e2e tests for the three install scenarios of cluster 0 (SNO, Compact, HA)
+* Run Openshift compliance tests against these installed clusters
+* Unit tests for new code to reach 50% code coverage
```
hardys, line 231: Are unit tests relevant here - it's going to be a bunch of ansible roles/tasks, right?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
...
+
+#### Launch a VM with the Ephemeral Agent Based Installer Image
+
+As a developer, I need to launch a VM with the built ephemeral agent based installer
+image so that I can test and validate this image.
+
+Both existing frameworks don't have an easy way to do this. Which makes sense because
+they predate the ephemeral agent based installer project. There also isn't an easy way
+to use an existing routine to just launch a generic ISO image which would have been
+something we could have re-purposed.
```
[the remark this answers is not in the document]
romfreiman: and test infra is basically does the iso boot with TF.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
zaneb: It would seem unfortunate indeed if we were to end up rewriting the vm-setup roles from metal3-dev-env.

This might even be an opportunity to start converting more of the first couple of scripts in dev-scripts to ansible, which would benefit both projects.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Drawbacks
+
+The biggest argument against it is that by creating a new framework we don't
+inherit the work that was done previously in validating Assisted Installer.
+
+## Alternatives
+
+Alternatively we can re-factor one of the existing frameworks to include this
+new project.
```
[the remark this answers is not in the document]
andfasano: I would tend to agree on this point. Even though dev-scripts has some complexity, it seems covering most of the side tasks that may be required - for that maybe it could help trying to identify as much as possible the existing gaps

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
dhellmann: The thing being proposed will automate the setup of infrastructure resources (libvirt, mirror registry, etc.), right? And then deploying the cluster and running any tests would be 2 steps that are independent of that setup? Maybe we need a different name than "e2e testing framework."

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
...
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Create a new e2e testing framework for the new agent based installer project. 
+Currently, for this project there would be potentially two existing e2e frameworks 
+that could be used, dev-scripts and assisted-test-infra. However, both of these have 
+significant challenges in re-purposing them for this new project. 
+
+The main challenge with either of these existing frameworks is that the environment
```
dhellmann, line 38: On a purely formatting note, it's nice to keep the "summary" section short (~1 paragraph). I think your first paragraph here is OK for that, and the rest of this section could move down to either the "motivation" or "proposal" sections.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
dhellmann: I agree that the roles are decoupled enough that they could be used independently of the deployment steps. There are other questions about their fitness that we could answer here, though.

Do they already do something close enough to what is needed that they could be used without other significant refactoring?

Is there any concern that reusing metal3-dev-env would tie this project to the expectations or assumptions of metal3 upstream?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
...
+#### Launch a VM with the Ephemeral Agent Based Installer Image
+
+As a developer, I need to launch a VM with the built ephemeral agent based installer
+image so that I can test and validate this image.
+
+Both existing frameworks don't have an easy way to do this. Which makes sense because
+they predate the ephemeral agent based installer project. There also isn't an easy way
+to use an existing routine to just launch a generic ISO image which would have been
+something we could have re-purposed. 
+
```
dhellmann, line 139: https://gist.github.com/beekhof/35137d9f8e93b4512d64e734110cd4aa is an example of another option not already considered here. For the agent-based installer, there's no particular reason to create the VMs before you're ready to boot them from the ISO.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
...
+#### Removing a deprecated feature
+
+### Upgrade / Downgrade Strategy
+
+### Version Skew Strategy
+
+Since the ephemeral agent based installer is targeted to be released with OCP it would 
+be easiest to version the e2e framework the same as OCP x.y. This is the strategy
+that openshift-tests takes.
```
dhellmann, line 255: Do you anticipate needing to have version numbers for these tools?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,275 @@
...
+- name: Run e2e-workflow Test steps
+  import_playbook: example-e2e-workflow_test.yml
+
+- name: Run e2e-workflow Post steps
+  import_playbook: example-e2e-workflow_test.yml
+```
+
+This is just a general idea and would like to keep this pretty open ended at the moment.
+The first priority will be to run these playbooks locally but how to understand this in the CI
+is being taken into consideration.
```
dhellmann, line 223: It would be useful to have a little more detail about what the tools need to do, to get an idea of how that might be broken up into roles and how much, if any, of what already exists would be useful to reuse.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
hardys: That would work, but FWIW the reason I'm suggesting reuse of the existing metal3-dev-env roles is those include support for more complex scenarios, e.g multiple disks, configurable support for multiple networks, enable/disable DHCP, etc - it's not just how to boot a VM with an ISO.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
dhellmann: I think based on the in-person conversations we've had, this paragraph can be dropped.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,329 @@
...
+
+These challenges will be explained in more detail in the User Stories section below.
+
+## Motivation
+
+The new agent based installer project will need an e2e testing framework for validation.
+We should look at the existing frameworks and understand what re-purposing them will 
+entail for this new project.
+
+This enhancement looks at the existing e2e frameworks dev-scripts and assisted-test-infra
```
dhellmann, line 57: Linking the names to the repos here would help readers not already familiar with those tools.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,329 @@
...
+This enhancement looks at the existing e2e frameworks dev-scripts and assisted-test-infra 
+closely since this is the previous work in this space. The motivation is to expand on
+what we have learned from those frameworks try to come up with a solution that we can 
+reuse for new installers without these challenges in the future.
+
+### Goals
+
+* This proposal will focus using this new e2e framework in the new ephemeral agent based 
+installer image project.
+* After discussing with the Metal Platform team we will be improving parts of dev-scripts
```
dhellmann, line 66: This bullet point is specific enough that it feels like the proposal, rather than a goal.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,329 @@
...
+## Proposal
+
+Create an e2e testing framework for the ephemeral agent based installer image.
+
+### User Stories
+
+#### Run Assisted Installer Outside k8s
+
+As a developer, I need to run Assisted Installer without installing a k8s platform first 
+so that I can test the installation of a users first cluster, cluster 0.
```
dhellmann, line 88: Is this user story still relevant? It feels very focused on an implementation detail, unlike the other stories.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,329 @@
...
+
+#### Pre Environment Setup 
+
+* The e2e framework will be a collection of Ansible playbooks and roles
+* Using roles to define reusable steps scoped to units of work
+* Use playbooks to set up common e2e workflows and chains for the project
+
+In discussions regarding this enhancement we identified the dev-scripts has components that are
+a part of the Pre Environment Setup and Environment Teardown phases of the framework. We will
+move forward with adding and potentially refactoring scripts in this repository for these phases.
```
dhellmann, line 182: Did those discussions result in identifying any specific changes needed before dev-scripts could be used?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,329 @@
...
+
+Currently there is a `make clean` flow for starting new locally. We will need to add similar
+functionally for the flows that will be used for the agent-based installer
+
+### Open Questions [optional]
+
+### Test Plan
+
+* e2e tests for the three install scenarios of cluster 0 (SNO, Compact, HA)
+* Run Openshift compliance tests against these installed clusters
```
dhellmann, line 273: If 2 teams will be using dev-scripts for slightly different purposes, we will want test jobs to run on changes in the dev-scripts repo to ensure neither team accidentally breaks the workflow of the other.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,329 @@
...
+## Alternatives
+
+~~Alternatively we can re-factor one of the existing frameworks to include this
+new project.~~
+
+Based on the feedback for this enhancement we will be extending parts of dev-scripts
+to include this new project. However, as it is outlined above not all components of
+this e2e framework will land in the dev-scripts repository. Specifically the components
+for the Build and Test phases will land in other repositories.
+
```
dhellmann, line 322: There were basically 3 options originally, right? Modify dev-scripts, modify test-infra, or create something completely new. The 2 options that were rejected should be described here briefly, including the reasons for not choosing them.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: I feel like this implementation detail is important to point out. It cuts to the root of the chicken/egg problem that we have with the existing Assisted Installer.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: Agreed. To be clear this is now 3 teams too :) So this risk is something to keep an eye on.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: I think in the most recent rewrite I clarified this. I don't think the e2e framework should have a seperate versioning scheme. Rather the e2e tests will need to match the different versions of the agent installer and OCP. Since not all features are available in OCP, etc.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: Yes I took a stab at pointing these out.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,337 @@
...
+* Re-factoring dev-scripts and assisted-test-infra for reasons not related to the new 
+ephemeral agent based installer image project.
+* There should be some overlap in the pieces that are refactored for use in the new project.
+However it is still a non-goal for outside installer to use this new framework in their e2e 
+tests for the upcoming release. We will plan to create a point of convergence after the
+ephemeral agent based installer project is out of prototyping and MVP phase.
+
+## Proposal
+
+Create an e2e testing framework for the ephemeral agent based installer image.
```
dhellmann, line 67: Reading through this document, I'm having trouble understanding what the actual plan is. I think there is a plan, but the details aren't making it onto the page in a way that would allow someone else to help with the work.

You could expand this section to make it clear that the plan is to update dev-scripts so it can be used to create dev/test setups and then to manage any new test suite that is created separately from those scripts. This would be a good place to talk about any specific changes that need to be made, too.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,337 @@
...
+
+The direction of this proposal is now to extend and refactor parts of dev-scripts
+for the agent-based installer.
+
+** There will be now a third team using dev-scripts for slightly different purposes.
+Which means the risk of breaking things in the central framework now is larger and
+has a larger "splash zone."
+
+We can mitigate this for the new ephemeral agent installer by adding tests into the
+dev-scripts repository itself. The purpose of this would be to find breakages in the
```
dhellmann, line 131: I think what we want, rather than tests in the repo, is to run the test jobs that **use** dev-scripts to test other components when PRs are submitted against the dev-scripts repo. For example, we would run an e2e metal IPI job so that we know changes to dev-scripts to support the agent-installer don't break IPI support. And, eventually, we would want a similar e2e job using the agent-installer to ensure that changes to support IPI don't break the agent-installer.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,337 @@
...
+tests across the different teams.
+
```
dhellmann, line 133: Another risk is that dev-scripts relies on an upstream repo, metal3-dev-env, that isn't completely within our control and changes there can break downstream jobs. I don't think that has happened very often, but see https://coreos.slack.com/archives/C01RJHA6BRC/p1646755848747089 for one example.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,337 @@
...
+We will be including existing e2e tests that have been created to validate the OCP cluster
+installed by the new agent based installer.
+
+#### Teardown
+
+Currently there is a `make clean` flow for starting new locally. We will need to add similar
+functionally for the flows that will be used for the agent based installer
+
+### Identified Overlap Points for Agent Installer
```
dhellmann, line 246: Understanding the existing overlap is a good starting place. I'm interested in understanding more about the refactoring proposed here, though, and especially any gaps that need to be closed. What does dev-scripts need to do in the future to support the agent-based installer that it doesn't do today?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
dhellmann: This section is still not accurate. We've established that at least dev-scripts has stages and that those stages can be run independently so that a cluster isn't deployed via IPI.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,337 @@
...
+
+#### Launch a VM with the Ephemeral Agent Based Installer Image
+
+As a developer, I need to launch a VM with the built ephemeral agent based installer
+image so that I can test and validate this image.
+
+Both existing frameworks don't have an easy way to do this. Which makes sense because
+they predate the ephemeral agent based installer project. There also isn't an easy way
+to use an existing routine to just launch a generic ISO image which would have been
+something we could have re-purposed.
```
dhellmann, line 113: This is a good example of a change we'll need to make that could be called out in the proposal or details sections.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,337 @@
...
+        defaults/
+        meta/
+    example_role/
+        tasks/
+        defaults/
+        meta/
+```
+
+This is pretty open ended and meant to be. I just believe that focusing on making
+reusable roles here for the steps is key.
```
dhellmann, line 184: Is this level of refactoring a prerequisite for making dev-scripts do something useful for the agent-based installer team? Or is this a longer term goal to clean up the current implementation?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,337 @@
...
+#### Failure Modes
+
+#### Support Procedures
+
+## Implementation History
+
+## Drawbacks
+
+The biggest argument against it is that by creating a new framework we don't
+inherit the work that was done previously in validating Assisted Installer.
```
dhellmann, line 300: Is this talking about creating a new test suite, or reusing dev-scripts instead of the AI test infra repo?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: Unsure of how to reword this paragraph to get this across. The existing e2e metal IPI jobs and e2e assisted ztp jobs are commonly referred to as tests. These tests though are exercising dev-scripts to do the thing.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: Doesn't happen often but has happened in the past year maybe 2-3 times from the Assisted Installer front, I believe.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: Offline we  have identified:

- Making a new make target for agent-installer
- Adding scripts for agent-installer to use those routines

Is this the level of detail you are looking for?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: My plan is to add the new things for agent-installer using these patterns. In some cases that would mean a refactoring of things in the current implementation in the future. Iterative improvements over time.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: It's talking about creating a new test suite.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: I have removed this paragraph.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: Made another pass at updating the document to reflect we are using the dev-scripts repo for this.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+* The Assisted Installer e2e tests that run in assisted-test-infra installs minikube first 
+and then runs Assisted Installer on top of it.
+
+Which means currently in both cases we first setup some form of k8s first. Given
+the goal of the ephemeral agent based installer project to install a users first cluster, 
+cluster 0, this is highly undesirable.
+
+**Implementation Plan:**
+
+* Create a new routine to run Assisted Installer outside of k8s.
```
zaneb, line 115: Agent-based Installer?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+
+In discussions regarding this enhancement we identified the dev-scripts has components that are
+a part of the Pre Environment Setup and Environment Teardown phases of the framework. We will
+move forward with adding and potentially refactoring scripts in this repository for these phases.
+
+#### Build
+
+The scripts/automation for building the agent-based installer should be inside the repository
+where the installer is being worked on and not dev-scripts. We will reference the built 
+container images and ISO in dev-scripts from those repositories. Currently this project is being
```
zaneb, line 196: How can we substitute a container image with a local build for development?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+logic will land in dev-scripts alongside the pre environment setup steps.
+
+#### Tests
+
+For the actual test scenarios themselves we should use Ginkgo/Gomega. This is 
+similar to the rest of Openshift. Well most of (if not all) the product is written in Golang
+so it makes sense to use a Golang framework for these.
+
+There is a repository created as a preliminary place to put these. Plan to use a similar
+structure to openshift-tests in this repository.
```
zaneb, line 211: Would the long-term plan be to move this to openshift-tests once we move the code being tested into the openshift/ org?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Open Questions [optional]
+
+### Test Plan
+
+* e2e tests for the three install scenarios of cluster 0 (SNO, Compact, HA)
+* Run Openshift compliance tests against these installed clusters
+* e2e tests in dev-scripts to cover these new flows made for agent installer
+
+### Graduation Criteria
```
zaneb, line 300: I think this whole section is N/A, since we don't ship test frameworks.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+* There should be some overlap in the pieces that are refactored for use in the new project.
+However it is still a non-goal for outside installer to use this new framework in their e2e 
+tests for the upcoming release. We will plan to create a point of convergence after the
+ephemeral agent based installer project is out of prototyping and MVP phase.
+
+## Proposal
+
+Create an e2e testing framework for the ephemeral agent based installer image in the 
+[dev-scripts](https://github.com/openshift-metal3/dev-scripts) repository and refactor 
+components as needed for the new project.
```
rwsu, line 71: I would suggest adding more details about the framework being proposed here. I had to scan down the document and saw there are more specific information in the Design Details. Being able to see what the framework being proposed looks like here, will aid in helping understand how they impact the User Stories discussed in the next section.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+* The Assisted Installer e2e tests that run in dev-scripts first use the IPI installer 
+to install OCP and then the infrastructure-operator on this OCP cluster making it a hub 
+cluster.
+
+* The Assisted Installer e2e tests that run in assisted-test-infra installs minikube first 
+and then runs Assisted Installer on top of it.
+
+Which means currently in both cases we first setup some form of k8s first. Given
+the goal of the ephemeral agent based installer project to install a users first cluster, 
+cluster 0, this is highly undesirable.
```
rwsu, line 111: Is there an existing e2e test where you are given a clean host with no cluster installed and then can then proceed to do whatever you want with it?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
rwsu, line 115: Would building the ISO happen here?

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+
+**Implementation Plan:**
+
+* Create a new routine to run Assisted Installer outside of k8s.
+
+* Create a new routine to launch a VM with the built ephemeral agent based installer
+image.
```
rwsu, line 118: What about routines to:
1. Check cluster installation progress
2. Check cluster installation is successful
3. Validate kubeconfig is available and it can be used to access newly formed cluster
4. Extract logs

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Create a new e2e testing framework for the new agent based installer project.
```
rwsu, line 33: I'm a bit confused. Is a new e2e testing framework still being proposed? The "Alternatives" section says it was abandoned and the plan is to extend dev-scripts.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+** Setting up VMs, networks, locations of images, image registries, etc
+* Build: 
+** Building the agent-based installer ISO and container images associated with it
+* Test:
+**  Test scenarios with pass/failure outputs
+* Reporting:
+** Reports comprised of test results
+* Environment Teardown
+** Removing VMs, networks, etc to start fresh again
+
```
rwsu, line 181: The parts that are likely to be reused in the framework would be pre-environment setup and teardown.  I would focus most of the time here to make them easy to customize and reuse. Each project would tend to having specific requirements for build, test, and reporting. Having a generic way to be able to plug those in would be a big help.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
andfasano: It doesn't seem a very common case, also per my experience. As previously mentioned, the metal3-dev-env pin and devscripts CI jobs help in mitigating such risk

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
andfasano: I don't see any particular blocking reason for not adopting it since the very beginning. Moreover, `openshift-tests` can be easily used in dev environment as a test runner.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+- name: Run e2e-workflow Pre steps
+  import_playbook: example-e2e-workflow_pre.yml
+
+- name: Run e2e-workflow Test steps
+  import_playbook: example-e2e-workflow_test.yml
+
+- name: Run e2e-workflow Post steps
+  import_playbook: example-e2e-workflow_test.yml
+```
```
andfasano, line 279: I'm missing the reason for this playbook (which seems to overlap somehow the ci-operator)? Why do you need this full stuff locally and not reuse all the resources provided by the CI?
Note that in any case it's possible to simply run the e2e locally via `openshift-tests`, once a cluster has been deployed (and I guess it has alreay been assumed that dev-scripts will cover the cluster deployment phase somehow)

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+
+**Parameter Passing and Defaults for Openshift CI**
+
+In Openshift CI the way we pass parameters is through the use of environment variables. 
+This means all of  these Ansible roles will need to have defaults and be aware of those 
+environment variables.
```
andfasano, line 285: `dev-scripts` offers an already quick baked-in integration via the the `DEVSCRIPTS_CONFIG` var (see for example [1]). This makes it easy and convenient configuring any job, in additional to the default e2e step default configuration (see [2]). So, as long as the new ansible playbook will be capable to consume the required environment variables then the integration could be performed in a quite direct way.

[1] https://github.com/openshift/release/blob/6ceedc2963da1362f693a4cf7e0ffb8b62a451a6/ci-operator/step-registry/baremetalds/e2e/ovn/ipv6/baremetalds-e2e-ovn-ipv6-workflow.yaml#L5
[2] https://github.com/openshift/release/blob/6ceedc2963da1362f693a4cf7e0ffb8b62a451a6/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh#L99-L104

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: If we can just plug directly into `openshift-tests` this seems like a good plan to me.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: Yes. We would need to build our ISO to test it here.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: Reworded this to only include the two options not chosen.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+- [ ] Test plan is defined
+- [ ] Operational readiness criteria is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+## Summary
+
+Create a new e2e testing framework for the new agent based installer project. 
+Currently, for this project there would be potentially two existing e2e frameworks 
+that could be used, [dev-scripts](https://github.com/openshift-metal3/dev-scripts)
```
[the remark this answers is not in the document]
lranjbar: Yeah I agree with these statements. I reworded this enhancement to drop the e2e testing framework and just focus on the development environment part. There is some confusion being created because there are three things that need to happen:

1) Create development environment
2) Onboard into Openshift CI
3) Add in e2e tests for agent installer

Really this enhancement is mostly focused on 1 and 2. As far as 3 goes I think we can just add some base cases and some of the existing Openshift suites that validate our installed cluster. If we need more e2e past that we can extend from there.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
```diff
@@ -0,0 +1,373 @@
...
+### Risks and Mitigations
+
+The direction of this proposal is now to extend and refactor parts of dev-scripts
+for the agent-based installer.
+
+** There will be now a third team using dev-scripts for slightly different purposes.
+Which means the risk of breaking things in the central framework now is larger and
+has a larger "splash zone."
+
+We can mitigate this for the new ephemeral agent installer by adding e2e tests into the
```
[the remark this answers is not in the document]
lranjbar: Here I mean we can run the e2e test flow that use dev-scripts as a base against PRs in dev-scripts repo to catch potential breakages in those flows due to changes in dev-scripts. It is rare but has happened in the past.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: I took this into account and put in more information about file names and things I've identified in the file structure part. I was hesitant before because those things may change when they get into code review in dev-scripts. I think though this will help clarify things a lot so I added it in.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: Not really at this point. You could do this by running the scripts in dev-script one by one yourself and stopping before 06_create_cluster.sh. I've done that before but it is a little unwieldy to do.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: I removed this part entirely. In general the pattern I was describing is that the developer would build the container images they want to test themselves. We would probably have scripts and/or Makefile targets to make that easy. But the root of it is that you might have several branches and your want to build different containers on the fly for your changes. You want more control than a git clone allows. If this makes sense.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: I added these as user stories.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: This is a pattern that I've copied and reused in different parts of Openshift CI. It does have a limitation in causing large enumerations of settings which is why I added a couple of other steps in the Assisted folder. But overall it is a pattern I really like. :)

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Refactored File and Folder Structure
+
+The following is describing general pattern of how to organize these things in the dev-scripts 
+repository. These files and folders will be added alongside existing scripts in the dev-scripts 
+repository. This example includes things that have been already identified as overlaps between 
+the two teams that will need to be added or refactoring. Things may change names or locations
+during the code review process this folder structure is added for clarification.
+
+**File and Folder Structure:**
```
andfasano, line 162: I think that here there could be the risk to over-engineering the solution, considering that in future things may evolve differently (for example, integration within the openshift installer). In addition there are parts that seem to overlap with already existing frameworks such as the `ci-operator` and already defined steps at OpenShift level (ie ci-operator steps registry)

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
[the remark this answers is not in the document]
andfasano: Ok, in any case I think this is a fundamental pre-requisites given that the `dev-scripts` it's pushed in its stream so that other CI component repo could be used

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
andfasano: I'm not sure if I'm missing something, but the content of `DEVSCRIPTS_CONFIG` is converted to a number of env var. Do you maybe mean to group particular configurations under a single var?

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Proposal
+
+Create an development environment for the ephemeral agent based installer image in the 
+[dev-scripts](https://github.com/openshift-metal3/dev-scripts) repository and refactor 
+components as needed for the new project.
+
+### User Stories
+
+* As a developer, I need to set up an environment so that I can simulate the installation of a
```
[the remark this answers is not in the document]
hardys: I'd say something like "emulate on-premise customer environments (including various network and server configurations), to test installation using the ephemeral agent based installer"

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,305 @@
...
+
+The direction of this proposal is now to extend and refactor parts of dev-scripts
+for the agent-based installer.
+
+** There will be now a third team using dev-scripts for slightly different purposes.
+Which means the risk of breaking things in the central framework now is larger and
+has a larger "splash zone."
+
+We can mitigate this for the new ephemeral agent installer by adding e2e tests into the
+dev-scripts repository that run the agent flows. The purpose of this would be to find
```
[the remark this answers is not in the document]
hardys: I think it means add e2e tests that run on PRs proposed to the dev-scripts repository

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,305 @@
...
+
+** The dev-scripts repository clones the [metal-dev-env](https://github.com/metal3-io/metal3-dev-env) 
+repository for most of its Ansible roles. Which means there is a risk of changes in this 
+repository can break things in our repository as well. We have less control over this repository 
+than others inside Openshift.
+
+This is partially mitigated already because the repository is cloned at a specific
+Git SHA. To further this we can add a test in dev-scripts that runs maybe once or twice
+a week that clones the HEAD of this repository. We would be alerted for breakages more
+frequently and we can mitigate them on a case-by-case basis.
```
hardys, line 122: So far we haven't found this additional testing to be necessary, but we do sometimes have to make changes to dev-scripts when moving the metal3-dev-env pin (for example increasing ansible/python version)

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,305 @@
...
+2) Add base e2e test flow to run the ephemeral agent installer image in Openshift CI.
+3) Write documentation for new developers joining the agent installer team.
+
+In this phase it should also become more clear what scripts and code are overlapping
+between the two teams.
+
+### Phase 2 (Walk): Update shell scripts to Ansible in dev-scripts and use them for Agent Installer
+
+1) Convert the shell scripts [01_install_requirements.sh](https://github.com/openshift-metal3/dev-scripts/blob/master/01_install_requirements.sh) and [02_configure_host.sh](https://github.com/openshift-metal3/dev-scripts/blob/master/02_configure_host.sh) 
+into Ansible roles and playbooks for reuse.
```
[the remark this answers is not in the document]
hardys: Yeah this seems desirable, but unrelated to the agent integration

enhancements/agent-installer/development-environment.md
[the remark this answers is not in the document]
lranjbar: Yes that's what I mean. Just put the e2e tests that use dev-scripts as a base on the dev-scripts repo so that PRs into dev-scripts don't break them.

enhancements/agent-installer/development-environment.md
zaneb: We'll need a separate step to https://github.com/openshift/release/blob/master/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh (just running `make` won't cut it), so what's the harm in putting most of the configuration for each topology/network scenario into the dev-scripts repo where developers can also make use of it, and having just 1 configuration option (i.e. which scenario to choose) in the ci-operator steps?

enhancements/agent-installer/development-environment.md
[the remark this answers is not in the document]
lranjbar: I'll put in the adjective change here. :) 

But, just to give some context the use of simulate vs emulate is probably because of my hardware background. Hardware emulation vs Software simulation is the context I'm used to having these words. In that context these are pretty different things despite both being software programs. What we are doing with the VM environment for installation is close to "a software based simulation of a customers installation environment." That's how I would describe it... but that isn't too far off from Steve's wording either.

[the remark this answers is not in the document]
zaneb: It's a real installation into a simulated environment (VMs and linux bridges instead of servers and switches).

enhancements/agent-installer/development-environment.md
lranjbar: This is the route I went with on the rewording of it. Closer to the intention of it all as well.

enhancements/agent-installer/development-environment.md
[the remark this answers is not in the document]
lranjbar: This was cut in the rewording of this section :)

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,305 @@
...
+frequently and we can mitigate them on a case-by-case basis.
+
+
+## Design Details
+
+### Phase 1 (Crawl): Add development scripts into dev-scripts for Agent Installer
+
+1) Add development scripts into dev-scripts for the new agent installer project using 
+existing conventions.
+2) Add base e2e test flow to run the ephemeral agent installer image in Openshift CI.
```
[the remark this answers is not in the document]
lranjbar: I left this open because we are still deciding where things go. Right now dev-scripts and fleeting for sure. Looks like we will soon add another to this list now. ;)

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,305 @@
...
+
+* As a developer, I need to launch the built ephemeral agent based installer image so that I can test 
+and validate this image during development.
+
+* As a developer, I need to install Openshift using the ephemeral agent based installer image 
+to validate the agent installer.
+
+* As a developer, I need to check cluster installation progress of the Openshift cluster being
+installed by agent installer and when this installation is complete to validate the Openshift
+cluster.
```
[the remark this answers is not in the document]
lranjbar: It's not really included in how you write a user story but I plan to use the existing Openshift tests to validate the installed cluster.

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,305 @@
...
+### User Stories
+
+* As a developer, I need to set up an environment so that I can simulate the installation of a 
+users first cluster using the ephemeral based agent installer image.
+
+* As a developer, I need to launch the built ephemeral agent based installer image so that I can test 
+and validate this image during development.
+
+* As a developer, I need to install Openshift using the ephemeral agent based installer image 
+to validate the agent installer.
```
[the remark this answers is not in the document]
lranjbar: Yeah I cut out some of them since it seemed redundant.

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
+---
+title: ephemeral-agent-installer-e2e-tests
```
dhellmann, line 2:
```suggestion
title: development-environment
```

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+In this enhancement we will explain our desire to reuse existing frameworks for the new 
+project while limiting the scope of modifications to them to the new project's requirements.
+
+### Goals
+
+* Create a development environment for the agent installer project that can be used for local
+development and for automating the flows in Openshift CI.
+* Create predefined configurations to deploy Openshift in HA, compact and SNO topologies.
+* Create predefined network configurations for Openshift including but not limited IP type 
+(IPv4, IPv6, dual-stack), L2 segments and DHCP configurations.
```
dhellmann, line 54: Will we say the enhancement isn't fully implemented until all of these profiles/configurations are available? Or is the goal to enable them to be created over time?

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+* A reusable Ansible environment configuration for development and testing teams to use with
```
dhellmann, line 55:
```suggestion
* A reusable environment configuration for development and testing teams to use with
```

Let's not pre-select the implementation tool in the goal.

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,305 @@
...
+in dev-scripts test flows the beginning steps are very similar to the development
+flows we desire for the agent installer. In contrast the assisted-test-infra test flows
+set up extra things in the development environment that is undesirable for the agent installer.
+
+### Goals
+
+* Using this development environment in the new ephemeral agent based installer image project.
+* Re-factoring parts of dev-scripts that are needed for this ephemeral agent based installer.
+* Convergence of the underlying scripts and environment setup code between Metal Platform (IPI) 
+team and the new Agent Installer teams.
```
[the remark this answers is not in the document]
dhellmann: I think this thread can be resolved?

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+## Proposal
+
+Create an development environment for the ephemeral agent based installer image in the 
+[dev-scripts](https://github.com/openshift-metal3/dev-scripts) repository and refactor 
+components as needed for the new project.
+
+### User Stories
+
+* As a developer, I need to set up an environment so that I can simulate the installation of a 
+users first cluster using the ephemeral based agent installer image.
```
dhellmann, line 75:
```suggestion
* As a developer, I need to set up an environment so that I can install a 
cluster locally using the agent-based installer for development and testing
```

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+
+* As a developer, I need to launch the built ephemeral agent based installer image so that I can 
+test and validate this image during development.
+
+* As a developer, I need to check cluster installation progress of the Openshift cluster being
+installed by agent installer and when this installation is complete to validate the Openshift
+cluster.
+
+* As a developer, I want to extract logs from agent installer and the Openshift cluster under
+install to troubleshoot and debug.
```
dhellmann, line 85: Will this goal influence the design of the solution?

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+Not applicable.
+
+### Risks and Mitigations
+
+** There will be now a third team using dev-scripts for slightly different purposes.
+Which means the risk of breaking things in the central framework now is larger and
+has a larger "splash zone."
+
+We can mitigate this for the new ephemeral agent installer by adding e2e tests into the
+dev-scripts repository that run the agent flows. The purpose of this would be to find
```
dhellmann, line 98:
```suggestion
We can mitigate this for the new ephemeral agent installer by running e2e tests using the agent flows for changes to the
dev-scripts repository. The purpose of this would be to find 
```

enhancements/agent-installer/development-environment.md
dhellmann: Let's not sign up for more work than we need. I like the idea of keeping this as a suggestion, but not a commitment of something we'll do.

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+piece as needed.
+
+### Phase 1 (Crawl): Add development scripts into dev-scripts for Agent Installer
+
+1) Add development scripts into dev-scripts for the new agent installer project using 
+existing conventions.
+2) Add Ansible roles and tasks to wrap existing shell scripts to make them easier
+to integrate with an Ansible solution.
+3) Add base e2e test flow to run the ephemeral agent installer image in Openshift CI.
+4) Write documentation for new developers joining the agent installer team.
```
dhellmann, line 132: Documentation in dev-scripts?

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+
+This is partially mitigated already because the repository is cloned at a specific
+Git SHA. To further this we can add a test in dev-scripts that runs maybe once or twice
+a week that clones the HEAD of this repository. We would be alerted for breakages more
+frequently and we can mitigate them on a case-by-case basis.
+
+
+## Design Details
+
+### General Strategy
```
dhellmann, line 114: I like the phase breakdown.

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+Not applicable.
+
+### Version Skew Strategy
+
+* Currently the dev-scripts repository is not versioned. This will stay the same.
+
+* However for the e2e test scenarios themselves it makes sense to version these.
+Since the ephemeral agent based installer is targeted to be released with OCP it would 
+be easiest to version the e2e framework the same as OCP x.y. This is the strategy
+that openshift-tests takes.
```
dhellmann, line 181: Do you expect to need new tests? Or a new test job that uses the agent-based installer to set up the cluster and then run the existing e2e tests?

dhellmann (commented):
This is making good progress. I have a few wording suggestion changes and a couple of clarifying questions.

enhancements/agent-installer/development-environment.md
lranjbar: My view is that we absolutely need this set because this is most the common configurations we use for OCP installs. The two that aren't: L2 segments and DHCP are linked in with work that is planned for Agent Installer. So it is fair to say we aren't complete until we get these in.

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: An update to this comment. We decided to go with a more piece by piece flow for some of these things. This is now described in the design details section

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: This question got lots in the shuffle. I meant that the Ansible playbooks need to use the same environment variables as `DEVSCRIPTS_CONFIG` and `ASSISTED_CONFIG` where it makes sense. [1]

[1]: https://github.com/openshift/assisted-service/blob/master/deploy/operator/ztp/assisted-installer-crds-playbook.yaml#L7-L31

enhancements/agent-installer/development-environment.md
lranjbar: Yes. It just means add new documentation for the new things we are adding for agent installer. How to use dev-scripts to setup your development environment to write/test agent installer. If we don't want to call it out I can remove it.

enhancements/agent-installer/development-environment.md
lranjbar: For this enhancement the strategy I'm suggesting is the latter. If we find we need new test scenarios we can revise the enhancement to include them. Reusing what we have already is more of a baseline we need to get to before adding new tests.

enhancements/agent-installer/development-environment.md
dhellmann: Yeah, I'm not questioning whether need them. I'm questioning whether we would say "we have neither a development environment tool nor CI" until those are written.

It's safe to break this work down into more steps, and for this enhancement to say the goal is to provide 1 basic configuration and then enable us to build those "advanced" configurations later. You wouldn't need separate enhancements for the later work, but you might track them in separate tickets.

enhancements/agent-installer/development-environment.md
dhellmann: It's a good thing to call out, I was just curious where those docs would go.

enhancements/agent-installer/development-environment.md
lranjbar: It shouldn't influence the design of the development environment. The development environment would just be using existing API calls against assisted-service, must-gather against the Openshift cluster, etc to get this information.

enhancements/agent-installer/development-environment.md
lranjbar: Okay I'll pair this one down to be a suggestion.

enhancements/agent-installer/development-environment.md
dhellmann: OK. You can leave it in, but generally we want to limit the goals list to those that apply directly to the planned work so we can map the proposed changes back to the goals and make sure we're meeting them all.

enhancements/agent-installer/development-environment.md
lranjbar: Ahh okay I understand the question now. 🤔 I'll split them into different phases. It isn't too hard to do most of them because a lot of this already defined in Openshift CI

enhancements/agent-installer/development-environment.md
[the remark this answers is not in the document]
andfasano: I think that (in the long term) the number of configurations * OpenShift versions * e2e conformance suits required for an exhaustive vaidation will make it impractical to verify it locally, leaving this responsibility to the CI rehearsals.
During the development activity I'd like to have a quick / gross way to verify the current changes, so probably we could identify a number of minimum must-have checks required - and possibly common to all the possible configurations (ie cluster is up), but that's probably something we could leave as an implementation details and here just leave "verify that installation is complete"

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+
+1) Add development scripts into dev-scripts for the new agent installer project using 
+existing conventions.
+2) Add Ansible roles and tasks to wrap existing shell scripts to make them easier
+to integrate with an Ansible solution.
+3) Add base e2e test flow to run the ephemeral agent installer image in Openshift CI.
+4) Write documentation for new developers joining the agent installer team.
+
+In this phase it should also become more clear what scripts and code are overlapping
+between the two teams.
```
andfasano, line 135: nit: more than "between the two teams" I'd suggest to change it to "with the existing code base" (ie right now more than two teams are using dev-scripts for their dev / test environments)

enhancements/agent-installer/development-environment.md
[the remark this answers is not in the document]
lranjbar: I like this idea of defined minimum checks. That would help with onboarding new developers and learning to develop against OCP. This could be something we define for the future... We should at least put it in the backlog.

My desire to have the configurations local is more about running the configuration I'm currently working on while I'm developing. I want to have that ability to do that on demand and easily. For instance if I'm working on a feature that is specific to SNO clusters, I would only run SNO based flows locally while developing. When I push up my work the CI will run compact cluster and HA clusters checks against my code to make sure I didn't break anything. After that is finished I might work on a feature that is focused on a different type of cluster configuration than SNO so I wouldn't run SNO based flows. I realize this is my personal preference in developing but I think we can accommodate both styles without much issue.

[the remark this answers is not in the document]
lranjbar: I also just realized this thread is more about validations of the OCP clusters than configurations.

Local validations I'd like to see:
- Kubeconfig is created and accessible
- Control Plane came up
- Cluster is installed
- A generic way to capture the error output when cluster fails to install

enhancements/agent-installer/development-environment.md
lranjbar: Yeah that's a good point with disconnected vs connected the changes between the two are pretty advanced configurations 🤔

enhancements/testing/ephemeral-agent-installer-e2e-tests.md
lranjbar: I took reporting out of the scope of this enchancement. In the future linking into the existing reports should be looked at.

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Design Details
+
+### General Strategy
+
+* Using the scope of the agent installer project we will identify possible refactors to 
+existing scripts and new functionality required by this new project.
+* New functionality we will write in Ansible, unless it is highly coupled with existing
+shell scripts.
+* We will create Anisble roles to wrap the existing code making it easier to use within
```
zaneb, line 120: typo: Ansible

enhancements/agent-installer/development-environment.md
```diff
@@ -0,0 +1,221 @@
...
+a pure Ansible solution
+* The existing bash scripts will be left as is. Refactoring them will be done piece by 
+piece as needed.
+
+### Phase 1 (Crawl): Add development scripts into dev-scripts for Agent Installer
+
+1) Add development scripts into dev-scripts for the new agent installer project using
```
rwsu, line 127: Can you clarify what are the development scripts? Perhaps explain their purpose.

rwsu (commented):
Requesting minor clarification, otherwise lgtm.

zaneb (commented):
Last call for reviews. I'm planning to approve this tomorrow if there a no more major changes to incorporate.

enhancements/agent-installer/development-environment.md
lranjbar: I moved this information into the design details section under phase 1 in the update.

enhancements/agent-installer/development-environment.md
lranjbar: I added the three scenarios in an update. This is not meant to be an exhaustive list just to get us started.

dhellmann:
I'm happy with this version. I'll leave it open for other reviewers to comment before we approve it.

PR #1035, opened 2022-02-16
Pods Egress DSCP QoS proposal

oribon:
Signed-off-by: Ori Braunshtein <obraunsh@redhat.com>

### DIFF
--- a/enhancements/network/egress-qos.md
+++ b/enhancements/network/egress-qos.md
@@ -1,0 +1,216 @@
+---
+title: egress-qos
+authors:
+  - "@oribon"
+reviewers:
+  - "@trozet"
+  - "@danwinship"
+  - "@tssurya"
+  - "@abhat"
+approvers:
+  - "@danwinship"
+  - "@trozet"
+  - "@knobunc"
+api-approvers:
+  - "@trozet"
+creation-date: 2022-02-16
+last-updated: 2022-04-14
+status: implementable
+tracking-link:
+  - https://issues.redhat.com/browse/SDN-2097
+---
+
+# OVN Pods Egress DSCP QoS
+
+## Summary
+
+Not all traffic has the same priority, and when there is contention for bandwidth, there should be a mechanism for objects outside the cluster to prioritize the traffic.
+To enable this, we will use Differentiated Services Code Point (DSCP) which allows us to classify packets by setting a 6-bit field in the IP header, effectively marking the priority of a given packet relative to other packets as "Critical", "High Priority", "Best Effort" and so on.
+
+By introducing a new CRD `EgressQoS`, users could specify a DSCP value for packets originating from pods on a given namespace heading to a specified CIDR.
+The CRD will be Namespaced, with one resource allowed per namespace.
+The resources will be watched by ovn-k, which in turn will configure OVN's [QoS Table](https://man7.org/linux/man-pages/man5/ovn-nb.5.html#QoS_TABLE). 
+
+## Motivation
+
+Telco customers require support for DSCP marking capability for some of their 5G applications, giving some pods precedence over others.
+The QoS markings will be consumed and acted upon by objects outside of the OpenShift cluster to optimize traffic flow throughout their networks.
+
+### Goals
+
+- Provide a mechanism for users to set DSCP on egress traffic coming from specific namespaces.
+
+### Non-Goals
+
+- Ingress QoS.
+
+- Consolidating with current `kubernetes.io/egress-bandwidth` and `kubernetes.io/ingress-bandwidth` annotations.
+Nonetheless, the work done here does not interfere with the current bandwidth QoS mechanism.
+
+- The DSCP marking does not need to be handled or acted upon by OpenShift, just added to selected headers.
+
+- Marking East/West traffic, exposing the DSCP value from the inner packet to the outer geneve packet.
+
+## Proposal
+
+To achieve egress DSCP marking on pods, we introduce a new namespace-scoped CRD `EgressQoS` which allows specifying a set of QoS rules, each has a DSCP value, a destination CIDR and a PodSelector.
+Traffic coming from pods on the namespace heading to each destination CIDR that match the selector will be marked with the corresponding DSCP value.
+Not specifying a PodSelector will apply a rule to all pods in the namespace.
+
+### Implementation Details/Notes/Constraints
+
+A new API `EgressQoS` under the `k8s.ovn.org/v1` version will be added to `pkg/crd`.
+
+A new controller in OVN-K will watch `EgressQoS`, `Pod` and `Node` objects, which will create the relevant QoS objects and attach them to all of the node local switches in the cluster in OVN - resulting in the necessary flows to be programmed in OVS.
+
+In order to not create an OVN QoS object per pod in the namespace, the controller will also manage AddressSets. 
+For each QoS rule specified in a given `EgressQoS` it'll create an AddressSet, adding only the pods
+whose label matches the PodSelector to it, making sure that new/updated/deleted matching pods are also added/updated/deleted accordingly.
+Rules that do not have a PodSelector will leverage the namespace's AddressSet.
+
+For example, using LGW mode and assuming there's a single node `node1` and the following `EgressQoS` is created:
+
+```yaml
+kind: EgressQoS
+apiVersion: k8s.ovn.org/v1
+metadata:
+  name: default
+  namespace: default
+spec:
+  egress:
+  - dscp: 46
+    dstCIDR: 1.2.3.4/32
+    podSelector:
+      matchLabels:
+        priority: Critical
+  - dscp: 30
+    dstCIDR: 0.0.0.0/0
+```
+the equivalent of:
+```bash
+ovn-nbctl qos-add node1 to-lport 1000 "ip4.src == <default_ns-egress-qos-1000 address set> && ip4.dst == 1.2.3.4/32" dscp=46
+ovn-nbctl qos-add node1 to-lport 999 "ip4.src == <default_ns address set> && ip4.dst == 0.0.0.0/0" dscp=30
+```
+will be executed.
+Creating a new Pod that matches the first rule's podSelector in the namespace results in its IPs being added to that rule's Address Set.
+
+In addition it'll watch nodes to decide if further updates are needed, for example:
+when another node `node2` joins the cluster, the controller will attach the existing
+`QoS` object to its node local switch.
+
+IPv6 will also be supported, given the following `EgressQoS`:
+```yaml
+apiVersion: k8s.ovn.org/v1
+kind: EgressQoS
+metadata:
+  name: default
+  namespace: default
+spec:
+  egress:
+  - dscp: 48
+    dstCIDR: 2001:0db8:85a3:0000:0000:8a2e:0370:7330/124
+```
+and a single pod with the IP `fd00:10:244:2::3` in the namespace, the controller will create the relevant QoS object that will result in a similar flow to this on the pod's node:
+```bash
+ cookie=0x6d99cb18, duration=63.310s, table=18, n_packets=0, n_bytes=0, idle_age=63, priority=555,ipv6,metadata=0x4,ipv6_src=fd00:10:244:2::3,ipv6_dst=2001:db8:85a3::8a2e:370:7330/124 actions=mod_nw_tos:192,resubmit(,19)
+```
+
+### User Stories
+#### Story 1
+
+As a user of OpenShift, I should be able to mark egress traffic coming from a specific namespace with a valid DSCP value.
+
+### API Extensions
+
+A new namespace-scoped CRD is introduced:
+
+```go
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:path=egressqoses
+// +kubebuilder::singular=egressqos
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// EgressQoS is a CRD that allows the user to define a DSCP value
+// for pods egress traffic on its namespace to specified CIDRs.
+// Traffic from these pods will be checked against each EgressQoSRule in
+// the namespace's EgressQoS, and if there is a match the traffic is marked
+// with the relevant DSCP value.
+type EgressQoS struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   EgressQoSSpec   `json:"spec,omitempty"`
+	Status EgressQoSStatus `json:"status,omitempty"`
+}
+
+// EgressQoSSpec defines the desired state of EgressQoS
+type EgressQoSSpec struct {
+	// a collection of Egress QoS rule objects
+	Egress []EgressQoSRule `json:"egress"`
+}
+
+type EgressQoSRule struct {
+	// DSCP marking value for matching pods' traffic.
+	// +kubebuilder:validation:Maximum:=63
+	// +kubebuilder:validation:Minimum:=0
+	DSCP int `json:"dscp"`
+
+	// DstCIDR specifies the destination's CIDR. Only traffic heading
+	// to this CIDR will be marked with the DSCP value.
+	// This field is optional, and in case it is not set the rule is applied
+	// to all egress traffic regardless of the destination.
+	// +optional
+	DstCIDR *string `json:"dstCIDR,omitempty"`
+
+	// PodSelector applies the QoS rule only to the pods in the namespace whose label
+	// matches this definition. This field is optional, and in case it is not set
+	// results in the rule being applied to all pods in the namespace.
+	// +optional
+	PodSelector metav1.LabelSelector `json:"podSelector,omitempty"`
+}
+```
+
+### Test Plan
+
+* Unit tests coverage
+
+* IPv4/IPv6 E2E that validates egress traffic from a namespace is marked with the correct DSCP value by creating and deleting `EgressQoS`, setting up src pods and host-networked destination pods.
+  * Traffic to the specified CIDR should be marked.
+  * Traffic to an address not contained in the CIDR should not be marked.
+
+### Risks and Mitigations
+N/A
+## Design Details
+N/A
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+Dev Preview: 4.11
+Tech Preview: 4.12
+#### Tech Preview -> GA
+GA: 4.13
+#### Removing a deprecated feature
+N/A
+
+### Upgrade / Downgrade Strategy
+N/A
+### Version Skew Strategy
+N/A
+
+### Operational Aspects of API Extensions
+N/A
+
+#### Failure Modes
+N/A
+
+#### Support Procedures
+N/A
+
+## Implementation History
+N/A
+
+## Drawbacks
+N/A
+## Alternatives
+N/A


enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,107 @@
...
+
+# OVN Pods Egress DSCP QoS
+
+## Summary
+
+Not all traffic has the same priority, and when there is contention for bandwidth, there should be a mechanism for objects outside the cluster to prioritize the traffic.
+To enable this, we will use Differentiated Services Code Point (DSCP) which allows us to classify packets by setting a 6-bit field in the IP header, effectively marking the priority of a given packet relative to other packets as "Critical", "High Priority", "Best Effort" and so on.
+
+By introducing a new CRD `PodQos`, users could specify a DSCP value for packets originating from pods on a set of namespaces heading to a specified CIDR.
+The CRs will be watched by ovn-k, which in turn will configure OVN's [QoS Table](https://man7.org/linux/man-pages/man5/ovn-nb.5.html#QoS_TABLE).
```
danwinship, line 22: "The resources" or "The objects". Don't say "The CRs" when it's not actually relevant that they're Custom. (People do that a lot, but they shouldn't.)

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,107 @@
...
+- Consolidating with current `kubernetes.io/egress-bandwidth` and `kubernetes.io/ingress-bandwidth` annotations.
+Nonetheless, the work done here does not interfere with the current bandwidth QoS mechanism. # Q: Is this point relevant?
+
+- The DSCP marking does not need to be handled or acted upon by OpenShift, just added to selected headers.
+
+- Marking East/West traffic, exposing the DSCP value from the inner packet to the outer geneve packet. # Q: is this a goal? If so does it mean we allow dstCIDR to be empty / expect the user to specify an internal pod ip? https://www.mail-archive.com/ovs-dev@openvswitch.org/msg59787.html
+
+## Proposal
+
+To achieve egress DSCP marking on pods, we introduce a new CRD `PodQos`, which lets users specify a namespaceSelector, DSCP value and destination CIDR - meaning traffic coming from pods on specific namespaces heading to the destination CIDR will be marked with the given DSCP value.
```
danwinship, line 46: Why per-namespace? Why not per-pod? Why allow a single PodQoS object to apply to multiple namespaces, rather than requiring a separate PodQoS for each namespace? Why per-destination-CIDR rather than per-destination-port or based on time of day or anything else?

Even if the current decisions solve the problem for the current customer, will they solve the problem for all customers?

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,107 @@
...
+
+```yaml
+kind: PodQos
+apiVersion: k8s.ovn.org/v1
+metadata:
+  name: ns1-46
+  namespace: default # Q: Should this be cluster-scoped?
```
danwinship, line 53: As currently defined, yes, because this object has no connection to the `default` namespace

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,107 @@
...
+    namespaceSelector:
+      matchLabels:
+        kubernetes.io/metadata.name: ns1
+    dscp: 46
+    dstCIDR: 0.0.0.0/0 # Q: Do we want this CIDR / slice of CIDRs/IPs / single IP / ... ?
+```
+
+OVN-K watches these resources, creating/updating/deleting QoS objects in nbctl accordingly.
+For example, assuming there's a single pod `app` in namespace `ns1` on node `node1` and the above `PodQos` created the equivalent of:
+`ovn-nbctl qos-add <node1> from-lport <priority> "inport == \"ns1_app\" && ip4.dst == 0.0.0.0/0" dscp=46` # Q: Will we have a fixed priority?
```
danwinship, line 67: You should not need to match individual ports; you can match the address set for the namespace

enhancements/network/egress-qos.md
oribon: I went with per-namespace and not per-pod because that's the current requirements (which tbh are pretty vague right now), while per-pod(s) could be added later on.
Rethinking with your comments, we could have the CRD namespace scoped, with subsections corresponding to a CIDR/+port, leaving room for further granularity like podSelectors in the future, while enforcing 1 PodQos per namespace.
That way it might be less ambiguous in comparison to a cluster scoped CRD - it'll be clearer from where each Namespace gets its QoS.
Something like:
```yaml
kind: PodQos
apiVersion: k8s.ovn.org/v1
metadata:
  name: default
  namespace: ns1
spec:
  egress:
    - dstCIDR: 0.0.0.0/0
      port: 80
      dscp: 46
      priority: 1 # needed if overlapping destinations are used?
      # PodSelector could be another field here in the future?
    - dstCIDR: 1.2.3.4/32
      port: 443
      dscp: 32
      priority: 2
...
```
Is this approach better/cleaner?

IMO using dstCIDR/+port (or none) to determine when to apply DSCP makes sense, and tbh I'm not too familiar with other possible requirements to tell if there's anything better as a starting point or if it contradicts with another option.

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,107 @@
...
+---
+
+# OVN Pods Egress DSCP QoS
+
+## Summary
+
+Not all traffic has the same priority, and when there is contention for bandwidth, there should be a mechanism for objects outside the cluster to prioritize the traffic.
+To enable this, we will use Differentiated Services Code Point (DSCP) which allows us to classify packets by setting a 6-bit field in the IP header, effectively marking the priority of a given packet relative to other packets as "Critical", "High Priority", "Best Effort" and so on.
+
+By introducing a new CRD `PodQos`, users could specify a DSCP value for packets originating from pods on a set of namespaces heading to a specified CIDR.
```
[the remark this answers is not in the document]
danwinship: If I was going to nitpick the name, it would be the "Pod" part more than the "QoS" part... (Isn't it more like "NamespaceQos"? or "NamespaceEgressQos"?)

Maybe as we nail down the use case (and possible future changes to it) some more a better name will become clear.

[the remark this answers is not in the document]
tssurya: I like "EgressQoS", just like we have egressIPs, egressFirewalls and maybe in the future egressGateways, fits into the picture since afterall it is QoS for egress traffic, restricting to Pod or Namespace at this point might hurt us in the future if we think of changing its scope?

tssurya (commented):
/cc

enhancements/network/egress-qos.md
oribon: > Like, if this object is _only_ going to be used to describe QoS-related stuff that is _only_ acted on by out-of-cluster components, then that suggests one design. But if we wanted to handle QoS-related stuff more generally, then that might suggest a different design, etc.

Well, I think our main (or maybe only) focus is the former. Going by that I updated the file to reflect that, pretty much following the `EgressFirewall` approach - having a namespace-scoped CRD `EgressQos`, affecting all of the pods in the resource's namespace. I think it makes sense in this context, and keeps this egress only - leaving room for maybe small additions like marking by destination port if we'd like.

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+---
+
+# OVN Pods Egress DSCP QoS
+
+## Summary
+
+Not all traffic has the same priority, and when there is contention for bandwidth, there should be a mechanism for objects outside the cluster to prioritize the traffic.
+To enable this, we will use Differentiated Services Code Point (DSCP) which allows us to classify packets by setting a 6-bit field in the IP header, effectively marking the priority of a given packet relative to other packets as "Critical", "High Priority", "Best Effort" and so on.
+
+By introducing a new CRD `EgressQos`, users could specify a DSCP value for packets originating from pods on a given namespace heading to a specified CIDR.
```
tssurya, line 23: nit:
```suggestion
By introducing a new CRD `EgressQoS`, users could specify a DSCP value for packets originating from pods on a given namespace heading to a specified CIDR.
```

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+- Consolidating with current `kubernetes.io/egress-bandwidth` and `kubernetes.io/ingress-bandwidth` annotations.
+Nonetheless, the work done here does not interfere with the current bandwidth QoS mechanism.
+
+- The DSCP marking does not need to be handled or acted upon by OpenShift, just added to selected headers.
+
+- Marking East/West traffic, exposing the DSCP value from the inner packet to the outer geneve packet.
+
+## Proposal
+
+To achieve egress DSCP marking on pods, we introduce a new CRD `EgressQos`, which allows specifying a set of DSCP values and destination CIDRs - meaning traffic coming from pods on the namespace heading to each destination CIDR will be marked with the given DSCP value.
```
tssurya, line 49:
```suggestion
To achieve egress DSCP marking on pods, we introduce a new CRD `EgressQoS`, which allows specifying a set of DSCP values and destination CIDRs - meaning traffic coming from pods on the namespace heading to each destination CIDR will be marked with the given DSCP value.
```

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+
+```yaml
```
tssurya, line 51: Keep the proposal brief and move this yaml into impementation details section maybe?

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+spec:
+  egress:
+  - dscp: 46
+    dstCIDR: 0.0.0.0/0
+    priority: "?" # Do we add a priority field to handle overlapping CIDRs? Should we handle that in the code instead?
+  - dscp: 30
+    dstCIDR: 1.2.3.4/32
+```
+
+OVN-K watches these resources, creating/updating/deleting QoS objects in nbctl accordingly.
```
tssurya, line 66: in OVN accordingly so that the necessary flows can be programmed?
nbctl is just a utility.

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+For example, assuming there's a single pod `app` in namespace `default` on node `node1` and the above `EgressQos` created the equivalent of:
+```bash
+ovn-nbctl qos-add node1 from-lport 1 "ip4.src == <default_ns_address_set> && ip4.dst == 0.0.0.0/0" dscp=46
```
tssurya, line 69: again move the details in implementation section?

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+### User Stories
+
+### API Extensions
+
+A new namespace-scoped CRD is introduced.
+### Implementation Details/Notes/Constraints
+
+- Adding a new API `EgressQos` under the `k8s.ovn.org/v1` version to `pkg/crd`.
+
+- Adding a controller that watches `EgressQoses`, `Pods`, producing the relevant QoS objects in nbdb and attaching them to the join switch or to the relavant logical switches in LGW.
```
tssurya, line 85: I know we spoke about this offline, but we need to decide if its going to be on node switch or join switch depending on the gateway or if there isn't much perf impact for both modes we can probably leave it on the node switch.

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+kind: EgressQos
+apiVersion: k8s.ovn.org/v1
+metadata:
+  name: default
+  namespace: default
+spec:
+  egress:
+  - dscp: 46
+    dstCIDR: 0.0.0.0/0
+    priority: "?" # Do we add a priority field to handle overlapping CIDRs? Should we handle that in the code instead?
```
tssurya, line 61: I wonder how we do it for firewall code today?

enhancements/network/egress-qos.md
tssurya, line 85:
```suggestion
- Adding a controller in OVN-K that watches `EgressQoS` and `Pod` objects, which will create the relevant QoS objects in OVN and flows in OVS. These rules will be added on node local switches in the relevant nodes or something???
```

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Risks and Mitigations
+
+## Design Details
+
+### Test Plan
+
+* Unit tests coverage
+
+* IPv4/IPv6 E2E that validates egress traffic from a namespace is marked with the correct DSCP value by creating and deleting `EgressQoses`, setting up src pods and host-networked destination pods.
```
tssurya, line 95:
```suggestion
* IPv4/IPv6 E2E that validates egress traffic from a namespace is marked with the correct DSCP value by creating and deleting `EgressQoS`, setting up src pods and host-networked destination pods.
```

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+For example, assuming there's a single pod `app` in namespace `default` on node `node1` and the above `EgressQos` created the equivalent of:
+```bash
+ovn-nbctl qos-add node1 from-lport 1 "ip4.src == <default_ns_address_set> && ip4.dst == 0.0.0.0/0" dscp=46
+ovn-nbctl qos-add node1 from-lport 2 "ip4.src == <default_ns_address_set> && ip4.dst == 1.2.3.4/32" dscp=30
+```
+will be executed.
+
+In addition it'll watch pods to decide if further updates are needed, for example attaching the QoS to another switch.
+
+### User Stories
```
tssurya, line 76: We should fill this with the telco use cases?

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Test Plan
+
+* Unit tests coverage
+
+* IPv4/IPv6 E2E that validates egress traffic from a namespace is marked with the correct DSCP value by creating and deleting `EgressQoses`, setting up src pods and host-networked destination pods.
+  * Traffic to the specified CIDR should be marked.
+  * Traffic to an address not contained in the CIDR should not be marked.
+
+### Graduation Criteria
```
tssurya, line 99: Mark the irrelevant sections with N/A

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+
+#### Dev Preview -> Tech Preview
+
```
tssurya, line 102: This is Dev Preview in 4.11 right?
Tech Preview in 4.12 and GA in 4.13? Is that the goal?

tssurya (changes requested):

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
+---
+title: OVN Pods Egress DSCP QoS
+authors:
+  - "@oribon"
+reviewers:
+  - "@trozet"
+  - "@danwinship"
+  - "@tssurya"
+approvers:
+  - TBD
```
tssurya, line 10: suggestion: you can make Tim approver for OVN-K side, Dan for API side, Ben Bennett who leads networking.

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+- Provide a mechanism for users to set DSCP on egress traffic coming from specific namespaces.
+
+### Non-Goals
+
+- Ingress QoS.
+
+- Consolidating with current `kubernetes.io/egress-bandwidth` and `kubernetes.io/ingress-bandwidth` annotations.
+Nonetheless, the work done here does not interfere with the current bandwidth QoS mechanism.
+
+- The DSCP marking does not need to be handled or acted upon by OpenShift, just added to selected headers.
```
mccv1r0, line 43: So the external world hurries up to get a packet to OCP that we treat as best effort?

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+will be executed.
+
+In addition it'll watch pods to decide if further updates are needed, for example attaching the QoS to another switch.
+
+### User Stories
+
+### API Extensions
+
+A new namespace-scoped CRD is introduced.
+### Implementation Details/Notes/Constraints
```
dcbw, line 81: We need to address IPv6 too. When you add examples, could you add IPv6 ones too?

enhancements/network/egress-qos.md
msherif1234: sorry trying to understand this feature, ovnk will mark and the some entity outside the cluster will check DSCP and take some actions so we are talking about service traffic right so CIDR here are ClusterIP or LBServiceIP right ?

enhancements/network/egress-qos.md
msherif1234: how can u refelect the priority setting in the packet header ? setting dscp should be enough IMO

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,119 @@
...
+### Non-Goals
+
+- Ingress QoS.
+
+- Consolidating with current `kubernetes.io/egress-bandwidth` and `kubernetes.io/ingress-bandwidth` annotations.
+Nonetheless, the work done here does not interfere with the current bandwidth QoS mechanism.
+
+- The DSCP marking does not need to be handled or acted upon by OpenShift, just added to selected headers.
+
+- Marking East/West traffic, exposing the DSCP value from the inner packet to the outer geneve packet.
```
msherif1234, line 45: Hmm who inside the cluster will be looking at the marked header ? unless we have ingress ACL that filter on dscp and take actions but this seems to be irrelevant to this proposal ?

enhancements/network/egress-qos.md
msherif1234: we should also be clear what we will be using for from-lport I assum QoS has the same limitation like ACL and it only applies to switch's lp so which switch port will be used here nodeSwitch or JoinSwitch I assume its the nodeSwitch but that needs to be called out in the implementation detail section.

enhancements/network/egress-qos.md
msherif1234: oh @tssurya I was thinking about the same exact thing before I read your comment, since it seems we will move the egress ACL to nodeSwitch then it seems like having QoS there also will make sense ?

enhancements/network/egress-qos.md
oribon: Yeah some entity outside of the cluster will take actions based on the dscp value, but the CIDRs here are these entities - destinations that are external.
The purpose is to mark egress traffic from specific pods based on the relevant rule in the namespace's EgressQoS: by dstCIDR and dscp value.

enhancements/network/egress-qos.md
oribon: done

enhancements/network/egress-qos.md
oribon: done

enhancements/network/egress-qos.md
oribon: whoops, fixed

enhancements/network/egress-qos.md
oribon: > I wonder how we do it for firewall code today?

iiuc it relies on the rules to be in a descending order: https://github.com/ovn-org/ovn-kubernetes/blob/a24a233130fb87e07557d8c1579ad52a861fde96/go-controller/pkg/ovn/egressfirewall.go#L341
We can do that here and/or maybe sorting the rules first by mask/prefix length?


> how can u refelect the priority setting in the packet header ? setting dscp should be enough IMO

this priority relates to the one in the QoS table: https://man7.org/linux/man-pages/man5/ovn-nb.5.html#QoS_TABLE , iiuc we should differentiate similar matches using this.

enhancements/network/egress-qos.md
oribon: Not sure I understand, but this PR concerns only marking egress traffic from specified pods. We won't do anything from Openshift's side for incoming traffic regarding DSCP related/unrelated to what's added here.

enhancements/network/egress-qos.md
oribon: This is a non-goal, thought it's worth writing it here that we don't care about any traffic other than traffic that exits the cluster. Do you think I should remove this point?

enhancements/network/egress-qos.md
oribon: sure added a short one for now

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,187 @@
...
+---
+
+# OVN Pods Egress DSCP QoS
+
+## Summary
+
+Not all traffic has the same priority, and when there is contention for bandwidth, there should be a mechanism for objects outside the cluster to prioritize the traffic.
+To enable this, we will use Differentiated Services Code Point (DSCP) which allows us to classify packets by setting a 6-bit field in the IP header, effectively marking the priority of a given packet relative to other packets as "Critical", "High Priority", "Best Effort" and so on.
+
+By introducing a new CRD `EgressQoS`, users could specify a DSCP value for packets originating from pods on a given namespace heading to a specified CIDR.
```
abhat, line 23: Should the API design also allow for ingress specification, or would that be a separate CRD?

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,187 @@
...
+The CRD will be Namespaced, with one resource allowed per namespace.
```
abhat, line 24: The API should probably allow for namespace selector or pod selectors to allow for a more granular control or do we want this to be purely a per-namespace thing?

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,187 @@
...
+type EgressQoSSpec struct {
+	// a collection of Egress QoS rule objects
+	Egress []EgressQoSRule `json:"egress"`
+}
+
+type EgressQoSRule struct {
+	// Dscp marking value for matching pods' traffic.
+	// +kubebuilder:validation:Maximum:=63
+	// +kubebuilder:validation:Minimum:=0
+	Dscp int `json:"dscp"`
```
abhat, line 139: s/Dscp/DSCP?

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,187 @@
...
+
+	// DstCIDR specifies the destination's CIDR. Only traffic heading to this CIDR will be marked with DSCP.
+	DstCIDR string `json:"dstCIDR"`
```
abhat, line 142: Should this be optional? In other words, if the user wants all egress traffic marked with the same QoS value, there should be no need to specify 0.0.0.0/0

abhat: Need kubebuilder validation for addresses and valid dscp code values.

enhancements/network/egress-qos.md
oribon: In the current design Ingress would have to be in a separate CRD https://github.com/openshift/enhancements/pull/1035#discussion_r812840859

enhancements/network/egress-qos.md
oribon: I think it would be clearer having the CRD namespace-scoped and maybe a bit easier to implement that way. We'll achieve per-namespace egress dscp both ways, but won't have to manage/watch namespaces.
I believe pod selectors is a field we could add later as part of the egress rule if we'd want.

enhancements/network/egress-qos.md
oribon: done

enhancements/network/egress-qos.md
oribon: > Should this be optional? In other words, if the user wants all egress traffic marked with the same QoS value, there should be no need to specify 0.0.0.0/0

I don't have a strong opinion about this, do you think defaulting to 0.0.0.0/0 or ::/0 is clearer?



> Need kubebuilder validation for addresses and valid dscp code values.

For the dscp I added one, for addresses I'm a bit hesitant about writing a good enough regex that covers all IPv4/6 CIDRs. In any case, we'll parse the addresses with `net.ParseCIDR`.

abhat: Check how the service CIDR and cluster IP subnets are validated in network.config.openshift.io object in `openshift/api`.

abhat: If the 80% users want all traffic in the namespace to be marked a certain way, then the dst cidr field seems like an overhead. OTOH, if specifying only a subset of destinations for which the packets should be marked with a certain DSCP value is the more common case, then I would lean on keeping the field mandatory.

enhancements/network/egress-qos.md
trozet: If we only care about egress traffic, then we should be implementing this at the join switch. Otherwise we are going to marking east west packets for no reason.

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,187 @@
...
+```yaml
+kind: EgressQoS
+apiVersion: k8s.ovn.org/v1
+metadata:
+  name: default
+  namespace: default
+spec:
+  egress:
+  - dscp: 46
+    dstCIDR: 0.0.0.0/0
```
trozet, line 68: what does 0.0.0.0/0 represent in OVN? Is that the equivalent of "any"? If so what happens with overlapping CIDRs?

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,187 @@
...
+  - dscp: 30
+    dstCIDR: 1.2.3.4/32
```
trozet, line 70: why do we only match on destination CIDR and then assume the source is all addresses in this namespace? I think it would be better to add srcCIDR option as well so that only certain pods may be selected.

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,187 @@
...
+### Implementation Details/Notes/Constraints
+
+A new API `EgressQoS` under the `k8s.ovn.org/v1` version will be added to `pkg/crd`.
+
+A new controller in OVN-K will watch `EgressQoS` and `Pod` objects, which will create the relevant QoS objects in OVN and result in the necessary flows to be programmed in OVS. By listing the pods in the namespace these QoS rules will be attached only to the relevant node local switches.
+
+For example, assuming there's a single pod `app1` in namespace `default` on node `node1` and the following `EgressQoS` is created:
+
+```yaml
+kind: EgressQoS
```
trozet, line 60: I presume in the future that:
"Consolidating with current `kubernetes.io/egress-bandwidth` and `kubernetes.io/ingress-bandwidth` annotations." will end up being consolidated into a QOS CRD. In that case I suspect the relevant field, like egress-bandwidth would be added under this CRD type. Do we want to end up having 2 separate EgressQoS and IngressQoS CRDs, or should we just have one QoS CRD?

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,187 @@
...
+  egress:
+  - dscp: 46
+    dstCIDR: 0.0.0.0/0
+  - dscp: 30
+    dstCIDR: 1.2.3.4/32
+```
+the equivalent of:
+```bash
+ovn-nbctl qos-add node1 from-lport 1 "ip4.src == <default_ns_address_set> && ip4.dst == 0.0.0.0/0" dscp=46
+ovn-nbctl qos-add node1 from-lport 2 "ip4.src == <default_ns_address_set> && ip4.dst == 1.2.3.4/32" dscp=30
```
trozet, line 75: note this "2" is priority:
[--may-exist] qos-add switch direction priority match [dscp=dscp]

how will the CRD determine the priority to use for these?

trozet:
@oribon can you resolve some of the resolved comments so it cleans up this review a bit?

enhancements/network/egress-qos.md
oribon: I thought following what egressFirewall is doing (https://github.com/ovn-org/ovn-kubernetes/blob/a24a233130fb87e07557d8c1579ad52a861fde96/go-controller/pkg/ovn/egressfirewall.go#L341) - relying on the rules of the CRD to be in a descending order priority-wise. The user would have to put them in an order that'll work like having a specific CIDR in a higher priority than a broader one that includes it.

enhancements/network/egress-qos.md
oribon: yes that would be "any". we're still not sure if we should allow the field to be empty and that'll default it to 0.0.0.0/0 or ::/0
regarding overlapping CIDRs that'll need to be handled by the priority setting: https://github.com/openshift/enhancements/pull/1035#discussion_r825411903

enhancements/network/egress-qos.md
oribon: Wouldn't srcCIDR be a bit complicated for the user? Adding a PodSelector per rule is possible with the current design but I left it out for now as it seems that it is not a requirement.

enhancements/network/egress-qos.md
oribon: I think that we want to keep this CRD focused on egress QoS, so if we do decide to have some ingress QoS stuff it'll have to be under a different CRD.
https://github.com/openshift/enhancements/pull/1035#discussion_r811211259

oribon:
> @oribon can you resolve some of the resolved comments so it cleans up this review a bit?

done

enhancements/network/egress-qos.md
oribon: done, added some words about LGW vs SGW

enhancements/network/egress-qos.md
abhat: that API is very confusing to me. do we want to explore an explicit priority field?

enhancements/network/egress-qos.md
abhat: Actually revisiting my conversation with the customer, pod selector is very much a requirement. We want to be able to say `Pod with X label gets DSCP marking of Y for all of its traffic` as a functional MVP. namespace selector can come later perhaps and if we are making the CRD namespace scoped, then we will be OK.

oribon: Added a PodSelector field for the QoS rules, and a bit about how we can use AddressSets to make sure matching pods are marked with the right DSCP.

enhancements/network/egress-qos.md
oribon: we could, but I'm not sure how useful it could be for the user.
with the rules for some namespace being something like :
```
  - dscp: 46
    dstCIDR: 1.2.3.4/32
  - dscp: 40
    dstCIDR: 1.2.3.0/24
  - dscp: 30
    dstCIDR: 0.0.0.0/0
```
would it add any functionality to let the user set some priorities here? would it make any difference if he sets them 3,2,1 or 80,75,5?
looking at a larger set of rules, is it simpler to observe the priority of each rule to understand what takes precedence or rather relying on the order (or some other option)?

trozet: I think we should just make it work like egress firewall

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,209 @@
...
+metadata:
+  name: default
+  namespace: default
+spec:
+  egress:
+  - dscp: 46
+    dstCIDR: 1.2.3.4/32
+    podSelector:
+      matchLabels:
+        priority: Critical
```
abhat, line 83: What does the `priority: Critical` mean in the context of the `podSelector`?

abhat: Oh it is just an example of matching pods with labels as critical?

oribon: yes :sweat_smile:

enhancements/network/egress-qos.md
oribon: after also talking with Tim and Dan this field should be optional, updated.
thanks :)

trozet (approved):
/approve

abhat:
/hold
Need additional API review.

enhancements/network/egress-qos.md
```diff
@@ -0,0 +1,214 @@
...
+To achieve egress DSCP marking on pods, we introduce a new namespace-scoped CRD `EgressQoS` which allows specifying a set of QoS rules, each has a DSCP value, a destination CIDR and a PodSelector.
+Traffic coming from pods on the namespace heading to each destination CIDR that match the selector will be marked with the corresponding DSCP value.
+Not specifying a PodSelector will apply a rule to all pods in the namespace.
+
+### Implementation Details/Notes/Constraints
+
+A new API `EgressQoS` under the `k8s.ovn.org/v1` version will be added to `pkg/crd`.
+
+A new controller in OVN-K will watch `EgressQoS`, `Pod` and `Node` objects, which will create the relevant QoS objects in OVN and result in the necessary flows to be programmed in OVS. To which logical switches these QoS objects are attached to differs a bit between SGW/LGW modes:
+* In SGW mode it is enough to attach the QoS objects only to the single distributed `join` switch, as all pods egress traffic goes through it.
```
trozet, line 61: do we need to put it on 2 different switches (depending on gateway mode)? It would be easier to just do this on the node switch for both modes I think. Also, is the qos evaluation done after load balancer DNAT so that your dst match would be post service DNAT?

trozet: I know I previously made a comment that steered you towards splitting to worker and join switch for different modes, but I think since we match on dest ip its OK to do them both on the worker switch.

@dceara pointed out that the qos eval is done pre-LB DNAT, so matching on destinations that are k8s endpoints woudln't work. Maybe its worth just noting that here?

dceara: Is there any reason why the traffic can't be dscp-marked in the "egress" pipeline (egress from OVN perspective)?  If we use "to-lport" QOS rules then there's no issue, those are applied after LB DNAT, so they could match on endpoint IPs.

oribon: > I know I previously made a comment that steered you towards splitting to worker and join switch for different modes, but I think since we match on dest ip its OK to do them both on the worker switch.
> 
> @dceara pointed out that the qos eval is done pre-LB DNAT, so matching on destinations that are k8s endpoints woudln't work. Maybe its worth just noting that here?

changed it to use node switches regardless of sgw/lgw.



> Is there any reason why the traffic can't be dscp-marked in the "egress" pipeline (egress from OVN perspective)? If we use "to-lport" QOS rules then there's no issue, those are applied after LB DNAT, so they could match on endpoint IPs.

no reason, changed it to to-lport as it seems that it still works

PR #1036, opened 2022-02-17
Insights Operator Prometheus Alerts for Insights Recommendations

natiiix:


enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Goals
+
+- New Prometheus metric with labels containing aforementioned properties of
+  the recommendation that will indicate if the recommendation is currently active.
+- New Prometheus alert based on this metric, defined using a template that fills
+  the information from labels into the description of the alert.
+- The user should be able to find out when was a certain recommendation active
+  by querying the Prometheus metric (and plot this in Grafana, for example).
+- Configuration option allowing the user to disable/enable this functionality.
```
tremes, line 57: Maybe I missed that but I think we should explicitly mention that it will be opt-in feature (disabled by default)

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,267 @@
...
+  endpoint moves elsewhere, the URL will be broken and an updated version of the
+  alert template will have to be backported. The process can take several
+  months, and there is no guarantee that the user will perform the update.
+
+## Design Details
+
+### Open Questions [optional]
+
+1. Which operator conditions should fall under the unhealthy Insights Operator
+   state?
```
tremes, line 176: Apart of the common clusteroperator conditions, we have these [special conditions](https://github.com/openshift/insights-operator/blob/master/pkg/controller/status/conditions.go#L8) so I think that whenever there's any condition related to a failure to obtain the new data (basically `UploadDegraded` and/or `InsightsDownloadDegraded`)

natiiix: The upload and download are certainly import to this functionality, and the `Disabled` status is definitely something we'd need to watch out for, but I'm not too sure about the `SCANotAvailabe`. Should that also be considered a fatal status, which should cause the alerts to be hidden?

tremes: No I don't think so. `SCANotAvailable` is basically a situation when we can't download the SCA certs for some reason.

bparees:
This looks reasonable and the impact is largely contained to the insights team, so i don't know that additional approval outside the insight team itself is needed, but interested parties would probably include:

@simonpasquier (metrics/monitoring)
@sdodson @wking (telemetry/fleet health observation)

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
bparees: @sdodson this might give you the most cause for concern:  if the insights operator can't reach the reporting service, it may report itself degraded, which would mean upgrades can't complete.  That sort of coupling to an external service could be fragile.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+- The alert should only fire if IO has been healthy at any point in the past 10
+  minutes. Healthy, in this context, means that there is no major error
+  happening (crashlooping, degraded operator status, etc.). A temporary error
+  state should not cause the alerts to be disabled.
+- If the user intentionally disables IO, the alerts should never fire. If IO is
+  disabled while some of the Insights alerts are firing, the alerts should
+  immediately stop firing (a part of their condition will be that IO is
+  enabled).
+- If IO was disabled by an accident, the user will be notified via the existing
+  alert.
```
simonpasquier, line 140: which alert is it?

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+
+- The latest successfully fetched and parsed report is too old
+  (`Meta.LastCheckedAt` is more than 8 hours ago)
+  - Reset the Insights recommendation metrics to avoid showing alerts about
+    recommendations that are no longer hitting.
+- Unable to fetch a report from Smart Proxy (endpoint unreachable, auth error, etc.)
+  - Log error, do not touch the Insights recommendation metrics.
+- The fetched report cannot be decoded into the expected structure, or the data
+  are otherwise invalid
+  - Log error, do not touch the Insights recommendation metrics.
```
simonpasquier, line 152: Not directly related to this proposal but do we already alert on these failure scenarios?

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+metrics will not be directly affected for as long as the last report is valid.
+If IO has been disabled, or crashing for more then 10 minutes, the alerts will
+be disabled to avoid presenting the user with outdated information.
+
+### Risks and Mitigations
+
+- Disabling and re-enabling an alert (e.g, due to IO or Smart Proxy outage)
+  can trigger user-defined actions, even though there has not been any actual
+  change to the status of the recommendation. This could be very annoying if
+  it were to happen often or repeatedly over a short period of time.
```
simonpasquier, line 166: The alerting expression should be written in a way that temporary outages don't affect the alert.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+- The URL to detailed information about an active Insights recommendation will
+  be hard-coded into the alert template or the Insights Operator. If the
+  endpoint moves elsewhere, the URL will be broken and an updated version of the
+  alert template will have to be backported. The process can take several
+  months, and there is no guarantee that the user will perform the update.
```
simonpasquier, line 171: So it means that if the endpoint location changes ever, we need to setup a redirection to the new location?

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+      recommendation (do not repeat alerts for previously spotted recommendations)
+    - If a recommendation was previously registered as active and the latest
+      report no longer contains it, the metric will either be reset or
+      unregistered from the Prometheus instance if possible. (remove alerts for
+      no longer hitting recommendations)
+    - If a recommendation was previously registered as inactive or the metric
+      for this recommendation is absent and the latest Insights report contains
+      it, a metric will be updated/created to indicate that the recommendation
+      is currently hitting. (indicate new recommendations)
+4. The metric will have labels containing the ID, severity (based on the total
```
simonpasquier, line 112: It would help to have a concrete example of the metric.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+   risk property), a human-readable name, and a link to a detailed description
+   of the recommendation and its remediation steps (Insights Advisor URL
+   generated based on the cluster ID and rule ID by IO) and its value will be
+   0/1 to indicate if the recommendation is inactive/active respectively.
+5. There will be a generic/templated info-level alert present on the cluster,
+   which will present the user with brief information about the hitting
+   recommendation (its human-readable name and severity),
+   and the link for details and remediation steps.
```
simonpasquier, line 120: A preliminary version of the alerting rule would be useful too.

simonpasquier (commented):
This looks reasonable to me. I've made a few comments to ask for more details.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+message, etc.).
+
+### Goals
+
+- New Prometheus metric with labels containing aforementioned properties of
+  the recommendation that will indicate if the recommendation is currently active.
+- New Prometheus alert based on this metric, defined using a template that fills
+  the information from labels into the description of the alert.
+- The user should be able to find out when was a certain recommendation active
+  by querying the Prometheus metric (and plot this in Grafana, for example).
```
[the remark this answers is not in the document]
natiiix: Thank you for bringing this to my attention. I had no idea when I was writing this. The change makes perfect sense to me.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+   which are then processed by the CCX data pipeline. (standard behavior)
+2. After a successful archive upload, IO waits for the report to become
+   available, fetches it from the Smart Proxy and processes it. (This is an
+   extension of the current behavior, where we only look at the number of
+   active recommendations per severity level.) (needs to be implemented)
+3. In-cluster Prometheus metrics indicating the status of recommendations
+   being active are set accordingly.
+    - If a recommendation was previously registered as active and is present
+      in the latest Insights report, IO performs no additional action for this
+      recommendation (do not repeat alerts for previously spotted recommendations)
```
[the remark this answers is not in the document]
natiiix: The way it is written, it describes the end-user experience rather than how the Insights Operator should behave. These parenthesized notes were not written by me (they were added by a colleague when we discussed the behaviour), and are not necessary anymore, but an earlier design idea assumed that the alerts would be short-lived, and we wouldn't want to "revive" them, hence the note. I can remove this one if you would prefer that.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+      no longer hitting recommendations)
+    - If a recommendation was previously registered as inactive or the metric
+      for this recommendation is absent and the latest Insights report contains
+      it, a metric will be updated/created to indicate that the recommendation
+      is currently hitting. (indicate new recommendations)
+4. The metric will have labels containing the ID, severity (based on the total
+   risk property), a human-readable name, and a link to a detailed description
+   of the recommendation and its remediation steps (Insights Advisor URL
+   generated based on the cluster ID and rule ID by IO) and its value will be
+   0/1 to indicate if the recommendation is inactive/active respectively.
```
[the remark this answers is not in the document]
natiiix: This is a good question. At maximum, the number of external/customer-facing recommendations that exist at that moment (if the IO will somehow make sure that inactive recommendation metrics are removed). There are quite a few of those, but I do not know the exact number from the top of my head. I would never expect to see more than low 10s of recommendations hitting at once, unless the cluster is in a completely broken state.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: The `InsightsDisabled` alert defined here: https://github.com/openshift/insights-operator/blob/a9e61607f9a3f3656abb6a13747e7f5fdfff4d9b/manifests/08-prometheus_rule.yaml#L14

I will add the information to the enhancement.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: Yes, definitely, but it is very difficult to determine how long we should wait before disabling the alert as a result of an outage that is too long. If we say that 8 hours, as proposed, is enough, then repeated 9 hour outages, in between which the IO runs for an hour, let's say, would cause the alert to be repeatedly disabled and enabled. I would not want to unnecessarily over-engineer the alerts by making the expression too complex, because this is not such a big feature, and it will already be quite complicated as is.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: Yes, if we want to support outdated versions of the Insights Operator indefinitely.

natiiix: Yes, if we want to support outdated versions of the Insights Operator indefinitely.

natiiix: Yes, if we want to support outdated versions of the Insights Operator indefinitely.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: We are not really use about what information would be needed to be passed via the labels exactly, so I tried to provide as much information as we currently have. Let's say we would have something like `InsightsRecommendationActive[id="<id>" name="<human readable name>" severity="critical"] = 1`. This is obviously a very rough example of what it might look like.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: Same as with the metric. These two will be established during the PoC development, as we have no experience with exposing more complicated metrics and defining templated alerts, so we do not know yet how much information we will be able to pass.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Open Questions [optional]
+
+1. Which operator conditions should fall under the unhealthy Insights Operator
+   state?
+1. How long should IO wait before considering a report to be no longer valid
+   (outdated)?
+1. If uploads keep failing, how and when will the alerts be disabled? (because
+   the current design assumes that the metrics are only modified after receiving
+   a report after a successful upload)
```
sdodson, line 183: I'd handle this via alerting on IO report freshness. Whenever there's no report fresher than 2-4x the reporting interval raise an info level alert. Once we cross over to 24hrs or something escalate to warning. But continue reporting metrics based on the most recently received report, is that stateful?

Does Insights have any latency SLOs? ie: condition on cluster happens at 00:00, how long does Insights have to expose that to the admin in either OCM or in-cluster?

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+  by querying the Prometheus metric in the OCP console.
+- Configuration option allowing the user to disable/enable this functionality.
+  Since this feature consumes resources on both sides (cluster and Smart Proxy),
+  it should only be enabled on clusters where the user actually wants to use it. (It should be disabled by default, unless the user explicitly enables it.)
+- The enabling should be done using the same approach as the rest of the
+  Insights Operator configuration (i.e., currently using the `support` secret).
+
+### Non-Goals
+
+- Insights Operator fetching additional data, from the Smart Proxy or any other source.
```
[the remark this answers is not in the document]
natiiix: It's a feature added for the users, not for us, so the motivation to enable it would be that they want to use it. Enabling it by default does not make sense for us because most users will likely not be interested in it, and we wouldn't want to spam them with alerts they don't care about, but most importantly because this feature costs us additional resources (fetching the Insights report adds load onto the Smart Proxy), which would be wasted if the user didn't care about the alerts and didn't bother disabling the feature. I see not benefit from enabling this by default. We expect that no one would use this without configuring some notifications via Alertmanager, so there will always be some configuration necessary to actually utilize the feature for its intended purpose.

I am not really sure what the difference between the two options is, but it is saying that the Insights Operator will be fetching the exact same resources it is fetching already. We are already getting the full report from the Smart Proxy, and will continue to do that. The only difference is that we will extract more data from it than we currently do. The reason I added this point was to clarify that we do not want/need to access a new endpoint, and neither does the Smart Proxy need to expose additional information in the reports it returns. The Smart Proxy will not have to be modified in any way, and IO will be making the same exact request, and will expect the same exact data.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: Okay, so if I understand correctly, the Insights Operator should never be in a Degraded state, because it never directly impacts the performance of the server.

If you believe it is a non-issue to keep the alerts enabled, and add an alert regarding the report "freshness", then I suppose it would be a reasonable option from the IO/implementation side as well. My only worry about this is that we will have to track several conditions separately (IO disabled, IO crashlooping, old report).

Regarding Insights SLOs, I will ask around to get a proper answer, but I believe there is _some_ interval within which an archive should typically be processed and a report become available for it. Typically, it happens in the order of minutes since its upload.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
tremes: I don't think we want to backport it. I guess enabling it by default will introduce a couple of new alerts, which someone (SRE?) might find a little bit annoying (even though they will be only info alerts). I think the motivation for enabling it is to prefer to use the alertmanager (and its own notification configuration) as opposed to the "common" Insights advisor.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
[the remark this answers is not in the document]
tremes: As far as I know we have around 68 external (customer facing) healthchecks right now.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
tremes: Yes we have [InsightsDownloadDegraded](https://github.com/openshift/insights-operator/blob/master/pkg/controller/status/conditions.go#L14) clusteroperator condition and I think it's a good idea to introduce an alert based on this failure.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+      in the latest Insights report, IO performs no additional action for this
+      recommendation.
+    - If a recommendation was previously registered as active and the latest
+      report no longer contains it, the metric will either be reset or
+      unregistered from the Prometheus instance if possible. (remove alerts for
+      no longer hitting recommendations)
+    - If a recommendation was previously registered as inactive or the metric
+      for this recommendation is absent and the latest Insights report contains
+      it, a metric will be updated/created to indicate that the recommendation
+      is currently hitting. (indicate new recommendations)
```
wking, line 111: I think all of this can be collapsed into "Insights operator serves a metric time-series for any recommendations in the most recently retrieved report", and the alert rule you layer on top can handle do some aggregation to bridge outages due to the insights operator pod rolling, etc.  For example, [here][1] is the cluster-version operator looping over ClusterOperator conditions and writing a time-series with different labels for each hit, and [here][2] is the `ClusterOperatorDegraded` rule built on top of that metric.  openshift/cluster-version-operator#643 was an attempt to make a similar alert robust to CVO-pod restarts, but I think we need to add some time aggregation to that still to actually get it working.

[1]: https://github.com/openshift/cluster-version-operator/blob/0e9bc4ef03004fcc2bb0c58d39e5b49445a9f8f6/pkg/cvo/metrics.go#L478-L489
[2]: https://github.com/openshift/cluster-version-operator/blob/0e9bc4ef03004fcc2bb0c58d39e5b49445a9f8f6/install/0000_90_cluster-version-operator_02_servicemonitor.yaml#L88-L100

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+   0/1 to indicate if the recommendation is inactive/active respectively.
+5. There will be a generic/templated info-level alert present on the cluster,
+   which will present the user with brief information about the hitting
+   recommendation (its human-readable name and severity),
+   and the link for details and remediation steps.
+
+It is worth pointing out that if any key part of the Prometheus stack is not
+running as expected on the cluster, then this feature will not work. Prometheus
+manages both the metrics and the alerts, so without it, no alerts can be
+generated by IO.
```
wking, line 125: While we're processing these, you could also dump them into your `ClusterOperator` `status.conditions[]` or other in-cluster Kube-API resource.  But doesn't have to get sorted out in this enhancement request.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+
+**Unhealthy Insights Operator scenario:**
+
+- If something goes wrong with IO, we would not want the Insights alerts to
+  remain hanging on the cluster indefinitely.
+- The alert should only fire if IO has been healthy at any point in the past 10
+  minutes. Healthy, in this context, means that there is no major error
+  happening (crashlooping, degraded operator status, etc.). A temporary error
+  state should not cause the alerts to be disabled.
```
wking, line 134: You can get the fail-silent behavior by not serving the metrics unless you are happy with them.  You can bridge expected outages (e.g. insights pod restarts) with some over-time aggregation in the rule's `expr`.  I don't think you need anything fancier than that here; Prometheus and the Alertmanager will clear the firing alert if the metric that fed it goes away beyond the short interpolation window.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+- If the user intentionally disables IO, the alerts should never fire. If IO is
+  disabled while some of the Insights alerts are firing, the alerts should
+  immediately stop firing (a part of their condition will be that IO is
+  enabled).
```
wking, line 138: I'd make this a condition of the metric serving "the Insights operator will only serve rule metrics when submitting Insights data is enabled", instead of trying to build something into the alert rule's `expr`.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: I'm also concerned with latencies, especially with the "`Meta.LastCheckedAt` is more than 8 hours ago" mentioned below.  My impression is that worst case is:

1. Something goes wrong on the cluster.
2. Telemetry shipped up, so central Insights engine may already notice and have a report ready, but maybe not.
3. Insights operator collects Insights that include the bad thing.
4. Insights operator pushes that Insights tarball up to Red Hat.
5. Red Hat starts processing the new Insights tarball.
6. Insights operator asks for a recent Insights report.  Red Hat hasn't finished processing the upload from the previous step, so we get a stale report.
7. No Insights recommendations in the returned report.
8. Time passes (max 4h between upload attempts?).
9. Red Hat finishes processing the recent Insights tarball and produces recommendations.
10. New gather, new push, new report retrieved.
11. There are Insights recommendations in this new report!  Start serving metric time series for them, alerts go off.
12. Admin sees the alert and mitigates the issue.

And then that whole cycle again before the fact that the fix is landed makes it up off the cluster, gets analyzed, comes back down, and the alert clears.  Depending on caching in the recent-report-server, something like [`If-Modified-Since`][1] would make polling for a new report a bit cheaper, and maybe we could do that more frequently than we upload Insights tarballs themselves?

[1]: https://datatracker.ietf.org/doc/html/rfc7232#section-3.3

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: My personal preference would be to stop serving Insights-rule time series immediately.  Having some [time-aggregation in your alert rule][1] would keep the alerts firing for a few minutes, and if a replacement report retrieval had gone through by then, they'd continue firing without interruption.  Otherwise, you'd have the new alert about "we're having trouble pulling Insights reports", and folks would know they were flying blind, and have a link to the hosted Insights dashboard where they could check on things if they were nervous.

[1]: https://github.com/openshift/enhancements/pull/1036#discussion_r818208034

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+   a report after a successful upload)
+
+### Test Plan
+
+Due to its nature, this feature will probably be best tested using e2e tests
+that involve other components. The Prometheus stack will be necessary to
+evaluate the metrics and alerts.
+
+The e2e tests could work as follows:
+- break some property of the cluster,
```
wking, line 192: Alternatively, public rule for "you look like an OCP CI cluster", and then we add something to an insights e2e-operator and/or an openshift/origin suite that says "yup, Insights is telling my I'm an OCP CI cluster"?

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+
+#### Dev Preview -> Tech Preview
+
+#### Tech Preview -> GA
+
+#### Removing a deprecated feature
+
+### Upgrade / Downgrade Strategy
+
+Downgrading to an older version will remove the alert template definition, if
```
wking, line 215: this doesn't happen automatically; you'd need to teach the alert-resource-creator in the older release to know that it should remove the alerts.  If it's the CVO pushing those rules on your behalf, that means manifests with the [`release.openshift.io/delete` annotation][1].  On the other hand, [rollbacks are not supported][2], and with the older Insights operator no longer serving the backing metrics, having the stale alert rule lying around doesn't seem like that big a deal.

[1]: https://github.com/openshift/enhancements/blob/20183fad84f682c83e60386593c0eca717ee5bc9/enhancements/update/object-removal-manifest-annotation.md#proposal
[2]: https://github.com/openshift/openshift-docs/blame/d4762f0f626a4dddb9d7330e63a3bb6cb73f5bb5/modules/update-upgrading-cli.adoc#L160-L162

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,268 @@
...
+N/A
+
+#### Failure Modes
+
+N/A
+
+#### Support Procedures
+
+If there are active Insights recommendations, but no alerts are firing, check
+the status of the Prometheus stack and the Insights Operator. If both of them
```
wking, line 242: Hah, is this a new rule for "the cluster doesn't seem to know about the rules I told it last time"?

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
tremes: Yes this is a good point. The key point is the number 6 IMO. I think we really need to check the time of the Insights report (and if it's older than let's say 15 min then do not update the metrics? or remove them?). Now the time information is part of the report content (HTTP body) and it would be probably better as some HTTP header, but I wouldn't poll the service (serving the reports) more often.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
dofinn: given the info state - potentially combined with a custom label for explicit identification? SRE should have no issue with this as we can route on sev + label accordingly.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: I basically agree with what @tremes said here. The 8 hours were chosen rather arbitrarily and based on the previous feedback, I'm leaning towards the idea of keep the alerts enabled, and adding another metric/alert to indicate if we expected to receive a new report but do not have it yet (10 minutes since last successful upload, for example), which would stay on until we receive a report is within that time frame from its respective upload time.

The polling is something we wanted to avoid in order to prevent unnecessary load on the Smart Proxy. Instead, if the report fetched after a successful upload is older than expected, we will wait (e.g., with an exponential back-off) and try to get it later. We will keep trying until the next archive upload, but after some time (e.g., 10 minutes, as mentioned above) we would set the metric to indicate that the report is delayed.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: This point was regarding an unhealthy IO state. If that happens, how would we remove the metrics? If IO is crashlooping, then it cannot really do anything about its metrics, right? So that's why I wanted an IO state check to be a part of the alerting expression.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: Okay, I guess this is a non-issue either way, then. IO is managed by CVO, and the alert definitions would be shipped with it, so presumably also managed by CVO. Thanks for pointing this out, and I believe it should be fine to add that annotation to the alert template when we write it.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: That would probably work as well, but I think breaking some generally available cluster configuration might be easier and will resemble the end-user-experience more closely. I feel like this would add additional complexity to an otherwise fairly simple feature.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: I'm not really sure what metric you mean right now, and how this could be used to make sure there are no Insights recommendation alerts when Insights are disabled.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: Some context on [the metrics architecture][1], in case that's helpful (I'm not monitoring, so I might be wrong):

1. You declare a ServiceMonitor, pointing at your Service
2. Prometheus notices the ServiceMonitor, and periodically stops by to scrape any metrics you are serving.
3. If the scrape fails, Prometheus will extrapolate/interpolate to paper over small gaps.
3. If the scrape succeeds, any metrics you are currently serving get stored in the Prom database, and any metrics you are no longer serving get removed from the Prom database.

So:

1. You get a report.
2. You set up `insights_recommendation{someLabel="A",...} 1`, `insights_recommendation{someLabel="B",...} 1`, etc., based on the things you saw in that report.
3. Prom comes by and scrapes those up.
4. Prom comes by again, for another round of scraping.
5. Fetch a new report!  You set up `insights_recommendation{someLabel="A",...} 1`, `insights_recommendation{someLabel="C",...} 1`.  No mention of `B` in this report.  You don't have to clear anything.
6. Prom comes by and scrapes those up.  Hey, a new `C`!  Hmm, I guess `B` is gone.

As long as you aren't needlessly churning label values on the metrics you're serving, you won't be triggering notification churn.  And if I/O crashes halfway through a metrics update, you'll be starting over with a fresh report-fetch and metrics setup, and as far as Prom is concerned that just looks like a healthy pod/container roll.

[1]: https://prometheus.io/docs/introduction/overview/#components

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: Seems simpler to me than finding a reliable way to break something.  But 🤷 , it will be your CI, so if you would rather maintain a cluster-breaking workflow than a CI-focused insights rule and test-case looking for that rule, I have no problem with that.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: if you use the timestamp of the report (or the Insights upload that fed the report) as the metric value, you could have the alert say:

>`$DURATION` ago, there was an Insights recommendation `$SLUG`.  More details at `$URL`

that makes the stale-ness explicit, and my only concern is "users might think a stale alert was talking about their current cluster".  I'm ok with infinitely stale alerts if they explicitly declare how stale they are.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: you don't have to remove metrics.  Unless you are actively serving them, [Prom will notice][1].

[1]: https://github.com/openshift/enhancements/pull/1036#discussion_r819175113

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: See [the scrape architecture note][1].  If you stop serving metrics when the user disables Insights, then Prom will notice that the metrics are gone, and the alert `expr` which are looking for the metrics will have no matches, and any previously-firing alerts will resolve, and no new alerts will start firing for that rule.

[1]: https://github.com/openshift/enhancements/pull/1036#discussion_r819175113

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: Maybe it's worth considering the feedback loop for Insights rules?  Is there something visible to support so they know how to get in touch if they feel like a particular rule is misfiring or not giving sufficient context or something?

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: Hmmm, so instead of having 0/1 as the metric values, we would use the timestamp from the latest report that contained the recommendation. When thinking about adding the timestamp to the metric, I was thinking of putting it into a label, but that seemed like a poor idea because it would change with every report. Putting it into the value instead makes a lot more sense, and if the Prometheus works the way you described it below, the recommendations that were not in the latest report would simply not have their respective metrics set. I'll have to discuss this with the team, and more importantly with someone who has a more experience with Prometheus to see if this could work for us.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: I don't really understand what you mean by not having to clear anything. The

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: Well, my concern was that the creation of another rule tends to be a lengthy process. It doesn't have to be a cluster-breaking workflow, per se, as there are recommendations even for some fairly minor things. Either way, I would ask around to see which option would be easier to implement and maintain into the future, and I will keep your suggestion in mind.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: When you say support, do you mean the customer's support or our support? There is no way to provide feedback through the feature described in this enhancement, but there is a way of providing feedback through the Insights tab in OCM and/or Insights Advisor. I'm not 100% sure about it because I don't really come into contact with either on a daily basis, but that is what I have heard.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: I'm a bit fuzzy on how Prom's Go metrics server is set up, but the way it seems to work in the CVO is:

1. Something calls `Collect` on us periodically (a metrics-serving goroutine?) to ask what metrics it should currently be serving.
2. Inside `Collect`, we [push into the provided channel][1] anything we thing we should be serving right now.
3. The metrics server or whatever called `Collect` updates its internal cache to match.

Periodically Prom's scraper comes around, and gets whatever was in the metrics-server's cache at that time.  So no CVO-specific steps to clear anything on the metrics-serving side either.  That's magically happening in vendored code in step 3.

[1]: https://github.com/openshift/cluster-version-operator/blob/0e9bc4ef03004fcc2bb0c58d39e5b49445a9f8f6/pkg/cvo/metrics.go#L478-L489

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: This section is for "customer opened a support case with Red Hat, Red Hat support thinks this feature is the problem, what do they do next?".  You could narrowly scope that to "the Insights rules are great, and the only reason support would read this section was if there was a bug in the pipeline around getting those down into the cluster and shipping out notifications".  That matches the scope of the actual enhancement, so I don't see any problems with that.  The notes you have are useful.  You could add a step about looking at the metrics if the alert behavior seemed suspicious.  Or hitting the metrics Service endpoint to completely cut Prom components out of the "maybe the Insights operator isn't serving the metrics we expect" diagnostics.  But whatever, I'm agnostic to how detailed you want to be here.

You could also widen the scope here to cover "now the customer is getting poked with repeated notifications as a flaky Insights rule flops between matching and not matching".  That would be a problem already, but this enhancement would make it more visible, so it might be worth confirming that support knows to (and how to) contact an Insights-rule maintainer if they suspect something like that is going on.  I'm agnostic about whether you do this, but "I could not think of anything more meaningful to put into this section" suggested you were feeling this section was a bit sparse, so I was generating some candidates in case you were looking for additional work ;).

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
tremes: That's really interesting. We don't implement any `Collect` like this AFAIK (we only do [this](https://github.com/openshift/insights-operator/blob/master/pkg/insights/insightsclient/insightsclient.go#L245)) and I am wondering if we should. The interface is defined here https://github.com/prometheus/client_golang/blob/main/prometheus/collector.go#L27

natiiix: Okay, I see what you mean, more or less, in the CVO code. Unfortunately, that's not how IO manages it metrics at the moment, so we will probably have to either refactor the IO code to follow this pattern or work around it in some other way, as was originally suggested in the enhancement. From my understanding, and I'll have to double-check this, IO simply sets the metrics ad-hoc, and doesn't really have this `Collect` callback at all.

simonpasquier: I would expect that the insights operator implements the custom collector pattern with const metrics. The operator will probably have an internal representation of the recommendations and it needs to translate them into metrics whenever Prometheus scraping happens. See https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#hdr-Custom_Collectors_and_constant_Metrics for more details.
Trying to use the `GaugeVec`, `CounterVec`, ... types is prone to errors because their interfaces don't make it easy to remove labels.

wking: > Trying to use the `GaugeVec`, `CounterVec`, ... types is prone to errors because their interfaces don't make it easy to remove labels.

Will we need to remove labels for this use-case?  It sounded like we'd have a recommendation ID, name/slug, active-ness, and detail URI.  And that we'd have those same labels in every exported entry.  That sounds pretty much identical to the CVO's `cluster_operator_conditions` to me, and we are using `NewGaugeVec` for that, and exporting information for multiple conditions for multiple operators simultaneously.

natiiix: So the suggestion is to use individual const metrics for each recommendation, with their respective labels? I don't know if removing labels, as you mention, is something we need to do in IO.

I'm not really very familiar with const metrics because we have not used this Collector pattern before in IO. Would it be okay to use a gauge type with the const metrics and store the timestamp of the most recent report with the recommendation in it? It's not exactly clear to me from the documentation if it's only "const" in the context of how it's created or if the Prometheus service will prevent us from changing it once it has been created (because with each report, we may need to update that timestamp).

Then again, I don't think it makes sense to store the timestamp in each metric because their values would all be the same. We could stick to the binary metrics for recommendations and have another metric contain the timestamp of the last report. Since every recommendation was either in the latest report or it isn't active, this should not be an issue.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,277 @@
...
+  immediately stop firing (the in-cluster will detect that IO is not active
+  during its period scraping, which will remove the Insights metrics from its
+  database).
+- If IO was disabled by an accident, the user will be notified via the existing
+  `InsightsDisabled` alert, which is defined as a part of the Insights Operator.
+
+**Unhealthy Smart Proxy scenario:**
+
+If the processing of a successfully uploaded archive takes unexpectedly long, a
+Prometheus metric indicating this delay should be set. This would most likely be
```
tremes, line 151: > If the processing of a successfully uploaded archive takes unexpectedly long, a Prometheus metric indicating this delay should be set

I would reuse our existing `InsightsDownloadDegraded` cluster operator condition in this case and then introduce a new alert based on this condition as discussed. So I wouldn't create any new metric.

natiiix: Sure, that makes sense, and it will be easier to manage together instead of dealing with two separate indicators of the download delay.

tremes:
One minor typo. Otherwise it looks good to me. Please squash the commits.
/approve

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,286 @@
...
+3. Prometheus metrics indicating the status of recommendations being active are
+   set accordingly. Only metrics for active recommendations are reported during
+   the metric collection. If a recommendation is no longer active, the metric
+   will not be collected, and the in-cluster Prometheus is expected to delete
+   it, which in turns should cancel the corresponding alert.
+4. The metric will have labels containing the ID, severity (based on the total
+   risk property), a human-readable name, and a link to a detailed description
+   of the recommendation and its remediation steps (Insights Advisor URL
+   generated based on the cluster ID and rule ID by IO) and its value will be
+   0/1 to indicate if the recommendation is inactive/active respectively.
```
wking, line 116: The 0/inactive case is stale, isn't it?  From the previous point, "Only metrics for active recommendations are reported during the metric collection.".

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
wking: #1037 might include these in `status` for the new resource.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,286 @@
...
+   and the link for details and remediation steps.
+
+It is worth pointing out that if any key part of the Prometheus stack is not
+running as expected on the cluster, then this feature will not work. Prometheus
+manages both the metrics and the alerts, so without it, no alerts can be
+generated by IO.
+
+An example of what the Insights recommendation metric may look like:
+```text
+insights_recommendation_active{rule_id="ccx_rules_ocp.external.rules.empty_prometheus_db_volume.report",error_key="PROMETHEUS_DB_VOLUME_IS_EMPTY",severity="moderate",description="Prometheus metrics data will be lost when the Prometheus pod is restarted or recreated",info_url="http://example.com/"} 1646416494
```
wking, line 129: `description` here is a longer string than I am used to seeing in Prom metrics.  And it doesn't seem to be called out in your earlier:

>  ...will have labels containing the ID, severity (based on the total risk property), a human-readable name, and a link to a detailed description of the recommendation and its remediation steps (Insights Advisor URL...)

You might also want to consider `rule_name` or just `name` or some such instead of `error_key`.  Insights rules might not even be errors, right?

natiiix: The `rule_name` and `error_key` are field names that we use throughout the Insights stack, so those are not just some arbitrarily chosen label names.

Description is the name of the field that contains the human-readable name mentioned in the quoted paragraph, and its length was one of the concerns I had initially, but after some discussions, we came to the conclusion that it should not be an issue.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
natiiix: You are right, I will update this part.

enhancements/insights/prometheus-alerts-for-insights-recommendations.md
```diff
@@ -0,0 +1,286 @@
...
+  disabled while some of the Insights alerts are firing, the alerts should
+  immediately stop firing (the in-cluster will detect that IO is not active
+  during its period scraping, which will remove the Insights metrics from its
+  database).
+- If IO was disabled by an accident, the user will be notified via the existing
+  `InsightsDisabled` alert, which is defined as a part of the Insights Operator.
+
+**Unhealthy Smart Proxy scenario:**
+
+If the processing of a successfully uploaded archive takes unexpectedly long, a
```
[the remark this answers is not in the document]
natiiix: Thanks for pointing that out. I will fix that.

bparees:
team doing the implementation should apply the final /lgtm

PR #1038, opened 2022-02-21
Refine how a resource specifies multiple capabilities

bparees:


enhancements/installer/component-selection.md
```diff
@@ -198,12 +198,12 @@ The CVO will filter out(not apply/reconcile) resources that are annotated with k
...
-a `ServiceMonitor` which is part of both the `Monitoring` capability and the `Foo` capability.  If either
-of those capabilities are excluded, the `ServiceMonitor` should not be created.  (If Monitoring is excluded
-then no ServiceMonitors should be created, if only Foo is excluded, then only Foo's ServiceMonitors should
-be excluded).
+If a resource participates in multiple capabilities, it should specify all capabilities as comma separated
+values in the annotation.  The resource will be filtered if any of the annotations on the resource are matched
+by a configured filter.  This allows a component to define a `ServiceMonitor` which is part of both the 
+`Monitoring` capability and the `Foo` capability.  If either of those capabilities are excluded, the 
+`ServiceMonitor` should not be created.  (If Monitoring is excluded then no ServiceMonitors should be created,
+if only Foo is excluded, then only Foo's ServiceMonitors should be excluded).
```
bparees, line 206: @wking @deads2k thanks to @jottofar we realized the EP was suggesting multiple annotations w/ the same key, which obviously can't work (originally the annotation was supposed to include the cap name in the key).

Do you see any challenges w/ this scheme instead?  e.g.:

`capability.openshift.io/name=console,monitoring`

bparees:
reasserting the comment i invalidated w/ a new commit:


@wking @deads2k thanks to @jottofar we realized the EP was suggesting multiple annotations w/ the same key, which obviously can't work (originally the annotation was supposed to include the cap name in the key).

Do you see any challenges w/ this scheme instead? e.g.:

capability.openshift.io/name=console,monitoring

enhancements/installer/component-selection.md
wking: Change looks good to me :)

deads2k:
@wking suggested a `+` as a delimiter.

I personally prefer an explicit name and separate requires, but asking a likely consumer of the API showed his preference as a single list, so I'll defer to the actual API consumer here and reserve an "i told you so" for 18 months from now :)

/approve

bparees:
I like using a `+` delimeter to make the `and` behavior clearer.  i'll change it to that.

bparees:
I also sort of like the idea of `capability.openshift.io/requires` (i'm not a fan of provides because the resource doesn't provide the cap, it's part of providing the cap), but fundamentally the concept feels fuzzy either way:

"to apply this resource, we require you enable this cap", but also "applying this resource provides(part of) this cap"

PR #1042, opened 2022-02-23
Change cap annotation delimiter to +

bparees:


bparees:
self-merging as the comments in https://github.com/openshift/enhancements/pull/1038 seemed to be in support of this.

PR #1000, opened 2022-01-07
Update enhancement proposal to support update of user-defined AWS tags 

TrilokGeer:
This PR addresses design considerations to support update of user-defined AWS tags.
Also, details are listed for graduation criteria for tech preview and GA.

PR #1044, opened 2022-02-25
Revert "Update enhancement proposal to support update of user-defined AWS tags "

sdodson:
Reverts openshift/enhancements#1000


sdodson:
https://github.com/openshift/enhancements/pull/1000#issuecomment-1050825760 indicates that there were still items which needed to be reconciled.

PR #1048, opened 2022-03-02
Dual Stack VIPs for on-prem deployments

cybertron:
Originally the on-prem loadbalancer architecture supported only single-stack
ipv4 or ipv6 deployments. When dual stack support was added for the baremetal
platform, we carried over the VIP configuration that only allowed for a single
VIP to be specified. The thinking was that in dual stack clusters every node
would have access to either ipv4 or ipv6 endpoints. While this is true of the
nodes, it is not necessarily true of the applications that run in the cluster
and may need to interact with the API or ingress services.

This design is proposing the addition of a second VIP for dual stack clusters
so API and ingress are accessible from either ip version.

### DIFF
--- a/enhancements/network/on-prem-dual-stack-vips.md
+++ b/enhancements/network/on-prem-dual-stack-vips.md
@@ -1,0 +1,339 @@
+---
+title: dual-stack-vips
+authors:
+  - @cybertron
+reviewers:
+  - @creydr
+approvers:
+  - @patrickdillon
+  - @kirankt
+  - @shardy
+  - someone from MCO
+api-approvers:
+  - ???
+creation-date: 2022-03-01
+last-updated: 2022-03-02
+tracking-link:
+  - https://issues.redhat.com/browse/SDN-2213
+see-also:
+  - "/enhancements/on-prem-service-load-balancers.md"
+replaces:
+superseded-by:
+---
+
+# Dual Stack VIPs
+
+## Summary
+
+Originally the on-prem loadbalancer architecture supported only single-stack
+ipv4 or ipv6 deployments. When dual stack support was added for the baremetal
+platform, we carried over the VIP configuration that only allowed for a single
+VIP to be specified. The thinking was that in dual stack clusters every node
+would have access to either ipv4 or ipv6 endpoints. While this is true of the
+nodes, it is not necessarily true of the applications that run in the cluster
+and may need to interact with the API or ingress services.
+
+This design is proposing the addition of a second VIP for dual stack clusters
+so API and ingress are accessible from either ip version.
+
+## Motivation
+
+Customer applications on dual stack clusters may only have ipv6 connectivity,
+but the cluster VIPs will be ipv4 (at this time ipv6 VIPs are not supported
+in dual stack clusters).
+
+### Goals
+
+Create both ipv4 and ipv6 VIPs on dual stack clusters so single stack
+applications have the connectivity they require.
+
+### Non-Goals
+
+Nothing in particular. The problem being solved here is fairly contained, so
+scope creep should not be a problem.
+
+## Proposal
+
+### User Stories
+
+As a deployer of a dual stack OpenShift cluster, I want to run ipv6-only
+applications that need access to the API and ingress VIPs.
+
+### API Extensions
+
+New fields will have to be added to the
+[platform status section](https://github.com/openshift/api/blob/24043075985b3dc87190174241f3c4d0416b8ea2/config/v1/types_infrastructure.go#L465)
+of the infrastructure object to persist the specified additional VIPs.
+
+### Implementation Details/Notes/Constraints [optional]
+
+This should be a relatively simple feature to implement since most of the
+new code, manifests, and templates will be slight modifications of the
+existing loadbalancer ones. The main thing we need to determine is what the
+structure of the install-config and api changes will be. That will have an
+impact on the implementation of these new VIPs, but any of the designs
+proposed below should be doable. It's mostly a question of picking the one
+that will provide the best user experience and, secondarily, the simplest
+implementation.
+
+### Risks and Mitigations
+
+Minimal risk. This is just adding another VIP, something we already have in
+our deployments. The main concern would be logic errors arising out of
+the need to handle multiple VIPs, which (if they happen) will need to be
+addressed as bugs.
+
+## Design Details
+
+This feature will require changes in a few different components:
+
+Installer - New fields will need to be added to install-config for deployers
+to specify the additional VIPs.
+
+API - New fields will need to be added to the platform status section of the
+infrastructure object.
+
+machine-config-operator - The on-prem loadbalancer templates and manifests
+will need to be updated to support the additional VIPs.
+
+baremetal-runtimecfg - There is logic that depends on the ip version of the
+VIP. If there are multiple VIPs, this will have to be modified to account for
+that. In addition, the render code will need the new values wired in.
+
+### install-config
+
+Currently the VIPs are specified as follows:
+
+```
+platform:
+  baremetal:
+    apiVIP: 192.168.111.5
+    ingressVIP: 192.168.111.4
+```
+
+There are a few options for how we could add the new VIPs (we need to pick
+one and move the others to the alternatives section).
+
+#### Just add v6 versions of the VIP fields
+```
+platform:
+  baremetal:
+    apiVIP: 192.168.111.5
+    apiVIPv6: f00::5
+    ingressVIP: 192.168.111.4
+    ingressVIPv6: f00::4
+```
+This is a minimal change, but it leaves us with somewhat of a problem in that
+we can't assume the old fields always contain a v4 address since previous
+releases allowed either v4 or v6. We may be able to deal with that by
+detecting the appropriate ip version for the old fields and populating new
+fields on the api side. E.g. if apiVIP is v4, we populate
+`platformStatus.apiServerInternalIPv4` and if it is v6 we populate
+`apiServerInternalIPv6` instead. That way the only place we need to worry about
+the difference is in the installer. We would need a validation to ensure a v6
+VIP isn't specified in both fields.
+
+Note that this assumes we replace the existing api fields with two new ones
+per VIP and do not reuse the existing ones at all. I think we're going to want
+to do that regardless of what install-config layout we use so we don't need
+logic on the backend in machine-config-operator and baremetal-runtimecfg to
+handle older cluster configs.
+
+#### Deprecate the old fields and add new v4 and v6 fields
+```
+platform:
+  baremetal:
+    # Deprecated
+    apiVIP: 192.168.111.5
+    # Deprecated
+    ingressVIP: 192.168.111.4
+
+    apiVIPv4: 192.168.111.5
+    apiVIPv6: f00::5
+    ingressVIPv4: 192.168.111.4
+    ingressVIPv6: f00::4
+```
+This is more of a 1-1 mapping to what I anticipate the platformStatus api
+will look like. We would still need compatibility logic to handle old-style
+install-configs that set the deprecated fields, but having no overlap between
+the old and new-style configs might simplify the logic somewhat.
+
+#### Add plural fields that take a list of VIPs
+```
+platform:
+  baremetal:
+    # Deprecated
+    apiVIP: 192.168.111.5
+    # Deprecated
+    ingressVIP: 192.168.111.4
+
+    apiVIPs:
+    - 192.168.111.5
+    - f00::5
+    ingressVIPs:
+    - 192.168.111.4
+    - f00::4
+```
+This is essentially a variation on the previous option where the two separate
+fields per VIP are combined into a list field with a length of up to 2. In
+this case we'd need some logic to determine whether the VIPs in the list are
+v4 or v6, but I expect that would only be a little more complicated than the
+other options. This has the advantage of being extensible if we ever wanted
+to add more addresses, but I admit to having a hard time coming up with a use
+case where more than an ipv4 and ipv6 address would be needed.
+
+### openshift/api
+
+We will need to make changes to the
+[baremetal platformStatus](https://github.com/openshift/api/blob/354aa98a475c1fcd60b41aaee735da95d7318b42/config/v1/0000_10_config-operator_01_infrastructure.crd.yaml#L311)
+to add new fields for the new VIPs. Unlike the install-config layout, I think
+it is clearly better to add discrete new fields for this feature. That way,
+backend code in machine-config-operator and baremetal-runtimecfg will not
+need to detect whether the values in the old fields are from a new (where
+the field will have to be v4) or old (where the existing field can be either
+v4 or v6) configuration.
+
+Ideally we would transition existing values to the new fields on upgrade.
+I'm unsure whether a mechanism exists for us to do that, however. If not, then
+some logic (probably in MCO) will be needed to map values from the old fields
+to the appropriate ip version.
+
+The new structure of platformStatus would look like this:
+
+```
+baremetal:
+  description: BareMetal contains settings specific to the BareMetal platform.
+  type: object
+  properties:
+    # Deprecated
+    apiServerInternalIP:
+      description: apiServerInternalIP is an IP address...
+      type: string
+    apiServerInternalIPv4:
+      description: apiServerInternalIPv4 is an IP address...
+      type: string
+    apiServerInternalIPv6:
+      description: apiServerInternalIPv6 is an IP address...
+      type: string
+    # Deprecated
+    ingressIP:
+      description: ingressIP is an external IP...
+      type: string
+    ingressIPv4:
+      description: ingressIPv4 is an external IP...
+      type: string
+    ingressIPv6:
+      description: ingressIPv6 is an external IP...
+      type: string
+    nodeDNSIP:
+    ...
+```
+
+### machine-config-operator
+
+We will need to wire in the new VIPs to the manifests and configuration
+templates. There should be conditionals around the v4 and v6 sections
+so only the appropriate ones are enabled based on the provided VIPs.
+
+I believe haproxy already listens on both ipv4 and ipv6 so there should be no
+changes needed in that configuration. We'll need to verify that as part of
+this implementation though. Note that as of this writing, the apiserver
+does not support dual stack anyway so all traffic through haproxy will end
+up on one or the other ip version.
+
+As noted above, it would be best to migrate the existing VIP fields to the new
+v4 and v6 fields to simplify the logic in consumers. We may be able to do that
+in the
+[merge code run on upgrade](https://github.com/openshift/machine-config-operator/blob/080919dc992d600706b679eb118393ee293496f7/lib/resourcemerge/machineconfig.go#L68).
+I'm unsure whether that can modify the actual infrastructure object, but
+it can massage the data put into controllerconfig, which is where we actually
+get the values.
+
+### baremetal-runtimecfg
+
+There are a number of places in runtimecfg where a VIP is used to determine
+which ip version to use (for example, when deciding whether to use the ipv4
+or ipv6 addresses in the unicast_peer list). This logic will need to be
+updated to handle the fact that there may be both v4 and v6 VIPs.
+
+Additionally, the new VIPs will need to be wired in to the rendering code.
+
+### Open Questions [optional]
+
+- See above options for install-config structure
+- If we add the v4 and v6 VIP fields to the api as entirely new fields, is
+  there a way for us to take the values from the old fields in existing
+  clusters and map them to the appropriate v4 or v6 field on upgrade?
+
+### Test Plan
+
+Most of this feature will be covered by the existing dual stack test jobs.
+A few added tests will be needed to verify the functionality of the new VIPS.
+
+### Graduation Criteria
+
+We do not anticipate needing a graduation process for this feature. The
+internal loadbalancer implementation has been around for a number of
+releases at this point and we are just extending it.
+
+#### Dev Preview -> Tech Preview
+
+NA
+
+#### Tech Preview -> GA
+
+NA
+
+#### Removing a deprecated feature
+
+NA
+
+### Upgrade / Downgrade Strategy
+
+Upgrades and downgrades will be handled the same way they are for the current
+internal loadbalancer implementation. On upgrade, existing VIP configurations
+will be maintained. We will not (and cannot) automatically add new VIPs to a
+cluster on upgrade. If a deployer of an older dual stack cluster wants the new
+VIP functionality that will have to be a separate operation from upgrade.
+
+### Version Skew Strategy
+
+This will also be handled as it is today. The keepalived configuration used
+for the new version must work with the old version until all nodes have been
+upgraded. If breaking changes are needed, there will be a migration mechanism
+that runs after upgrade is complete (similar to what we did in the multicast
+to unicast migration).
+
+### Operational Aspects of API Extensions
+
+NA
+
+#### Failure Modes
+
+NA
+
+#### Support Procedures
+
+NA
+
+## Implementation History
+
+4.11: Initial implementation
+
+## Drawbacks
+
+Nothing significant. A very small amount of compute resources will be used
+to manage the new VIPs. Even this can be avoided by simply not specifying
+the second VIP if it is not needed.
+
+## Alternatives
+
+Implement this somewhat like the
+[existing workaround](https://github.com/yboaron/kepalived-v6) - create a
+separate instance of keepalived that is responsible for ipv6 only. This
+results in some duplication of containers and configs, but it would probably
+simplify the code changes.
+
+## Infrastructure Needed [optional]
+
+None
--- a/enhancements/network/on-prem-dual-stack-vips.md
+++ b/enhancements/network/on-prem-dual-stack-vips.md
@@ -15,7 +15,7 @@ api-approvers:
   - @danwinship
   - @aojea
 creation-date: 2022-03-01
-last-updated: 2022-03-24
+last-updated: 2022-04-01
 tracking-link:
   - https://issues.redhat.com/browse/SDN-2213
 see-also:
@@ -160,7 +160,7 @@ baremetal:
       description: ingressIP is an external IP...
       type: string
     ingressIPSecondary:
-      description: ingressIPv4 is an additional external IP...
+      description: ingressIPSecondary is an additional external IP...
       type: string
     nodeDNSIP:
     ...
@@ -405,7 +405,7 @@ baremetal:
       description: ingressIP is an external IP...
       type: string
     ingressIPs:
-      description: ingressIPv4 is a list of  external IPs...
+      description: ingressIPs is a list of  external IPs...
       type: array
     nodeDNSIP:
     ...
--- a/enhancements/network/on-prem-dual-stack-vips.md
+++ b/enhancements/network/on-prem-dual-stack-vips.md
@@ -52,9 +52,11 @@ applications have the connectivity they require.
 
 ### Non-Goals
 
-Making the `kubernetes.default` Service dual stack is not supported and will
-not be addressed by this work. We are only adding dual stack access to the
-API via the VIPs.
+* Making the `kubernetes.default` Service dual stack is not supported and will
+not be addressed by this work.
+
+* Configuration of any VIPs beyond a second ipv6 one for dual stack clusters.
+MetalLB is a better solution to creating arbitrary loadbalancers.
 
 ## Proposal
 
@@ -104,6 +106,8 @@ baremetal-runtimecfg - There is logic that depends on the ip version of the
 VIP. If there are multiple VIPs, this will have to be modified to account for
 that. In addition, the render code will need the new values wired in.
 
+assisted-service - the hive API extension is used for agent-based cluster creation as a stand-alone tool ("Infrastructure Operator") and as part of RHACM. It will need new API fields that correspond to the changes in the install-config.
+
 ### install-config
 
 Currently the VIPs are specified as follows:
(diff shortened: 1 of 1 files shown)


enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,339 @@
...
+changes needed in that configuration. We'll need to verify that as part of
+this implementation though. Note that as of this writing, the apiserver
+does not support dual stack anyway so all traffic through haproxy will end
+up on one or the other ip version.
+
+As noted above, it would be best to migrate the existing VIP fields to the new
+v4 and v6 fields to simplify the logic in consumers. We may be able to do that
+in the
+[merge code run on upgrade](https://github.com/openshift/machine-config-operator/blob/080919dc992d600706b679eb118393ee293496f7/lib/resourcemerge/machineconfig.go#L68).
+I'm unsure whether that can modify the actual infrastructure object, but
```
cgwalters, line 248: Hmm, I don't think the MCO should be mutating the infrastructure object.

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,339 @@
...
+it can massage the data put into controllerconfig, which is where we actually
+get the values.
```
cgwalters, line 250: Right yes, the controllerconfig is just an internal implementation detail of the MCO we can change at will.

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,339 @@
...
+the difference is in the installer. We would need a validation to ensure a v6
+VIP isn't specified in both fields.
+
+Note that this assumes we replace the existing api fields with two new ones
+per VIP and do not reuse the existing ones at all. I think we're going to want
+to do that regardless of what install-config layout we use so we don't need
+logic on the backend in machine-config-operator and baremetal-runtimecfg to
+handle older cluster configs.
+
+#### Deprecate the old fields and add new v4 and v6 fields
```
cgwalters, line 142: I like this option the best.  I'd say though it's probably good if this is a relatively "quiet" deprecation at least to start; the cost of carrying the auto-detection code seems low.

cgwalters (approved):
Very well written enhancement, overall LGTM!  Just some minor things.

(Feel free to at least skim https://github.com/openshift/enhancements/pull/1032 and review/approve in trade for this review :wink: )

enhancements/network/on-prem-dual-stack-vips.md
cybertron: Yeah, I'm looking at this deprecation more as a "don't use this anymore" than "this is going away". From previous discussions with the installer team I'm not sure we're supposed to break old install-configs anyway, so we may have to carry it indefinitely.

enhancements/network/on-prem-dual-stack-vips.md
cybertron: It certainly doesn't seem right. If there's somewhere more appropriate that we could put migration code I'd be open to that, but if not I think we can live with a slight mismatch between what's in infrastructure and what's in controllerconfig.

kikisdeliveryservice: I think you'd be looking to make changes here in createDiscoveredControllerConfigSpec() to make sure that the correct fields are populated in the controller config: https://github.com/openshift/machine-config-operator/blob/61159678d9bf051a5e8a017210a349f8c643b910/pkg/operator/render.go#L111

enhancements/network/on-prem-dual-stack-vips.md
kikisdeliveryservice: +1 Agree with Colin. This also kind of sounds like the Platform/PlatformStatus thing we had to do early on. Where old clusters had an empty new field and we had to populate it correctly for born-in 4.1-ish clusters so they wouldnt break. Overall, we can't break people but it's not the first time we had to do something like this so it should be ok.

cybertron: Cool, this is the one I would tend to prefer as well since it matches what I'm proposing for the api side, and it's good to hear we're all in agreement so far.

enhancements/network/on-prem-dual-stack-vips.md
cybertron: Thanks, I'll add that link to this section as well.

cybertron (commented):
Just a note that I have fixed the lint errors locally, but I'm holding off on pushing a new version of this to avoid orphaning comments. I'll push the cleaned up version after I get someone from the api and installer teams to look at this.

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,339 @@
...
+Originally the on-prem loadbalancer architecture supported only single-stack
+ipv4 or ipv6 deployments. When dual stack support was added for the baremetal
+platform, we carried over the VIP configuration that only allowed for a single
+VIP to be specified. The thinking was that in dual stack clusters every node
+would have access to either ipv4 or ipv6 endpoints. While this is true of the
+nodes, it is not necessarily true of the applications that run in the cluster
+and may need to interact with the API or ingress services.
+
+This design is proposing the addition of a second VIP for dual stack clusters
+so API and ingress are accessible from either ip version.
```
aojea, line 37: cc @danwinship for API server dual-stack https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2438-dual-stack-apiserver

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,339 @@
...
+
+We will need to wire in the new VIPs to the manifests and configuration
+templates. There should be conditionals around the v4 and v6 sections
+so only the appropriate ones are enabled based on the provided VIPs.
+
+I believe haproxy already listens on both ipv4 and ipv6 so there should be no
+changes needed in that configuration. We'll need to verify that as part of
+this implementation though. Note that as of this writing, the apiserver
+does not support dual stack anyway so all traffic through haproxy will end
+up on one or the other ip version.
```
danwinship, line 242: FTR, the apiserver situation is slightly more complicated than this: it doesn't support having the `kubernetes.default` Service be dual-stack, and it doesn't have any way of recognizing / talking about dual-stack IPs, but it _can_ take advantage of Linux's support for dual-stack listening sockets. Currently cluster-kube-apiserver-operator explicitly passes `bindNetwork: tcp4` or `bindNetwork: tcp6` in the kube-apiserver config to prevent that, but if we changed it to pass `bindNetwork: tcp` instead, then we'd get an apiserver that accepted connections on both the IPv4 and IPv6 IPs of its node (even though it would only create `Endpoints` for one family or the other).

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,339 @@
...
+      description: ingressIPv6 is an external IP...
+      type: string
+    nodeDNSIP:
+    ...
+```
+
+### machine-config-operator
+
+We will need to wire in the new VIPs to the manifests and configuration
+templates. There should be conditionals around the v4 and v6 sections
```
danwinship, line 235: So if you're going to be fiddling with all this stuff anyway [we may be able to fix UPI node IP detection too](https://bugzilla.redhat.com/show_bug.cgi?id=1949827#c59):

> Currently, for UPI, if the MCO knows the apiserver IP at install time, it will write it into the nodeip-configuration service, and so we will pick the IP with the most direct route to the apiserver (just like in the IPI case). However, in some cases (notably vSphere UPI), the MCO does not know the apiserver IP at install time and thus cannot do this. From what I can tell though, the apiserver IP is still *known* (eg, to the installer) in this case, it's just not recorded anywhere that the MCO can see at install time. So if we fix the install-time config plumbing so that the MCO always has access to the configured apiserver IP, then it should always be able to pass that to nodeip-configuration, and we should always get the right IP. (So this probably requires one or more additions to openshift/api, plus changes to openshift/installer to fill in the new API, and changes to openshift/machine-config-operator to consume the new API. It should not require any changes to openshift/baremetal-runtimecfg.)

I haven't dug too deeply into this but it seems like while we may need to continue having platform-specific `spec` VIP fields, we might be able to have platform-_independent_ `status` VIP fields? (Which would really simplify [this](https://github.com/openshift/machine-config-operator/blob/bbe78cfe1eb9bfa9dc5e8fe27fb8155e648dd848/pkg/operator/render.go#L258).)

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,339 @@
...
+    apiVIPv6: f00::5
+    ingressVIPv4: 192.168.111.4
+    ingressVIPv6: f00::4
+```
+This is more of a 1-1 mapping to what I anticipate the platformStatus api
+will look like. We would still need compatibility logic to handle old-style
+install-configs that set the deprecated fields, but having no overlap between
+the old and new-style configs might simplify the logic somewhat.
+
+#### Add plural fields that take a list of VIPs
```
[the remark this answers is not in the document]
cybertron: Hmm, so I was envisioning the pseudo-config template to have a conditional second VIP:
```
{{ if Cluster.APIVIPv4 }}
vrrp_instance apiv4 {
  virtual_ipaddress { {{ Cluster.APIVIPv4 }} }
  unicast_peers {
    {{- range .LBConfig.Backendsv4 -}}
    {{.Address}}
    {{- end}}
  }
}
{{ end }}
{{ if Cluster.APIVIPv6 }}
vrrp_instance apiv6 {
  # Same thing but s/v4/v6/
}
```
But we should be able to do the same thing as a loop, as long as we can keep the VIP list and backend list consistent with each other. I'm not exactly sure how to iterate lists in parallel in a Go template, but I assume there's some way to do it. If not we may have to modify the runtimecfg template data structure so there's a single top-level list.

Not having specific v4 and v6 VIP fields would also fix the UX issue you pointed out where the user needs to make sure to set the right field in single stack deployments.

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,339 @@
...
+### openshift/api
+
+We will need to make changes to the
+[baremetal platformStatus](https://github.com/openshift/api/blob/354aa98a475c1fcd60b41aaee735da95d7318b42/config/v1/0000_10_config-operator_01_infrastructure.crd.yaml#L311)
+to add new fields for the new VIPs. Unlike the install-config layout, I think
+it is clearly better to add discrete new fields for this feature. That way,
+backend code in machine-config-operator and baremetal-runtimecfg will not
+need to detect whether the values in the old fields are from a new (where
+the field will have to be v4) or old (where the existing field can be either
+v4 or v6) configuration.
```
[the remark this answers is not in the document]
cybertron: If we're using a list for install-config, I wonder if we just make these lists too. If we're going to use the secondary terminology for the extra VIP we might as well do the same in install-config and then we don't have to deprecate any fields, which might be a point in favor of this approach. Maybe I'll write these two options up as alternatives and see if that shakes out any details that would make one clearly better than the other.

We have the flexibility to mix and match data structures from install-config -> api -> render data, but it's simpler if we avoid bouncing between discrete fields and lists.

enhancements/network/on-prem-dual-stack-vips.md
cybertron: Yeah, this was a bit hand-wavy because I couldn't remember exactly what you said was missing from dual stack in the apiserver. I'll rewrite this to be more accurate.

enhancements/network/on-prem-dual-stack-vips.md
cybertron: Hmm, that may be possible. I'm not familiar with UPI installer data, but if it's just adding a field somewhere it's probably doable.

cybertron (commented):
Thanks for all the discussion! I'll try to get a new revision of this pushed tomorrow that incorporates the feedback so far.

cybertron:
Okay, after some more mulling I think the primary/secondary arrangement is going to be best for both user experience and ease of implementation. Not having to deprecate any fields limits the impact of this on users, and also removes the need to migrate any values from existing api records. We can always convert the two fields into a list for processing if we need to.

I've updated the enhancement to reflect this, and left the other options as subsections in the alternatives section for reference.

@danwinship @kikisdeliveryservice @cgwalters please take another look and let me know if you agree with my conclusions.

danwinship (commented):
seems simple. one question:

kikisdeliveryservice:
From an MCO perspective this LGTM

/lgtm

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+
+I believe haproxy already listens on both ipv4 and ipv6 so there should be no
+changes needed in that configuration. We'll need to verify that as part of
+this implementation though.
+
+We could also modify the apiserver configuration so it listens on both v4 and
+v6, but since there's currently no way to set up the `kubernetes.default`
+Service as dual stack this doesn't accomplish much. Traffic to the
+secondary IP version will have to come in via the VIP, which already has
+haproxy to handle both versions.
```
aojea, line 195: does this mean that haproxy does NAT64?

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+the difference is in the installer. We would need a validation to ensure a v6
+VIP isn't specified in both fields.
+
+Note that this assumes we replace the existing api fields with two new ones
+per VIP and do not reuse the existing ones at all. I think we're going to want
+to do that regardless of what install-config layout we use so we don't need
+logic on the backend in machine-config-operator and baremetal-runtimecfg to
+handle older cluster configs.
+
+#### Deprecate the old fields and add new v4 and v6 fields
```
kirankt, line 306: This would be my pick for the installer API change. The fields are unambiguous to the end user. Granted that this option is a little more intensive to implement, it shouldn't be too complicated. We can start using the new v4/v6 fields right away and add the correct default (for ipv4, ipv4 or dualstack) to the deprecated fields.

kirankt (commented):
Overall the proposal is clear and the broad idea is fine by me. Although I vote for alternate 2 design for install config changes. I'd like to see input from @patrickdillon and @staebler as well. Thanks

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+
+Making the `kubernetes.default` Service dual stack is not supported and will
+not be addressed by this work. We are only adding dual stack access to the
+API via the VIPs.
+
+## Proposal
+
+### User Stories
+
+As a deployer of a dual stack OpenShift cluster, I want to run ipv6-only
```
romfreiman, line 63: Another significant story is hub/spoke mode where the spoke is ipv6 only

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+
+### baremetal-runtimecfg
+
+There are a number of places in runtimecfg where a VIP is used to determine
+which ip version to use (for example, when deciding whether to use the ipv4
+or ipv6 addresses in the unicast_peer list). This logic will need to be
+updated to handle the fact that there may be both v4 and v6 VIPs.
+
+Additionally, the new VIPs will need to be wired in to the rendering code.
+
```
romfreiman, line 205: Lets add a section here how will it affect assisted installer / ZTP. Once of the main goals of this work is to allow ZTP deployments of similar topologies

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+    apiVIPv6: f00::5
+    ingressVIPv4: 192.168.111.4
+    ingressVIPv6: f00::4
+```
+This is more of a 1-1 mapping to what I anticipate the platformStatus api
+will look like. We would still need compatibility logic to handle old-style
+install-configs that set the deprecated fields, but having no overlap between
+the old and new-style configs might simplify the logic somewhat.
+
+#### Add plural fields that take a list of VIPs
```
mkowalski, line 325: I'm wondering if this data structure wouldn't be the most in-line with other network parameters configurable nowadays. We have already `clusterNetworks`, `serviceNetworks` and so on that are plural, what does not necessarily mean that they accept every single combination.

When in Assisted Installer we were implementing dual-stack support, we had similar concerns - whether fields that in single-stack accept 1 entry and in dual-stack exactly 2 entries deserve to be lists or should we take the simplest path by introducing `-secondary` suffix. We picked the former as a way of keeping the configuration elegant for the price of internal validations to be a bit more complex.

enhancements/network/on-prem-dual-stack-vips.md
cybertron: I'm not sure how it works, I just know that it did work when we added v6 vips with https://github.com/yboaron/kepalived-v6

enhancements/network/on-prem-dual-stack-vips.md
cybertron: Dan mentioned the same thing in his comments on a previous revision and if we were doing a clean sheet design I'd absolutely prefer this. With the existing config though, I think there's also an elegance to not having multiple fields that configure the same thing. Yes, we can handle that, but is this enough better to justify the complication of deprecating the old field and complicating the backend implementation? I don't think so, but I don't have strong feelings about this. Most of the options presented here are perfectly valid (except the one Kiran noted above) so I'm open to the alternatives if others feel strongly about it.

enhancements/network/on-prem-dual-stack-vips.md
cybertron: This was the frontrunner initially, but Dan pointed out that if we're going to deprecate the old fields it would be more consistent with the other network settings to make the new fields lists (as in the next alternative).

That said, I'm not opposed to any of these options (except the one you noted above), so if the installer team prefers this I'm not inclined to argue too strenuously. :-)

enhancements/network/on-prem-dual-stack-vips.md
cybertron: Is there a meaningful difference between "spoke" and "application"?

cybertron (commented):
Okay, I think the diversity of opinions here shows how most of the alternatives would work fine. Each option has it benefits and drawbacks, so I'm not dying on any of these hills. However, we do need to pick one that everyone finds acceptable so we can progress with this work.

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+
+```yaml
+platform:
+  baremetal:
+    apiVIP: 192.168.111.5
+    # Optional, and only for dual stack clusters
+    apiVIPSecondary: fd2e:6f44:5dd8:c956::5
+    ingressVIP: 192.168.111.4
+    # Optional, and only for dual stack clusters
+    ingressVIPSecondary: fd2e:6f44:5dd8:c956::4
```
sadasu, line 128: Can we call these new configs apiVIPv6 and ingressVIPv6 instead of calling them "Secondary" ? 

If it is necessary to mark one of them as primary & the other secondary, we should have another optional config like primaryApiVIP with a default value of apiVIP. Assuming the v6 VIPs as secondary seems arbitrary.

enhancements/network/on-prem-dual-stack-vips.md
romfreiman: Different cluster. Different user story. But not siginificant

enhancements/network/on-prem-dual-stack-vips.md
romfreiman: The main goal of this enhancement is to cover hub/spoke scenario AFAIK. Im not saying it should be covered by this enhancement, but it should be considered.

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+
+Implement this somewhat like the
+[existing workaround](https://github.com/yboaron/kepalived-v6) - create a
+separate instance of keepalived that is responsible for ipv6 only. This
+results in some duplication of containers and configs, but it would probably
+simplify the code changes.
+
+### Alternative install-config layouts
+
+#### Just add v6 versions of the VIP fields
```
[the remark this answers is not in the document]
sadasu: This alternative seems like it is the most unambiguous to the user. 

While describing the preferred solution, you mentioned that it would be beneficial to mark one address as the primary (and the other as secondary). So, in that case, for this alternative, would we also need another optional config item to point to the promaryVIP? For example, primaryApiVIP could have the values apiVIP or apiVIPv6. If unspecified it defaults to apiVIP.

enhancements/network/on-prem-dual-stack-vips.md
sadasu: Most unambiguous. Again we need a way for the user to indicate what the primary VIP is if that is even required.

sadasu:
Thanks for the really nice explanation around each of the options and their pro and cons. My vote goes to the use of plurals or list in the install config and api because it appears to be future proof (in case there is ever a case where we would need more than 1 v4 and 1 v6 address).

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+this case we'd need some logic to determine whether the VIPs in the list are
+v4 or v6, but I expect that would only be a little more complicated than the
+other options. This has the advantage of being extensible if we ever wanted
+to add more addresses, but I admit to having a hard time coming up with a use
+case where more than an ipv4 and ipv6 address would be needed.
+
+The list format could also help with processing of the VIPs. Instead of two
+distinct v4 and v6 sections for each VIP, we could loop over the list and
+use essentially the same configuration for both.
+
```
sadasu, line 352: The preferred solution points to the need to have a primary VIP and how that becomes easy with that solution. This solution with plurals/list has a similar advantage if we instruct the user to list the primary address as the 1st in the list.

enhancements/network/on-prem-dual-stack-vips.md
aojea: ah, cool, it was mostly curiosity

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+    - 192.168.111.4
+    - f00::4
+```
+This is essentially a variation on the previous option where the two separate
+fields per VIP are combined into a list field with a length of up to 2. In
+this case we'd need some logic to determine whether the VIPs in the list are
+v4 or v6, but I expect that would only be a little more complicated than the
+other options. This has the advantage of being extensible if we ever wanted
+to add more addresses, but I admit to having a hard time coming up with a use
+case where more than an ipv4 and ipv6 address would be needed.
```
mhrivnak, line 347: I could imagine an on-prem scenario where there is both an application network plus an independent management network. Since we blur the lines of the k8s API, using it both for infrastructure management and application management, it would be necessary to have the API service listening on both networks.

Likewise with RHACM you could imagine a hub-and-spoke model where the hub (management) cluster has special access to multiple otherwise-separate networks so it can manage infrastructure in multiple locations / facilities / regions etc. It would benefit from having VIPs on those multiple networks.

I don't know that there's been any direct demand for the above patterns, but they seem plausible.

mhrivnak (changes requested):
Since this will require similar API changes in assisted-service, whose API is an extension of hive's API and is a supported method for installing openshift, the assisted-service team would like to decide together what the API change should look like in both places. I think it makes sense to consider the assisted-service changes as an integral part of this enhancement.

sadasu:
I would like to propose an alternative that actually combines 2 proposals in the enhancement. Keep the existing apiVIP in the install_config and add **additionalApiVIPs** as a list. That has all the advantages listed for preferred solution with the additional advantage of easy extensibility.

 ```yaml
platform:
  baremetal:
    apiVIP: 192.168.111.5
    # Optional
    additionalApiVIPs:
    - f00::5
```

Changes to the Infrastructure API are again a combination what has already been detailed in the 2 sections.

**Review by cybertron — COMMENTED** [review]
Oh, my inline comments didn't get posted. Here you go!

File: enhancements/network/on-prem-dual-stack-vips.md [hunk 18]
```diff
@@ -0,0 +1,421 @@
...
+
+## Motivation
+
+Customer applications on dual stack clusters may only have ipv6 connectivity,
+but the cluster VIPs will be ipv4 (at this time ipv6 VIPs are not supported
+in dual stack clusters).
+
+### Goals
+
+Create both ipv4 and ipv6 VIPs on dual stack clusters so single stack
```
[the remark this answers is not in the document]
mhrivnak: I see. So this is referring to a single-stack application running outside the cluster that needs to interact with the api or ingress? I read it as a single-stack application running in the cluster.

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+`platformStatus.apiServerInternalIPv4` and if it is v6 we populate
+`apiServerInternalIPv6` instead. That way the only place we need to worry about
+the difference is in the installer. We would need a validation to ensure a v6
+VIP isn't specified in both fields.
+
+Note that this assumes we replace the existing api fields with two new ones
+per VIP and do not reuse the existing ones at all. I think we're going to want
+to do that regardless of what install-config layout we use so we don't need
+logic on the backend in machine-config-operator and baremetal-runtimecfg to
+handle older cluster configs.
```
staebler, line 304: There should not be any logic needed in machine-config-operator or baremetal-runtimecfg to handle older cluster configs. The installer would need to populate the new fields with the values from the deprecated fields. The backend should be able to rely on the new fields always being used.

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+```yaml
+platform:
+  baremetal:
+    apiVIP: 192.168.111.5
+    apiVIPv6: f00::5
+    ingressVIP: 192.168.111.4
+    ingressVIPv6: f00::4
+```
+This is a minimal change, but it leaves us with somewhat of a problem in that
+we can't assume the old fields always contain a v4 address since previous
```
staebler, line 291: This is not a problem, as far as I can tell.

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+releases allowed either v4 or v6. We may be able to deal with that by
+detecting the appropriate ip version for the old fields and populating new
+fields on the api side. E.g. if apiVIP is v4, we populate
+`platformStatus.apiServerInternalIPv4` and if it is v6 we populate
+`apiServerInternalIPv6` instead. That way the only place we need to worry about
+the difference is in the installer. We would need a validation to ensure a v6
+VIP isn't specified in both fields.
```
staebler, line 298: That validation, or some variant of it, is needed regardless of which option is selected.

enhancements/network/on-prem-dual-stack-vips.md
staebler: This alternative is my preference. It keeps it consistent with other fields. Deprecation is not terribly complex to achieve. As discussed above, the backend implementation should not be effected by deprecated fields.

enhancements/network/on-prem-dual-stack-vips.md
cgwalters: MCO doesn't parse the install config for day 2 operations.

cybertron: This might be a terminology problem. When I say "cluster config", I'm referring to the Infrastructure and ControllerConfig objects that are populated with the values specified in install-config. I will be more explicit in the next revision.

The main thing is that if we deprecate a field in install-config, that will require deprecation of an API field and migration of the values on the backend as well. We can do that, but it is a small drawback to some of the options.

staebler: OK. It was confusing to include cluster config then in the discussion of the merits of alternative designs for the install config. I view the layout of the install-config fields and the layout of the config.openshift.io API fields as separate concerns. If there is nothing using the install-config in-cluster, then the backend arguments has even less bearing as criteria for which install-config alternative to select.

cybertron:
I've pushed a new revision that changes the implementation to be based on a list of VIPs and addresses some other comments.

kikisdeliveryservice:
Since last review, I don't think anything has changed for us on the MCO end, later revisions seem fine, so still lgtm

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -0,0 +1,421 @@
...
+infrastructure object.
+
+machine-config-operator - The on-prem loadbalancer templates and manifests
+will need to be updated to support the additional VIPs.
+
+baremetal-runtimecfg - There is logic that depends on the ip version of the
+VIP. If there are multiple VIPs, this will have to be modified to account for
+that. In addition, the render code will need the new values wired in.
+
+assisted-service - the hive API extension is used for agent-based cluster creation as a stand-alone tool ("Infrastructure Operator") and as part of RHACM. It will need new API fields that correspond to the changes in the install-config.
```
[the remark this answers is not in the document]
cybertron: Given that we will need to have backward compatibility code in the installer and MCO to handle existing install-configs and clusters, is this a blocker?

I ask because this is already at high risk of missing 4.11 and we're not likely to be able to absorb a significant amount of additional work. If we could land the underlying feature in 4.11 and wire it in to assisted in 4.12 that would help a lot.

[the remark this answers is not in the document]
mkowalski: No no, this is not a blocker.

Also, all the work I described above will be handled by the Assisted Installer team, so on your side no additional work would be needed. I just wanted to make sure that changes required in Assisted are written down somewhere at the moment of accepting this change of the structure.

Assisted Installer releases with ACM and not with OCP, so our freezes are usually around a month later than OCPs. So if you deliver for 4.11, Assisted will catch-up around ACM 2.6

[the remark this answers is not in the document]
cybertron: Excellent, thank you! I will get this added.

cybertron:
Added implementation details for assisted-service based on the earlier discussion. I think this should be ready to go now.

cybertron:
@danwinship Since you're an approver in this repo could you also do that? We have agreement on the design from all of the affected teams now. Thanks.

PR #1051, opened 2022-03-07
Update enhancement proposal to support update of user-defined AWS tags

TrilokGeer:
This PR addresses design considerations to support update of user-defined AWS tags.
Also, details are listed for graduation criteria for tech preview and GA.

### DIFF
--- a/enhancements/api-review/custom-tags-aws.md
+++ b/enhancements/api-review/custom-tags-aws.md
@@ -9,11 +9,12 @@ reviewers:
   - @sinnykumari
   - @dmage
   - @staebler
+  - @tkashem
   - @tjungblu
 approvers:
   - @tjungblu
 creation-date: 2021-03-24
-last-updated: 2022-02-10
+last-updated: 2022-03-10
 status: implementable
 ---
 
@@ -75,6 +76,7 @@ Motivations include but are not limited to:
   2. Image registry.
   3. Ingress LB.
   4. IAM credentials by CCO in mint mode of operation.
+- Openshift is bound by the AWS resource that allows the fewest number of tags. An AWS S3 bucket can only have at most 10 tags.
 
 ## Proposal
 
@@ -119,28 +121,41 @@ In the future we will want to introduce a spec field where a customer can specif
   `.spec.platformSpec.aws` is a mutable field and `.status.platformStatus.aws` is immutable.
   `.status.platformStatus.aws` will have older version tags defined and is required for upgrade case.
 
-- All operators that create AWS resources (ingress, cloud credential, storage, image registry, machine-api) will apply these tags to all AWS resources they create.\
-  Operator must update AWS resource to match requested user-defined tags provided in `.spec.platformSpec.aws` (or `.status.platformStatus.aws`).\
+- All operators that create AWS resources (ingress, cloud credential, storage, image registry, machine-api) will apply these tags to all AWS resources they create.
+  Operator must update AWS resource to match requested user-defined tags provided in `.spec.platformSpec.aws` (or `.status.platformStatus.aws`).
+
   Operator must consider `.status.platformStatus.aws` to support upgrade scenarios.
 
 #### Create tags scenarios
 `resourceTags` that are specified in  `.spec.platformSpec.aws` of the Infrastructure resource will merge with user-defined tags in an individual component.
 
 In the case where a user-defined tag is specified in the Infrastructure resource and
-1) There is already a tag with the same key present for AWS resource, the value from the AWS resource will be replaced.\
-   For example,\
-   Existing tag for AWS resource = `key_infra1 = value_comp1`\
-   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_infra1`\
-   Action = Existing tag for AWS resource is updated to reflect new value.\
-   Final tag set to AWS resource = `key_infra1 = value_infra1`\
+1) There is already a tag with the same key present for AWS resource, the value from the AWS resource will be replaced.
+
+   For example,
+
+   Existing tag for AWS resource = `key_infra1 = value_comp1`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_infra1`
+
+   Action = Existing tag for AWS resource is updated to reflect new value.
+
+   Final tag set to AWS resource = `key_infra1 = value_infra1`
+
    Event action = An event is generated to notify user about the action status (success/failure) to update tags for the AWS resource.
 
-2) There is no tag with same key present, new tag is created for the AWS resource. In case of limit reached, a validation error is generated.\
-   For example,\
-   Existing tag for AWS resource = `key_infra1 = value_comp1`\
-   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_infra1`\
-   Action = A new tag is created for AWS resource.\
-   Final tag set to AWS resource = `key_infra1 = value_infra1`\
+2) There is no tag with same key present, new tag is created for the AWS resource. In case of limit reached, a validation error is generated.
+
+   For example,
+
+   Existing tag for AWS resource = `key_infra1 = value_comp1`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_infra1`
+
+   Action = A new tag is created for AWS resource.
+
+   Final tag set to AWS resource = `key_infra1 = value_infra1`
+
    Event action = An event is generated to notify user about the action status (success/failure) to create tags for the AWS resource.
 
 #### Update tags scenarios
@@ -148,51 +163,79 @@ Users can update the user-defined tags by editing `.spec.platformSpec.aws` of th
 On update of `resourceTags`, AWS resource is not created or restarted.
 
 In the case where a user-defined tag is specified in the Infrastructure resource and
-1) There is already a tag with the same key and value present for AWS resource, the tag for the AWS resource will not be updated.\
-   For example,\
-   Existing tag for AWS resource = `key_infra1 = value_update1`\
-   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_update1`\
-   Action = There is no update for AWS resource.\
-   Final tag set to AWS resource = `key_infra1 = value_update1`\
+1) There is already a tag with the same key and value present for AWS resource, the tag for the AWS resource will not be updated.
+
+   For example,
+
+   Existing tag for AWS resource = `key_infra1 = value_update1`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_update1`
+
+   Action = There is no update for AWS resource.
+
+   Final tag set to AWS resource = `key_infra1 = value_update1`
+
    Event action = An event is generated to notify user about the action status (success/failure) to update tags for the AWS resource.
 
-2) There is already a tag with the same key and different value present for AWS resource, the user-define tag value for the AWS resource will be updated.\
-   For example,\
-   Existing tag for AWS resource = `key_infra1 = value_old`\
-   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_update1`\
-   Action = Existing tag for AWS resource is updated to reflect new value.\
-   Final tag set to AWS resource = `key_infra1 = value_update1`\
+2) There is already a tag with the same key and different value present for AWS resource, the user-define tag value for the AWS resource will be updated.
+
+   For example,
+
+   Existing tag for AWS resource = `key_infra1 = value_old`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_update1`
+
+   Action = Existing tag for AWS resource is updated to reflect new value.
+
+   Final tag set to AWS resource = `key_infra1 = value_update1`
+
    Event action = An event is generated to notify user about the action status (success/failure) to update tags for the AWS resource.
 
 #### Delete tags scenarios
 Tags are deleted when the user sets the user-defined tag value to an empty string in `.spec.platformSpec.aws.resourceTags`.
-Also refer to `Precedence` scenario to understand cases where deletion of user-defined tags is not allowed .\
-For example,\
-Existing tag for AWS resource = `key_infra1 = value_old`\
-New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 =`\
-Action = Existing tag for AWS resource is deleted.\
-Final tag set to AWS resource = deleted\
+Also refer to `Precedence` scenario to understand cases where deletion of user-defined tags is not allowed .
+
+For example,
+
+Existing tag for AWS resource = `key_infra1 = value_old`
+
+New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 =`
+
+Action = Existing tag for AWS resource is deleted.
+
+Final tag set to AWS resource = deleted
+
 Event action = An event is generated to notify user about the action status (success/failure) to delete tags for the AWS resource.
 
 #### Precedence scenarios
 1) User-defined tags on kube resources MUST continue to take precedence during create and update. User-defined tags found on kube resources must not be deleted by methods described in `Delete tags scenarios`.
    A warning is generated to inform the user about action not being applied on the list of user tags.
 
-   For example,\
-   Existing tag in kube resource object like machine CR = `key_infra1 = custom_value`\
-   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value1`\
-   Action = The tag for AWS resource is maintained to `key_infra1 = custom_value`.\
-   Final tag set to AWS resource = `key_infra1 = custom_value`\
+   For example,
+
+   Existing tag in kube resource object like machine CR = `key_infra1 = custom_value`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value1`
+
+   Action = The tag for AWS resource is maintained to `key_infra1 = custom_value`.
+
+   Final tag set to AWS resource = `key_infra1 = custom_value`
+
    Event action = An event is generated to notify user about the action status (failure) to update tags for the AWS resource. A warning also must be generated about list of user-defined tags on which action is not applicable.
 
 2) When a new user-defined tag is added on kube resources, it MUST override the tag in `.spec.platformSpec.aws`. User-defined tags found on kube resources must not be deleted by methods described in `Delete tags scenarios`.
    A warning is generated to inform the user about action not being applied on the list of user tags.
 
-   For example,\
-   Existing tag in Infrastructure CR = `key_infra1 = custom_value`\
-   New tag added in kube resource object like machine CR = `key_infra1 = value1`\
-   Action = The tag for AWS resource is maintained to `key_infra1 = value1`.\
-   Final tag set to AWS resource = `key_infra1 = custom_value`\
+   For example,
+
+   Existing tag in Infrastructure CR = `key_infra1 = custom_value`
+
+   New tag added in kube resource object like machine CR = `key_infra1 = value1`
+
+   Action = The tag for AWS resource is maintained to `key_infra1 = value1`.
+
+   Final tag set to AWS resource = `key_infra1 = custom_value`
+
    Event action = An event is generated to notify user about the action status (success) to update tags for the AWS resource. A warning also must be generated about list of user-defined tags on which action did a override.
 
 3) `.spec.platformSpec.aws` take precedence over `.status.platformStatus.aws`. User-defined tags must be merged from `.status.platformStatus.aws` and set to AWS resource.
@@ -205,34 +248,48 @@ Event action = An event is generated to notify user about the action status (suc
    The resource's tag will be reconciled by its owning operator to the value from `.spec.platformSpec.aws.resourceTags`.
    The user-defined tag will be overwritten with value from `.spec.platformSpec.aws.resourceTags` when there is an update to `.spec.platformSpec.aws.resourceTags`.
 
-   For example,\
-   Edited existing tag using external tool for AWS resource = `key_infra1 = value_comp1`\
-   Previous tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_infra1`\
-   Action = Update existing tag with value from `.spec.platformSpec.aws.resourceTags`.\
-   Final tag set to AWS resource = `key_infra1 = value_infra1`\
+   For example,
+
+   Edited existing tag using external tool for AWS resource = `key_infra1 = value_comp1`
+
+   Previous tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_infra1`
+
+   Action = Update existing tag with value from `.spec.platformSpec.aws.resourceTags`.
+
+   Final tag set to AWS resource = `key_infra1 = value_infra1`
+
    Event action = An event is generated to notify user about the action status (success/failure) to update tags for the AWS resource.
 
 2) User deletes the user-defined tag from `.spec.platformSpec.aws.resourceTags`
-   The user-defined tag which is removed from spec, will not be reconciled or managed by operators.\
+   The user-defined tag which is removed from spec, will not be reconciled or managed by operators.
+
    User can update user-defined tag key:value using external tool. The user-defined tag will not be overwritten.
 
-   For example,\
-   Existing tag for AWS resource = `key_infra1 = value1`\
-   New tag request = `.spec.platformSpec.aws.resourceTags` has no user-defined tag with key `key_infra1`\
-   Action = No change in tags for the AWS resource.\
-   Final tag set to AWS resource = `key_infra1 = value1`\
+   For example,
+
+   Existing tag for AWS resource = `key_infra1 = value1`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has no user-defined tag with key `key_infra1`
+
+   Action = No change in tags for the AWS resource.
+
+   Final tag set to AWS resource = `key_infra1 = value1`
+
    Event action = No event
 
-3) User sets user-defined tag created using external tools to empty string for deleting for AWS resource.\
+3) User sets user-defined tag created using external tools to empty string for deleting for AWS resource.
+
    There is no validation check involved for creator tool of user-defined tag. Any user-defined tag listed by user in `.spec.platformSpec.aws.resourceTags` is considered for create/update/delete accordingly.
 
-4) Any user-defined tag set using `.spec.platformSpec.aws.resourceTags` in `Infrastructure.config.openshift.io/v1` type has scope limited to cluster-level.
+4) Any user-defined tag set using `.spec.platformSpec.aws.resourceTags` in `Infrastructure.config.openshift.io/v1` type will affect all AWS resources managed.
 
 5) User-defined tags are not synced from `.spec.platformSpec.aws.resourceTags` to `.status.platformStatus.aws.resourceTags` for the following reasons.
+
 - User-defined tags in `.spec.platformSpec.aws.resourceTags` can be "removed without delete", "delete", "update".
   `.spec.platformSpec.aws.resourceTags` to `.status.platformStatus.aws.resourceTags` sync is not required,
   as there will be no versioning of the user-defined tags required to override the user-defined tags in Infrastructure CR.
   Instead, the user-defined tags (if supported in resource operator spec field, e.g: machine) can override the user-defined tags in Infrastructure CR.
+
 - As `.spec.platformSpec.aws.resourceTags` has the actual values. Sync to `.status.platformStatus.aws.resourceTags` was required in earlier design to identify if the
   tag on the resource was created using Infrastructure CR or external tool. Identifying the creator tool was done to restrict user from editing the same user-tag kv pair using multiple tools.
   This inherently poses many scenarios of conflict which will result in user-defined tag kv pair being inconsistent across cluster when applied using Infrastructure CR. Hence, user will be confused which tool to be used to update tag.
@@ -248,65 +305,11 @@ Event action = An event is generated to notify user about the action status (suc
 The developer must also carefully consider the upgrade and rollback ordering
 between the webhook and CRD.
 
-The Infrastructure resource example to use spec for api changes
-
-```go
-// AWSPlatformSpec holds the desired state of the Amazon Web Services Infrastructure provider.
-// This only includes fields that can be modified in the cluster.
-type AWSPlatformSpec struct {
-    // Existing fields
-    ...
-    // ResourceTags is a list of additional tags to apply to AWS resources created for the cluster.
-    // See https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html for information on tagging AWS resources.
-    // AWS supports a maximum of 10 tags per resource. OpenShift reserves 5 tags for its use, leaving 5 tags
-    // available for the user.
-    // ResourceTags field is mutable and items can be removed.
-    // +kubebuilder:validation:MaxItems=10
-    // +optional
-    ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
-}
-
-```
-
-```yaml
-spec:
-description: AWS contains settings specific to the Amazon Web Services Infrastructure provider.
-type: object
-properties:
-    resourceTags:
-    description: ResourceTags is a list of additional tags to apply to AWS resources created for the cluster. See https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html for information on tagging AWS resources.
-    type: array
-    maxItems: 10
-    items:
-        description: AWSResourceTag is a tag to apply to AWS resources created for the cluster.
-        type: object
-        required:
-            - key
-            - value
-        properties:
-            key:
-                description: key is the key of the tag
-                type: string
-                maxLength: 128
-                minLength: 1
-                pattern: ^[0-9A-Za-z_.:/=+-@]+$
-            value:
-                description: value is the value of the tag. Some AWS service do not support empty values. Since tags are added to resources in many services, the length of the tag value must meet the requirements of all services.
-                type: string
-                maxLength: 256
-                minLength: 1
-                pattern: ^[0-9A-Za-z_.:/=+-@]+$
-```
-
 ### User Stories
 
 - As a security-conscious ROSA customer, I want to restrict the permissions granted to Red Hat in my AWS account by using AWS resource tags.
   Red Hat applies one well-known tag to all ROSA cluster resources to identify Red Hat-managed AWS resources (e.g. “red-hat-managed=true”).
   Customer can create AWS policies that restrict Red Hat access to tagged resource types based on tag (e.g. “RedHat role can only create/read/update/delete S3 buckets with a red-hat-managed=true tag”).
-
-
-- As a cluster administrator of OpenShift, I would like to be able to add a maximum of 10 user-defined tags for AWS resources provisioned by an OCP cluster.
-  Openshift is bound by the AWS resource that allows the fewest number of tags. An AWS S3 bucket can only have at most 10 tags.
 
 - As a cluster administrator of OpenShift, I would like to be able to delete user-defined tags managed by OpenShift for AWS resources.
   As there are limitations on the maximum number of user-defined tags that can be set on an AWS resource, the ability to delete user-defined tags enables user to replace and manage tags within limits allowed.
@@ -332,6 +335,57 @@ As the existing cluster might be configured with user-defined tags from kube res
 ### API Extensions
 
 This proposal edits `Infrastructure.config.openshift.io/v1` type
+
+The Infrastructure resource example to use spec for api changes
+
+```go
+// AWSPlatformSpec holds the desired state of the Amazon Web Services Infrastructure provider.
+// This only includes fields that can be modified in the cluster.
+type AWSPlatformSpec struct {
+    // Existing fields
+    ...
+    // ResourceTags is a list of additional tags to apply to AWS resources created for the cluster.
+    // See https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html for information on tagging AWS resources.
+    // AWS supports a maximum of 10 tags per resource. OpenShift reserves 5 tags for its use, leaving 5 tags
+    // available for the user.
+    // ResourceTags field is mutable and items can be removed.
+    // +kubebuilder:validation:MaxItems=10
+    // +optional
+    ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
+}
+
+```
+
+```yaml
+spec:
+    description: AWS contains settings specific to the Amazon Web Services Infrastructure provider.
+    type: object
+    properties:
+    resourceTags:
+        description: ResourceTags is a list of additional tags to apply to AWS resources created for the cluster. See https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html for information on tagging AWS resources.
+        type: array
+        maxItems: 10
+        items:
+            description: AWSResourceTag is a tag to apply to AWS resources created for the cluster.
+            type: object
+            required:
+                - key
+                - value
+                properties:
+                    key:
+                        description: key is the key of the tag
+                        type: string
+                        maxLength: 128
+                        minLength: 1
+                        pattern: ^[0-9A-Za-z_.:/=+-@]+$
+                    value:
+                        description: value is the value of the tag. Some AWS service do not support empty values. Since tags are added to resources in many services, the length of the tag value must meet the requirements of all services.
+                        type: string
+                        maxLength: 256
+                        minLength: 1
+                        pattern: ^[0-9A-Za-z_.:/=+-@]+$
+```
+
 
 ### Operational Aspects of API Extensions
 
@@ -415,7 +469,5 @@ The operator would need to monitor multiple kube resources, infrastructure CR an
 
 ## Future Work
 
-1. Update the Infrastructture CRD to enable validation  to check if any tag that starts with `kubernetes.io` or `openshift.io using
+1. Update the Infrastructture CRD to enable validation  to check if any tag that starts with `kubernetes.io` or `openshift.io` using
    CRD Validation Expression Language proposed here : https://github.com/kubernetes/enhancements/pull/2877/files.
-
-
--- a/enhancements/api-review/custom-tags-aws.md
+++ b/enhancements/api-review/custom-tags-aws.md
@@ -405,7 +405,7 @@ NA
 User can add tags during installation for creation of resources with tags. Installer creates infrastructure manifests with `resourceTags` to `.spec.platformSpec.aws` of the `Infrastructure.config.openshift.io`.
 Operators that operate on aws resources must consider the tags from `.spec.platformSpec.aws`, `.status.platformStatus.aws` and existing AWS resource tags.
 
-The values from `status.platformStatus.aws` will used to only support older version.
+The values from `status.platformStatus.aws` will be used to only support older version.
 
 `.status.platformStatus.aws` will be deprecated in the future versions.
 
--- a/enhancements/api-review/custom-tags-aws.md
+++ b/enhancements/api-review/custom-tags-aws.md
@@ -330,7 +330,7 @@ type AWSResourceTagSpec struct {
     // +kubebuilder:validation:Pattern=`^(?!openshift.io)(?!kubernetes.io)([0-9A-Za-z_.:\/=+-@]+$)`
     // +required
     Key string `json:"key"`
-    // value is the value of the AWS tag. The value must consist only of alphabets, numbers and special characters(_.:\/=+-@).
+    // value is the value of the AWS tag. The value must consist only of alphanumeric and certain special characters(_.:\/=+-@).
     // value must be no more than 256 characters in length.
     // Some AWS service do not support empty values. Since tags are added to resources in many services, the
     // length of the tag value must meet the requirements of all services.
--- a/enhancements/api-review/custom-tags-aws.md
+++ b/enhancements/api-review/custom-tags-aws.md
@@ -321,8 +321,8 @@ type AWSPlatformSpec struct {
 // Existing type
 // AWSResourceTagSpec is a tag to apply to AWS resources created for the cluster.
 type AWSResourceTagSpec struct {
-    // key is the key of the AWS tag. The value must consist only of alphabets, numbers and special characters(_.:\/=+-@).
-    // key should not consist "openshift.io" and "kubernetes.io" which are reserved.
+    // key is the key of the AWS tag. The value must consist only of alphanumeric and certain special characters(_.:\/=+-@).
+    // key should not start with "openshift.io" or "kubernetes.io". Both of these prefixes are reserved for use by the platform.
     // key must be no more than 128 characters in length.
     // +kubebuilder:validation:Required
     // +kubebuilder:validation:MinLength=1
--- a/enhancements/api-review/custom-tags-aws.md
+++ b/enhancements/api-review/custom-tags-aws.md
@@ -321,14 +321,17 @@ type AWSPlatformSpec struct {
 // Existing type
 // AWSResourceTagSpec is a tag to apply to AWS resources created for the cluster.
 type AWSResourceTagSpec struct {
-    // key is the key of the tag
+    // key is the key of the AWS tag. The value must consist only of alphabets, numbers and special characters(_.:\/=+-@).
+    // key should not consist "openshift.io" and "kubernetes.io" which are reserved.
+    // key must be no more than 128 characters in length.
     // +kubebuilder:validation:Required
     // +kubebuilder:validation:MinLength=1
     // +kubebuilder:validation:MaxLength=128
     // +kubebuilder:validation:Pattern=`^(?!openshift.io)(?!kubernetes.io)([0-9A-Za-z_.:\/=+-@]+$)`
     // +required
     Key string `json:"key"`
-    // value is the value of the tag.
+    // value is the value of the AWS tag. The value must consist only of alphabets, numbers and special characters(_.:\/=+-@).
+    // value must be no more than 256 characters in length.
     // Some AWS service do not support empty values. Since tags are added to resources in many services, the
     // length of the tag value must meet the requirements of all services.
     // +kubebuilder:validation:Required
--- a/enhancements/api-review/custom-tags-aws.md
+++ b/enhancements/api-review/custom-tags-aws.md
@@ -231,7 +231,7 @@ Event action = An event is generated to notify user about the action status (suc
 
    New tag added in kube resource object like machine CR = `key_infra1 = value1`
 
-   Action = The tag for AWS resource is maintained to `key_infra1 = value1`.
+   Action = The tag for AWS resource is updated to `key_infra1 = value1`.
 
    Final tag set to AWS resource = `key_infra1 = custom_value`
 
--- a/enhancements/api-review/custom-tags-aws.md
+++ b/enhancements/api-review/custom-tags-aws.md
@@ -2,13 +2,18 @@
 title: apply-user-defined-tags-to-all-aws-resources-created-by-openshift
 authors:
   - "@gregsheremeta"
+  - "@tgeer"
 reviewers:
-  - @decarr
-  - @bparees
+  - @joelspeed
+  - @Miciah
+  - @sinnykumari
+  - @dmage
+  - @staebler
+  - @tjungblu
 approvers:
-  - @decarr 
+  - @tjungblu
 creation-date: 2021-03-24
-last-updated: 2021-04-09
+last-updated: 2022-02-10
 status: implementable
 ---
 
@@ -16,11 +21,11 @@ status: implementable
 
 ## Release Signoff Checklist
 
-- [ ] Enhancement is `implementable`
-- [ ] Design details are appropriately documented from clear requirements
-- [ ] Test plan is defined
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
 - [ ] Operational readiness criteria is defined
-- [ ] Graduation criteria for dev preview, tech preview, GA
+- [x] Graduation criteria for dev preview, tech preview, GA
 - [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
 
 ## Summary
@@ -31,34 +36,49 @@ have the ability to apply user defined tags to many resources created by OpenShi
 Note: this enhancement is slightly retroactive. Work has already begun on this. See
 - https://github.com/openshift/api/pull/864
 - https://github.com/openshift/cluster-ingress-operator/pull/578
+- https://github.com/openshift/api/pull/1064/
 
 ## Motivation
 
 Motivations include but are not limited to:
 
-- allow admin, compliance, and security teams to keep track of assets and objects created by OpenShift,
+- Allow admin, compliance, and security teams to keep track of assets and objects created by OpenShift,
    both at install time and during continuous operation (Day 2)
-- in a Managed OpenShift environment such as Red Hat OpenShift on AWS (ROSA), allow easy differentiation
+- In a Managed OpenShift environment such as Red Hat OpenShift on AWS (ROSA), allow easy differentiation
    between Red Hat-managed AWS objects and customer-managed objects
-- *allow for the restriction of permissions granted to Red Hat in an AWS account by AWS resource tags.
+- *Allow for the restriction of permissions granted to Red Hat in an AWS account by AWS resource tags.
    For example, see https://issues.redhat.com/browse/SDE-1146 - "IAM users and roles can only operate on resources with specific tags"*
 
 ### Goals
 
-- the administrator or service (in the case of Managed OpenShift) installing OpenShift can pass an arbitrary
+- The administrator or service (in the case of Managed OpenShift) installing OpenShift can pass an arbitrary
    list of user-defined tags to the OpenShift Installer, and everything created by the installer and all other
    bootstrapped components will apply those tags to all resources created in AWS, for the life of the cluster, and where supported by AWS.
-- tags must be applied at creation time, in an atomic operation. It isn't acceptable to create an object and,
-   after some period of time, apply the tags post-creation.
+- User-defined tags can be applied at creation time, in an atomic operation.
+- User-defined tags can be updated post creation.
 
 ### Non-Goals
 
-- to reduce initial scope, tags are applied only at creation time and not reconciled. If an administrator manually
-   changes the tags stored in the infrastructure resource, behavior is undefined. See below.
-- to reduce initial scope, we are not implementing this for clouds other than AWS. We will not take any actions
-   to prohibit that later.
+- To reduce initial scope, the proposal does not apply for clouds other than AWS. There will be no actions taken to prohibit that later.
+
+### Limitations
+
+- User-defined tags for PV volumes cannot be updated but only set during creation of volumes using `--extra-tags`. Present CSI spec does not support `UpdateVolume` to update volume metadata.
+   Restarting csi driver to initiate multiple `CreateVolume` will depend on underlying storage layer idempotency implementation for the operation. It is recommended to use an interface from CSI spec which enforces idempotency guarantees for volume metadata updates.
+   In case of AWS EC2, `CreateVolume` returns `IdempotentParameterMismatch` error when identical request is not found for the same client token.
+   Also, the CSI spec `https://github.com/container-storage-interface/spec/blob/master/spec.md` mentions grpc error code
+   `6 ALREADY_EXISTS` for `a volume corresponding to the specified volume name already exists but is incompatible with the specified capacity_range, volume_capabilities, parameters, accessibility_requirements or volume_content_source`.
+- User-defined tags cannot be updated for vpc, security groups, elb(not managed by in-cluster operators), route53, subnet resources as there is no operator managing the resources post installation.
+   User-defined tags cannot be updated to an AWS resource which is not managed by an operator. In this proposal, the changes proposed and developed will be part of openshift-* namespace. External operators are not in scope.
+- User-defined tags can be updated on the following AWS resources.
+  1. EC2 instances for master and worker nodes.
+  2. Image registry.
+  3. Ingress LB.
+  4. IAM credentials by CCO in mint mode of operation.
 
 ## Proposal
+
+### Existing design details
 
 New `experimentalPropagateUserTags` field added to `.platform.aws` of install config to indicate that the user tags should be applied to AWS
 resources created by in-cluster operators.
@@ -79,53 +99,7 @@ The userTags field is intended to be set at install time and is considered immut
 
 If the userTags field is changed post-install, there is no guarantee about how an in-cluster operator will respond to the change. Some operators may reconcile the change and change tags on the AWS resource. Some operators may ignore the change. However, if tags are removed from userTags, the tag will not be removed from the AWS resource.
 
-### User Stories
-
-https://issues.redhat.com/browse/SDE-1146 - IAM users and roles can only operate on resources with specific tags
-As a security-conscious ROSA customer, I want to restrict the permissions granted to Red Hat in my AWS account by using
-AWS resource tags.
-
-### Risks and Mitigations
-
-## Design Details
-
-### Test Plan
-
-TBD
-
-### Graduation Criteria
-
-TBD
-
-#### Dev Preview -> Tech Preview
-
-TBD
-
-#### Tech Preview -> GA
-
-TBD
-
-#### Removing a deprecated feature
-
-This enhancement does not remove or deprecate any features.
-
-### Upgrade / Downgrade Strategy
-
-On upgrade:
-
-The new status field won't be populated since it is only populated by the installer and that can't have happened if the cluster was installed from a prior version. Components that consume the new field should take no action since they will see no additional tags.
-
-On downgrade:
-
-The status field may remain populated, components may or may not continue to tag newly created resources w/ the additional tags depending on whether or not a given component still has logic to respect the status tags, after the downgrade.
-
-### Version Skew Strategy
-
-TBD
-
-## Implementation History
-
-## Drawbacks
+#### Existing drawbacks
 
 If a customer decides they do not want these tags applied to new resources, there is no clean way to address that need:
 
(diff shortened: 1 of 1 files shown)


TrilokGeer:
This PR is in continuation to update based on review suggestions in : https://github.com/openshift/enhancements/pull/1000

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +99,323 @@ The userTags field is intended to be set at install time and is considered immut
...
+The operator would reconcile on changes to the infrastructure CR to ensure that resources owned by the cluster have tags matching what is in the infrastructure CR.
+That would get tricky when it comes to resources owned by in-cluster operators, though, particularly with tags that may have different values in the corresponding kube resources.
+The operator would need to monitor multiple kube resources, infrastructure CR and multiple instances of different AWS resource types.
 
 ## Infrastructure Needed [optional]
 
-TBD
+## Future Work
+
+1. Update the Infrastructture CRD to enable validation  to check if any tag that starts with `kubernetes.io` or `openshift.io using
```
JoelSpeed, line 418: You've missed the closing backtick after openshift.io

JoelSpeed (commented):
This is looking a look better, thanks for the updates!

It definitely needs some input from the installer team before it merges, might also be worth giving a heads up to the other teams this affects, ingress, cloud credential, storage and image registry. I can approve on behalf of the Machine API team

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +99,323 @@ The userTags field is intended to be set at install time and is considered immut
...
-TBD
+- Upgrade/downgrade testing
+- Sufficient time for feedback
+- Available by default
+- Stress testing for scaling and tag update scenarios
 
 #### Removing a deprecated feature
 
-This enhancement does not remove or deprecate any features.
+`experimentalPropagateUserTags` field will be set for deprecation.
```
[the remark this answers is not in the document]
staebler: The "experimental" in the name of the field does not change the API guarantees in any way.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +99,323 @@ The userTags field is intended to be set at install time and is considered immut
...
+  The install validation will fail if there is any tag that starts with `kubernetes.io` or `openshift.io` in `userTags`. In the existing design, kubernetes.io and openshift.io namespaces are not allowed as part of user-defined tags.
+
+  The `resourceTags` field will be populated by the installer using the entries from `userTags` field.
+
+  `.spec.platformSpec.aws` is a mutable field and `.status.platformStatus.aws` is immutable.
+  `.status.platformStatus.aws` will have older version tags defined and is required for upgrade case.
+
+- All operators that create AWS resources (ingress, cloud credential, storage, image registry, machine-api) will apply these tags to all AWS resources they create.\
+  Operator must update AWS resource to match requested user-defined tags provided in `.spec.platformSpec.aws` (or `.status.platformStatus.aws`).\
+  Operator must consider `.status.platformStatus.aws` to support upgrade scenarios.
```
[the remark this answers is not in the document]
TrilokGeer: Removed, but now there are extra blanks.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +99,323 @@ The userTags field is intended to be set at install time and is considered immut
...
+   Existing tag for AWS resource = `key_infra1 = value1`\
+   New tag request = `.spec.platformSpec.aws.resourceTags` has no user-defined tag with key `key_infra1`\
+   Action = No change in tags for the AWS resource.\
+   Final tag set to AWS resource = `key_infra1 = value1`\
+   Event action = No event
+
+3) User sets user-defined tag created using external tools to empty string for deleting for AWS resource.\
+   There is no validation check involved for creator tool of user-defined tag. Any user-defined tag listed by user in `.spec.platformSpec.aws.resourceTags` is considered for create/update/delete accordingly.
+
+4) Any user-defined tag set using `.spec.platformSpec.aws.resourceTags` in `Infrastructure.config.openshift.io/v1` type has scope limited to cluster-level.
```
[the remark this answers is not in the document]
TrilokGeer: rephrased to make it less ambiguous

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +99,323 @@ The userTags field is intended to be set at install time and is considered immut
...
+// AWSPlatformSpec holds the desired state of the Amazon Web Services Infrastructure provider.
+// This only includes fields that can be modified in the cluster.
+type AWSPlatformSpec struct {
+    // Existing fields
+    ...
+    // ResourceTags is a list of additional tags to apply to AWS resources created for the cluster.
+    // See https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html for information on tagging AWS resources.
+    // AWS supports a maximum of 10 tags per resource. OpenShift reserves 5 tags for its use, leaving 5 tags
+    // available for the user.
+    // ResourceTags field is mutable and items can be removed.
```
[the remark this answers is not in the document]
TrilokGeer: As the behavior is better explained in enhancement, I think, it is better to add link to enhancement. Suggestion will be added in the development PR https://github.com/openshift/api/pull/1064/

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +99,323 @@ The userTags field is intended to be set at install time and is considered immut
...
-https://issues.redhat.com/browse/SDE-1146 - IAM users and roles can only operate on resources with specific tags
-As a security-conscious ROSA customer, I want to restrict the permissions granted to Red Hat in my AWS account by using
-AWS resource tags.
+- As a security-conscious ROSA customer, I want to restrict the permissions granted to Red Hat in my AWS account by using AWS resource tags.
+  Red Hat applies one well-known tag to all ROSA cluster resources to identify Red Hat-managed AWS resources (e.g. “red-hat-managed=true”).
+  Customer can create AWS policies that restrict Red Hat access to tagged resource types based on tag (e.g. “RedHat role can only create/read/update/delete S3 buckets with a red-hat-managed=true tag”).
+
+
+- As a cluster administrator of OpenShift, I would like to be able to add a maximum of 10 user-defined tags for AWS resources provisioned by an OCP cluster.
+  Openshift is bound by the AWS resource that allows the fewest number of tags. An AWS S3 bucket can only have at most 10 tags.
```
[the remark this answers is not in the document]
TrilokGeer: I think, it is better listed under proposal section.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +99,323 @@ The userTags field is intended to be set at install time and is considered immut
...
+As the existing cluster might be configured with user-defined tags from kube resources, the precedence is given to user-defined tags in kube resources.
+- As a cluster administrator of OpenShift, I expect user-defined tags added in kube resources override the entries in `.spec.platformSpec.aws.resourceTags`.
+
+- As a cluster administrator of OpenShift, I expect user-defined tags added in kube resources must not be overridden by update or delete actions.
+
+
+
+### API Extensions
+
+This proposal edits `Infrastructure.config.openshift.io/v1` type
```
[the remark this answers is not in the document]
TrilokGeer: Correct, updated the same.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Thanks for locating, updated the tick.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +99,323 @@ The userTags field is intended to be set at install time and is considered immut
...
+
+6) User applies user-defined tag using `.spec.platformSpec.aws.resourceTags`. Later, user modifies the user-defined tag for AWS resource using external tool.
+   In this case, the desired value is not set immediately for AWS resource by the owning operator. There is eventual consistency maintained by the owning operator.
+   The time taken to reconcile the modified user-defined tag on AWS resource to desired value vary across owning operators.
+
+7) Post-installation of the cluster, validation for kubernetes.io or openshift.io namespaced tags in the `.spec.platformSpec.aws.resourceTags` is not done by API server.
+   The individual AWS resource component operator fails to apply the user tags. Developing a production-ready webhook is a substantial task. To support webhook life-cycle management and monitoring, instrumentation, alerting and integration with the packaging/releases processes must be supported.
+   For validating fixed strings it is better to have an expression based check done by API server which is already part of upstream as alpha feature for CRD Validation Expression Language.
```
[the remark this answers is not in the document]
TrilokGeer: It is better to have validation check at api server to avoid stale entries in spec field and additional triggers for removing it.
As user tags can be updated without validation on AWS direclty, to raise a warning or overwrite, there needs to be a validation check at the operator.
Kube resource objects are not validated and allow namespace strings (`kubernetes.io` and `openshift.io`) to be used in user tags. Operators need to validate to avoid breaking change.

As the webhook implementation might become obsolete in next versions, allowing stale entries in spec field for now, does not allow invalid usage of namespace strings.

TrilokGeer:
cc @Miciah @sinnykumari @dmage @staebler @patrickdillon @tkashem , can you please give a read on behalf of respective components to help check for any component-specific caveats?

enhancements/api-review/custom-tags-aws.md
```diff
@@ -31,35 +37,51 @@ have the ability to apply user defined tags to many resources created by OpenShi
...
+   Also, the CSI spec `https://github.com/container-storage-interface/spec/blob/master/spec.md` mentions grpc error code
+   `6 ALREADY_EXISTS` for `a volume corresponding to the specified volume name already exists but is incompatible with the specified capacity_range, volume_capabilities, parameters, accessibility_requirements or volume_content_source`.
+- User-defined tags cannot be updated for vpc, security groups, elb(not managed by in-cluster operators), route53, subnet resources as there is no operator managing the resources post installation.
+   User-defined tags cannot be updated to an AWS resource which is not managed by an operator. In this proposal, the changes proposed and developed will be part of openshift-* namespace. External operators are not in scope.
+- User-defined tags can be updated on the following AWS resources.
+  1. EC2 instances for master and worker nodes.
+  2. Image registry.
+  3. Ingress LB.
+  4. IAM credentials by CCO in mint mode of operation.
+- Openshift is bound by the AWS resource that allows the fewest number of tags. An AWS S3 bucket can only have at most 10 tags.
```
staebler, line 79:
```suggestion
- OpenShift is bound by the AWS resource that allows the fewest number of tags. An AWS S3 bucket can have at most 10 tags.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+If a customer decides they do not want these tags applied to new resources, there is no clean way to address that need:
+
+1. they would have to get help from support to edit the status field
+2. they would have to manually remove the undesired tags from any existing resources
+
+In the future we will want to introduce a spec field where a customer can specify and edit these tags, which will be reflected into status when changed, and we will expect consuming components to reconcile changes to the set of tags by applying net new tags to existing resources (but they will still not remove tags that are dropped from the list of tags).
+
+### New proposal to support update of user-defined tags (or userTags)
+
+- `experimentalPropagateUserTags` field will be set for deprecation in this proposal version. The values in `experimentalPropagateUserTags` are ignored in this version.
```
staebler, line 115: We cannot simply ignore the field. If the field is set to "false", then we **must not** apply the user tags to resources created in-cluster.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+
+   Final tag set to AWS resource = `key_infra1 = value_infra1`
+
+   Event action = An event is generated to notify user about the action status (success/failure) to update tags for the AWS resource.
+
+2) There is no tag with same key present, new tag is created for the AWS resource. In case of limit reached, a validation error is generated.
+
+   For example,
+
+   Existing tag for AWS resource = `key_infra1 = value_comp1`
```
staebler, line 151: This case is for when there is no existing tag for the AWS resource.
```suggestion
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_update1`
+
+   Action = There is no update for AWS resource.
+
+   Final tag set to AWS resource = `key_infra1 = value_update1`
+
+   Event action = An event is generated to notify user about the action status (success/failure) to update tags for the AWS resource.
+
+2) There is already a tag with the same key and different value present for AWS resource, the user-define tag value for the AWS resource will be updated.
```
staebler, line 180:
```suggestion
2) There is already a tag with the same key and different value present for AWS resource, the AWS resource will be updated with the user-defined tag value.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_update1`
+
+   Action = Existing tag for AWS resource is updated to reflect new value.
+
+   Final tag set to AWS resource = `key_infra1 = value_update1`
+
+   Event action = An event is generated to notify user about the action status (success/failure) to update tags for the AWS resource.
+
+#### Delete tags scenarios
+Tags are deleted when the user sets the user-defined tag value to an empty string in `.spec.platformSpec.aws.resourceTags`.
```
staebler, line 195: Are we certain that there is no use case for a user wanting a tag with an empty value on a resource?

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 =`
+
+Action = Existing tag for AWS resource is deleted.
+
+Final tag set to AWS resource = deleted
+
+Event action = An event is generated to notify user about the action status (success/failure) to delete tags for the AWS resource.
+
+#### Precedence scenarios
+1) User-defined tags on kube resources MUST continue to take precedence during create and update. User-defined tags found on kube resources must not be deleted by methods described in `Delete tags scenarios`.
```
staebler, line 211: What are "user-defined tags on kube resources"? Is that the same as "user-defined tags in an individual component"? If so, let's be consistent with terminology.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+
+   Existing tag in kube resource object like machine CR = `key_infra1 = custom_value`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value1`
+
+   Action = The tag for AWS resource is maintained to `key_infra1 = custom_value`.
+
+   Final tag set to AWS resource = `key_infra1 = custom_value`
+
+   Event action = An event is generated to notify user about the action status (failure) to update tags for the AWS resource. A warning also must be generated about list of user-defined tags on which action is not applicable.
```
staebler, line 224: Why is the event a failure? I am on-board with there being an event, but I am having a hard time rationalizing that the event is a failure. This seems like a reasonable use case that is behaving successfully in the expected manner.  

Think about if the user updates the infrastructure resource and the component resource at the same time. How would a controller know to generate a failure event in that case instead of a success event?

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+
+   For example,
+
+   Existing tag in Infrastructure CR = `key_infra1 = custom_value`
+
+   New tag added in kube resource object like machine CR = `key_infra1 = value1`
+
+   Action = The tag for AWS resource is maintained to `key_infra1 = value1`.
+
+   Final tag set to AWS resource = `key_infra1 = custom_value`
```
staebler, line 237:
```suggestion
   Final tag set to AWS resource = `key_infra1 = value1`
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+2) When a new user-defined tag is added on kube resources, it MUST override the tag in `.spec.platformSpec.aws`. User-defined tags found on kube resources must not be deleted by methods described in `Delete tags scenarios`.
+   A warning is generated to inform the user about action not being applied on the list of user tags.
+
+   For example,
+
+   Existing tag in Infrastructure CR = `key_infra1 = custom_value`
+
+   New tag added in kube resource object like machine CR = `key_infra1 = value1`
+
+   Action = The tag for AWS resource is maintained to `key_infra1 = value1`.
```
staebler, line 235:
```suggestion
   Action = The tag for AWS resource is updated to `key_infra1 = value1`.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+
+   Final tag set to AWS resource = `key_infra1 = custom_value`
+
+   Event action = An event is generated to notify user about the action status (success) to update tags for the AWS resource. A warning also must be generated about list of user-defined tags on which action did a override.
+
+3) `.spec.platformSpec.aws` take precedence over `.status.platformStatus.aws`. User-defined tags must be merged from `.status.platformStatus.aws` and set to AWS resource.
+
+4) When `.status.platformStatus.aws` has a same user-defined tag as in `.spec.platformSpec.aws`,the entry must be removed from `.status.platformStatus.aws` before deleting user-defined tag from AWS resource.
```
staebler, line 243: The language of this sentence is ambiguous. What does it mean that "the entry must be removed"? Does it mean that the operation of changing the infrastructure resource will fail? Does it mean that there is undefined behavior? Does it mean that some other action will not happen until the entry is removed?

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+   Setting user-defined tag to empty string in `.spec.platformSpec.aws` will not override the user-defined tag in `.status.platformStatus.aws`.
+
+#### Caveats
+1) User updates a resource's tag using an external tool when there is an entry in `.spec.platformSpec.aws.resourceTags`
+   The resource's tag will be reconciled by its owning operator to the value from `.spec.platformSpec.aws.resourceTags`.
+   The user-defined tag will be overwritten with value from `.spec.platformSpec.aws.resourceTags` when there is an update to `.spec.platformSpec.aws.resourceTags`.
```
staebler, line 249: Or an update to the tags defined on the component, right? For example, if the user updates the tags on a Machine, will all of the tags on the AWS Instance be updated at that point?

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+   Previous tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_infra1`
+
+   Action = Update existing tag with value from `.spec.platformSpec.aws.resourceTags`.
+
+   Final tag set to AWS resource = `key_infra1 = value_infra1`
+
+   Event action = An event is generated to notify user about the action status (success/failure) to update tags for the AWS resource.
+
+2) User deletes the user-defined tag from `.spec.platformSpec.aws.resourceTags`
+   The user-defined tag which is removed from spec, will not be reconciled or managed by operators.
```
staebler, line 264:
```suggestion
   The user-defined tag which is removed from spec will not be reconciled or managed by operators.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has no user-defined tag with key `key_infra1`
+
+   Action = No change in tags for the AWS resource.
+
+   Final tag set to AWS resource = `key_infra1 = value1`
+
+   Event action = No event
+
+3) User sets user-defined tag created using external tools to empty string for deleting for AWS resource.
```
staebler, line 280: I think it would be helpful to give the full example for this scenario.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+
+   There is no validation check involved for creator tool of user-defined tag. Any user-defined tag listed by user in `.spec.platformSpec.aws.resourceTags` is considered for create/update/delete accordingly.
+
+4) Any user-defined tag set using `.spec.platformSpec.aws.resourceTags` in `Infrastructure.config.openshift.io/v1` type will affect all AWS resources managed.
```
staebler, line 284:
```suggestion
4) Any user-defined tag set using `.spec.platformSpec.aws.resourceTags` in `Infrastructure.config.openshift.io/v1` type will affect all managed AWS resources.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+  as there will be no versioning of the user-defined tags required to override the user-defined tags in Infrastructure CR.
+  Instead, the user-defined tags (if supported in resource operator spec field, e.g: machine) can override the user-defined tags in Infrastructure CR.
+
+- As `.spec.platformSpec.aws.resourceTags` has the actual values. Sync to `.status.platformStatus.aws.resourceTags` was required in earlier design to identify if the
+  tag on the resource was created using Infrastructure CR or external tool. Identifying the creator tool was done to restrict user from editing the same user-tag kv pair using multiple tools.
+  This inherently poses many scenarios of conflict which will result in user-defined tag kv pair being inconsistent across cluster when applied using Infrastructure CR. Hence, user will be confused which tool to be used to update tag.
+
+6) User applies user-defined tag using `.spec.platformSpec.aws.resourceTags`. Later, user modifies the user-defined tag for AWS resource using external tool.
+   In this case, the desired value is not set immediately for AWS resource by the owning operator. There is eventual consistency maintained by the owning operator.
```
staebler, line 298: This is in strict opposition to the behavior that is described earlier in the EP. If there is eventual consistency, then without any further user action, the tag would eventually be updated on the AWS resource. The EP has stated previously that the tag on the AWS resource will not be updated until the user-defined tags are modified.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+  `.spec.platformSpec.aws.resourceTags` to `.status.platformStatus.aws.resourceTags` sync is not required,
+  as there will be no versioning of the user-defined tags required to override the user-defined tags in Infrastructure CR.
+  Instead, the user-defined tags (if supported in resource operator spec field, e.g: machine) can override the user-defined tags in Infrastructure CR.
+
+- As `.spec.platformSpec.aws.resourceTags` has the actual values. Sync to `.status.platformStatus.aws.resourceTags` was required in earlier design to identify if the
+  tag on the resource was created using Infrastructure CR or external tool. Identifying the creator tool was done to restrict user from editing the same user-tag kv pair using multiple tools.
+  This inherently poses many scenarios of conflict which will result in user-defined tag kv pair being inconsistent across cluster when applied using Infrastructure CR. Hence, user will be confused which tool to be used to update tag.
+
+6) User applies user-defined tag using `.spec.platformSpec.aws.resourceTags`. Later, user modifies the user-defined tag for AWS resource using external tool.
```
staebler, line 297: Let's make sure that the term "user-defined tag" is only being used for setting the tags on the kube resource.
```suggestion
6) User applies user-defined tag using `.spec.platformSpec.aws.resourceTags`. Later, user modifies the tag on the AWS resource using external tool.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+
+- As a cluster administrator of OpenShift, I would like to be able to update user-defined tags managed by OpenShift for AWS resources. User will need to update the user-defined tags to different set of values during ongoing operations.
+  This can be supported by allowing edit on mutable `.spec.platformSpec.aws.resourceTags` field in Infrastructure CR.
+
+- As a cluster administrator of OpenShift, I would like to be able to remove user-defined tags from `.spec.platformSpec.aws.resourceTags` without deleting for AWS resources.
+  This enables, user-defined tags to be modified on AWS without being overridden by OpenShift operators. The user can adopt new 3rd party component to manage tags.
+
+- As a cluster administrator of OpenShift, I expect user-defined tags added in Infrastructure CR are reconciled and desired user-defined tags maintained on AWS resources.
+  The latest updates on `.spec.platformSpec.aws.resourceTags` field must be reconciled to AWS resources and desired user-defined tags maintained when being managed by Openshift.
```
staebler, line 324:
```suggestion
  The latest updates on `.spec.platformSpec.aws.resourceTags` field must be reconciled to AWS resources and desired user-defined tags maintained when being managed by OpenShift.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+  Any modifications which are external to Openshift must be reconciled to desired user-defined tags from `.spec.platformSpec.aws.resourceTags`.
```
staebler, line 325:
```suggestion
  Any modifications which are external to OpenShift must be reconciled to desired user-defined tags from `.spec.platformSpec.aws.resourceTags`.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
+#### Support Procedures
+
+NA
+
 ## Design Details
 
+User can add tags during installation for creation of resources with tags. Installer creates infrastructure manifests with `resourceTags` to `.spec.platformSpec.aws` of the `Infrastructure.config.openshift.io`.
+Operators that operate on aws resources must consider the tags from `.spec.platformSpec.aws`, `.status.platformStatus.aws` and existing AWS resource tags.
+
+The values from `status.platformStatus.aws` will used to only support older version.
```
staebler, line 409:
```suggestion
The values from `status.platformStatus.aws` will be used to only support older version.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +101,373 @@ The userTags field is intended to be set at install time and is considered immut
...
-
-1. they would have to get help from support to edit the status field
-2. they would have to manually remove the undesired tags from any existing resources
+If a user decides that they do not want these tags applied to new resources, there is no clean way to address that need:
 
-In the future we will want to introduce a spec field where a customer can specify and edit these tags, which will be reflected into status when changed, and we will expect consuming components to reconcile changes to the set of tags by applying net new tags to existing resources (but they will still not remove tags that are dropped from the list of tags).
+1. User would have to get help from support to edit the status field for older version which uses `experimentalPropagateUserTags`.
+2. User would have to manually remove the undesired user-defined tags from some existing resources even though user-defined tags can be marked for deletion by setting tag value to empty string.
+3. User would have to manually update user-defined tags for AWS resources which are not managed by an operator.
```
staebler, line 458: I would like to see an overview of how we would expect a user to go about updating non-managed AWS resources. I expect that we are going to need to provide guidance in the docs anyway.

staebler (commented):
This is looking much nicer. Thanks for the persistence on this, @TrilokGeer.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: The tags on kube resources override from infra CR.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Agree. Thanks.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Yes, agreed. Updated all to use "kube resource" terminology.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: The term "action status" is ambiguous here. Rephrased to "request status". The "failure" status is added in event as the tag is not updated as per the request.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Rephrased to add more on user usage  PoV.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: The statement is ambiguous in the context. It was supposed to be a note, that "updates on Infra CR are still applied post reconcile". I think, there is no need to bring it out explicitly.

enhancements/api-review/custom-tags-aws.md
staebler: The issue here is that there is no request. There is only spec. Controllers have no way of reliably reacting to a specific edge. Controllers can only react to the current spec, perhaps with knowledge of the previous seen spec.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Ok, added an example for the same.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: As per the general desired value reconciliation, it is better to reconcile any updates. In the previous, version it was discarded taking into consideration for rate limiting and traffic generated. After discussion with @JoelSpeed , concluded that it is not much of a concern on rate limiters for the APIs and is implementation specific e.g: avoiding  adding any new API calls in existing reconcile loops.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Generally, cluster admins tend to set tags with default value set to empty string during creation of resources. Later based on the usecase, the values are updated in these tags or set to empty string. The OpenShift installer fails on validation for using empty strings.

@staebler , 
1) do you suggest to keep the behaviour same with updates? i.e, fail validation on empty strings?
2) is it better to consider value as an optional field and consider "nil"/null value field for deletion?
In case of cloud providers, the value of a tag can be set to empty string but cannot be null.

enhancements/api-review/custom-tags-aws.md
JoelSpeed: > The EP has stated previously that the tag on the AWS resource will not be updated until the user-defined tags are modified.

This seems to be against kube-patterns doesn't it? We have a desired state defined within the cluster, if anyone modifies the actual state, a controller puts it back. 

If you want to have the controllers only act when there's an update to the tags defined on the infra resource, then each controller will need to keep track of which tags it previously applied, then calculate a diff, and ensure the ordering is consistent to avoid spurious diffs. This seems overly complex and isn't something we would normally do in OpenShift, are we really sure we want to introduce this now?

enhancements/api-review/custom-tags-aws.md
TrilokGeer: As there are multiple sources of tags information (kube resource, infra CR, AWS resource), controllers must use work queues efficiently to react. Yes, there are race conditions possible, but the precedence should be resolved eventually.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,397 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+
+   Final tag set to AWS resource = `key_infra1 = value1`
+
+   Event action = No event
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 =`
+
+   Action = No change in tags for the AWS resource.
+
+   Final tag set to AWS resource = `deleted
```
JoelSpeed, line 296: You've missed a closing backtick on this line which is messing up your formatting below

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,397 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+// AWSPlatformSpec holds the desired state of the Amazon Web Services Infrastructure provider.
+// This only includes fields that can be modified in the cluster.
+type AWSPlatformSpec struct {
+    // Existing fields
+    ...
+    // ResourceTags is a list of additional tags to apply to AWS resources created for the cluster.
+    // See https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html for information on tagging AWS resources.
+    // AWS supports a maximum of 10 tags per resource. OpenShift reserves 5 tags for its use, leaving 5 tags
+    // available for the user.
+    // ResourceTags field is mutable and items can be removed.
```
JoelSpeed, line 369: You need to explain what happens when a tag is removed within this comment. Think of this block as user facing documentation.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,397 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+    // +kubebuilder:validation:MaxItems=10
```
JoelSpeed, line 370: As OpenShift reserves 5, should this not be `MaxItems=5`?

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Thanks for locating it.

enhancements/api-review/custom-tags-aws.md
staebler: My suggestion would be to separate the tags to delete entirely from the tags to maintain. That way it is clearer to the user and less likely from them to experience surprises.

enhancements/api-review/custom-tags-aws.md
staebler: I am not concerned about the precedence. I am concerned about the statement that the controller will report a failure event when the infra CR is updated while a kube resource has a overriding value for the tag.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Do you think it is better to delete tags on AWS resources when removed from spec? This behavior will be consistent with usual CRUD-like object handling.

As tag update/delete can be overridden by entries in kube resource objects, making tags "unmanaged" wiithout deleting requires tricky methods to handle delete and maintain the limits for the number of tags. It is better to NOT support the following user story. 

```
- As a cluster administrator of OpenShift, I would like to be able to remove user-defined tags from `.spec.platformSpec.aws.resourceTags` without deleting for AWS resources.
  This enables, user-defined tags to be modified on AWS without being overridden by OpenShift operators. The user can adopt new 3rd party component to manage tags.
```
@JoelSpeed @staebler , Do you foresee any usecase that might require the above user story? Third-party tools operate without any OpenShift specific awareness of tag management, the user story does not seem to add much benefit. In this case, user would be required to delete the tag from Openshift and re-tag AWS resources using the third-party tools.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Oh, I misunderstood. Correct, status should be successful.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,400 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+   Action = The tag for AWS resource is updated to `key_infra1 = value1`.
+
+   Final tag set to AWS resource = `key_infra1 = value1`
+
+   Event action = An event is generated to notify user about the request status (success) to update tags for the AWS resource. A warning also must be generated about list of user-defined tags on which action did a override.
+
+3) `.spec.platformSpec.aws` take precedence over `.status.platformStatus.aws`. User-defined tags must be merged from `.status.platformStatus.aws` and set to AWS resource.
+
+4) Before deleting user-defined tag from AWS resource, the entry must be removed from `.status.platformStatus.aws`, if any, as operator will reconcile user-defined tag to the AWS resource.
+   Setting user-defined tag to empty string in `.spec.platformSpec.aws` will not override the user-defined tag in `.status.platformStatus.aws`.
```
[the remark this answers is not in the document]
TrilokGeer: The value in spec takes precedence only for updates. The status is maintained immutable in the upgrades also.
A central infra controller to sync spec to status field. The controller becomes a SPOF. As the different operators do not update request status back to the central controller, avoiding SPOF is better. As the status field is set for deprecation, future revisions of the feature may not require this function.

enhancements/api-review/custom-tags-aws.md
staebler: I agree with Joel. I don't think that we can support removal of AWS tags when a tag is removed from the Infrastructure's `resourceTags`.

enhancements/api-review/custom-tags-aws.md
staebler: I think we have two choices here.

1) We guarantee that a tag on an AWS resource will be restored within a certain period of time after the resource is updated externally.
2) Or, we guarantee that a tag on an AWS resource will be restored after either the `resourceTags` field on the Infrastructure resource is updated or the tags on the kube resource is updated. The controller may restore the tag on the AWS resource prior to that. But there is no guarantee that the controller will restore the tag unless there is a change to a resource in the cluster that prompts the controller to check whether the tag needs to e updated.

I think that option (1) is the better user experience, but I understand if there are some controllers that may have limitations that preclude that as a viable option. With either option that we choose, the EP needs to be consistent.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Yes, from UX perspective, it definitely helps user if there is more detailed and discrete tag information.

Following reasons user may find different lists confusing
1) There is no deterministic time to remove the entry from delete list.
2) Conflicting entries are possible which adds confusion to user if the delete list tag entry is not removed.
3) Conflict scenarios with override is possible. e.g: One of the scenarios is, when the kube resource overriding entry is removed and there is no entry in maintain list and delete list.

The following might help user to track tag history or requests,
1) For tag history, velero backups can be used to save.
2) For monitoring operations on tags, events are generated for tag add/update/delete requests.
Also, based on specific events, alerts can be triggered.
3) For monitoring tag changes on AWS resource, AWS CloudWatch Events can be explored.

It is less complex to remove entry from spec for deletion and let user opt for one of the above methods to track tags.

@JoelSpeed, Do you think it is ok to go ahead with the above methods for tracking tag?

enhancements/api-review/custom-tags-aws.md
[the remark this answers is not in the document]
staebler: Why would the value in spec only take precedence on updates?

enhancements/api-review/custom-tags-aws.md
staebler: I do not know of any. OpenShift uses the kubernetes.io/cluster/... tag on all resources. Some operators may add other tags. The 5 reservations is for future use so that we could add more tags if needed.

enhancements/api-review/custom-tags-aws.md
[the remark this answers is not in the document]
TrilokGeer: As delete is not supported for status. Only updates are considered for precedence.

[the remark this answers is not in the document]
staebler: I don't follow how delete not being supported for status is relevant. The precedence applies to the current values in the spec and not how the actions performed to get to the current values.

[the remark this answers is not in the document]
TrilokGeer: You are right, precedence is not related to actions. 

> The value in spec takes precedence only for updates. The status is maintained immutable in the upgrades also. A central infra controller to sync spec to status field becomes a SPOF. As the different operators do not update request status back to the central controller, avoiding SPOF is better. The status field is set for deprecation, future revisions of the feature may not require this function.

The meaning in the conversation here is, in case of delete operations, status value is not updated. Tags in spec override tags in status when there is an update operation. Delete operation fails if there is a tag of same key is found in status.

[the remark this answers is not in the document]
TrilokGeer: > The status is maintained immutable in the upgrades also.
> 
> A central infra controller to sync spec to status field becomes a SPOF. As the different operators do not update request status back to the central controller, avoiding SPOF is better. The status field is set for deprecation, future revisions of the feature may not require this function.

If the tag from status is not removed, on the next reconcile operator will add the tag entry back to AWS resource.

Is it ok to go ahead with above limitations about status field and allow delete when there is tag with same key in status?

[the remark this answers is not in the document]
staebler: The scenario being discussed here is the one where the user sets the value to an empty string. It is not the scenario where the tag is removed from `resourceTags`. The next reconcile will still see that the spec is directing the operator to delete the tag from the AWS resource, regardless of what is in the infra status.

enhancements/api-review/custom-tags-aws.md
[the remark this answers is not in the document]
sanchezl: Given that is seems that there will be no "one size fits all" logic to how the tags will be applied (e.g. create vs update and resource specific limitations) I don't think it is unreasonable to simply have each operator also validate the keys names while they are at it. A helper can be provided in library go to help with this.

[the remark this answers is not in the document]
sanchezl: Another option would be for the config-controller to copy only valid resourceTag entries into the `infrastructure.status.platformStatus.aws.resourceTags` and have operators observe those values instead of the `platformSpec.aws.resourceTags`.

enhancements/api-review/custom-tags-aws.md
JoelSpeed: I certainly think option 1 is the preferred choice as it has better UX and is, IMO, easier to implement. Certainly for Machine API, we reconcile the status of our AWS resources at most every 10 minutes, lets check when the other controllers do so and see if we can put an upper time bound on this.

Is there a value we are happy with for the upper time bound? 10/30/60 mins?

enhancements/api-review/custom-tags-aws.md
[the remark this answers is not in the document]
JoelSpeed: I'm of the strong opinion that if we ever want validation, it needs to be validated centrally and from the beginning of the introduction of the feature (we had these limitations on the previous iteration, we shouldn't lose them now).

While a webhook introduced now may be obsoleted in the future, the alpha feature may never graduate. I see this as a blocker for this feature and I believe the API reviewers will block the API implementation without this

Is there an API PR created already?

enhancements/api-review/custom-tags-aws.md
JoelSpeed: > the MaxItems includes both Openshift and user tags as a combined limit.

Can you explain why it includes both? At the moment, as a user, I could add 10 tags here and go over the limit of 5 that I'm allowed. I think what we want to do is limit this list to 5 items which allows the operators to add their own tags (eg the cluster tag and name tags) without going over the 10 tag limit.

enhancements/api-review/custom-tags-aws.md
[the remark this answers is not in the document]
JoelSpeed: To clarify some points here, I don't disagree with what @sanchezl is saying. We have a number of different scenarios to work with here.

1) Tags in the platformStatus - these are validated by the installer to ensure that they meet the namespace requirements, therefore any operator can apply them blindly without validation
2) Tags on a local object (eg Machine) - these must be validated by the operator when they are merged with the platformSpec and platformStatus tags, however, not every operator has these. 
3) Tags on the platformSpec
  a) If these are validated centrally, any operator applying them can apply them blindly, therefore, if they don't have their own local tags, they need to do no validation and can blindly apply both platformSpec and platformStatus tags. If they don't have their own local tags, they probably don't have a way to report that the tags are invalid (I would expect those with this ability to validate via a webhook)
  b) If these aren't validated centrally, each operator must validate them and reject anything that is invalid, this is async and so now we are left with the problem of, how is this reported to the user? Some operators will have status objects they put this on, some may not. What happens if the only reporting mechanism an operator has currently is via events? The UX of this, because its async, will not be as good as option a

enhancements/api-review/custom-tags-aws.md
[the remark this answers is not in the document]
TrilokGeer: As discussed over call, experimentalPropagateUserTags field will be supported. The field needs to be made optional.In the absence of  experimentalPropagateUserTags , default behaviour will be to consider user tags for propagation to AWS resources.

enhancements/api-review/custom-tags-aws.md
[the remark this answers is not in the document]
TrilokGeer: After discussion, the validation the is considered is default validation method supported : https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Based on the discussion with @JoelSpeed and @staebler ,  the change is going to be ambiguous to the user. It may also involve versioning the tags. It is decided to not consider in this enhancement.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -79,65 +99,323 @@ The userTags field is intended to be set at install time and is considered immut
...
+
+```go
+// AWSPlatformSpec holds the desired state of the Amazon Web Services Infrastructure provider.
+// This only includes fields that can be modified in the cluster.
+type AWSPlatformSpec struct {
+    // Existing fields
+    ...
+    // ResourceTags is a list of additional tags to apply to AWS resources created for the cluster.
+    // See https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html for information on tagging AWS resources.
+    // AWS supports a maximum of 10 tags per resource. OpenShift reserves 5 tags for its use, leaving 5 tags
```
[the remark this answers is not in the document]
JoelSpeed: Please resolve this comment

enhancements/api-review/custom-tags-aws.md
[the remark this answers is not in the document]
JoelSpeed: The comments on these fields are generated into user facing documentation so they should be as comprehensive as possible. Please add the comment as per my suggestion and fill the additional details required.

enhancements/api-review/custom-tags-aws.md
JoelSpeed: This still needs to be fixed, it should be 5 items

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,373 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+    // ResourceTags field is mutable and items can be removed.
+    // When a tag is removed from ResourceTags, the tag still persists on AWS resources.
+    // +kubebuilder:validation:MaxItems=10
+    // +optional
+    ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
+}
+
+// Existing type
+// AWSResourceTag is a tag to apply to AWS resources created for the cluster.
+type AWSResourceTag struct {
```
JoelSpeed, line 322: On the call, we discussed using a new type so that, in the future, if the spec and status need to drift, we don't add extra fields to the status that aren't required.
```suggestion
// AWSResourceTagSpec is a tag to apply to AWS resources created for the cluster.
type AWSResourceTagSpec struct {
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,373 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+    // key is the key of the tag
```
JoelSpeed, line 323: Bearing in mind that this is user facing documentation, please expand this to include sufficient detail such that a user can understand what value they are supposed to put in here
```suggestion
    // key is the key of the AWS tag. The value must consist only of <these characters> and must be no more than 128 characters in length.
    // Using kubernetes.io or openshift.io prefixes is prohibited. These tag namespaces are reserved for use by the platform.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,373 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=128
+    // +kubebuilder:validation:Pattern=`^(?!openshift.io)(?!kubernetes.io)([0-9A-Za-z_.:\/=+-@]+$)`
+    // +required
+    Key string `json:"key"`
+    // value is the value of the tag.
+    // Some AWS service do not support empty values. Since tags are added to resources in many services, the
+    // length of the tag value must meet the requirements of all services.
```
JoelSpeed, line 332: As above, include the requirements for the characters that are allowed and the length please. Remember to write it in a user friendly way

JoelSpeed (commented):
I have some wants for the documentation of the API to be updated, but otherwise this is good to go as far as I'm concerned

enhancements/api-review/custom-tags-aws.md
staebler: I care less about stipulating a specific upper time bound and more about ensuring that the EP is consistent with expectations. I am fine with leaving this part as saying that we ensure eventual consistency, so long as the previous part of the EP claiming that the tags on the AWS resource will only be updated in response to a spec change is updated to match this part.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,377 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+    // +kubebuilder:validation:MaxItems=5
+    // +optional
+    ResourceTags []AWSResourceTag `json:"resourceTags,omitempty"`
+}
+
+// Existing type
+// AWSResourceTagSpec is a tag to apply to AWS resources created for the cluster.
+type AWSResourceTagSpec struct {
+    // key is the key of the AWS tag. The value must consist only of alphabets, numbers and special characters(_.:\/=+-@).
+    // key should not consist "openshift.io" and "kubernetes.io" which are reserved.
```
JoelSpeed, line 325: Please update wording to
```suggestion
    // key is the key of the AWS tag. The value must consist only of alphanumeric and certain special characters(_.:\/=+-@).
    // key should not start with "openshift.io" or "kubernetes.io". Both of these prefixes are reserved for use by the platform.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,377 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+    // key must be no more than 128 characters in length.
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=128
+    // +kubebuilder:validation:Pattern=`^(?!openshift.io)(?!kubernetes.io)([0-9A-Za-z_.:\/=+-@]+$)`
+    // +required
+    Key string `json:"key"`
+    // value is the value of the AWS tag. The value must consist only of alphabets, numbers and special characters(_.:\/=+-@).
+    // value must be no more than 256 characters in length.
```
JoelSpeed, line 334:
```suggestion
    // value is the value of the AWS tag. The value must consist only of alphanumeric and certain special characters(_.:\/=+-@).
    // value must be no more than 256 characters in length.
```

JoelSpeed (commented):
/lgtm

I'm happy with the API for this now, lets continue and address anything further on the API PR

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,377 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+If a customer decides they do not want these tags applied to new resources, there is no clean way to address that need:
+
+1. they would have to get help from support to edit the status field
+2. they would have to manually remove the undesired tags from any existing resources
+
+In the future we will want to introduce a spec field where a customer can specify and edit these tags, which will be reflected into status when changed, and we will expect consuming components to reconcile changes to the set of tags by applying net new tags to existing resources (but they will still not remove tags that are dropped from the list of tags).
+
+### New proposal to support update of user-defined tags (or userTags)
+
+- `experimentalPropagateUserTags` field will be set for deprecation in this proposal version.
```
staebler, line 115: Missing the part where the `experimentalPropagateUserTags` will be changed to a `*bool`, where an omitted value means that the user tags *will* be propagated to the cluster.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,377 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+
+  The install validation will fail if there is any tag that starts with `kubernetes.io` or `openshift.io` in `userTags`. In the existing design, kubernetes.io and openshift.io namespaces are not allowed as part of user-defined tags.
+
+  The `resourceTags` field will be populated by the installer using the entries from `userTags` field.
```
staebler, line 119: This should be clear that the installer will only populate the field in the spec.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,377 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+
+   Existing tag for AWS resource = `key_infra1 = value_update1`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_update1`
+
+   Action = There is no update for AWS resource.
+
+   Final tag set to AWS resource = `key_infra1 = value_update1`
+
+   Event action = An event is generated to notify user about the request status (success/failure) to update tags for the AWS resource.
```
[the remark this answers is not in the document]
TrilokGeer: Yes, for example, it can be added to workqueue to be processed on the next reconciliation.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,377 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+
+   Existing tag in kube resource object like machine CR = `key_infra1 = custom_value`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value1`
+
+   Action = The tag for AWS resource is maintained to `key_infra1 = custom_value`.
+
+   Final tag set to AWS resource = `key_infra1 = custom_value`
+
+   Event action = An event is generated to notify user about the request status (success) to update tags for the AWS resource. A warning also must be generated about list of user-defined tags on which action is not applicable.
```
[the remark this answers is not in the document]
TrilokGeer: The event is generated when
1) There is update on kube resource and Infrastructure resource. There are no subsequent events generated when there are conflicts between kube resource and infrastructure resource.
2) Every time when there is conflict between kube/infrastructure vs AWS resource.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,377 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+  `.spec.platformSpec.aws.resourceTags` to `.status.platformStatus.aws.resourceTags` sync is not required,
+  as there will be no versioning of the user-defined tags required to override the user-defined tags in Infrastructure CR.
+  Instead, the user-defined tags (if supported in resource operator spec field, e.g: machine) can override the user-defined tags in Infrastructure CR.
+
+- As `.spec.platformSpec.aws.resourceTags` has the actual values. Sync to `.status.platformStatus.aws.resourceTags` was required in earlier design to identify if the
+  tag on the resource was created using Infrastructure CR or external tool. Identifying the creator tool was done to restrict user from editing the same user-tag kv pair using multiple tools.
+  This inherently poses many scenarios of conflict which will result in user-defined tag kv pair being inconsistent across cluster when applied using Infrastructure CR. Hence, user will be confused which tool to be used to update tag.
+
+6) User applies user-defined tag using `.spec.platformSpec.aws.resourceTags`. Later, user modifies the tag on the AWS resource using external tool.
```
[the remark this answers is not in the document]
TrilokGeer: Yes, merged the Caveat(6) to Caveat(1) about eventual consistency.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Ok, added the change to EP.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Correct, thanks for locating it. Updated the EP accordingly.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,378 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+
+   Existing tag in kube resource object like machine CR = `key_infra1 = custom_value`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value1`
+
+   Action = The tag for AWS resource is maintained to `key_infra1 = custom_value`.
+
+   Final tag set to AWS resource = `key_infra1 = custom_value`
+
+   Event action = An event is generated to notify user about the request status (success) to update tags for the AWS resource. A warning also must be generated about list of user-defined tags on which action is not applicable.
```
staebler, line 207:
```suggestion
   When there is a conflict between the user tags specified in the Infrastructure resource and the kube resource, the event generated on updates to the AWS resource tags will include details about the conflict. Note that if the conflict does not cause an update to the AWS resource tags, then the conflict will not be registered in an event.

   For example,

   Existing tag in `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value1`

   Tags specified in a kube resource updated to include a different value `key_infra1 = value2`

   Action = The tag for AWS resource is updated to `key_infra1 = value2`.

   Final tag set to AWS resource = `key_infra1 = value2`

   Event action = The event generated to notify the user about the request status (success) to update tags for the AWS resource will include details stipulating that the user tag specified for "key_infra1" in the Infrastructure resource was superseded by the tag specified for "key_infra1" in the kube resource.
```

enhancements/api-review/custom-tags-aws.md
```diff
@@ -31,35 +37,51 @@ have the ability to apply user defined tags to many resources created by OpenShi
...
+- User-defined tags can be updated post creation.
 
 ### Non-Goals
 
-- to reduce initial scope, tags are applied only at creation time and not reconciled. If an administrator manually
-   changes the tags stored in the infrastructure resource, behavior is undefined. See below.
-- to reduce initial scope, we are not implementing this for clouds other than AWS. We will not take any actions
-   to prohibit that later.
+- To reduce initial scope, the proposal does not apply for clouds other than AWS. There will be no actions taken to prohibit that later.
+
```
LalatenduMohanty, line 64: Does it also include resources created or used by OLM operators? I guess it is not. We should clearly mention that we are not trying to track resources from OLM operators.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -31,35 +37,51 @@ have the ability to apply user defined tags to many resources created by OpenShi
...
 
 ### Goals
 
-- the administrator or service (in the case of Managed OpenShift) installing OpenShift can pass an arbitrary
+- The administrator or service (in the case of Managed OpenShift) installing OpenShift can pass an arbitrary
    list of user-defined tags to the OpenShift Installer, and everything created by the installer and all other
    bootstrapped components will apply those tags to all resources created in AWS, for the life of the cluster, and where supported by AWS.
-- tags must be applied at creation time, in an atomic operation. It isn't acceptable to create an object and,
-   after some period of time, apply the tags post-creation.
+- User-defined tags can be applied at creation time, in an atomic operation.
```
LalatenduMohanty, line 58: s/in an atomic operation/as an atomic operation/

enhancements/api-review/custom-tags-aws.md
```diff
@@ -31,35 +37,51 @@ have the ability to apply user defined tags to many resources created by OpenShi
...
+- User-defined tags can be updated post creation.
```
LalatenduMohanty, line 59: When you say the tags can be updated post creation, who will be updating those ? I am assuming some controller will update those? Because if we let user edit these tags then it would be hard to keep track of these.

enhancements/api-review/custom-tags-aws.md
LalatenduMohanty: Also saw `User-defined tags cannot be updated to an AWS resource which is not managed by an operator. In this proposal, the changes proposed and developed will be part of openshift-* namespace. External operators are not in scope` , so please this to the non goals

LalatenduMohanty (changes requested):

staebler (commented):
LGTM

enhancements/api-review/custom-tags-aws.md
staebler: The controller that owns the AWS resource would be the one that updates the tags on the AWS resource, on behalf of a spec change by the user to either the Infrastructure resource or the kube resource for the AWS resource.

LalatenduMohanty: Sounds good to me.

dmage:
Should operators monitor created AWS resources and stomp manual changes of their tags?

staebler:
> Should operators monitor created AWS resources and stomp manual changes of their tags?

The expectation of an operator is that it will eventually stomp manual changes. The operator is not expected to actively monitor the AWS resources. The operator would be expected to check the tags on the AWS resources as part of its regular validation that the AWS resource is aligned with the spec of the kube resource.

See Caveat (1). https://github.com/openshift/enhancements/pull/1051/files#diff-9e7c9d80b69a46de2319af1610f54e1c0f04896901283d2e878bef4ecf915240R227

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Thanks for the review, updated the EP accordingly.

frobware:
LGTM

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,376 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+// Existing type
+// AWSResourceTagSpec is a tag to apply to AWS resources created for the cluster.
+type AWSResourceTagSpec struct {
+    // key is the key of the AWS tag. The value must consist only of alphanumeric and certain special characters(_.:\/=+-@).
+    // key should not start with "openshift.io" or "kubernetes.io". Both of these prefixes are reserved for use by the platform.
+    // key must be no more than 128 characters in length.
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=128
+    // +kubebuilder:validation:Pattern=`^(?!openshift.io)(?!kubernetes.io)([0-9A-Za-z_.:\/=+-@]+$)`
```
sanchezl, line 329: Negative look-ahead `(?! ... )` is not supported by the Go standard `regexp` library. OpenShift provides a [CRD admission plugin library](https://github.com/openshift/kubernetes/tree/master/openshift-kube-apiserver/admission/customresourcevalidation) that can be used to validate this field.

sanchezl (changes requested):

sanchezl (approved):
lgtm

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,376 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+
+   Action = No change in tags for the AWS resource.
+
+   Final tag set to AWS resource = `key_infra1 = value1`
+
+   Event action = No event
+
+3) Any user-defined tag set using `.spec.platformSpec.aws.resourceTags` in `Infrastructure.config.openshift.io/v1` type will affect all managed AWS resources.
+
+4) User-defined tags are not synced from `.spec.platformSpec.aws.resourceTags` to `.status.platformStatus.aws.resourceTags` for the following reasons.
```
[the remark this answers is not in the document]
TrilokGeer: Correct.

enhancements/api-review/custom-tags-aws.md
TrilokGeer: Ok, updated the same in https://github.com/openshift/api/pull/1064.

akhil-rane (commented):
lgtm

tjungblu:
Glad to see this is finally done. Thanks for your review folks, and thanks for your persistence @TrilokGeer.

/lgtm
/approve

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,376 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
+On update of `resourceTags`, AWS resource is not created or restarted.
+
+In the case where a user-defined tag is specified in the Infrastructure resource and
+1) There is already a tag with the same key and value present for AWS resource, the tag for the AWS resource will not be updated.
+
+   For example,
+
+   Existing tag for AWS resource = `key_infra1 = value_update1`
+
+   New tag request = `.spec.platformSpec.aws.resourceTags` has `key_infra1 = value_update1`
```
bparees, line 171: these two values are the same.  should the first one be "value_compupdate1" and the second one "value_infraupdate1"?

staebler: It is intentional that the two values are the same. This scenario is the following.
```
There is already a tag with the same key and value present for AWS resource
```
This scenario is demonstrating that adding the tag to the Infrastructure resource will not cause an update of the AWS resource when the AWS resource already has the matching tag.

We could argue whether this scenario is an interesting enough scenario to describe for a level-driven system.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -73,71 +95,376 @@ Note existing unchanged behavior: The installer will apply these tags to all AWS
...
 
 ## Alternatives
 
-TBD
+1. An operator can be designed to behave similar to how the installer's destroy behaves, namely by using the resource tagging API to search for resources with the kubernetes.io/.../cluster=owned tag.
+   The operator would reconcile on changes to the infrastructure CR to ensure that resources owned by the cluster have tags matching what is in the infrastructure CR.
+   That would get tricky when it comes to resources owned by in-cluster operators, though, particularly with tags that may have different values in the corresponding kube resources.
+   The operator would need to monitor multiple kube resources, infrastructure CR and multiple instances of different AWS resource types.
+
+2. A way to delete tags was considered when the user sets the user-defined tag value to an empty string in `.spec.platformSpec.aws.resourceTags`.
```
jewzaam, line 453: What about a new property like `.spec.platformSpec.aws.resourceTagsDelete` where user enumerates tags specifically to be deleted?  If there's something set on a kube resource the deletion request is ignored for that resource.

I worry we'll get to a point somewhere in the future where we need to change how we tag resources for ROSA.  If we have to build a tool to delete tags it negates the value of this enhancement.

bparees: @deads2k sketched out an api that would allow for deletion as:

```
spec.tags -- this is the current user desire to create.
spec.orphanTagsOnDeletion -- this is a list of tags for the platform to stop managing and not remove
status.tags -- this is the list of tags for the platform to create
status.tagsToRemove -- this is the list of tags for the platform to remove

a user adds to spec.tags and it is vetted and moved to status.tags

a user removes from spec.tags, it is compared against status.tags, and if there is no entry in orphanedTags, it is removed from status.tags and added to status.tagsToRemove.  if there is a matching entry in orphanTags, it is simply removed from status.tags
```

jewzaam: @bparees is that something to be addressed in another enhancement or just dropped?  I don't see delete handled in this enhancement hence my comment.  Deletion with logic to identify what should be managed by the platform vs external tools and having the ability to transition to an external tool is what is keeping delete from being included, from what I can tell.  If we then step away from having any logic and leave it to users to explicitly state what tags need deleted then the decision on what to keep vs delete is left to the administrator.

bparees: > @bparees is that something to be addressed in another enhancement or just dropped? 

neither, it's something that came up for discussion after this EP update had merged, and when the api changes were being reviewed.  I pasted it here so you could have some idea of how delete would work, if we were to include it as a capability and see what thoughts you'd have on it if that was the approach we took.

jewzaam: Thanks.  Seems a reasonable approach, just inverse of my initial post.  But the key thing is giving the control to the user with explicit list of tags to orphan.  Sure.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -31,35 +37,51 @@ have the ability to apply user defined tags to many resources created by OpenShi
...
+   Also, the CSI spec `https://github.com/container-storage-interface/spec/blob/master/spec.md` mentions grpc error code
+   `6 ALREADY_EXISTS` for `a volume corresponding to the specified volume name already exists but is incompatible with the specified capacity_range, volume_capabilities, parameters, accessibility_requirements or volume_content_source`.
+- User-defined tags cannot be updated for vpc, security groups, elb(not managed by in-cluster operators), route53, subnet resources as there is no operator managing the resources post installation.
+   User-defined tags cannot be updated to an AWS resource which is not managed by an operator. In this proposal, the changes proposed and developed will be part of openshift-* namespace. External operators are not in scope.
+- User-defined tags can be updated on the following AWS resources.
+  1. EC2 instances for master and worker nodes.
+  2. Image registry.
+  3. Ingress LB.
+  4. IAM credentials by CCO in mint mode of operation.
+- OpenShift is bound by the AWS resource that allows the fewest number of tags. An AWS S3 bucket can have at most 10 tags.
```
bparees, line 79: @sdodson @deads2k I had a conversation w/ SD about their needs/expectations and in general how we make this an api that an admin won't be surprised by the behavior of.

one of the thoughts would be rather than having a single "tags" abstraction field that:
1) will only update tags on some resources (ones we have active operator management of)
2) has to be constrained by the tag limit of the lowest resource tag limit (10)
3) needs to compete w/ machine spec which considers itself the only owner of ec2 tags, if you provide tags in that api (though possibly this could be changed)

we should instead break this out into specific api fields for each resource type we can actually manage the tags of (and no resource field for the ones we cannot).

it would make for a bit of an ugly api (and beg the next question of: why not just manage the tags via a config resource for each operator, which we already do in some cases, rather than specify them here where it's an action at a distance thing)

Another thing we discussed(possibly independent of that api shape decision) would be whether the operators that actually manage the resources should go degraded if they can't apply the requested tags.  This would at least give some feedback to the admin that the changes they tried to assert via the api did not get applied.  Unfortunately it's "action at a distance".  It would be better if the infrastructure resource reported the errors back, but introducing that sort of two way dependency between whatever manages the infra resource object, and whatever manages the aws resources, doesn't feel great to me.

thoughts?  Fundamentally I think it's important that if we're going to provide a tag-updating api interface, it be obvious to the users of that interface what aws resources will/won't be updated and what, in general, the limitations of that api are.  Otherwise they are going to add tags and then be surprised when some resources get the tags and others don't.  Which makes me think a generic "tag your aws resources via this field" api(regardless of how it is documented) is problematic.

ideas?

sdodson: That sounds like a pretty horrible user experience. What became of the approach that we set one tag, everywhere we can, and then additional tags are managed outside of OpenShift however you prefer?

bparees: > What became of the approach that we set one tag, everywhere we can, and then additional tags are managed outside of OpenShift however you prefer?

it's all still on the table, but the problem w/ that approach is that it's not obvious when interacting with the api, which resources will get the new/updated tag and which will not.  It requires reading the documentation, assuming the documentation is correct and exhaustive to understand if a resource falls into the category of:

1) we can fully manage/update the tags
2) we can create the tags at install time, but we can't update tags
3) we can't manage the tags at all (not even create)
4) we can manage the tags as long as you don't exceed N of them, so even though you'd normally expect this tag you added to get applied to X, it didn't because X was over its tag limit, while Y did get the tag because it's not over the limit.

a single "put your tag desires here" api field doesn't get us there, not without at least a related "status" field that tells you what you actually got.  And maintaining a status field on the infra resource will be difficult since it needs to go ask each operator that manages a resource what tags the operator has actually applied to the resources.  (infra resource becomes an aggregation over the various resource-managing operators, for resources that *have* an operator that manages them)

bparees: ah, yeah i sort of pitched that to SD.  They weren't a huge fan (basically punting tooling to them, and i think there were some more explicit concerns about how it would be done), but I think it's technically still on the table, depending on what SD and customers come back and tell us are their actual requirements.

@jewzaam @jharrington22 can better speak to their thoughts/concerns on that approach, though.

jewzaam: @jharrington22  and I talked over this and we'd like to keep tagging as a part of OCP.  Not having delete is fine initially.  At some point it's possible we care about this.  Our requirement is we can set a tag on all resources created by the installer and created by OCP after installation.  We require a single tag and it's the same on all resources.  Anything beyond that is a customer use case and I'm personally OK with not tackling those at this time.  Having a tag set on all resources will allow us to create policies limiting permissions in the customer AWS account to just the cluster created/managed resources.

PR #1053, opened 2022-03-07
Add agent-based installer concept

zaneb:
The agent-based installer is an installation method that will use a bootable, ephemeral installer image running on the hosts that are to become part of the cluster. The user will generate the image using a command-line tool. The image itself will contain components (such as [assisted-service](https://github.com/openshift/assisted-service/#readme) and [assisted-agent](https://github.com/openshift/assisted-installer-agent#readme)) of the Assisted and Multi-Cluster Engine (MCE) installation methods. The installer will be usable in both a fully-automated workflow, where configuration is provided upfront, and an interactive workflow where the user can interact with a GUI similar to the hosted Assisted installation service.

### DIFF
--- a/enhancements/agent-installer/agent-based-installer.md
+++ b/enhancements/agent-installer/agent-based-installer.md
@@ -1,0 +1,337 @@
+---
+title: agent-based-installer
+authors:
+  - "@zaneb"
+reviewers:
+  - "@avishayt"
+  - "@celebdor"
+  - "@hardys"
+  - "@lranjbar"
+  - "@mhrivnak"
+  - "@patrickdillon"
+  - "@pawanpinjarkar"
+  - "@romfreiman"
+  - "@rwsu"
+  - "@sdodson"
+approvers:
+  - "@dhellmann"
+creation-date: 2022-03-03
+last-updated: 2022-03-14
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - https://issues.redhat.com/browse/AGENT-9
+---
+
+# Agent-based Installer
+
+## Summary
+
+The agent-based installer is an installation method for on-premises clusters,
+that will use a bootable, ephemeral installer image running on the hosts that
+are to become part of the cluster. The user will generate the image using a
+command-line tool. The image itself will contain components (such as
+[assisted-service](https://github.com/openshift/assisted-service/#readme) and
+[assisted-agent](https://github.com/openshift/assisted-installer-agent#readme))
+of the Assisted and Multi-Cluster Engine (MCE) installation methods. The
+installer will be usable in both a fully-automated workflow, where
+configuration is provided upfront, and an interactive workflow where the user
+can interact with a GUI similar to the hosted Assisted installation service.
+
+## Motivation
+
+Every on-premises user has a first cluster ("Cluster 0"), and many of them will
+be installed in a disconnected environment. While powerful tools exist for
+creating multiple on-premises clusters (Hive, HyperShift, MCE), they all run in
+a cluster themselves, which must somehow be provisioned without these tools.
+
+Each of the current installation options has limitations that make it
+unpalatable to a subset of users. IPI is highly opinionated, and suffers from
+providing limited visibility of configuration problems in on-premises
+environments. UPI exposes users to the full complexity of OpenShift with little
+guidance. The (hosted) Assisted service is not available for IPv6-only (until
+Quay.io supports IPv6) or disconnected environments. For some users, none of
+these options is optimal.
+
+On-premises users need a simple way to deploy their first cluster, even in a
+disconnected environment, using their own automation (or even manual
+intervention) to boot hosts.
+
+### Goals
+
+- Install clusters in fully-disconnected environments.
+- Make use of assisted installation components to pre-validate the installation
+  environment.
+- Perform reproducible cluster builds from configuration artifacts, as well as
+  interactive installation.
+- Require no machines in the cluster environment other than those that are to
+  become part of the cluster.
+- Install clusters with single-node, compact, and high-availability topologies.
+- Be agnostic to the tools used to provision (virtual) machines so that users
+  can use their own (possibly pre-existing) tooling for provisioning.
+- Support install configurations with platform `none`, `baremetal`, or
+  `vsphere`.
+- Run on either Linux (any distro) or MacOS, in either x86-64 or arm64
+  architectures.
+- Install to machines with either x86-64 or arm64 architectures, independent of
+  the architecture the CLI tool runs on.
+
+### Non-Goals
+
+- Replace the IPI installer in any capacity.
+- Automate booting of the image on (virtual) machines.
+- Support install configurations for cloud-based platforms.
+- Provide an interactive API to drive automated deployments.
+- Operate in disconnected environments without the need to separately mirror
+  the release payload to a registry that is usable from the installed cluster.
+- Provide a mechanism to add and remove nodes from the cluster after it is
+  installed.
+- Generate image formats other than ISO.
+- Create minimal ISO images with a separate rootfs.
+- Run on Windows.
+
+## Proposal
+
+A command-line tool will enable users to build a custom RHCOS image, initially
+in ISO format, containing the components needed to install.
+
+A prerequisite for disconnected users is that the release image, and possibly
+an extra image containing installer components, is mirrored to a local registry
+that is accessible from the network environment where the cluster is to be
+deployed. The command-line tool will
+[retrieve](https://github.com/openshift/machine-os-images/) the base ISO from
+the container registry, and add an Ignition file to produce a cluster-specific
+install ISO. Users can then boot this image on the hosts that are to become
+part of the cluster using any method they choose - this could be existing
+automation that they have or manually booting using physical or virtual media.
+Container images required for installation will be pulled from the same
+registry as the base ISO, as will the OpenShift release that is being
+installed. One host (hereafter referred to as "Node 0") will be selected to run
+the assisted-service. All hosts will run the assisted agent. A single ISO will
+be used for all hosts in the cluster.
+
+In the automated workflow, all configuration data for the cluster to be
+installed (including the expected number of hosts) will be provided as input to
+the command-line tool at the time the ISO is generated. Users will (optionally)
+be able to supply custom manifests to augment the initial configuration, in
+addition to data required for installation such as the network config.
+Installation will proceed automatically once the requisite hosts have
+registered with Node 0 and passed validations. An API running on Node 0 will
+enable automated tools to monitor progress, detect errors, and retrieve logs
+until the OpenShift control plane is available.
+
+In the interactive flow, the ISO will be generated without collecting
+cluster-specific data, or with only enough data to configure networking so that
+the hosts will be able to retrieve container images from the registry and
+communicate with each other once they are booted. Users will interact with
+assisted-service on Node 0 to define and install the cluster. Unlike the hosted
+service, each instance of assisted-service will work with only a single
+cluster.
+
+### User Stories
+
+- As a user in a disconnected environment with no existing management cluster,
+  I want to deploy a cluster using my own automation for provisioning.
+- As a user in a disconnected environment with no existing management cluster,
+  I want to interactively deploy a cluster and receive immediate feedback about
+  configuration errors and problems.
+- As a user in a disconnected environment with no existing management cluster,
+  I want to deploy a cluster to a vSphere where I have permissions only to set
+  the boot image for one or more existing VMs.
+- As a user in a disconnected environment with no existing management cluster
+  and strict security protocols, I  want to deploy a cluster using only
+  physical media and access to the host consoles via a KVM switch.
+
+### API Extensions
+
+N/A
+
+### Implementation Details/Notes/Constraints [optional]
+
+We must allow the user to provision hosts themselves, either manually or using
+automated tooling of their choice. The ISO format offers the widest range of
+compatibility. If necessary, the user can decompose the ISO into a minimal
+version, or into PXE artifacts, using `coreos-installer`.
+
+Building a single ISO to boot all of the hosts makes it considerably easier for
+the user to manage. In some PXE environments this may even be compulsory. It
+also avoids the need for large amounts of storage.
+
+We must be able to install without access to any compute resource other than
+the machines that are to join the cluster. Thus, all run-time activities (that
+is, those that must take place in the same network environment as the eventual
+cluster) must happen within the ISO image once it is booted.
+
+The location of the release image will be customisable to allow for
+disconnected installation. Disconnected and IPv6-only users will be required to
+mirror the release image to a local registry that is accessible both from the
+environment where the ISO is generated and the environment the cluster will be
+deployed in. This could, but need not, be set up using
+[oc-mirror](https://github.com/openshift/oc-mirror#readme). The agent-based
+installer itself will not attempt to automate this aspect of the deployment.
+
+The components of the agent-based installer itself may be included in the
+release payload, or in a separate operator payload that is also mirrored to the
+local registry.
+
+No additional components will be added to the installed cluster, other than
+those explicitly requested by the user. In particular, the agent-based
+installer will **not** automatically add MCE for day 2 scale out of the
+cluster.
+
+Further details will be discussed in future enhancements.
+
+### Risks and Mitigations
+
+The agent-based installer adds yet *another* method of installing OpenShift,
+and overlaps some other installation methods in applicability. Thus, in some
+cases it will contribute to even more decisions the user has to make before
+getting started.
+
+No mechanism currently exists for monitoring progress of an assisted
+installation in which the assisted-service is stopped before the installation
+is complete (as it must in this case, when Node 0 reboots to join the cluster).
+It will have to be created. There may be a gap in debug coverage as Node 0 is
+installed.
+
+## Design Details
+
+### Open Questions [optional]
+
+- Should the command-line tool be a subcommand of openshift-installer, or a
+  standalone binary?
+- Will the containers for the assisted components be retrieved from the same
+  release payload, or will we require the user to do additional mirroring (e.g.
+  of MCE) as well?
+
+### Test Plan
+
+The agent-based installer will be covered by end-to-end testing using virtual
+machines (in a baremetal-like configuration), automated by some variation on
+the metal platform
+[dev-scripts](https://github.com/openshift-metal3/dev-scripts/#readme). This is
+similar to the testing employed on both the baremetal IPI and assisted
+installation flows.
+
+### Graduation Criteria
+
+**Note:** *Section not required until targeted at a release.*
+
+TBD
+
+#### Dev Preview -> Tech Preview
+
+- Ability to utilize the enhancement end to end
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+- Enumerate service level indicators (SLIs), expose SLIs as metrics
+- Write symptoms-based alerts for the component(s)
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+- User facing documentation created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+**For non-optional features moving to GA, the graduation criteria must include
+end to end tests.**
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+N/A
+
+### Version Skew Strategy
+
+N/A
+
+### Operational Aspects of API Extensions
+
+N/A
+
+#### Failure Modes
+
+N/A
+
+#### Support Procedures
+
+N/A
+
+## Implementation History
+
+Major milestones in the life cycle of a proposal should be tracked in
+`Implementation History`.
+
+## Drawbacks
+
+In disconnected environments, users will still need to mirror the release
+payload to an accessible local registry, which must be long-running (i.e. it
+continues to be available once the cluster is up).
+
+Since the installation tools cannot continue running to the completion of
+building the cluster (Node 0 must eventually be rebooted to join the cluster),
+it will necessarily not be possible to use a single API to track the entire
+process of installation.
+
+Despite the fact that RHCOS is an operating system designed exclusively to run
+OpenShift, users will still need to wait for OpenShift to be *installed* into
+it after boot. A total of three boots (which are very slow on baremetal due to
+RAM testing at startup) will still be required for each host to come up as an
+OpenShift node.
+
+## Alternatives
+
+### Run assisted-service outside of the cluster hosts
+
+Rather than running assisted-service on one of the hosts that is to become part
+of the cluster, we could require the user to provide somewhere else to run it
+in the same network environment, in either a Podman pod on a RHEL host or in a
+virtual machine. This avoids the complexity of running the installation
+components on a host that will itself have to be rebooted and come up as a
+member of the cluster control plane.
+
+However, there exist environments where the boot image is transported in on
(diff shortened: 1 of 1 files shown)


enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+last-updated: 2022-03-03
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - https://issues.redhat.com/browse/AGENT-9
+---
+
+# Agent-based Installer
+
+## Summary
+
+The agent-based installer is an installation method that will use a bootable,
```
dhellmann, line 28: It's covered well in the rest of the doc but it would be good to mention in the summary that the current focus is on-prem platforms.

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Goals
+
+- Install clusters in fully-disconnected environments.
+- Make use of assisted installation components to pre-validate the installation
+  environment.
+- Require no machines in the cluster environment other than those that are to
+  become part of the cluster.
+- Install clusters with single-node, compact, and high-availability topologies.
+- Be agnostic to the tools used to provision (virtual) machines.
```
dhellmann, line 65: You could elaborate on this by pointing out that the reason we want to be agnostic is that we have customers who want to bring their own provisioning tools.

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+### Non-Goals
+
+- Replace the IPI installer in any capacity.
+- Automate booting of the image on (virtual) machines.
+- Support install configurations for cloud-based platforms.
+- Provide an interactive API to drive automated deployments.
+- Install minor versions of OpenShift that are older than the agent-based
+  installer version.
+- Operate in disconnected environments without the need to separately mirror
+  the release payload to a registry that is usable from the installed cluster.
```
dhellmann, line 78: We have some future goals that could be called out as non-goals for this enhancement that would be covered by other enhancements later

- Support multiple image formats
- Create minimal bootable images and separate root filesystems

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+from the payload of the OpenShift release that is being installed.
+
+In the automated workflow, all configuration data for the cluster to be
+installed (including the expected number of hosts) will be provided at the time
+the ISO is generated. Installation will proceed automatically once the
+requisite hosts are available and have passed validations. An API running on
+Node 0 will enable automated tools to monitor progress, detect errors, and
+retrieve logs until the OpenShift control plane is available.
+
+In the interactive flow, the ISO will be generated with only enough data to
```
dhellmann, line 100: I still have questions about whether what's described here is the best UX we can provide. Let's put the interactive flow on the non-goals list for this document and focus on the automated flow.

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+Since the installation tools cannot continue running to the completion of
+building the cluster (Node 0 must eventually be rebooted to join the cluster),
+it will necessarily not be possible to use a single API to track the entire
+process of installation.
+
+Despite the fact that RHCOS is an operating system designed exclusive to run
+OpenShift, users will still need to wait for OpenShift to be *installed* into
+it after boot. A total of three boots (which are very slow on baremetal due to
+RAM testing at startup) will still be required for each host to come up as an
+OpenShift node.
```
dhellmann, line 228: What is the third boot, after booting the live ISO with the agent and then rebooting to run as a node?

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+  and strict security protocols, I  want to deploy a cluster using only
+  physical media and access to the host consoles via a KVM switch.
+
+### API Extensions
+
+N/A
+
+### Implementation Details/Notes/Constraints [optional]
+
+Further details will be discussed in future enhancements.
```
dhellmann, line 128: This section would be a good place to explain the reasoning behind each of the decisions described in the Proposal section.

"We must be able to run without a service on a host outside of the cluster, therefore we will build images that are self-orchestrating."

"We must support user-provided provisioning tools, therefore we will produce artifacts like ISOs that are ready to be consumed by those tools with minimal integration."

"We must support user-provided manifests like the openshift-installer does and will not have a REST API service, therefore the inputs to the image generator tool will be files."

"We want to minimize the storage requirements for provisioning images, therefore we will build one image that can be used on all hosts instead of creating host-specific images."

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+A command-line tool will enable users to build a custom RHCOS image, initially
+in ISO format, containing the components needed to install. The base ISO will
+be [retrieved](https://github.com/openshift/machine-os-images/) from the
+release payload. Users can then boot this image on the hosts that are to become
+part of the cluster using any method they choose - this could be existing
+automation that they have or manually booting using physical media. One host
+(hereafter referred to as "Node 0") will be selected to run the
+assisted-service. All hosts will run the assisted agent. A single ISO will be
+used for all hosts in the cluster. Required container images will be pulled
+from the payload of the OpenShift release that is being installed.
```
dhellmann, line 91: There's also an open question about this point below. Maybe here just say that the images will be pulled from the same local registry?

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+**For non-optional features moving to GA, the graduation criteria must include
+end to end tests.**
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+N/A
```
dhellmann, line 190: This section could talk about the goal of supporting versionable manifests as inputs so that the API of the image generator follows our existing support and deprecation guarantees for anyone integrating the tool with their workflow.

enhancements/agent-installer/agent-based-installer.md
zaneb: 1. Boot with live ISO, install to disk
2. MCO updates ostree on disk to match machine OS content shipped in cluster
3. Boot as Node

dhellmann:
> Thanks for creating this Enhancement PR and articulating the scope, approaches and decisions we are making. It may be useful to add a diagram of what the automated workflow looks like for the user.

+1, I have a slide I can give you as a basis for that, @zaneb 

> For goals, it may be worth adding it's targetting initial cluster (cluster 0) in fully-disconnected environments, so folks don't interpret this to mean any cluster (cluster 1, cluster 2, etc.)
> 
> It's unclear what the how or what input is provided in the statement "all configuration data for the cluster to be installed (including the expected number of hosts) will be provided at the time the ISO is generated."

If the team has done enough experimentation to feel comfortable making a decision about that, let's write it up as a separate enhancement.

> We should add some info about what the pre-reqs (mirrored registry), and also what happens after the cluster is stood up, specifically, what's the direction the user takes for day 2 activities even if it's to point to other tooling options to continue).

I think the mirror prereq is mentioned. Let's handle day 2 activities as another enhancement.

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+be installed in a disconnected environment. While powerful tools exist for
+creating multiple on-premises clusters (Hive, HyperShift, MCE), they all run in
+a cluster themselves, which must somehow be provisioned without these tools.
+
+Each of the current installation options has limitations that make it
+unpalatable to a subset of users. IPI is highly opinionated, and suffers from
+providing limited visibility of configuration problems in on-premises
+environments. UPI exposes users to the full complexity of OpenShift to the user
+with little guidance. The (hosted) Assisted service is not available for
+IPv6-only or disconnected environments. For some users, none of these options
```
romfreiman, line 51: It does available for dual stack and will be available to ipv6 once quay will be. 
I would limit it to disconnected?

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+- Be agnostic to the tools used to provision (virtual) machines.
+- Support install configurations with platform `none`, `baremetal`, or
+  `vsphere`.
+
+### Non-Goals
+
+- Replace the IPI installer in any capacity.
+- Automate booting of the image on (virtual) machines.
+- Support install configurations for cloud-based platforms.
+- Provide an interactive API to drive automated deployments.
```
romfreiman, line 74: :(

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+- Install minor versions of OpenShift that are older than the agent-based
```
romfreiman, line 75: Why do you mind?

enhancements/agent-installer/agent-based-installer.md
romfreiman: Why not? Isn't it built in already? Ipxe and minimal iso is there. I would say that releasing without minmal iso is sbuoptimal

romfreiman: It's a good point. Is it acceptable releasing without minimal iso support? 
@julienlim

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+
+Instead of doing an 'installation' as such, we could provide a pre-built
+Single-Node OpenShift image with the Multi-Cluster Engine (and potentially a
+registry with a mirror of the release image) installed. Users could then use
+MCE to expand the cluster to the desired size.
+
+However, scaling etcd from 1 to multiple nodes carries a significant risk of
+data loss (unlike scaling from 3 to 5 or more nodes). While this is likely
+acceptable during installation, it is not at any other time and there are thus
+no plans to implement this scenario in the etcd operator.
```
avishayt, line 256: This is unfortunate.  Having this would mean that we could install using the same code/logic as expanding a cluster (via MAPI/CAPI).

enhancements/agent-installer/agent-based-installer.md
zaneb: @romfreiman why would that be a goal?

enhancements/agent-installer/agent-based-installer.md
zaneb: AIUI users can build a minimal ISO from the ISO using coreos-installer, and they can certainly build PXE artifacts. We might want to document this but I don't think we have to support it directly in the tool to have a viable installer.

dhellmann: Right, remember this doc isn't the only place we'll talk about work on this project. Right now we're not focusing on multiple image formats. We know we'll need to, and we'll want a separate document when we do.

enhancements/agent-installer/agent-based-installer.md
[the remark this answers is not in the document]
dhellmann: This document does not describe how we would run an API service that an orchestration tool would use to trigger provisioning because we have requirements that have led us to conclude that the tool can't work that way. Some of the details I've suggested adding in my comment below (https://github.com/openshift/enhancements/pull/1053/files#r821861322) will make that clearer.

It's possible this bit of information would fit better in the proposal or details section (maybe in the rough overview of the workflow that @julienlim asked about), since it's a conclusion rather than a starting condition for the design.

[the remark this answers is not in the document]
zaneb: I included it in the non-goals because it isn't a goal: no user has requested it and nobody afaik has articulated a need for it. If it _were_ a goal then we would need to adjust the design and/or compromise on other goals. Nevertheless, it comes up a lot so it is worth being explicit about this as a starting condition.

This doesn't necessarily mean that it is an anti-goal, but the _conclusion_ is that due to the other constraints we have we're not going to have one.

romfreiman: Even just because we're in 2022 and everything in the world is API driven

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+providing limited visibility of configuration problems in on-premises
+environments. UPI exposes users to the full complexity of OpenShift to the user
+with little guidance. The (hosted) Assisted service is not available for
+IPv6-only or disconnected environments. For some users, none of these options
+is optimal.
+
+On-premises users need a simple way to deploy their first cluster, even in a
+disconnected environment, using their own automation to boot hosts.
+
+### Goals
```
romfreiman, line 57: I assume progress monitoring is one of the goals as well, right?

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+
+### API Extensions
+
+N/A
+
+### Implementation Details/Notes/Constraints [optional]
+
+Further details will be discussed in future enhancements.
+
+### Risks and Mitigations
```
romfreiman, line 130: I think installation progress reporting is a risk, no?

enhancements/agent-installer/agent-based-installer.md
romfreiman: hmm are u sure about the 3rd one?

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+data loss (unlike scaling from 3 to 5 or more nodes). While this is likely
+acceptable during installation, it is not at any other time and there are thus
+no plans to implement this scenario in the etcd operator.
+
+### Deliver the assisted components separately
+
+Rather than delivering the assisted components as part of the release image, we
+could ship them separately and have the ISO pull them either directly from MCE
+or from an operator in the OLM catalog. In disconnected environments, this
+would require the user to perform an extra step to mirror an additional image
```
romfreiman, line 263: They still mirror acm, operators...

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+obtained in a different way from the release image.
+
+This would also permit installation of older versions of OpenShift, at least
```
romfreiman, line 266: Theoretically, and practiallcaly, u can always use the same iso as a discovery entry point - anyhow the os-content is overwritten by MCO and there are no optimizations there. 
Few challenges are ignition version (that might be not backwards compatible) and drivers difference between the live iso and rhcos node os

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+with little guidance. The (hosted) Assisted service is not available for
+IPv6-only or disconnected environments. For some users, none of these options
+is optimal.
+
+On-premises users need a simple way to deploy their first cluster, even in a
+disconnected environment, using their own automation to boot hosts.
+
+### Goals
+
+- Install clusters in fully-disconnected environments.
```
romfreiman, line 59: Isnt multi version support is a requirement?

enhancements/agent-installer/agent-based-installer.md
zaneb: @cgwalters can you confirm/deny?

enhancements/agent-installer/agent-based-installer.md
zaneb: Nominally it isn't, but I'll get back to you in about 90 minutes :smile:

enhancements/agent-installer/agent-based-installer.md
zaneb: I agree, but we have to get the ISO from somewhere. Currently it's in the release payload, so that's the obvious place because there are no extra steps for the user.

enhancements/agent-installer/agent-based-installer.md
cgwalters: That's correct, 3 boots today.  If condensing boots 2 and 3 is important, it would be quite viable to do an "offline" update of the metal image before booting it.  There's some risks to this (e.g. suddenly we'd actually be using a potentially new Ignition version) but I think we can generally cover it in CI.

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,279 @@
...
+- Make use of assisted installation components to pre-validate the installation
+  environment.
+- Require no machines in the cluster environment other than those that are to
+  become part of the cluster.
+- Install clusters with single-node, compact, and high-availability topologies.
+- Be agnostic to the tools used to provision (virtual) machines.
+- Support install configurations with platform `none`, `baremetal`, or
+  `vsphere`.
+
```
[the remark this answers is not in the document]
dhellmann: Probably not the initial version because our first user isn't on ARM. It's on the roadmap for a near-future version.

zaneb:
I've updated with most of the suggestions from comments. The remaining ones are mostly requesting more detail about the automated flow - that's valid, but I think it will be better to put it all into a dedicated enhancement so that all the info about the automated flow is together. In this one I'll keep the automated and interactive flows at the same (low) level of detail. I think we are close to being able to merge this after another round of reviews.

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,328 @@
...
+assisted-service. All hosts will run the assisted agent. A single ISO will be
+used for all hosts in the cluster. Required container images will be pulled
+from the same local registry as the base ISO, as will the OpenShift release
+that is being installed.
+
+In the automated workflow, all configuration data for the cluster to be
+installed (including the expected number of hosts) will be provided as input to
+the command-line tool at the time the ISO is generated. Users will (optionally)
+be able to supply custom manifests to augment the initial configuration, in
+addition to data required for installation such as the network config.
```
imiller0, line 106: Will the input support the full range of what is possible with current Assisted Service CRs (ClusterDeployment, AgentClusterInstall, NMStateConfig, InfraEnv and extra installation manifests)?

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,308 @@
...
+Each of the current installation options has limitations that make it
+unpalatable to a subset of users. IPI is highly opinionated, and suffers from
+providing limited visibility of configuration problems in on-premises
+environments. UPI exposes users to the full complexity of OpenShift to the user
+with little guidance. The (hosted) Assisted service is not available for
+IPv6-only (until Quay.io supports IPv6) or disconnected environments. For some
+users, none of these options is optimal.
+
+On-premises users need a simple way to deploy their first cluster, even in a
+disconnected environment, using their own automation to boot hosts.
```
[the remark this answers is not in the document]
zaneb: I agree with you on the importance of automation, but that's not really the point here - the point is which parts of the automation the _user_ has control over.
If you want OpenShift to provide complete automation then IPI is available.
If you want to automate booting the hosts yourself and have OpenShift automate everything else, then the agent-based installer is for you, and if you are in a disconnected environment it is the only game in town.

dhellmann:
> I think we are close to being able to merge this after another round of reviews.

Yes, I agree. Let's give reviewers a last chance to comment and then I'll approve it Wednesday afternoon if there is no  feedback requiring significant changes.

/cc @avishayt @celebdor @hardys @lranjbar @mhrivnak @patrickdillon @pawanpinjarkar @rwsu @sdodson

enhancements/agent-installer/agent-based-installer.md
imiller0: Great. Thanks!

enhancements/agent-installer/agent-based-installer.md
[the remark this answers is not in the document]
imiller0: I think I'm well aligned with what both of you are saying. My intent was that there is significant value in being able to drive this installer within an automated framework where the user has a set of inputs (declarative CRs) which result in the ISO.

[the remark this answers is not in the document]
zaneb: Added this as a goal as suggested below.

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,330 @@
...
+
+Every on-premises user has a first cluster ("Cluster 0"), and many of them will
+be installed in a disconnected environment. While powerful tools exist for
+creating multiple on-premises clusters (Hive, HyperShift, MCE), they all run in
+a cluster themselves, which must somehow be provisioned without these tools.
+
+Each of the current installation options has limitations that make it
+unpalatable to a subset of users. IPI is highly opinionated, and suffers from
+providing limited visibility of configuration problems in on-premises
+environments. UPI exposes users to the full complexity of OpenShift to the user
```
mhrivnak, line 49: typo here. `s/users to //`

patrickdillon:
Relationship to `openshift-install` is stated as an open question, which makes sense as we continue to work through that design. So this LGTM as it stands.

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,330 @@
...
+  the release payload to a registry that is usable from the installed cluster.
+- Provide a mechanism to add and remove nodes from the cluster after it is
+  installed.
+- Generate image formats other than ISO.
+- Create minimal ISO images with a separate rootfs.
+- Run on Windows.
+
+## Proposal
+
+A command-line tool will enable users to build a custom RHCOS image, initially
```
mhrivnak, line 92: I think this paragraph would be easier to follow if it were more linear.
* An openshift release needs to be available in a local mirror
* The tool gets the base RHCOS image from that release
* The tool creates a new ISO
* The new ISO gets booted (fixes some current ambiguity on line 95 about which is "this" ISO that gets booted)
* Other required container images get pulled from the registry after boot

mhrivnak (commented):
Generally this looks good to me. It would be good to see the comment threads, and questions asked within, resolved before merging.

enhancements/agent-installer/agent-based-installer.md
dhellmann: Let's leave this for the detailed enhancement to be written next.

enhancements/agent-installer/agent-based-installer.md
dhellmann: It is, but that's well outside of the scope of work for this team.

enhancements/agent-installer/agent-based-installer.md
dhellmann: We'll get into details about that in the next enhancement.

enhancements/agent-installer/agent-based-installer.md
```diff
@@ -0,0 +1,337 @@
...
+the assisted-service. All hosts will run the assisted agent. A single ISO will
+be used for all hosts in the cluster.
+
+In the automated workflow, all configuration data for the cluster to be
+installed (including the expected number of hosts) will be provided as input to
+the command-line tool at the time the ISO is generated. Users will (optionally)
+be able to supply custom manifests to augment the initial configuration, in
+addition to data required for installation such as the network config.
+Installation will proceed automatically once the requisite hosts have
+registered with Node 0 and passed validations. An API running on Node 0 will
```
rwsu, line 117: Is the API the assisted-service REST-API?

rwsu (changes requested):
Looks good to me. Requested a minor clarification.

cgwalters (approved):

enhancements/agent-installer/agent-based-installer.md
dhellmann: Yes, that's right.

avishayt:
LGTM, thanks

dhellmann:
I think we have achieved consensus on the general approach. We'll be creating some more low-level enhancements very soon, so there will be more opportunities to ask questions and dig into details there.

/approve
/lgtm

enhancements/agent-installer/agent-based-installer.md
zaneb: Probably, but the vagueness was intentional (we may rule out e.g. a k8s API running on node0 and started prior to bootstrapping, but not in this enhancement).

PR #1055, opened 2022-03-08
updated the flag values with the corresponding format

sairameshv:
The Kube API Server flags ending with *-seconds require `int` values and hence updated the format of the values
Reference: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/


sairameshv:
cc: @harche

harche:
/lgtm 

This is in line with https://github.com/openshift/api/pull/1136#discussion_r821599219 

/cc @soltysh

soltysh (approved):
/lgtm
/approve

PR #1056, opened 2022-03-08
BUILD-406: shared resource validating admission EP

gabemontero:
EP that stems from early testing and demoing of the use of RHEL entitlements in OpenShift Builds via the new Shared Resources CSI Driver and Controller, which is tech preview in 4.10

This EP proscribes 3 major validations around the creation/modification of SharedConfigMaps and SharedSecrets, as well as the use of CSI volumes provisioned by the Shared Resource CSI Driver, via new k8s ValidatingAdmissionWebhooks.

While similar validations currently or will exist in the controller / CSI driver, use of the webhook provides better usability when dealing with violations of the validations.

### DIFF
--- a/enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
+++ b/enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
@@ -1,0 +1,524 @@
+---
+title: shared-resource-validation
+authors:
+  - "@gabemontero"
+reviewers:
+  - "@deads2k"
+  - "@jsafrane"
+  - "@adambkaplan"
+  - "@coreydaley"
+  - "@soltysh"
+  - Barnaby Court
+  - Chris Snyder
+approvers:
+  - "@deads2k"
+  - "@jsafrane"
+  - "@adambkaplan"
+api-approvers: # in case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers)
+  - "@deads2k"
+creation-date: 2022-03-08
+last-updated: 2022-03-29
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - https://issues.redhat.com/browse/BUILD-406
+see-also:
+  - "/enhancements/cluster-scope-secret-volumes/csi-driver-host-injections.md"
+replaces:
+  - "N/A"
+superseded-by:
+  - "N/A"
+---
+
+# Shared Resource Validation
+
+## Summary
+
+This Enhancement Proposal details improvements to existing validations currently performed, as well as the introduction
+of new validations to be performed, on artifacts related to the Shared Resource CSI Driver and Operator component introduced 
+as Tech Preview in 4.10.  As a quick reminder, the associated API for "Shared Resources", namely SharedSecrets and SharedConfigMaps, are 
+cluster scoped objects whose referenced namespace scoped Secrets and Configmaps are mounted into Pod Spec'able objects
+via use of the associated Shared Resource CSI Driver.
+
+With the introduction of the Shared Resources API in 4.10, any restrictions around use of the new CRDs or the declaring
+of Volumes using the new Driver in Pods, was done at the CSI Driver level.  That is often not the 
+most efficient location to enforce such restrictions, as the entire Kubelet provisioning path for the Pod must be exercised 
+for the error to be caught.  Also, the user does not learn of the error upon creation of the Pod Spec'able object.
+Instead, the user must inspect Pod status and Events to learn of the problem.
+
+This proposal intends to improve on the usability around error discovery by employing validations called during API object
+creation and modification.
+
+## Motivation
+
+### Goals
+
+- Ensure names of pre-supplied SharedSecrets and SharedConfigMaps by OCP components will not conflict with any SharedSecrets or SharedConfigMaps created by users.
+- Whenever possible prevent unsupported use of read-write volumes, as well as any future restrictions around the CSI Volume API or Shared Resource API
+- Provide pluggable means of validation of the associated Secrets and ConfigMaps of SharedSecrets and SharedConfigMaps.
+
+And provide validation failures on API object creation/modification for each of these whenever possible.
+
+### Non-Goals
+
+- Shared Resource controller, CSI Driver, Operator, nor Webhooks will determine the "Readiness" of any underlying `Secret` 
+or `ConfigMap` of a `SharedSecret` or `SharedConfigMap`.  Such determinations are the sole provenance of the Operators/Controllers/etc.
+that create those items.
+
+## Proposal
+
+See upstream doc 
+[https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook)
+for explanation of `ValidationAdmissionWebhooks`
+
+We will employ this plug point for inserting the code that performs any validations added.
+
+### User Stories
+
+As a developer using the Shared Resource CSI Driver
+I want validation for `readOnly: true` to be set on volumes for the Shared Resource CSI Driver to occur prior
+to the Pod Spec'able object getting persisted to etcd, so that pods are not stuck in "Creating" state waiting for a mount that will never succeed.
+Rather, I learn of the problem on the API Object PUT/POST.
+
+As an OpenShift operator maintainer or cluster administrator
+I want to reserve the `openshift-` prefix for SharedSecrets and SharedConfigMaps
+so that future OpenShift operators can create system-level shared resources and not have name conflicts 
+with items created by OCP users.
+
+As a developer, OpenShift operator maintainer, or cluster administrator
+I want SharedSecrets and SharedConfigMaps pre-populated by OCP Cluster Operators to make some indication on those
+SharedSecrets and SharedConfigMaps that the referenced Secrets and ConfigMaps are valid, functional, and ready for use.
+
+As a developer using the Shared Resource CSI Driver I want any validations to be provisioned and lifecycled in as much
+as possible the same fashion as the driver itself for consistency (i.e. via Cluster Storage Operator).
+
+As a developer or administrator using the Shared Resource CSI Driver I want sufficient metrics on the validations' prevention of incorrect Pods,
+SharedConfigMaps, and SharedSecrets, as a means of understanding the need for correcting misuse..
+
+As a developer or administrator using the Shared Resource CSI Driver I want necessary debug of the validation mechanism to be captured by
+OCP must-gather process.
+
+
+### API Extensions
+
+Quick note:  When discussing API, 'Shared Resource' on its own == SharedConfigMap / SharedSecret, and 'Reference' == the corresponding
+Secret or ConfigMap.
+
+A [ValidatingAdmissionWebhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook)
+will be used to inspect all created Shared Resources, and make sure their names do not start
+with the reserved prefix "openshift-" unless their Shared Resource name and Reference key are found in a 
+well known list of OCP pre-populated Shared Resource.  That list will be in well known ConfigMaps that resides 
+in the managed CSI Driver namespace.  Components / operators other than Shared Resources that want to create pre-populated
+SharedSecrets or SharedConfigMaps should submit PRs against the Shared Resource CSI Driver Operator Repo to update the 
+corresponding ConfigMap with the necessary information.  This is akin to how we create in-payload image streams via the PRs against the 
+samples operator today.
+
+There will be a "ocp-sharedsecret-list" and "ocp-sharedconfigmap-list" ConfigMap.  The key of an entry is the Shared Resource
+name.  The value should be of the form "namespace:name" in order to capture.  These ConfigMaps will be managed by the
+CVO/CSO infrastructure via there definition in the Shared Resourece CSI Driver Operator assets folder in the CSO repository.
+
+A brief example:
+
+```yaml
+kind: ConfigMap 
+apiVersion: v1
+metadata:
+  name: ocp-sharedsecret-list
+  namespace: openshift-cluster-csi-drivers
+data:
+  openshift-etc-pki-entitlement: openshift-config-managed:etc-pki-entitlement
+```
+
+The "openshift-*" naming restriction is new as compared to the initial tech preview release of Shared Resources in 4.10.
+
+A `ValidatingAdmissionWebhook` will be used to inspect 
+all created Pods, and k8s Pod derived types (Deployments, DaemonSets, etc.) with CSI volume mounts referencing the Shared
+Resource CSI Driver and make sure no read-write Shared Resource CSI Volumes are specified.  In other words, the volume attribute
+'readOnly: true' setting must exist.  Note, the understanding is that there is no "duck typing" in vanilla k8s or OpenShift that we can leverage to
+capture all "Pod Spec'able" API objects.
+
+For Builds, BuildRequests, and BuildConfigs, the existing [Aggregated OpenShfit API Server Build Admission plugin](https://github.com/openshift/openshift-apiserver/blob/master/pkg/build/apis/build/validation/validation.go)
+could be updated to enforce this restriction before the build controller in the OCM gets involved in translating a Build, BuildConfig,
+or BuildRequest into a Build Pod.  The Pod level ValidatingAdmissionWebhook could catch this as well, but catching the 
+error case sooner will provide even better usability.  In particular `oc start-build` invocations would now see the error
+immediately.
+
+The 'readOnly: true' for Pods restriction exists in 4.10, but is only enforced at the CSI driver level.
+
+On the associated `ValidatingWebhookConfiguration`, per guidance from David Eads, which should utilize the label selector
+capabilities to help filter which Pods are addressed by our new `ValidatingAdmissionWebhook`.  Consider the following 
+example:
+
+```yaml
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+...
+webhooks:
+  - name: pod.csi.sharedresource.openshift.io
+    failurePolicy: Ignore
+    namespaceSelector:
+      matchExpressions:
+        - key: runlevel
+          operator: NotIn
+          values: ["0", "1"]
+        - key: openshift.io/run-level
+          operator: NotIn
+          values: ["0", "1"]
+        - key: csi.sharedresource.openshift.io/skip-validation
+          operator: NotIn
+          values: ["true"]
+    rules:
+      - operations: [ "CREATE" ]
+        apiGroups: []
+        apiVersions: []
+        resources: [ "pods" ]
+        scope: "*"
+      - operations: [ "CREATE", "UPDATE" ]
+        apiGroups: [ "apps" ]
+        apiVersions: [ "v1", "v1beta1" ]
+        resources: [ "deployments", "daemonset", ....]
+        scope: "*"
+      - operations: [ "CREATE, "UPDATE" ]
+        apiGroups: [ "apps.openshift.io" ]
+        apiVersions: [ "v1" ]
+        resources: [ "deploymentconfigs" ]
+        scope: "*"
+    ...
+```
+
+Let's dive into the 2 different label
+- The first `matchExpression` is the means of preventing the use of our new webhook for any Pods of the core OCP control pane.  Because of this
+label selector we will filter out components like [the K8s API Server Operator Namespace manifest](https://github.com/openshift/cluster-kube-apiserver-operator/blob/d80b406ec4844328e04253f9ed5aec3f4a29834a/manifests/0000_20_kube-apiserver-operator_00_namespace.yaml#L11)
+- The second `matchExpression` is the means of preventing the use of our new webhook for any Pods in other OCP namespaces we want to exclude.
+Initially, we are thinking only the [CSI Drivers namespace](https://github.com/openshift/cluster-storage-operator/blob/2a3fbb68afd066d16582b09f9fd834c7fd03457e/manifests/02_csi_driver_operators_namespace.yaml) 
+will be excluded (as that is where we will also host our webhook), so
+to allow for the use of Shared Resources in OCP operators in the future for accessing "popular" Secrets and ConfigMaps (think the Global Proxy certs).
+
+
+Next, the idea of validating the References of any future well-known, pre-populated Shared Resources
+(like the entitlement secret) by other OCP CVO Operators was also raised during the review of the 4.10 blog post fof Shared Resources.  
+
+The discussion landed around these points:
+- A golang constant will declare the name of a k8s Condition that other components (i.e. OC Operators besides Shared Resources) will
+set on any Shared Resource that they pre-populate into OCP clusters.
+- REMINDER: the status structs of Shared Resources have an array of Conditions.  See 
+[here](https://github.com/openshift/api/blob/dbc82f2a3bc8b07e39628d36aee19d5a62847e2c/sharedresource/v1alpha1/types_shared_configmap.go#L83-L88) and
+[here](https://github.com/openshift/api/blob/dbc82f2a3bc8b07e39628d36aee19d5a62847e2c/sharedresource/v1alpha1/types_shared_secret.go#L81-L86)
+- The components who create pre-populated Shared Resources should have a controller that performs whatever validations
+they deem possible to verify the References, and those controllers are what will update the Status Condition arrays
+for the pre-populated Shared Resources
+
+With the initial implementation, 2 new constants will be defined along with the current status API, with godoc explaining
+the intent of these conditions, and who manages them:
+- `Available`: indicates that the Reference exists
+- `Ready`: indicates that the Reference is ready to be consumed, or facilitate the purpose for which it was created (i.e. the associated user can access his entitled content with the entitlement secret)
+
+Who sets the conditions:
+- 'Available': the Shared Resource controller can set this as part of its existing watches on the Shared Resources (creation, update, and relist).  Updating Shared Resources on 
+Reference delete events is *possible*, though the implementation details are trickier there, so this EP will not call for that (though we can revisit during implementation and its PR review),
+in the name of "eventual consistency".  An OCP operator which is pre-populating Shared Resources for References it manages is allowed to set it as well if it likes, but it is not required.
+- `Ready`: if the Reference is not available, the Shared Resource controller *could* by extension set Ready to false; however, we absolutely want to avoid the slippery slope of consumers 
+thinking that the Shared Resource controller deals with the "readiness" of the underlying Reference of the Shared Resource; so this EP is putting the stake in the ground that only the 
+contributing OCP operator maintains `Ready`.
+
+At this time, neither the CSI Driver / Operator nor any Webhooks will react to external controllers/operators manipulating
+the "validity" conditions.  See the alternatives section for potential considerations for doing so.  But human inspection,
+controller reaction via watches, as well as metrics/alerts, are all possible off of this condition.
+
+To further illustrate, consider this specific example:
+- the Insights Operator starts creating (via their CVO manifest list, or Operator logic) a SharedSecret to encapsulate the entitlement Secret (namespace: openshift-config-managed, name: etc-pki-entitlement) they maintain
+- Insights Operator may in fact be already performing some "validations" around that secret
+- but if not, we work with them to add or augment what they have
+- but with their validations, if Insights Operator discover something wrong with the entitlement secret, the Insights Operator updates the status of the SharedSecret they create with the appropriate conditions, setting true/false, reason, message on our well known "validity" condition, explaining something is wrong
+- and if we get a support case that says "entitled builds don't work", we will ask them to check the conditions on the SharedSecret that Insights Operator manages, where
+  the Reason/Message Insgiht Operator puts there helps the customer figure out what he needs to do to fix his entitlement credentials (maybe the customer has to log onto console.redhat.com and do something) and then his entitlement builds work
+
+This is very similar to what CVO operators do with the well known conditions CVO defines for the ClusterOperator object in order 
+to report their status.
+
+### Implementation Details/Notes/Constraints [optional]
+
+Generally, basic field checking of the API Object created or modified when the webhook is called should be sufficient
+for the 'readOnly: true' and 'openshift-*' restriction enforcement.
+
+No special handling of existing Shared Resources whose names start with 'openshift-*' and were created by customers when OCP was at a 
+4.10.x level is required, since with "TechPreviewNoUpgrade", we have no upgrade to worry about.  A customer will have to 
+create a new 4.11.x cluster, and if they attempt to create a Shared Resource whose name starts with "openshift-*", 
+the 4.11.x webhook will prevent it, and they will simply have to pick a new name.
+
+The webhook will handle the "openshift-*" naming restriction for creates of Shared Resources.
+The webhook will read the "allow list" ConfigMaps on startup (the "allow list" ConfigMap should not change except on upgrade, as 
+it is CSO Operator managed item) and so the webhook's validations should not involve any RPC calls, thus minimizing SLI
+impact on the API Server.
+
+Existing checks in the controller for the presence of "readOnly: true" check will be kept.  Analogous checks for 
+who creates "openshift-*" Shared Resources will also exist in the controller.  This covers the case when the webhook
+might be down.
+
+### Risks and Mitigations
+
+Since Shared Resources is Tech Preview 4.10, we can introduce validation controls that "break existing API usage" in later releases than 4.10.
+In particular, any users who were creating SharedSecrets of their own that start with "openshift-".  But we have to 
+have all such restrictions in place before moving from Tech Preview to GA.  The 4.10 blog post does give a warning
+that we might restrict use of Shared Resources names starting with "openshift-*"
+
+And again, the "readOnly: true" validation was already in effect in 4.10, but only enforced at the controller.
+
+That said, even with Tech Preview, release notes are warranted if we go down the path of invalidating of 
+Shared Resources that were "OK" in 4.10.  And user push back is at least conceivable on a case by case basis.
+
+Any Pod level validation logic will get utilized *A LOT*, so it better work.  See the [Failure Modes](#failure-modes)
+section for our rationale on being OK with taking this check on.
+
+## Design Details
+
+
+### Open Questions [optional]
+
+- Metrics for some persistent volumes were found at [https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#kube-controller-manager-metrics](https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#kube-controller-manager-metrics)
+but nothing was found for in-line CSI ephemeral volumes other than the alpha level feature [External Health Monitor](https://kubernetes-csi.github.io/docs/volume-health-monitor.html).  Are we missing anything?
+- Are their any issues with preventing UPDATES to the workload types if they specify CSI volumes where 'readOnly: true' is not set?
+
+
+### Test Plan
+
+The existing Shared Resource CSI Driver e2e tests will be tweaked to look for read-write failures on object creation in addition
+examining events and Pod status.  We'll see if we can force a mode where the webhook is disabled and the controller checks still work.
+
+New e2e tests will be added to our suite to ensure any Shared* objects whose names start with
+"openshift-" must match up with the list of allowed items in the designated ConfigMap in the OpenShift CSI Driver namespace.  
+Again if tenable will have flavors for both when the webhook is present and when only the controller is present.
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+
+N / A
+
+#### Tech Preview -> GA
+
+Adding just this feature will not result in this feature moving from Tech Preview as introduced in 4.10 to GA.
+
+We also need the storage team's work and SCC/Pod Policy and formalizing the approach for unprivileged users 
+creating objects with use of the Shared Resource CSI Driver.
+
+Nor is it the gating factor on the API moving past the v1alpha1 level introduced in 4.10.
+Ideally, we want a period of usage in the field and feedback from users before making the call for API 
+version promotion from v1alpha to v1beta or v1.
+
+#### Removing a deprecated feature
+
+N / A
+
+### Upgrade / Downgrade Strategy
+
+Technically, with still requiring TechPreviewNoUpgrade, we don't have to worry about upgrade / downgrade yet.
+
+However, some thoughts on preparing ourselves for when we promote from tech preview to GA:
+- As a cluster storage operator managed item, upgrades/downgrades follow the same pattern established there, both in how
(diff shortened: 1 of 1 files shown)


enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,407 @@
...
+
+## Design Details
+
+
+### Open Questions [optional]
+
+ 1. Is there a "Pod Spec'able" filter wrt configuring validating admission webhooks, so we catch all the subtypes of Pod?  Or do we have to list each k8s type that has Volumes in the spec?  Do not think there is a "pod spec'able" filter, but let's confirm.
+ 2. Existing k8s metrics cited below are sufficient for execution times for each request/response interaction when our new admission validations
+    are called, correct? If not, does it make sense to have new metrics be histograms around the admission validations execution times?
+ 3. Validating webhooks can define / maintain metrics and a metric service, right?
```
jsafrane, line 183: Yes, from metrics perspective, a webhook it's just a Pod that exposes a metric endpoint.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,407 @@
...
+The existing or lack there of the ValidatingAdmissionWebhook should not impact a 4.10 version of the current Controller/CSI Driver.
+Either the kubelet calls it or it does not call it during upgrades.  Either it gets watch events from Kubernetes or it does not
+during upgrades.  The enhancement here in no way depends on any elements of how the kubelet and a CSI driver operator that 
+do not exist in 4.10, when Shared Resources were introduced.
+
+The CSI Driver does make a SAR request, as well as access the contents of the associated Secret or ConfigMap, during its processing.
+But if the API Server is not up to honor those requests, then any requests from the kubelet to mount our volumes will be denied.
+But that is no different than the level of service provided in 4.10.
+
+### Version Skew Strategy
```
jsafrane, line 246: Until the CSI driver is behind I don't think it's necessary to add convoluted conditions to validate existing Shared Resources, there can't be any version skew.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,407 @@
...
+## Design Details
+
+
+### Open Questions [optional]
+
+ 1. Is there a "Pod Spec'able" filter wrt configuring validating admission webhooks, so we catch all the subtypes of Pod?  Or do we have to list each k8s type that has Volumes in the spec?  Do not think there is a "pod spec'able" filter, but let's confirm.
+ 2. Existing k8s metrics cited below are sufficient for execution times for each request/response interaction when our new admission validations
+    are called, correct? If not, does it make sense to have new metrics be histograms around the admission validations execution times?
+ 3. Validating webhooks can define / maintain metrics and a metric service, right?
+ 4. Since we are going to have backup validations continue in the controller, can the webhook and controller share metrics, or do we need them to maintain separately named, but logically equivalent metrics?
```
jsafrane, line 184: They can share metrics, still, I am not sure there is a need for the controller + extra conditions.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,407 @@
...
+
+That said, even with Tech Preview, release notes are warranted if we go down the path of invalidating of 
+Shared Resources that were "OK" in 4.10.  And user push back is at least conceivable on a case by case basis.
+
+## Design Details
+
+
+### Open Questions [optional]
+
+ 1. Is there a "Pod Spec'able" filter wrt configuring validating admission webhooks, so we catch all the subtypes of Pod?  Or do we have to list each k8s type that has Volumes in the spec?  Do not think there is a "pod spec'able" filter, but let's confirm.
```
jsafrane, line 180: I wish I knew the answer. IMO, PodSpec can be embedded in CRDs / external API servers and there is no easy way to find them.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,407 @@
...
+corresponding ConfigMap with the necessary information.  This is akin to how we create in-payload image streams via the PRs against the 
+samples operator today.
+
+There will be a "ocp-sharedsecret-list" and "ocp-sharedconfigmap-list" ConfigMap.  The key of an entry is the Shared Resource
+name.  The value should be of the form "namespace:name" in order to capture. 
+
+The "openshift-*" naming restriction is new as compared to the initial tech preview release of Shared Resources in 4.10.
+
+A ValidatingAdmissionWebhook will be used to inspect all created Pods, and k8s Pod derived types (Deployments, DaemonSets, etc.), 
+as well Builds and BuildConfigs, and make sure no read-write Shared Resource CSI Volumes are specified.
```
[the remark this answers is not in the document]
gabemontero: yep although I cannot find any linked references, there have been discussions around this idea

*IF* I recall correctly, I thought we received a general warning from @deads2k that we should stay away from mutating webhooks

Now, if I'm not remembering the outcome of the conversation correctly, or if @deads2k 's stance has changed, or if there are exceptions to that stance which this scenario falls under, then yes, just mutating a volume specs to insure `readOnly: true` for use of the Shared Resource CSI driver does have the feel of "helping usability" even further.

@deads2k - WDYT?

thanks

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: thanks for confirmation

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: Per my agreement at https://github.com/openshift/enhancements/pull/1056#discussion_r823895693 I will clean this up too

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: Yeah aside from not worrying about 4.10 customer instances with "openshift-*" names, the EP template does asks questions around "what happens if the new component is down" and I do like the idea of the 4.11 version of the controller being able to enforce our restrictions if the webhook is inexplicably down for some reason, but enough of the API server is available so that pods with volume mounts to our driver can still be started.

Let's revisit, as well as see what the other reviewers say, once I make the simplifications around the "openshift-*" naming scenario.

Thanks.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: yeah we'll see if any of the other assignees to this PR know ... thanks

gabemontero:
ok @jsafrane I have pushed a new commit that is mostly my attempt to address your comments that I immediately signed off on (the mutating one is still pending for me at least on some feedback from @deads2k based on what I think he has told me in the past)

There are also some minor tweaks based on my taking another pass at reinterpreting the EP template.

thanks

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
adambkaplan: Unfortunately I don't think there is a notion of "duck typing" APIs for webhooks - you need to specify the top level APIs to listen to. That would mean we watch things that we ship in OpenShift that reference pod volumes

- Pod itself
- Basic workload types (Deployment, StatefulSet, Job, CronJob, DaemonSet)
- More "inner" workload types (ReplicaSet, ReplicationController)?
- BuildConfig and Build

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+
+The discussion landed around these points:
+- A golang constant will declare the name of a k8s Condition that other components (i.e. OC Operators besides Shared Resources) will
+set on any Shared Resource that they pre-populate into OCP clusters.
+- REMINDER: the status structs of Shared Resources have an array of Conditions.  See 
+[here](https://github.com/openshift/api/blob/dbc82f2a3bc8b07e39628d36aee19d5a62847e2c/sharedresource/v1alpha1/types_shared_configmap.go#L83-L88) and
+[here](https://github.com/openshift/api/blob/dbc82f2a3bc8b07e39628d36aee19d5a62847e2c/sharedresource/v1alpha1/types_shared_secret.go#L81-L86)
+- The components who create pre-populated Shared Resources should have a controller that performs whatever validations
+they deem possible to verify the References, and those controllers are what will update the Status Condition arrays
+for the pre-populated Shared Resources
```
adambkaplan, line 132: It seems odd to me that a separate controller would need to update status. What status conditions will these controllers set?

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+will be used to inspect all created/modified Shared Resources, and make sure their names do not start
+with the reserved prefix "openshift-" unless their Shared Resource name and Reference key are found in a 
+well known list of OCP pre-populated Shared Resource.  That list will be in well known ConfigMaps that resides 
+in the managed CSI Driver namespace.  Components / operators other than Shared Resources that want to create pre-populated
+SharedSecrets or SharedConfigMaps should submit PRs against the Shared Resource CSI Driver Operator Repo to update the 
+corresponding ConfigMap with the necessary information.  This is akin to how we create in-payload image streams via the PRs against the 
+samples operator today.
+
+There will be a "ocp-sharedsecret-list" and "ocp-sharedconfigmap-list" ConfigMap.  The key of an entry is the Shared Resource
+name.  The value should be of the form "namespace:name" in order to capture.
```
adambkaplan, line 112: May help to provide an example here with the Insights operator's cluster entitlement:

```
kind: ConfigMap
apiVersion: v1
metadata:
  name: ocp-sharedsecret-list
  namespace: <namespace of the webhook>
data:
  openshift-etc-pki-entitlement: openshift-config-managed:etc-pki-entitlement
```

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+OCP must-gather process.
+
+
+### API Extensions
+
+Quick note:  When discussing API, 'Shared Resource' on its own == SharedConfigMap / SharedSecret, and 'Reference' == the corresponding
+Secret or ConfigMap.
+
+A [ValidatingAdmissionWebhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook)
+will be used to inspect all created/modified Shared Resources, and make sure their names do not start
```
adambkaplan, line 103: Inspecting object updates could be problematic. Let's say an item is _removed_ from our allowlist on upgrade. Then we could have a `SharedSecret` or `SharedConfigMap` that is valid on creation, but is invalid on update after the upgrade removes the .

I don't think we want to block object updates if it already exists.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+with the reserved prefix "openshift-" unless their Shared Resource name and Reference key are found in a 
+well known list of OCP pre-populated Shared Resource.  That list will be in well known ConfigMaps that resides 
+in the managed CSI Driver namespace.  Components / operators other than Shared Resources that want to create pre-populated
+SharedSecrets or SharedConfigMaps should submit PRs against the Shared Resource CSI Driver Operator Repo to update the 
+corresponding ConfigMap with the necessary information.  This is akin to how we create in-payload image streams via the PRs against the 
+samples operator today.
+
+There will be a "ocp-sharedsecret-list" and "ocp-sharedconfigmap-list" ConfigMap.  The key of an entry is the Shared Resource
```
adambkaplan, line 111: We should also note that these lists will be managed by the CVO. Users will not be able to change them without the CVO restoring the original allowlists.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+for the pre-populated Shared Resources
+
+At this time, neither the CSI Driver / Operator nor any Webhooks will react to external controllers/operators manipulating
+the "validity" condition.  See the alternatives section for potential considerations for doing so.  But human inspection,
+controller reaction via watches, as well as metrics/alerts, are all possible off of this condition.
+
+### Implementation Details/Notes/Constraints [optional]
+
+Generally, basic field checking of the API Object created or modified when the webhook is called should be sufficient
+for the 'readOnly: true' and 'openshif-*' restriction enforcement.
```
adambkaplan, line 141: nit:
```suggestion
for the 'readOnly: true' and 'openshift-*' restriction enforcement.
```

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: Yeah I need to do a better job of describing what @deads2k and I discussed.  I'll post a concrete example here (maybe I even put some form of that in the EP).  Assuming that concrete example helps clear things up for you @adambkaplan, I'll then move on to improving the EP itself.

So consider:
- the Insights Operator starts creating a SharedSecret to encapsulate the entitlement secret they maintain
- they seem to already being doing some "validations" around that secret 
- maybe then even have more in the works?
- maybe they agree to do some items that we or @deads2k suggest?
- if they discover something wrong with the entitlement secret, the insights operator updates the status of the SharedSecret they create for using it, setting true/false, reason, message on our well known "validity" condition, explaining someting is wrong
- Very similar to what CVO operators do with the well known conditions CVO defines for the ClusterOperator object
- then OpenShift Build development, we we get the case that says "entitled builds don't work", will ask them to check the conditions on the SharedSecret that Insights Operator creates
- the Reason/Message there helps the customer figure out what he needs to do to fix his entitlement credentials (maybe he has to log onto console.redhat.com and do something) and then his entitlement builds work

Does that help @adambkaplan ?  If it does, including perhaps you accepting the CVO operator analogy, do you think you are on board?

thanks

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: will do in next push

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: good point .... I'll reduce to create and take out the modify bit

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+
+So the 4.10 impact for Shared Resources on the API Server storage i.e. 'etcd' was deemed "no problem".
+
+#### New SLIs for the new admission validations to help cluster administrators and support
+
+Separate metrics for
+= SharedSecret creation/modification failures because of an openshift-* name.  Metric labels for share name, secret name, secret namespace
+- SharedConfigMap creation/modification failures because of an openshift-* name.  Metric labels for share name, configmap name, configmap namespace
+- SharedConfigMap/SharedSecret having the validity condition go to invalid.  Metric labels for share name, reference name and namespace
+- Pod specable creation/modification failure for Shared Resource CSI Volumes without 'readOnly: true'.  Metric labels for namespace, Group/Kind/Version, share name.
```
adambkaplan, line 265: We absolutely need a response time histogram for the webhook if we are reacting all pods, possibly with separate labels for accepted/rejected. If SD team gets a report that it is taking a long time for pods to be created, we need to see to rule out that this webhook was the cause. SLO would need to be something like "99% of all requests return a response within 1 second."

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: I am pretty sure I do that elsewhere, but will cross reference with this line, what the EP template says, and reiterate here in some form of fashion to make sure it is clear.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: cool thanks

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+
+Per open questions: since we are validating at both the webhook and controller (in case webhook is down), do they need separately
+named metrics (i.e. "..._webhook_total" and "..._controller_total") for the same logical items in the above list?
+
+Deferring on any metrics around associated Secret/ConfigMap validation until some of the open questions around that topic
+are answered.
+
+Alert 1:  using the `rate()` prometheus function, we can compare our failure counts above with existing metrics from 4.10
+wrt successful mounts, and build alerts where, for example, if failures / successes over 1 minute > 0.1, fire.  A pseudo-code
+example: `rate(read_only_failures[1m] / rate(successful_mount[1m]) > 0.1`.  Debating info only to moderate priority for the alert
```
adambkaplan, line 275: Is there a general metric for volume mounts succeeding/failing?

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: ok cool thanks for the confirmation

when we do get down to implementation, we should work with metrics team wrt a new metric vs. the generic webhook one I found

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: I have not found one.   I'll look again.  If still no luck, I'll add an open question, reach out to the storage team.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+
+Incorrectly preventing the create/update of Shared Resources at the moment appears to be relatively minor on the problem
+scale for cluster administrators and support.  But the error returned wil be directly available to the user.  That, along
+with the CLI invocation used, and any associated YAML for the object, should lead to straight forward diagnostics for 
+support.
+
+The enforcement of 'readOnly: true' in particular is worth discussing in this section.  It seems like very simple 
+pattern matching / field checking, but unanticipated bugs when performing this check
+in a ValidatingAdmissionWebhook could be impactful, as we are ultimately gating whether a Pod is allowed or not.  If 
+there is a bug which results in valid Pods being rejected, that could have broad and major consequences.
```
adambkaplan, line 326: If we are enforcing things at the `Pod` level, then ultimately every workload will run through this check.
This could potentially lead to undesirable behavior if the workload managing the pod does not gracefully handle repeated failed pod creation.

The same issue will apply to workload APIs that the webhook is not aware of. For instance, a Tekton TaskRun may create a faulty Pod and not report this error in a friendly manner.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+
+##### Detect Failure modes and Possible Symptoms
+
+- If the webhook is not running, kube-apiserver logs will show errors like "failed to call admission webhook xyz".
+- Cluster Storage Operator / CSI Shared Resource Operator will degrade with message "Failed to launch webhook server" and reason "WehhookServerFailed".
+- The k8s metric `webhook_admission_duration_seconds("<name of our new admission controller>", "validating", "put", "false")`
+    will show >1s latency and alert `WebhookAdmissionLatencyHigh` will fire.
+- Otherwise, we should be able to fall back on must gather.
+
+##### Disable the API extension (e.g. remove ValidatingWebhookConfiguration `<names of the our new webhook configs>`)
```
adambkaplan, line 350: Is this achievable - won't our operator deploy the webhook and reconcile the `ValidatingWebhookConfiguration` object?

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+- What consequences does it have for newly created workloads?
+
+For properly constructed Shared Resources and Pods with Shared Resource CSI Volume Mounts, incorrect specifications
+will still be denied, but the errors will not manifest on Pod creation, but when Pod containers are trying to start.
+
+## Implementation History
+
+Initial submission of this EP
+
+## Drawbacks
```
adambkaplan, line 374: Adding a webhook that watches all pod creation risks cluster stability.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+
+There is a 'csi-snapshot-webhook' instance of the 'ValidatingWebhookConfiguration' type in the 'openshift-cluster-storage-operator'
+namespace that actually employs a CVO based operator for its management.  Per confirmation with Jan/Storage team,
+management of any new webhooks for Shared Resources should be managed by the Shared Resources CSI Driver Operator, where
+that operator is managed by the Cluster Storage Operator.  Assets and manifests in those two repos will
+be updated as needed so that there is permission to create the webhook related API Objects.
+
+So we do not envision needing new repositories or a new operator for managing the webhook(s).
+
+The current core repository of [https://github.com/openshift/csi-driver-shared-resource](https://github.com/openshift/csi-driver-shared-resource)
```
adambkaplan, line 411: +1 to this

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,413 @@
...
+
+## Drawbacks
+
+- Customers prove inexplicably resistant to reserving "openshift-" naming prefixes for Shared Resources (though they seem to have accepted similar restrictions for namespaces)
+
+## Alternatives
+
+First, currently see no reasons to use the older method (like in the early 3.0 days) of an admission plugin inside openshift-apiserver
+for validation, like you see for older API like Builds, DeploymentConfigs, ImageStreams, and Templates, could be found.
+But of course any SME guidance around this is welcome.
```
adambkaplan, line 382: I think a reason to use an admission plugin instead of webhook is the fact that we will be validating pods. Validating pods requires a level of availability that we already guarantee with openshift-apiserver. Openshift-apiserver already has a mutating admission webhook for imagestreams, this would not be that much different.

When I was working on the draft KEP for CSI volume security, we went down the path of adding an admission plugin to k8s precisely because we were validating pods. We can use feature gates to enable/disable the admission plugin.

gabemontero: Right (forgot that precise element of the plugin/webhook comparison).  So a couple of follow ups @adambkaplan :
1) does it make sense to use validating webhooks for the SharedResource openshift-* name validation, or just use an admission plugin for both
2) I know we have examples of such admission plugins for the OpenShift API types in the OpenShift API server today, but I don't remember in the KEP CSI volume discussions, not that I was involved in all of them, if there is an example of OpenShift injecting one for a K8 type ... i.e. injecting one in the k8s API server.  Is that maybe how we prevent namespace creation from the command line with names starting with "openshift-" ?  Do you recall @adambkaplan 

I did not bring up this question with @deads2k during my exchange with him a few days ago.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: I mentioned that concern earlier but sure I can mention it here, though per the other thread, maybe we switch on that for the Pod validations.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: Well I'm actually just following the template, trying to address the questions it asks.  I just to converted a bullet point in the template to a heading for readability.

Can't seem to do line numbers when referencing lines in markdown files, but if you search for `disable the API extension (e.g. remove MutatingWebhookConfiguration xyz, remove APIService foo)` in https://github.com/openshift/enhancements/blob/db2ce78a020c2998096f7763bddbe108e40ce3ca/guidelines/enhancement_template.md you'll see the question.

As to "achievability", other than disabling the CSO (I imagine that has the same consequence support wise as disabling the CVO, but don't know that with certainty) is the only means for that I can think of.

Maybe for crazy critical support situations we will suggest that to customers to prevent clusters from going belly up ??

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: OK so I take it this comment is plus one'ing my concern @adambkaplan :-)

Your point about admission plugins vs. webhooks and getting back to the choice item that came up with the KEP are key.  Presumably with the reviewers here between you, me, @deads2k and @jsafrane will reach a final consensus.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: thanks for the additional confirmation @adambkaplan as well ascalling out duck typing, as that is what in part prompted my thinking ... I should have used that phrase as well (just forgot).

Maybe I'll add that phrase at some point, but more than likely I just remove this questions and state the reality.  That said, I should also replace this question with bubbling up the admission webhook vs. plugin question, per your other comments wrt the concerns you and I both voiced on the impact of a misbehaving webhook.

thanks again

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: I will be posting what I did find as part of an open question on this topic.... summary, "sort of" but it is only alpha level at this time https://kubernetes-csi.github.io/docs/volume-health-monitor.html

gabemontero:
@adambkaplan - first pass at updates based on your comments from last week
@jsafrane - I also added the "should we be mutating" item to the open questions

unless @deads2k manages to find time to look at this in the next few days, I'll try to reach out to him in slack on some of the open questions that are either a) I need a quick refresh on conversations a few months ago, or b) he is most like the most authoritative opinion

gabemontero:
although github per their status page is a bit compromised as I type, FYI I just pushed a new commit with metrics related updates per slack conversations I've had with the monitoring team and some of the reviewers here.

I don't see the new commit registered yet, but presumably it will show up eventually.  I'll monitor throughout the day and repush if need be.

gabemontero:
@adambkaplan @jsafrane @deads2k I've updated the EP based on the exchanged @deads2k and I had in https://coreos.slack.com/archives/CE4L0F143/p1647612099320099

With that exchanged, I've settled on the validating webhook implementation choice, along with adding needed details on the associated configuration I picked up form @deads2k.  As part of this, I pruned a good chunk of the open questions, and moved some of the comments here in the PR along those open questions to the alternatives section.

PTAL - thanks

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,486 @@
...
+
+So the 4.10 impact for Shared Resources on the API Server storage i.e. 'etcd' was deemed "no problem".
+
+#### New SLIs for the new admission validations to help cluster administrators and support
+
+Separate metrics for
+= SharedSecret creation/modification failures because of an openshift-* name.  Metric labels for share name, secret name, secret namespace
+- SharedConfigMap creation/modification failures because of an openshift-* name.  Metric labels for share name, configmap name, configmap namespace
+- SharedConfigMap/SharedSecret having the validity condition go to invalid.  Metric labels for share name, reference name and namespace
+- Pod specable creation/modification success/failure count for Shared Resource CSI Volumes without 'readOnly: true'.  Metric labels for namespace, Group/Kind/Version, share name,
```
adambkaplan, line 323: These metrics all have unbound cardinality. Is this a concern?

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,486 @@
...
+
+The controller's examination of the Validity condition in its existing Shared Resource reconciliation loop should 
+have minimal/negligible impact there.  And the controller's removal of data is a behavior whose path length is 
+already established in 4.10, as we remove data if the service account has its 'use' permissions on the Shared Resource
+revoked.
+
+#### Measuring / Verifying impact on existing API Server SLIs
+
+From existing metric discovery, what we will depend on:
+- the EP template example in "Support Procedures" of using `webhook_admission_duration_seconds("<name of our new admission controller>", "validating", "put", "false")`, with >1s latency resulting in the `WebhookAdmissionLatencyHigh` alert firing, appears to be the valid choice addressing our new webhooks' impact on the API Object create/modify flows.  Is that correct?
```
adambkaplan, line 361: This feels correct - if our admission controller has high latency, we fire an alert.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
adambkaplan: This might be the best option we have in terms of "emergency support":
- Scale the CVO deployment to 0
- Scale the cluster-storage-operator deployment to 0
- Remove the ValidatingAdmissionWebhook configuration.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,486 @@
...
+capabilities to help filter which Pods are addressed by our new `ValidatingAdmissionWebhook`.  Consider the following 
+example:
+
+```yaml
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+...
+webhooks:
+  - name: podsharedresourcereadonlywebhook.openshift.com
+    objectSelector:
```
adambkaplan, line 152: Be sure to add the `failurePolicy` - any timeout/500 errors should be ignored ("fail open")
```suggestion
    failurePolicy: Ignore # Do not fail object creation if this webhook times out or returns 5xx errors
    objectSelector:
```

See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#validatingwebhook-v1-admissionregistration-k8s-io

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,486 @@
...
+
+### Goals
+
+- Ensure names of pre-supplied SharedSecrets and SharedConfigMaps by OCP components will not conflict with any SharedSecrets or SharedConfigMaps created by users.
+- Whenever possible prevent unsupported use of read-write volumes, as well as any future restrictions around the CSI Volume API or Shared Resource API
+- Provide pluggable means of validation of the associated Secrets and ConfigMaps of SharedSecrets and SharedConfigMaps.
+
+And provide validation failures on API object creation/modification for each of these whenever possible.
+
+### Non-Goals
```
adambkaplan, line 59: Perhaps add non-goals around the "is the underlying object valid" feature.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,486 @@
...
+    objectSelector:
+      matchExpressions:
+        - {key: openshift.io/run-level, operator: NotIn, values: ["0", "1"]} 
+        - {key: openshift.io/exempt-from-shared-resource-readonlytrue-webhook, operator: NotIn, values: ["true"]}
+    rules:
+      - operations: [ "CREATE" ]
+        apiGroups: []
+        apiVersions: []
+        resources: [ "pods", "deployments", "daemonsets", ....]
+        scope: "*"
```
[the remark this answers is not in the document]
gabemontero: good catch @adambkaplan thanks

my next push will include:
- I've split out apps from pods in my pseudo yaml here
- I also added openshift deploymentconfig as a separate item; yes, they are not "preferred", but they still exist and are used
- I added an open question for your query to @soltysh in case he does not get to it before I push the update
-  I added him as a reviewer in the doc
- I'll assign him to the PR when I've pushed the changes

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,486 @@
...
+    ...
+```
+
+Let's dive into the 2 different label
+- The first `matchExpression` is the means of preventing the use of our new webhook for any Pods of the core OCP control pane.  Because of this
+label selector we will filter out components like [the K8s API Server Operator Namespace manifest](https://github.com/openshift/cluster-kube-apiserver-operator/blob/d80b406ec4844328e04253f9ed5aec3f4a29834a/manifests/0000_20_kube-apiserver-operator_00_namespace.yaml#L11)
+- The second `matchExpression` is the means of preventing the use of our new webhook for any Pods in other OCP namespaces we want to exclude.
+Initially, we are thinking only the [CSI Drivers namespace](https://github.com/openshift/cluster-storage-operator/blob/2a3fbb68afd066d16582b09f9fd834c7fd03457e/manifests/02_csi_driver_operators_namespace.yaml) 
+will be excluded (as that is where we will also host our webhook), so
```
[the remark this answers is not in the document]
gabemontero: it is more the  latter, but to be clear, with the use if `NotIn`, we would be validating by default, except for pods for namespaces in the `NotIn` list .... for the detailed walk through, revisit the thread between David and me at  https://coreos.slack.com/archives/CE4L0F143/p1647612099320099 and in particular the existing example he pointed me to https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_20_kube-apiserver-operator_00_namespace.yaml#L11 and his comments https://coreos.slack.com/archives/CE4L0F143/p1647615292929869?thread_ts=1647612099.320099&cid=CE4L0F143

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: sure, I took the fork in the road around "optimal debuggability" from a support perspective, but do not feel super strong about it either.

But I'm fine with removing the labels for each of them, and making them global in nature.  Still has value in that alerts based of them growing will trigger the need for more debug.

I'll proceed removing those labels @adambkaplan but certainly chime in with any "meet in the middle labeling" you like.

gabemontero:
updates for you comments pushed @adambkaplan 

/assign @soltysh

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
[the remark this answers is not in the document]
adambkaplan: I found it - the admission controller config has a `namespaceSelector` option. This is how we ignore pods in these namespaces: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,532 @@
...
+capabilities to help filter which Pods are addressed by our new `ValidatingAdmissionWebhook`.  Consider the following 
+example:
+
+```yaml
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+...
+webhooks:
+  - name: pod.csi.sharedresource.openshift.com
+    failurePolicy: Ignore
```
adambkaplan, line 156: Also add `namespaceSelector:`:

```suggestion
    failurePolicy: Ignore
    namespaceSelector:
      matchExpressions:
        - key: runlevel
          operator: NotIn
          values: ["0", "1"]
        - key: openshift.io/run-level
          operator: NotIn
          values: ["0", "1"]
```

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,532 @@
...
+    objectSelector:
+      matchExpressions:
+        - {key: openshift.io/run-level, operator: NotIn, values: ["0", "1"]} 
+        - {key: csi.sharedresource.openshift.io/skip-validation, operator: NotIn, values: ["true"]}
```
adambkaplan, line 160: Drop the run-level match, covered by the namespace selector.

```suggestion
      matchExpressions:
        - {key: csi.sharedresource.openshift.io/skip-validation, operator: NotIn, values: ["true"]}
```

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,532 @@
...
+On the associated `ValidatingWebhookConfiguration`, per guidance from David Eads, which should utilize the label selector
+capabilities to help filter which Pods are addressed by our new `ValidatingAdmissionWebhook`.  Consider the following 
+example:
+
+```yaml
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+...
+webhooks:
+  - name: pod.csi.sharedresource.openshift.com
```
adambkaplan, line 155: nit - we use .io TLD within OpenShift
```suggestion
  - name: pod.csi.sharedresource.openshift.io
```

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
adambkaplan: Now that I see the full impact of my idea above, I'm less convinced adding something to `spec` is a good idea. Leaving `Ready` as an unmanaged condition that the CSI driver checks is perhaps the best option. We would want to document this in the status API.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
gabemontero: actually I'm moving the skip-validation item to the namespace selector as well, as that was my intent:  allow for marking of  entire namespaces to be skipped, independent of the run level stuff.

gabemontero:
updates from you comments yesterday pushed @adambkaplan

adambkaplan (approved):
/approve

Hitting approve on this EP. At this point we have vetted the configuration of the webhook and the scope of changes to be made. I'm ready to lgtm once the commits have been squashed.

enhancements/cluster-scope-secret-volumes/shared-resource-validation.md
```diff
@@ -0,0 +1,524 @@
...
+Given the controller and webhook will be in separate Pods and have separate metric service endpoints registering metrics
+with prometheus, that is the implementation reality.  Any alerts defined can take into account both _webhook and _controller
+metrics of the same logical significance.
+
+Alerts:  using the `rate()` prometheus function, we can compare our new metrics above along with existing metrics from 4.10
+wrt invalid names and failed/successful mounts.  We'll define alerts where, for example, if failures / successes over 1 minute > 0.1, fire.  A pseudo-code
+example: `rate(read_only_failures[1m] / rate(successful_mount[1m]) > 0.1`.  Debating info only to moderate priority for the alert.
+We'll finalize during implementation and update EP accordingly.
+
+Alert 2: if any Shared Resources have an invalid Reference.  Info level priority for user created Shared Resources.  Considering
```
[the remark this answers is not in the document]
gabemontero: we can debate during implementation

gabemontero:
last suggestion pulled in and commits squashed @adambkaplan

adambkaplan (approved):
/lgtm

PR #1057, opened 2022-03-11
Enhancement for enabling 4-node cluster deployments as a day0 operation

flaper87:
Signed-off-by: Flavio Percoco <flavio@redhat.com>

### DIFF
--- a/enhancements/installer/compact-plus-worker.md
+++ b/enhancements/installer/compact-plus-worker.md
@@ -59,9 +59,9 @@ List the specific goals of the proposal. How will we know that this has succeede
 
 ## Proposal
 
-- Set the InfrastructureTopology to have the same value as the
-  ControlPlaneTopology when the number of workers is 1, instead of setting its
-  value to SingleReplica as it's done today.
+- Set the `InfrastructureTopology` to have the same value as the
+  `ControlPlaneTopology` when the number of workers is 1, instead of setting its
+  value to `SingleReplica` as it's done today.
 
 - Set masters as schedulable when workers are less or equal to 1. Today,
   masters are set to schedulable only when there are no workers available.
--- a/enhancements/installer/compact-plus-worker.md
+++ b/enhancements/installer/compact-plus-worker.md
@@ -21,7 +21,7 @@ superseded-by: []
 
 ## Summary
 
-This enhancement proposes to allow for deploying a 4-nodes cluster as a day 0
+This enhancement proposes to allow for deploying a 4-node cluster as a day 0
 operation, where the infrastructure workload can be deployed with HA enabled.
 The 4-nodes cluster would be composed by 3 schedulable master nodes and a
 single worker, which provide users with a highly available cluster.
@@ -40,7 +40,7 @@ topology, which won't allow the various services to run as HA.
 To achieve the desired topology described in this enhancement,OpenShift cluster
 administrators can only deploy a 4-nodes clusters by first deploying a compact
 cluster (3 schedulable masters) and then adding a host as a day 2 operation.
-Another option is for them to This provides the same results as aimed in this
+Another option is for them to <missing text here> This provides the same results as aimed in this
 enhancement but with a poor and slower user experience.
 
 Today, because of the current limitations in the installer, it takes
@@ -65,8 +65,8 @@ List the specific goals of the proposal. How will we know that this has succeede
 
 ## Proposal
 
-- Set the `InfrastructureTopology` to have the same value as the
-  `ControlPlaneTopology` when the number of workers is 1, instead of setting its
+- Set the `infrastructureTopology` to have the same value as the
+  `controlPlaneTopology` when the number of workers is 1, instead of setting its
   value to `SingleReplica` as it's done today.
 
 - Set masters as schedulable when workers are less or equal to 1. Today,
@@ -81,7 +81,7 @@ List the specific goals of the proposal. How will we know that this has succeede
 
 ### API Extensions
 
-- Sets the `infrastructureTopology` field to the value of`ControlPlaneTopology` in the
+- Sets the `infrastructureTopology` field to the value of `controlPlaneTopology` in the
   `config.openshift.io/v1/infrastructure` CR called `cluster`, which is created
   during the install process, when the number of workers is 1.
 
--- a/enhancements/installer/compact-plus-worker.md
+++ b/enhancements/installer/compact-plus-worker.md
@@ -63,8 +63,8 @@ List the specific goals of the proposal. How will we know that this has succeede
   ControlPlaneTopology when the number of workers is 1, instead of setting its
   value to SingleReplica as it's done today.
 
-- Use the `ControlPlaneTopology` topology when the number of workers is 1,
-  instead of the `SingleReplicaTopologyMode` as it's done today.
+- Set masters as schedulable when workers are less or equal to 1. Today,
+  masters are set to schedulable only when there are no workers available.
 
 ### User Stories
 
--- a/enhancements/installer/compact-plus-worker.md
+++ b/enhancements/installer/compact-plus-worker.md
@@ -168,7 +168,7 @@ masters as well as other workloads. Depending on the amount of traffic, and the
 resources available to the cluster, this may result in degraded functionality
 in the masters, which finally affects the general health of the cluster.
 
-With this enhancement, we will be making it easier for users to create such a
+With this enhancement, we will be making it easier for users to create such
 4-nodes cluster deployments, which could make users believe it is supported or
 recommended.
 
--- a/enhancements/installer/compact-plus-worker.md
+++ b/enhancements/installer/compact-plus-worker.md
@@ -1,0 +1,188 @@
+---
+title: Compact + Worker deployment as day0 operation
+authors:
+  - "@flaper87"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+api-approvers: # in case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers)
+  - TBD
+creation-date: 2022-03-08
+last-updated: 2022-03-08
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - TBD
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Compact + Worker deployment as day0 operation
+
+## Summary
+
+This enhancement proposes to allow for deploying a 4-nodes cluster as a day 0
+operation. The 4-nodes cluster would be composed by 3 schedulable master nodes
+and a single worker, which provide users with a highly available cluster.
+
+## Motivation
+
+There are environments where only 4 nodes are available where 4-nodes clusters
+are being deployed. We could serve OpenShift cluster administrators of these
+environments better by allowing them to deploy a 4-nodes cluster as a day0
+operation.
+
+Today, in such environments, OpenShift cluster administrators can only deploy a
+4-nodes clusters by first deploying a compact cluster (3 schedulable masters)
+and then adding a host as a day 2 operation. This provides the same results as
+aimed in this enhancement but with a poor and slower user experience.
+
+Today, because of the current limitations in the installer, it takes
+administrators significantly longer to achieve the desired topology, when only
+4 nodes are available. Furthermore, by requiring more steps, we are making the
+process more prone to user errors.
+
+### Goals
+
+List the specific goals of the proposal. How will we know that this has succeeded?
+
+- Enable the installer to deploy a 4-nodes cluster composed by 3 schedulable
+  masters and one worker.
+
+### Non-Goals
+
+- Define new infrastructure topologies
+
+- Define new architectures
+
+- Provide official support for the 4-nodes cluster deployments
+
+## Proposal
+
+- Set the InfrastructureTopology to have the same value as the
+  ControlPlaneTopology when the number of workers is 1, instead of setting its
+  value to SingleReplica as it's done today.
+
+- Use the `ControlPlaneTopology` topology when the number of workers is 1,
+  instead of the `SingleReplicaTopologyMode` as it's done today.
+
+### User Stories
+
+- As an OpenShift cluster administrator, I want to be able to install a 4-nodes
+  cluster as a single installation step so that my cluster will be able to
+  handle workloads that require replication.
+
+
+### API Extensions
+
+- Sets the `infrastructureTopology` field to the value of`ControlPlaneTopology` in the
+  `config.openshift.io/v1/infrastructure` CR called `cluster`, which is created
+  during the install process, when the number of workers is 1.
+
+- Sets the `mastersSchedulable` field to true in the
+  `config.openshift.io/v1/scheduler` CR called `cluster`, which is created
+  during the install process, when the number of workers is 1.
+
+### Implementation Details/Notes/Constraints [optional]
+
+What are the caveats to the implementation? What are some important details that
+didn't come across above. Go in to as much detail as necessary here. This might
+be a good place to talk about core concepts and how they relate.
+
+### Risks and Mitigations
+
+This should not introduce newer risks other than the existing ones in a compact
+cluster deployment.
+
+## Design Details
+
+### Open Questions [optional]
+
+Nothing at the moment
+
+### Test Plan
+
+
+- Add unit tests to ensure the Infrastructure and Scheduler settings are properly set for the different combinations
+
+- Add a periodic job (nightly?) job to run OpenShift tests on a 4-node cluster
+
+### Graduation Criteria
+
+**Note:** *Section not required until targeted at a release.*
+
+TBD
+
+#### Dev Preview -> Tech Preview
+
+TBD
+
+#### Tech Preview -> GA
+
+TBD
+
+#### Removing a deprecated feature
+
+TBD
+
+### Upgrade / Downgrade Strategy
+
+This enhancement does not have an impact on existing clusters as it's intended
+to affect only day0 operations.
+
+### Version Skew Strategy
+
+Does not apply
+
+### Operational Aspects of API Extensions
+
+
+- Allows for a 4-nodes, highly available, cluster to be deployed as a day0
+  task. It reduces the time required for such deployments to be created.
+
+- Does not impact general deployment functionalities in a negative way.
+
+- Allowing workloads on the master nodes may affect the master's health if the
+  traffic and workloads are too high.
+
+#### Failure Modes
+
+Same as a compact cluster
+
+#### Support Procedures
+
+- The same support procedures used for a compact cluster should be used for a
+  4-nodes cluster. I would appreciate help filling in this section with more
+  details.
+
+
+## Implementation History
+
+N/A
+
+## Drawbacks
+
+A 4-nodes cluster (a.k.a 3+1) is generally not a recommended architecture. By
+having schedulable masters users will be routing ingress traffic through the
+masters as well as other workloads. Depending on the amount of traffic, and the
+resources available to the cluster, this may result in degraded functionality
+in the masters, which finally affects the general health of the cluster.
+
+With this enhancement, we will be making it easier for users to create such a
+4-nodes cluster deployments, which could make users believe it is supported or
+recommended.
+
+## Alternatives
+
+- Don't set masters as schedulable automatically. Instead, cluster
+  administrators will have to explicitly set masters as schedulable in the
+  deployment configuration. The rest of the proposal would remain the same,
+  switching the cluster to `ControlPlaneTopology` if there are 3 schedulable
+  masters and 1 worker.
+
+- Don't do anything and force users to do this in a 2 steps operation: first
+  deploy a compact cluster and then add a worker node.
+
+## Infrastructure Needed [optional]
+
+N/A
--- a/enhancements/installer/compact-plus-worker.md
+++ b/enhancements/installer/compact-plus-worker.md
@@ -22,20 +22,26 @@ superseded-by: []
 ## Summary
 
 This enhancement proposes to allow for deploying a 4-nodes cluster as a day 0
-operation. The 4-nodes cluster would be composed by 3 schedulable master nodes
-and a single worker, which provide users with a highly available cluster.
+operation, where the infrastructure workload can be deployed with HA enabled.
+The 4-nodes cluster would be composed by 3 schedulable master nodes and a
+single worker, which provide users with a highly available cluster.
 
 ## Motivation
 
 There are environments where only 4 nodes are available where 4-nodes clusters
 are being deployed. We could serve OpenShift cluster administrators of these
 environments better by allowing them to deploy a 4-nodes cluster as a day0
-operation.
+operation and make sure their infrastructure workload is highly available.
 
-Today, in such environments, OpenShift cluster administrators can only deploy a
-4-nodes clusters by first deploying a compact cluster (3 schedulable masters)
-and then adding a host as a day 2 operation. This provides the same results as
-aimed in this enhancement but with a poor and slower user experience.
+Today, it is possible to deploy a 4-nodes cluster. The resulting cluster will
+be functional but the infrastructure workload will be run as a SingleReplica
+topology, which won't allow the various services to run as HA.
+
+To achieve the desired topology described in this enhancement,OpenShift cluster
+administrators can only deploy a 4-nodes clusters by first deploying a compact
+cluster (3 schedulable masters) and then adding a host as a day 2 operation.
+Another option is for them to This provides the same results as aimed in this
+enhancement but with a poor and slower user experience.
 
 Today, because of the current limitations in the installer, it takes
 administrators significantly longer to achieve the desired topology, when only


enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,188 @@
...
+
+- Define new architectures
+
+- Provide official support for the 4-nodes cluster deployments
+
+## Proposal
+
+- Set the InfrastructureTopology to have the same value as the
+  ControlPlaneTopology when the number of workers is 1, instead of setting its
+  value to SingleReplica as it's done today.
```
omertuc, line 64:
```suggestion
- Set the `InfrastructureTopology` to have the same value as the
  `ControlPlaneTopology` when the number of workers is 1, instead of setting its
  value to `SingleReplica` as it's done today.
```

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,188 @@
...
+
+- Use the `ControlPlaneTopology` topology when the number of workers is 1,
```
omertuc, line 66: This looks like a duplicate of the first bullet?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Drawbacks
+
+A 4-nodes cluster (a.k.a 3+1) is generally not a recommended architecture. By
+having schedulable masters users will be routing ingress traffic through the
+masters as well as other workloads. Depending on the amount of traffic, and the
+resources available to the cluster, this may result in degraded functionality
+in the masters, which finally affects the general health of the cluster.
+
+With this enhancement, we will be making it easier for users to create such a
```
omertuc, line 171:
```suggestion
With this enhancement, we will be making it easier for users to create such
```

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+resources available to the cluster, this may result in degraded functionality
+in the masters, which finally affects the general health of the cluster.
+
+With this enhancement, we will be making it easier for users to create such
+4-nodes cluster deployments, which could make users believe it is supported or
+recommended.
+
+## Alternatives
+
+- Don't set masters as schedulable automatically. Instead, cluster
```
flaper87, line 183: I think it may be worth it to elaborate on the pros/cons of this alternate proposal. 

One of the pros is certainly that it will let the user be explicit about their desired control plane topology. However, this may result in the user still being able to deploy a 3+1 topology like today where the infrastructure topology is SingleReplica. 

Maybe some questions to add to the open questions section could be:

- Do we want to protect the user from having infra workloads that are not HA?
- Do we want the user to decide whether their infra workload should be HA or not?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+  details.
+
+
+## Implementation History
+
+N/A
+
+## Drawbacks
+
+A 4-nodes cluster (a.k.a 3+1) is generally not a recommended architecture. By
```
dhellmann, line 171: If this isn't a recommended configuration, we should have some more justification for supporting it explicitly than "users asked for it." Can you add more details about the use case(s) above?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+## Alternatives
+
+- Don't set masters as schedulable automatically. Instead, cluster
+  administrators will have to explicitly set masters as schedulable in the
+  deployment configuration. The rest of the proposal would remain the same,
+  switching the cluster to `ControlPlaneTopology` if there are 3 schedulable
+  masters and 1 worker.
+
+- Don't do anything and force users to do this in a 2 steps operation: first
+  deploy a compact cluster and then add a worker node.
```
dhellmann, line 190: Is there no way to do this just by adding/editing manifests used by the installer? So, not in `install-config` but in a separate manifest?

enhancements/installer/compact-plus-worker.md
dhellmann: Removing the automatic scheduling behavior would also be a backwards-incompatible change, right?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+administrators can only deploy a 4-nodes clusters by first deploying a compact
+cluster (3 schedulable masters) and then adding a host as a day 2 operation.
+Another option is for them to This provides the same results as aimed in this
+enhancement but with a poor and slower user experience.
+
+Today, because of the current limitations in the installer, it takes
+administrators significantly longer to achieve the desired topology, when only
+4 nodes are available. Furthermore, by requiring more steps, we are making the
+process more prone to user errors.
+
```
dhellmann, line 50: Does it make any difference what deployment approach or platform is being used?

dhellmann (commented):
Is there any overlap between the topology changes proposed here and the changes in https://github.com/openshift/enhancements/pull/1041 for changing how ingress is scheduled? If we did #1041, would it help with this?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+- Enable the installer to deploy a 4-nodes cluster composed by 3 schedulable
+  masters and one worker.
+
+### Non-Goals
+
+- Define new infrastructure topologies
+
+- Define new architectures
+
+- Provide official support for the 4-nodes cluster deployments
```
sdodson, line 64: Then why are we doing this? Could the same not be achieved by having compute pool of size 1, modifying the manifests to make masters schedulable?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Proposal
+
+- Set the `InfrastructureTopology` to have the same value as the
+  `ControlPlaneTopology` when the number of workers is 1, instead of setting its
+  value to `SingleReplica` as it's done today.
+
+- Set masters as schedulable when workers are less or equal to 1. Today,
+  masters are set to schedulable only when there are no workers available.
```
sdodson, line 73: What or who will these things?

enhancements/installer/compact-plus-worker.md
sdodson: you can certainly do this by modifying manifests.

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+operation and make sure their infrastructure workload is highly available.
+
+Today, it is possible to deploy a 4-nodes cluster. The resulting cluster will
+be functional but the infrastructure workload will be run as a SingleReplica
+topology, which won't allow the various services to run as HA.
+
+To achieve the desired topology described in this enhancement,OpenShift cluster
+administrators can only deploy a 4-nodes clusters by first deploying a compact
+cluster (3 schedulable masters) and then adding a host as a day 2 operation.
+Another option is for them to This provides the same results as aimed in this
```
staebler, line 43: What is the other option? It is missing.
```suggestion
Another option is for them to <missing text here> This provides the same results as aimed in this
```

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+enhancement but with a poor and slower user experience.
+
+Today, because of the current limitations in the installer, it takes
+administrators significantly longer to achieve the desired topology, when only
```
staebler, line 47: Significantly longer? Is this longer in terms of the time that an administrator has to be actively doing something? Or is this longer in terms of wall clock time before an administrator has the cluster configured in the way that they want?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+4 nodes are available. Furthermore, by requiring more steps, we are making the
+process more prone to user errors.
+
+### Goals
+
+List the specific goals of the proposal. How will we know that this has succeeded?
+
+- Enable the installer to deploy a 4-nodes cluster composed by 3 schedulable
```
staebler, line 55: This goal needs to be refined. This goal is already achieved in the current state of OpenShift. Should there be something in here about HA infrastructure components?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+  masters and one worker.
+
+### Non-Goals
+
+- Define new infrastructure topologies
```
staebler, line 60: Why is this a non-goal? I could see this as a considered and rejected solution. But this is not really a goal.

enhancements/installer/compact-plus-worker.md
staebler: I agree that this non-goal seems at direct odds with goal of the EP.

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+
+- Provide official support for the 4-nodes cluster deployments
+
+## Proposal
+
+- Set the `InfrastructureTopology` to have the same value as the
+  `ControlPlaneTopology` when the number of workers is 1, instead of setting its
+  value to `SingleReplica` as it's done today.
+
+- Set masters as schedulable when workers are less or equal to 1. Today,
```
staebler, line 72: Are there any use cases for a cluster where there is actually only one worker and that workload are only schedulable to that one worker and not the masters?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Compact + Worker deployment as day0 operation
+
+## Summary
+
+This enhancement proposes to allow for deploying a 4-nodes cluster as a day 0
```
staebler, line 24: Throughout the document, replace "4-nodes cluster" with "4-node cluster".
```suggestion
This enhancement proposes to allow for deploying a 4-node cluster as a day 0
```

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+- Set the `InfrastructureTopology` to have the same value as the
+  `ControlPlaneTopology` when the number of workers is 1, instead of setting its
+  value to `SingleReplica` as it's done today.
+
+- Set masters as schedulable when workers are less or equal to 1. Today,
+  masters are set to schedulable only when there are no workers available.
+
+### User Stories
+
+- As an OpenShift cluster administrator, I want to be able to install a 4-nodes
```
staebler, line 77: How many customers want to install 4-node cluster? Of those, how many want workloads on the masters?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+  cluster as a single installation step so that my cluster will be able to
+  handle workloads that require replication.
+
+
+### API Extensions
+
+- Sets the `infrastructureTopology` field to the value of`ControlPlaneTopology` in the
```
staebler, line 84:
```suggestion
- Sets the `infrastructureTopology` field to the value of `controlPlaneTopology` in the
```

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+- Define new infrastructure topologies
+
+- Define new architectures
+
+- Provide official support for the 4-nodes cluster deployments
+
+## Proposal
+
+- Set the `InfrastructureTopology` to have the same value as the
+  `ControlPlaneTopology` when the number of workers is 1, instead of setting its
```
staebler, line 69:
```suggestion
- Set the `infrastructureTopology` to have the same value as the
  `controlPlaneTopology` when the number of workers is 1, instead of setting its
```

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+
+- Define new infrastructure topologies
+
+- Define new architectures
+
+- Provide official support for the 4-nodes cluster deployments
+
+## Proposal
+
+- Set the `InfrastructureTopology` to have the same value as the
```
staebler, line 68: Shouldn't we instead always set the `infrastructureTopology` field to "HighlyAvailable" when the replica count for workers is 1? Even when the replica count for masters is 1, there are 2 nodes for infrastructure--namely the one master and the one worker.

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+single worker, which provide users with a highly available cluster.
+
+## Motivation
+
+There are environments where only 4 nodes are available where 4-nodes clusters
+are being deployed. We could serve OpenShift cluster administrators of these
+environments better by allowing them to deploy a 4-nodes cluster as a day0
+operation and make sure their infrastructure workload is highly available.
+
+Today, it is possible to deploy a 4-nodes cluster. The resulting cluster will
```
staebler, line 36: Could you talk some about what services would not be run as HA and what the implications of that are on a cluster with only a single worker node?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+TBD
+
+### Upgrade / Downgrade Strategy
+
+This enhancement does not have an impact on existing clusters as it's intended
+to affect only day0 operations.
+
+### Version Skew Strategy
+
+Does not apply
```
staebler, line 141: Is this the proper place to discuss that a user that does an install with a given install-config.yaml will get a quite different cluster depending on whether the OpenShift version targeted makes their masters schedulable or not?

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+## Drawbacks
+
+A 4-nodes cluster (a.k.a 3+1) is generally not a recommended architecture. By
+having schedulable masters users will be routing ingress traffic through the
+masters as well as other workloads. Depending on the amount of traffic, and the
+resources available to the cluster, this may result in degraded functionality
+in the masters, which finally affects the general health of the cluster.
+
+With this enhancement, we will be making it easier for users to create such
+4-nodes cluster deployments, which could make users believe it is supported or
```
staebler, line 178: If it is not supported, then why are we adding ways to explicitly make it easier to do?

enhancements/installer/compact-plus-worker.md
romfreiman: It becomes 2 step, rather than 1. 
Timewise, it's basically a delta between installing a clusters and adding a worker to it - of course, on BM, it takes longer  - server boot time - 3 reboots if I recall that in case of single installation will happen almost in parallel (1 reboot in parallel to the control plane, and 2 only after the api moves to the control plane to server the ignition.

enhancements/installer/compact-plus-worker.md
romfreiman: I agree

enhancements/installer/compact-plus-worker.md
romfreiman: I doubt. I would say either all nodes are schedulable or only workers.

enhancements/installer/compact-plus-worker.md
romfreiman: Our whole ZTPFW is based on 4 nodes appliance

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+reviewers:
+  - TBD
+approvers:
+  - TBD
+api-approvers: # in case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers)
+  - TBD
+creation-date: 2022-03-08
+last-updated: 2022-03-08
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - TBD
```
dhellmann, line 14: Is there a Jira ticket with more background for this work?

enhancements/installer/compact-plus-worker.md
dhellmann: I think we're currently doing validation to prevent this configuration via IPI deployments on all platforms, right? So are we talking about removing that validation and doing something different? Or is this limited to the metal ZTP deployment scenarios?

enhancements/installer/compact-plus-worker.md
dhellmann: That's just the initial customer's use case, right? It will eventually support other configurations?

enhancements/installer/compact-plus-worker.md
staebler: My scenario fits under your two options. It is the second. Only workers are schedulable.

enhancements/installer/compact-plus-worker.md
staebler: Can we add this to the use cases with specifics of why they need or want a 4-node cluster?

enhancements/installer/compact-plus-worker.md
flaper87: I am not aware of scenarios where only one worker is schedulable. For a 4-nodes cluster, then all available nodes are schedulable.

Let me know if I misunderstood your question :)

enhancements/installer/compact-plus-worker.md
flaper87: > That's just the initial customer's use case, right? It will eventually support other configurations?

It's a specific partner's use-case, at least for now. More may come with the same needs and, yes, there are others that want a different configuration already.

@staebler yes, I will add it. :+1:

enhancements/installer/compact-plus-worker.md
flaper87: I will rephrase this because I don't think saying "it is generally not recommended" is fair here. IT's not recommended for specific cases due to the motivations I mentioned in this paragraph. I do think it can be ok for other scenarios. 

I will add more details

enhancements/installer/compact-plus-worker.md
flaper87: Goes back to the point I made in the `non-goals` section. To be honest, it's unclear if this architecture is really supported or not. 

The installer won't let you do  it in one go due to the issues mentioned in this enhancement but users can still get a 4-nodes cluster in two steps. It's unclear where the "supportability" of a topology is defined

enhancements/installer/compact-plus-worker.md
flaper87: > Removing the automatic scheduling behavior would also be a backwards-incompatible change, right?

I meant that we could keep the current behavior (make masters schedulable if there are no workers) and not expand as proposed in this enhancement.

flaper87 (commented):
> Is there any overlap between the topology changes proposed here and the changes in #1041 for changing how ingress is scheduled? If we did #1041, would it help with this?

It may help with the scheduling of the ingresses but it won't solve the problem for other infrastructure operators that need to be replicated.

enhancements/installer/compact-plus-worker.md
hardys: @dhellmann I actually don't think we're doing validation to prevent this (I just tried it for baremetal IPI and the installer accepted 1 worker replica)

So I think all we're talking about is extending the existing [scheduler manifest](https://github.com/openshift/installer/blob/master/pkg/asset/manifests/scheduler.go#L74) special-case so the controlplane nodes are schedulable when zero or 1 compute replicas are specified.

If we think this is specific to ZTP then it may be reasonable to just modify the installer-generated manifests (and this perhaps limits the test impact, e.g we probably don't want to test 3+1 for every platform in CI?)

enhancements/installer/compact-plus-worker.md
staebler: After you run `openshift-install create manifests --dir $INST_DIR`, there will be worker machineset manifests under `$INST_DIR/openshift` where you can change the number of worker replicas desired.

enhancements/installer/compact-plus-worker.md
staebler: > For example, a user could achieve a 4 nodes cluster in 2 steps today. Does it make a 4 nodes cluster officially supported? Probably not.

I don't know why a 4-node cluster where the user added nodes after installation would not be supported. The user is certainly permitted to add and remove worker nodes post-installation.

enhancements/installer/compact-plus-worker.md
staebler: I suppose what I am caught up on is the "significantly" part. Can we remove that color unless we are providing specifics to define what we deem "significant"?

enhancements/installer/compact-plus-worker.md
staebler: That is a detail of the proposed solution and not a non-goal. I could see where you would perhaps want a goal to be that we do not want to add any more infrastructure topologies. But I would question the reasoning behind that, too.

enhancements/installer/compact-plus-worker.md
staebler: I just want to make really sure that we are not going to be taking away functionality from a subset of users by changing the default behavior.

enhancements/installer/compact-plus-worker.md
hardys: Or you can specify a single worker replica in the install-config then modify `manifests/cluster-scheduler-02-config.yml` after running `create manifests` IIRC

enhancements/installer/compact-plus-worker.md
hardys: Perhaps it's worth considering test coverage - we have coverage of compact (3+0) and HA (3+2) clusters, do we have any coverage of the 3+1 case anywhere?  (I think the answer is no, but we may want to change that if we want to consider this configuration fully supported, or do we expect to only support/test in the context of ZTP?)

enhancements/installer/compact-plus-worker.md
dhellmann: Ah, perhaps that validation was removed or changed when we added support for compact clusters. Or maybe I'm imagining it.

enhancements/installer/compact-plus-worker.md
dhellmann: Adding those links for context would be good.

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
...
+### API Extensions
+
+- Sets the `infrastructureTopology` field to the value of `controlPlaneTopology` in the
+  `config.openshift.io/v1/infrastructure` CR called `cluster`, which is created
+  during the install process, when the number of workers is 1.
+
+- Sets the `mastersSchedulable` field to true in the
+  `config.openshift.io/v1/scheduler` CR called `cluster`, which is created
+  during the install process, when the number of workers is 1.
+
```
dhellmann, line 91: I pointed out the relationship to #1041 in an earlier comment. There the proposal is to add a new API field to control how ingress is deployed because even in an environment where we could run 2 controllers (both nodes are schedulable), we want to keep ingress coming through the control plane node.

In this proposal, instead of saying the infrastructureTopology is the same as the controlPlaneTopology, we could say that rather than counting only workers the infrastructureTopology setting should be based on the _total number of nodes on which infrastructure pods could be scheduled_ (including the control plane, when those nodes are schedulable).

In a SNO configuration, the field would be set to singleReplica, as it is today.

In a SNO+1 configuration, after this change the field would be set to HA. If we accept the API proposed in #1041, we would keep ingress off of the worker node, but other infrastructure pods could run there. Is that desirable?

In a compact cluster, both fields would be set to HA. I believe that's consistent with the current behavior, right?

In a 3+1 cluster, both fields would be set to HA. That's the goal of this enhancement.

In an HA cluster, both fields would be set to HA. That's consistent with the current behavior.

enhancements/installer/compact-plus-worker.md
flaper87: > I suppose what I am caught up on is the "significantly" part. Can we remove that color unless we are providing specifics to define what we deem "significant"?

Sure, I will replace it with `it takes administrators at >2h longer` and remove the significantly

enhancements/installer/compact-plus-worker.md
flaper87: > If we think this is specific to ZTP then it may be reasonable to just modify the installer-generated manifests (and this perhaps limits the test impact, e.g we probably don't want to test 3+1 for every platform in CI?)

I don't think this is ZTP specific. Let's ignore that this came from a ZTP environment and focus on the hardware we are dealing with. Imagine an enclosure that only has 4 blades in it. The option to add another blade is out of the question. This leaves the administrator with the option to either do the installation in 2 steps or have to go through the process of patching manifests. 

I understand the concern on the test impact of this change.

enhancements/installer/compact-plus-worker.md
flaper87: > If we think this is specific to ZTP then it may be reasonable to just modify the installer-generated manifests (and this perhaps limits the test impact, e.g we probably don't want to test 3+1 for every platform in CI?)

I guess I don't where the line between supported and unsupported architectures is drawn.

FWIW, I am going to remove this line from the non-goals. I'm just taking it as an opportunity to think through what's really supported and what's not. 

I think @hardys  may be right in considering the architectures that we run tests on as the ones that are "officially" supported. 

As far as testing a 3+1 deployment like the one proposed here, I wonder if the 3+0 one that exists already would be enough. The difference between a 3+0 and a 3+1 would be that there is one node that is just a worker (instead of Master,Worker) so, in terms of architecture, they are both the same.

enhancements/installer/compact-plus-worker.md
flaper87: > In this proposal, instead of saying the infrastructureTopology is the same as the controlPlaneTopology, we could say that rather than counting only workers the infrastructureTopology setting should be based on the total number of nodes on which infrastructure pods could be scheduled (including the control plane, when those nodes are schedulable).

This is the gist of this proposal and what I think the final change would look like.

Admittedly, this enhancement could be rephrased in terms of the changes desired on the way the InfrastructureTopology is chosen. I decided to define the enhancement around the 4-nodes "concept" because of the changes proposed to the way we automatically flag masters as schedulable.

I think both changes are important. Just doing the change to how the InfrastructureTopology is chosen would still allow people to create a 4-nodes cluster that is unhealthy, which would result in a bad user experience.

enhancements/installer/compact-plus-worker.md
flaper87: > Supportability means that if a user performs the actions and ends up having a problem, then we will help them fix the problem without first requiring that they undo the actions. The fact that it takes the user two steps to get there has no bearing on whether it is something that we do or do not support.

I will remove references of the term "support" in this enhancement. 

However, I think it's more complicated than how you are phrasing it here. There's always a gap between what developers think is supported and what the user thinks is supported. Many of the actions we take will communicate to the user whether we think something is ok or not. Finally, our actions in cases like this will help us provide better support (to @hardys point on what the test matrix would look like).

enhancements/installer/compact-plus-worker.md
hardys: IMO we should have at least some testing of the 3+1 topology, somewhere - testing 3+0 is not sufficient IMO since the worker hosts are deployed and potentially configured differently (different MCP/ignition, different deploy tooling in the IPI case) 

Also the runtime behavior is different to compact clusters - for example the ingress VIP can end up on either a controlplane or compute host, and may move between them

enhancements/installer/compact-plus-worker.md
staebler: I don't want to get caught up on this detail, but I am dumbfounded as to why it would take more than 2 hours to add a worker to a cluster. There are a number of ways to accomplish this via one-line change to installer manifests. And even if you wait until the cluster is installed, the change should be small to make and the rollout of the machines should not take too long. Maybe I am missing something in the context of ZTP that adds to the burden, as I am more used to cloud platforms.

enhancements/installer/compact-plus-worker.md
staebler: Do you think that if we implemented this EP that we would no longer support the 3+1 configuration where the control planes nodes are not schedulable for workloads? If so, should we actively reject that configuration in-cluster?

enhancements/installer/compact-plus-worker.md
flaper87: @dhellmann @staebler @hardys thanks for your comments so far. 

I have been looking more into this and what options are available for the user.

I wonder if it would be better for the discussion to narrow things a bit more and focus on the part that would, at least, unblock the functionality. There are 2 items that I think would help us move forward. These items are not mutually exclusive, but I think they can be worked on independently:

- Allow for setting the `InfrastructureTopology` through manifests:
    - I wrote down my findings in my last comment on [this thread](https://github.com/openshift/enhancements/pull/1057/files?diff=unified&w=1#r824987520). Unless I am missing something, there is currently no way for the `InfrastructureTopology` to be set through manifests
- Count schedulable masters as workers when selecting the InfrastructureTopology


The above leaves out, from this enhancement, the switch of the master nodes to schedulable when there is one worker. This will still need to be an explicit action (override on the manifests) done by the user.

Since setting masters as schedulable requires modifying a manifest, it may make more sense allow for setting the InfrastructureTopology through manifests too. This, however, means API changes to the Infrastructure CR will be required. Again, this is all assuming my findings are correct.

In my opinion, I think overriding manifests is not the greatest use experience. I understand the value as it helps with providing a more explicit support matrix. On the other hand, making assumptions like 1 worker == SingleReplica (especially when they cannot be changed after the fact) is also not a good experience.

enhancements/installer/compact-plus-worker.md
staebler: > Is it possible to overwrite the Status using manifests?

Yes, you can set the status in the cluster-infrastructure-02-config.yml manifest.

flaper87: > > Is it possible to overwrite the Status using manifests?
> 
> Yes, you can set the status in the cluster-infrastructure-02-config.yml manifest.

Then I must have done something wrong. I will retry and get back. Thanks for the feedback

enhancements/installer/compact-plus-worker.md
```diff
@@ -0,0 +1,194 @@
+---
+title: Compact + Worker deployment as day0 operation
+authors:
+  - "@flaper87"
+reviewers:
+  - TBD
+approvers:
+  - TBD
```
dhellmann, line 8:
```suggestion
  - "@dhellmann"
```

dhellmann (commented):
You can add me as approver for this one.

/assign

enhancements/installer/compact-plus-worker.md
flaper87: I took more time to do an e2e test with assisted installer and manifests override. I managed to get a 4-node cluster deployed but there are multiple limitations with the current approach. 

1. The current approach requires applying custom manifests
2. openshift-installer applies manifests in order rather than merging them. This means that overriding the Infrastructure CR requires passing a fully populated manifest. 
3. The InfrastructureTopology is a `Status` field, which is commonly read as an output that came from some internal operations. Overriding it just doesn't feel right.

The way custom manifests is implemented in Assisted Service is that the custom manifest is pushed to the API, then it's made part of the bootstrap ignition. In my tests I created a file called `55-infrastructure-override.yaml` that was added to the `manifests` dir that the `openshift-installer` read from. This manifest overwrote the original one generated by the installer.

I got it to work in my test but I don't think this is a good approach. There are fields in the Infrastructure CR that are computed and they are dynamic (i.e infrastructureName). To override the Infrastructure CR, we would have to develop an infrastructure name that will differ from the one generated by the installer. Here's an example of what it looks like:

image

In the screenshot above, the infrastructure CR has the name I picked but the machines were created with the name computed by the installer.

Looking ahead (and taking a step back):

I think as a first approach we could start by exposing the `InfrastructureTopology` somehow through the deployment process. Either by putting it in the install-config or by exposing it in the Infrastructure CRD spec. We can leave any conversation about setting masters as schedulable and counting schedulable masters as workers for future enhancements.

Would this be something reasonable? @staebler @dhellmann 

The impact of this work would be smaller and it would allow automated flows to customize deployments without having to consider this architecture supported or having to expand the test matrix. 

For completeness, here's a screenshot showing the deployment tasks, the nodes, and operators:

image

staebler: > 1. The current approach requires applying custom manifests

This is an accepted and valid way of interacting with the OpenShift installer.

> 2. openshift-installer applies manifests in order rather than merging them. This means that overriding the Infrastructure CR requires passing a fully populated manifest.

The OpenShift installer supports editing the manifests that the OpenShift Installer generates. The OpenShift Installer does not require, or even condone, the user to provide a fully populated manifest.

> 3. The InfrastructureTopology is a Status field, which is commonly read as an output that came from some internal operations. Overriding it just doesn't feel right.

The fact that it is a Status field is a consequence of using the difficulty of updating Status fields as a placeholder for immutability. In this case, there is no restriction against or recommendation to not edit the Status fields.

> The way custom manifests is implemented in Assisted Service is that the custom manifest is pushed to the API, then it's made part of the bootstrap ignition.

Can the Assisted Service be modified to support editing manifests in addition to the current behavior of overwriting manifests?

dhellmann: > > 1. The current approach requires applying custom manifests
> 
> This is an accepted and valid way of interacting with the OpenShift installer.

I agree, manifests are the preferred way of doing "unusual" things like this.

> > 2. openshift-installer applies manifests in order rather than merging them. This means that overriding the Infrastructure CR requires passing a fully populated manifest.
> 
> The OpenShift installer supports editing the manifests that the OpenShift Installer generates. The OpenShift Installer does not require, or even condone, the user to provide a fully populated manifest.
> 
> > 3. The InfrastructureTopology is a Status field, which is commonly read as an output that came from some internal operations. Overriding it just doesn't feel right.
> 
> The fact that it is a Status field is a consequence of using the difficulty of updating Status fields as a placeholder for immutability. In this case, there is no restriction against or recommendation to not edit the Status fields.

I'm not sure I'd go quite that far. We could say that for this use case it's OK to change the value in the manifest, but I'm not sure we want to recommend that users do that themselves as a regular practice. I know we can't stop them from changing it, but that's different from recommending it.

> > (dhellmann) In this proposal, instead of saying the infrastructureTopology is the same as the controlPlaneTopology, we could say that rather than counting only workers the infrastructureTopology setting should be based on the total number of nodes on which infrastructure pods could be scheduled (including the control plane, when those nodes are schedulable).
> 
> (flaper87) This is the gist of this proposal and what I think the final change would look like.
> 
> Admittedly, this enhancement could be rephrased in terms of the changes desired on the way the InfrastructureTopology is chosen. I decided to define the enhancement around the 4-nodes "concept" because of the changes proposed to the way we automatically flag masters as schedulable.
> 
> I think both changes are important. Just doing the change to how the InfrastructureTopology is chosen would still allow people to create a 4-nodes cluster that is unhealthy, which would result in a bad user experience.

I don't think you fully answered my earlier question. It seems to me that the problem isn't the actual number of nodes being deployed, but is rather that the way we compute the value for the infrastructureTopology is incorrect because it doesn't count all of the nodes that it should.

I think that change, combined with ZTP explicitly setting the control plane nodes to be schedulable, should be sufficient for what you're trying to do in this 4 node case and that it would work correctly for other size clusters, too.

Would there be any negative side-effects of using that approach?

romfreiman: > Can the Assisted Service be modified to support editing manifests in addition to the current behavior of overwriting manifests?
Nice. Never thought in this direction (and honestly, no one ever needed this functionality so far). But worth exploring regardless (unrelated to this enhancement of course). BTW, I assume hive does not support editing neither, right?

staebler: > > Can the Assisted Service be modified to support editing manifests in addition to the current behavior of overwriting manifests?
> 
> Nice. Never thought in this direction (and honestly, no one ever needed this functionality so far). But worth exploring regardless (unrelated to this enhancement of course). BTW, I assume hive does not support editing neither, right?

You are correct that Hive does not current support editing either. They have a use case for it to support Managed ODF, where they would want to edit the worker machineset manifests to add a security group.

staebler: > > > (flapper87) 5\. The InfrastructureTopology is a Status field, which is commonly read as an output that came from some internal operations. Overriding it just doesn't feel right.
> > 
> > 
> > (staebler) The fact that it is a Status field is a consequence of using the difficulty of updating Status fields as a placeholder for immutability. In this case, there is no restriction against or recommendation to not edit the Status fields.
> 
> (dhellman) I'm not sure I'd go quite that far. We could say that for this use case it's OK to change the value in the manifest, but I'm not sure we want to recommend that users do that themselves as a regular practice. I know we can't stop them from changing it, but that's different from recommending it.

Note that the user is not mutating the Status of a resource. The user is editing the Status of a manifest representing a resource to be created. The Status is not representative of the status of a resource, which is something that I agree we would to want to suggest that a user should edit. By the nature of installer manifests, none of the Status fields in the manifests are representative of the status of a resource. I see no meaningful difference between the user editing a Spec field or a Status field of the manifest for the Infrastructure resource.

flaper87: > > (flaper87) 2. openshift-installer applies manifests in order rather than merging them. This means that overriding the Infrastructure CR requires passing a fully populated manifest.
> 
> (staeblerer) The OpenShift installer supports editing the manifests that the OpenShift Installer generates. The OpenShift Installer does not require, or even condone, the user to provide a fully populated manifest.

Can you point me to an example of how I can pass a partially-populated manifest to the installer? 

I tried passing an Infrastructure CR with just the `infrastructureTopology` and it ended up overriding the entire CR, leaving all the other fields in the status empty.
 
> > (flaper87) The way custom manifests is implemented in Assisted Service is that the custom manifest is pushed to the API, then it's made part of the bootstrap ignition.
> 
> (staeblerer) Can the Assisted Service be modified to support editing manifests in addition to the current behavior of overwriting manifests?

I thought of this but I didn't want to "just do it on our side" but rather bring the discussion here so that we could have a chance to think whether it would make sense to expose it through the installer or not. I hear that it's ok to implement our own patching and logic, which would help us work around this limitation.

> (dhellmann) I don't think you fully answered my earlier question. It seems that the problem isn't the actual number of nodes being deployed, but rather that the way we compute the value for the infrastructureTopology is incorrect because it doesn't count all of the nodes that it should.
> 
> I think that change, combined with ZTP explicitly setting the control plane nodes to be schedulable, should be sufficient for what you're trying to do in this 4 node case and that it would work correctly for other size clusters, too.
>
> Would there be any negative side-effects of using that approach?

It would be sufficient as it would pick the right InfrastructureTopology. This would bring us back to the points @hardys brought up with regards to what is tested and what isn't. Right now it seems that only compact and full cluster deployments are being tested. 

This is fundamentally what I am trying to achieve in this proposal, to have the installer take into account schedulable masters and workers when picking the topology. Admittedly my wording was not the best.

If this is fine and an acceptable solution, I would be happy to reword the enhancement so that we can focus on that, without exposing the `InfrastructureTopology`. The reason I proposed the latter in my previous comment is that I wanted the installer to make fewer assumptions about what the environment looks like and what the customer wants. especially for fields that are immutable for the rest of the cluster lifecycle.

enhancements/installer/compact-plus-worker.md
LalatenduMohanty: We should remove this point from non-goals. If it is in non-goals there is very less motivation for implementing this.

LalatenduMohanty (changes requested):

PR #1059, opened 2022-03-14
Converge ZTP and Metal3 flows

dtantsur:


enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,221 @@
...
+- In the *live ISO* flow, it is not possible to notify Ironic (and thus
+  baremetal-operator) when the actual installation finishes. From the
+  perspective of Metal3 components, the installation is done once the ISO is
+  connected. As a result, Ironic cannot disconnect the ISO from the machine's
+  BMC, nor can it update the BMC's boot device settings properly.
+- The *live ISO* flow disables Ironic *inspection* process and has to simulate
+  it via the assisted components. This creates an unnessesary coupling between
+  the assisted service and the baremetal-operator's hardware inventory format
+  (which, in turn, is derived from the Ironic's inventory format).
+- The *standard* flow is highly extensible via the
```
romfreiman, line 53: can u explain what does it mean? what extensibility?

dtantsur: I definitely can, but it will be a dive into Ironic's features, and I wonder if it's beneficial for this enhancement? I'm referring here to the ability for us to plug additional code in pretty much any stage of the provisioning through a custom agent plugin.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+- Some features, namely firmware and RAID configuration, require using the
+  Ironic agent, which is not present in the *live ISO* flow.
+- Supporting (i)PXE boot with the *live ISO* flow requires creating new API
+  in baremetal-operator specifically for this purpose.
+- In the *live ISO* flow, it is not possible to notify Ironic (and thus
+  baremetal-operator) when the actual installation finishes. From the
+  perspective of Metal3 components, the installation is done once the ISO is
+  connected. As a result, Ironic cannot disconnect the ISO from the machine's
+  BMC, nor can it update the BMC's boot device settings properly.
+- The *live ISO* flow disables Ironic *inspection* process and has to simulate
```
flaper87, line 49: To me, the motivation is more along the lines of unifying the inspection step rather than the fact that Assisted Service has one. The way this is phrased makes it sound like assisted created its own inspection process, despite there being one. In truth, assisted already had an inspection step, which is now being used for the ZTP flow.

The motivation for the above is that there was already a way to disable inspection in ZTP and not one in the assisted agent.

This distinction is important because I don't think the AI inspection step will be disabled, as it populates the internal data used for validations. What will happen is that the baremetal operator will use Ironic's inspection data and AI will use the assisted agent one. This could be done today but it's just slower for the user because the two agents need to be booted separtedly.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+  assisted service's `InfraEnv` resources, when they are present for the given
+  `BareMetalHost`.
+- When `InfraEnv` resources are present, the generated images will contain
+  Ignition configuration for both agents. The Ironic agent will be started
+  automatically and handle coordination with Ironic and baremetal-operator.
+- [Ironic agent][ironic-agent] currently contains a custom deployment plugin
+  (*custom deploy method* in baremetal-operator terms, *custom deploy step* in
+  Ironic terms) that invokes `coreos-installer` for the image installation.
+  A new custom deploy method `install_assisted` will be added to start the
+  assisted agent at the right moment.
```
flaper87, line 81: MAy be worth mentioning that Ironic will disable AIA's systems service as part of the ignition customization.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+  the Ironic agent configuration. Short-term, we'll probably document this
+  as a known issue. Long-term, we may need to update `InfraEnv` with a new
+  field.
+
+## Design Details
+
+### Open Questions
+
+- We are going to put the user's pull secret directly to the `InfraEnv`
+  resources as part of the Ignition override. Is it acceptable?
```
flaper87, line 156: Can you elaborate on who is `we` here? Do you mean that ICC will be injecting the PS?

The pullSecret is already requested as part of the ZTP flow, so what you are describing here is not a new thing.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+### Non-Goals
+
+- Converging two agent implementations (Ironic and assisted) into a single one.
+- Making assisted components mandatory for all bare-metal deployments.
+
+## Proposal
+
+- Building CoreOS images for the *standard* flow is handled by
+  [image-customization-controller][icc] (ICC). It will be updated to use the
+  assisted service's `InfraEnv` resources, when they are present for the given
```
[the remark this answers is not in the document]
dtantsur: I think I mention it below in more details.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+  `BareMetalHost`.
```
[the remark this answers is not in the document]
dtantsur: An optional dependency.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+Image:
+  URL: url/of/assisted/agent.iso
+  DiskFormat: live-iso
+```
+
+The new process will be instead triggered by:
+
+```yaml
+CustomDeploy:
+  Method: install_assisted
```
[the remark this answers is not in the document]
dtantsur: This string comes from Ironic, not from kubernetes, and fits its standard, as well as the pre-existing `install_coreos` step.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+Currently, the installation is initiated by applying the following change to
+the `BareMetalHost`'s spec:
+
+```yaml
+Image:
+  URL: url/of/assisted/agent.iso
+  DiskFormat: live-iso
+```
+
+The new process will be instead triggered by:
```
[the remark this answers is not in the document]
dtantsur: I *think* it's the user. @flaper87?

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+  resources as part of the Ignition override. Is it acceptable?
+- Is there a chance of version mismatch between the CoreOS `initrd` from the
+  assisted service and the root filesystem provided by
+  [coreos-downloader][coreos-downloader]?
+
+### Test Plan
+
+The feature will be covered by the existing ZTP e2e tests.
+
+### Graduation Criteria
```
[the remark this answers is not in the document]
dtantsur: All I can write here is "GA" since there is no staging possible or planned.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+For the new flow to apply, it must be explicitly requested through the
+`BareMetalHost` API. Otherwise, the old flow is used.
+
+There are no plans to remove the code that supports the *live ISO* flow from
+Ironic, Metal3 or OpenShift.
+
+### Version Skew Strategy
+
+The image-customization-controller will need to be updated before the new flow
+will be usable.
```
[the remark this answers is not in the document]
dtantsur: I think so. Given that ZTP releases trail OCP releases (and are tied to them), it should not be a big problem.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+
+- [Updated image building
+  process](https://github.com/openshift/image-customization-controller/pull/42)
+- [Ironic agent
+  plugin](https://github.com/openshift/ironic-agent-image/pull/40)
+
+## Drawbacks
+
+- The converged flow is more complex than each of the old flows.
+- The image-customization-controller becomes aware of the assisted-service API.
```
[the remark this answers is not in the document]
dtantsur: It's not coupling, the dependency is optional.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+N/A
+
+#### Removing a deprecated feature
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+This change does not introduce new resources, so upgrading or downgrading them
+is not a concern.
```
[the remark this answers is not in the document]
dtantsur: The linkage is via labels, which are already set (by BMAC or a user - Flavio may know that).

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
flaper87: BMAC will set this, so assisted service

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,249 @@
...
+- [Updated image building
+  process](https://github.com/openshift/image-customization-controller/pull/42)
+- [Ironic agent
+  plugin](https://github.com/openshift/ironic-agent-image/pull/40)
+
+## Drawbacks
+
+- The converged flow is more complex than each of the old flows.
+- The image-customization-controller becomes aware of (although does not
+  require) the assisted-service API.
```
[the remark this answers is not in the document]
dhellmann: "assisted-service API" means something specific elsewhere, so I threw me off to see it phrased that way here. "ZTP API" or "InfraEnv" seem less confusing in context here.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+### Implementation Details/Notes/Constraints
+
+#### Image building
+
+ICC will follow different code paths depending on whether the `BareMetalHost`
+has the `infraenvs.agent-install.openshift.io` annotation linking it to an
+`InfraEnv` resource. In both cases it will generate an Ignition configuration
+to start and configure the Ironic agent. In the ZTP case, the corresponding
+`InfraEnv` will be updated with the generated Ignition configuration (merging
+it with any user-provided configuration).
```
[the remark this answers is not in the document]
dhellmann: Right, so what I would expect to see is some way for some ZTP component to indicate to ICC that it needs to do something other than the default behavior, without anything that is tightly coupled to ZTP being baked into ICC. Is there some way to pass data from ZTP to ICC in a way that ICC can include it in the image, for example?

OR is this a step on a path to making this coordinated behavior the new default for ICC? If that's the case, then saying it in the design sets a different context for reviewing the proposed changes.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
zaneb: I assume we're excluding the upgrade impacts on the BMAC from this enhancement?

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,249 @@
...
+## Drawbacks
+
+- The converged flow is more complex than each of the old flows.
+- The image-customization-controller becomes aware of (although does not
+  require) the assisted-service API.
+
+## Alternatives
+
+- Keep both flows separate, implement all required feature twice.
+
```
zaneb, line 239: An alternative we discussed is for the ICC to retrieve the Ignition from assisted-service and build the image including the IPA config itself, instead of the passing the IPA config to assisted-service and having that assisted-image-service build the image.

(A drawback to this is that minimal ISO support has not yet been added to ICC.)

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dtantsur: > what I would expect to see is some way for some ZTP component to indicate to ICC that it needs to do something other than the default behavior

That's what we do. ICC uses a label on PreprovisioningImage (which is copied from a label on BareMetalHost) to decide to follow the alternative path.

> Is there some way to pass data from ZTP to ICC in a way that ICC can include it in the image, for example?

Doing this will require the data to go through baremetal-operator, which is an upstream project that is not specific to OpenShift (unlike ICC).

> OR is this a step on a path to making this coordinated behavior the new default for ICC?

No, the default behavior will still be using Ironic only and building the image ourself. I think we'll end up with 2 paths for the time being.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,249 @@
...
+### Goals
+
+- ZTP follows the *standard* Metal3 flow instead of *live ISO*.
+- (i)PXE boot, firwmare and RAID settings work for ZTP without introducing
+  special API in Metal3 components.
+
+### Non-Goals
+
+- Converging two agent implementations (Ironic and assisted) into a single one.
+- Making assisted components mandatory for all bare-metal deployments.
```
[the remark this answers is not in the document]
dtantsur: Yeah, but we don't have anti-goals, so I've put it here to prevent people from asking :)

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,249 @@
...
+  as a known issue. Long-term, we may need to update `InfraEnv` with a new
+  field.
+
+## Design Details
+
+### Open Questions
+
+- ICC will store the user's pull secret directly to the Ignition override
+  field the `InfraEnv` resource (as a base64 encoded text, not as a reference
+  to secret). Is it acceptable?
```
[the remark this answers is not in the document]
dtantsur: I suspect assisted does it in code. Since it does not understand what our Ignition is doing, it has no way to inject it there.

Maybe we can start the ironic agent in a way that reuses the secret. But it would mean coupling our code with the way the assisted agent is started. Dunno. @flaper87?

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dtantsur: At the very least, someone else will have to write it. I don't understand BMAC even remotely well enough.

romfreiman:
Somewhere we have to state that the firmware update path will 'break' discovery flow by first updating the firmware and only then go through validations. It's sub-optimal, but might make (some) sense in ZTP

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dhellmann: I'm trying to get you to be more precise in the wording. "assisted-service API" usually means the REST API, but that doesn't seem to be the case in this instance. If it is just the InfraEnv, say that. If it is the InfraEnv plus other CRDs, list them or mention the operator that owns them.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dhellmann: And if the label is present but the CRD is not installed?

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dhellmann: > The API here is really the PreprovisioningImage CRD, and it is internal to Metal³. The BMO creates the resource and retrieves the image URL from it.
> ICC is our OpenShift-specific controller for PreprovisioningImage. As such, it is the right place to be putting logic about other OpenShift-specific features, such as ZTP. Since the CRD is defined by Metal³, the only way to provide additional information (other than having it try to call in to services that may or may not exist) is via an annotation.
> This actually works out quite well because ZTP already annotates the BMHs, and Metal³ can copy that annotation to the PreprovisioningImage to make determining when to use ZTP highly efficient.

There are 2 ways to treat an API like that: separate controllers handle separate instances (like Machine, or Ingress, or storage classes) or one controller incorporates multiple sets of logic (as is proposed here). I'm suggesting that building logic into a core OCP component to know about a layered product (ZTP) introduces an unnecessary level of coupling between the layers.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
zaneb: assisted-service passes it in the file `/root/.docker/config.json` (which is where `podman login` would save it): https://github.com/openshift/assisted-service/blob/master/internal/ignition/ignition.go#L240-L248
i-c-c passes it in the file `/etc/authfile.json`: https://github.com/openshift/image-customization-controller/blob/main/pkg/ignition/service_config.go#L58-L62
All we have to do is use the same file for both.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
zaneb: Why would a user add an annotation pointing to an InfraEnv on a Host in a cluster that is not running ZTP and has no InfraEnv CRD?
There's no way that any API anywhere can reliably guess what the user meant when they provide garbage input. All we can do is keep attempting to eventually converge, and report any errors in a way that will hopefully make sense to the user.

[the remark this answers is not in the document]
dhellmann: Users do all sorts of crazy things. The controller needs to be able to cope with that case by not crashing, at the very least.

enhancements/baremetal/ztp-metal3.md
avishayt: AIA == Assisted Installer Agent

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,249 @@
...
+  A new custom deploy method `install_assisted` will be added to start the
+  assisted agent at the right moment.
+- The assisted service will stop explicitly disabling *inspection* on
+  `BareMetalHost`s it manages, will start using a different process based
+  on the  new `install_assisted` deploy method and stop rebooting the machine
+  in the end of the process.
+
+### User Stories
+
+N/A (this change itself does not add new user stories)
```
avishayt, line 94: Users wanting firmware/RAID with ZTP?

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,249 @@
...
+
+All required API extensions are already present.
+
+### Implementation Details/Notes/Constraints
+
+#### Image building
+
+ICC will follow different code paths depending on whether the `BareMetalHost`
+has the `infraenvs.agent-install.openshift.io` annotation linking it to an
+`InfraEnv` resource. In both cases it will generate an Ignition configuration
```
avishayt, line 106: Do we need some coordination here between ICC and BMAC when an InfraEnv is updated?

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dtantsur: > And if the label is present but the CRD is not installed?

I think it will be a normal error propagated through `PreprovisioningImage`, but I can double-check.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
flaper87: What @zaneb said! Assisted passes the PS as part of the image itself. We need to make sure to use the same for both.

enhancements/baremetal/ztp-metal3.md
flaper87: > Do we need some coordination here between ICC and BMAC when an InfraEnv is updated?

Ideally, with this work, BMAC should be able to stop syncing the BMH with the `InfraEnv` and the `Agent`. This means, the agent inspection will only be used by AI (instead of it being propagated to the BMH). The InfraEnv URL will be read by ICC.

Finally, we will likely stop setting the `baremetalhost.metal3.io/detached` annotation. At the very least, we would set it if and only if the deployment succeeded but I don't think we will need to set this annotation anymore

enhancements/baremetal/ztp-metal3.md
hardys: +1 we could also perhaps have stories around environments where the in-band efibootmgr approach doesn't work reliably, which would be solved by having Ironic managed the ISO attach/detach like in the regular virtualmedia flow.

Another story would be PXE boot for ZTP customers (which could be solved without doing this, but that becomes much easier as a side-effect of this alignment?)

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,239 @@
...
+
+- We are going to put the user's pull secret directly to the `InfraEnv`
+  resources as part of the Ignition override. Is it acceptable?
+- Is there a chance of version mismatch between the CoreOS `initrd` from the
+  assisted service and the root filesystem provided by
+  [coreos-downloader][coreos-downloader]?
+
+### Test Plan
+
+The feature will be covered by the existing ZTP e2e tests.
```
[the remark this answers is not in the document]
hardys: It may be helpful to identify and list the ZTP e2e tests which currently exist - do we need some new ones?

For example should we have CI coverage for the newly-enabled PXE flow and/or RAID/firmware config?

How can we validate on upgrade that existing ZTP deployments aren't impacted (maybe this can be manual QE validation vs CI testing)?

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
hardys: Outlining the BMAC upgrade impact (and how we'll ensure test coverage) seems like a good idea, e.g we'll need to be very sure that BMAC leaves existing deployed BMH resources alone, that scale-out post upgrade works as expected, etc - @flaper87 can you help with that?

enhancements/baremetal/ztp-metal3.md
zaneb: I don't think there's any connection between the `detached` annotation and this work.
Detaching the host means that ironic in the hub cluster won't continue to probe it (for power state) and metal³ won't attempt to manage it. You should do this whenever there is a BaremetalHost created in a spoke cluster that is self-managing the host, and probably not if there is not. Either way though, it is independent of this enhancement.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
zaneb: _Something_ will still have to set the CustomDeploy method at provisioning time. I don't recall where we got to on which part of assisted-service is expected to do that.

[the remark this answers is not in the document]
flaper87: > Something will still have to set the CustomDeploy method at provisioning time. I don't recall where we got to on which part of assisted-service is expected to do that.

Fair point, I knew I was forgetting something. Who sets it in the `install_coreos` case? The user? If so, then we can let the user do this for assisted to. Otherwise, BMAC or ICC can do it.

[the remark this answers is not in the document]
zaneb: > Who sets it in the `install_coreos` case?

CAPBM

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
zaneb: I guess we could list as an alternative that i-c-c could just ignore all CRs with an InstallEnv label, and a separate controller (probably running in the same process as BMAC) could reconcile those ones. The complex parts of the controller code are provided by upstream Metal³ anyway, so you'd only have to write a very simple ImageProvider plugin.

I'm not sure what that buys us though. A second place to vendor the controller code just adds opportunities for it to get out of sync. i-c-c would still need to know about the layered product anyway so it could ignore the label. And if the user did apply the label erroneously as you asked about in the other thread, we'd have no opportunity to supply an error message and  instead the CR would just sit there not getting reconciled forever.

romfreiman:
> Do you mean that this is what we'd have to do without the change proposed in this enhancement, and that's one of the motivations for doing this? If so I agree.

@zaneb hmm I think regardless of this work we'll not be able to run validations prior to applying bios update. @dtantsur  @flaper87 am I wrong here? (I would love to be wrong in this case)

dtantsur:
> we'll not be able to run validations prior to applying bios update

Correct. Not in this iteration at least.

zaneb:
> > we'll not be able to run validations prior to applying bios update
> 
> Correct. Not in this iteration at least.

OK apparently the plan has changed since I last heard it then, because I thought this was explicitly one of the goals.

Now that you mention it, the text in this enhancement saying that the AIA will only be started once the host in provisioned is indeed incompatible with running the validations before the firmware settings. Is the current plan to have the BMAC trigger the provisioning unconditionally (i.e. without waiting for validation) just as it does today.

My thought on this was always that we would have BMAC apply the firmware settings and then start the provisioning of the host only after validations had been run. Did we just put that in the 'too hard' basket?

romfreiman:
> OK apparently the plan has changed since I last heard it then, because I thought this was explicitly one of the goals.

It was. But I dont think we have a solution for that in this mode :(

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dtantsur: Listing this as an alternative.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dtantsur: It does not have to come at a different stage, it just has to happen *after* the BIOS settings are written to the BMH, otherwise they won't be applied.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dtantsur: The location `/root/.docker/config.json` seems standard. Does it mean that `podman pull` will just work without an explicit authfile argument, @flaper87?

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dtantsur: I'll write a few basic things here, but I'd like input from assisted folks.

dtantsur:
I think I've addressed most of the comments that do not need further discussion. Please take a look.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,286 @@
...
+  the given `BareMetalHost`.
+- When `InfraEnv` resources are present, the generated images will contain
+  Ignition configuration for both agents. The Ironic agent will be started
+  automatically and handle coordination with Ironic and baremetal-operator.
+  The configuration will also supress automatic start-up of the assisted
+  agent.
+- [Ironic agent][ironic-agent] currently contains a custom deployment plugin
+  (*custom deploy method* in baremetal-operator terms, *custom deploy step* in
+  Ironic terms) that invokes `coreos-installer` for the image installation.
+  A new custom deploy method `install_assisted` will be added to start the
```
zaneb, line 90: If we don't start the assisted agent until the BMH is provisioned, then we need to continue using the current ZTP workflow where BMH provisioning happens immediately (before we actually decide to install a node on a host).

This removes many of the benefits of the original idea, including the benefit of being able to apply firmware config changes only after validation has happened and installation is definitely going ahead.

I think we should at least document as an alternative that both agents run in parallel and `install_assisted` unblocks the installation step in some other way.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,286 @@
...
+### API Extensions
+
+All required API extensions are already present.
+
+### Implementation Details/Notes/Constraints
+
+#### Image building
+
+ICC will follow different code paths depending on whether the `BareMetalHost`
+has the `infraenvs.agent-install.openshift.io` annotation linking it to an
```
zaneb, line 119: It's a label, not an annotation, right? We shouldn't use those terms interchangeably.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,286 @@
...
+`InfraEnv` resource. In both cases it will generate an Ignition configuration
+to start and configure the Ironic agent. In the ZTP case,
+image-customization-controller will update the corresponding `InfraEnv` with
+the generated Ignition configuration (merging it with any user-provided
+configuration).
```
zaneb, line 124: TBH, this feels like a mistake. Why should the configuration, which is an implementation detail, be via an API that is both readable and writable by the user?

Since the i-c-c already contains a web server, perhaps it could serve up the parts of the ignition that are not host-specific and are needed by assisted (i.e. excluding stuff like the pull secret) at `http://metal3-image-customization-service.openshift-machine-api.svc.cluster.local/assisted-ignition.json`, and assisted-service could just retrieve it from there. That way we solve the problem of having one canonical source for the IPA ignition config at runtime, without i-c-c having to write to the InfraEnv.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,286 @@
...
+  firmware/BIOS settings and iPXE boot) twice.
+- The image-customization-controller could just ignore all CRs with an
+  `InfraEnv` link, and a separate controller (probably running in the same
+  process as BMAC) could reconcile those ones. A second place to vendor the
+  controller code just adds opportunities for it to get out of sync. The
+  image-customization-controller will still need to know about the `InfraEnv`
+  API anyway so it could ignore hosts with the label.
+- ICC may retrieve the Ignition from assisted-service and build the image
+  including the Ironic agent config itself, instead of the passing it to
+  assisted-service and having that assisted-image-service build the image.
```
zaneb, line 274: An advantage of this is that in future, a given InfraEnv could work both with hosts that are booted by Metal³ and those booted manually by the user.

In the current proposal, once any hosts in an InfraEnv have corresponding BMHs, i-c-c will update the InfraEnv ignition to contain IPA and not start AIA. This makes the ISO in the InfraEnv useless for a user who wants to manually boot it and use it as a discovery image.

Today this is not a problem, but it could limit us in future.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,286 @@
...
+
+The first agent to start will be the Ironic agent since it is responsible
+for coordinating the whole flow and (indirectly through Ironic) updating the
+`BareMetalHost`'s status. *Inspection* and additional configuration (RAID etc)
+will be handled by the Ironic agent alone.
+
+BMAC will be able to stop synchronizing the inventory to `BareMetalHost`
+objects since this task will be handled by Ironic and baremetal-operator.
+
+Currently, the installation is initiated by BMAC (the assisted component
```
mhrivnak, line 145: BMAC doesn't initiate installation. It just boots a live ISO. Importantly, it is just one of many other ways a user could automate the booting of the live ISO.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,286 @@
...
+CustomDeploy:
+  Method: install_assisted
+```
+
+This instructs Ironic to find and execute a *custom deploy step* called
+`install_assisted`, which will be provided by an OpenShift specific Ironic
+agent plugin. This step will use D-Bus to order systemd to start the assisted
+agent and watch its status.
+
+If the assisted agent exits successfully, the Ironic agent will report success
```
mhrivnak, line 167: We should take care how "success" is defined and captured. I'm not sure we want to guarantee that if the agent exits 0, that means installation is proceeding, or anything else other than no errors occurred. It's also possible (I'm not sure if this is currently the case) that the agent won't exit until it's time to reboot, in which case there may not be time for the ironic agent to notice and report success.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
zaneb: Yes.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
dtantsur: `install_assisted` is consistent with `install_coreos` we have now (and with other deploy steps in Ironic that are verbs, not nouns).

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,286 @@
...
+### Implementation Details/Notes/Constraints
+
+#### Image building
+
+ICC will follow different code paths depending on whether the `BareMetalHost`
+has the `infraenvs.agent-install.openshift.io` annotation linking it to an
+`InfraEnv` resource. In both cases it will generate an Ignition configuration
+to start and configure the Ironic agent. In the ZTP case,
+image-customization-controller will update the corresponding `InfraEnv` with
+the generated Ignition configuration (merging it with any user-provided
```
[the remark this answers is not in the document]
dtantsur: Okay. Will assisted service be open to adding a new field for us OR pulling the Ignition from a URL like Zane suggests in another comment?

enhancements/baremetal/ztp-metal3.md
dtantsur: We will need to make sure the agent does not reboot, otherwise this whole idea may not work out.

enhancements/baremetal/ztp-metal3.md
dtantsur: Well, it does initiate the installation from our (= BMO) perspective. I don't think we should add here a whole range of options AI provides.

enhancements/baremetal/ztp-metal3.md
dtantsur: cc @flaper87

enhancements/baremetal/ztp-metal3.md
zaneb: s/installation/provisioning/ would be more accurate

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,320 @@
...
+
+- The new annotation with accept a URI instead of a full Ignition configuration
+  or a link to a secret because the configuration can be quite large when
+  many hosts are present.
+
+- ICC will replicate the graceful wait period logic from the assisted service:
+  the image will be built not earlier than 60 seconds from the last change
+  to the `PreprovisioningImage` object. This is done for two reasons: to give
+  the user enough time to apply all changes that may affect the build and to
+  reduce the probability of a race between BMAC and ICC.
```
dhellmann, line 192: Tracking modification time on CRs can be tricky. Is this really necessary? If so, let's describe how to track the update time.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,320 @@
...
+For the new flow to apply, it must be explicitly requested through the
+`BareMetalHost` API. Otherwise, the old flow is used.
+
+There are no plans to remove the code that supports the *live ISO* flow from
+Ironic, Metal3 or OpenShift.
+
+### Version Skew Strategy
+
+The image-customization-controller will need to be updated before the new flow
+will be usable. Given that ZTP releases trail OpenShift releases (and are tied
```
dhellmann, line 256: We'll have to allow ZTP and OCP to be upgraded independently and so ZTP will have to account for different versions of BMO with different expectations. How will it do that with no API version increment for BMH?

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,320 @@
...
+
+N/A
+
+### Upgrade / Downgrade Strategy
+
+This change does not introduce new resources, so upgrading or downgrading them
+is not a concern.
+
+For the new flow to apply, it must be explicitly requested through the
+`BareMetalHost` API. Otherwise, the old flow is used.
```
dhellmann, line 248: How long will the old workflow be kept in place?

dhellmann (commented):
I think this version captures what we've talked about elsewhere. I have a few questions about version skew and upgrades.

I would also like to understand how this step fits into the larger context where the two systems are more fully converged for workflows other than ZTP, but I think that's asking for more discussion and another enhancement.

enhancements/baremetal/ztp-metal3.md
dtantsur: Aha, assisted is using `Status.CreatedTime`, not modification time. Which is, indeed, a different thing, but fits this use case reasonably well.

enhancements/baremetal/ztp-metal3.md
dtantsur: Sorry, I've complete misread your question....

There are no immediate plans to deprecate the live ISO workflow. I think I mention it somewhere else.

enhancements/baremetal/ztp-metal3.md
dhellmann: I don't know if we've ever had to deal with this before, have we? ZTP needed some changes, but it just didn't work until the version of OCP that had those changes.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,321 @@
...
+## Proposal
+
+- Building CoreOS images for the *standard* flow is handled by
+  [image-customization-controller][icc] (ICC). It will be updated to optionally
+  accept additional Ignition configuration via a new annotation
+  `baremetal.openshift.io/ignition-override-uri`.
+
+- Additionally, ICC will recognize the label linking `BareMetalHost`s to
+  `InfraEnv` resources from the assisted service. When such a label is present,
+  ICC will wait for the non-empty Ignition override URI to be provided.
```
zaneb, line 93: I don't know that we need to wait for it to be non-empty. Just waiting for the annotation to be present should be sufficient, and still leave users with an out.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,321 @@
...
+
+After this enhancement, BMAC will instead apply:
+
+```yaml
+CustomDeploy:
+  Method: install_assisted
+```
+
+This instructs Ironic to find and execute a *custom deploy step* called
+`install_assisted`, which will be provided by an OpenShift specific Ironic
```
zaneb, line 163: I don't think this step should be called `install_assisted` if all it does is start the assisted agent and not actually begin an install. Something like `start_assisted` might be more appropriate.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,321 @@
...
+  effects of any changes. In addition, the *live iso* flow is not tested
+  upstream by the Metal3 project.
+
+### Goals
+
+- ZTP follows the *standard* Metal3 flow instead of *live ISO*.
+- (i)PXE boot, firwmare and RAID settings work for ZTP without introducing
+  special API in Metal3 components.
+
+### Non-Goals
```
zaneb, line 78: A non-goal is supporting per-host network configuration through the BMH PreprovisioningNetworkData API.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,321 @@
...
+
+There are no plans to remove the code that supports the *live ISO* flow from
+Ironic, Metal3 or OpenShift.
+
+### Version Skew Strategy
+
+The image-customization-controller will need to be updated before the new flow
+will be usable. Given that ZTP releases trail OpenShift releases (and are tied
+to them), it should not become an issue.
+
```
zaneb, line 259: Should mention here that the version of the ISO will always be the one shipped with the hub cluster, not necessarily the same version we expect the spoke cluster to run on.
We think this is fine because MCO on the spoke will replace the ostree before starting the host as a node anyway.
It does mean that we'll have to be careful about Ignition versions when specifying the override ignition - it can't be newer than what is supported by the hub cluster.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,321 @@
...
+### Operational Aspects of API Extensions
+
+#### Failure Modes
+
+- When image building fails in an `InfraEnv` resource, this failure will be
```
zaneb, line 264: This no longer reflects the plan.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,321 @@
...
+## Alternatives
+
+- Keep both flows separate, implement all required features (currently
+  firmware/BIOS settings and iPXE boot) twice.
+
+- The image-customization-controller could just ignore all CRs with an
+  `InfraEnv` link, and a separate controller (probably running in the same
+  process as BMAC) could reconcile those ones. A second place to vendor the
+  controller code just adds opportunities for it to get out of sync. The
+  image-customization-controller will still need to know about the `InfraEnv`
```
zaneb, line 300: s/will/would/

enhancements/baremetal/ztp-metal3.md
zaneb: Should probably also document "Perform firmware config changes only _after_ running host validations" as a non-goal, since if it were a goal then this plan wouldn't be sufficient.

enhancements/baremetal/ztp-metal3.md
dtantsur: > I don't know if we've ever had to deal with this before, have we? 

Well, at some point we introduced the live ISO support itself. I don't think BMAC actually checks the schema to validate the support is there.

dtantsur:
Most of the comments addressed. The version skew problem is the biggest issue now, more ideas are welcome.

enhancements/baremetal/ztp-metal3.md
dhellmann: Did we just assume the new BMAC would use the new BMO?

If we're introducing a behavior without an API version update because that API is an annotation and the CRD is owned upstream, then I don't see a safer way to check for the behavior than looking at the version of something. The brute force way would be to check the ClusterVersion resource for the OCP version. Something more subtle might involve having one of the operators drop an annotation on a configuration resource when the behavior is available. Perhaps someone on the API team can suggest alternatives based on their experience.

dhellmann (commented):
I think the version skew question is important enough to figure out before we merge this. I think the rest of it is ready to go.

enhancements/baremetal/ztp-metal3.md
zaneb: We could have the CBO put something in the status of the Provisioning CR.
Or even just checking the CBO version with CVO (rather than the cluster version) would be an improvement.

flaper87 (commented):
looks good to me

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,334 @@
...
+  Ironic terms) that invokes `coreos-installer` for the image installation.
+  A new custom deploy method `start_assisted_install` will be added to start
+  the assisted agent at the right moment.
+
+- The assisted service (more specifically, BMAC) will stop explicitly disabling
+  *inspection* on `BareMetalHost`s it manages, will start using a different
+  process based on the  new `start_assisted_install` deploy method and stop
+  rebooting the machine in the end of the process.
+
+- BMAC will also set the Ignition URI with the assisted agent configuration on
```
flaper87, line 112: I was a bit confused when reading this paragraph. It may be worth saying that BMAC will set the `ignition-override-url` annotation that is mentioned in the first paragraph of the proposal.

flaper87 (commented):
still looks good, just a nit based on my last reading :P

enhancements/baremetal/ztp-metal3.md
eranco74: We can make sure the agent won't reboot the host.
There is an issue where the agent service gets restarted during an agent upgrade (not common but can happen), in this case, the Ironic agent might interpret this restart as:  assisted-agent served its purpose and report success.
I think we should have a better API here for this communication.
Placing a file in a known location (perhaps even adding some info in the file success/failure, what failed, etc) should do the trick.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,325 @@
...
+For the new flow to apply, it must be explicitly requested through the
+`BareMetalHost` API (by setting `CustomDeploy` instead of `Image`). Otherwise,
+the old flow is used.
+
+There are no plans to remove the code that supports the *live ISO* flow from
+Ironic, Metal3 or OpenShift.
+
+### Version Skew Strategy
+
+The baremetal-operator will need to be updated before the new flow will be
```
eranco74, line 256: This is a prerequisite only for PXE or it's relevant for the live-ISO as well?
In other words, can the assisted-service set the `start_assisted_install` deploy method and update the PreprovisioningImage status with an image containing both agents  on HUB cluster in version 4.10 (or older)

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,325 @@
...
+  plugin](https://github.com/openshift/ironic-agent-image/pull/40)
+
+## Drawbacks
+
+- The converged flow is more complex than each of the old flows.
+
+## Alternatives
+
+- Keep both flows separate, implement all required features (currently
+  firmware/BIOS settings and iPXE boot) twice.
```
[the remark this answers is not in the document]
dtantsur: BIOS is intended here (to avoid confusion because of our usage of  "BIOS")

enhancements/baremetal/ztp-metal3.md
dtantsur: > set the start_assisted_install deploy method 

This deploy method will only be added in 4.11, so no.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,325 @@
...
+
+- The image-customization-controller could just ignore all CRs with an
```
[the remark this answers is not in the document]
dtantsur: Mmmm, right, forgot to remove.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,325 @@
...
+  `KernelURL` to store the matching kernel.
+
+  `ExtraKernelParams` to provide a link to the root file system. BMO will be
+  updated to pass the parameters to the node.
+
+- [Ironic agent][ironic-agent] currently contains a custom deployment plugin
+  (*custom deploy method* in baremetal-operator terms, *custom deploy step* in
+  Ironic terms) that invokes `coreos-installer` for the image installation.
+  A new custom deploy method `start_assisted_install` will be added to start
+  the assisted agent at the right moment.
```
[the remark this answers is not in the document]
dtantsur: yeah, a bit of hand-waving here. I don't know BMAC well enough to say for sure when it starts the deployment.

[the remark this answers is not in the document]
eranco74: I guess this will happen the same way BMAC  updates the BMH Image.URL today (by looking the InfraEnv label on the BMH)
[https://github.com/openshift/assisted-service/blob/4eee641dccec201432e154189a67b09[…]19186e8/internal/controller/controllers/bmh_agent_controller.go](https://github.com/openshift/assisted-service/blob/4eee641dccec201432e154189a67b09f719186e8/internal/controller/controllers/bmh_agent_controller.go#L678)
It should work just fine.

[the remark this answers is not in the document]
avishayt: From my understanding, BMAC adds the custom deploy method as soon as it first reconciles a BMH that is linked to an InfraEnv.  This has the effect of IPA starting Assisted Agent "at the right moment", which means after any bios/raid/whatever configurations have been completed - correct?

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,325 @@
...
+- Building CoreOS images for the *standard* flow is handled by
+  [image-customization-controller][icc] (ICC). ICC will start ignoring all
+  `PreprovisioningImage`s that have the `infraenvs.agent-install.openshift.io`
+  label (copied from the corresponding `BareMetalHost` by BMO).
+
+- A new controller will be created as part of the assisted service to
+  handle `PreprovisioningImage`s that do have the InfraEnv label. Its exact
+  implementation will be defined by the assisted team. The assisted agent
+  configuration will be updated to not start the agent automatically in the
+  presence of the Ironic agent.
```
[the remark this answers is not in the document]
eranco74: Do we always generate (and set) all artifacts (ISO + (i)PXE artifacts)?

dhellmann:
I think the latest draft matches what we've said in conversations elsewhere.

enhancements/baremetal/ztp-metal3.md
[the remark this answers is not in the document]
avishayt: I believe we're not "generating" anything ahead of time.  I'm not sure if the URL changes for things that aren't changed (like kernel) but I'm not sure how it affects the flow?

enhancements/baremetal/ztp-metal3.md
flaper87: I agree with @mhrivnak here that just the exit code may not be enough. @eranco74 and I met this morning and here's a proposal that may be worth exploring:

We could have both agents communicate through dbus. Have a single path there to set a flag that says that IAI has finished the installation. This should be fairly simple to do. The IPA PoC I wrote already uses dbus to start and monitor AIA. In the assisted installer we will also mount the dbus socket so that we can check whether ironic-agent is present, which gives it everything needed to send a message through that socket.

We would change https://github.com/openshift/ironic-agent-image/pull/40 to only start AIA and then check the dbus path we would add instead of monitoring AIA's unit status.

I would really like to avoid a file-based API/interaction, especially if we already have other methods for IPC in place.

dtantsur: What should Ironic do if the dbus path is not present? Assume the agent is dead? Wait a bit?

tsorya: @flaper87 i am not sure i understand why to use dbus in that case? honestly i prefer using file as it is much more visible and debuggable. More of it , it will allow much more control over agent as it will not run in case file exists for example and such mechanism exists all over the installation process and not sure why we need a new one. Maybe i don't understand what dbus gives us though

tsorya: in any case we can't depend on agent service not being restarted

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,328 @@
...
+  Method: start_assisted_install
+```
+
+This instructs Ironic to find and execute a *custom deploy step* called
+`start_assisted_install`, which will be provided by an OpenShift specific
+Ironic agent plugin. This step will use D-Bus to order systemd to start the
+assisted agent and watch its status.
+
+If the assisted agent exits successfully, the Ironic agent will report success
+to Ironic and baremetal-operator. Otherwise, a failure will be recorded.
```
tsorya, line 185: what will be done with failure? I mean what status we will have in the host? how user should continue from that state? Just not sure what error mean in that case

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,328 @@
...
+
+- Converging two agent implementations (Ironic and assisted) into a single one.
+- Supporting per-host network configuration through the BMH
+  PreprovisioningNetworkData API in the ZTP case.
+- Performing firmware config changes only after the assisted service runs host
+  validations.
+
+## Proposal
+
+- Building CoreOS images for the *standard* flow is handled by
```
zaneb, line 91: This first sentence needs some work, because it's directly contradicted by the second.

enhancements/baremetal/ztp-metal3.md
```diff
@@ -0,0 +1,328 @@
...
+  [image-customization-controller][icc] (ICC). ICC will start ignoring all
+  `PreprovisioningImage`s that have the `infraenvs.agent-install.openshift.io`
+  label (copied from the corresponding `BareMetalHost` by BMO).
+
+- A new controller will be created as part of the assisted service to
```
zaneb, line 96: This might be out of scope for this enhancement, but I'd consider putting it in the assisted-image-service (rather than assisted-service) so it has direct access to the image-building APIs. This could potentially avoid e.g. the need to pass the ironic agent config in the InfraEnv spec that is supposed to belong to the user.

avishayt: Good point, we should consider that.  Also agree that it probably doesn't need to be included here though.

enhancements/baremetal/ztp-metal3.md
dtantsur: Could you elaborate? Here I'm contrasting the standard flow with the ZTP flow.

enhancements/baremetal/ztp-metal3.md
dtantsur: It's going to propagate to the BMH object (which will probably cause BMO to retry, but I don't want this document to list all the details about BMO).

enhancements/baremetal/ztp-metal3.md
eranco74: The ignition config for the assisted discovery iso is defined in the assisted-service.
We can definitely add the ironic agent config to it.
Unlike our initial thought there is no need to add the ironic config to the infraEnv

PR #1062, opened 2022-03-16
Update the CSI migration enhancement

bertinatto:
Now the CSI StorageClass is preferably the default one.

CC @openshift/storage

### DIFF
--- a/enhancements/storage/csi-migration.md
+++ b/enhancements/storage/csi-migration.md
@@ -7,7 +7,7 @@ reviewers:
 approvers:
   - "@openshift/openshift-architects"
 creation-date: 2020-07-29
-last-updated: 2021-03-31
+last-updated: 2022-03-17
 status: implementable
 see-also: https://github.com/openshift/enhancements/pull/463
 replaces:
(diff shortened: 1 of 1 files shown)


bertinatto:
/hold
to update other sections

dhellmann:
> /hold
> to update other sections

@bertinatto Since you're updating an existing doc, you don't have to add the sections that have been added to the template. We can override the test job to let this land. When you're ready, let me know and I can do that.

enhancements/storage/csi-migration.md
```diff
@@ -133,12 +136,36 @@ spec:
...
+In order to achieve that, these are the main changes introduced in OCP 4.11 with regards to SCs:
+
+* The *defaultstorageclasscontroller* from CSO will be completely removed and we will not create any new in-tree SCs.
+  * SCs that were previously created by this controller will not be owned by any operator and may very well be deleted by the user.
+* A new special-purpose controller will be added to the CSI Driver Operator.
+  * This controller will make sure that a CSI SC always exists.
+  * And it will set that SC as the default one when appropriate.
+  * Also, it will make sure there is only **one** default SC.
+
+In a more detaild way, this is how the new controller will operate and determine whether the CSI SC should be the default one:
```
dobsonj, line 158: typo: detailed

enhancements/storage/csi-migration.md
```diff
@@ -133,12 +136,36 @@ spec:
...
+
+* On every sync, thie controller fetches all existing SCs.
```
dobsonj, line 160: typo: this?

dobsonj (commented):
Couple of minor nits, but overall it looks fine to me.

enhancements/storage/csi-migration.md
bertinatto: Fixed

enhancements/storage/csi-migration.md
bertinatto: Fixed

enhancements/storage/csi-migration.md
```diff
@@ -133,12 +136,36 @@ spec:
...
+* If there are **no** pre-existing SCs in the cluster, then it creates a CSI SC **and** set it as the default.
```
gnufied, line 161: Can you clarify which "type" of SC it will look for?

jsafrane: I would rephrase it to:

* If there are **no** pre-existing **default** SCs in the cluster, then it creates a CSI SC **and** set it as the default.
* Otherwise just create the CSI SC, but don't make it default.

bertinatto: Done, but I think the **default**  above is unnecessary noise (if the are no pre-existing SCs, there can be no default SC anyway).

gnufied: But, there can be a case where there are existing SCs but they are not default and if we are going to create CSI storageclass then it should be default. So that is why we need to mention the default part.

jsafrane:
lgtm-ish

enhancements/storage/csi-migration.md
```diff
@@ -133,12 +136,36 @@ spec:
...
+  * This controller will make sure that a CSI SC always exists.
+  * And it will set that SC as the default one when appropriate.
+  * Also, it will make sure there is only **one** default SC.
+
+In a more detailed way, this is how the new controller will operate and determine whether the CSI SC should be the default one:
+
+* On every sync, this controller fetches all existing SCs.
+* If there are **no** pre-existing **default** SCs in the cluster, then it creates a CSI SC **and** set it as the default.
+  * Otherwise, create the CSI SC but do not make it default.
+* If there are pre-existing SCs in the cluster and more than one is set as default, it will set an appropriate message in the CR asking the user to make amendments.
```
jsafrane, line 163: cc @RomanBednar, we did not discuss this part. But it would be a trivial addition to the controller that you're going to write.

enhancements/storage/csi-migration.md
bertinatto: IMO we shouldn't make it default in that case. Otherwise, this would not only be a behavior change, but it could also break some use cases.

For example, a cluster admin set all SCs as **non** default to prevent users from using dynamic provisioning. I recall we even recommended this to customers, since we [intentionally don't overwrite the existing SC annotations](https://github.com/bertinatto/cluster-storage-operator/blob/f8eda8eeaba9ec2f275431e95555fa05a8c6fa81/pkg/operator/defaultstorageclass/controller.go#L173). What do you think? I'm really bad at taking notes during meetings, so I might have missed something (forgive me if that's the case).

gnufied: Yes - In that case we aren't creating any new SCs though. So that is what we talked about in the meeting, *only if* we are creating *new* SCs then we should mark it as default if there is no existing default exists in the cluster.

bertinatto: Oh, I see... thanks for clarifying

jsafrane:
/lgtm
/approve
From storage team perspective

jsafrane:
/hold cancel
We decided not to update to the latest template, it would be a too big change.

enhancements/storage/csi-migration.md
RomanBednar: @jsafrane noted, thanks

PR #1063, opened 2022-03-17
add stage-based exit codes to the openshift-installer

deads2k:
Describing the limitations on exit-codes from openshift-installer, where they could be used, what they should not be used for, and how this doesn't preclude producing a different interaction later.

@patrickdillon @sdodson @staebler 
@dgoodwin do you know Ken's github name?

### DIFF
--- a/enhancements/installer/coarse-grained-exit-codes.md
+++ b/enhancements/installer/coarse-grained-exit-codes.md
@@ -1,0 +1,162 @@
+---
+title: coarse-grained-exit-codes
+authors:
+  - "@deadsk2"
+reviewers:
+  - "@staebler"
+approvers:
+  - "@pdillon"
+  - "@sdodson"
+api-approvers: # in case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers)
+  - TBD
+creation-date: 2022-03-16
+last-updated: 2022-03-16
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - TBD
+see-also:
+replaces:
+superseded-by:
+---
+
+# Coarse Grained Exit Codes from the Installer
+
+## Summary
+
+`openshift-installer` callers will now get more granular exit codes.
+Callers of `openshift-installer` can expect non-zero exit codes *that may change over time* as
+the `openshift-installer` evolves over time.
+`openshift-installer` provides *no guarantee* that exit codes will not change as result of changes like
+greater granularity, lower granularity, re-organization, or other needs.
+`openshift-installer` *does guarantee* that a particular exit code will not be re-used to have a
+different meaning in the same y-level version.
+`openshift-installer` will make reasonable efforts to avoid re-using a particular exit code to have
+different meanings across y-level versions, but for sufficiently compelling reasons may do so.
+
+## Motivation
+
+Automated categorization of failures can assist in efforts like
+1. reliability determination of provision on various clouds
+2. directing effort to proper teams depending on failure modes
+3. tracking of which installation stages have problems on various platforms
+4. automated retry of infrastructure failures by CI step registry
+
+These are reasons why the data is useful and by providing it, other teams will be able to build these
+secondary system.
+Other teams *must properly handle previously unknown exit codes*.
+
+### Goals
+
+1. Coarse grained (order of 10 or less values), stage-based exit codes to assist in categorizing
+   failures, but not debugging failures.
+
+### Non-Goals
+
+1. Detailed, machine-parseable, cause-based failure information is outside the scope of this enhancement.
+
+## Proposal
+
+The installer knows what stage of the installation has failed when it exits.
+The stages include
+1. install-config handling
+2. infrastructure creation
+3. bootstrapping
+4. wait-for-cluster-install
+5. destroy
+6. everything else
+
+The installer will not introspect errors to provide deeper analysis, but these stages (and whatever future
+ones the install team identifies) may have unique exit codes when they fail.
+Categorizing failures in this way will allow machine categorization of the type of failure by other systems.
+The CI step registry is a good example of a consumer, but issuing a destroy after an infrastructure creation
+and retrying the `openshift-install` seems like a fairly common use-case since infrastructure creation
+failures often happen for environmental reasons and are reasonably likely to succeed on a retry.
+
+### Known Exit Codes
+
+Key things to remember:
+1. Callers of `openshift-installer` can expect non-zero exit codes *that may change over time* as
+   the `openshift-installer` evolves over time.
+2. `openshift-installer` provides *no guarantee* that exit codes will not change as result of changes like
+   greater granularity, lower granularity, re-organization, or other needs.
+3. `openshift-installer` *does guarantee* that a particular exit code will not be re-used to have a
+   different meaning in the same y-level version.
+4. `openshift-installer` will make reasonable efforts to avoid re-using a particular exit code to have
+   different meanings across y-level versions, but for sufficiently compelling reasons may do so.
+5. Consumers must properly handle previously unknown exit codes.
+
+| Stage | Exit Code |
+| --- | --- | 
+| Generic | whatever other value is produced |
+| infrastructure creation | 3 |
+| bootstrapping | 4 | 
+| wait-for-cluster-install | 5 |
+| install-config verification | TBD |
+
+### User Stories
+
+### API Extensions
+
+### Implementation Details/Notes/Constraints [optional]
+
+Approval of the implementation lies with the installer team, but [openshift/installer#5702](https://github.com/openshift/installer/pull/5702)
+demonstrates using `errors.Unwrap` and logical `errors.Is` to keep all return code logic contained in the command package.
+This keeps exit code logic from leaking into the installer library itself.
+
+### Risks and Mitigations
+
+## Design Details
+
+### Open Questions [optional]
+
+### Test Plan
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+
+#### Tech Preview -> GA
+
+#### Removing a deprecated feature
+
+### Upgrade / Downgrade Strategy
+
+The `openshift-installer` is only excecuted once and does not leave any persisted data
+in the cluster itself, so there is not an upgrade/downgrade concern for changing
+existing non-zero return codes into other non-zero return codes.
+
+### Version Skew Strategy
+
+The `openshift-installer` is only excecuted once and does not leave any persisted data
+in the cluster itself, so there is no version skew concern for changing
+existing non-zero return codes into other non-zero return codes.
+
+Consumers that installer multiple levels of `openshift-installer` may see exit codes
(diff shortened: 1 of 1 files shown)


dgoodwin:
Ken is @xueqzhan

enhancements/installer/coarse-grained-exit-codes.md
```diff
@@ -0,0 +1,159 @@
...
+| infrastructure creation | 3 |
+| bootstrapping | 4 | 
+| wait-for-cluster-install | 5 |
+| destroy | |
+
+### User Stories
+
+### API Extensions
+
+### Implementation Details/Notes/Constraints [optional]
```
patrickdillon, line 100: As shown by #5702, this handling is done by `logrus`. All relevant `logrus` calls are contained in the `cmd` package. Keeping the semantics of the error codes at this high-level does not require significant integration with other packages. Retrieving errors with more detail would require a much more sophisticated design.

patrickdillon, line 100: As shown by #5702, this handling is done by `logrus`. All relevant `logrus` calls are contained in the `cmd` package. Keeping the semantics of the error codes at this high-level does not require significant integration with other packages. Retrieving errors with more detail would require a much more sophisticated design.

enhancements/installer/coarse-grained-exit-codes.md
```diff
@@ -0,0 +1,159 @@
...
+5. Consumers must properly handle previously unknown exit codes.
+
+| Stage | Exit Code |
+| --- | --- | 
+| Generic | whatever other value is produced |
+| install-config verification | |
+| infrastructure creation | 3 |
+| bootstrapping | 4 | 
+| wait-for-cluster-install | 5 |
+| destroy | |
```
patrickdillon, line 94: Adding destroy was my suggestion, but on second thought a generic error for destroy is probably not worth the added complexity. These "stages"--except destroy--are all part of the `create cluster` flow. Destroy is only run explicitly so it seems that the Generic error code would capture just as much information.

patrickdillon, line 94: Adding destroy was my suggestion, but on second thought a generic error for destroy is probably not worth the added complexity. These "stages"--except destroy--are all part of the `create cluster` flow. Destroy is only run explicitly so it seems that the Generic error code would capture just as much infromation.

patrickdillon (commented):
Looks good to me. Left a few notes. 

Are there any particular clients we need to get feedback from?

enhancements/installer/coarse-grained-exit-codes.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Known Exit Codes
+
+Key things to remember:
+1. Callers of `openshift-installer` can expect non-zero exit codes *that may change over time* as
+   the `openshift-installer` evolves over time.
+2. `openshift-installer` provides *no guarantee* that exit codes will not change as result of changes like
+   greater granularity, lower granularity, re-organization, or other needs.
+3. `openshift-installer` *does guarantee* that a particular exit code will not be re-used to have a
+   different meaning in the same y-level version.
```
dhellmann, line 82: How will a user know what the exit codes for a given version mean? Docs only? Sub-command help? Other?

deads2k:
Updated for comments.



> Are there any particular clients we need to get feedback from?

An explicit ack from @xueqzhan and/or @dgoodwin .  Given flexibility on the consumer side (which they must logically have today), I think future changes can be addressed with minor tweaks (probably just table entries).

enhancements/installer/coarse-grained-exit-codes.md
```diff
@@ -0,0 +1,159 @@
...
+| Stage | Exit Code |
+| --- | --- | 
+| Generic | whatever other value is produced |
+| install-config verification | |
+| infrastructure creation | 3 |
+| bootstrapping | 4 | 
+| wait-for-cluster-install | 5 |
+| destroy | |
+
+### User Stories
```
dhellmann, line 95: If there is already a consumer in mind, describing that use case here would provide helpful context.

enhancements/installer/coarse-grained-exit-codes.md
deads2k: > How will a user know what the exit codes for a given version mean? Docs only? Sub-command help? Other?

Keeping this enhancement up to date seems reasonable.  I wouldn't get carried away with documenting what amount to best effort codes.

enhancements/installer/coarse-grained-exit-codes.md
deads2k: > If there is already a consumer in mind, describing that use case here would provide helpful context.

openshift CI.  Present and ready.

https://github.com/openshift/release/blob/master/ci-operator/step-registry/gather/must-gather/gather-must-gather-commands.sh#L26

xueqzhan:
> Updated for comments.
> 
> > Are there any particular clients we need to get feedback from?
> 
> An explicit ack from @xueqzhan and/or @dgoodwin . Given flexibility on the consumer side (which they must logically have today), I think future changes can be addressed with minor tweaks (probably just table entries).

This design sufficiently satisfies the requirement of our use case. Thanks @deads2k!

enhancements/installer/coarse-grained-exit-codes.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Proposal
+
+The installer knows what stage of the installation has failed when it exits.
+The stages include
+1. install-config handling
+2. infrastructure creation
+3. bootstrapping
+4. wait-for-cluster-install
+5. destroy
```
staebler, line 64: What is destroy? Do you mean the step of destroying the bootstrap? Or do yo mean destroying the entire cluster?

enhancements/installer/coarse-grained-exit-codes.md
staebler: Hive/OSD seem like prime candidates to use these codes, too. We should get some input from them about how the codes are defined.
/cc @2uasimojo

dgoodwin:
SGTM

enhancements/installer/coarse-grained-exit-codes.md
```diff
@@ -0,0 +1,162 @@
...
+3. `openshift-installer` *does guarantee* that a particular exit code will not be re-used to have a
+   different meaning in the same y-level version.
+4. `openshift-installer` will make reasonable efforts to avoid re-using a particular exit code to have
+   different meanings across y-level versions, but for sufficiently compelling reasons may do so.
+5. Consumers must properly handle previously unknown exit codes.
+
+| Stage | Exit Code |
+| --- | --- | 
+| Generic | whatever other value is produced |
+| install-config verification | |
```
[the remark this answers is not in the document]
deads2k: > Was the intention to use the error code of 2 here?

No.  This was a stage that @patrickdillon suggested, but I wasn't ready to assign it because it wasn't in the first cut of the PR.  I'll reshuffle to add this to the end.

enhancements/installer/coarse-grained-exit-codes.md
```diff
@@ -0,0 +1,162 @@
...
+## Proposal
+
+The installer knows what stage of the installation has failed when it exits.
+The stages include
+1. install-config handling
+2. infrastructure creation
+3. bootstrapping
+4. wait-for-cluster-install
+5. destroy
+6. everything else
```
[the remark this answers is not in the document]
deads2k: > Thanks for not sneaking in "cvo reports that the cluster installed but some cluster operators are now reporting degraded". 

I considered this and decided against it for a couple reasons.  I'll enumerate here and if you like, I'll add them to the KEP to help place further boundaries on what (in my opinion) is reasonable.
1. having "operators are degraded" or "operators are progressing" would be seemingly handy for consumers, but those conditions are not mutually exclusive, making the actual return values, one, the other, or both.
2. when the cluster fails and the installer can determine "operator state is bad" and communicates that as exit code 5, the caller can investigate the operator status themselves.

2uasimojo:
> Hive/OSD seem like prime candidates to use these codes, too. We should get some input from them about how the codes are defined.
> /cc @2uasimojo

I'm not seeing a strong motivation for hive to make use of this:
- The cross-version guarantees are explicitly limited (you said it twice; I'm thinking you might be serious).
- Hive doesn't predetermine the installer version it invokes.
- We could discover the version and map to the appropriate set of error codes as [suggested](https://github.com/openshift/enhancements/blob/be943b74508b64651964576ba6987f29edea09d2/enhancements/installer/coarse-grained-exit-codes.md#drawbacks)... **but** it would have to be worth the trouble.

Today we worry about two states: success and failure. On failure, we invoke a destroy and (maybe) try again from the beginning. We detect and categorize failures by scraping the logs, and expose some logic allowing the consumer to configure retries based on regex-matching those scrapings. That's brittle, but finer-grained than what these coarse exit codes would support, and can be adapted/adjusted by config (as opposed to code) in effectively realtime.

If this KEP _were_ a thing, there are a couple of hive use cases I can see being possible -- though again, perhaps not worth the trouble of implementing:
- Today when we invoke that destroy on failure, there are cases where the destroy fails repeatably and we're stuck. We've been told to file those as bugs against the installer (destroy should be idempotent). However, if we got the "install config processing" exit code, we could be "sure" there's nothing to clean up, and could simply bail without bothering to invoke the destroy.
- If I get an error code in one of the later stages, can I reinvoke the installer and have it idempotently pick up where it left off? That would be a time-saver for sure. Though again, having to understand which exit codes that applies to, for which versions of the installer, might be more trouble than it's worth to save that time.

/cc @akhil-rane @abutcher @suhanime @newtonheath @gregsheremeta in case your opinions should differ significantly from mine :)

deads2k:
> Hive/OSD seem like prime candidates to use these codes, too. We should get some input from them about how the codes are defined.
/cc @2uasimojo 

Looks like they are inclined to pass, but the TRT/CI use-case remains, so there's still an interested and ready consumer with a PR and staffing to complete and maintain the work with no conflict with future hive desires.

enhancements/installer/coarse-grained-exit-codes.md
sdodson: Maybe human readable stderr output like:
`exit(1) -- Failed to parse install-config`

enhancements/installer/coarse-grained-exit-codes.md
```diff
@@ -0,0 +1,162 @@
...
+   different meanings across y-level versions, but for sufficiently compelling reasons may do so.
+5. Consumers must properly handle previously unknown exit codes.
+
+| Stage | Exit Code |
+| --- | --- | 
+| Generic | whatever other value is produced |
+| infrastructure creation | 3 |
+| bootstrapping | 4 | 
+| wait-for-cluster-install | 5 |
+| install-config verification | TBD |
```
sdodson, line 93: Whenever we turn this TBD into something else can we get rid of the word verification unless we envision tying this to things like quota verification, etc. I don't know if this is just limited to ensuring inputs conform to the install-config api or more than that.

sdodson:
/lgtm
I don't see anything that should inhibit implementation

PR #1064, opened 2022-03-18
Add enhancement proposal for SDN live migration

pliurh:
Signed-off-by: Peng Liu <pliu@redhat.com>

### DIFF
--- a/enhancements/network/sdn-live-migration.md
+++ b/enhancements/network/sdn-live-migration.md
@@ -1,0 +1,466 @@
+---
+title: sdn-live-migration
+authors:
+  - "@pliurh"
+reviewers:
+  - "@danwinship"
+  - "@trozet"
+  - "@dcbw"
+  - "@russellb"
+approvers:
+  - "@danwinship"
+  - "@trozet"
+  - "@dcbw"
+  - "@russellb"
+api-approvers:
+  - "@danwinship"
+  - "@trozet"
+  - "@dcbw"
+  - "@russellb"
+creation-date: 2022-03-18
+last-updated: 2023-09-25
+tracking-link:
+  - https://issues.redhat.com/browse/SDN-2612
+status: implementable
+---
+
+# SDN Live Migration
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+
+## Summary
+Migrating the CNI network provider network of a running cluster from
+OpenShift SDN to OVN Kubernetes without service interruption. During the
+migration, we will partition the cluster into two sets of nodes controlled by
+different network plugins. We will utilize the Hybrid overlay feature of
+OVN Kubernetes to connect the networks of the two CNI network plugins. So that
+pods on each side can still talk to pods on the other side.
+
+## Motivation
+
+For some Openshift users, they have very high requirements on service
+availability. The current SDN migration solution, which will cause a service
+interruption, is not acceptable.
+
+### Goals
+
+- Migrate the cluster network provider from OpenShift SDN to OVN Kubernetes for
+  an existing cluster.
+  - The OVN Kubernetes will be running in the IC multi-zone mode.
+- This solution will work on all platforms managed by the SD team.
+  - ARO (Azure Red Hat OpenShift)
+  - OSDv4 (OpenShift Dedicated) on GCP
+  - OSDv4 (OpenShift Dedicated) on AWS
+  - ROSA Classic (Red Hat OpenShift Service on AWS)
+- This is an in-place migration without requiring extra nodes.
+- The impact on workload of the migration will be similar to that of an OCP
+  upgrade.
+- The solution can work at scale, e.g., in a large cluster with 100+ nodes.
+- The migration operation will be able to be rolled back if needed.
+- The migration is fully automated with zero human interaction
+- The migration shall be triggered by a declarative network API
+- There shall be failure indicator and corresponding documentation on what to do
+  next.
+
+### Non-Goals
+
+- Support for migration to other network providers
+- The necessary GUI change in Openshift Cluster Manager
+- Ensure the following features remain working during a live migration:
+  - Egress IP
+  - Egress Router
+  - Multicast
+- Migration from SDN Multi-tenant mode
+- Support Hypershift clusters
+
+## Pre-requisites
+
+- OVN-IC multi-zone mode in OCP is dev-complete.
+
+## Proposal
+
+The key problem with doing a live SDN migration is that we need to maintain the
+connectivity of the cluster network during the migration when pods are attached
+to different networks. We propose to utilize the OVN Kubernetes hybrid overlay
+feature to connect the networks owned by OpenShift SDN and OVN Kubernetes.
+
+- We will run different plugins on different nodes, but both plugins will know
+  how to reach pods owned by the other plugin, so all pods, services, etc.
+  remain connected.
+- During migration, CNO will take original-plugin nodes one by one and convert
+  them to destination-plugin nodes, rebooting them in the process.
+- The cluster network CIDR will remain unchanged, as will the node host subnet
+  of each node.
+- NetworkPolicy will work correctly throughout the migration.
+
+### Limitations
+
+- The following features, which are only supported by OpenShift SDN but not by
+  OVN Kubernetes, will stop working when the migration is started.
+  - Multitenant Isolation
+- The Egress Router feature is supported by both OpenShift SDN and OVN
+  Kubernetes, but with different designs and implementations. Therefore, even
+  with the same name, it has different APIs and configuration logic between
+  OpenShift SDN and OVN Kubernetes. Also, the supported platforms and modes are
+  different between the two network providers. So users need to evaluate before
+  conducting the live migration and have to migrate the configuration manually
+  after the migration is complete.
+- We have supported automated conversion for the configuration of the following
+  features. When the migration is done, these features will remain functional.
+  - Multicast
+  - Egress IP
+  - Egress Firewall
+- During the migration, when the cluster is running with both OVN Kubernetes and
+  OpenShift SDN,
+  - Multicast and Egress IP will be temporarily disabled for both CNIs.
+  - Egress Firewall shall remain functional, as the implementation of this
+    feature is per-node-based in both CNIs.
+
+### User Stories
+
+The service delivery (SD) team (which manages OpenShift services ARO, OSD, ROSA)
+has a unique set of requirements around downtime, node reboots, and a high
+degree of automation. Specifically, SD needs a way to migrate their managed
+fleet in a way that is no more impactful to the customer's workloads than an OCP
+upgrade and that can be done at scale in a safe, automated way that can be made
+self-service and not require SD to negotiate maintenance windows with customers.
+The current migration solution needs to be revisited to support these
+(relatively) more stringent requirements.
+
+### Risks and Mitigations
+
+## Design Details
+
+The existing OVN Kubernetes hybrid overlay feature was developed for hybrid
+Windows/Linux clusters. Each OVN Kubernetes node manages an external-to-OVN OVS
+bridge, named br-ext, which acts as the VXLAN source and endpoint for packets
+moving between pods on the node and their cluster-external destination. The
+br-ext SDN switch acts as a transparent gateway and routes traffic towards
+Windows nodes.
+
+In the SDN live migration use case, we can enhance this feature to connect the
+nodes managed by different CNI plugins. To minimize the implementation effort
+and maintainability of the code, we will try to reuse the hybrid overlay code
+and only make necessary changes to both CNI plugins. OVN Kubernetes nodes
+(OVN-IC zones) will have full-mesh connections to all OpenShift SDN Nodes
+through VXLAN tunnel.
+
+On the OVN Kubernetes side, all the cross-CNI traffic shall follow the same path as the current hybrid overlay implementation. For OVN Kubernetes, we need to make the following enhancements:
+
+1. We need to prevent OVN Kubernetes from allocating a subnet for each host. We
+   need to reuse the host subnet allocated by OpenShift SDN.
+2. We need to modify OVN Kubernetes to allow overlapping between the cluster
+   network and the Hybrid overlay CIDR.
+3. OVN Kubernetes shall be able to handle adding/removing hybrid overlay nodes
+   on the fly.
+4. We need to allow `hybrid-overlay-node` to run on the Linux nodes using
+   OpenShift SDN as the CNI plugin; currently, it is designed to only run on
+   Windows nodes. It is responsible for:
+   - collecting the MAC address of the host primary interface and setting the
+    node annotation `k8s.ovn.org/hybrid-overlay-distributed-router-gateway-mac`.
+   - removing the pod annotations added by OVN Kubernetes for pods running on
+    the local node.
+
+On the OpenShift SDN side, when a node is converted to OVN Kubernetes, it will
+be almost transparent to the control-plane of OpenShift SDN. But we still need
+to introduce a 'migration mode' for OpenShift SDN by:
+
+1. Change ingress NetworkPolicy processing to be based entirely on pod IPs
+   rather than using namespace VNIDs, since packets from OVN nodes will have the
+   VNID 0 set.
+2. To be compatible with Windows node VXLAN implementation, OVN Kubernetes
+   hybrid overlay uses the peer host interface MAC as the VXLAN inner dest MAC
+   for egress traffic. Therefore, when packets arrive at the br0 of the SDN
+   node, they cannot be forwarded to the pod interface correctly. We need to add
+   a flow to change the dst MAC to the pod interface MAC for such ingress
+   traffic.
+
+### MTU Consideration
+
+There are two related but distinct MTU values to consider: the hardware MTU and
+the cluster MTU. The cluster MTU is the MTU value for pod interfaces. It is
+always less than your hardware MTU to account for the cluster network overlay
+overhead. The overhead is 100 bytes for OVN Kubernetes and 50 bytes for SDN. For
+example, if the hardware MTU is 1500, then the auto-probed cluster MTU would be
+1400 for OVN Kubernetes and 1450 for OpenShift SDN. During the migration, this
+MTU mismatch would break some cross-CNI traffic.
+
+To resolve this issue, during the live migration process, CNO will update the
+routable MTU to make sure the 2 CNI shares the same overlay MTU.
+
+### The Traffic Path
+
+#### Packets going from OpenShift SDN to OVN Kubernetes
+
+On the SDN side, it doesn't need to know if the peer node is a SDN node or an
+OVN node. We reuse the existing VXLAN tunnel rules on the SDN side.
+- Egress NetworkPolicy rules and service proxy happen as normal.
+- When the packet reaches table 90, it will hit a "send via vxlan" rule that was
+  generated based on a HostSubnet object.
+
+On the OVN side:
+- OVN accepts the packet via the VXLAN tunnel, ignores the VNID set by SDN, and
+  then just routes it normally.
+- Ingress NetworkPolicy processing will happen when the packet reaches the
+  destination pod's switch port, just like normal.
+  - Our NetworkPolicy rules are all based on IP addresses, not "logical input
+    port", etc., so it doesn't matter that the packets came from outside OVN and
+    have no useful OVN metadata.
+
+#### Packets going from OVN Kubernetes to OpenShift SDN
+
+On the OVN side:
+- The packet just follows the same path as the hybrid overlay.
+
+On the SDN side:
+- We have to change ingress NetworkPolicy processing to be based entirely on pod
+  IPs rather than using namespace VNIDs since packets from OVN nodes won't have
+  the VNID set. There is already code to generate the rules that way, though,
+  because egress NetworkPolicy already works that way.
+
+### Workflow Description
+
+1. The admin kicks off the migration process by updating the custom resource
+   `network.config`:
+   
+   ```bash
+   $ oc patch Network.config.openshift.io cluster --type='merge' --patch '{"metadata":{"annotations":{"network.openshift.io/network-type-migration":""}},"spec":{"networkType":"OVNKubernetes"}}'
+   ```
+   
+   CNO will set the `NetworkTypeMigrationInProgress` condition to TRUE in the
+   `status` of the network.config CR. If there are any unsupported features
+   (refer to the [Limitations](#limitations) section) enabled in the cluster,
+   CNO will set an the `NetworkTypeMigrationInProgress` condition to `False`
+   with warning message and won't start the live migration. User shall disable
+   these unsupported features before starting the live migration.
+
+2. CNO will check the current hardware MTU and the cluster MTU to determine a
+   routable MTU during the migration. e.g.
+   - If the current hardware MTU is 1500, and the cluster MTU is not specified
+     explicitly. CNO will set `.spec.migration.mtu` of network.operator CR with
+     MTU of OVN-K, 
+     - `.spec.migration.mtu.network.from` 1450 `to` 1400
+     - `.spec.migration.mtu.machine.from` 1500 `to` 1500
+   - If the current hardware MTU is 9000, and the cluster MTU is specified
+     explicitly to 1250 for OpenShiftSDN. CNO will set `.spec.migration.mtu` of
+     network.operator CR with MTU of OVN-K,
+     - `.spec.migration.mtu.network.from` 1250 `to` 1200
+     - `.spec.migration.mtu.machine.from` 9000 `to` 9000
+   
+   This will trigger MCO to apply a new machine config to each MCP and reboots
+   all the nodes.
+
+3. CNO watches the MCP, wait until the MCs with routable MTU is applied to all
+   the nodes. CNO then sets the `NetworkTypeMigrationMTUReady`
+   condition to `TRUE` in the status of the `network.config` CR.
+
+4. CNO patches the network.operator CR with
+   `{"spec":{"migration":{"networkType":"OVNKubernetes","mode":"Live"}}}`
+
+5. CNO then redeploys openshift-sdn in migration mode. It
+   will add a condition check to the wrapper script of the sdn container to see
+   if the bridge `br-ex` exists on the node. If `br-ex` exists, it means the
+   node has already been updated by MCO, therefore it's ready for running OVN
+   Kubernetes. The sdn pod would just sleep infinity rather than actually
+   launching the `openshift-sdn-node` process.
+
+   CNO will also deploy OVN Kubernetes to the cluster with hybrid overlay
+   enabled. To avoid racing, the host subnet allocation will be disabled in
+   OVN Kubernetes.
+      
+   The ovnkube-node container wrapper script will be rendered with the logic for
+   migration:
+
+   - If `br-ex` doesn't exist, it means the node has not yet been updated by
+     MCO, so instead of starting the `ovnkube-node` process, it would run the
+     `hybrid-overlay-node` process. `hybrid-overlay-node` can add the node
+     annotation `k8s.ovn.org/hybrid-overlay-distributed-router-gateway-mac`,
+     which is required by hybrid overlay.
+
+     The script will also add the node annotation
+     `k8s.ovn.org/hybrid-overlay-node-subnet`, according to the HostSubnet CR of
+      this node.
+
+   - If `br-ex` does exist, it means the node has been updated by MCO. So the
+     `ovnkube-node` can be started.
+
+     The script will also add the node annotation `k8s.ovn.org/node-subnets`
+     according to the HostSubnet CR of this node.
+     
+   CNO will wait until OVN-K is deployed in the cluster and set the
+   `NetworkTypeMigrationTargetCNIAvailable` condition to `TRUE` in the status of
+   the network.config CR.
+
+6. CNO updates the `.status.migration.networkType` of the network.config CR.
+   This change will trigger MCO to apply a new MachineConfig to each MCP:
+
+   - MCO will render a new MachineConfig for each MCP.
+   - MCO will cordon, drain and reboot the node.
+   - When the node is up, the `br-ex` bridge will be created. It means the node
+     is ready for OVN Kubernetes. So the local zone OVN Kubernetes components
+     can be started. The ovnkube-node container wrapper script will also add the
+     node annotation `k8s.ovn.org/node-subnets`, according to the HostSubnet CR
+     of this node. On the local node, OVN Kubernetes will add all SDN nodes as
+     hybrid overlay nodes to the local zone and all OVN nodes as OVN-IC remote
+     zones. On a remote OVN node, OVN Kubernetes will remove a hybrid overlay
+     node and add an OVN-IC remote zone. Pods will be recreated on the node
+     using ovn-kubernetes as the default CNI plugin.
+   - MCO will uncordon the node. New pods can be scheduled to this node.
+
+   The above process will be repeated for each node until all the nodes have
+   been applied to the new MachineConfig and converted to OVN Kubernetes.
+
+7. CNO watches the MCPs, wait until the new MCs is applied to all the nodes. Set
+   the `NetworkTypeMigrationTargetCNIInUse` condition to `TRUE` in the status of
+   the network.config CR.
+
+8. CNO removes the `spec.migration` field of network.operator CR. It will
+   trigger the CNO to:
+
+   - delete the openshift-sdn DaemonSets and the related resources
+     (CustomResources, ConfigMaps, etc.).
+   - redeploy ovn-kubernetes in "normal" mode (no migration mode script, hybrid
+     overlay disabled, host subnet allocation enabled etc.).
+   - remove the hybrid overlay related labels and node hybrid overlay
+     annotations from the nodes.
+
+9. CNO wait until OVN-K is redeployed Set the
+   `NetworkTypeMigrationOriginalCNIPurged` condition to `TRUE` in the status of
+   the network.config CR. Then set the `NetworkTypeMigrationInProgress`
+   condition to `FALSE` with the reason `NetworkTypeMigrationCompleted` in the
+   status of the network.config CR. This indicates that the migration is
+   completed.
+
+### API
+
+For `Network.config.openshift.io` CRD, we will use the CR annotation
+`network.openshift.io/network-type-migration:` to gate the modification to
+`spec.networkType` field. If the annotation absents (default), CNO will block
+the change to take effect. If it exists, CNO takes it as the trigger of starting
+SDN live migration.
+
+The following conditions will be added to the `.status` of the
+`Network.config.openshift.io` CR:
+
+- NetworkTypeMigrationInProgress
+- NetworkTypeMigrationMTUReady
+- NetworkTypeMigrationTargetCNIAvailable
+- NetworkTypeMigrationTargetCNIInUse
+- NetworkTypeMigrationOriginalCNIPurged
+
+These conditions will be utilized by CNO to assess the present status of live
+migration. These conditions besides `NetworkTypeMigrationInProgress` will be set
+to `FALSE` when a live migration is initiated and reset to `UNKNOWN` once the
+live migration is successfully finished. Users can also employ these conditions
+to track the progress of live migration and troubleshoot any errors that may
+occur.
+
+The `LiveMigrationProgressing` condition will be set to `TRUE` when a live
+migration is initiated. It will be set to `FALSE` with with the reason
+`NetworkTypeMigrationCompleted` when the live migration is successfully
+completed. If the `LiveMigrationProgressing` condition remains stuck in `FALSE`,
+and there is an error in the reason field. it indicates an error in the live
+migration process. This serves as an indicator to the user that they need to
+address and resolve the issue before CNO can rollback or proceed the live
+migration.
+
+A new field `spec.migration.mode` will be introduced to the CRD
+`Network.operator.openshift.io`. The supported values are `Live` and `Offline`.
+If unset, the default mode is Offline.
+
+```json
+{ 
+  "spec": { 
+    "migration": {
+      "networkType": "OVNKubernetes",
+      // mode indicates whether we're doing a live migration or not.
+      "mode": "Live"
+    }
+  } 
+}
+```
+
+### Rollback
+
+Users shall be able to rollback to openshift-sdn after the migration is
+complete. The migration is bidirectional, so users can follow the same procedure
+as above to conduct the rollback.
+
+The HostSubnet CRs will be removed when the migration is done. So during the
+rollback, the HostSubnet CRs will be recreated by the wrapper script of
+ovnkube-node according to the node annotation `k8s.ovn.org/node-subnets`.
+
+### Lifecycle Management
+
+This is a one-time operation for a cluster; therefore, there is no lifecycle
+management.
+
+### Test Plan
+
+We need to setup a CI job that can run the sig-network test against a cluster
+that is in an intermediate state of live migration; specifically, some nodes are
+running with OVN Kubernetes and the rest are running with OpenShift SDN. So that
+we can ensure the cluster network remains functioning during the migration.
+
+We also need to work closely with the Performance & Scale test team and the SD
+team to test the live migration on a large (100+ nodes) cluster.
+
+### Graduation Criteria
+
+Graduation criteria follows:
+
+#### Dev Preview -> Tech Preview
+
+- End user documentation, relative API stability
+- Sufficient test coverage
+- Gather feedback from users rather than just developers
+
+#### Tech Preview -> GA
+
+- More testing (upgrade, scale)
+- Add gating CI jobs on the relevant GitHub repos
+- Sufficient time for feedback
+
+#### Removing a deprecated feature
+N/A
+
+### Upgrade / Downgrade Strategy
+
+This is a one-time operation for a cluster; therefore, there is no upgrade /
+downgrade strategy.
+
+### Drawbacks
+N/A
+
+### Version Skew Strategy
+N/A
+
+### API Extensions
+N/A
+
+### Operational Aspects of API Extensions
+N/A
+
+#### Failure Modes
+N/A
+
+#### Support Procedures
+N/A
+
+## Implementation History
+N/A
+
+## Alternatives
+
+Instead of switching the network provider for an existing cluster, we can spin
+up a new cluster and move the workload to it.
+
+## Infrastructure Needed
+N/A


enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,358 @@
...
+  `machineconfiguration.openshift.io/state` node annotation. It will be set by
+  MCD if the reboot is blocked. It indicts that draining node is done, and the
+  rebooting is paused.
+
+When we kick off the migration from CNO, CNO will update the
+`status.networkType` field of the `network.config` CR. It will trigger MCO to
+apply new MachineConfig to each node:
+1. MCO will 
+   - rerender the MachineConfig for each node
+   - try to cordon and drain the node one node
```
jewzaam, line 233:
```suggestion
   - try to cordon and drain nodes based on MachineConfigPools.spec.maxUnavailable
```

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,358 @@
...
+2. CNO finds a node is being drained by MCO by watching the MCO node
+   annotations, it will:
+   - pause the node reboot by annotate the node with
+     `machineconfiguration.openshift.io/rebootPausedBy: network-operator`.
```
jewzaam, line 237: Risk here with watching for a drain vs proactively annotating nodes.  Possibility if there are issues in CNO the pause annotation won't be set before reboot.  Can the annotation be set before this kicks off?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,358 @@
...
+
+The above process will be repeated for each node, until all the nodes have been
+applied to the new MachineConfig and converted to OVN Kubernetes.
+
+#### Migration Cleanup
+
+Once migration is complete, CNO will:
+- delete the openshift-sdn DaemonSets
+- redeploy ovn-kubernetes in “normal” mode (no migration mode config, no node
+  affinity).
```
jewzaam, line 261: Nodes are uncordoned at this point, so OVN is redeployed with workloads deployed to the node.  Is it possible to switch this over node-by-node in the migration steps above to reduce workload impacts?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,358 @@
...
+- Add CI job
+- Sufficient time for feedback
+
+#### Removing a deprecated feature
+N/A
+
+### Upgrade / Downgrade Strategy
+
+This is a one-time operation for a cluster, therefore no upgrade / downgrade
+strategy.
```
jewzaam, line 328: A stated goal is "The migration operation shall be able to be rolled back if needed." so how will the rollback be handled?  I don't see a better section to capture this in.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,358 @@
...
+    migration.network.openshift.io/plugin: OpenShiftSDN
+    k8s.ovn.org/hybrid-overlay-node-subnet: 10.129.0.0/23
+    ```
+
+   (CNO knows how to figure out the host subnet by looking at openshift-sdn
+   HostSubnet objects. When migrating in the reverse direction, it knows to look
+   at the ovn-kubernetes annotations.)
+
+3. CNO will redeploy the openshift-sdn DaemonSets, enabling “migration mode”,
+   and adding node affinity to the node DaemonSet:
```
jewzaam, line 185: What is the impact to the cluster as these updates happen?  If any possible disruption it should be called out in the Risks section above, even if it can be couched as no worse than any other DS rollout.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,358 @@
...
+The current migration solution needs to be revisited to support these
+(relatively) more stringent requirements.
+
+### Risks and Mitigations
+
+## Design Details
+
+The existing ovn-kubernetes hybrid overlay feature was developed for hybrid
+Windows/Linux clusters. Each ovn-kubernetes node manages an external-to-OVN OVS
+bridge, named br-ext, which acts as the VXLAN source and endpoint for packets
```
jewzaam, line 100: Is there an overhead / performance impact of using VXLAN and therefore an impact to workloads on the cluster while migration is in progress?

mwoodson: if so, is it temp? and if so, how long will the impact last?

enhancements/network/sdn-live-migration.md
danwinship: This step is about fixing the state of the `ovnkube-master` and `ovnkube-node` Deployment objects. While the Deployment controller will roll the update out node-by-node, there's nothing we can do specially, since we're just talking about about updating two API objects.

Also note that all of the nodes are already running ovn-kubernetes at this point, so this is just a question of restarting the existing ovn-kubernetes binaries on each node. It should be very quick/non-disruptive.

enhancements/network/sdn-live-migration.md
danwinship: openshift-sdn-node will be restarted on each node. The restart itself is immediate. After restarting, the node process will need to reload its caches, which may take a few seconds in large clusters (eg thousands of pods), during which time it would lag in processing updates to the cluster. (So, eg, a pod launched at this time may take an extra few seconds to become ready.)

In clusters with less than thousands of pods the restart is likely to be completely unnoticeable.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,358 @@
...
+
+At this point, neither plugin is willing to run on a node that has no
+“migration.network.openshift.io/plugin” label at all. This means that if a new
+node gets added to the cluster in the middle of the migration process, neither
+plugin will get deployed to it until after the migration of the older nodes is
+complete, and the node affinity is removed.
+
+#### Migration
+
+In order to run ovn-kubernetess in shared-gw mode, we need MCO to update the
```
cgwalters, line 217: OK a lot going on here.  Would it help if we had the SDN daemonset actually *live reconfigure* the service?  That's fully supported by systemd and the MCO.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,358 @@
...
+ovs-configuration.service. To apply the MachineConfig update, a reboot will be
+triggered by MCO. To coordinate the draining and rebooting between CNO and MCO,
+we need to modify MCO as following:
+- add a `machineconfiguration.openshift.io/rebootPausedBy: network-operator`
+  node annotation to MCO, so that CNO can block MCO from rebooting the node
```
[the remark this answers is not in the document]
cgwalters: I like that idea a lot better indeed.

Now, I do think eventually that the MCO should offer more control to administrators and 3rd party code.  See e.g. https://github.com/openshift/machine-config-operator/pull/2163

But short term, if we can avoid trying to "wedge" things in between the MCO changes that helps.

enhancements/network/sdn-live-migration.md
pliurh: I refactored the process, PTAL.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,358 @@
...
+“migration.network.openshift.io/plugin” label at all. This means that if a new
+node gets added to the cluster in the middle of the migration process, neither
+plugin will get deployed to it until after the migration of the older nodes is
+complete, and the node affinity is removed.
+
+#### Migration
+
+In order to run ovn-kubernetess in shared-gw mode, we need MCO to update the
+ovs-configuration.service. To apply the MachineConfig update, a reboot will be
+triggered by MCO. To coordinate the draining and rebooting between CNO and MCO,
```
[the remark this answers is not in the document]
pliurh: I refactored the process, PTAL.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,331 @@
...
+- We will run different plugins on different nodes, but both plugins will know
+  how to reach pods owned by the other plugin, so all pods/services/etc remain
+  connected.
+- During migration, CNO will take original-plugin nodes one by one and convert
+  them to destination-plugin nodes, rebooting them in the process.
+- The cluster network CIDR will not change. In fact, none of the node host
+  subnets will change.
+- NetworkPolicy will work correctly throughout the migration.
+
+### Limitations
```
russellb, line 79: Maybe describe the features called out in "non-goals" in this section in more detail?

pliurh: added more description below. PTAL

russellb: Thanks.

What if the implementation in the CNO did validations before any migration begins?  It could check the cluster for anything it knows may not work?  I know some things are difficult to check for (like multicast traffic), but some things should be relatively straight forward if it's just looking at the kube API.

pliurh: I think it is a good idea. I added it to the `Migration Setup` section. Thanks.

enhancements/network/sdn-live-migration.md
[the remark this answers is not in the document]
pliurh: I took @danwinship 's suggestion and updated the document.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,369 @@
...
+On the OVN side:
+- OVN accepts the packet via the VXLAN tunnel, ignore the VNID set by SDN and
+  then just routes it normally.
+- Ingress NetworkPolicy processing will happen when the packet reaches the
+  destination pod’s switch port, just like normal.
+  - Our NetworkPolicy rules are all based on IP addresses, not “logical input
+    port”, etc, so it doesn’t matter that the packets came from outside OVN and
+    have no useful OVN metadata.
+
+#### Packets going from OVN Kubernetes to OpenShift SDN
```
dcbw, line 179: We'll also have to modify the hybrid overlay code to handle SDN nodes having HostSubnets within the ClusterCIDR; because Windows HO doesn't do that. So I guess ovnkube will have to learn about HostSubnet resources, or maybe the SDN code could annotate its nodes with the right bits so ovnkube can grab them subnets.

pliurh: Instead of letting ovnkube-master read the HostSubnet CR, I plan to run the `hybrid-overlay-node` binary on the SDN nodes. It will be responsible for reading the HostSubnets CR and annotating the node. So it won't affect the current Windows HO logic.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,369 @@
...
+
+On the ovn side:
+- The packet just follow the same path of hybrid overlay. The packet just has to
+  get routed out the VXLAN tunnel with VNID 0.
+
+On the sdn side:
+- We have to change ingress NetworkPolicy processing to be based entirely on pod
```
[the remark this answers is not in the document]
pliurh: I am not quite sure how much effort it would be. According to @danwinship 's comments, it should not be too much.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,369 @@
...
+  subnets will change.
+- NetworkPolicy will work correctly throughout the migration.
+
+### Limitations
+
+- Multicast may not work during the migration when the CNI plugins are running
+  in parallel, but it shall work after the migration is complete.
+- The following features which are only supported by OpenShift SDN but not by
+  OVN Kubernetes, will stop working when the migration is started.
+  - Multitenant Isolation
```
dcbw, line 96: @pliurh we probably want to block migration from SDN multitenant to OVNK, because I can't see how that would work at all during a migration.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,369 @@
...
+   is designed to only run on windows nodes. 
+
+On the OpenShift SDN side, when a node is converted to OVN Kubernetes,
+it shall be almost transparent to the control-plane. But we need to introduce a
+'migration mode' for OpenShift SDN by:
+
+1. Change ingress NetworkPolicy processing to be based entirely on pod IPs
+   rather than using namespace VNIDs, since packets from ovn nodes will have
+   the VNID 0 set.
+2. To be compatible with Windows node VXLAN implementation, OVN-K hybrid overlay
```
dcbw, line 154: We could figure out a way to disable this for ovnk...

pliurh: I don't know how to make it work. Even we disable it on the OVN-K side, the egress packets still cannot have the real pod MAC. In OVN-K, for all the traffic to HO nodes, the traffic will come to the cluster_router first and then route to the local br-ext bridge. So when the packets arrive at the br-ext bridge, both the src and dst mac will be changed as we cross the node subnet.

Even if we somehow manage to set the egress MACs to the real ones, I assume it will require much more code changes than we do it on the SDN side.

BTW, here is the patch I prepared for SDN https://github.com/openshift/sdn/compare/master...pliurh:sdn:live-migration.

enhancements/network/sdn-live-migration.md
trozet: oh this is why you need to run hybrid-overlay-node on linux nodes. You should make that clear in my earlier comment that it is for linux nodes running openshift-sdn

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,377 @@
...
+   bridge `br-ex` exists on the node . If `br-ex` exists, it means the node has
+   already been updated by MCO, therefore it's ready for running OVN Kubernetes,
+   the sdn pod would just do "sleep infinity" rather than actually launching the
+   `openshift-sdn-node` process.
+
+3. CNO will also deploy OVN Kubernetes to the cluster with hybrid overlay
+   enabled. The ovnkube-node DaemonSet wrapper script also has the logic to
+   check if the bridge br-ex exists on the node. If `br-ex` doesn't exist, it
+   means the node has yet been updated by MCO, so instead of starting the
+   `ovnkube-node` process, it would run `hybrid-overlay-node` process that
```
trozet, line 222: cool

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,377 @@
...
+   resource `network.operator`.
+
+2. CNO will redeploy the openshift-sdn in the __migration mode__. It will add a
+   condition check to the wrapper script of the sdn DaemonSet, to see if the
+   bridge `br-ex` exists on the node . If `br-ex` exists, it means the node has
+   already been updated by MCO, therefore it's ready for running OVN Kubernetes,
+   the sdn pod would just do "sleep infinity" rather than actually launching the
+   `openshift-sdn-node` process.
+
+3. CNO will also deploy OVN Kubernetes to the cluster with hybrid overlay
```
trozet, line 218: so at this step will ovnkube-master be brought up on the control nodes without blocking/waiting for anything?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,377 @@
...
+   ovn-kubernetes as the default CNI plugin.
+
+The above process will be repeated for each node, until all the nodes have been
+applied to the new MachineConfig and converted to OVN Kubernetes.
+
+#### Migration Cleanup
+
+Once migration is complete, CNO will:
+- delete the openshift-sdn DaemonSets
+- redeploy ovn-kubernetes in "normal" mode (no migration mode config, no node
```
trozet, line 271: but still hybrid overlay enabled?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,377 @@
...
+CR by adding:
+
+CNO will deploy both OpenShift SDN and OVN Kubernetes pods in 'migration' mode.
+
+```json
+{ 
+  "spec": { 
+    "migration": {
+      "networkType": "OVNKubernetes",
+      "isLive": true
```
trozet, line 287: what is the significance of "isLive"? Is this a new field with this enhancement?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,377 @@
...
+
+For some Openshift users, they have very high requirement on service
+availability. The current SDN migration solution, that will cause a service
+interruption, is not acceptable.
+
+### Goals
+
+- Migrate the cluster network provider from OpenShift SDN to OVN Kubernetes for a 
+  existing cluster.
+- This's in-place migration without requiring extra nodes.
```
trozet, line 56: nit: This is

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,377 @@
...
+  SDN and OVN Kubernetes. Also, the supported platforms and modes are different
+  between the two network providers. So users need to evaluate before conducting
+  the live migration, and have to migrate the configuration manually after the
+  migration is complete.
+- The following features are supported by both OpenShift SDN and OVN Kubernetes.
+  And we have support automated converting for the configuration. However, users
+  shall not expect these feature can function as normal during the live
+  migration phase.
+  - Egress IP
+  - Egress Firewall
```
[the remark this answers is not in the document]
pliurh: Yes. This is a 4.12 epic https://issues.redhat.com/browse/NP-20.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,377 @@
...
+On the OVN Kubernetes side, all the cross CNI traffic shall follow the same path
+of current hybrid overlay implementation. For OVN Kubernetes, we need to do
+following enhancements:
+
+1. We need to modify OVN-K to allow overlapping between the cluster network and
+   the Hybrid overlay CIDR. So that we can reuse the cluster network in the
+   migration.
+2. We need to modify OVN-K to allow modifying the hybrid overlay CIDR on the
+   fly.
+3. We need to allow `hybrid-overlay-node` running on linux nodes, currently it
```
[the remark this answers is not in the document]
pliurh: This is for the nodes using OpenShiftSDN as the CNI plugin. So no `ovnkube-node` process will be running on those nodes. We want to run hybrid-overlay-node binary on those nodes to get the `distributed-router-gateway-mac`. Here's the PR for ovn-k https://github.com/ovn-org/ovn-kubernetes/pull/3037.

enhancements/network/sdn-live-migration.md
pliurh: will do

enhancements/network/sdn-live-migration.md
pliurh: No, it doesn't need to. At this moment, `ovnkube-node` process won't be running on any node. We only have the ovn-k control-plane deployed.

enhancements/network/sdn-live-migration.md
pliurh: hybrid overlay disabled

dhellmann:
@pliurh this design was closed by the bot without merging but I see the ticket it is linked to is marked in progress. Should this design be fixed up and merged?

dhellmann:
This pull request is closed with lifecycle/rotten. The associated Jira ticket, SDN-2612, has status "In Progress". Should the PR be reopened, updated, and merged? If not, remove the lifecycle/rotten label to have this bot ignore it in the future.

dhellmann:
> This pull request is closed with lifecycle/rotten. The associated Jira ticket, SDN-2612, has status "In Progress". Should the PR be reopened, updated, and merged? If not, remove the lifecycle/rotten label to have this bot ignore it in the future.

I'll remove the lifecycle label so you don't get these automated messages any more.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,404 @@
...
+  - Multitenant Isolation
+- The Egress Router feature is supported by both OpenShift SDN and OVN
+  Kubernetes but with different design and implementation. Therefore even with
+  the same name, it have different API and configure logic between OpenShift
+  SDN and OVN Kubernetes. Also, the supported platforms and modes are different
+  between the two network providers. So users need to evaluate before conducting
+  the live migration, and have to migrate the configuration manually after the
+  migration is complete.
+- The following features are supported by both OpenShift SDN and OVN Kubernetes.
+  And we have support automated converting for the configuration. However, users
```
danwinship, line 113: How "automated" is automated? When does this happen? (The "Workflow Description" below doesn't mention it.)

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,404 @@
...
+#### Migration
+
+We can kick off the migration from CNO, by:
+
+```bash
+$ oc patch Network.config.openshift.io cluster --type='merge' --patch '{"spec":{"networkType":"OVNKubernetes"}'
+```
+
+CNO will update the `status.networkType` field of the `network.config` CR. It
+will trigger MCO to apply new MachineConfig to each node:
```
danwinship, line 258: what new MachineConfig?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,404 @@
...
+- delete the openshift-sdn DaemonSets
+- redeploy ovn-kubernetes in "normal" mode (no migration mode config, no node
+  affinity, hybrid overlay disabled).
+- remove the migration-related labels from the nodes
+
+### API
+
+A new field `spec.migration.isLive` will be introduced to the CRD
+`networks.operator.openshift.io`. Users can set this flag to conduct a live
+migration or a offline migration.
```
danwinship, line 295: Why do we want to continue supporting the offline migration?

(If we are keeping this field, I feel like `live` is a better name than `isLive`. It seems wrong to say "is" since the migration isn't actually happening yet at that point... it's more like `willBeLive`...)

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,404 @@
...
+### Lifecycle Management
+
+This is a one-time operation for a cluster, therefore no lifecycle management.
+
+### Test Plan
+
+We need to setup a CI job which can run the sig-network test against a cluster
+that is in an intermediate state of live migration, specifically some nodes are
+running with OVN Kubernetes, and the rests are running with OpenShift SDN. So
+that we can ensure the cluster network remains functioning during the migration.
```
danwinship, line 348: Depending on how long it takes to do a migration vs how long it takes to run the e2e suite, you could possibly have a job that just migrates the cluster back and forth until the e2e suite completes?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,404 @@
...
+
+- Support for migration to other network providers
+- The necessary GUI change in Openshift Cluster Manager.
+- Ensure the following features remain working during live migration:
+  - Egress IP
+  - Egress Router
+  - Egress Firewall
+  - Multicast
+- Migration from SDN Multi-tenant mode
+- Migrate to OVN Kubernetes single-zone IC mode.
```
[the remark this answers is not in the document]
pliurh: It means all nodes are in one zone. I am not sure if we will support this in OCP. If not, I can remove it. @tssurya

enhancements/network/sdn-live-migration.md
pliurh: Do we test the upgrade in this way?

enhancements/network/sdn-live-migration.md
pliurh: Initially, the live migration is only for clusters managed by the SD team. For other customers, they will keep using the offline migration, until we're confident to have live migration work on all platforms.

enhancements/network/sdn-live-migration.md
pliurh: The new MachineConfig is rendered for running OVN-K.

enhancements/network/sdn-live-migration.md
pliurh: This has already been implemented https://issues.redhat.com/browse/NP-20.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,404 @@
...
+to introduce a 'migration mode' for OpenShift SDN by:
+
+1. Change ingress NetworkPolicy processing to be based entirely on pod IPs
+   rather than using namespace VNIDs, since packets from ovn nodes will have
+   the VNID 0 set.
+2. To be compatible with Windows node VXLAN implementation, OVN-K hybrid overlay
+   uses the peer host interface MAC as VXLAN inner dest MAC for egress traffic.
+   Therefore when packets arrive at the br0 of the SDN node, they cannot be
+   forwarded to the pod interface correctly. We need to add flows for each pod
+   to change the dst MAC to the pod interface MAC.
```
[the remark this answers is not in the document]
pliurh: No, it wasn't. I had an unmerged patch for that. And yes, you're right we can use one flow to do this.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,404 @@
...
+  between the two network providers. So users need to evaluate before conducting
+  the live migration, and have to migrate the configuration manually after the
+  migration is complete.
+- The following features are supported by both OpenShift SDN and OVN Kubernetes.
+  And we have support automated converting for the configuration. However, users
+  shall not expect these feature can function as normal during the live
+  migration phase.
+  - Multicast
+  - Egress IP
+  - Egress Firewall
```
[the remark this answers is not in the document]
Makdaam: I'm also wondering if "stop working" means fail open or fail closed. It should be explicit in documentation for our customers to make an informed decision on when and how to proceed.

enhancements/network/sdn-live-migration.md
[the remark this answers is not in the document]
Makdaam: I think making it explicit that it's an OVN IC zone, and not for example an AWS availability zone would be helpful.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,404 @@
...
+
+### The Traffic Path
+
+#### Packets going from OpenShift SDN to OVN Kubernetes
+
+On the SDN side, it doesn't need to know if the peer node is a SDN node or a OVN
+node. We reuse the existing VXLAN tunnel rules one the SDN side.
+- Egress NetworkPolicy rules and service proxying happen as normal
+- When the packet reaches table 90, it will hit a "send via vxlan" rule that was
+  generated based on a HostSubnet object.
```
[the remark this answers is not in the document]
pliurh: HostSubnet CRs will be removed when the migration is done. So in rollback or the reverse migration(It isn't our target) scenario, we will use the bootstrap script in ovnkube-node container to create the HostSubnet CRs. I'll add the detail in the enhancement.

pliurh:
@danwinship @Makdaam I updated the doc according to your comments. PTAL.

enhancements/network/sdn-live-migration.md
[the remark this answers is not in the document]
danwinship: (I meant specifically "what does this imply" in terms of like "Does this mean X% of users won't be able to use the migration?" "Does this mean early adopters won't be able to use the migration but late-adopters will?" etc.)

enhancements/network/sdn-live-migration.md
danwinship: ok but you haven't mentioned any MachineConfig existing or being created yet, or what it contains, etc

enhancements/network/sdn-live-migration.md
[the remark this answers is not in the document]
danwinship: yeah, removing it sounds right then

enhancements/network/sdn-live-migration.md
pliurh: I added more details.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,434 @@
...
+There're two related but distinct MTU values to consider, the hardware MTU and
+cluster MTU. The cluster MTU is the MTU value of pod interfaces. It is always
+less than your hardware MTU to account for the cluster network overlay overhead.
+The overhead is is 100 bytes for OVN Kubernetes and 50 bytes for SDN. For
+example, the hardware MTU is 1500, then the auto-probed cluster MTU would be
+1400 for OVN Kubernetes and 1450 OpenShift SDN. During the migration, this MTU
+mismatch would break some cross CNI traffic.
+
+To avoid this issue, before starting the migration, users need to decrease the
+cluster MTU by 50 bytes following this instruction
```
Makdaam, line 193: Will the migration tool provide a check for the users that confirms the network MTU has enough headspace vs. the machine MTU before proceeding?

pliurh: Yes, we can add this check to CNO. But if the check result shows that an MTU migration is required, the user shall do it before conducting the migration.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,428 @@
...
+      "live": true
+    }
+  } 
+}
+```
+
+On removing of the `spec.migration` field, CNO will start the migration cleanup
+and redeploy the target CNI in 'normal' mode.
+
+```
```
trozet, line 351: this causes the entire rest of the doc to be in a code block

trozet (commented):
There are bunch of spelling and format issues with this enhancement, but as far as the general gist of how the migration would work it seems sound.

/approve

enhancements/network/sdn-live-migration.md
pliurh: Fixed.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,425 @@
...
+   nodes to the local zone and all OVN nodes as OVN-IC remote zones. On a remote
+   OVN node, OVN Kubernetes will remove a hybrid overlay node and add an OVN-IC
+   remote zone.
+5. MCO will uncordon the node. Pods will be recreated on the node using
+   ovn-kubernetes as the default CNI plugin.
+
+The above process will be repeated for each node until all the nodes have been
+applied to the new MachineConfig and converted to OVN Kubernetes.
+
+#### Migration Cleanup
```
jcaamano, line 314: Is this a manual step where the user will patch the net work config the remove the migration setting or is this step automatically triggered by CNO?

jewzaam: Could a section be added that captures all the manual steps in one place?  I'd like to be sure we understand this when looking at the implications to managed openshift.  Thanks!

pliurh: I added the Operation Procedure chapter. PTAL.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,475 @@
...
+
+   ```bash
+   $ oc get mcp
+   ```
+
+5. To finalize the live migration, cleanup original CNI and redeploy the target
+   CNI in 'normal' mode, enter the following command
+
+   ```bash
+   $ oc patch Network.operator.openshift.io cluster --type='merge' --patch '{"spec":{"migration": null, "defaultNetwork":{ "openshiftSDNConfig": null}}'
```
jcaamano, line 369: @pliurh This made me wonder if CNO will translate from  defaultNetwork.openshiftSDNConfig and set to defaultNetwork.OVNKubernetesConfig?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,475 @@
...
+```
+
+CNO will:
+- delete the openshift-sdn DaemonSets and the related resources
+  (CustomResources, ConfigMaps, etc.).
+- redeploy ovn-kubernetes in "normal" mode (no migration mode script, hybrid
+  overlay disabled, etc.).
+- remove the migration-related labels and node annotations from the nodes.
+
+#### Operation Procedure
```
jewzaam, line 329: General feedback on the manual touchpoints, we need to permit this to be done for any managed openshift (OSD/ROSA/ARO) whenever customer is ready.  We won't have SRE manually executing these steps.  It needs some controller to manage.  Is there a reason why this isn't handled as part of the migration automatically?  We need a way to express intent to start migration and an indicator when it's done.  In the middle if there are problems sufficient signal needs to be present so alerts can surface to SRE (or ops teams for self managed).  Each alert needs associated with clear steps on how to triage and fix known issues.  Of course, not everything will be known, we'll discover new things.  But some starting point will be needed else we go in blind.

pliurh: These steps can be automated and integrated into OCM. But that is out of the scope of this enhancement. As @Makdaam and I agreed, the SD team shall do the necessary UI and other changes in OCM. The cluster network operator and machine config operator will report errors in the corresponding ClusterOperator CRs. The detailed troubleshooting steps will be in the OCP document when the code is ready. The enhancement only contains the design of the feature.

enhancements/network/sdn-live-migration.md
pliurh: No. Since I don't see there are any common settings that need to be translated, except MTU. In the offline migration, users are allowed to configure `defaultNetwork.OVNKubernetesConfig` before the migration. But if unset, CNO will just set it with default values. We will follow the same behavior in the live migration.

Regarding the MTU, CNO will check if it's possible to use the same cluster MTU as OSDN in OVN-K before migration. If not CNO will block the migration.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,476 @@
...
+the cluster MTU. The cluster MTU is the MTU value for pod interfaces. It is
+always less than your hardware MTU to account for the cluster network overlay
+overhead. The overhead is 100 bytes for OVN Kubernetes and 50 bytes for SDN. For
+example, if the hardware MTU is 1500, then the auto-probed cluster MTU would be
+1400 for OVN Kubernetes and 1450 for OpenShift SDN. During the migration, this
+MTU mismatch would break some cross-CNI traffic.
+
+To avoid this issue, before starting the migration, users need to decrease the
+cluster MTU by 50 bytes following this instruction:
+[changing-cluster-network-mtu](https://docs.openshift.com/container-platform/4.13/networking/changing-cluster-network-mtu.html).
```
cblecker, line 193: This is tricky, because this isn't a fully automated process either.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,476 @@
...
+   - If `br-ex` does exist, it means the node has been updated by MCO. So the
+     `ovnkube-node` can be started.
+
+     The script will also add the node annotation `k8s.ovn.org/node-subnets`
+     according to the HostSubnet CR of this node.
+
+#### Migration
+
+When the condition 'Progressing' of the network Cluster Operator is switched
+from `True` to `False`, the admin can kick off the migration from CNO by:
```
cblecker, line 281: Similar to the MTU procedure, is there a reason a human needs to do a multi-step process here? Is there a reason we can't declaratively say "move to OVN" and all the steps are done by operators?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,476 @@
...
+   remote zone.
+5. MCO will uncordon the node. Pods will be recreated on the node using
+   ovn-kubernetes as the default CNI plugin.
+
+The above process will be repeated for each node until all the nodes have been
+applied to the new MachineConfig and converted to OVN Kubernetes.
+
+#### Migration Cleanup
+
+Once migration is complete, the admin can trigger the migration cleanup by:
```
cblecker, line 316: Same here.. we're again stating that a human has to do this, rather than an operator just cleaning up when it's done

enhancements/network/sdn-live-migration.md
pliurh: A fully automated process is not the target of this enhancement. As you can see in the procedure, both CNO and MCO are involved. So the automation shall use the APIs provided by CNO and MCO to do the work at a higher level.
For SD clusters, @Makdaam plans to implement the migration procedure in https://github.com/openshift/osdctl

enhancements/network/sdn-live-migration.md
pliurh: The answer is similar to the above.

enhancements/network/sdn-live-migration.md
jewzaam: osdctl is an SRE tool, it isn't used in automation afaik but could be wrong.  It's not the right place to think about automatino for the SD use case.  we cannot have SRE running migrations for anything, it needs to be a controller somewhere, preferably in-cluster.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,476 @@
...
+- The cluster network CIDR will remain unchanged, as will the node host subnet
+  of each node.
+- NetworkPolicy will work correctly throughout the migration.
+
+### Limitations
+
+- The following features, which are only supported by OpenShift SDN but not by
+  OVN Kubernetes, will stop working when the migration is started.
+  - Multitenant Isolation
+- The Egress Router feature is supported by both OpenShift SDN and OVN
```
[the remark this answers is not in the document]
pliurh: We support egress IP and egress FW metrics in OpenShift SDN. But not egress Router and Multicast.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,443 @@
...
+   [Limitations](#limitations) section) enabled in the cluster. CNO will
+   set an the `LiveMigrationProgressing` condition to `False` with warning message
+   and won't start the live migration. User shall disable these unsupported
+   feature before starting the live migration.
+
+2. CNO will check the current hardware MTU and the cluster MTU to see if there
+   is enough free space for OVN-K overlay overhead. If not, sets the
+   `.status.migration.mtu` of network.config CR with MTU of OVN-k, e.g.:
+   - `.status.migration.mtu.network.from` 1450 `to` 1400
+   - `.status.migration.mtu.machine.from` 1500 `to` 1500
```
cblecker, line 249: Is this dynamic based on what the current settings are? Will it be restored afterwards to the previous setting? While most clusters within a cloud will have similar settings, the use of certain features (e.g. AWS Local Zones) necessitates a lower network overlay MTU (ref https://docs.openshift.com/container-platform/4.13/installing/installing_aws/installing-aws-localzone.html#edge-machine-pools-aws-local-zones_installing-aws-localzone)

enhancements/network/sdn-live-migration.md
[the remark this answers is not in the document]
pliurh: I am not sure if there metrics/telemetry available. @martinkennelly may know better.

Those CRs/CRDs will be removed after the migration. During the migration, it will be there but have no function. For egressIP, egressFW, and multicast. we will convert the CRs to new ones that are used by the target CNI.

enhancements/network/sdn-live-migration.md
pliurh: Yes. The new cluster MTU will be calculated based on the value of the machine MTU and the current cluster MTU.

enhancements/network/sdn-live-migration.md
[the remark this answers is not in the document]
martinkennelly: That metric works for OVN-K and SDN.

enhancements/network/sdn-live-migration.md
pliurh: For AWS localzone instance, the underlay/machine MTU is 1300. It will not be changed by the migration. The migration only changes the overlay/cluster MTU which is the MTU used by the pod interfaces. So in your case, the calculated cluster MTU is 1250 for O-SDN, and 1200 for OVN-K. So after the migration. The cluster will be 1200.

cblecker: Okay, I understand now a bit more clearly although not all of your statement about local zones is correct. The machines themselves have an MTU of 9001 like other AWS nodes, but you do need to lower the overlay network MTU because the inter-zone networking in AWS for local zones to non-local zones is 1300, minus any overhead.

So for an SDN cluster with local zone nodes, I would expect the cluster/overlay MTU to be 1250, and for OVN it would be 1200 like you indicate. And for a default installed AWS cluster with no local zones, the overlay network MTU would be 8951 for SDN, and 8001 for OVN.

Thanks for clarifying!

enhancements/network/sdn-live-migration.md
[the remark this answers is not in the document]
cblecker: Thank you, these are helpful. I can see us setting up some monitoring/alerting around these metrics (we have the ability to push custom telemetry to/from the managed fleet) in order to determine usage of these features and warn users ahead of the migration.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,442 @@
...
+     `k8s.ovn.org/hybrid-overlay-node-subnet`, according to the HostSubnet CR of
+      this node.
+
+   - If `br-ex` does exist, it means the node has been updated by MCO. So the
+     `ovnkube-node` can be started.
+
+     The script will also add the node annotation `k8s.ovn.org/node-subnets`
+     according to the HostSubnet CR of this node.
+     
+   CNO wait until OVN-K is deployed in the cluster Set the
```
jcaamano, line 291: nit: `... will wait ... and set ...`

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,442 @@
...
+  the VNID set. There is already code to generate the rules that way, though,
+  because egress NetworkPolicy already works that way.
+
+### Workflow Description
+
+1. The admin kicks off the migration process by updating the custom resource
+   `network.config`:
+   
+   ```bash
+   $ oc patch Network.config.openshift.io cluster --type='merge' --patch '{"metadata":{"annotations":{"network.openshift.io/live-migration":""}},"spec":{"networkType":"OVNKubernetes"}}'
```
jcaamano, line 235: Should the annotation read as `network-type-live-migration` and statuses read as `NetworkTypeLiveMigration...` so it is clear it is about network type migration? Perhaps they could be condensed and remove the `live` part of it if there is no confusion possible with the offline migration since that one does not have similar annotations or statuses.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,442 @@
...
+7. CNO watches the MCPs, wait until the new MCs is applied to all the nodes. Set
+   the `LiveMigrationTargetMachineConfigApplied` condition to `TRUE` in the
+   status of the network.config CR.
+
+8. CNO removes the `spec.migration` field of network.operator CR. It will
+   trigger the CNO to:
+
+   - delete the openshift-sdn DaemonSets and the related resources
+     (CustomResources, ConfigMaps, etc.).
+   - redeploy ovn-kubernetes in "normal" mode (no migration mode script, hybrid
```
jcaamano, line 323: is this when host subnet allocation will be re-enabled? Maybe worth mentioning.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,442 @@
...
+     overlay disabled, etc.).
+   - remove the migration-related labels and node annotations from the nodes.
```
jcaamano, line 325: I guess you mean hybrid overlay annotations?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,442 @@
...
+SDN live migration.
+
+The following conditions will be added to the `.status` of the
+`Network.config.openshift.io` CR:
+
+- LiveMigrationProgressing
+- LiveMigrationRoutableMTUApplied
+- LiveMigrationTargetCNIReady
+- LiveMigrationTargetMachineConfigApplied
+- LiveMigrationOriginalCNIPurged
```
jcaamano, line 348: I am unsure about the need and the meaning of some these status entries. Are they merely informative, or are they needed for something else, like enabling specific troubleshooting procedures that will be documented? Or do they follow a design pattern?

I am concerned that the value of these statuses can be confusing if there several migration procedures are attempted. Should they be reset to false when a new migration or rollback is started?

I will leave it up to @cblecker and @jewzaam as I am sure they are more knowledgeable than me on how statuses can be leveraged.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,442 @@
...
+the change to take effect. If it exists, CNO takes it as the trigger of starting
+SDN live migration.
+
+The following conditions will be added to the `.status` of the
+`Network.config.openshift.io` CR:
+
+- LiveMigrationProgressing
+- LiveMigrationRoutableMTUApplied
+- LiveMigrationTargetCNIReady
+- LiveMigrationTargetMachineConfigApplied
```
jcaamano, line 347: `LiveMigrationTargetMachineConfigApplied` reads to me like an implementation detail. Could we come up with a different name? Maybe `LiveMigrationTargetCNIInUse`

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,442 @@
...
+`spec.networkType` field. If the annotation absents (default), CNO will block
+the change to take effect. If it exists, CNO takes it as the trigger of starting
+SDN live migration.
+
+The following conditions will be added to the `.status` of the
+`Network.config.openshift.io` CR:
+
+- LiveMigrationProgressing
+- LiveMigrationRoutableMTUApplied
+- LiveMigrationTargetCNIReady
```
jcaamano, line 346: `LiveMigrationTargetCNIReady` might be ambiguous. The target CNI is redeployed several times after the step where this is set to true and each of those deployments could fail on its own while this would still read as true.

Maybe `LiveMigrationTargetCNIConfigured` or `LiveMigrationTargetCNIPrepared` or LiveMigrationTargetCNIAvailable`?

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,442 @@
...
+For `Network.config.openshift.io` CRD, we will use the CR annotation
+`network.openshift.io/live-migration:` to gate the modification to
+`spec.networkType` field. If the annotation absents (default), CNO will block
+the change to take effect. If it exists, CNO takes it as the trigger of starting
+SDN live migration.
+
+The following conditions will be added to the `.status` of the
+`Network.config.openshift.io` CR:
+
+- LiveMigrationProgressing
```
jcaamano, line 344: Is it ok that `LiveMigrationProgressing` is never set to false if it is stuck or failed for whatever reason? Should this read differently, like `LiveMigrationInProgress`.

enhancements/network/sdn-live-migration.md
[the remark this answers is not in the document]
martinkennelly: I am creating a doc to give to you folks that will have this: https://issues.redhat.com/browse/SDN-4178
Id appreciate any feedback. Please watch this story.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,442 @@
...
+   [Limitations](#limitations) section) enabled in the cluster. CNO will
+   set an the `LiveMigrationProgressing` condition to `False` with warning message
+   and won't start the live migration. User shall disable these unsupported
+   feature before starting the live migration.
+
+2. CNO will check the current hardware MTU and the cluster MTU to see if there
+   is enough free space for OVN-K overlay overhead. If not, sets the
+   `.status.migration.mtu` of network.config CR with MTU of OVN-k, e.g.:
+   - `.status.migration.mtu.network.from` 1450 `to` 1400
+   - `.status.migration.mtu.machine.from` 1500 `to` 1500
```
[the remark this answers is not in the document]
pliurh: This a mistake. CNO shall set the `spec.migration` of the `network.operator` CR.

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,442 @@
...
+   
+   This will trigger MCO to apply a new machine config to each MCP and reboots
+   all the nodes.
+
+3. CNO watches the MCP, wait until the MCs with routable MTU is applied to all
+   the nodes. CNO then sets the `LiveMigrationRoutableMTUApplied` condition to
```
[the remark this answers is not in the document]
pliurh: In this design, we try to make the `network.config` API be declarative. The `network.operator` API still works in the old way. So it would need extra effort to make the MTU migration be declarative and automated.

enhancements/network/sdn-live-migration.md
pliurh: These conditions will be used by CNO to determine the current state of live migration. These conditions will be set to ` FALSE` when a live migration is triggered and reset to `UNKNOW` when a live migration is completed. Also, they will be used for troubleshooting.

enhancements/network/sdn-live-migration.md
pliurh: ok, let use `LiveMigrationTargetCNIInUse`

enhancements/network/sdn-live-migration.md
pliurh: Yes, if `LiveMigrationProgressing` got stuck in `FALSE`. It means the is an error in the live migration. It's an indicator to the user that he needs to fix the issue first.

enhancements/network/sdn-live-migration.md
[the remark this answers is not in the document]
jcaamano: I was not suggesting to cover a declarative MTU migration in this spec, but rather think ahead of time whether some statuses should be shared in case that ever happens in the future.

In any case though, `RoutableMTU` looks as an implantation detail as well. Could it just read `LiveMigrationMTUApplied` or `LiveMigrationMTUReady` or something like that?

enhancements/network/sdn-live-migration.md
jcaamano: Not sure if this is is a yes or a no. I noticed no changes here.

enhancements/network/sdn-live-migration.md
jcaamano: Perhaps worth mentioning somewhere as well. I read that the only reason for this to be set to false is if non-migratable features are in use.

Also, if there is no mention of any specific implementation needed to support the remediation of such cases, I guess that means that the procedure is supposed to automatically resume progressing once the user has fixed the issue or that otherwise it needs to be handled on a case by case basis through support? I think it would be useful to have a mention about this as well.

enhancements/network/sdn-live-migration.md
jcaamano: Isn't all this worth mentioning here or is it a common design pattern with status conditions? You mentioned to me there might be documentation around this. Could you refer me to it? Thanks!

pliurh: Yes, I can add this here. Here's the document about conditions https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

enhancements/network/sdn-live-migration.md
pliurh: I added more details. PTAL.

cblecker (approved):
Approved on behalf of SD Architecture as of 318dfb59bc1267562baa1db804900cbcf34e5ff6

enhancements/network/sdn-live-migration.md
```diff
@@ -0,0 +1,460 @@
...
+
+There are two related but distinct MTU values to consider: the hardware MTU and
+the cluster MTU. The cluster MTU is the MTU value for pod interfaces. It is
+always less than your hardware MTU to account for the cluster network overlay
+overhead. The overhead is 100 bytes for OVN Kubernetes and 50 bytes for SDN. For
+example, if the hardware MTU is 1500, then the auto-probed cluster MTU would be
+1400 for OVN Kubernetes and 1450 for OpenShift SDN. During the migration, this
+MTU mismatch would break some cross-CNI traffic.
+
+To avoid this issue, before starting the migration, users need to decrease the
```
kyrtapz, line 195: Is MTU migration automated or not?
This sentence suggests that it is not but there is also this one:
```
CNO will check the current hardware MTU and the cluster MTU to see if there is enough free space for OVN-K overlay overhead
```

pliurh: MTU migration will be part of the live migration, so it will be automated. Let me make the sentences more clear.

jcaamano: Just a clarification: it is only automated in the scope of this procedure because it doesn't contemplate the need to update the physical network MTU. This probably worth adding as well.

PR #1071, opened 2022-03-23
baremetal: Add bootstrap-external-ip

honza:


enhancements/baremetal/bootstrap-external-ip.md
```diff
@@ -0,0 +1,130 @@
...
+
+- As a user, I want to be able to deploy an IPI bare metal cluster in an
+  environment where no DHCP cluster is running
+- As a user, I want to specify a static IP address for the bootstrap VM
+
+### API Extensions
+
+We will add a new field to the `install-config.yaml` file called
+`bootstrapExternalIP`.  This is similar to the existing
+`bootstrapProvisioningIP` field.
```
sadasu, line 66: Do we need to add additional validations for this new field? If so, what would they be?

enhancements/baremetal/bootstrap-external-ip.md
```diff
@@ -0,0 +1,130 @@
+---
+title: bootstrap-external-ip
+authors:
+  - "@honza"
+reviewers:
+  - "@hardys"
+  - "@dtantsur"
+  - "@dhellmann"
+  - TBD
```
zaneb, line 9: Should probably add @patrickdillon since this touches the install-config

enhancements/baremetal/bootstrap-external-ip.md
```diff
@@ -0,0 +1,130 @@
...
+- Improve the level of customization of bootstrap VM networking
+- Improve user-friendliness when specifying a static IP for the bootstrap VM
+
+### Non-Goals
+
+- Provide means for configuring every aspect of bootstrap networking
+
+## Proposal
+
+We will add a new field, `bootstrapExternalIP`, to allow for further
```
zaneb, line 52: Maybe put 'static' in the name?

enhancements/baremetal/bootstrap-external-ip.md
zhouhao3: I think just verifying it's in the correct IP format should be fine.

zhouhao3: I think just verifying it's in the correct IP format should be fine.

zhouhao3 (commented):
Seems good to me. 
Is there a track issue for this proposal? I can help with some tasks for the implementation of this proposal if needed.

enhancements/baremetal/bootstrap-external-ip.md
```diff
@@ -0,0 +1,126 @@
...
+We will add two new fields to the `install-config.yaml` file called
+`bootstrapExternalStaticIP`, and `bootstrapExternalGateway`.  This is similar to
+the existing `bootstrapProvisioningIP` field.
+
+### Risks and Mitigations
+
+## Design Details
+
+### Test Plan
+
```
hardys, line 75: We should follow up and add CI coverage for static IP deployments, I know @derekhiggins was [working on that](https://github.com/openshift-metal3/dev-scripts/pull/1342) but I'm not sure of the latest status

enhancements/baremetal/bootstrap-external-ip.md
```diff
@@ -0,0 +1,126 @@
...
+- More testing (upgrade, downgrade, scale)
+- Sufficient time for feedback
+- Available by default
+- Backhaul SLI telemetry
+- Document SLOs for the component
+- Conduct load testing
+- User facing documentation created in [openshift-docs](https://github.com/openshift/openshift-docs/)
+
+**For non-optional features moving to GA, the graduation criteria must include
+end to end tests.**
```
hardys, line 100: Most of this section does not seem relevant, the new installer API will go straight to GA and automates and already-possible deployment flow (but we should add e2e coverage as mentioned above)

zaneb:
I agree with Steve's suggestion of stating which section of the install-config this will go in. Otherwise LGTM.

enhancements/baremetal/bootstrap-external-ip.md
hardys: We should validate [similar to other fields](https://github.com/openshift/installer/blob/master/pkg/types/baremetal/validation/platform.go#L410..L414) e.g

* IP syntax is valid
* IP is in the MachineNetwork
* IP doesn't conflict with other IPs (e.g API/Ingress VIP IPs)

hardys:
This version lgtm other than the markdownlint failure.

We should follow up with the test-plan details though when we make progress on getting an e2e job in place that exercises the static-ip scenario (if you have to update this again perhaps mention there is work in-progress to add such coverage, and that this will get tested via that new job when it's done)

enhancements/baremetal/bootstrap-external-ip.md
sadasu: Thanks! I was wondering if there is some correlation with the `networkConfig` field. Since nothing is mentioned in the doc about it, I am assuming that they can occur independently. We could configure `bootstrapExternalStaticIP`, and
`bootstrapExternalGateway` irrespective of whether the `networkConfig` field is used to configure StaticIPs on hosts within the cluster.

enhancements/baremetal/bootstrap-external-ip.md
```diff
@@ -0,0 +1,135 @@
...
+### Workflow Description
+
+The **cluster creator** will modify the `install-config.yaml` to configure
+static networking for the bootstrap node.
+
+### API Extensions
+
+We will add two new fields to the baremetal platform section of the
+`install-config.yaml` file called `bootstrapExternalStaticIP`, and
+`bootstrapExternalGateway`.  This is similar to the existing
```
zaneb, line 75: bootstrapExternalStaticGateway

zaneb (commented):
/lgtm

PR #1073, opened 2022-03-24
Fix links to make relative to page

johnbwork:
Current links to pull requests, not to actual .md files.

### DIFF
--- a/enhancements/network/dpu/smart-nic-ovn-offload.md
+++ b/enhancements/network/dpu/smart-nic-ovn-offload.md
@@ -31,7 +31,7 @@ status: implementable
 
 ## Summary
 
-See [DPU Support in OCP (Overview)](https://github.com/openshift/enhancements/pull/738). This document describes the
+See [DPU Support in OCP (Overview)](.enhancements/network/dpu/ocp-on-dpu.md). This document describes the
 OCP modification in Tenant cluster in details.
 
 ## Terminology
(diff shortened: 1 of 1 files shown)


enhancements/network/dpu/smart-nic-ovn-offload.md
```diff
@@ -31,7 +31,7 @@ status: implementable
 
 ## Summary
 
-See [DPU Support in OCP (Overview)](https://github.com/openshift/enhancements/pull/738). This document describes the
+See [DPU Support in OCP (Overview)](.enhancements/network/dpu/ocp-on-dpu.md). This document describes the
```
dhellmann, line 34: Do those links work with the `.` at the front? Should that be `./` or even `/`?

dhellmann: In [at least one place](https://github.com/openshift/enhancements/blob/master/enhancements/baremetal/an-slo-for-baremetal.md?plain=1#L282) we use a full URL to link to another doc but that feels excessive. In [this other case](https://github.com/openshift/enhancements/blob/master/enhancements/builds/disable-jenkins-pipeline.md?plain=1#L211), a relative path is used. And [over here](https://github.com/openshift/enhancements/blob/master/enhancements/builds/volume-mounted-resources.md?plain=1#L68), `/enhancements/...` is used.

danwinship: That stackoverflow answer is only talking about exactly `"."` vs `"./"`. You always need the slash if you are forming a full path.

> Check what the link looks like on my forked file here: https://github.com/johnbwork/enhancements/blob/patch-1/enhancements/network/dpu/smart-nic-ovn-offload.md

right, it doesn't work: you get a link to `https://github.com/johnbwork/enhancements/blob/patch-1/enhancements/network/dpu/.enhancements/network/dpu/ocp-on-dpu.md` with a `.` in there that makes it not work.

`./ocp-on-dpu.md` should work

PR #1076, opened 2022-03-29
Add on-prem internal loadbalancer ports to host port registry

cybertron:
The internal loadbalancer has been using a few ports since it was
first implemented. These need to be reflected in the registry.

### DIFF
--- a/dev-guide/host-port-registry.md
+++ b/dev-guide/host-port-registry.md
@@ -9,7 +9,7 @@ reviewers:
 approvers:
   - "@russelb"
 creation-date: 2020-08-26
-last-updated: 2022-01-06
+last-updated: 2022-05-09
 status: informational
 ---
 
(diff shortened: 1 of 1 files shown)


cybertron:
/hold

When I went to add our ports I discovered we were conflicting with the kubernetes node port range. I have patches up to fix that, but until they go in these ports are not accurate.

https://github.com/openshift/baremetal-runtimecfg/pull/175
https://github.com/openshift/machine-config-operator/pull/3044

cybertron:
/hold cancel

I'm not sure it makes that much difference what order these patches merge in, just as long as they all eventually make it in.

cgwalters:
> I'm not sure it makes that much difference what order these patches merge in, just as long as they all eventually make it in.

Definitely.  We're Kubernetes!  Eventual consistency!

cybertron:
/remove-lifecycle stale

We still need this.

dhellmann:
/approve

@cybertron maybe you can get someone else on the network team to lgtm?

dev-guide/host-port-registry.md
```diff
@@ -142,6 +144,7 @@ Ports are assumed to be used on all nodes in all clusters unless otherwise speci
 | 10302 | various CSI drivers | storage | 4.7 | healthz |
 | 10303 | various CSI drivers | storage | 4.9 | healthz |
 | 11443 | kube-scheduler | workloads || recovery-controller |
+| 20000  | haproxy | sdn | 4.7 | on-prem internal loadbalancer, stats port |
```
danwinship, line 147: while I guess you _can_ use that port, for the others we tried to keep the convention that the localhost-only port was 20000 more than the externally-accessible port...

cybertron: That works for me, I wasn't aware of the convention before now. I've updated this and the runtimecfg patch accordingly.

PR #1077, opened 2022-03-31
Add a priority label to alerts

sradco:
Added a new optional label called `priority`.
The 'priority` is different than the severity.
Severity indicates the impact the alert has on the component or system.
Priority indicates the importance of the alert and the order in which an alert should be fixed.

Signed-off-by: Shirly Radco <sradco@redhat.com>

### DIFF
--- a/enhancements/monitoring/alerting-consistency.md
+++ b/enhancements/monitoring/alerting-consistency.md
@@ -108,7 +108,8 @@ these alerts if they are not actionable.
 * Alert names MUST be CamelCase, e.g.: `PrometheusRuleFailures`
 * Alert names SHOULD be prefixed with a component, e.g.: `AlertmanagerFailedReload`
   * There may be exceptions for some broadly scoped alerts, e.g.: `TargetDown`
-* Alerts MUST include a `severity` label indicating the alert's urgency.
+* Alerts MUST include a `severity` label indicating the impact the alert has
+    on the operation of a component or the system.
   * Valid severities are: `critical`, `warning`, or `info` — see below for
     guidelines on writing alerts of each severity.
 * Alerts MUST include `summary` and `description` annotations.
@@ -123,6 +124,10 @@ these alerts if they are not actionable.
   * Runbook style documentation for resolving critical alerts is required.
     These runbooks are reviewed by OpenShift SREs and currently live in the
     [openshift/runbooks][2] repository.
+* Alerts CAN include a `priority` label indicationg the alert's level of importance and the order in which it should be fixed.
+  * Valid priorities are: `high`, `medium`, or `low`.
+    Higher the priority the sooner the alert should be resolved.
+  * If the alert doesn't include a `priority` label, we can assume it is a `medium` priority alert.
 
 ### Critical Alerts
 
@@ -228,7 +233,6 @@ Example info alert: [MultipleContainersOOMKilled][5]
     namespace: kube-system
     severity: info
 ```
-
 This alert fires if multiple containers have been terminated due to out of
 memory conditions in the last 15 minutes.  This is something the administrator
 should be aware of, but may not require immediate action.


sradco:
@bison @michaelgugino Please review my PR and let me know what you think.|
In CNV we plan to use the `priority` label in order to identify warning alerts that are important and display them to the users.
Its for cases where the alert is not critical and doesn't require immediate attention, but is still very important and should be addressed with a high importance.

simonpasquier (commented):
I guess that KubeVirt already uses this label but could you link to concrete examples?

The severity label already carries a notion of urgency (critical: someone needs to act now, warning: someone needs to be looked at it but it doesn't require an interruption) and I'm not sure that another field is necessary. For example, what would be the case for a warning/high alert vs. warning/low? Could critical alerts have priority other than high?

cc @openshift/sre-alert-sme for their inputs.

rhrazdil:
@sradco has pointed to a document somewhere (unfortunately not in this PR, do you still have the link Shirly?) where the distinction between priority and urgency is explained.

The example IIRC was, if you have a component, that's completely broken due to a bug, you'd say that the bug has super high severity. But, consider that the component is not used by any of your customers. That probably means the nobody cares and thus priority of fixing it is low.

This make a world of sense to me, however, since the `priority` field is completely general, what if other alerts start to use it.
Wouldn't it mean that other alerts also get displayed in the UI?
@yaacov @sradco Have you considered that we may not want to display everything that has a priority label?

simonpasquier:
> The example IIRC was, if you have a component, that's completely broken due to a bug, you'd say that the bug has super high severity. But, consider that the component is not used by any of your customers. That probably means the nobody cares and thus priority of fixing it is low.

In this case, the priority would be dependent on the customer environment and not something that the component can determine by itself.

yaacov:
> In this case, the priority would be dependent on the customer environment and not something that the component can determine by itself.

@simonpasquier 
Yes, usually it is, that is why we propose it to be an optional label. usually the entity setting the alert rule will not know the priority and only a cluster admin can know it.

In our case we have an exception, the alert, `CnaoNmstateMigration` - "Nmstate will be removed from CNAO"
is `low` or `warning` severity,  because the cluster works find without any need to take immediate action or functionality loss,
but we want to raise the awareness of users to this alert, if they will not act on it soon, next update will break there cluster.

In this case we want to have a "warning" alert, but we want the cluster admin to take immediate action although there is currently no functionality loss.

simonpasquier:
> In our case we have an exception, the alert, CnaoNmstateMigration - "Nmstate will be removed from CNAO"
is low or warning severity, because the cluster works find without any need to take immediate action or functionality loss,
but we want to raise the awareness of users to this alert, if they will not act on it soon, next update will break there cluster.

Hmm I'm not familiar how the KubeVirt operator works and what is available from OLM but I feel that the operator should prevent the upgrade in the first place rather than expecting that an alert would stop the cluster admin from running the upgrade?

rhrazdil:
> In our case we have an exception, the alert, CnaoNmstateMigration - "Nmstate will be removed from CNAO"
is low or warning severity, because the cluster works find without any need to take immediate action or functionality loss,
but we want to raise the awareness of users to this alert, if they will not act on it soon, next update will break there cluster.

Let me correct that, there is no functionality loss. The upgrade would be denied by the HCO operator.
The aim of the alert is to make most users to install the operator before they get to upgrade to 4.11, to ensure smooth upgrade.

dofinn:
I am against adding this field from an SRE perspective. Alerting is complicated enough already without adding an extra field to create more granular response process. 

> In this case we want to have a "warning" alert, but we want the cluster admin to take immediate action although there is currently no functionality loss.

In this defintion, the proposed alert should be marked as severity of Critical based on the description of the desired response to this condition. 

If you want to display any alert to a user, why not just use a label for that?

```
display: true
```

That way, this abstraction avoids muddying severity response standards.

sradco:
@dofinn Hi,

Im actually against adding something like 
```display: true```

Since alerts are not used only in the OCP UI, they should remain clean from specific environment details. The user can set the alertmanager to get the alerts and they will get all the related labels with it and it will have no value for them.

There is a clear difference between severity and priority.
It is not meant to be used with critical alerts. But for warning alerts it can create a bit more order in the clutter.

Having this alert as critical is definitely not an option, since we should not wake up people in the middle of the night to fix this.

Severity is objective, Priority is subjective and can change over a period of time depending on the change in the project situation

Please see this article that describe the differences in a very clear way:
https://www.guru99.com/defect-severity-in-software-testing.html

yaacov:
> If you want to display any alert to a user, why not just use a label for that?

We display all alerts

"critical" alerts are pushed up and displayed more prominently, and admin know they need to assign high priority to fix them,
"warning" severity tells the admins they can assign the priority according to there capacity and cluster needs,
In our case we want to hing the admin that they need to assign higher priority to a specific "warning" alert over other alerts with the same severity,

in our case "CnaoNmstateMigration" is not "critical", it does not break the cluster, and it should not trigger a "critical" alert response.
it is a "warning", but we want to hint the admin it should be handled with higher priority than other "warning" currently active. 
the "priority" label does that, it hints admins that they need assign higher priority to that alert over others without overriding the "severity" mechanisms.

sradco:
@rhrazdil Hi, regarding your comment about other alerts using this label and its affect on the UI.
For that I opened https://github.com/openshift/enhancements/pull/1078.

We already added 2 labels that indicate the operator name and component.
This is very useful for many use cases:
1. Additional information to the user what operator generated the alert
2. Being able to filter them in the UI like we want to do now.
3. Being able to filter them in Telemetry and have better granularity for examining alert and catching issues.

dofinn:
> Since alerts are not used only in the OCP UI, they should remain clean from specific environment details.
Use of Priority is a environment specific variable? 

I understand your goals and can sympathize with them, but from the perspective of SRE at scale (where we dont use the OCP UI) its hard to extrapolate this value. SRE either needs to paged for immediate attention (sev=Critical), else it goes into a queue for processing (sev=Warning).

yaacov:
> If this would be useful for other operators we can have a section like "Optional labels" / "Additional labels" (Other suggestion are welcome) that explains what this label means and how to use it.

makes sense to me too

dofinn:
> Otherwise this will result in many different labels in the wild with no standardization, which will make this a pain for users and will not provide the additional benefits as mentioned above.
I want to emphasize that currently every operator has the freedom to add what ever label that they decide on. I'm just trying to make some standardization and provide helpful information that saves others time.

If others are planning on using these mechanisms specifically to be flagged in the OCP console, then standardizing makes sense.

simonpasquier:
I'd be ok if this goes into a different section so it's clear that these labels are optional. What I want to avoid is adding more labels than what's necessary. In most cases, the existing labels should be enough to triage and route the alert efficiently.

simonpasquier:
And regarding the `priority` label, I'd need to see a concrete use case to be convinced that it's a useful addition.
If I read https://github.com/openshift/enhancements/pull/1077#issuecomment-1087522582 correctly, the operator would block the upgrade anyway.

yaacov:
> regarding the priority label, I'd need to see a concrete use case to be convinced that it's a useful addition.

In our use case "CnaoNmstateMigration", we want to raise "CnaoNmstateMigration" priority above the regular alerts in the "warnings" queue without setting the severity to "critical".

"CnaoNmstateMigration" is not critical, we do not need to address it as critical alert,
On the other hand, upgrade of the system is blocked, this may be an important functionality loss in some cases.
We want to set the severity to "warning", because we don't want to abuse the "critical" settings, and we want to hint admins that this alert should be processed with higher priority when looking at the pool of warning alerts. 

( It's an optional flag, alert maintainers will only use it when they need such granularity, for most alerts it is not needed. )

rhrazdil:
@simonpasquier @sradco  @dofinn  What else do we need to move forward?

PR #1078, opened 2022-03-31
Add operator name and component labels

sradco:
Added 2 recommended labels for operators alerts.
`kubernetes_operator_part_of` - Name of the operator the alert was fire fired from.
`kubernetes_operator_component` - Name of the component/sub operator  the alert was fire fired from.

Signed-off-by: Shirly Radco <sradco@redhat.com>

### DIFF
--- a/enhancements/monitoring/alerting-consistency.md
+++ b/enhancements/monitoring/alerting-consistency.md
@@ -123,6 +123,10 @@ these alerts if they are not actionable.
   * Runbook style documentation for resolving critical alerts is required.
     These runbooks are reviewed by OpenShift SREs and currently live in the
     [openshift/runbooks][2] repository.
+* Operator Alerts are RECOMMENDED to include `kubernetes_operator_part_of` label
+  indicating the operator name the alert is related to.
+* Operator Alerts are RECOMMENDED to include `kubernetes_operator_component` label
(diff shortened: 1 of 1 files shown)


sradco:
@bison @michaelgugino Please review my PR and let me know what you think.

enhancements/monitoring/alerting-consistency.md
```diff
@@ -123,6 +123,10 @@ these alerts if they are not actionable.
   * Runbook style documentation for resolving critical alerts is required.
     These runbooks are reviewed by OpenShift SREs and currently live in the
     [openshift/runbooks][2] repository.
+* Operator Alerts are RECOMMENDED to include `kubernetes_operator_part_of` label
+  indicating the operator name the alert is related to.
+* Operator Alerts are RECOMMENDED to include `kubernetes_operator_component` label
```
simonpasquier, line 128: I presume we're talking about the operand? If yes, how is it different from the `service` or `job` labels?

sradco: @simonpasquier I see that we don't set the these labels.  Can you please provide me with how is it being set in other operators?

Also, I didn't see documentation for these labels. I must say that from the name it is not that clear what they exactly mean and also in most cases I see they are set to the same value.

PR #1082, opened 2022-04-04
dev-guide/cluster-version-operator/dev/operators: Link capabilities enhancement

wking:
Sentence of very-high-level context, so folks have some idea of when they'd want to click through to the enhancement with all the details.  Following the pattern we pivoted to in 3d4a9f6b59 (#974).


bparees:
/approve
/lgtm

in case you want feedback from others:
/hold

LalatenduMohanty (approved):
/lgtm

PR #1087, opened 2022-04-10
tools: add metadata linting step to linter job

dhellmann:
- Add a metadata-lint command to the tools directory for checking enhancements metadata.
- Update the existing job to run the new tool against enhancements markdown files that are changed in the current pull request.
- Require an approver in the metadata.
- Require at least one reviewer in the metadata.
- Require a string that looks like a URL in the tracking-links metadata.
- Update the comments in the metadata section of the template to provide better guidance about the new linter rules.

### DIFF
--- a/hack/metadata-lint.sh
+++ b/hack/metadata-lint.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -x
+
+# Ensure the enhancement metadata includes the required information
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# We only look at the files that have changed in the current PR, to
+# avoid problems when the template is changed in a way that is
+# incompatible with existing documents.
+CHANGED_FILES=$(git log --name-only --pretty= "${PULL_BASE_SHA:-origin/master}.." -- \
+                    | (grep '^enhancements.*\.md$' || true) \
+                    | sort -u)
+
+(cd tools && go build -o ../enhancement-tool -mod=mod ./main.go)
+
+./enhancement-tool metadata-lint ${CHANGED_FILES}
--- a/tools/cmd/metadataLint.go
+++ b/tools/cmd/metadataLint.go
@@ -1,0 +1,85 @@
+/*
+Copyright © 2022 Doug Hellmann <dhellmann@redhat.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/openshift/enhancements/tools/enhancements"
+
+	"github.com/spf13/cobra"
+)
+
+// metadataLintCmd represents the metadataLint command
+var metadataLintCmd = &cobra.Command{
+	Use:       "metadata-lint",
+	Short:     "Check the metadata of the enhancement files given as arguments",
+	ValidArgs: []string{"enhancement-filename"},
+	Run: func(cmd *cobra.Command, args []string) {
+		errorCount := 0
+		for _, filename := range args {
+			reportError := func(msg string) {
+				fmt.Fprintf(os.Stderr, "%s: %s\n", filename, msg)
+				errorCount++
+			}
+			content, err := ioutil.ReadFile(filename)
+			if err != nil {
+				reportError(fmt.Sprintf("%s", err))
+				continue
+			}
+			metadata, err := enhancements.NewMetaData(content)
+			if err != nil {
+				reportError(fmt.Sprintf("%s", err))
+				continue
+			}
+
+			// Doc title needs to match filename, minus the ".md" extension
+			fileBase := filepath.Base(filename)
+			fileBase = fileBase[0 : len(fileBase)-3]
+			if fileBase != metadata.Title {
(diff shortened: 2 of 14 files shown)


tools/enhancements/metadata.go
```diff
@@ -43,3 +44,79 @@ func (s *Summarizer) GetMetaData(pr int) (*MetaData, error) {
...
+			Data: m.Approvers,
+		},
+		{
+			Name: "api-approvers",
+			Data: m.APIApprovers,
+		},
+	} {
+		valid := 0
+		for _, value := range field.Data {
+			if value == "TBD" {
```
aravindhp, line 110: Can't we check if this is a valid GitHub id?

tools/enhancements/metadata.go
```diff
@@ -43,3 +44,79 @@ func (s *Summarizer) GetMetaData(pr int) (*MetaData, error) {
...
+	results := []string{}
+
+	reportError := func(msg string) {
+		results = append(results, msg)
+	}
+
+	// Must have one valid tracking link and no TBD
+	foundLink := false
+	for _, trackingLink := range m.TrackingLinks {
+		if trackingLink == "TBD" {
```
wking, line 59: We can probably ban `TBD` in all metadata properties, right?

dhellmann: We could. I was focusing on the fields that seemed most important right now, based on issues I've seen with PRs.

hack/metadata-lint.sh
```diff
@@ -0,0 +1,20 @@
...
+set -o pipefail
+
+# We only look at the files that have changed in the current PR, to
+# avoid problems when the template is changed in a way that is
+# incompatible with existing documents.
+CHANGED_FILES=$(git log --name-only --pretty= "${PULL_BASE_SHA:-origin/master}.." -- \
+                    | (grep '^enhancements.*\.md$' || true) \
+                    | sort -u)
+
+(cd tools && go build -o ../enhancement-tool -mod=mod ./main.go)
```
aravindhp, line 18: We should ideally build the binary in the Dockerfile and just run it here.

aravindhp (commented):
Thanks for doing this, @dhellmann. Just a few non-blocking comments.

hack/metadata-lint.sh
dhellmann: Ideally, yeah. Right now that Dockerfile is used for several things, each of which is installed separately. So, I was following the existing pattern. Maybe we can review that separately.

tools/cmd/metadataLint.go
```diff
@@ -0,0 +1,85 @@
...
+			if err != nil {
+				reportError(fmt.Sprintf("%s", err))
+				continue
+			}
+
+			// Doc title needs to match filename
+			fileBase := filepath.Base(filename)
+			fileBase = fileBase[0 : len(fileBase)-3]
+			if fileBase != metadata.Title {
+				reportError(fmt.Sprintf("title %s and filename %s do not match",
```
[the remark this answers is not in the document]
dhellmann: How's this?

```suggestion
				reportError(fmt.Sprintf("the title %s and the file base name %s must match",
```

tools/enhancements/metadata.go
```diff
@@ -43,3 +44,79 @@ func (s *Summarizer) GetMetaData(pr int) (*MetaData, error) {
...
+			reportError("'TBD' is not a valid value for tracking-link")
+			continue
+		}
+		if trackingLink == "" {
+			reportError("empty string for tracking-link")
+			continue
+		}
```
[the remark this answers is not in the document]
dhellmann: I was trying to differentiate between a missing value and an invalid value. Do you think people would understand "could not parse tracking link ''" as meaning the value was empty and it can't be?

tools/enhancements/metadata.go
dhellmann: We can, but doing that reliably would require giving the job a github token (to avoid API load limits) and specifying how to parse the name out of the comment. I didn't want to get into that level of complexity. I trust reviewers to read the values that are there, so this check is just ensuring someone provides a value.

hack/metadata-lint.sh
aravindhp: That works for me.

tools/cmd/metadataLint.go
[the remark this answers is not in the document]
aravindhp: That's much clearer.

tools/enhancements/metadata.go
```diff
@@ -43,3 +44,79 @@ func (s *Summarizer) GetMetaData(pr int) (*MetaData, error) {
...
+
+	// Must have one valid tracking link and no TBD
+	foundLink := false
+	for _, trackingLink := range m.TrackingLinks {
+		if trackingLink == "TBD" {
+			reportError("'TBD' is not a valid value for tracking-link")
+			continue
+		}
+		if trackingLink == "" {
+			reportError("empty string for tracking-link")
```
dhellmann, line 64:
```suggestion
			reportError("tracking-link must contain at least one valid URL")
```

tools/enhancements/metadata.go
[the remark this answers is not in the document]
dhellmann: Done

tools/enhancements/metadata.go
dhellmann: Done

tools/cmd/metadataLint.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+			metadata, err := enhancements.NewMetaData(content)
+			if err != nil {
+				reportError(fmt.Sprintf("%s", err))
+				continue
+			}
+
+			// Doc title needs to match filename
+			fileBase := filepath.Base(filename)
+			fileBase = fileBase[0 : len(fileBase)-3]
```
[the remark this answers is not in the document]
dhellmann: Done

dhellmann:
I believe I have made updates based on all of the comments on the earlier drafts.

aravindhp (commented):
/lgtm

PR #1092, opened 2022-04-19
Correct description of cap separation character

bparees:


wking (approved):
/lgtm

wking:
Follows up on #1042.

bparees:
this is as happy as i'm going to make the new linting rules when run against an old EP.

PR #1093, opened 2022-04-19
enhancements/update/update-blocker-lifecycle: Pivot to "update recommendations"

wking:
Users are occasionally confused by the "blocked edges" wording, which is internal shorthand.

The fact that update recommendations are tracked as a directed graph (hence "edges") isn't all that relevant for most users, so talk about "updates" instead of "edges".

While we occasionally drop update recommendations to reduce the chances that users update into serious issues, from [docs][1]:

> When an update recommendation is supported, it remains supported for the life of 4.10, even if the update recommendation is later dropped or made conditional.

Updating is still possible, we just don't think it's a good idea (unless maybe you're elbow deep in a support case, and a support tech decides to recommend an update based on context that is not available to the update service).  So replace "blocked" with "not recommended", to convey the "still possible" aspect.

Also ask for some more specifics in the impact statement template, as we gear up to support conditional updates fff9d3c349 (#821).  I haven't gone as far as "tell us what PromQL to use in our conditional recommendation", but I'm easing things in that direction.

[1]: https://docs.openshift.com/container-platform/4.10/updating/understanding-upgrade-channels-release.html#upgrade-version-paths_understanding-upgrade-channels-releases

### DIFF
--- a/enhancements/update/update-blocker-lifecycle/README.md
+++ b/enhancements/update/update-blocker-lifecycle/README.md
@@ -9,15 +9,15 @@ approvers:
 api-approvers:
   - None
 creation-date: 2020-09-11
-last-updated: 2021-06-17
+last-updated: 2022-04-19
 status: implementable
 ---
 
 # Update-blocker Lifecycle
 
 We occasionally have bugs which impact update success or the stability of the target release.
-When that happens, we protect users by [removing the impacted updates from our update recommendations][graph-data-block].
-This enhancement describes the process used to identify these bugs and determine which edges should be blocked because of them.
+When that happens, we protect users by [removing update recommendations or qualifying recommendations them with conditional risks][graph-data-block].
+This enhancement describes the process used to identify these bugs and clarify the resulting update risks.
 
 ## Release Signoff Checklist
 
@@ -29,14 +29,14 @@ This enhancement describes the process used to identify these bugs and determine
 
 ## Summary
 
-The lifecycle for update blockers looks like:
+The lifecycle for recommendation changes looks like:
 
 <div style="text-align:center">
   <img src="flow.svg" width="100%" />
 </div>
 
 Currently all tracking through the lifecycle is manual, and it is tedious for graph-admins to audit bugs with `UpgradeBlocker` to see where they are in the lifecycle and, when necessary, poke component teams about outstanding impact statement requests.
-Having an explicit, machine-readable lifecycle reduces the chances that suspected update blockers fall through the cracks by clarifying the responsible parties for moving the bug to the next stage, which supports tracking and automated reminders.
+Having an explicit, machine-readable lifecycle reduces the chances that issues fall through the cracks by clarifying the responsible parties for moving the bug to the next stage, which supports tracking and automated reminders.
 
 With the changes from this enhancement, the queues become:
 
@@ -48,7 +48,7 @@ With the changes from this enhancement, the queues become:
 
 ### Goals
 
-* Clearly define, in a machine-readable fashion, the currently responsible party for bugs in the update-blocker lifecycle.
+* Clearly define, in a machine-readable fashion, the currently responsible party for bugs in the update-recommendation lifecycle.
 
 ### Non-Goals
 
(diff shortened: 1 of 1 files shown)


enhancements/update/update-blocker-lifecycle/README.md
```diff
@@ -5,17 +5,19 @@ authors:
...
+last-updated: 2022-04-19
 status: implementable
 ---
 
 # Update-blocker Lifecycle
 
 We occasionally have bugs which impact update success or the stability of the target release.
-When that happens, we protect users by [removing the impacted updates from our update recommendations][graph-data-block].
-This enhancement describes the process used to identify these bugs and determine which edges should be blocked because of them.
+When that happens, we protect users by [removing or qualifying the impacted updates from our update recommendations][graph-data-block].
```
LalatenduMohanty, line 19: I do not think `we protect users by qualifying the impacted updates from our update recommendations` communicates the same thing you want to communicate

enhancements/update/update-blocker-lifecycle/README.md
```diff
@@ -69,35 +71,35 @@ Write tooling that automatically:
...
+We're asking the following questions to evaluate whether or not this bug warrants changing update recommendations from either the previous X.Y or X.Y.Z.
 The ultimate goal is to avoid delivering an update which introduces new risk or reduces cluster functionality in any way.
 Sample answers are provided to give more context and the ImpactStatementRequested label has been added to this bug.
 When responding, please remove ImpactStatementRequested and set the ImpactStatementProposed label.
 The expectation is that the assignee answers these questions.
 
-Who is impacted?  If we have to block upgrade edges based on this issue, which edges would need blocking?
-* example: Customers upgrading from 4.y.Z to 4.y+1.z running on GCP with thousands of namespaces, approximately 5% of the subscribed fleet
-* example: All customers upgrading from 4.y.z to 4.y+1.z fail approximately 10% of the time
+Who is impacted?  Which 4.y.z to 4.y'.z' updates?  Which types of clusters?  How does a cluster admin determine they are vulnerable?
```
LalatenduMohanty, line 80: `How does a cluster admin determine they are vulnerable?` should be in a separate line.

enhancements/update/update-blocker-lifecycle/README.md
```diff
@@ -69,35 +71,35 @@ Write tooling that automatically:
...
+* example: Customers upgrading from 4.y.Z to 4.y+1.z running on GCP with thousands of namespaces, approximately 5% of the subscribed fleet.  Check your vulnerability with `oc ...` or the following PromQL `count (...) > 0`.
+* example: All customers upgrading from 4.y.z to 4.y+1.z fail approximately 10% of the time.  Check your vulnerability with `oc adm upgrade` to show your current cluster version.
 
-What is the impact?  Is it serious enough to warrant blocking edges?
-* example: Up to 2 minute disruption in edge routing
-* example: Up to 90 seconds of API downtime
-* example: etcd loses quorum and you have to restore from backup
+What is the impact?  Is it serious enough to warrant removing update recommendations?  How does a cluster admin determine they are getting bitten?
```
LalatenduMohanty, line 84: `How does a cluster admin determine they are vulnerable?` and `How does a cluster admin determine they are getting bitten?` is almost the same thing IMO.

LalatenduMohanty (changes requested):

enhancements/update/update-blocker-lifecycle/README.md
wking: the distinction is that "vulnerable" would go into conditional update preconditions.  E.g. "if I have `$CONFIG`, and I update to `$VERSION`, I'll be at risk".  Like [high control plane memory and 4.9.<28 makes you vulnerable to etcd data inconsistency][1].  "Getting bitten" is "the bad thing is currently happening to you" and goes into the recovery KCS.  Like "[`etcdctl endpoint status -w fields --cluster` shows persistent, divergent revisions][1]".  We can protect folks from vulnerabilities without waiting for them to get bitten.  Some folks who we fail to protect from vulnerabilities will _still_ not get bitten.

[1]: https://access.redhat.com/solutions/6849521

enhancements/update/update-blocker-lifecycle/README.md
wking: Reworded to:

```markdown
When that happens, we protect users by [removing update recommendations or qualifying recommendations them with conditional risks][graph-data-block].
```

with af8037bf -> 92525969.

enhancements/update/update-blocker-lifecycle/README.md
wking: Yup.  I've pushed 92525969 -> 18de1352 making all of this more explicit.

enhancements/update/update-blocker-lifecycle/README.md
wking: Line is gone with 92525969 -> 18de1352; please open a new thread if you have line-wrapping issues with the new content :)

sdodson:
/lgtm
This amends an existing enhancement

PR #1097, opened 2022-04-25
Move drawbacks section up to a place of prominence, rework a few other section orderings

bparees:
I think this creates a better flow:

motivation - why do we want to do this
user stories - who is asking for it/what do they want to do
goals - concrete interpretation of the stories
non-goals - concrete interpretation of the stories
proposal - how we plan to do it
drawbacks - reasons not to do it (technical or business), subsection of Proposal

while bringing the "drawbacks" section up to front and center instead of buried
at the end.

I would probably argue for moving "alternatives" up from where it is now (near the bottom) also, but i'm ok letting it sit for now

### DIFF
--- a/guidelines/enhancement_template.md
+++ b/guidelines/enhancement_template.md
@@ -98,16 +98,6 @@ focus discussion and make progress. Highlight anything that is being
 deferred to a later phase of implementation that may call for its own
 enhancement.
 
-## Drawbacks
-
-The idea is to find the best form of an argument why this enhancement should
-_not_ be implemented.
-
-What are the reasons we might not want to undertake this proposal, and how do 
-we overcome them?  Does this proposal implement a behavior that's new/unique/novel?
-Is it poorly aligned with existing user expectations?  Will it be a significant
-maintenance burden?  Is it likely to be superceded by something else in the near future?
-
 ## Proposal
 
 This is where we get down to the nitty gritty of what the proposal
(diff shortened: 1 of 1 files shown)


bparees:
@dhellmann i'm willing to split this out into separate PRs if more discussion is needed, but this fixes some of the structural ordering issues that had been bothering me about the current EP section ordering

dhellmann:
I like this. I'll end up having to change my PR adding the workflow section, but I can do that.

/approve

/cc @kikisdeliveryservice

kikisdeliveryservice (commented):
+1 moving user stories up and general +1 in moving drawbacks up but I feel like it would be better moved up to right under the Proposal section? I'd understand the drawbacks better, if I read the proposal first no?

bparees:
> I feel like it would be better moved up to right under the Proposal section? I'd understand the drawbacks better, if I read the proposal first no?

i think it depends on what the drawbacks are.  We have both "drawbacks" and "risks" (and also alternatives, which is related).

in my mind, the "drawbacks" are more about drawbacks to the concept/idea of doing this overall, and thus comes right after the motivation for why we *do* think we should be doing this.

risks are where you talk about limitations/challenges/technical obstacles to the design/implementation discussed in the proposal

zaneb:
> in my mind, the "drawbacks" are more about drawbacks to the concept/idea of doing this overall, and thus comes right after the motivation for why we _do_ think we should be doing this.

I guess I don't really agree here. Every design choice comes with trade-offs and it's important to go into a decision with a clear idea of what trade-offs you're making. Putting the "Drawbacks" before the "Proposal" (so that it is implicitly responding only to the idea of doing anything at all, as opposed to the concrete proposal) seems like an invitation to not think too hard about that.

Maybe this calls for something like an additional "Counter-arguments" subsection under "Motivations" instead?

bparees:
@kikisdeliveryservice @zaneb ok i think i'm convinced, i've pushed a new commit that relocates the drawbacks section to be part of the proposal, but retains the additional description about what the Drawbacks section should cover.

I debated trying to keep it split (i.e. an "anti-motivation" section, and then a drawbacks section) but that's probably not worth splitting and i couldn't come up with a good word for "anti-motivation" other than drawbacks anyway.

kikisdeliveryservice (approved):
/lgtm

PR #1102, opened 2022-04-29
Dual stack VIPs: Add rules for openshift/api fields on upgrades

creydr:
In #1048 we added and deprecated fields in openshift/api. 
To have consistent APIs regardless of upgrade or a fresh installation, we need one component to update the `XYPlatformStatus` fields correctly. This PR adds details how the values of these fields get updated.

### DIFF
--- a/enhancements/network/on-prem-dual-stack-vips.md
+++ b/enhancements/network/on-prem-dual-stack-vips.md
@@ -2,6 +2,7 @@
 title: dual-stack-vips
 authors:
   - @cybertron
+  - @creydr
 reviewers:
   - @creydr
   - @dougsland
@@ -15,7 +16,7 @@ api-approvers:
   - @danwinship
   - @aojea
 creation-date: 2022-03-01
-last-updated: 2022-04-08
+last-updated: 2022-08-02
 tracking-link:
   - https://issues.redhat.com/browse/SDN-2213
 see-also:
@@ -257,6 +258,21 @@ internal loadbalancer implementation. On upgrade, existing VIP configurations
 will be maintained. We will not (and cannot) automatically add new VIPs to a
 cluster on upgrade. If a deployer of an older dual stack cluster wants the new
 VIP functionality that will have to be a separate operation from upgrade.
+
+CNO takes care updating the new fields (`apiServerInternalIPs` &
+`ingressInternalIPs`) and old fields (`apiServerInternalIP` &
+`ingressInternalIP`) in [openshift/api](#openshiftapi) to have a consistent API
+between versions and therefore keeping clients consuming the old API functional.
+
+The following table shows the rules how the fields are set:
+
+| Case | Initial value of new field | Initial value of old field | Resulting value of new field | Resulting value of old field | Description |
+| ---- | -------------------------- | -------------------------- | ---------------------------- | ---------------------------- | ----------- |
+| 1    | _empty_                    | foo                        | [0]: foo                     | foo                          | `new` field is empty, `old` with value: set `new[0]` to value from `old` |
+| 2    | [0]: foo <br />[1]: bar    | _empty_                    | [0]: foo <br />[1]: bar      | foo                          | `new` contains values, `old` is empty: set `old` to value from `new[0]` |
+| 3    | [0]: foo <br />[1]: bar    | foo                        | [0]: foo <br />[1]: bar      | foo                          | `new` field contains values, `old` contains `new[0]`: we are fine, as `old` is part of `new` |
+| 4    | [0]: foo <br />[1]: bar    | bar                        | [0]: foo <br />[1]: bar      | bar                          | `new` contains values, `old` contains `new[1]`: we are fine, as `old` is part of `new` |
+| 5    | [0]: foo <br />[1]: bar    | baz                        | [0]: foo <br />[1]: bar      | foo                          | `new` contains values, `old` contains a value which is not included in `new`: new values take precedence over old values, so set `old` to value from `new[0]` (and log a warning) |
 
 ### Version Skew Strategy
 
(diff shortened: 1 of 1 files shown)


creydr:
Hello @sinnykumari,
as you are working on [MCO-322](https://issues.redhat.com/browse/MCO-322), this might be elevant for you too.
Could you PTAL here too?
Thanks in advance 

/assign @sinnykumari

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -258,6 +265,21 @@ will be maintained. We will not (and cannot) automatically add new VIPs to a
 cluster on upgrade. If a deployer of an older dual stack cluster wants the new
 VIP functionality that will have to be a separate operation from upgrade.
 
+MCO takes care updating the new fields (`apiServerInternalIPs` &
+`ingressInternalIPs`) and old fields (`apiServerInternalIP` &
+`ingressInternalIP`) in [openshift/api](#openshiftapi) to have a consistent API
```
sinnykumari, line 270: I didn't fully understand this. Did you mean to say that MCO will keep up-to date these new fields from openshift/api (cluster infrastructure CRD) into MCO controllerconfig ? Note that controllerconfig is MCO internal.

creydr: @sinnykumari: What I mean is that MCO will keep the values between the old fields (`apiServerInternalIP` & `ingressInternalIP`) and the new fields (`apiServerInternalIPs` & `ingressInternalIPs`) consistent. So users consuming the old API fields (`apiServerInternalIP` & `ingressInternalIP`) will get "updated" values. The rules of the update logic between these fields are described in the table later and will be implemented by @cybertron in https://github.com/openshift/machine-config-operator/pull/3269
The PR for the API changes is https://github.com/openshift/api/pull/1152

sinnykumari: got it, thanks for clarifying

sinnykumari:
LGTM from MCO side
/lgtm

sinnykumari:
Sorry, have another question from https://github.com/openshift/api/pull/1152#issuecomment-1109859075 which says:

> Had a quick talk in the team and we are not planning to set the new fields on upgrades. In MCO we will set the corresponding new fields in the controllerconfigs, but not in the infrastructure objects.

Why are we no updating infrastructure object during upgrade accordingly? In MCO, we don't mutate and diverge from what is set at cluster level objects e.g. proxy config, infra platform status, etc.

/lgtm cancel

creydr:
> Sorry, have another question from [openshift/api#1152 (comment)](https://github.com/openshift/api/pull/1152#issuecomment-1109859075) which says:
> 
> > Had a quick talk in the team and we are not planning to set the new fields on upgrades. In MCO we will set the corresponding new fields in the controllerconfigs, but not in the infrastructure objects.
> 
> Why are we not updating infrastructure object during upgrade accordingly? In MCO, we don't mutate and diverge from what is set at cluster level objects e.g. proxy config, infra platform status, etc.
> 
> /lgtm cancel

@sinnykumari this was the reason for this enhancement. First we thought about not updating the fields (what the comment is about), but then we noticed that people using the old API would have problems when the fields don't get updated. So we came up with this EP.

sinnykumari:
Adding summary of follow-up conversation happened on slack:
MCO role is to write content on the node and perform read only operation on any cluster level objects , not the other way around. As a result, MCO team is hesitant to add sync operation in MCO to modify cluster level infra object.
@wking brought up nice suggestion on possibility of using https://github.com/openshift/cluster-config-operator managing Infrastructure object. Perhaps that could be made pluggable, with sub-packages owned by different teams.

@creydr / @cybertron will investigate on using cluster-config-operator to update infra object field related to this EP.

creydr:
After checking with the API team on the CCO, we decided that CNO is a better place to have the sync logic. I updated this EP accordingly.

creydr:
Following up our discussion on [slack](https://coreos.slack.com/archives/CDCP2LA9L/p1659365304001269), we will add the sync logic to the cluster-network-controller.

PR for API changes: https://github.com/openshift/api/pull/1152
PR for sync controller in CNO: https://github.com/openshift/cluster-network-operator/pull/1519 (WIP, as we need the API changes to be merged before)

/assign @trozet @danwinship 
/cc @cybertron @vpickard

enhancements/network/on-prem-dual-stack-vips.md
```diff
@@ -258,6 +265,21 @@ will be maintained. We will not (and cannot) automatically add new VIPs to a
...
+
+| Case | Initial value of new field | Initial value of old field | Resulting value of new field | Resulting value of old field | Description |
+| ---- | -------------------------- | -------------------------- | ---------------------------- | ---------------------------- | ----------- |
+| 1    | _empty_                    | foo                        | [0]: foo                     | foo                          | `new` field is empty, `old` with value: set `new[0]` to value from `old` |
+| 2    | [0]: foo <br />[1]: bar    | _empty_                    | [0]: foo <br />[1]: bar      | foo                          | `new` contains values, `old` is empty: set `old` to value from `new[0]` |
+| 3    | [0]: foo <br />[1]: bar    | foo                        | [0]: foo <br />[1]: bar      | foo                          | `new` field contains values, `old` contains `new[0]`: we are fine, as `old` is part of `new` |
+| 4    | [0]: foo <br />[1]: bar    | bar                        | [0]: foo <br />[1]: bar      | bar                          | `new` contains values, `old` contains `new[1]`: we are fine, as `old` is part of `new` |
```
danwinship, line 280: The original `apiServerInternalIP` and `ingressInternalIP` values had to have been the cluster's primary IP family, so `old` should not be `new[1]`... either `old` is wrong or `new` is in the wrong order.

PR #1104, opened 2022-04-29
WIP: enhancements/olm: Recursive ClusterOperators

wking:
This enhancement proposes using ClusterOperator all over the place, not just for passing information up to the CVO.  For example, OLM would have ClusterOperators for its own operands, so it could hear about how they were doing.  Generic structures for everything!  Naive expectations that specifics and details won't matter! ;)

### DIFF
--- a/enhancements/olm/recursive-cluster-operators.md
+++ b/enhancements/olm/recursive-cluster-operators.md
@@ -1,0 +1,415 @@
+---
+title: recursive-cluster-operators
+authors:
+  - "wking"
+reviewers:
+  - "bparees, interesting guy"
+approvers:
+  - TBD
+api-approvers: # In case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers). If there is no API change, use "None"
+  - TBD
+creation-date: 2022-04-29
+last-updated: 2022-04-29
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - TBD
+---
+
+# Recursive ClusterOperators
+
+## Summary
+
+This enhancement proposes using ClusterOperator all over the place, not just for passing information up to the [cluster-version operator (CVO)][cluster-version-operator].
+
+## Motivation
+
+For all of OpenShift v4, we have had a [cluster-version operator (CVO)][cluster-version-operator] for managing OpenShift-core components in OpenShift clusters.
+The CVO recieves direction from cluster administrators and reports results via [ClusterVersion][].
+The CVO manages [hundreds of in-cluster resources via manifests](/dev-guide/cluster-version-operator/user/reconciliation.md#manifest-graph), including tens of "second level operators".
+Management includes:
+
+* Monitoring operand status, so we can complain to the admin (via ClusterVersion and alerts) if an operand is having difficulty.
+* Passing configuration changes (e.g. admin requests an update via ClusterVersion `spec.desiredUpdate.version="4.9.99"`, etc.), on to operands (e.g. setting `spec.containers[].image` in the etcd operator Deployment).
+* Managing interactions between operands (e.g. waiting for the Kubernetes API operator to report completed updates before asking dependent components like the machine-config operator to update).
+
+This is a generic operator pattern, and not something that is unique to the CVO.
+
+In many situations, operand resource types have a `status` that reports on their direct status.
+For example, the Deployment controller sets [`status` properties on managed Deployments][Deployment-status], like "`Progressing=True` with `reason=NewReplicaSetAvailable`; we are [complete][Deployment-complete]".
+That includes information about underlying ReplicaSets, which the Deployment controller manages on behalf of the Deployment, and recursively from Pods, which the ReplicaSet controller manages on behalf of the ReplicaSets.
+But it does not include information from an operand branching out to manage something directly, like the etcd operator managing etcd members.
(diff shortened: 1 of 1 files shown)


enhancements/olm/recursive-cluster-operators.md
```diff
@@ -0,0 +1,415 @@
...
+For example, OLM is is the business of installing and managing operators.
+Something installs and manages the Deployment operator.
+The monitoring operator installs and manages a Prometheus operator.
+All of these operators need a way to get information back out of their operands about how their operating is going, so they can in turn pass that information up to their parents.
+The ClusterOperator type is a very generic `status` object, and would be a good vessel for transmitting this information.
+
+### Tracking ownership
+
+No need for a `spec.parent` property or anything; we already have `metadata.ownerReferences`.
+Might need a namespaced variant?  I dunno how many operators can actually be installed with the same name in parallel namespaces.
```
jewzaam, line 57: Assume identical operator installs in multiple namespaces.  For example, ROSA/OSD addons today provision their own monitoring stack using prometheus operator (not UWM).  I think they use the same pattern.  So if 2+ addons are installed that need prometheus you have prometheus in 2+ namespaces.

enhancements/olm/recursive-cluster-operators.md
```diff
@@ -0,0 +1,415 @@
...
+* ...
+
+Life is great.
+
+#### CVO
+
+CVO probably dogfoods this pattern by growing its own ClusterOperator where it sets `conditions` and things.
+We'd keep `history`, `availableUpdates`, and other things that don't fit in ClusterOperator in our ClusterVersion.
+We'd probably set ourselves as `ownerReferences` if nobody else claimed us.
+But for OSD, [the managed upgrade operator][managed-upgrade-operator] drives the CVO, so it would add itself to `ownerReferences`, remove the CVO, and the CVO would respect "someone that's not me has taken ownership of my ClusterOperator; I guess I know my boss's name now".
```
jewzaam, line 85: This is not making sense to me.  Managed upgrade operator (MUO) is installed via OLM.  It would therefore have a CO with OLM as the owner.  MUO will manipulate spec of ClusterVersion CR.  Why does any hackery around ownership need to be done?

enhancements/olm/recursive-cluster-operators.md
```diff
@@ -0,0 +1,415 @@
...
+* Cluster level.  OLM sounds interesting.  Click.
+* OLM level.  ExampleOpA sounds interesting.  Click.
+* ExampleOpA level.  ExampleOpB sounds interesting.  Click.
+* ...
+
+Life is great.
+
+#### CVO
+
+CVO probably dogfoods this pattern by growing its own ClusterOperator where it sets `conditions` and things.
```
jewzaam, line 82: This implies a CO without an owner or where the owner is itself.  Moving conditions out of ClusterVersion will have impacts on a lot of tooling.  I don't see the benefit in this change.

enhancements/olm/recursive-cluster-operators.md
```diff
@@ -0,0 +1,415 @@
...
+### Upstream
+
+If we are pitching this as "hey, this is a fundamental gap", it makes sense to take it upstream.
+I have no idea what's upstream in this space; maybe there's already something there.
+
+### User Stories
+
+#### OLM
+
+OLM starts creating ClusterOperator for each of its operand operators, just like the CVO does today.
```
jewzaam, line 68: I really like this idea and having it automated.  Must handle that some operators already creating their own COs outside of CVO.  For example the aro-operator (for Azure Red Hat OpenShift) creates a CO called "aro".

enhancements/olm/recursive-cluster-operators.md
wking: Yeah, could do with with two `Operator` (or whatever you call the namespaced variant) in different namespaces that both use the `prometheus` name, or with two `ClusterOperator` that use different names `openshift-prometheus` and `openshift-user-workload-prometheus`, or whatever).

enhancements/olm/recursive-cluster-operators.md
wking: > This is not making sense to me. Managed upgrade operator (MUO) is installed via OLM.

Ahh, ownership cycle, with CVO driving OLM driving MUO driving the CVO.  MUO's config is what admins are supposed to be touching, not ClusterVersion, so I'd select MUO as the top of that cycle.  `ownerReferences` is just a clear way for MUO to declare "I'm driving this thing" on ClusterVersion, in parallel with `managedFields`.

enhancements/olm/recursive-cluster-operators.md
wking: `ownerReferences` only works when you're referencing in-cluster owners.  It can't point at human admins.  It probably can't even point across the HyperShift control-plane/managed-compute boundary.  So the `ownerReferences` there do need to remain empty or self-referential as a marker for "there is no in-cluster owner for this".  Benefit of the self-referential `ownerReferences` is that it is clearly distinct from "`ownerReferences` is empty because the owner didn't think to set it".

And I agree about the impact of dropping `conditions` from ClusterVersion, which is why I say "but backwards compat means that will probably never happen" a few lines below this sentence.

iNecas:
I like the idea of unified way of reporting about operators health and I can see as it would make it easier to deal with the observability data. Is the `recursive` the best way to describe this proposal though? I've seen very little actual recursion in the text itself (or I was reading it wrong?)

PR #1110, opened 2022-05-06
Track host ports used by metallb

fedepaol:
The metallb speakers are hostnetworked pods. There are a few ports that
need to be exposed outside of the nodes, for metrics and for leader
election, and another couple for the local non-encrypted metrics.

Here we start tracking them.

Note: the process of pinning to the right ports is still in progress (to the point that metrics are not working :P ), but we'll bring them back to 4.9.


fedepaol:
cc @russellb

PR #1114, opened 2022-05-10
WIP: Add option to manage DNS to ingresscontrollers

thejasn:
This enhancement provides the customer the ability to completely
disable DNS management on the Ingress Controller so that DNS can be
managed externally by them.

Signed-off-by: thejasn <thn@redhat.com>


enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+---
+
+# Configurable DNS Management for LoadBalancerService Ingress Controllers
+
+## Summary
+
+The ingress operator currently assumes that it's going to create and
+manage DNS using the required cloud provider integration, but in some
+circumstances when the DNS zone may be different from the cluster DNS zone
+and may not even hosted in the cloud provider. In such scenarios the operator
```
alebedev87, line 31:
```suggestion
and may not even be hosted in the cloud provider. In such scenarios the operator
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+after provisioning the required resources(load balancer, haproxy) in the cloud
```
alebedev87, line 32: `haproxy` is not provisioned in the cloud provider, it's a deployment in OpenShift.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+provider reports a degraded state because the provided DNS can't be created.
+
+This enhancement aims to provide the end-user the ability to completely disable
+DNS management on an Ingress Controller.
+
+## Motivation
+
+End users need to have an option to manage DNS on their own instead of relying
+on the `cluster-ingress-operator`. This is quite a common requirement where
+1. Customers may need deploy ingress controllers on domains hosted outside of
```
alebedev87, line 42:
```suggestion
1. Customers may need to deploy ingress controllers on domains hosted outside of
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+   cloud providers.
+2. Partners may need to deploy ingress controllers for domains from their own 
+   customers (with domains not hosted in the same cloud provider).
+
+With no support for such scenarious currently, customers end up with clusters
+in degraded state. The solution to this is to add support
```
alebedev87, line 48: I think that `degraded state` is just one of the possible results: when the DNS zone was not created by the installer.
The scenarios described above imply the usage of an external (to the cloud provider) DNS which doesn't exclude the existence of the cluster's DNS zone. So the controller may easily be in non degraded state but not willing to setup the DNS using the cluster's zone.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+    // <snip>
+
+    // manageDNS indicates if the lifecyle of the wildcard DNS record
+    // associated with the load balancer service will be managed by
+    // the ingress operator.
+    //
+	// +kubebuilder:default:=true
+	// +kubebuilder:validation:Optional 
+    // +optional
+    ManageDNS bool `json:"manageDNS"`
```
alebedev87, line 102: Just to foresee the potential remarks from the API team.
Maybe a generalization of the boolean would be better percieved, see also [the Kube conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) for `bool`.

I can think of something like:
```go
// +kubebuilder:default:="managed"
// +kubebuilder:validation:Optional
// +optional
DNS LoadBalancerDNSmanagement `json:"dns"`
}

type LoadBalancerDNSManagement string

var (
    ManagedLoadBalancerDNS LoadBalancerDNSManagement = "managed"
    UnmanagedLoadBalancerDNS LoadBalancerDNSManagement = "unmanaged"
)
```

WDYT?

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+}
+```
+
+Re-uses existing state `DNSManaged` of the controller to denote to denote if
```
alebedev87, line 106:
```suggestion
Re-uses existing state `DNSManaged` of the controller to denote if
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+mutually exclusive i.e. if customer has opted to disable DNS management cluster
+wide (as mentioned under UPI based installations), this additional flag is of
+no value and need not be set since *DNSRecord* created is silently reconciled 
+by the dns controller.
+
+### Risks and Mitigations
+
+The manual creation of the required DNS records will need to be done prior to
+making this change as the canary controller will try to ensure the if the ingress
+domains are reachable, otherwise the `ingress` *ClusterOperator* results in a
```
alebedev87, line 125:
```suggestion
domains are reachable, otherwise the `cluster-ingress-operator` gets into a 
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+wide (as mentioned under UPI based installations), this additional flag is of
+no value and need not be set since *DNSRecord* created is silently reconciled 
+by the dns controller.
+
+### Risks and Mitigations
+
+The manual creation of the required DNS records will need to be done prior to
+making this change as the canary controller will try to ensure the if the ingress
+domains are reachable, otherwise the `ingress` *ClusterOperator* results in a 
+degraded state. This needs to be documented so as to avoid unecessary support cases.
```
alebedev87, line 126:
```suggestion
degraded state. This needs to be documented to avoid unnecessary support cases.
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Drawbacks
+
+There is an inconsistency in how the existing support is added to disable
```
alebedev87, line 130: I find it quite consistent. Before the proposed solution the ingress controller was always managing `DNSRecord` CRs for its domain. So it keeps managing them fulfilling its contract with the customer no matter what. The fact that the zone may not exist has to be handled but it doesn't free the ingress controller from its contract. Only the new API field you are adding may do this.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+controller [silently skips](https://github.com/openshift/cluster-ingress-operator/blob/972f09b9dbb181ae5c414da2a990b57c60fde9d8/pkg/operator/controller/dns/controller.go#L261-L311)
+publishing the record to the zone. Where as in the proposed solution, 
+a wildcard *DNSRecord* is **not** created when `manageDNS` is set to `false`. 
+
+## Design Details
+
+### Test Plan
+
+- Test by updating the default ingresscontroller to be associated with a manually
+  created dns record and ensure correct working of test workload.
```
alebedev87, line 142: I imagine that the cluster provisioned by the CI will have the cluster DNS zone and the cluster ingress operator installed, so the wildcard for the routes will already be there when this test case will hit. This means that even if you setup the wildcard DNS somewhere else and disable the dns management for the default ingress controller: how would you make sure that the DNS resolution is still not going from the old DNS records (created at the installation)?

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Upgrade / Downgrade Strategy
+
+On upgrades, this field will default to `true` which is consistent with the
+older versions of the ingress operator.
+
+On downgrades, 
+- If downgrading to an unsupported version while having the `manageDNS` set to
+  `false` could result in failing to create a wildcard *DNSRecord* and therby
+  resulting in the ingress *ClusterOperator* in a degraded state.
```
alebedev87, line 170: `unsupported version` means the cluster ingress operator which didn't know anything about this field? If so, won't it create non existing `DNSRecord` CRs?

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,199 @@
...
+N/A.
+
+### Operational Aspects of API Extensions
+
+N/A.
+
+#### Failure Modes
+
+- If the customer doesn't create the appropriate manual records needed
+  for the `default` ingresscontroller, then the ingress *ClusterOperator*
```
alebedev87, line 186:
```suggestion
  for the `default` ingresscontroller, then the `cluster-ingress-operator`
```
Just to be consistent with the previous writings of the cluster ingress operator.

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: Within the test case, we would be updating the `ingresscontroller` to update the `domain` and `manageDNS` field to the required values. Followed by creating a new DNS zone and record which is associated to the default *LoadBalancerService* `router-default`'s external IP. 

The update will ensure the older *DNSRecord* is deleted and using something like an echo server (workload) the new domain will be tested.

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: I checked this out as well, looks like there is another issue where updating the [domain](https://github.com/openshift/cluster-ingress-operator/blob/972f09b9dbb181ae5c414da2a990b57c60fde9d8/pkg/operator/controller/ingress/controller.go#L342-L355) in the ingresscontroller doesn't result in [updating](https://github.com/openshift/cluster-ingress-operator/blob/972f09b9dbb181ae5c414da2a990b57c60fde9d8/pkg/operator/controller/ingress/dns.go#L97) the wildcard DNS.  

If we fix this issue, the dns_controller will end up with an error when it tries to publish the new record to the old zone. This would lead to the `DNSReadyIngressConditionType` to be set to false, resulting in degraded state.

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: So the reason why I feel it is inconsistent is because, once this new field is introduced, some controllers will not have any *DNSRecord*s associated with them at all compared to the scenario where the customer setup the installer without zones but still the ingresscontrollers have *DNSRecords* associated with them just to denote the association between the domain and the target external IP. (although since it's using a fake provider, the association isn't actually ensured by the operator).

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: Yes, correct. These *DNSRecord* CRs will put the operator in a degraded state. At this point, the only way to recover is to disable DNS management cluster wide. I'll mention the same as well.

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: Yeah, this would make sense. Future use case could involve denoting managed by externalDNS. I'll make the required changes.

enhancements/ingress/configurable-dns-loadbalancerservice.md
alebedev87: I think I see what you mean. In my opinion this comes from the unprecise wording in the UPI markdown docs:

> If you don't want [the ingress operator](https://github.com/openshift/cluster-ingress-operator) to manage DNS records on your behalf, remove the privateZone and publicZone sections from the DNS configuration:

They are putting it like the cluster ingress operator explicitly stops doing any DNS related things if no cluster zone is present. While in reality the operator keeps doing what it was doing, but since there is no zone - no actual DNS records is created. The final result is the same, I agree, but the intention is different. I see it as a sort of separation of concerns: zones are managed in one place, ingress operator does its job in another place; and only this lonely doc puts an explicit link between these two.

Well, overall, I  don't think it's a real drawback, it's maybe a slightly unprecise doc but I don't think it's something we should worry about for this EP.

enhancements/ingress/configurable-dns-loadbalancerservice.md
alebedev87: Yes, the degraded state is indeed quite a probable outcome if the above mentioned scenarios are tried to be implemented without the proposed solution. I think what poked into my eye is the wording `degraded cluster` while I thought more of `degraded ingress operator`.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,245 @@
...
+### Upgrade / Downgrade Strategy
+
+On upgrades, `.loadBalancer.dns` will default to `Managed` which is consistent with the
+older versions of the ingress operator.
+
+On downgrades, 
+- If downgrading to an unsupported version while having the `.loadBalancer.dns` set to
+  `Unmanaged` could result in failing to create a wildcard *DNSRecord* and thereby
+  resulting in the `cluster-ingress-operator` in a degraded state. To recover
+  DNS management will have to disabled cluster wide ([docs](https://github.com/openshift/installer/blob/master/docs/user/aws/install_upi.md#remove-dns-zones)).
```
sherine-k, line 216:
```suggestion
  DNS management will have to be disabled cluster wide ([docs](https://github.com/openshift/installer/blob/master/docs/user/aws/install_upi.md#remove-dns-zones)).
```
I think we should only suggest to disable DNS management cluster wide if we have only one (unmanaged DNS) ingress controller. In the event of a mix of managed and unmanaged, this might lead to the cluster not reconciling DNS records it is responsible for.
I also had a question concerning the downgrade: Does the degraded state come from the fact that the controller will try to recreate an existing DNS record (for a domain it doesnt have)? Or does it come from the canaries ? 
In the latter case, canaries should still be able to reach this externally managed DNS wildcard record.

sherine-k (commented):
Hi @thejasn,
as promised, a few typo fixes suggested.
along with 1 question...
Great job

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: > I think we should only suggest to disable DNS management cluster wide if we have only one (unmanaged DNS) ingress controller. In the event of a mix of managed and unmanaged, this might lead to the cluster not reconciling DNS records it is responsible for.

Added that as a condition.

> I also had a question concerning the downgrade: Does the degraded state come from the fact that the controller will try to recreate an existing DNS record (for a domain it doesnt have)? Or does it come from the canaries ?
In the latter case, canaries should still be able to reach this externally managed DNS wildcard record.

This depends on which controller has been configured to unmanaged. 
* If the default controller has the flag, will result in canaries not able to reach the domain. If the external domain has been    created, then it errors when trying to create the DNS record. (currently hinges on [open question](https://github.com/openshift/enhancements/blob/f528e26c377c355164080be3a9df11490b591988/enhancements/ingress/configurable-dns-loadbalancerservice.md#open-questions)) 
* If it is not the default controller, canaries are not in the picture, so it only errors when trying to create the DNS record.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,245 @@
...
+# Configurable DNS Management for LoadBalancerService Ingress Controllers
+
+## Summary
+
+The ingress operator currently assumes that it's going to create and
+manage DNS using the required cloud provider integration, but in some
+circumstances when the DNS zone may be different from the cluster DNS zone
+and may not even be hosted in the cloud provider. In such scenarios the operator
+after provisioning the required resources in the cloud provider reports a 
+degraded state because the provided DNS can't be created.
```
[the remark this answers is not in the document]
thejasn: Fixed.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,245 @@
...
+#### As a cluster admin, when updating an ingresscontroller from managed DNS to unmanaged DNS, I want the previous DNS records to be deleted
+
+Upon updating `.loadBalancer.dns` to `Unmanaged` the older DNS records that were created
+must be deleted by the ingress operator.
+
+### Goals
+
+- Provide the ability to opt in for DNS management by the `cluster-ingress-operator`.
+- Recover from degraded state of the `cluster-ingress-operator` during upgrades
+  from OCP 3.x to 4.x where DNS was being managed externally,
```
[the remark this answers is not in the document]
thejasn: Fixed.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### Modifying the default ingresscontroller
+
+The default ingresscontroller will be created with domain present in
+*ingress.config.openshift.io/cluster* `.spec.domain`. If we want to 
+update the domain present on the ingresscontroller with a domain
+outside the defined cluster DNS zone,
+1. The new domain that needs to be associated with the ingresscontroller
+    must be created prior to making any changes.
+2. The ingresscontroller must be edited with a new domain and
```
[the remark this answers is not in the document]
thejasn: Fixed.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,245 @@
...
+wide (as documented
+[here](https://github.com/openshift/installer/blob/master/docs/user/aws/install_upi.md#remove-dns-zones)),
+this field is of no value and need not be set since *DNSRecord* created is silently
+reconciled by the DNS controller.
+
+### Risks and Mitigations
+
+The manual creation of the required DNS records will need to be done prior to
+making this change as the canary controller will try to ensure the if the ingress
+domains are reachable, otherwise the `cluster-ingress-operator` gets into a
```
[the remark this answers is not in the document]
thejasn: Fixed.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,245 @@
...
+// LoadBalancerStrategy holds parameters for a load balancer.
+type LoadBalancerStrategy struct {
+    // <snip>
+
+    // dns indicates if the lifecyle of the wildcard DNS record
+    // associated with the load balancer service will be managed by
+    // the ingress operator.
+    //
+    // +kubebuilder:default:="Managed"
+    // +kubebuilder:validation:Optional
```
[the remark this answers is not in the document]
thejasn: Fixed.

enhancements/ingress/configurable-dns-loadbalancerservice.md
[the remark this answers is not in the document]
thejasn: Fixed.

alebedev87:
/retitle [NE-568] Add option to manage DNS to ingresscontrollers

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,316 @@
...
+// DNSRecordSpec contains the details of a DNS record.
+type DNSRecordSpec struct {
+  // <snip>
+
+  // managementPolicy denotes the current policy applied on the dns record.
+  // Records that have policy set as "Unmanaged" are ignored by the ingress
+  // operator and should be managed at the discretion of the cluster admin.
+  //
+  // +kubebuilder:default:="Managed"
+  // +kubebuilder:validation:Optional
```
brandisher, line 165: Sharing a bit of feedback I've gotten from the API team on this: https://github.com/openshift/api/pull/1110#discussion_r836645159

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,316 @@
...
+type LoadBalancerStrategy struct {
+    // <snip>
+
+    // dns indicates if the lifecyle of the wildcard DNS record
+    // associated with the load balancer service will be managed by
+    // the ingress operator.
+    //
+    // +kubebuilder:default:="Managed"
+    // +kubebuilder:validation:Optional
+    // +kubebuilder:validation:Enum=Managed;Unmanaged
```
brandisher, line 136: I would clear this with the API team before changing but I did get a few bits of feedback on this sort of approach with my API PR:
* [Describe what an empty value means](https://github.com/openshift/api/pull/1110#discussion_r818889153)
* [Additional Discussion](https://github.com/openshift/api/pull/1110#pullrequestreview-925198254)

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,316 @@
...
+### Implementation Details/Notes/Constraints
+
+Based on `.loadBalancer.dns`, the controller decides whether to ensure creating
+a wildcard *DNSRecord*. If the customer is updating `.loadBalancer.dns` from 
+`"Managed"` to `"Unmanaged"` the previously created *DNSRecord* will be updated
+appropriately.
+
+The existing state of the ingresscontroller will be synced to the *DNSRecord* as
+well, to clearly state which dns record is unmanaged. The cluster admin can choose
+to delete the *DNSRecord* at his own discretion.
```
brandisher, line 199:
```suggestion
The existing `.loadBalancer.dns` value of the ingresscontroller will be replicated to the *DNSRecord*'s `.spec.managementPolicy` as well, to clearly state which dns record is unmanaged. The cluster admin can choose to delete the *DNSRecord* at his own discretion. 
```

This seemed like what the intent of the wording was but feel free to adjust/ignore if that's incorrect. I assumed that what you were trying to state was that the management policy of the IC will be replicated to the management policy of the corresponding DNSRecord.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,316 @@
...
+Based on `.loadBalancer.dns`, the controller decides whether to ensure creating
+a wildcard *DNSRecord*. If the customer is updating `.loadBalancer.dns` from 
+`"Managed"` to `"Unmanaged"` the previously created *DNSRecord* will be updated
+appropriately.
+
+The existing state of the ingresscontroller will be synced to the *DNSRecord* as
+well, to clearly state which dns record is unmanaged. The cluster admin can choose
+to delete the *DNSRecord* at his own discretion. 
+
+Once the *DNSRecord* is set to `Unmanaged`, any updates done to it will be ignored
```
brandisher, line 201: One implementation note on this; it'll be important that any `DNSRecord`s skipped during the reconcile cause a log to be emitting indicating which record was skipped and why. This'll be important for the RH Support team when troubleshooting ingress issues related to this feature.

Similarly in the `Unmanaged` -> `Managed` case, having a log indicating that the `DNSRecord` was reconciled and had it's status changed will be important for troubleshooting.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,316 @@
...
+if not all components are properly updated. This should be done at the discretion
+of the cluster admin. The manual creation of the required DNS records will need to be
+done prior to making this change, as the canary controller will try to ensure that the
+ingress domains are reachable, and otherwise this sets the `cluster-ingress-operator`
+in a degraded state.
+
+### Drawbacks
+
+After an ingresscontroller is updated from `Managed` to `Unmanaged`, the previous
+*DNSRecord* is left in an `Unmanaged` state which could result in orphaned resources
```
brandisher, line 225: I'm curious about what "orphaned" means in this context. Does it mean that if your IC's `.spec.dns` is set to `Unmanaged` and you delete your IC that the `DNSRecord` CRs will stay on the cluster? If that's the case, this may be a good opportunity for an alert indicating that there are orphaned resources in cases where you have DNSRecord objects with no corresponding IC. I haven't looked too closely at whether that can happen though so it'd be worth validating first.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,316 @@
...
+the wildcard *DNSRecord*. 
+> 1. Can this logic be changed to support moving from managed DNS to
+     unmanaged DNS on the `default` ingresscontroller?
+     
+  __Closed:__ The operator will continue to retain immutability of `.spec.domain`
+              and will not support updating domain when migrating from `Managed` to 
+              `Unmanaged` DNS.
+  
+
+### Test Plan
```
brandisher, line 244: This may or may not be the right spot for it but the orphaned `DNSRecord` case should be tested as well.

brandisher (changes requested):
@thejasn overall I understand the intent but I did have a few questions/comments and some feedback to share. For any of my API related comments, defer to the API Team for final say. I mostly wanted to share my recent API experience to help move the needle.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,316 @@
...
+
+    // dns indicates if the lifecyle of the wildcard DNS record
+    // associated with the load balancer service will be managed by
+    // the ingress operator.
+    //
+    // +kubebuilder:default:="Managed"
+    // +kubebuilder:validation:Optional
+    // +kubebuilder:validation:Enum=Managed;Unmanaged
+    // +optional
+    DNS LoadBalancerDNSManagementPolicy `json:"dns"`
```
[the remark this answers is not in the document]
thejasn: Are you suggesting sharing the type variable between the fields? I wanted to avoid that, hence the different names. Sharing the type doesn't seem correct, since it would couple the two objects `LoadBalancerStrategy` and `DNSRecord` when any new types are added (eg: `externalDNS`). WDYT?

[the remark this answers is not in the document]
brandisher: Ah, that's a great question! I wasn't as clear as I intended to be; this is the change I'm suggesting for clarity.

```suggestion
    DNSManagementPolicy LoadBalancerDNSManagementPolicy `json:"dns"`
```
That way you're maintaining the distinction in type between `LoadBalancerDNSManagementPolicy` and `DNSManagementPolicy` but when a cluster admin is reading the API/config it's clear that they're changing the `loadBalancerStrategy.dnsManagementPolicy` which is, in my opinion, more clear than `loadBalancerStrategy.dns`. Also just a reminder that ultimately the API team should make the call on the shape of the API but I suspect they'd err on the side of making configuration options more obvious.

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: Yes, your understanding is correct. So by alert do you mean some kind of `Event` or do you mean a `Log`? (Both are feasible)

enhancements/ingress/configurable-dns-loadbalancerservice.md
alebedev87: I believe that the idea of using both `+optional` and `+kubebuilder:validation:Optional` comes from the support of different code generators. `+kubebuilder:*` markers work only for kubebuilder while `+optional` is a universal convention which can enable the generation done by a different code generator.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,316 @@
...
+
+- Provide the ability to opt in for DNS management by the `cluster-ingress-operator`.
+- Recover from degraded state of the `cluster-ingress-operator` during upgrades
+  from OCP 3.x to 4.x where DNS was being managed externally.
+
+### Non-Goals
+
+- Provide a day-0 solution for cluster installations involving external DNS
+  management by the customer.
+- DNS migration from managed to unmanaged.
```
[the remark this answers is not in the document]
thejasn: Ahh, this was meant to state traffic management is not going to be considered during moving from `Managed` to `Unmanaged`. I've updated the point, to make it more clear.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,320 @@
...
+    UnmanagedDNS.
+   
+    
+__Note__: Similarly, to create a new ingresscontroller where DNS is managed
+          externally, the same workflow can be followed. No *DNSRecord*s will be
+          created in the flow.
+
+### API Extensions
+
+The ingresscontroller CRD is extended by adding an optional field `DNS`
```
alebedev87, line 122:
```suggestion
The ingresscontroller CRD is extended by adding an optional field `DNSManagementPolicy`
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: Looking at the existing alerts, they seem to be very critical in nature, i.e. scenarios where the ingresscontroller is completely non-functional or where it has incorrect configs. In contrast to that, having an alert for orphaned *DNSRecord*s can't really be justified when these orphaned resources don't impact the functioning of the ingresscontroller. 

One of the main reasons we decided not to delete the resource was to enable the cluster admin to create the required DNS record manually looking at the *DNSRecord* (has all the required information, otherwise would need to gather information from the routes and ingresscontroller).

enhancements/ingress/configurable-dns-loadbalancerservice.md
alebedev87: 1. No, I don't think so.
2. No. Yes, I agree that the kubebuilder specific marker is preferred for the code generation.

When I saw them both specified I was also wondering whether this redundancy is really needed, however I thought that keeping them would still be needed because:

- it's consistent with the other OpenShift APIs, [operator's ingress one too](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go) (even though just `+optional` is used in some places).
- it's part of [Kube API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) (OpenAPI documentation is mentioned as a tool which may need it)

enhancements/ingress/configurable-dns-loadbalancerservice.md
brandisher: > having an alert for orphaned DNSRecords can't really be justified when these orphaned resources don't impact the functioning of the ingresscontroller.

I think we're coming from different perspectives and that's maybe muddying the water a bit. From an operator perspective, I agree with you. No need to alert on the operator if the operator is functioning normally. However, consider the customer or SD perspective when an IC is accidentally deleted or they discover (without an alert) that they unexpectedly have _n_ orphaned DNSRecords. Maybe the middle-ground in this is to expose the metric so it can be captured and queried but do not create an alert. Then we're solving both needs by making the orphaned DNSRecord count easily retrievable without necessarily indicating to cluster admins that they need to take urgent action. What do you think?

I'd be curious whether @Miciah has an opinion on this.

enhancements/ingress/configurable-dns-loadbalancerservice.md
brandisher: Yeah for sure. To be clear I'm only sharing for the sake of sharing; not necessarily anything actionable here :smile:

enhancements/ingress/configurable-dns-loadbalancerservice.md
brandisher: ack, understood. I think it may be worth getting some more formal guidance on this from the API team. With that said, we can close this thread and let the API team decide what needs to happen (if anything). Thanks @alebedev87 !

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: I'm still not very sure as to what problem we are attempting to solve with this metric. Are we trying to solve the problem where cluster admins forget to clean up the *DNSRecord* post migrating from `Managed` to `Unmanaged` or are we looking for a way to provide more information to cluster admins about the current state of ingresscontroller (especially about orphaned *DNSRecord* count)?  

If it is the latter, and we want to give more information about the *DNSRecords* (counts) it still doesn't seem to justify a metric, especially when the transition from `Managed` to `Unmanaged` is more of a one-time operation (I don't see customers frequently toggling this). And if we are talking about exposing other details (`IC is accidentally deleted`) better to take that up separately. 

If it is the former, I think documentation is sufficient, since this procedure of migrating from `Managed` to `Unmanaged` is very infrequent. We can maybe have `WARN` logs on the controller where we ignore the *DNSRecord* to nudge the cluster admin to clean up.

brandisher: Stepping back a bit, if the question from a cluster admin is "How do I see if I have orphaned DNSRecord objects?" How would you answer that question?

thejasn: The cluster admin can fetch all *DNSRecords* and filter where `managementPolicy` is set to `Unmanaged`. 
```bash
oc get dnsrecord -A  -o=jsonpath="{.items[?(@.spec.managementPolicy=='Unmanaged')].metadata.name}"
``` 
Maybe we can introduce a new status into *DNSRecord*, which can be used as filter.

brandisher: @thejasn thanks for the time on the call! It sounds like the current implementation allows for DNSRecords to be orphaned when an IC is deleted (correct me if I'm wrong) so in order to avoid that, the implementation of this EP will ensure that DNSRecords do not get orphaned when an IC is deleted. That'll remove the need for a metric/alert around orphaned DNSRecord objects and result in a better UX since we won't be leaving objects around. One question this raises for me though is, do we have any use case for leaving the DNSRecord around in an orphaned state?

thejasn: @brandisher Existing implementation doesn't allow orphaned DNSRecords, the ingresscontroller adds a finalizer to the *DNSRecord*, so that it can consequently be cleaned up during deletion of the IC. This EP will introduce the behavior where we would end up with *DNSRecords* which aren't associated with any IC (orphaned), these can be queried by the cluster admin as shown above  and proceed to delete. 

We will ensure the cluster admin deletes the *DNSRecord* by not removing the finalizer on the *DNSRecord* when the IC was marked for deletion. So the workflow would be something like, 
1. Set the field to `managementPolicy: Unmanaged`
2. Later on, when the admin deletes IC (tries to orphan the *DNSRecord*)
3. Delete the finalizer on the *DNSRecord* manually so that IC can also be deleted. 
This would ensure the there are no orphaned resources. 

Regarding the adding a status/condition on *DNSRecord*, had a discussion with @Miciah, and we concluded that querying using the spec is sufficient and adding a new condition is redundant since the content is pretty much the same.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,348 @@
...
+- This will trigger a reconcile of the controller, resulting in updating
+  the following conditions on the ingress operator
+  - `DNSManaged` condition will be set to false and reason updated to
+    UnmanagedDNS.
+- The DNSRecord will also be updated as part of the same reconcile, 
+  `.spec.dnsManagementPolicy` will be set to `Unmanaged` and the following conditions
+  will be updated,
+  - `DNSReady` condition will be set to true and reason updated to
+    UnmanagedDNS.
+- Post successfully updating the ingresscontroller, the associated *DNSRecord*
```
thejasn, line 123: Added deletion workflow.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,348 @@
...
+The existing `.loadbalancer.dnsManagementPolicy` value of the ingresscontroller
+will be replicated to the *DNSRecord*'s `.spec.dnsManagementPolicy` as well, to
+clearly state which dns record is unmanaged. The cluster admin can choose to delete
+the *DNSRecord* at his own discretion (this will involve also manually removing the finalizer).
+
+Once the *DNSRecord* is set to `Unmanaged`, any updates done to it will be ignored
+by the operator, but it is to be noted that if the ingresscontroller is updated from
+`Unmanaged` to `Managed` any changes done will be lost during re-sync.
+
+If the ingresscontroller is deleted, and the associated *DNSRecord* still exists
```
thejasn, line 222: Clarify the implementation details on how deletion of ingresscontrollers will be done.

alebedev87: I know that there is probably no easy way out. But I think it still a bit contradictory. In [this comment](https://github.com/openshift/enhancements/pull/1114/files#r886931980) we say that `DNSRecord` CR won't be created/updated/deleted, however the deletion will still happen: `deletionTimestamp` will be added. Also the same comment says that it's the full responsibility of the cluster admin to deal with `DNSRecord` CRs however the cluster admin won't do `oc delete dnsrecord` it has to be `oc edit dnsrecord` which is kinda not intuitive.

I think I'm rather for one of the following:
- "fully orphaned" unmanaged `DNSRecord` CR: cluster admin has to do `aws route53 delete resource record` + `oc delete dnsrecord`. The orphaned `DNSRecord` CRs can be found with [the command you suggested](https://github.com/openshift/enhancements/pull/1114/files#r881753249).
- "cloud provider orphaned" unmanaged `DNSRecord` CR: cluster admin has to do `aws route53 delete resource record` only. `DNSRecord` will be cleared from the finalization label without doing the actual finalization.

@thejasn, @brandisher: WDYT guys?

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+```
+
+Re-uses existing condition `DNSManaged` of the controller to denote if
+the wildcard *DNSRecord* associated with the controller is managed/unmanaged. 
+
+### Implementation Details/Notes/Constraints
+
+Based on `.loadbalancer.dnsManagementPolicy`, the controller decides whether to ensure creating
+a wildcard *DNSRecord*. If the customer is updating `.loadbalancer.dnsManagementPolicy` from 
+`"Managed"` to `"Unmanaged"` the previously created *DNSRecord* will be updated
```
Miciah, line 209: The terminology is overloaded, so better to be specific.
```suggestion
Based on an IngressController's `spec.endpointPublishingStrategy.loadBalancer.dnsManagementPolicy` field, the ingress operator decides whether to ensure creating
a wildcard *DNSRecord* CR. If the user is updating the IngressController's `dnsManagementPolicy` field from 
`"Managed"` to `"Unmanaged"`, the previously created *DNSRecord* CR will be updated
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+appropriately.
+
+The existing `.loadbalancer.dnsManagementPolicy` value of the ingresscontroller
+will be replicated to the *DNSRecord*'s `.spec.dnsManagementPolicy` as well, to
+clearly state which dns record is unmanaged. The cluster admin can choose to delete
+the *DNSRecord* at his own discretion (this will involve also manually removing the finalizer).
```
Miciah, line 215: This needs to be updated per https://github.com/openshift/api/pull/1200#issuecomment-1144081676.  
```suggestion
clearly state which dns record is unmanaged. If cluster admin were to delete
the *DNSRecord* CR, the operator would recreate it (again reflecting and respecting the IngressController's `spec.endpointPublishingStrategy.loadBalancer.dnsManagementPolicy` field).
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+
+Once the *DNSRecord* is set to `Unmanaged`, any updates done to it will be ignored
+by the operator, but it is to be noted that if the ingresscontroller is updated from
+`Unmanaged` to `Managed` any changes done will be lost during re-sync.
```
Miciah, line 219: I needed to think about this to make sure there wasn't a race condition here:

1. `oc -n openshift-ingress-operator patch ingresscontrollers/default --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"loadBalancer":{"dnsManagementPolicy":"Unmanaged"}}}}'`
2. `oc -n openshift-ingress patch dnsrecords/default-wildcard --type=merge --patch='{"spec":{"recordTTL":123}}'`
3. `oc -n openshift-ingress-operator patch ingresscontrollers/default --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"loadBalancer":{"dnsManagementPolicy":"Managed"}}}}'`

We should be fine because the operator will patch the DNSRecord CR's `spec.recordTTL` field at the same time that the operator patches the CR's `spec.dnsManagementPolicy` field, so the dns controller will never will never observe the user-patched TTL at the same time that it observes "Managed".  

It shouldn't _really_ matter because users should _never_ patch the DNSRecord CR, but... you never know.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+`Unmanaged`, the *DNSRecord* CR will be deleted as per current implementation but the
+DNS record on the cloud provider will not be deleted by the operator and will require
+the cluster admin to manually delete it. 
+
+The support provided by the installer and this new field aren't entirely
+mutually exclusive, i.e. if customer has opted to disable DNS management cluster
+wide (as documented
+[here](https://github.com/openshift/installer/blob/master/docs/user/aws/install_upi.md#remove-dns-zones)),
+this field is of no value and need not be set since *DNSRecord* created is silently
+reconciled by the DNS controller.
```
Miciah, line 231:
```suggestion
this field is of no value and need not be set since reconciliation of the *DNSRecord*
CR is a no-op when no DNS zones are configured.
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+
+__Note:__ Appropriate logging is a must on all controllers clearly indicating the
+current DNS management policy set on both ingresscontroller and DNSRecord and any
+subsquent effects such as skipping reconciliation, etc must also be logged.
```
Miciah, line 235:
```suggestion
subsequent effects, such as skipping reconciliation etc., must also be logged.
```
(It's also important to set clear status in the CRs, which may provide sufficient information on its own or may be the cue for the cluster admin to check the logs.)

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+### Test Plan
+
+- Test by creating a new secondary ingresscontroller with a custom domain
+  and `.loadbalancer.dnsManagementPolicy` set to `Unmanaged`. Ensure proper conditions
+  and connectivity.
+- Test deletion of ingresscontrollers to ensure correct behaviour of the associated
+  *DNSRecord* CR and the DNS record on the cloud provider.  
+- Test the following update paths on a custom ingresscontroller
+  - Updating `.loadbalancer.dnsManagementPolicy` from `Managed` -> `Unmanaged` -> `Managed` and
+    to ensure no conflicts during creation/recreation of DNS record on the cloud provider.
```
Miciah, line 278: Why "recreation"?  Is the test going to delete the DNS record after setting `dnsManagementPolicy` to "Unmanaged"?

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+  - Updating `Unmanaged` -> `Managed` and to ensure creation of new *DNSRecord*
+    and correct conditions.
```
Miciah, line 280: Is the idea here to create the IngressController with "Unmanaged", update it to "Managed", and verify that the operator creates the DNS record upon the update to the IngressController?  Here, "DNSRecord" means the actual DNS record and not the CR, right?

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+On downgrades, there are 2 possible variations
+* When downgrading to an unsupported version, having the `.loadbalancer.dnsManagementPolicy` set to
+  `Unmanaged`,
+  * If the cluster domain is not the same as the domain in the ingresscontroller,
+    could result in failing to create the wildcard *DNSRecord* associated on the 
+    ingresscontroller and thereby result in the `cluster-ingress-operator` entering 
+    a degraded state. To recover DNS management will have to be disabled cluster 
+    wide ([docs](https://github.com/openshift/installer/blob/master/docs/user/aws/install_upi.md#remove-dns-zones)).
+  * When the cluster domain is the same as the domain in the ingresscontroller,
+    this will not cause any issues during downgrades, but a new wildcard *DNSRecord*
```
Miciah, line 312: If I understand correctly, the CR would already exist, but the DNSRecord CR's `spec.dnsManagementState` field would go away, and the downgraded ingress operator's dns controller would proceed to create the _actual_ DNS record.
```suggestion
    this will not cause any issues during downgrades, but a new wildcard DNS record
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+
+N/A.
+
+### Upgrade / Downgrade Strategy
+
+On upgrades, `.loadbalancer.dnsManagementPolicy` will default to `Managed` which is consistent with the
+older versions of the ingress operator.
+
+On downgrades, there are 2 possible variations
+* When downgrading to an unsupported version, having the `.loadbalancer.dnsManagementPolicy` set to
```
Miciah, line 304: The word "unsupported" is confusing here and below.  
```suggestion
* When downgrading to a version of OpenShift without this enhancement, having the `.loadbalancer.dnsManagementPolicy` set to
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+### Operational Aspects of API Extensions
+
+N/A.
+
+#### Failure Modes
+
+- Application workloads will experiance connectivity issues, when replacing/re-creating
+  an ingresscontroller from `Managed` to `Unmanaged` since the associated deployment,
+  routes, loadbalancer will be recreated/reconfigured. Eventually, once all components
+  reach stable conditions, workload connectivity should be restored.
```
Miciah, line 331:
```suggestion
- Application workloads will experience connectivity issues, when replacing/re-creating
  an ingresscontroller from `Managed` to `Unmanaged` since the operator must recreate
  the associated deployment and loadbalancer and the cluster administrator must
  reconfigure DNS. Eventually, once all components reach stable conditions, workload
  connectivity should be restored.
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+
+### Workflow Description
+
+This feature is designed as Day-2 solution and is geared towards cluster
+admins. It mainly applies to scenarios where specific ingresscontrollers
+need to be configured to not manage DNS records associated with them.
+
+__Note__: This feature is only supported on new or non-default ingresscontrollers.
+The default ingresscontroller can be modified/updated at the discretion of
+the cluster admin.
```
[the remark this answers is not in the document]
thejasn: Yes, the cluster admin can choose to change the `dnsManagementPolicy` of the default ingresscontroller, but that is out of the scope of this feature since that could potentially involve updating the domain as well.

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: Yes, that is what I had in mind. Thinking of testing both the scenarios to verify there were no collisions during recreation.

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: Yes, that is what would happen.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+  must be created prior to making any changes.
+- The ingresscontroller must be edited to set `.loadbalancer.dnsManagementPolicy` to `Unmanaged`.
+- This will trigger a reconcile of the controller, resulting in updating
+  the following conditions on the ingress operator
+  - `DNSManaged` condition will be set to false and reason updated to
+    UnmanagedDNS.
+- The DNSRecord will also be updated as part of the same reconcile, 
+  `.spec.dnsManagementPolicy` will be set to `Unmanaged` and the following conditions
+  will be updated,
+  - `DNSReady` condition will be set to true and reason updated to
```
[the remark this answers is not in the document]
thejasn: Yeah, that makes more sense, I've changed it to `DNSReady=Unknown`.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,343 @@
...
+the cluster admin. 
+
+#### Modifying/Updating an ingresscontroller
+
+The ingresscontroller will need to be deleted and recreated with the updated domain
+present in if required domain is _not_ the same as cluster domain configured at
+*ingress.config.openshift.io/cluster* `.spec.domain`. If not needed, the workflow
+defined below is sufficient,
+- The new domain that needs to be associated with the ingresscontroller
+  must be created prior to making any changes.
```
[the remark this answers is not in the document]
thejasn: Yes, I've re-worded it to be more clear.

enhancements/ingress/configurable-dns-loadbalancerservice.md
thejasn: Thanks, Had missed this instance.

enhancements/ingress/configurable-dns-loadbalancerservice.md
```diff
@@ -0,0 +1,346 @@
...
+clearly state which dns record is unmanaged. If cluster admin were to delete
+the *DNSRecord* CR, the operator would recreate it (again reflecting and respecting the 
+IngressController's `spec.endpointPublishingStrategy.loadBalancer.dnsManagementPolicy` field).
+
+Once the *DNSRecord* is set to `Unmanaged`, any updates done to it will be ignored
+by the operator, but it is to be noted that if the ingresscontroller is updated from
+`Unmanaged` to `Managed` any changes done will be lost during re-sync.
+
+If the ingresscontroller is deleted when `.loadbalancer.dnsManagementPolicy` is set to
+`Unmanaged`, the *DNSRecord* CR will be deleted as per current implementation but the
```
Miciah, line 224:
```suggestion
`Unmanaged`, the *DNSRecord* CR will be deleted as per the current implementation, but the
```

enhancements/ingress/configurable-dns-loadbalancerservice.md
Miciah: Thanks, that makes sense; it just isn't clear from the test plan as written.

thejasn: The test plan might need to be revisited. I've mostly implemented this feature and turns out we can't have updates from `Unmanaged` to `Managed` since `domain` cannot be updated. Essentially, to update an ingresscontroller from `Managed` to `Unmanaged` is a DELETE and CREATE. 

Although it is possible to update from `Managed` to `Unmanaged` users would be forced to re-use the domain leading to again DELETE and CREATE of the ingresscontroller so that a new unmanaged domain can be given. 

So I'm thinking it is better to make `dnsManagementPolicy` also not updatable similar to `domain` (figured this would be better than making `domain` updatable`), then the test plan would be straight forward. Just test with an Unmanaged ingresscontroller.

Miciah: I don't understand.  Why would it matter that `domain` cannot be updated?  The operator should still be able to do an upsert when the management state is changed from `Unmanaged` to `Managed`, the same as it does an upsert if a managed DNSRecord's targets change.  On some platforms, an upsert may actually be implemented as discrete delete and create steps within a transaction, but that's an implementation detail that is abstracted behind the `dns.Provider` interface.

Miciah:
Thanks!
/approve
/lgtm

PR #1116, opened 2022-05-11
lint: only apply template lint rules to new files

dhellmann:
Ignore existing files during template lint. We don't expect authors to 
update to use the new template, so this change ignores existing files so we
don't have to manually override the job failure each time.

/assign @bparees
/cc @kikisdeliveryservice

### DIFF
--- a/hack/template-lint.sh
+++ b/hack/template-lint.sh
@@ -14,7 +14,12 @@ TEMPLATE=${SCRIPTDIR}/../guidelines/enhancement_template.md
 # We only look at the files that have changed in the current PR, to
 # avoid problems when the template is changed in a way that is
 # incompatible with existing documents.
-CHANGED_FILES=$(${SCRIPTDIR}/find_changed.sh)
+CHANGED_FILES=$(${SCRIPTDIR}/find_new.sh)
+
+if [ -z "$CHANGED_FILES" ]; then
+    echo "OK, no new enhancement files found"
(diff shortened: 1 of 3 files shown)


dhellmann:
I think Zane has needed to manually override a failure recently, too.

/cc @zaneb

hack/template-lint.sh
```diff
@@ -14,7 +14,12 @@ TEMPLATE=${SCRIPTDIR}/../guidelines/enhancement_template.md
 # We only look at the files that have changed in the current PR, to
 # avoid problems when the template is changed in a way that is
 # incompatible with existing documents.
-CHANGED_FILES=$(${SCRIPTDIR}/find_changed.sh)
+CHANGED_FILES=$(${SCRIPTDIR}/find_new.sh)
```
bparees, line 17:
```suggestion
NEW_FILES=$(${SCRIPTDIR}/find_new.sh)
```

hack/template-lint.sh
```diff
@@ -14,7 +14,12 @@ TEMPLATE=${SCRIPTDIR}/../guidelines/enhancement_template.md
...
+
+if [ -z "$CHANGED_FILES" ]; then
```
bparees, line 19:
```suggestion
if [ -z "$NEW_FILES" ]; then
```

bparees (commented):
naming nits

PR #1118, opened 2022-05-13
enhancements/machine-api/azure-ultra-disks: update api

damdo:
Update of the api in the enhancement after the changes to the api introduced by:
- [machine: azureprovider: AzureUltraSSDCapabilityState const rename #1145](https://github.com/openshift/api/pull/1145)
- [machine: azureprovider: data disk: add deletion policy #1149](https://github.com/openshift/api/pull/1149)


JoelSpeed (commented):
/approve

PR #1122, opened 2022-05-16
template: update instructions for milestones

dhellmann:
We don't track the status values in enhancements any more, so the
instructions about those don't make sense. Replace that step with
instructions to keep the doc up to date if the details change.

### DIFF
--- a/guidelines/enhancement_template.md
+++ b/guidelines/enhancement_template.md
@@ -29,16 +29,11 @@ To get started with this template:
    should desire this enhancement.
 1. **Create a PR.** Assign it to folks with expertise in that domain to help
    sponsor the process.
-1. **Merge at each milestone.** Merge when the design is able to transition to a
-   new status (provisional, implementable, implemented, etc.). View anything
-   marked as `provisional` as an idea worth exploring in the future, but not
-   accepted as ready to execute. Anything marked as `implementable` should
-   clearly communicate how an enhancement is coded up and delivered. If an
-   enhancement describes a new deployment topology or platform, include a
-   logical description for the deployment, and how it handles the unique aspects
-   of the platform. Aim for single topic PRs to keep discussions focused. If you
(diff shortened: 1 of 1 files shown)


guidelines/enhancement_template.md
```diff
@@ -29,16 +29,9 @@ To get started with this template:
...
-   marked as `provisional` as an idea worth exploring in the future, but not
-   accepted as ready to execute. Anything marked as `implementable` should
-   clearly communicate how an enhancement is coded up and delivered. If an
-   enhancement describes a new deployment topology or platform, include a
-   logical description for the deployment, and how it handles the unique aspects
-   of the platform. Aim for single topic PRs to keep discussions focused. If you
-   disagree with what is already in a document, open a new PR with suggested
-   changes.
+1. **Merge at each milestone.** Merge when the design is able to be
+   implemented. Come back and update the document if important details
```
bparees, line 33: does "able to be implemented" mean:

a) the design is complete, all questions are answered, we know how to build it, and we got agreement from appropriate stakeholders that we want to build it and this is how we want to build it?

or does it mean (or also mean):

b) we are ready to start coding, we have resources assigned/item is prioritized?

fundamentally i think instead of saying "able to be implemented" we should be more explicit about when we expect EPs to get merged.

kikisdeliveryservice: Isn't it 1, given that we don't want (or allow??) PRs to stay open for a long time. But also we have asked that enhancements are for things that mostly fall into 2 ie you intend on working on it in the near future so avoid PRs being open indefinitely.

dhellmann: As Kirsten points out, the instructions in the readme say not to write an enhancement for something we aren't going to start working on "soon". I assume if that's the case we have the resources we need in place, and that group of people are the ones reviewing the enhancement.

I've updated this step further to talk about merging when there is consensus that it's ready. Let me know what you both think of the new draft.

PR #1123, opened 2022-05-17
CONSOLE-3159: API changes for i18n

jhadvig:
Story: https://issues.redhat.com/browse/CONSOLE-3159

/assign @spadgett @vojtechszocs @florkbr

### DIFF
--- a/enhancements/console/dynamic-plugins.md
+++ b/enhancements/console/dynamic-plugins.md
@@ -433,6 +433,35 @@ const VMHeading = () => {
 };
 ```
 
+In the `v1` API version, an additional field is introduced for enabling and loading i18n namespace that
+given dynamic plugin contains. The i18n namespace name follows `plugin__{plugin-name}` naming convention.
+If i18n is not enabled in the `ConsolePlugin` resource, no i18n namespaces will be loaded for the plugin.
+
+```yaml
+apiVersion: console.openshift.io/v1
+kind: ConsolePlugin
+metadata:
+  name: acm
+spec:
+  displayName: 'Advanced Cluster Management'
+  i18n:
+    enabled: true
+```
+
+In the 4.11 release, a `console.openshift.io/default-i18next-namespace` annotation
+is being introduced. The annotation indicates whether the `ConsolePlugin` contains
+localization resources. If the annotation is set to `"true"`, the localization
+resources from the i18n namespace named after the dynamic plugin (e.g. `plugin__kubevirt`),
+are loaded. If the annotation is set to any other value or is missing on the `ConsolePlugin`
+resource, localization resources are not loaded.
+Note: All API versions of `ConsolePlugin` are checked for this annotation. The annotation
+should only be used by plugins that need to work with 4.10 and 4.11. This annotation will
+be removed when 4.10 goes out of support in favor of the proper API.
+
+Prior to 4.11 release, localization resources are being loaded by default. In case these
+resources are not present in the dynamic plugin, the initial console load will be slowed 
+down. For more info check [BZ#2015654](https://bugzilla.redhat.com/show_bug.cgi?id=2015654)
+
 ### Error Handling
 
 Console will guard against runtime errors in plugins. All plugin components
@@ -487,9 +516,10 @@ dynamic.
 
 ### Graduation Criteria
 
-Initially, this will be an internal API. We will reevaluate as we receive
-implementation feedback from plugin teams and as the dynamic plugin model
-matures.
+| OpenShift | Maturity     | API version   |
+|-|-|-|
+| 4.8       | Alpha        | v1alpha1      |
+| 4.11      | GA (planned) | v1            |
 
 Static plugins are already a supported feature. Any existing static plugin that
 is migrated to a dynamic plugin will need to have the same support level.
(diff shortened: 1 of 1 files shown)


enhancements/console/dynamic-plugins.md
```diff
@@ -433,6 +433,33 @@ const VMHeading = () => {
...
+```yaml
+apiVersion: console.openshift.io/v1alpha1
+kind: ConsolePlugin
+metadata:
+  name: acm
+spec:
+  displayName: 'Advanced Cluster Management'
+  i18n:
+    namespaces:
+      - plugin__acm
```
spadgett, line 460: We'll need to decide if (1) we want to support multiple namespaces and (2) if we should ensure that namespace names are unique (i.e. by including the plugin name). It would be good to get stakeholder feedback and discuss at the next sync.

cc @christianvogt @bipuladh

enhancements/console/dynamic-plugins.md
```diff
@@ -433,6 +433,33 @@ const VMHeading = () => {
 };
 ```
 
+For a 4.11 release a `console.openshift.io/default-i18next-namespace` annotation,
```
spadgett, line 436: I'd make it clear that this annotation should only be used by plugins that need to work with 4.10 and 4.11. We'll eventually drop this annotation when 4.10 goes out of support in favor of the proper API.

enhancements/console/dynamic-plugins.md
christianvogt: An argument for multiple namespaces is that a shared library is being used and manages it's own namespace, then the host plugin must register the shared library namespace with it's plugin because the console must relay the call to the correct service hosting the locales for that shared library.

This however has potential to clash with other plugins that may be using the same library but different versions.

How can the console ensure that the shared library localization requests go to the correct service if multiple plugins use the same shared library with the same namespace?

The onus is also then on a plugin to prevent clashing with other plugins. This is problematic. The uniqueness of plugin is in its name and nothing else.

enhancements/console/dynamic-plugins.md
```diff
@@ -433,6 +433,34 @@ const VMHeading = () => {
 };
 ```
 
+In the 4.11 release a `console.openshift.io/default-i18next-namespace` annotation,
```
spadgett, line 436: I would move this below discussion of the API property we're introducing since it's not the recommended way and only here for backwards compatibility.

enhancements/console/dynamic-plugins.md
```diff
@@ -433,6 +433,34 @@ const VMHeading = () => {
...
+is being introduced. It annotation will indicate whether the dynamic plugin containes 
+localization resources, or not. If the annotation value is set to `true`, the localization
+resources, from the namespace named after the dynamic plugin (eg. `plugin__kubevirt`),
+are loaded. If the annotation is set to `false`, or it is missing on the dynamic plugin CR,
+the localization resources are not loaded.
```
spadgett, line 441:
```suggestion
In the 4.11 release, a `console.openshift.io/default-i18next-namespace` annotation
is being introduced. The annotation indicates whether the `ConsolePlugin` contains
localization resources. If the annotation is set to `"true"`, the localization
resources from the i18n namespace named after the dynamic plugin (e.g. `plugin__kubevirt`),
are loaded. If the annotation is set to any other value or is missing on the `ConsolePlugin` resource,
localization resources are not loaded.
```

enhancements/console/dynamic-plugins.md
```diff
@@ -433,6 +433,34 @@ const VMHeading = () => {
...
+Note: All API versions of dynamic plugins CRD are checked for this annotation. This annotation
+should only be used by plugins that need to work with 4.10 and 4.11. This annotation will
+be eventually dropped this annotation when 4.10 goes out of support in favor of the proper API.
```
spadgett, line 444:
```suggestion
Note: All API versions of `ConsolePlugin` are checked for this annotation. The annotation
should only be used by plugins that need to work with 4.10 and 4.11. This annotation will
be removed when 4.10 goes out of support in favor of the proper API.
```

enhancements/console/dynamic-plugins.md
```diff
@@ -433,6 +433,34 @@ const VMHeading = () => {
...
+
+Prior to 4.11 release, localization resources are being loaded by default. In case these
+resources are not present in the dynamic plugin, the initial console load will be slowed 
+down. For more info check [BZ#2015654](https://bugzilla.redhat.com/show_bug.cgi?id=2015654)
+
+For `v1` API version, an additional field is introduced, for enabling and loading i18n namespace that
+given dynamic plugin contains. The i18n namespace name follows `plugin__{plugin-name}` schema.
```
spadgett, line 451:
```suggestion
In the `v1` API version, an additional field is introduced for enabling and loading i18n namespace that
given dynamic plugin contains. The i18n namespace name follows `plugin__{plugin-name}` naming convention. If i18n is not enabled in the `ConsolePlugin` resource, no i18n namespaces will be loaded for the plugin.
```

spadgett (approved):
/lgtm
/hold
to give others a chance to comment

enhancements/console/dynamic-plugins.md
```diff
@@ -433,6 +433,35 @@ const VMHeading = () => {
...
+metadata:
+  name: acm
+spec:
+  displayName: 'Advanced Cluster Management'
+  i18n:
+    enabled: true
+```
+
+In the 4.11 release, a `console.openshift.io/default-i18next-namespace` annotation
+is being introduced. The annotation indicates whether the `ConsolePlugin` contains
```
bipuladh, line 452: Where is this annotation being introduced? Should plugins add it to the CSV? Maybe we should mention that.

spadgett: It is added to the `ConsolePlugin` resource.

jhadvig: So the annotation will be checked for DynamicPlugin CR.

PR #1125, opened 2022-05-19
AGENT-40: Enhancement for generating admin kubeconfig

zaneb:
We will generate the admin kubeconfig file in the agent installer at the
same time as the agent ISO. The CA keys used to sign the client
certificate in the kubeconfig will be added to the ISO and eventually
installed into the cluster so that the credentials in the admin
kubeconfig will be accepted.


enhancements/agent-installer/automated-workflow-kubeconfig.md
```diff
@@ -0,0 +1,253 @@
...
+  command to wait for the cluster to complete, with the preceding steps in the
+  automated workflow limited to only running the `agent create image` command
+  and booting the ISO on the hosts.
+
+### Goals
+
+* Generate a working admin kubeconfig as part of the `openshift-install agent
+  create image` command.
+
+### Non-Goals
```
rwsu, line 72: Should making the kubeadmin password available be a non-goal?

dhellmann: Why wouldn't we include that password, too?

enhancements/agent-installer/automated-workflow-kubeconfig.md
```diff
@@ -0,0 +1,253 @@
...
+cluster-manifests, rather than the install-config, because the install-config
+is optional in the agent-based installer flow.
+
+The actual ignition files used for bootstrapping the cluster are not created by
+the agent-based installer binary, but rather by the assisted-service running
+openshift-installer again. In order to ensure that the cluster that is
+installed accepts the credentials in the kubeconfig, we must supply all assets
+that match the following rules:
+
+If we define an 'implicated' asset as one that:
```
dhellmann, line 90: I'm having trouble following this section. I think I understand what you're trying to convey, but the logic might be easier to follow if it started with the why instead of the rules.

Why is it important that some assets cannot be recreated?

Why are the relationships and interdependencies between assets relevant?

enhancements/agent-installer/automated-workflow-kubeconfig.md
```diff
@@ -0,0 +1,253 @@
...
+
+* Contains some additional data that cannot be recreated from its dependencies alone (unlike, e.g. the CABundle assets); and
+* Can be written to and loaded from disk; and
```
dhellmann, line 93: Are there any assets that cannot be recreated and also cannot be written to and loaded from disk?

enhancements/agent-installer/automated-workflow-kubeconfig.md
```diff
@@ -0,0 +1,253 @@
...
+* Either:
+  * Is depended on (directly or indirectly) by the implicating asset; or
```
dhellmann, line 95: "Is depended on" is a bit awkward. Does this just mean that we need to include the dependencies of the other assets that we're treating the same way?

enhancements/agent-installer/automated-workflow-kubeconfig.md
```diff
@@ -0,0 +1,253 @@
...
+
+Note that the modifications to the installer to read certificates from disk
+need to occur in the installer present in the payload of the release being
+installed, i.e. on the `master` branch, not the `agent-installer` branch where
+the agent-based installer is being developed.
+
+The work directory for the installer is a tmpdir created by assisted-service on
+the fly with a randomised name. Therefore it is not possible to pre-populate
+the certificates into the directory, and assisted-service must be involved in
+copying the files.
```
dhellmann, line 164: Is that going to be the case for all manifests? Could we change the service to use either a well-defined location or a random one that both the service and agent can know?

enhancements/agent-installer/automated-workflow-kubeconfig.md
```diff
@@ -0,0 +1,253 @@
...
+* `tls.KubeAPIServerServiceNetworkSignerCertKey`
+* `tls.AdminKubeConfigSignerCertKey`
+* `tls.AdminKubeConfigClientCertKey`
+
+When generating the agent ISO, the certificate and key files for those same
+assets will be built in to the ISO.
+
+The assisted-service will copy the certificates from a location provided by the
+environment variable `ASSISTED_SERVICE_CLUSTER_TLS_CERTS_DIR` to the
+installer's work directory before generating the manifests/ignition.
```
dhellmann, line 130: Do we only want this bypass approach to apply to the auth-related manifests, rather than making it a generic way to use other manifests without passing them through the API?

enhancements/agent-installer/automated-workflow-kubeconfig.md
zaneb: As someone who never uses the console, TBH I'm not sure how this works in a _normal_ installation. It appears that in IPI/UPI it is written to the file `auth/kubeadmin-password`? What, if anything, does assisted do? Presumably it's easy enough to set a new password if you have the admin kubeconfig.

enhancements/agent-installer/automated-workflow-kubeconfig.md
zaneb: > I'm having trouble following this section.

Me too, and I wrote it :frowning_face: 

The idea was to explain the methodology used to generate the graphs below.

> I think I understand what you're trying to convey, but the logic might be easier to follow if it started with the why instead of the rules.
> 
> Why is it important that some assets cannot be recreated?
> 
> Why are the relationships and interdependencies between assets relevant?

e.g. kubeconfig.AdminClient depends indirectly on tls.AdminKubeConfigSignerCertKey. That's not readable from disk, so to supply our own AdminClient asset we'd also have to modify all of the other assets that depend (indirectly) on AdminKubeConfigSignerCertKey - which in this case means bootstrap.Bootstrap - or also supply all of _it's_ dependencies.

enhancements/agent-installer/automated-workflow-kubeconfig.md
zaneb: Yes, in fact all of the interesting ones cannot be loaded from disk and that's exactly the point where this enhancement proposes to intervene.

enhancements/agent-installer/automated-workflow-kubeconfig.md
zaneb: Yes, this is the easy part, before it gets recursive in the next bullet point.

enhancements/agent-installer/automated-workflow-kubeconfig.md
dhellmann: And we don't want the auth-related manifests to go through that same API?

I'm not arguing one way or the other, but it would be good to have that laid out explicitly with the reasoning.

zaneb: To be clear, the tls assets are _not_ manifests. They are not kubectl applied to the cluster, and they don't live in the `manifests` directory. They're just raw certificate and private key files.

zaneb:
> The important thing is that it's OK with the installer team and that they will support it.

They already merged it :smile: openshift/installer#5872

enhancements/agent-installer/automated-workflow-kubeconfig.md
dhellmann: OK, thanks for clarifying that.

enhancements/agent-installer/automated-workflow-kubeconfig.md
avishayt: It is available via REST API and UI just like the kubeconfig, and when running as an operator it is stored in a secret.

romfreiman: hmm how exposing a kubeadmin password can be not a goal? We want to provide CUs friendly experience, rather than adding more steps for them to perform the installation. 
And yes, I assume most of the customers will go to console.

enhancements/agent-installer/automated-workflow-kubeconfig.md
```diff
@@ -0,0 +1,253 @@
...
+N/A
+
+## Implementation History
+
+The changes to the installer to read the certificates/keys from disk were
+implemented in
+[openshift/installer#5872](https://github.com/openshift/installer/pull/5872)
+and are available in OpenShift 4.11 nightlies from Feature Freeze on.
+
+## Alternatives
```
mhrivnak, line 243: Just one more idea to toss in here:

We could let the CA cert, kubeconfig, and kubeadmin password all still be generated the usual way. Then in the new cluster, run a temporary web service just for making those available. It could require a pre-defined token that is generated by the installer, which might be easier to add to the cluster. The service itself might be run based on including an extra manifest.

There are various strategies for ensuring the service stops after the hand-off is complete, so the service doesn't run forever. For example:

- user accesses a known URL served by the new cluster and presents their token. This could be via a UI or just an API response.
- The access service responds with a limited kubeconfig that can only access a specific namespace and with limited capability.
- The user creates an empty secret with a known name in that namespace.
- The access service sees the secret, shuts down its http endpoint, populates the secret with the cluster-admin kubeconfig and kubeadmin password, and exits.
- The user reads the secret.

Such an exchange can be easily automated. This may or may not be an easier approach, but it seems worth at least considering and listing as an alternative.

enhancements/agent-installer/automated-workflow-kubeconfig.md
```diff
@@ -0,0 +1,253 @@
...
+need to occur in the installer present in the payload of the release being
+installed, i.e. on the `master` branch, not the `agent-installer` branch where
+the agent-based installer is being developed.
+
+The work directory for the installer is a tmpdir created by assisted-service on
+the fly with a randomised name. Therefore it is not possible to pre-populate
+the certificates into the directory, and assisted-service must be involved in
+copying the files.
+
+### Risks and Mitigations
```
mhrivnak, line 166: If the secrets are included in the ISO, then they're only as safe as the ISO is, which can be difficult to guarantee. If a user wants to pxe-boot or similar, it's likely they'll have to serve the ISO unauthenticated on their local network.

dhellmann (commented):
/approve
/hold

I think this is ready to merge, but want to leave a little more time for the reviewers to give a final sign off or propose wording updates. Since I'll be out next week, I'll add the approved label now and place a hold for @zaneb or @celebdor to remove Tuesday or Wednesday.

enhancements/agent-installer/automated-workflow-kubeconfig.md
```diff
@@ -0,0 +1,270 @@
...
+similar problems, although it could work at any time. It could also result in
+nodes being tainted due to the ssh access.
+
+A service running in the cluster could provide access to the admin kubeconfig
+suing some sort of ad hoc authentication, not dependent on the admin kubeconfig
+signing key. It could perhaps also be removed when no longer required. This
+would not be consistent with the other installer workflows, and could require
+complex work on the part of the user when the cluster is started in an
+environment not connected to the one where the ISO was generated (so the
+installer might not be available to provide automation).
```
mhrivnak, line 270: It would probably make more sense as an `oc` feature.

enhancements/agent-installer/automated-workflow-kubeconfig.md
```diff
@@ -0,0 +1,270 @@
...
+Agent installer->>Certificate assets: Generate()
+Agent installer->>Kubeconfig asset: Generate()
+Certificate assets-->>Kubeconfig asset: certificates
+Kubeconfig asset-->>Cluster creator: kubeconfig
+Agent installer->>Agent ISO: Generate()
+Certificate assets-->>Agent ISO: certificates
+Agent ISO-->>Cluster creator: agent.iso
+Agent ISO-->>Assisted Service: certificates
+Assisted Service-->>OpenShift Installer: certificates
+OpenShift Installer-->>Assisted Service: kubeconfig
```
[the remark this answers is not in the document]
zaneb: It's the one that can be downloaded from the assisted API, but there's no practical difference.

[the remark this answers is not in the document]
dhellmann: OK, thanks

zaneb:
I've addressed all of the comments, so I'll remove the hold and the next LGTM wins it.
/hold cancel

rwsu (commented):
/lgtm

PR #1126, opened 2022-05-19
cluster-logging: Logs-based alerts

periklis:
### Summary
This document proposes a solution and a delivery plan for pushing logs-based alerts from OpenShift Logging Loki to Cluster-Monitoring AlertManager.

cc @simonpasquier @jan--f @alanconway @jcantrill

### DIFF
--- a/enhancements/cluster-logging/logs-based-alerts.md
+++ b/enhancements/cluster-logging/logs-based-alerts.md
@@ -1,0 +1,300 @@
+---
+title: logs-based-alerts
+authors:
+  - "@periklis"
+reviewers:
+  - "@alanconway"
+  - "@xperimental"
+  - "@jcantrill"
+  - "@simonpasquier"
+  - "@fpetkovski"
+approvers:
+  - "@alanconway"
+  - "@simonpasquier"
+  - "@fpetkovski"
+api-approvers:
+  - "@alanconway"
+creation-date: "2022-05-16"
+last-updated: "2022-05-19"
+tracking-link:
+  - https://issues.redhat.com/browse/LOG-2510
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Logs-based Alerts
+
+## Summary
+
+This document proposes a solution and a delivery plan for pushing logs-based alerts from OpenShift Logging Loki to Cluster-Monitoring AlertManager. It complements the existing [user-workload monitoring stack][uwm-docs] and [alerting stack][alerting-docs], enabling a full self-service experience for platform and workload logs-based monitoring.
+
+## Motivation
+
+Since OpenShift 4.6, application owners can configure alerting rules based on metrics themselves as described in [User Workload Monitoring][user-workload-monitoring-enhancement] (UWM)
+enhancement. The rules are defined as `PrometheusRule` resources and can be based on platform and/or application metrics.
+
+To expand the alerting capabilities on logs as an observability signal, cluster admins and application owners should be able to configure alerting rules as described in the [Loki Rules][loki-rules-docs] docs and in the [Loki Operator Ruler][loki-operator-ruler-support] upstream enhancement.
+
+[AlertingRule][alertingrule-crd] CRD fullfills the requirement to define alerting rules for Loki similar to `PrometheusRule`.
+
+[RulerConfig][rulerconfig-crd] CRD fullfills the requirement to connect the Loki Ruler component to notify a list of Prometheus AlertManager hosts on firing alerts.
+
+### User Stories
+
+* As an application owner, I want to use AlertingRule custom resources to notify AlertManager on firing alerts from my workload logs.
+* As a cluster admin, I want to use AlertingRule custom resources to notify AlertManager on firing alerts from the platform logs.
+* As a cluster admin, I want to use AlertingRule custom resources to notify AlertManager on firing alerts from the audit logs.
+
+### Goals
+
+* Application Owners can configure alert notifications for application alerts based on workload logs.
+* Cluster admins can configure alert notifications for platform alerts based on platform logs.
+* Cluster admins can configure alert notifications for platform alerts based on audit logs.
+
+### Non-Goals
+
+* Specific alerting support on the OCP console for Logs-based alerts. This should be seamless in the existing alerting UIs.
+* Recording rules to compile metrics from logs.
+
+## Proposal
+
+We plan to leverage the `AlertingRule` and `RulerConfig` custom resource definitions already exposed by the Loki Operator so that:
+- cluster admins can configure how the platform logs can notify on alertworthy platform and/or audit logs.
+- application owners can configure how the workload logs can notify on alertworthy logs.
+
+### Workflow Description
+
+#### Logs-based Alerts for Application Owners
+
+1. The application owner creates an AlertingRule custom resource on a workload project, e.g.
+
+```yaml
+apiVersion: loki.grafana.com/v1beta1
+kind: AlertingRule
+metadata:
+  name: my-workload-alerts
+  namespace: my-workload-ns
+  labels:
+    openshift.io/cluster-monitoring: "true"
+spec:
+  tenantID: application
+  groups:
+    - name: MyApplication
+      rules:
+        - alert: MyApplicationHighPercentageError
+          expr: |
+            sum(rate({kubernetes_namespace_name="my-workload-ns", kubernetes_pod_name=~"my-workload.*"} |= "error" [1m])) by (job)
+              /
+            sum(rate({kubernetes_namespace_name="my-workload-ns", kuberentes_pod_name=~"my-workload.*"}[1m])) by (job)
+              > 100
+          for: 5m
+          labels:
+            severity: warning
+          annotations:
+            summary: My Application High Errors.
+            description: My application throws a high amount of error logs.
+```
+
+2. The application owner can only define AlertingRules custom resources in project with granted access to them and only for the tenant `application`.
+3. The application owner can inspect firing alerts from the above rule in Developer > Observe > Alerts.
+
+#### Logs-based Alerts for Cluster Admins
+
+1. The cluster admin creates an AlertingRule custom resource on a platform project, e.g.
+
+```yaml
+apiVersion: loki.grafana.com/v1beta1
+kind: AlertingRule
+metadata:
+  name: my-operator-alerts
+  namespace: openshift-operators-redhat
+  labels:
+    openshift.io/cluster-monitoring: "true"
+spec:
+  tenantID: infrastructure
+  groups:
+    - name: MyOperator
+      rules:
+        - alert: MyOperatorHighPercentageError
+          expr: |
+            sum(rate({kubernetes_namespace_name="openshift-operators-redhat", kubernetes_pod_name=~"my-operator.*"} |= "error" [1m])) by (job)
+              /
+            sum(rate({kubernetes_namespace_name="openshift-operators-redhat", kuberentes_pod_name=~"my-operator.*"}[1m])) by (job)
+              > 100
+          for: 5m
+          labels:
+            severity: warning
+          annotations:
+            summary: My Operator High Errors.
+            description: My operator throws a high amount of error logs.
+```
+
+2. The cluster admin can define AlertingRule custom resources all projects for tenants `application`, `infrastructure` and `audit`
+3. The cluster admin can inspect firing alerts from the above rule in Administrator > Observe > Alerting > Alerts.
+
+### API Extensions
+
+As mentioned above this document proposes to leverage two existing CRDs namely `AlertingRule` and `RulerConfig` provided by the Loki Operator.
+
+### Implementation Details/Notes/Constraints
+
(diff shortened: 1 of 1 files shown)


enhancements/cluster-logging/logs-based-alerts.md
```diff
@@ -0,0 +1,300 @@
...
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Logs-based Alerts
+
+## Summary
+
+This document proposes a solution and a delivery plan for pushing logs-based alerts from OpenShift Logging Loki to Cluster-Monitoring AlertManager. It complements the existing [user-workload monitoring stack][uwm-docs] and [alerting stack][alerting-docs], enabling a full self-service experience for platform and workload logs-based monitoring.
```
simonpasquier, line 30: (nit) s/AlertManager/Alertmanager/g :)

enhancements/cluster-logging/logs-based-alerts.md
```diff
@@ -0,0 +1,300 @@
...
+              > 100
+          for: 5m
+          labels:
+            severity: warning
+          annotations:
+            summary: My Application High Errors.
+            description: My application throws a high amount of error logs.
+```
+
+2. The application owner can only define AlertingRules custom resources in project with granted access to them and only for the tenant `application`.
```
simonpasquier, line 99: > only for the tenant `application`.

will this be enforced?

enhancements/cluster-logging/logs-based-alerts.md
```diff
@@ -0,0 +1,300 @@
...
+
+### Goals
+
+* Application Owners can configure alert notifications for application alerts based on workload logs.
+* Cluster admins can configure alert notifications for platform alerts based on platform logs.
+* Cluster admins can configure alert notifications for platform alerts based on audit logs.
+
+### Non-Goals
+
+* Specific alerting support on the OCP console for Logs-based alerts. This should be seamless in the existing alerting UIs.
```
simonpasquier, line 57: I would need to understand more about the design but the OCP console gets alerting/recording rules as well as alerts via the Thanos Querier service. Only silences are pulled directly from the Alertmanager API service.
I'm not sure how/if the Thanos querier can know about the Loki rulers.

enhancements/cluster-logging/logs-based-alerts.md
```diff
@@ -0,0 +1,300 @@
...
+```
+
+2. The cluster admin can define AlertingRule custom resources all projects for tenants `application`, `infrastructure` and `audit`
+3. The cluster admin can inspect firing alerts from the above rule in Administrator > Observe > Alerting > Alerts.
+
+### API Extensions
+
+As mentioned above this document proposes to leverage two existing CRDs namely `AlertingRule` and `RulerConfig` provided by the Loki Operator.
+
+### Implementation Details/Notes/Constraints
```
simonpasquier, line 140: Until now we've treated the Alertmanager deployment as an implementation detail and didn't make any promise on the stability (e.g. we could change the authentication method or require different permissions if necessary). If the Loki Ruler becomes another source of alerts, we would need to be more careful (not saying it's bad but this needs to be taken into account on the monitoring side). Something that comes to my mind for instance: in 4.11 we're adding the possibility to deploy a standalone Alertmanager for user-defined alerts so the Alertmanager destination for alerts could be the Alertmanager service in openshift-monitoring namespace or the Alertmanager service in the openshift-user-workload-monitoring namespace.

enhancements/cluster-logging/logs-based-alerts.md
```diff
@@ -0,0 +1,300 @@
...
+1. The application owner creates an AlertingRule custom resource on a workload project, e.g.
+
+```yaml
+apiVersion: loki.grafana.com/v1beta1
+kind: AlertingRule
+metadata:
+  name: my-workload-alerts
+  namespace: my-workload-ns
+  labels:
+    openshift.io/cluster-monitoring: "true"
```
simonpasquier, line 79: I expect that we don't need this label?

enhancements/cluster-logging/logs-based-alerts.md
periklis: I have not much experience with AdmissionWebhooks, but I would to think about this like:
- Application owners: `oc create -n my-workload-ns alertingrule my-workload-alerts` where `spec.tenantID: application` succeeds
- Only members of `cluster-admins` can create with `spec.tenantID: infrastructure` and/or `spec.tenantID: audit`.

WDYT?

enhancements/cluster-logging/logs-based-alerts.md
periklis: This is also a bit I am still investigating (See Open Questions section). AFAIK the Loki ruler exposes a Prometheus-compatible API (See [grafana/loki](https://github.com/grafana/loki/blob/main/pkg/loki/modules.go#L774-L775)). This looks like to declare compatibility from AlertManager calling the ruler or other tools like `cortextool`. 

However, I bet we will not succeed thanos-querier calling Loki-Ruler if it comes into evaluating the underlying expressions anyhow, because thanos does not understand LogQL. (cc @bwplotka WDYT???)

enhancements/cluster-logging/logs-based-alerts.md
periklis: Yeah exactly this is main reason why I want to raise this as an enhancement proposal first. Placing the Loki Ruler as extra notifier will basically raise the levels on an OCP in-cluster component like Alertmanager in terms of availability. To push this a little bit further, we will have the same issue with Prometheus when Loki's recording rules will use the remote-write-client to push metrics to cluster-monitoring's/user-workload-monitoring's prometheus. But let's frame this for AlertManager first.

My questions to you folks are:
- What does it take to make raise availability of alertmanager as a component that can be used from components like Loki ruler that are not part of OCP but belong to a layered product like OpenShift Logging?
- How can we detect if users have one Alertmanager (cluster-monitoring) or more Alertmanagers (cluster-monitoring + user-workload-monitoring)?

enhancements/cluster-logging/logs-based-alerts.md
periklis: The Loki-Operators follows the same pattern as the Prometheus-Operator for detecting AlertingRules/RecordingRules based on selectors and namespaceSelectors. I would say keeping the same labels to detect Loki AlertingRules/RecordingRules in OpenShift Logging as you do with PrometheusRules in OpenShift Monitoring, would make it easier for users for both observability pieces. WDYT?

enhancements/cluster-logging/logs-based-alerts.md
simonpasquier: This seems feasible but it probably deserves input from the API folks. FWIW we run an admission webhook for PrometheusRules: this isn't trivial but still doable (think about proper HA for instance).

alanconway (commented):
Aside from the flow-control question this looks good to me.

enhancements/cluster-logging/logs-based-alerts.md
```diff
@@ -0,0 +1,300 @@
...
+[AlertingRule][alertingrule-crd] CRD fullfills the requirement to define alerting rules for Loki similar to `PrometheusRule`.
+
+[RulerConfig][rulerconfig-crd] CRD fullfills the requirement to connect the Loki Ruler component to notify a list of Prometheus AlertManager hosts on firing alerts.
+
+### User Stories
+
+* As an application owner, I want to use AlertingRule custom resources to notify AlertManager on firing alerts from my workload logs.
+* As a cluster admin, I want to use AlertingRule custom resources to notify AlertManager on firing alerts from the platform logs.
+* As a cluster admin, I want to use AlertingRule custom resources to notify AlertManager on firing alerts from the audit logs.
+
```
eranra, line 48: @periklis The area of consolidated observability is very important forward and is one of the main motivations for this enhancement. Can you add a user story that talks on a consolidated alert that encapsulates information from logs and metrics? THX :-)

enhancements/cluster-logging/logs-based-alerts.md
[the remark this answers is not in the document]
simonpasquier: There's no kind of flow control in Alertmanager as of today. I've checked the Loki code base quickly and it seems to use the same techniques as Prometheus to avoid hammering Alertmanager:
* It has a default delay of 1 minute before resending the same alert to Alertmanager.
* It keeps alerts in an in-memory queue (up to 10,000 alerts). If Alertmanager can't keep up on the receiving side, older alerts are dropped.

I would expect that Loki is impacted as well when we hit this "too many alerts" issue.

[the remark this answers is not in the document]
periklis: Literally Loki replicates 1:1 the Prometheus code if it is not using it directly :P

enhancements/cluster-logging/logs-based-alerts.md
periklis: @eranra This is a delicate but important aspect. To reach this state we need a separate enhancement proposal to connect the Loki Ruler to send log-based metrics to Prometheus via remote-write. I have this on my plan after we sort the easier part here namely the Loki-Ruler integration with AlertManager.

enhancements/cluster-logging/logs-based-alerts.md
jewzaam, line 30: `plan for pushing logs-based alerts from OpenShift Logging Loki to Cluster-Monitoring AlertManager`

RulerConfig appears to handle this, routing to arbitrary AM instances.  But the other part of this enhancement is not about routing to AM, but presenting alerts via rules (AlertingRule CR).  Is this going to create a metric that can be scraped?  I ask because I want to make sure there isn't an assumption that AM is the only way to get log based alerts.  Specifically in the context of Managed OpenShift I have a few examples for why having AlertingRule create a metric akin to ALERTS is useful.

ROSA/OSD support log forwarding as an addon at this time.  There are no plans to SRE any full logging stack as a part of the platform.  Cluster monitoring is only used by SRE, customers cannot modify prometheus or AM configuration in openshift-monitoring.  Making sure customers can control and manage log based alerts includes flexibility in how the alerts are consumed.  And ARO only consumes ALERTS metrics from cluster prometheus.  ARO does not configure AM at all and never will due to use of internal MSFT systems as a first party offering in Azure.  If ARO were to ever consume logging metrics for the platform (not planned, just an example) we need the ALERTS metric, not AM routing.

periklis: @jewzaam Do I understand correctly that specific SD use cases for alerting require that log-based queries to be transformed into metrics that get persisted into the cluster's Prometheus first?

periklis: @jewzaam Scraping an alert metric for Prometheus is not going to work as per Loki executing LogQL expressions. However what we can do is to execute the LogQL query in Loki as a Recording rule and push the resulting metric to Prometheus via Remote Write. Basically Remote Write is the other half of the `RulerConfig` CRD and I aim to publish this as a separate document.

__ Note:__ The reason I started writing the Loki-to-AlertManager design document first is that it is from in-cluster logging perspective a low-hanging fruit. OTH Remote-Write into ClusterMonitoring/UserWorkloadMonitoring Prometheus requires a more robust design from reliability standpoint. I say this as per now Prometheus in either `openshift-monitoring` and `openshift-user-workload-monitoring` namespaces on any cluster is a closed black-box. Metrics can be exported via federation or Remote-Write to other services (in-/out-cluster) but you cannot write metrics into any of these Prometheus instances. For good reasons `openshift-monitoring` is closed for clients as it represents a first-line of defense operating the cluster itself, thus having Loki as writing client could overload it (e.g. high cardinality metrics from noisy/bad-crafted log queries).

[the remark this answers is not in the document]
periklis: Yes exactly. Using the `LokiStack` CR

enhancements/cluster-logging/logs-based-alerts.md
```diff
@@ -0,0 +1,300 @@
...
+### Non-Goals
+
+* Specific alerting support on the OCP console for Logs-based alerts. This should be seamless in the existing alerting UIs.
+* Recording rules to compile metrics from logs.
+
+## Proposal
+
+We plan to leverage the `AlertingRule` and `RulerConfig` custom resource definitions already exposed by the Loki Operator so that:
+- cluster admins can configure how the platform logs can notify on alertworthy platform and/or audit logs.
```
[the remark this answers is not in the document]
periklis: I would say rather not, as hosting user-related AlertingRule and RecordingRule upstreams is hard to maintain and get it right for everybody. If on the other side you mean any known upstream kubernetes alerts, I would say provide me the links and I can re-consider.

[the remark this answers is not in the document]
dofinn: i dont know any of yet - would be interesting to understand if any OCP engineering has requirements and/or test cases using loki + alerts

periklis:
> /approve /lgtm /hold I think this is merge-worthy, the /hold is in case @periklis wants to make any further changes based on conversations above. @periklis feel free to /unhold when ready.

Thanks for labeling this properly. I will not merge this as it still needs a spike to inform how to show up the alerts in the OpenShift Console at the end. The current state describes on how the alerts land per tenant into Cluster Monitoring's Alertmanager but they cannot be picked up by the UI. The obstacle here is that OpenShift Console is not calling AlertManager directly but Thanos Querier to assure tenancy and access control for alerts between Administrators and Developers.

PR #1127, opened 2022-05-19
Proposal for hiding container mountpoints from systemd

lack:
After some time, this proposal is ready to see the light of day again.

We have been running the proof-of-concept with various telco partners and there is widespread acceptance of the approach and desire for the benefits, especially on resource-constrained systems like Single-Node OpenShift.

This incorporates all discussion from the previous version of this proposal (#593), and cleans up some specific implantation plan and test plan details.

Signed-off-by: Jim Ramsay <jramsay@redhat.com>

### DIFF
--- a/enhancements/hide-container-mountpoints.md
+++ b/enhancements/hide-container-mountpoints.md
@@ -1,0 +1,343 @@
+---
+title: hide-container-mountpoints
+authors:
+  - "@lack"
+reviewers:
+  - "@haircommander"
+  - "@mrunalp"
+  - "@umohnani8"
+approvers:
+  - "@haircommander"
+  - "@mrunalp"
+  - "@umohnani8"
+api-approvers:
+  - None
+creation-date: 2021-01-18
+last-updated: 2021-05-19
+tracking-link:
+  - https://issues.redhat.com/browse/CNF-5326
+---
+
+# Hide Container Mountpoints
+
+## Summary
+
+The current implementation of Kubelet and CRI-O both use the top-level
+namespace for all container and Kubelet mountpoints. However, moving these
+container-specific mountpoints into a private namespace reduced systemd
+overhead with no difference in functionality.
+
+## Motivation
+
+systemd scans and re-scans mountpoints many times, adding a lot to the CPU
+utilization of systemd and overall overhead of the host OS running OpenShift.
+Changing systemd to reduce its scanning overhead is tracked in [BZ
+1819868](https://bugzilla.redhat.com/show_bug.cgi?id=1819868), but we can work
+around this exclusively within OpenShift. Using a separate mount namespace for
+both CRI-O and Kubelet can completely segregate all container-specific mounts
+away from any systemd or other host OS interaction whatsoever.
+
+### User Stories
+
+As an OpenShift system administrator, I want systemd to consume less resources
+so that I can run more workloads on my system.
+
+As an OpenShift system administrator, I want to disable the
+mount-namespace-hiding feature so that I can fall back to the previous system
+behavior.
+
+As an OpenShift developer or support engineer, I want to inspect
+kubernetes-specific mountpoints as part of debugging issues.
+
+### Goals
+
+- Mounts originating in CRI-O, OCI hooks, Kubelet, and container volumeMounts
+  with `mountPropagation: Bidirectional` are no longer visible to systemd or
+  the host OS
+- Mounts originating in the host OS are still visible to CRI-O, OCI hooks,
+  Kubelet, and container volumeMounts with `MountPropagation: HostToContainer`
+  (or Bidirectional)
+- Mounts originating in CRI-O, OCI hooks, Kubelet, and container volumeMounts
+  with `mountPropagation: Bidirectional` are still visible to each other and
+  container volumeMounts with `MountPropagation: HostToContainer`
+- Restarting either `crio.service` or `kubelet.service` does not result in the
+  mount visibility getting out-of-sync
+
+### Non-Goals
+
+- Fix systemd mountpoint scanning overhead
+
+## Proposal
+
+### Workflow Description
+
+Generally speaking, the end-user experience should not be affected in any way
+by this proposal, as there is no outward API changes. There is some
+supportability difference, since anyone attempting to inspect the CRI-O or
+Kubelet mountpoints externally would need to be aware that these are now
+available in a different namespace than the default top-level systemd mount
+namespace.
+
+For Tech Preview, the feature must be enabled by adding a MachineConfig which
+enables the new `kubens.service` systemd unit which drives the whole feature.
+For GA, the feature would be enabled by default, and may be disabled by adding
+a MachineConfig that disables the `kubens.service` systemd unit.
+
+For any containers running in the system, there should be no observable
+difference in the behavior of the system.
+
+For any administrative shells or processes running outside of containers on the
+host, the Kubernetes-specific mountpoints will no longer be visible by default.
+Entering the new mount namespace via the `kubensenter` script will make these
+mountpoints visible again.
+
+### API Extensions
+
+No API changes required.
+
+The existing APIs available within MachineConfig objects can be used to
+enable/disable the `kubens.service` which in turn enables/disables this
+feature.
+
+### Implementation Details/Notes/Constraints
+
+We will create a separate mount namespace and cause both CRI-O and Kubelet
+to launch within it to hide their many many mounts from systemd by:
+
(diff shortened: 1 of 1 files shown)


dhellmann:
> After some time, this proposal is ready to see the light of day again.

You'll want to update the doc to use the newer template. Part of that will include ensuring that the metadata links to a ticket (usually in Jira) where the priority and implementation status are being tracked.

lack:
> > After some time, this proposal is ready to see the light of day again.
> 
> You'll want to update the doc to use the newer template. Part of that will include ensuring that the metadata links to a ticket (usually in Jira) where the priority and implementation status are being tracked.

Thanks!  The linter was a great help on helping me figure out what I needed to do to get this one back into shape.  Hopefully I've got it going in the right direction now!

enhancements/hide-container-mountpoints.md
```diff
@@ -0,0 +1,333 @@
...
+- If the namespace is not configured correctly to allow mounts from the OS to
+  be shared into Kubelet or CRI-O, system mountpoints will not be visible to
+  either Kubelet or CRI-O or runing containers.
+
+#### Support Procedures
+
+When this feature is enabled, a shell on a node and the 'oc debug' container
+will not have visibility of the Kubernetes mountpoints.
+
+- To start a shell within the container mount namespace, execute the
```
haircommander, line 273: I think updates to the debug pod MOTD or a wrapper script around the pod's chroot could be helpful

haircommander, line 273: I think updates to the debug pod MOTD or a wrapper script around the pod's chroot could be helpful

lack: I think I'd vote for a MOTD update first, since wrapping chroot would assume a user always wants to enter the host filesystem AND the hidden mountspace at the same time, whereas it may be that there's reason to have the control of doing one or the other or both.

enhancements/hide-container-mountpoints.md
```diff
@@ -0,0 +1,339 @@
...
+## Summary
+
+The current implementation of Kubelet and CRI-O both use the top-level
+namespace for all container and Kubelet mountpoints. However, moving these
+container-specific mountpoints into a private namespace reduced systemd
+overhead with no difference in functionality.
+
+## Motivation
+
+Systemd scans and re-scans mountpoints many times, adding a lot to the CPU
```
mrunalp, line 32: nit: Systemd -> systemd.

lack: Even at the beginning of a sentence?

enhancements/hide-container-mountpoints.md
```diff
@@ -0,0 +1,339 @@
...
+The 3rd guarantee is explicitly removed by this proposal.
+
+This means that:
+- Administrators who have connected to the host OS and want to inspect the
+  mountpoints originating from CRI-O, OCI hooks, Kubelet, or containers will
+  not be able to see them unless they explicitly enter the 2nd-level namespace.
+- Any external or 3rd-party tools which run in the host mount namespace but
+  expect to see mountpoints created by CRI-O, OCI hooks, Kubelet, or containers
+  would need to be changed to enter the specific container namespace in order
+  to see them.
```
mrunalp, line 184: We can call out that this would break security scanning products. Also, need to check what work we need to do if any in stackrock.

enhancements/hide-container-mountpoints.md
```diff
@@ -0,0 +1,339 @@
...
+### Initial proof-of-concept
+
+Original work [here](https://github.com/lack/redhat-notes/tree/main/crio_unshare_mounts).
+
+It has MC objects that create:
+- The new `container-mount-namespace.service` service
+- Override files for both `crio.service` and `kubelet.service` which add the
+  appropriate systemd dependencies upon the `container-mount-namespace.service`
+  and wrap ExecStart inside of `nsenter`
+- A convenience utility called `/usr/local/bin/nsenterCmns` which can be used by
```
mrunalp, line 319: nit: Maybe keep the utitlity name all small caps?

lack: This name already exists in the wild as pre-existing work; the proposed name of the similar utility will be `kubensenter`

enhancements/hide-container-mountpoints.md
lack: Who would be best suited to ask about any potential implications for stackrox?

enhancements/hide-container-mountpoints.md
```diff
@@ -0,0 +1,339 @@
...
+This means that:
+- Administrators who have connected to the host OS and want to inspect the
+  mountpoints originating from CRI-O, OCI hooks, Kubelet, or containers will
+  not be able to see them unless they explicitly enter the 2nd-level namespace.
+- Any external or 3rd-party tools which run in the host mount namespace but
+  expect to see mountpoints created by CRI-O, OCI hooks, Kubelet, or containers
+  would need to be changed to enter the specific container namespace in order
+  to see them.
+
+We will mitigate this by adding a helper application to easily enter the right
```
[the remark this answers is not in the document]
lack: In the "Support Procedures" section below, at the suggestion of @haircommander, I've recommended we update the MOTD for the `oc debug` container to mention this utility.  

Where else should we document this helper?

enhancements/hide-container-mountpoints.md
mrunalp: Adding @connorgorman for stackrox.

enhancements/hide-container-mountpoints.md
mrunalp: Yeah, since it is a name :) See https://www.freedesktop.org/wiki/Software/systemd/ #Spelling section.

enhancements/hide-container-mountpoints.md
connorgorman: Please see the following lines. I'm don't understand the proposal completely as I'm not sure which paths would start being hidden, but these are the paths that we are currently tracking including tapping into systemd to see certain services
https://github.com/stackrox/stackrox/blob/master/compliance/collection/file/file.go#L36-L92

connorgorman: @robbycochran Can you take a look at this enhancement and help determine if it will impact collector?

lack: @connorgorman and @robbycochran : Here's a quick run-down of how this proposal might affect you, and when it should not.

- If your code runs inside a container, you shouldn't notice a difference except if you're doing privileged operations to access the host OS's mount namespace.
- If your code runs on the host OS (or asks systemd on the host OS for its perspective), the main difference is that systemd and other host OS processes won't have immediate access to any mountpoints created by kubelet or cri-o, or any containers like CSI providers.  This includes the following general classes of mountpoints:
  - Overlay mounts under /var/lib/containers/storage/ and /run/containers/storage/
  - Namespace bindmounts under /run/*ns/
  - secrets and volumes mounted under /var/lib/kubelet/pods/

If you're in this 2nd group, you can always use 'nsenter' or 'setns()' to jump into the kubernetes-specific mount namespace, and all the mounts will then be visible again.

haircommander:
LGTM

enhancements/hide-container-mountpoints.md
robbycochran: @lack StackRox uses privileged containers to access the host OS directories, including `/var/run/docker.sock`, `/proc`, `/etc`, `/usr/lib`, `/sys`, and `/dev` and the other paths listed in the compliance file above.  However, we don't access any mount points created by kubelet or cri-o. So I don't think this change will impact us.

lack: @robbycochran I did see `/proc/mounts` listed in the file.go that @connorgorman posted above, and what you get back from there definitely depends on what mount namespace you're in...

connorgorman: @lack Just looked into the code /proc/mounts is only used in one case of the compliance container to check that there is a container partition as a part of the Docker benchmark. I think losing this one case is okay as this benchmark is less relevant moving forward without the deprecation of docker bench

enhancements/hide-container-mountpoints.md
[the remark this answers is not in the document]
mrunalp: I was thinkihng that we could document an example system service that joins the mount namespace using our helper application.

mrunalp (approved):

enhancements/hide-container-mountpoints.md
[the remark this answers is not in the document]
lack: That's a great idea!

PR #1132, opened 2022-05-23
pod-security-admission: clarify final restricted profile goal

s-urbaniak:
The final clarification of enabling restricted pod security profile by default was not clear enough, this fixes it.

/cc @bparees @deads2k @stlaz @mfojtik

### DIFF
--- a/enhancements/authentication/pod-security-admission.md
+++ b/enhancements/authentication/pod-security-admission.md
@@ -51,7 +51,7 @@ Note that PodSecurity admission is validating only while SCC is mutating admissi
 
 ### Goals
 
-[1] Enable PodSecurity admission
+[1] Enable PodSecurity admission with "restricted" pod security profile by default
 [2] Design an architecture that lets PodSecurity coexist with SCC logic
 
 ### Non-Goals
@@ -66,7 +66,7 @@ We propose introducing PodSecurity admission in a multi-step process. We suggest
 
 For OpenShift, the following high level migration plan is envisioned:
 
-1. (DONE) Enable the PodSecurity admission plugin in no-op mode but with the ability to audit policy violations.
+1. Enable the PodSecurity admission plugin in no-op mode but with the ability to audit policy violations.
 
 Initially, PodSecurity is configured with the following settings:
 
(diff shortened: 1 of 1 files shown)


enhancements/authentication/pod-security-admission.md
```diff
@@ -111,6 +117,18 @@ spec:
               warn-version: latest
 ```
 
+4. Increase enforce audit profile to "restricted"
```
bparees, line 120: did you mean just `enforce` here?  `audit` became restricted in an earlier step.

deads2k:
The overall flow seems reasonable, needs comments addressed before lgtm.

/approve

enhancements/authentication/pod-security-admission.md
s-urbaniak: yes, just `enforce`. As @stlaz mentioned in https://github.com/openshift/enhancements/pull/1132#pullrequestreview-981843217 `warn` and `audit` are really not necessary in this case.

enhancements/authentication/pod-security-admission.md
```diff
@@ -111,6 +117,18 @@ spec:
...
+
+Finally, after iterating over and adjusting workloads the "restricted" pod security profile is going to be enabled with the following settings:
+
+| Policy  | Profile |
+| --------| --------|
+| enforce | restricted |
+| warn    | restricted |
+| audit   | restricted |
```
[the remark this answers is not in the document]
s-urbaniak: fixed

enhancements/authentication/pod-security-admission.md
s-urbaniak: reasoning: step 3. (the step we're in right now in the cycle) is a "grace period" where we can introspect out-of-process audit logs (aka `audit`) and in-process client side warnings (aka `warn`) to have a signal what workload needs to be converted. Once we switch to `enforce` admission will hard-reject non-compliant workloads, never reaching `audit` or `warn`.

stlaz: I believe s/enforce audit profile/enforce profile/ is the suggested change which I would also like to see

stlaz (changes requested):

enhancements/authentication/pod-security-admission.md
s-urbaniak: fixed

PR #1133, opened 2022-05-25
Guidelines for the use of eBPF in OCP

danwinship:
People want to use eBPF in OCP. This is awesome and scary, and we need to come up with guidelines on how this will work.

This is currently very very WIP and I'll need help from various other teams to fill in all of the details.

### DIFF
--- a/enhancements/ebpf-in-ocp.md
+++ b/enhancements/ebpf-in-ocp.md
@@ -1,0 +1,211 @@
+---
+title: ebpf-in-ocp
+authors:
+  - "@danwinship"
+reviewers:
+  - "@TBD, for network observability"
+  - "@TBD, for other existing eBPF use case X"
+  - "@TBD, for other future eBPF use case Y"
+  - "@dave-tucker, for bpfd"
+approvers:
+  - TBD
+api-approvers:
+  - None
+creation-date: 2022-05-25
+last-updated: 2022-05-25
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - TBD
+see-also:
+replaces:
+superseded-by:
+---
+
+# Guidelines for the Use of eBPF in OCP
+
+## Summary
+
+OCP developers and customers alike would like to use eBPF in OCP
+clusters. At the present time, _some_ eBPF hooks are designed to be
+"shareable", but others are not (and some are shareable only when used
+in certain restricted ways). We need to create rules for how OCP will
+use eBPF and how we expect others to use (or not use it) to prevent
+conflicts later.
+
+## Motivation
+
+### User Stories
+
+#### Existing Use Case Number 1
+
+(Something about observability or tracing where people are already
+using eBPF in OCP.)
+
+#### Upcoming Network Observability Features
+
+(Something about upcoming network observability features.)
+
+#### Ensuring Network (and General) Supportability
+
+As an OCP developer or support engineer, I would like to be able to
+debug customer problems without having to worry that random
+"unsanctioned" eBPF is messing things up in inexplicable ways.
+
+#### Customer eBPF Bug Patching
+
+As an OCP customer, I want to work around an OCP/RHEL bug by using a
+`MachineConfig` (or something similar) to deploy a small eBPF program
+to every node in my cluster, without voiding my warranty.
+
+I might be OK with needing a support exception, since this is just to
+work around a bug that should eventually be fixed.
+
+#### Customer eBPF Feature Patching
+
+As an OCP customer, I want to add functionality to OCP/RHEL by using a
+`MachineConfig` (or something similar) to deploy a small (or
+not-so-small) eBPF program to every node in my cluster, without
+voiding my warranty.
+
+I might be OK with needing a support exception.
+
+(eg, concretely, there is a customer who wants to use eBPF to
+implement [OVS-style SLB
+Bonding](https://docs.openvswitch.org/en/latest/topics/bonding/#slb-bonding)
+on a kernel bond interface, using eBPF. The upstream Linux maintainer
+rejected a patch to implement it in the kernel driver directly, in
+part because it would be so easy to implement it with eBPF.)
+
+#### Cilium as a Supported Third-Party Network Plugin
+
+As an OCP customer, I want to deploy an OCP cluster using Cilium as a
+certified third-party network plugin, without worrying that any other
+OCP components are going to interfere with it (or vice versa).
+
+Cilium uses a wide variety of eBPF hooks, including FIXME.
+
+#### eBPF From OLM
+
+As a third-party developer of an operator that uses eBPF, I would like
+to make my operator available via OLM and have it interoperate
+smoothly with existing and future OCP releases.
+
+#### Future eBPF Data Plane Functionality
+
+As an OCP developer/product manager, I would like to introduce
+functionality in a future release that makes use of the XDP and/or tc
+eBPF hooks to modify data plane functionality, and have these new
+features interact in a predictable way with existing networking
+functionality, other operators, and local customizations.
+
+#### Security
+
+As a cluster administrator, I would like to make sure that no one is
+using eBPF that I don't know about, because I am in a highly regulated
+industry, and I can't have random people running mystery code in my
+kernel, even if it's sandboxed.
+
+### Goals
+
+- Categorize the existing, near-term-planned, and long-term-envisioned
+  uses of eBPF in OCP.
+
+- Create a set of rules for using eBPF in OCP to ensure that different
+  eBPF users do not clobber each other's programs and do not cause
+  system-wide instability or un-debuggability.
+
+  - We may need different rules for different kinds of eBPF
+    programs.
+
+### Non-Goals
+
+-
+
+## Proposal
+
+TBD
+
+### Workflow Description
+
+N/A - not an end-user feature
+
+### API Extensions
+
+TBD - none expected
+
+### Implementation Details/Notes/Constraints [optional]
+
+### Risks and Mitigations
+
+- Continuing to allow arbitrary eBPF without any oversight will likely
+  eventually lead to buggy clusters, difficult-to-debug problems, and
+  possibly security holes.
+
+- OTOH adopting a too-strict policy toward customer use of eBPF may
+  turn some customers away (and would be awkward, messaging-wise:
+  "eBPF is great! You can't use it!").
+
+  - We should look into what RHEL is doing to balance
+    "supportability" vs "letting the customer do what they want"
+    with eBPF.
+
+### Drawbacks
+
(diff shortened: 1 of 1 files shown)


enhancements/ebpf-in-ocp.md
```diff
@@ -0,0 +1,211 @@
...
+### User Stories
+
+#### Existing Use Case Number 1
+
+(Something about observability or tracing where people are already
+using eBPF in OCP.)
+
+#### Upcoming Network Observability Features
+
+(Something about upcoming network observability features.)
```
mariomac, line 45:
```suggestion
As an OCP developer, I want to attach the eBPF agent into different parts of
the system (e.g. Nodes' kernel) to be able to trace the network flows within
the system, as well as being able to correlate them with other parts of the
system configuration (e.g. Network Policies, security, etc.).
```

jotak: should we also describe the end user (OCP customer) use cases that we address? Something like: "as an OCP customer, I want to be able to trace network flows across pods and services in my cluster, with a resource overhead as low as possible." ?

danwinship: I don't think we need to get into the higher-level goals that led to people using eBPF... for purposes of this enhancement, we're starting with "I want to use eBPF to do X" and not really worrying about how you got to that point. At least for OCP-internal use cases. For customer use cases we may need to distinguish between "customer wants to use eBPF and we think it's reasonable" vs "customer wants to use eBPF and we think it's a bad idea". (eg, that's why I separated out the "eBPF for bugfixes" and "eBPF for new features" use cases, because we might have different opinions on them)

enhancements/ebpf-in-ocp.md
```diff
@@ -0,0 +1,211 @@
...
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - TBD
+see-also:
+replaces:
+superseded-by:
+---
+
+# Guidelines for the Use of eBPF in OCP
+
+## Summary
```
eranra, line 25: @danwinship should add something on listing and addressing support OCP platforms. So that eBPF programs are tested and documented as supported for a known set of OCP architectures (x86,arm ... )

eranra: @danwinship same with regards to versioning and required kernel levels... so that it will be possible to list dynamically all supported loaded eBPF programs and also address the issue of upgrading etc.

enhancements/ebpf-in-ocp.md
```diff
@@ -0,0 +1,211 @@
...
+on a kernel bond interface, using eBPF. The upstream Linux maintainer
+rejected a patch to implement it in the kernel driver directly, in
+part because it would be so easy to implement it with eBPF.)
+
+#### Cilium as a Supported Third-Party Network Plugin
+
+As an OCP customer, I want to deploy an OCP cluster using Cilium as a
+certified third-party network plugin, without worrying that any other
+OCP components are going to interfere with it (or vice versa).
+
```
astoycos, line 83: I feel like CI will be the best way to ensure this... should we have an "ebpf feature" CI workflow to ensure stable cooperation? Maybe a job running with cilium + a job running with any OCP specific ebpf features

astoycos: (obviously good documentation on what hooks we're using explicitly is important to) but yeah the problem of ebpf interoperability is definitely a complex one.

enhancements/ebpf-in-ocp.md
```diff
@@ -0,0 +1,211 @@
...
+As an OCP developer/product manager, I would like to introduce
+functionality in a future release that makes use of the XDP and/or tc
+eBPF hooks to modify data plane functionality, and have these new
+features interact in a predictable way with existing networking
+functionality, other operators, and local customizations.
+
+#### Security
+
+As a cluster administrator, I would like to make sure that no one is
```
astoycos, line 102: Currently only privileged pods would be able to run ebpf programs... So the admin should know about any of these right?  Also As we start running bpf programs it probably will make sense to have an OCP system pod that gives some stats on all of the bpf programs currently running in a cluster (maybe this is observability's Job)

enhancements/ebpf-in-ocp.md
vrutkovs: We do have Cilium jobs in CI - https://prow.ci.openshift.org/?job=*cilium*. They however run just the conformance suite, no custom eBPF tests

enhancements/ebpf-in-ocp.md
mariomac: Being more precise, it is not mandatory that pods are full-privileged to run eBPF programs. For example, the NetObserv eBPF agent runs in a non-privileged pod, in a container defining the following security context:

```yaml
        securityContext:
          capabilities:
            add:
              - BPF
              - PERFMON
              - NET_ADMIN
              - SYS_RESOURCE
          runAsUser: 0
```

williamcaban:
A use case for consideration: 
- There is a lot of interest from partners and customers for reducing their dependencies to DPDK while still getting "good enough" performance using eBPF based stack. Also, be able to do that without secondary interfaces. E.g. have the capability native on the primary Pod interface (eth0).
- Alternatives to a vhost-user and virtio-pmd, etc

enhancements/ebpf-in-ocp.md
```diff
@@ -0,0 +1,211 @@
...
+#### Existing Use Case Number 1
+
+(Something about observability or tracing where people are already
+using eBPF in OCP.)
+
+#### Upcoming Network Observability Features
+
+(Something about upcoming network observability features.)
+
+#### Ensuring Network (and General) Supportability
```
dofinn, line 47: This would be amazing for CEE supporting managed OpenShift

enhancements/ebpf-in-ocp.md
dofinn: additionally there is pixie https://docs.px.dev/about-pixie/pixie-ebpf/

enhancements/ebpf-in-ocp.md
sherine-k: Would SELinux profile (CAP_BPF) help keep the attack surface a little smaller than using privileged containers (when needed)?

enhancements/ebpf-in-ocp.md
```diff
@@ -0,0 +1,211 @@
...
+eBPF hooks to modify data plane functionality, and have these new
+features interact in a predictable way with existing networking
+functionality, other operators, and local customizations.
+
+#### Security
+
+As a cluster administrator, I would like to make sure that no one is
+using eBPF that I don't know about, because I am in a highly regulated
+industry, and I can't have random people running mystery code in my
+kernel, even if it's sandboxed.
```
netoptimizer, line 105: One option for this "cluster administrator" would be to write a BPF LSM hook program that denies loading BPF programs regardless of their security level/capabilities.
@tohojo wrote this example to disable BPF:
https://github.com/xdp-project/bpf-examples/blob/master/lsm-nobpf

enhancements/ebpf-in-ocp.md
```diff
@@ -0,0 +1,211 @@
...
+use eBPF and how we expect others to use (or not use it) to prevent
+conflicts later.
+
+## Motivation
+
+### User Stories
+
+#### Existing Use Case Number 1
+
+(Something about observability or tracing where people are already
```
dave-tucker, line 40: As an OCP customer I want to orchestrate one (or more) probes to run on my nodes for either monitoring application performing application debugging.

As an OCP customer who is unable to modify application code, I would like to add eBPF-based instrumentation for my application.

saschagrunert:
I think this would be still valuable
/reopen

enhancements/ebpf-in-ocp.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Drawbacks
+
+None. We really need guidelines here.
+
+## Design Details
+
+### Open Questions [optional]
+
+Many
```
dave-tucker, line 159: Open question:

Where do we stand with support on RHEL. The [latest release notes](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/9.0_release_notes/new-features) state:

>  The XDP and AF_XDP API parts of the libbpf library are not supported and may be removed in a future release. 

We may not use `libbpf` to load XDP or AF_XDP programs.

> The eXpress Data Path (XDP) feature, which provides access to the received packets before the kernel networking stack processes them. Red Hat supports XDP only if it is used through the libxdp library.

We must use libxdp BUT we do not have higher level language bindings for it (i.e for Go).

> The xdp-tools package, which contains user-space support utilities for the XDP feature and is supported on the AMD64 and Intel64 CPU architectures.

libxdp is not packaged for any architecture other than x86_64

tohojo: For RHEL9, libxdp is packaged for all archs, it just won't do multi-prog until the kernel support lands; but that's fine, it'll fall back automatically, so it should be totally fine to use it as-is for those archs. So the lack of Go bindings is the only blocker AFAICT...

dave-tucker: thanks for clarifying @tohojo
we'll have to figure out a design for libxdp-go that works seamlessly with `cillium/ebpf`
lets' chat about that next week and perhaps we can get something started

enhancements/ebpf-in-ocp.md
```diff
@@ -0,0 +1,211 @@
...
+rejected a patch to implement it in the kernel driver directly, in
+part because it would be so easy to implement it with eBPF.)
+
+#### Cilium as a Supported Third-Party Network Plugin
+
+As an OCP customer, I want to deploy an OCP cluster using Cilium as a
+certified third-party network plugin, without worrying that any other
+OCP components are going to interfere with it (or vice versa).
+
+Cilium uses a wide variety of eBPF hooks, including FIXME.
```
markmc, line 84: This talk on Cilium's experience seems relevant: https://lpc.events/event/16/contributions/1353/attachments/1068/2130/plumbers_2022_tc_bpf_links.pdf

tl;dr - they saw datadog stomping on Cilium's hooks in a customer environment; they believe bpf_link attachment semantics help with this problem; they propose a model for using bpf_link with tc

netoptimizer: Yes, @markmc this example of (Datadog) BPF stomping on (Cilium) BPF is exactly the problem I'm trying to highlight.

For legal/support reasons RH must document what BPF hooks our products use and state RH cannot support shared usage of said BPF hooks. Regardless this stomping _will occur in practice_ in production environments, leading to **hard-to-debug** scenarios.

**IMHO** we should implement a technical solution, that:

1. as minimum detects and reports if a products BPF hook get "stomped"
2. preferably detects and _denies_ others from taking over products BPF hook

PR #1134, opened 2022-05-25
installer: adding gcp-ipi-shared-vpc

jstuever:
This enhancement adds the capability to deploy a GCP cluster into a Shared VPC
(XPN) configuration using the Installer Provisioned Infrastructure (IPI)
workflow.

### DIFF
--- a/enhancements/installer/gcp-ipi-shared-vpc.md
+++ b/enhancements/installer/gcp-ipi-shared-vpc.md
@@ -1,0 +1,282 @@
+---
+title: gcp-ipi-shared-vpc
+authors:
+  - "@jstuever"
+reviewers:
+  - "@padillon"
+approvers:
+  - "@padillon"
+api-approvers:
+  - "@JoelSpeed"
+creation-date: 2022-05-24
+last-updated: 2022-09-21
+tracking-link:
+  - "https://issues.redhat.com/browse/CORS-1774"
+---
+
+# GCP IPI Shared VPC
+
+## Summary
+
+This enhancement adds the capability to deploy a GCP cluster into a [Shared VPC
+(XPN)](https://cloud.google.com/vpc/docs/shared-vpc) configuration using the
+Installer Provisioned Infrastructure (IPI) workflow.
+
+## Motivation
+
+The GCP Best Practices include having a separate “Service Project” for each
+application, and for those Service Projects to use a Shared VPC (XPN)
+configuration. In such a configuration, the VPC, Network, Subnetworks, and
+Firewall Rules exist in a different “Host Project”. This enables the
+application owners to have significant permissions in the Service Project while
+restricting access to the shared networking infrastructure. Several of our
+customers are asking for this to be possible from within the IPI workflow. In
+addition to this, tooling such as OSD and Hive depend on this workflow.
+
+### User Stories
+
+- As a user, I want the installer to create as many resources as possible in
+  the Host Project so I can provide sufficient credentials to maximize the
+automation of an install to a shared VPC.
+- As a user, I want the installer to use existing resources in the Host Project
+  so that I can create infrastructre in advance and minimize the privileges to
+the Host Project given to the installer when installing to a shared VPC.
+
+### Goals
+
+- Enable users to deploy GCP clusters with Shared VPC networking using the IPI
+  workflow.
+- Enable users to provide service account(s) with sufficient permissions in the
+  Host Project to enable provisioning the majority of the cluster resources.
+- Enable users to alternatively provide pre-provisioned resources to enable
+  minimal permission requirements in the Host Project.
+
+### Non-Goals
+
+- Provisioning and/or configuring the VPC, Network, and Subnetworks are to be
+  done manually by the user prior to running the installer.
+- Provisioning and/or configuring the Service Accounts for the Installer and
+  Cluster are to be done manually by the user prior to running the installer.
+- Providing solution(s) for manually managing DNS in the Host Project is to be
+  solved by another effort.
+
+## Proposal
+
+The installer should be able to install a GCP IPI cluster using Shared VPC
+networking by automating the majority of the current GCP UPI workflow. This
+requires some resources to be created and configured in advance, such as the
+network and subnets. Other resources can be created in advance when the
+necessary host project permissions are not desired, such as the firewall rules
+and/or DNS configuration. The installer, and the cluster-ingress-operator, will
+be updated to enable some of this functionality. They will become capable of
+creating and managing resources that exist in a different (Host) project from
+the cluster (Service Project). They will also become capable of consuming
+pre-existing firewall rules.
+
+### Workflow Description
+
+Prior to executing the installer, the Shared VPC, Network, and Subnets will be
+manually created in the Host Project and configured for Shared VPC (XPN)
+configuration with the Service Project. A service account will also be created
+in the Service Project and granted sufficient permissions in the Host Project.
+When minimal permissions are desired in the Host Project, additional resources
+will be manually created and configured to include: firewall rules, api compute
+address, ingress compute address, and the api, api-int, and \*.apps public
+and/or private DNS records.
+
+After all of these resources are created, they will be used in a manually
+created install-config.yaml to configure the cluster accordingly. At a minimum,
+the credentialsMode will be set to Passthrough, and the network,
+computeNetwork, controlPlaneNetwork, and networkProjectID will be configured to
+reference the Shared VPC network. When minimial permissions are desired in the
+Host Project, the additional parameters will be configured to enable the
+cluster to consume them as well.
+
+The installation process can then proceed as any other IPI workflow by running
+the create manifests, create ignition, and/or create cluster sub-commands.
+
+### API Extensions
+
+#### api.dns
+
+The current DNSZone struct does not appear to be able to store which project
+the zone is in. The ID variable is currently limited to a managedZone ID value.
+The cluster-ingress-operator will be updated to also accept a relative resource
+name in the following format. This will enable the operator to override the
+default project id with the one provided. No functional API changes are
+necessary.
+
+projects/{projectid}/managedZones/{zoneid}
+
+#### installconfig.platform.gcp.networkProjectID
+
+A new parameter will be added to the install-config to specify the project id
+of the network and subnets. This parameter will then be used where necessary to
+validate existing network project resources as well as provision additional
+resources using these network resources. The default will be an empty string
+and will indicate these resources (should) exist in the default cluster
+project.
+
+#### installconfig.platform.gcp.privateDNSZone
+
+A new parameter will be added to the install-config to specify the private DNS
+managed zone. This parameter will be used where necessary to validate existing
+DNS resources as well as provision additional private DNS records. The default
+will be an empty value, which will indicate these resources are to be created
+and managed in the cluster project identified by the projectID.
+
+#### installconfig.platform.gcp.publicDNSZone
+
+A new parameter will be added to the install-config to specify the public DNS
+managed zone. This parameter will be used where necessary to validate existing
+DNS resources as well as provision additional public DNS records. The default
+will be an empty value, which will indicate these resources are to be created
+and managed in the cluster project identified by the projectID.
+
+#### installconfig.platform.gcp.defaultMachinePlatform.tags
+
+A new parameter will be added to the install-config to specify additional
+networking tags to add to the cluster instances. The default will be an empty
+list, in which case no additional tags will be added.
+
+#### installconfig.controlPlane.platform.gcp.tags
+
+A new parameter will be added to the install-config to specify additional
+networking tags to add to control plane instances. This parameter will override
+any tags specified in the defaultMachinePlatform. The default will be an empty
+list, in which case no additional tags will be added.
+
+#### installconfig.compute.platform.gcp.tags
+
+A new parameter will be added to the install-config to specify additional
+networking tags to add to compute instances. This parameter will override any
+tags specified in the defaultMachinePlatform. The default will be an empty
+list, in which case no additional tags will be added.
+
+#### installconfig.platform.gcp.createFirewallRules
+
+A new parameter will be added to the install-config to enable toggling of the
+creation and management of firewall rules. When set to disabled, the installer
+will skip creation of firewall rules. The default is enabled in which firewall
+rules will be created as before. This toggle does not modify the behavior of
+the cluster-ingress-operator, which is to create firewall rules if permissions
+allow.
+
+### Implementation Details/Notes/Constraints
+
+#### VPC, Network, and Subnetworks
+
+Shared VPC network and subnet(s) are to be created and configured in advance.
+They can then be configured in the install config using the following
+parameters as enabled by prior work. A new parameter (networkProjectID) will be
+added to specify in which project they exist. This parameter will be consumed
+by the Terraform templates as well as cluster manifests in order to configure
+the cluster to use these resources.
+
+- installconfig.platform.gcp.computeSubnet
+- installconfig.platform.gcp.controlPlaneSubnet
+- installconfig.platform.gcp.network
+
+#### Service Accounts
+
+Service Account(s) are to be created and configured in advance. These accounts
+must have additional permissions granted in the Host Project. The specific
+permissions depend on the desired level of automation. The install config's
+credentialsMode parameter must be set to passthrough so the cluster uses these
+accounts.
+
+#### Firewall Rules
+
+Firewall Rules can either be created and configured in advance, or the service
+accounts above must have sufficient permissions granted to manage firewall
+rules in the Host Project. User specified tags will be added to the instances
+to enable cluster specific rules. A new install config parameter will be added
+to enable specifying if the Terraform templates should create firewall rules,
+or not.
+
+In order for the firewall rules to be created by the installer, the service
+account will require the following access in the project where the firewall
+rules are to be managed (i.e. networkProjectID)
+
+- Compute Network Admin
+- Compute Security Admin
+
+#### DNS Zones and Records
+
+DNS can either be created and configured in advance, or the service accounts
+above must have sufficient permissions to create and manage DNS resources in
+the Host Project.
+
+The ability to use pre-existing (aka external) DNS solutions is outside the
+scope of this enhancement. There are other efforts to provide this capability.
+
+New install config parameters will be added to specify the public and/or
+private DNS zones to use when managing cluster records. The default will be an
+empty value, which will result in the private zone and public/private records
+being managed in the service account as before.
+
+In order for the DNS Zones and Records to be managed by the installer, the
+service account will require the following access in the in the project(s)
+where the DNS Zones are located.
+
+- DNS Administrator
+
+### Risks and Mitigations
+
+Adding complexity to the Terraform templates may affect the existing IPI
+workflows. Our CI and QE processes should catch any issues here.
+
+### Drawbacks
+
+Extending the GCP IPI workflow to enable this functionality will likely require
+additional cycles to maintain and test as our product changes over time.
+
+## Design Details
+
+### Test Plan
+
+A CI job will be created to test this functionality, replacing the existing GCP
+UPI XPN workflow.
+
+### Graduation Criteria
+
+#### Dev Preview -> Tech Preview
+
+The work necessary to deliver this enhancement can be split up into multiple,
+indipendent epics. The majority of these will become Tech Preview as they are
+added to the installer.
+
+One exception is the network tags epic, which is immediately useful on it's own
+and will be released immediately GA.
+
+#### Tech Preview -> GA
+
+When all of the work is completed and proven effective through Tech Preview,
+the whole of this enhancement will move into GA.
+
+#### Removing a deprecated feature
+
+There are no deprecated features.
+
+### Upgrade / Downgrade Strategy
+
+This enhancement should not affect upgrade, as it only affects install time.
+
+### Version Skew Strategy
+
+There is no expected need to manage version skew.
+
+### Operational Aspects of API Extensions
+
+The api extensions are discussed in the proposal section above.
+
+#### Failure Modes
+
+The user provided parameters will be validated, thereby failing early and
+reducing failure modes in general.
+
+#### Support Procedures
+
+## Implementation History
+
+## Alternatives


enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,262 @@
...
+the create manifests, create ignition, and/or create cluster sub-commands.
+
+### API Extensions
+
+#### api.dns
+
+The current DNSZone struct does not appear to be able to store which project
+the zone is in. The ID variable can be used to store fully qualified path
+names, which will include the project. However, this may require modification
+of all consumers of this variable to be able to parse this information.
```
patrickdillon, line 106: It looks like the [current behavior for the cluster ingress operator is](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/dns/controller.go#L563) to use the project from [the cluster infrastructure object](https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L452).

We probably need to add a field to that object.

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,262 @@
...
+parameter is set to “none”, the cluster will not be configured to manage
+private DNS resources.
+
+#### installconfig.platform.gcp.publicManagedZone
+
+A new parameter will be added to the install-config to specify the public DNS
+managed zone. This parameter will be used where necessary to validate existing
+DNS resources as well as provision additional public DNS records. The default
+will be an empty string, which will indicate these resources are to be created
+and managed in the project identified by the networkProjectID (when publish is
```
patrickdillon, line 142: The public DNS zone can also go in the service project, right? 

1. Do we have a sense of whether users prefer creating the zones in the host/service project?
2. Would we create a knob to decide which project the public zone is going in?

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,262 @@
...
+
+A new parameter will be added to the install-config to specify a pre-existing
+compute address resource to be used when creating and managing the API load
+balancer(s).
+
+#### installconfig.platform.gcp.ingressAddress
+
+A new parameter will be added to the install-config to specify a pre-existing
+compute address resource to be used when creating and managing the default
+ingress load balancer(s).
```
patrickdillon, line 156: I think these are optional too, like the previous fields.

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,262 @@
...
+- installconfig.platform.gcp.computeSubnet
+- installconfig.platform.gcp.controlPlaneSubnet
+- installconfig.platform.gcp.network
+
+#### Service Accounts
+
+Service Account(s) are to be created and configured in advance. These accounts
+must have additional permissions granted in the Host Project. The specific
+permissions depend on the desired level of automation. The install config's
+credentialsMode parameter must be set to passthrough so the cluster uses these
```
patrickdillon, line 178: 1. The installer should default this to passthrough once we know this is XPN
2. a nit: but manual mode should be able to work here as well

patrickdillon:
From the installer perspective, we also need to think about how this would affect destroy code

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: The public zone can go in either, or an unrelated project. The knob we are providing is going to be the selfLink to an existing zone, which will include the project name of which it belongs.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: Correct. Updated the text to indicate this.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: I agree. I intended to enable the ingress operator developers to determine the field specifics.

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,262 @@
...
+- As a user, I want to use the IPI workflow to install a GCP cluster with
+  Shared VPC networking so I can streamline more of the installation process
+than is available with the UPI workflow.
+- As a user, I want to use a service account with sufficient permissions in the
+  Host Project so I can maximize the automation of the installation process
+while using the IPI workflow to install a GCP cluster with Shared VPC
+networking.
+- As a user, I want to create some resources in advance so I can minimize the
+  permission requirements in the Host Project while using the IPI workflow to
+install a GCP cluster with Shared VPC networking.
```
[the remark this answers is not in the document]
jstuever: Thanks. Updated.

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,262 @@
...
+tracking-link:
+  - "https://issues.redhat.com/browse/CORS-1774"
+---
+
+# GCP IPI Shared VPC
+
+## Summary
+
+This enhancement adds the capability to deploy a GCP cluster into a Shared VPC
+(XPN) configuration using the Installer Provisioned Infrastructure (IPI)
```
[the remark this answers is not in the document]
jstuever: Thanks, updated.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: That is something to think about, for sure. One of the things I've realized is that many of the changes we will be making are useful even when not using XPN. Therefore, I'm trying to make sure I think about each knob with this in mind. We might be able to default to passthrough when networkProjectID is not empty. On the other hand, having the user specifically set this to passthrough makes it clear to them that the credentials they are using will become the cluster credentials. I believe we've had parallel conversations before.

jstuever:
/cc @JoelSpeed
I added you as an api approver. Let me know if I need to modify this.

enhancements/installer/gcp-ipi-shared-vpc.md
JoelSpeed: Just in case it's useful, for MAPI we added a field to the network interface struct, if a user sets this to a different project ID than that of the Machine, we create the NIC in the other project, https://github.com/openshift/api/blob/6c4f186339a7a4f505e133e9b86acd0d5aa84686/machine/v1beta1/types_gcpprovider.go#L134

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,260 @@
...
+of all consumers of this variable to be able to parse this information.
+Alternatively, a second variable can be added to store the projectID and the
+consumers can use that information when linking to the DNS Zone.
+
+#### api.ingress
+
+The current ingress structures do not appear to be able to store an address to
+be used when creating the ingress load balancer(s). New field(s) may be needed
+to enable this functionality. They can then be passed to the gcp lagacy
+provider for creating the ingress load balancers.
```
JoelSpeed, line 111: Will this be by setting the information in the cloud-config?

Do we know if any additional changes are needed for the out of tree cloud provider?

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,260 @@
...
+compute address resource to be used when creating and managing the API load
+balancer(s). The default will be an empty string, in which case this address
+will be generated via the Terraform templates as per usual.
+
+#### installconfig.platform.gcp.ingressAddress
+
+A new parameter will be added to the install-config to specify a pre-existing
+compute address resource to be used when creating and managing the default
+ingress load balancer(s). THe default will be an empty string, in which case
+this address will be generated by the gcp lagacy provider as per usual.
```
JoelSpeed, line 154:
```suggestion
this address will be generated by the gcp legacy provider as per usual.
```

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,260 @@
...
+consumers can use that information when linking to the DNS Zone.
+
+#### api.ingress
+
+The current ingress structures do not appear to be able to store an address to
+be used when creating the ingress load balancer(s). New field(s) may be needed
+to enable this functionality. They can then be passed to the gcp lagacy
+provider for creating the ingress load balancers.
+
+#### installconfig.platform.gcp.networkProjectID
```
JoelSpeed, line 113: So, something that comes to mind is that we do already have some customers using XPN, though via UPI IIRC. Reading through this enhancement, it seems that all the networking resources end up in the second project, is that the desired outcome here or is there a use case where some resource might want to be in the second project, and others in the main project? (It's been a while since I looked at XPN in depth)

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,260 @@
...
+the credentialsMode will be set to Passthrough, and the network,
+computeNetwork, controlPlaneNetwork, and networkProjectID will be configured to
+reference the Shared VPC network. When minimial permissions are desired in the
+Host Project, the additional parameters will be configured to enable the
+cluster to consume them as well.
+
+The installation process can then proceed as any other IPI workflow by running
+the create manifests, create ignition, and/or create cluster sub-commands.
+
+### API Extensions
```
JoelSpeed, line 95: Are there any changes to openshift APIs that reside in cluster? This all seems to be install config changes if I'm understanding correctly.

Have we got some description of the ingress API changes required for this?

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,260 @@
...
+names, which will include the project. However, this may require modification
+of all consumers of this variable to be able to parse this information.
+Alternatively, a second variable can be added to store the projectID and the
+consumers can use that information when linking to the DNS Zone.
+
+#### api.ingress
+
+The current ingress structures do not appear to be able to store an address to
+be used when creating the ingress load balancer(s). New field(s) may be needed
+to enable this functionality. They can then be passed to the gcp lagacy
```
JoelSpeed, line 110:
```suggestion
to enable this functionality. They can then be passed to the gcp legacy
```

enhancements/installer/gcp-ipi-shared-vpc.md
JoelSpeed: Can we enumerate the permissions changes/additional permissions that might be needed somewhere within the enhancement?

JoelSpeed (commented):
Will the installer be populating the MachineSet/Control Plane Machines in a different way once this enhancement is implemented?

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: In my testing, I used the following access in the host project. However, I suspect we can tune this down to individual requirements, and for each "thing" we are enabling (such as DNS, firewall, etc). I can add them to the enhancement if you think that would be useful, but I was hoping to scope them out better as part of the effort(s) to add these features.
Compute Instance Admin (beta)
Compute Network Admin
Compute Security Admin
DNS Administrator

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: I believe the instance code is already good go go in regard to networking. The majority of the effort here is about the firewalls, DNS, and load balancers by extension.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: Correct. Most of what is done in the "GCP UPI XPN" workflow will be replicated here. This includes adding the network information to the `manifests/cloud-provider-config.yaml` (network-project-id, network-name, and subnetwork-name) during the 'create manifests` stage of the installer. For the private DNS zone, I believe we can assume it exists in the same network-project-id. However, for the public zone the plan is to be more flexible and allow any project, if possible.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: Thanks! Fixed.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: For GCP UPI XPN, everything that lands in the "host project" does so because it is a requirement for Shared VPC configuration. Anything that CAN exist in the Service Project will do so. This does not change for IPI... with the exception of the "public DNS zone", where we will actually be enabling it to exist in "any project."

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: There are API changes, as mentioned in this document... as well as changes for operators. I was hoping to have the team responsible for that work help iron those details out.

jstuever: For reference, the API changes are the sections above: api.dns, api.ingress

enhancements/installer/gcp-ipi-shared-vpc.md
JoelSpeed: > I can add them to the enhancement if you think that would be useful

In my experience, this kind of information is useful to have in the enhancements so that product and docs teams have an awareness. Can always add this list provisionally now and update later if the list changes or becomes clearer in its definition

enhancements/installer/gcp-ipi-shared-vpc.md
JoelSpeed: Sounds good, thanks for the explanation. Might be worth adding some context early in the enhancement that mentions we are implementing this based on the UPI flow and the changes specifically around the cloud provider config are also good to note, thinking about future readers such as the docs and qe teams

enhancements/installer/gcp-ipi-shared-vpc.md
JoelSpeed: Cool, once you've worked out what new API fields will be required and what you'd like those to look like (seems maybe only 1 or 2?), can you write up a small Go snippet for them to represent the actual code changes to the API code including documentation and kubebuilder tags based on the [API conventions docs](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md)

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: Added permissions as requested.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: Done.

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,294 @@
...
+Host Project, the additional parameters will be configured to enable the
+cluster to consume them as well.
+
+The installation process can then proceed as any other IPI workflow by running
+the create manifests, create ignition, and/or create cluster sub-commands.
+
+### API Extensions
+
+#### api.dns
+
```
miheer, line 99: Please specify a go snip as to how it will look.
Also from my understanding installer will first need to set fields related to this in a cluster config object eg. dns cluster config object and then have the ingress operator get the information from this. I think you will be also needing some fields in the Ingress Controller CR.

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,294 @@
...
+The current DNSZone struct does not appear to be able to store which project
+the zone is in. The ID variable can be used to store fully qualified path
+names, which will include the project. However, this may require modification
+of all consumers of this variable to be able to parse this information.
+Alternatively, a second variable can be added to store the projectID and the
+consumers can use that information when linking to the DNS Zone.
+
+#### api.ingress
```
miheer, line 107: Same as above we need go snips showing how the api will look.
You might be needing to add some fields in the ingress cluster object and may be the ingress controller CR.

miheer (changes requested):

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: I've removed the need for any API code changes.

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,294 @@
...
+- Enable users to alternatively provide pre-provisioned resources to enable
+  minimal permission requirements in the Host Project.
+
+### Non-Goals
+
+- Provisioning and/or configuring the VPC, Network, and Subnetworks are to be
+  done manually by the user prior to running the installer.
+- Provisioning and/or configuring the Service Accounts for the Installer and
+  Cluster are to be done manually by the user prior to running the installer.
+
```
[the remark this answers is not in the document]
jstuever: Correct, the users will perform these operations outside of the installer codebase.

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,294 @@
...
+## Proposal
+
+The installer should be able to install a GCP IPI cluster using Shared VPC
+networking by automating the majority of the current GCP UPI workflow. This
+requires some resources to be created and configured in advance, such as the
+network and subnets. Other resources can be created in advance when the
+necessary host project permissions are not desired, such as the firewall rules
+and/or DNS configuration. The installer, and some operators, will be updated to
```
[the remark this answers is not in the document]
jstuever: To my knowledge, it's only the cluster-ingress-operator. I'd like to leave this open ended here while spelling out exactly which operator(s) later in the document.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: As stated in the updated text, there is no code change in the API itself.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: This has been moved to another effort.

jstuever:
Please note I have removed all "manually managed" DNS from this enhancement. There is another effort to handle external DNS solutions and this effort can utilize that.

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,281 @@
...
+
+Firewall Rules can either be created and configured in advance, or the service
+accounts above must have sufficient permissions granted to manage firewall
+rules in the Host Project. User specified tags will be added to the instances
+to enable cluster specific rules. A new install config parameter will be added
+to enable specifying if the Terraform templates should create firewall rules,
+or not.
+
+In order for the firewall rules to be created by the installer, the service
+account will rquire the following access in the project where the firewall
```
JoelSpeed, line 197: Typo on require

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,281 @@
...
+
+### API Extensions
+
+#### api.dns
+
+The current DNSZone struct does not appear to be able to store which project
+the zone is in. The ID variable is currently limited to a managedZone ID value.
+The cluster-ingress-operator will be updated to also accept a relative resource
+name in the following format. This will enable the operator to override the
+default project id with the one provided. No API changes are necessary.
```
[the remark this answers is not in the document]
jstuever: I didn't find any validations in my testing. I was able to use a proof of concept to enable this functionality, which I had pushed as a [PR](https://github.com/openshift/cluster-ingress-operator/pull/827) for reference. This is achieved by submitting different values for the `ID` field in [config/v1/types_dns.go](https://github.com/openshift/api/blob/27d36913c63ba542694a4be51513004ac67cd241/config/v1/types_dns.go#L58-L68). For example: ID `installer-public-zone` vs Name `projects/installer/managedZones/installer-public-zone`.

The documentation refers to the google API `managedZone.GET` function for this field. The linked documentation specifies this "Can be the managed zone name or ID." Google defines the `ID` as the "short name" and the `Name` can be in various formats to include a [relative resource name](https://cloud.google.com/apis/design/resource_names#relative_resource_name). When I first started investigating this, I was surprised when it couldn't actually use a relative name. With all of this in consideration, it is not clear what documentation would be updated here. I feel like I'm adding the functionality that was already described.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: https://github.com/openshift/enhancements/commit/8143742c0c8959bdf0a11b457e1941c7316cc394

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,281 @@
...
+- Enable users to alternatively provide pre-provisioned resources to enable
+  minimal permission requirements in the Host Project.
+
+### Non-Goals
+
+- Provisioning and/or configuring the VPC, Network, and Subnetworks are to be
+  done manually by the user prior to running the installer.
+- Provisioning and/or configuring the Service Accounts for the Installer and
+  Cluster are to be done manually by the user prior to running the installer.
+- Privoding solution(s) for manually managing DNS in the Host Project is to be
```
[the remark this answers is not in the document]
jstuever: https://github.com/openshift/enhancements/compare/8143742c0c8959bdf0a11b457e1941c7316cc394..d4de735127bc4057e4bc6c76ff0f7c30b9909110

enhancements/installer/gcp-ipi-shared-vpc.md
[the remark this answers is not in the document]
jstuever: I've updated to specify that no functional changes are required. Hopefully this leaves room for documentation to be modified without specifying code changes.

https://github.com/openshift/enhancements/compare/e693f231bd26efcafbc85c77b54c1738d6966be2..8143742c0c8959bdf0a11b457e1941c7316cc394#diff-68fd28c15e2ae2952a3e25ef489ad068b1d2312baf7ff6f28ccf522099025d9aR106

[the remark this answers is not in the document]
JoelSpeed: OK LGTM

JoelSpeed:
From an API perspective this looks good, deferring remaining reviews/approvals to the teams affected by this enhancement, pretty much just installer right?

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,282 @@
...
+The ability to use pre-existing (aka external) DNS solutions is outside the
+scope of this enhancement. There are other efforts to provide this capability.
+
+New install config parameters will be added to specify the public and/or
+private DNS zones to use when managing cluster records. The default will be an
+empty value, which will result in the private zone and public/private records
+being managed in the service account as before.
+
+In order for the DNS Zones and Records to be managed by the installer, the
+service account will require the following access in the in the project(s)
```
miheer, line 219: Do you mean the ingress operator will need to have sufficient permissions as it will be the one who will fetch the values from the install-config yaml and create the resources for you in the Host Project and gcloud ? or will the installer create the dns zones and records using CCO in the host project ? cc @Miciah

miheer (changes requested):

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,282 @@
...
+
+The current DNSZone struct does not appear to be able to store which project
+the zone is in. The ID variable is currently limited to a managedZone ID value.
+The cluster-ingress-operator will be updated to also accept a relative resource
+name in the following format. This will enable the operator to override the
+default project id with the one provided. No functional API changes are
+necessary.
+
+projects/{projectid}/managedZones/{zoneid}
+
```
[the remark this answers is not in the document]
jstuever: The installer will update the `cluster-dns-02-config.yml` (ie DNS:cluster) to use this format for the spec.privateZone and spec.publicZone parameters, only when a projectid is provided that is different from the clusterid. Otherwise it will remain using the current format of just the zoneID.

And yes, the format is to enable specifying a projectID for the zone specified by zoneID. This is inline with the gcp definition for a resource name, specifically a `Relative Resource Name`
https://cloud.google.com/apis/design/resource_names

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,282 @@
...
+tags specified in the defaultMachinePlatform. The default will be an empty
+list, in which case no additional tags will be added.
+
+#### installconfig.platform.gcp.createFirewallRules
+
+A new parameter will be added to the install-config to enable toggling of the
+creation and management of firewall rules. When set to disabled, the installer
+will skip creation of firewall rules. The default is enabled in which firewall
+rules will be created as before. This toggle does not modify the behavior of
+the cluster-ingress-operator, which is to create firewall rules if permissions
```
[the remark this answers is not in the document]
jstuever: Correct, this is for the openshift-installer to handle and out of scope of cluster-ingress-operator.

enhancements/installer/gcp-ipi-shared-vpc.md
jstuever: The installer will create the private zone and add some records to both the private and public zone. It will need sufficient permissions to do so. The cluster-ingress-operator will create some records in both zones and will need permissions to do so. The workflow for this will require passthrough credentials mode to enable the user to provide a service account with pre-configured permissions to do both.

enhancements/installer/gcp-ipi-shared-vpc.md
```diff
@@ -0,0 +1,282 @@
...
+will be an empty value, which will indicate these resources are to be created
+and managed in the cluster project identified by the projectID.
+
+#### installconfig.platform.gcp.publicDNSZone
+
+A new parameter will be added to the install-config to specify the public DNS
+managed zone. This parameter will be used where necessary to validate existing
+DNS resources as well as provision additional public DNS records. The default
+will be an empty value, which will indicate these resources are to be created
+and managed in the cluster project identified by the projectID.
```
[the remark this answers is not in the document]
jstuever: We will be updating the dns.cluster manifest, as described above, which updates the dns object in cluster once applied via the bootstrapping process. I believe this is what you are referring to.

enhancements/installer/gcp-ipi-shared-vpc.md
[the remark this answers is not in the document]
jstuever: An example `cluster-dns-02-config.yaml` is as follows. In the example, you will see that the private zone is using the original syntax of a simple zoneID. This zone is being created by the installer as the `privateDNSZone` parameter was omitted in the `install-config.yaml`. On the other hand, the public zone is using the newer relative resource name syntax. The `publicDNSZone` was specified in the install-config.yaml. with `{ ID: my-public-zone, projectID: myproject }`

```
apiVersion: config.openshift.io/v1
kind: DNS
metadata:
  creationTimestamp: null
  name: cluster
spec:
  baseDomain: cluster.example.com
  privateZone:
    id: cluster-zpqfj-private-zone
  publicZone:
    id: project/myproject/managedZones/my-public-zone
status: {}
```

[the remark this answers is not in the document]
jstuever: An example of how this could be handled by the cluster-ingress-operator can be found in the following PR
https://github.com/openshift/cluster-ingress-operator/pull/827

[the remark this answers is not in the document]
jstuever: The proposed changes in the installer to accomplish this can be found int he following PR
https://github.com/openshift/installer/pull/6288/files#diff-26df287c8be6151b827ee3de6fd24366d8736d40089dafdbf7b403d20de22c57

patrickdillon:
/approve

This LGTM. @miheer are your questions answered re ingress?

PR #1141, opened 2022-06-02
MGMT-10403: SNO + Workers enhancement - adjust upgrade non-goal to be a goal

omertuc:
The non-goal in question is:

```markdown
Deal with expansion of clusters that have been installed before the
implementation of this enhancement (if possible, their expansion may be
addressed by an upgrade followed with documentation outlining the steps
that the user has to take to enable expansion, e.g. patch the default
`IngressController` to have node selector targetting the master pool
rather than the worker pool).
```

Having it be a non-goal seems to have been a mistake, as it's really
easy to solve and would save a lot of headache in terms of complexity
for users upgrading from 4.10 to 4.11 - we wouldn't have to document
the differences between a "true freshly installed 4.11" cluster and
a "cluster upgraded to 4.11", expanding it with workers should work
just the same.

Implementation of this adjustment:

https://github.com/openshift/cluster-ingress-operator/pull/767

### DIFF
--- a/enhancements/single-node/single-node-openshift-with-workers.md
+++ b/enhancements/single-node/single-node-openshift-with-workers.md
@@ -182,13 +182,6 @@ and operators degrade. User intervention is required to fix this.
 - Deal with expansion of the single-node control-plane by adding more
 control-plane nodes
 
-- Deal with expansion of clusters that have been installed before the
-implementation of this enhancement (if possible, their expansion may be
-addressed by an upgrade followed with documentation outlining the steps that
-the user has to take to enable expansion, e.g. patch the default
-`IngressController` to have node selector targetting the master pool rather
-than the worker pool).
-
 - Adjust the baremetal platform to support single control-plane node
 installations. The baremetal platform solves the "floating ingress" issue by
 using virtual IP addresses/keepalived.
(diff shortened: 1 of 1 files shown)


romfreiman:
Not having this enhancement will require adding explicit docs, extra logic and warnings in assisted installer SaaS and CIM UI, and might cause issues to customers that will miss the relevant documentation.

dhellmann:
Is there a Jira ticket for this scope change?

romfreiman:
@omertuc [MGMT-10403](https://issues.redhat.com/browse/MGMT-10403)

LalatenduMohanty (approved):
Looks good to me.

patrickdillon:
LGTM

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -399,9 +392,27 @@ In any other situations, the installer will set `DefaultPlacement` to
...
+the number of nodes in practice, the Cluster Ingress Operator will set the
+value of `DefaultPlacement` to be `ControlPlane`. Otherwise, it would set it to
+be explicitly `Workers`. Doing this means that after upgrading a cluster
+matching the above criteria from an older version that doesn't contain this
+enhancement to a newer version that does, the `DefaultPlacement` field will go
+from being empty (implying `Workers`) to being `ControlPlane`. This in turn, in
+most default configuratoins, will trigger a reconciliation of the default
+ingress Deployment, as the `nodePlacement` in the default `IngressController`
+is empty. This reconcilliation will trigger a rollout of the default Ingress
+pods with node selectors targetting the "master" node label.
```
Miciah, line 415: This should be fine because the OpenShift upgrade will incur a router pod update anyway for the new container image, if nothing else.

Miciah (commented):
The proposal looks good to me.

enhancements/single-node/single-node-openshift-with-workers.md
```diff
@@ -399,9 +392,27 @@ In any other situations, the installer will set `DefaultPlacement` to
...
+Operator would set the value of `DefaultPlacement` by itself, when it's unset.
+When `DefaultPlacement` is unset and the cluster is a none-platform single
+control-plane node cluster, as indicated by its infrastructure topologies and
+the number of nodes in practice, the Cluster Ingress Operator will set the
+value of `DefaultPlacement` to be `ControlPlane`. Otherwise, it would set it to
+be explicitly `Workers`. Doing this means that after upgrading a cluster
+matching the above criteria from an older version that doesn't contain this
+enhancement to a newer version that does, the `DefaultPlacement` field will go
+from being empty (implying `Workers`) to being `ControlPlane`. This in turn, in
+most default configuratoins, will trigger a reconciliation of the default
```
Miciah, line 412:
```suggestion
most default configurations, will trigger a reconciliation of the default
```

enhancements/single-node/single-node-openshift-with-workers.md
omertuc: Yes, it was not my intent to mention it in a negative light.

PR #1149, opened 2022-06-09
 [CFE-508] As a developer I want to update enhancement to support AWS user tag GA

TrilokGeer:
Updated enhancement to replace experimentalPropagateUserTags to propagateUserTags

### DIFF
--- a/enhancements/api-review/custom-tags-aws.md
+++ b/enhancements/api-review/custom-tags-aws.md
@@ -44,39 +44,26 @@ Note: this enhancement is slightly retroactive. Work has already begun on this. 
 Motivations include but are not limited to:
 
 - Allow admin, compliance, and security teams to keep track of assets and objects created by OpenShift,
-   both at install time and during continuous operation (Day 2)
+both at install time and during continuous operation (Day 2)
 - In a Managed OpenShift environment such as Red Hat OpenShift on AWS (ROSA), allow easy differentiation
-   between Red Hat-managed AWS objects and customer-managed objects
+between Red Hat-managed AWS objects and customer-managed objects
 - *Allow for the restriction of permissions granted to Red Hat in an AWS account by AWS resource tags.
-   For example, see https://issues.redhat.com/browse/SDE-1146 - "IAM users and roles can only operate on resources with specific tags"*
+For example, see https://issues.redhat.com/browse/SDE-1146 - "IAM users and roles can only operate on resources with specific tags"*
 
 ### Goals
 
 - The administrator or service (in the case of Managed OpenShift) installing OpenShift can pass an arbitrary
-   list of user-defined tags to the OpenShift Installer, and everything created by the installer and all other
-   bootstrapped components will apply those tags to all resources created in AWS, for the life of the cluster, and where supported by AWS.
-- User-defined tags can be applied at creation time, as an atomic operation.
-- User-defined tags can be updated post creation.
+list of user-defined tags to the OpenShift Installer, and everything created by the installer and all other
+bootstrapped components will apply those tags to all resources created in AWS, for the life of the cluster, and where supported by AWS.
+- Tags must be applied at creation time, in an atomic operation. It isn't acceptable to create an object and,
+after some period of time, apply the tags post-creation.
 
 ### Non-Goals
 
-- To reduce initial scope, the proposal does not apply for clouds other than AWS. There will be no actions taken to prohibit that later.
-
-### Limitations
-
-- User-defined tags for PV volumes cannot be updated but only set during creation of volumes using `--extra-tags`. Present CSI spec does not support `UpdateVolume` to update volume metadata.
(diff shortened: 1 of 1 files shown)


enhancements/api-review/custom-tags-aws.md
```diff
@@ -101,7 +88,73 @@ The userTags field is intended to be set at install time and is considered immut
 
 If the userTags field is changed post-install, there is no guarantee about how an in-cluster operator will respond to the change. Some operators may reconcile the change and change tags on the AWS resource. Some operators may ignore the change. However, if tags are removed from userTags, the tag will not be removed from the AWS resource.
 
-#### Existing drawbacks
+### Update design details
+
+The `experimentalPropagateUserTags` field is changed to `propagateUserTags` for GA version.
```
bparees, line 93: not clear if it's implied but we should have a few releases where we keep supporting the old field name too, before we drop it entirely, so consumers have a change to change their scripts/templates/configs.

this should also be covered in the migration section.

Technically we are under no obligation to provide such a transition since it was an experimental field, but it's basically no cost to us and avoids operations teams from being surprised if we stop respecting the experimental flag and they don't get the tag propagation behavior they expected, suddenly.

TrilokGeer: For the new field `propagateUserTags` , Can the flag used to propagate user flags `propagateUserTags` be removed and made a default propagation of user-defined tags ? Is there a scenario where a switch like `propagateUserTags` is useful ?

enhancements/api-review/custom-tags-aws.md
```diff
@@ -88,9 +88,10 @@ The userTags field is intended to be set at install time and is considered immut
 
 If the userTags field is changed post-install, there is no guarantee about how an in-cluster operator will respond to the change. Some operators may reconcile the change and change tags on the AWS resource. Some operators may ignore the change. However, if tags are removed from userTags, the tag will not be removed from the AWS resource.
 
-### Update design details
+### Updated design details
 
-The `experimentalPropagateUserTags` field is changed to `propagateUserTags` for GA version.
+A new field `propagateUserTags` is added in GA version. The `experimentalPropagateUserTags` field will be deprecated in future release versions.
+When both fields are set, `experimentalPropagateUserTags` takes precedence.
```
bparees, line 94: we'd deprecate it now(when the ga field is added). we'd remove it in a future release.

TrilokGeer: Can the flag used to propagate user flags `propagateUserTags` be removed and made a default propagation of user-defined tags ? Are there usecases that might require a switch like propagateUserTags . As similar switch needs to be considered for other cloud providers.

bparees: if i understand your suggestion (default to just always propagate usertags when usertags are specified in the install config?) that feels like a breaking behavior change and also i'm not seeing the benefit of taking the choice away from admins.

TrilokGeer: Thanks, the breaking behavior change perspective makes sense.

TrilokGeer:
@bparees @patrickdillon @sdodson @jerpeter1 , Hey folks,  can you please help with a final read and approvals, if ok?

bparees:
/lgtm

I will wait for feedback/consensus from some of your other reviewers before approving.

enhancements/api-review/custom-tags-aws.md
```diff
@@ -110,52 +176,97 @@ If a customer decides they do not want these tags applied to new resources, ther
...
+- In a Managed OpenShift environment such as Red Hat OpenShift on AWS (ROSA), allow easy differentiation
+between Red Hat-managed AWS objects and customer-managed objects
+- *Allow for the restriction of permissions granted to Red Hat in an AWS account by AWS resource tags.
+For example, see https://issues.redhat.com/browse/SDE-1146 - "IAM users and roles can only operate on resources with specific tags"*
+
+### Identified goal
+
+- The administrator or service (in the case of Managed OpenShift) installing OpenShift can pass an arbitrary
+list of user-defined tags to the OpenShift Installer, and everything created by the installer and all other
+bootstrapped components will apply those tags to all resources created in AWS, for the life of the cluster, and where supported by AWS.
```
sdodson, line 198: Later on we say that this arbitrary list is limited, at least currently, to 10 items. I wonder if that should be mentioned here or if that's just the perils of not reading the entire document.

sdodson:
/lgtm
I'm fine with it as is, feel free to address my comment about clarity or not.

patrickdillon:
This also
/lgtm

enhancements/api-review/custom-tags-aws.md
patrickdillon: `experimentalPropagateUserTags` is a boolean field in the install config--not a boolean pointer. Therefore it is impossible for `experimentalPropagateUserTags` to not be set, right? There is no difference between not setting the field, and setting the field to `false`.

patrickdillon: D'oh there is a whole section on converting this to a pointer. My mistake! re /lgtm

PR #1152, opened 2022-06-10
Add enhancement change for disabling SC creation

gnufied:
Allow disabling storageclass creation

### DIFF
--- a/enhancements/storage/csi-driver-install.md
+++ b/enhancements/storage/csi-driver-install.md
@@ -176,6 +176,30 @@ Upgrade from 4.5 with Manila CSI driver (GA in 4.5) works as follows:
 In case there was no AWS / Manila CSI driver operator running during the update, CSO installs the corresponding operator
 as during installation, so user has a CSI driver running after update.
 
+### Disabling creation of storageclass
+Starting with OCP 4.12, we are allowing users to disable creation of StorageClass in case they need CSI driver but don't
+necessarily need dynamic provisioning in the cluster.
+
+A cluster admin can edit and set `disableStorageClass` field to `true` in `ClusterCSIDriver` object of that driver to disable
+creation of storageclass. This can be done as day-2 operation which will result in deletion of any previously created
+storageclasses by the CSI operator.
+
+```golang
+type ClusterCSIDriverSpec struct {
+    OperatorSpec `json:",inline"`
+    // DisableStorageClass determines if creation of storageclass for this driver
+    // should be skipped. By default Openshift creates storageclass for every CSI
+    // driver it installs via CSI operator. Setting this value to true will skip
+    // automatic creation of storageclass for driver being installed.
+    // Defaults to false.
+    // +kubebuilder:default=false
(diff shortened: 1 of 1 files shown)


enhancements/storage/csi-driver-install.md
```diff
@@ -176,6 +176,27 @@ Upgrade from 4.5 with Manila CSI driver (GA in 4.5) works as follows:
 In case there was no AWS / Manila CSI driver operator running during the update, CSO installs the corresponding operator
 as during installation, so user has a CSI driver running after update.
 
+### Disabling creation of storageclass
+Starting with OCP 4.12, we are allowing users to disable creation of StorageClass in case they need CSI driver but don't
+necessarily need dynamic provisioning in the cluster.
+
+An user can edit and set `disableStorageClass` field to `true` in `ClusterCSIDriver` object of that driver to disable
```
dobsonj, line 183: Tiny nit: should be `A user`, but +1 to these updates in general.

enhancements/storage/csi-driver-install.md
```diff
@@ -176,6 +176,30 @@ Upgrade from 4.5 with Manila CSI driver (GA in 4.5) works as follows:
...
+```golang
+type ClusterCSIDriverSpec struct {
+    OperatorSpec `json:",inline"`
+    // DisableStorageClass determines if creation of storageclass for this driver
+    // should be skipped. By default Openshift creates storageclass for every CSI
+    // driver it installs via CSI operator. Setting this value to true will skip
+    // automatic creation of storageclass for driver being installed.
+    // Defaults to false.
+    // +kubebuilder:default=false
+    DisableStorageClass bool `json:"disableStorageClass,omitempty"`
```
jsafrane, line 196: Not sure if `DisableStorageClassCreation` is better.

enhancements/storage/csi-driver-install.md
```diff
@@ -176,6 +176,30 @@ Upgrade from 4.5 with Manila CSI driver (GA in 4.5) works as follows:
 In case there was no AWS / Manila CSI driver operator running during the update, CSO installs the corresponding operator
 as during installation, so user has a CSI driver running after update.
 
+### Disabling creation of storageclass
+Starting with OCP 4.12, we are allowing users to disable creation of StorageClass in case they need CSI driver but don't
+necessarily need dynamic provisioning in the cluster.
```
gcharot, line 181: I think we should also add the use case where admins don't like how we configure the storage class and want to use their own instead.

enhancements/storage/csi-driver-install.md
gcharot: +1 I think this is more intuitive

gnufied: As a user editable option - the longer the field name is,  the higher the chance is that they will make a typo and we will end up with a support case that it does not work as expected.

enhancements/storage/csi-driver-install.md
gnufied: added

PR #1166, opened 2022-07-04
Add enhacement for hypershift networking

enxebre:
This is to capture info in existing docs in a single public source of truth for HyperShift networking state of the world. Any further development should be captured and discussed here.
Follow up:
- MCE integration.
- Monitoring.
- Cert management.


csrwng:
@enxebre this is awesome. Please add an OWNERS file under the hypershift directory.

enxebre:
/approve

Approving myself on behalf of @csrwng https://github.com/openshift/enhancements/pull/1169.

Let's set this doc as the foundation and go from here.

